PR #19, opened 2016-01-28
ExamplePet and ExampleLightPet now disappear upon buff removal

gorateron:
ExamplePet and ExampleLightPet now disappear upon buff removal, for example when right clicking the buff to remove it manually.

### DIFF
--- a/ExampleMod/Projectiles/Pets/ExampleLightPet.cs
+++ b/ExampleMod/Projectiles/Pets/ExampleLightPet.cs
@@ -37,7 +37,7 @@ namespace ExampleMod.Projectiles.Pets
 		{
 			Player player = Main.player[projectile.owner];
 			ExamplePlayer modPlayer = (ExamplePlayer)player.GetModPlayer(mod, "ExamplePlayer");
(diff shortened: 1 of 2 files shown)


bluemagic123:
Wait, did they not already disappear on buff removal?

gorateron:
Nowhere in the AI of the pets are you checking if the player actually has the buff or not, meaning if the pet is spawned and the buff removed the pet will stay. This way it won't. Tested it myself with a custom pet, which was literally an ExampleMod copy with this addition to it.

JavidPack:
It works as it is, I just tested it, are you sure you didn't forget the ExamplePlayer portion in your own mod?

ExamplePlayer resets the booleans to false. If the buff is active, it will set the boolean to true. Then the projectile will die or live if projectile.timeLeft expires or not based on the booleans.

I specifically modeled them after the code in the source code so there would be no surprises.

gorateron:
Yup you're right, I was forgetting to reset the boolean in ResetEffects().. It works properly now. Can close this pull request

PR #26, opened 2016-05-17
Added TargetClosest hooks!

jasonjmcghee:


JavidPack:
Can you tell me a use case for this? I don't understand what this accomplishes.

jasonjmcghee:
Sure. Right now TargetClosest set's an NPCs target to be the position of the player. If someone wanted to extend the AI somehow (I've been working with it a lot lately), they would need to change the targeting. For example, you could have it use instantaneous position based on the player's velocity, so the npc could shoot a projectile that would hit the player based on it's current trajectory, or even target a projectile the player fired.

jasonjmcghee:
I could come up with many more examples, all relating to modifying vanilla NPC targeting.

JavidPack:
But, the target is just the index of the player in player array, not a position/Vector2

jasonjmcghee:
That's true. So you'd have to create a temporary mock player containing the minimum information necessary for being a target

jasonjmcghee:
The other way to do this is to create a getter for NPC.target and replace every reference to it and add the option to override it

Chicken-Bones:
I'm not in favor of temporary mock players, this is the place to implement proper hooks.

PS. looks like something went wrong with the patch offsets

bluemagic123:
I could see usage by just overriding what npc.target gets set to. But are three hooks really needed for this?

PR #77, opened 2016-12-24
Update App.config and Terraria.csproj (Resolves #65)

### LINKED ISSUE #65: Build works only when launched from VS

Murfalo:


JavidPack:
@Murfalo I saw you put the start program and directory in your Terraria.csproj, but if you run setup.bat, it'll automatically setup Terraria.csproj.user with that information, and it is the more logical place for this sort of information and is automatically populated correctly.

Why don't you update to the latest master branch here and run setup.bat and the "Regenerate Sources" button and see if you have any issues with that how it is.

Murfalo:
@JavidPack I added it directly because for some reason the WindowsRelease config was never being populated with the correct fields.  Let me double check really fast!

Chicken-Bones:
it would definitely be possible to add the debug launch parameters to the windows release build configuration. However I decided not to.
The WindowsDebug configuration is there for the explicit purpose of debugging. The release version has optimisations enabled which means it doesn't work as well with debugging. Yes if you launch it right, you can still put breakpoints, but why would you?

tldr, the release configuration is meant to be run directly.
I'll need a reason why you want to launch the release configuration from VS and not the debug one.

Murfalo:
It's just a small convenience and would help people seen in #65

Murfalo:
Oh dear, I rebased and it changed the pull request, whoops!  I'm just gonna go ahead and close this because it wasn't going to be merged and the request is no longer accurate.

PR #79, opened 2016-12-24
Mod.ChatInput broadcast option (Resolves #61)

### LINKED ISSUE #61: Mod.ChatInput -- Allow option to not send message to others
Change 
ChatInput(string text)
to something like
ChatInput(string text, ref bool broadcast)
so modders can decide to hide chat commands from chat.

Murfalo:


Chicken-Bones:
I'd actually noticed this myself, personally I would've been going for a return value, where returning true indicates the input has been handled "captured" but this'll do

PR #85, opened 2016-12-28
Refactor command system

Murfalo:
Lots of changes in this one!  I've refactored the command system in a way that I think makes it much more extensible and somewhat easier to use.  I also think it aligns better with the tModLoader style.

If everything looks good on this, I'm hoping to move forward and implement server commands and some utility functions such as the /help command.


patches/tModLoader/Terraria.ModLoader/CommandManager.cs
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		internal static string[] ParseCommand(string text) => text.Substring(1).Split(' ');
+
+		internal static void PrintUsage(ModCommand mc) => Main.NewText("Usage: " + mc.Usage);
+
+		internal static void ProcessInput(string input, CommandType type, ref bool show)
+		{
+			var cmd = ParseCommand(input);
+			var cmdList = GetCommandList(type, cmd[0]);
```
Chicken-Bones: throw in a .ToLower()

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
```diff
@@ -0,0 +1,69 @@
...
+			if (cmdList == null) return;
+			var args = cmd.Skip(1).ToArray();
+			var mc = cmdList[0];
+
+			show = mc.Show;
+			if (cmdList.Count > 1)
+				Main.NewText("Error: Multiple definitions of command /" + mc.Command);
```
Chicken-Bones: should show be overwritten here?

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
```diff
@@ -0,0 +1,69 @@
...
+			else if (!mc.VerifyArguments(args))
+				PrintUsage(mc);
```
Chicken-Bones: should show be overwritten here?

ExampleMod/build.txt
```diff
@@ -1,5 +1,6 @@
 author = bluemagic123
 version = 0.9.1
+languageVersion = 6
```
Chicken-Bones: @JavidPack any problems with this?

JavidPack: Well, we've been avoiding using c# 6 code in ExampleMod, not sure if it's a conscious decision or not, but new syntax could be confusing to new modders.

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
Murfalo: I figured if you were using a command that was designed to be hidden from chat you probably didn't want everyone seeing it if you messed up the syntax or something.  Also if there's multiple definitions I don't think it really matters either way, but if you want to change it that's fine with me.

JavidPack:
Also, we were discussing earlier how multiple commands with the same name shouldn't thrown an error, but rather require the user to clarify which command from which mod to use with something like modname1:commandname modname2:commandname. 

We could just pull this in and then change stuff though.

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
Murfalo: I just changed the dictionaries to be case insensitive is that all right?

Murfalo:
@JavidPack I agree! I didn't really know what should be the expected behavior for commands with the same name so I just had it throw an error for now.

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
Chicken-Bones: Yep

patches/tModLoader/Terraria.ModLoader/CommandManager.cs
Chicken-Bones: Acceptable

ExampleMod/build.txt
Chicken-Bones: I'll agree, as much as I personally love C# 6, and use it in tModLoader itself, we'll use the old way in ExampleMod. I'll request you just use the longhand property overrides

Chicken-Bones:
We can add the conflict resolution. I'll merge this pending reversion to C# 4.0

Murfalo:
@Chicken-Bones Just finished making the revisions you asked for!

gorateron:
Very nice changes! While you're at it, have you thought about automated argument parsing ?

Chicken-Bones:
Take a look at https://github.com/bluemagic123/tModLoader/commit/84d5d4b64653115271eec9b5cbaeeb62c0659c85

PR #94, opened 2017-01-07
SpriteFontX class and DrawString Extensions

CXUtk:


Chicken-Bones:
After some discussion, we believe ExampleMod is not the place for this. The primary purpose of tModLoader is to provide modifications to Terraria's base code that enable the creation of compatible mods. We understand that you have added it to ExampleMod and not tModLoader, however we don't believe there is enough interest to warrant adding it, and would like to keep the purpose of ExampleMod clean and simple.

We suggest just adding this personally to any mods you're making that require it, or putting it in a shared dll or library mod. If there is enough interest we may consider adding it to tModLoader as a core utility for everyone at a later date

PR #95, opened 2017-01-08
Added example laser beam weapon

CXUtk:


Chicken-Bones:
Looks good to me

JavidPack:
The item namespace and PNG location don't match up, the cs file and PNG file should be moved to the weapons folder so it actually loads. Did you test on mp?

CXUtk:
I think they are match, one is weapon and one is projectile. I actually tested them.

JavidPack:
What I mean is, the ExampleLaserWeapon ModItem is in ExampleMod.Items.Weapons namespace, but in the ExampleMod/Items/ folder. Autoloading the textures will fail since the namespace and folder of the png don't match.

PR #103, opened 2017-01-12
TagSerializable - allows easy serialization of classes into TagCompound

Shockah:


patches/tModLoader/Terraria.ModLoader.IO/TagSerializable.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Reflection;
+
+namespace Terraria.ModLoader.IO
+{
+	public interface TagSerializable
+	{
+		TagCompound SerializeData();
+	}
+
+	public interface ITagDeserializer<out T> where T : TagSerializable
```
Chicken-Bones: No-one's going to override this, just use TagDeserializer, so why don't we just use Func<TagCompound, T> for the DESERIALIZER field

patches/tModLoader/Terraria.ModLoader.IO/TagCompound.cs
```diff
@@ -73,6 +81,11 @@ public bool GetBool(string key)
 			return o as double? ?? o as float? ?? 0;
 		}
 
+		public T GetSerializable<T>(string key) where T : TagSerializable
+		{
+			return TagSerializables.Deserialize<T>(GetCompound(key));
+		}
```
Chicken-Bones: I will augment this with list compatibility.

patches/tModLoader/Terraria/Item_tML.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Terraria.ModLoader.IO;
+
+namespace Terraria
+{
+	public partial class Item : TagSerializable
```
Chicken-Bones: This is good practice, we should probably employ it for other classes @JavidPack @bluemagic123

patches/tModLoader/Terraria.ModLoader.IO/TagCompound.cs
Shockah: Yeah, I kinda forgot to do so myself. I can do it after work today, unless you want to do so yourself.

patches/tModLoader/Terraria/Item_tML.cs
JavidPack: k, I did some Google-ing and convinced myself it's an ok idea. I'm not sure how I feel about the _tML naming scheme though, but maybe it's the best idea for now.

patches/tModLoader/Terraria.ModLoader.IO/TagCompound.cs
Chicken-Bones: I'm happy to do it

patches/tModLoader/Terraria.ModLoader.IO/TagSerializable.cs
```diff
@@ -0,0 +1,86 @@
...
+			FieldInfo deserializerField = type.GetField("DESERIALIZER");
+			if (deserializerField == null)
+				throw new Exception(string.Format("Missing deserializer for type {0}.", type.FullName));
+
+			ITagDeserializer<T> deserializer = (ITagDeserializer<T>)deserializerField.GetValue(null);
+			TagCompound dataTag = tag.HasTag("data") ? tag.GetCompound("data") : null;
+			return deserializer.Deserialize(dataTag);
+		}
+
+		private static Type GetType(string name)
```
Chicken-Bones: I'd suggest caching this in a dict and resetting it on reload. Calling GetType on up to every mod for every tag may not be the most performant idea

patches/tModLoader/Terraria/Item_tML.cs
Shockah: To be honest I don't like the "_tML" suffix either, but I couldn't come up with anything better. Could possibly split it out into another folder and name it just "ClassName.cs", but I'm not sure whether I like the idea of partial classes being completely separated in different folders.

patches/tModLoader/Terraria.ModLoader.IO/TagSerializable.cs
Shockah: I'm not sure how costly GetType on each assembly would be (it doesn't sound like it would be THAT bad), but I agree it would be better. Need to make sure you're clearing the dictionary on mod reload though.

patches/tModLoader/Terraria.ModLoader.IO/TagSerializable.cs
Shockah: Thinking about it now I agree, I only made it like this because I was kind of "inspired" by how Android has it done with its Parcel class and Parcelable interface.

Chicken-Bones: Understandable, but Java lacks (or lacked) functional constructs

patches/tModLoader/Terraria/Item_tML.cs
JavidPack: No, I definitely don't think we should break the namespace and folder structure duality we have so far.

Chicken-Bones:
It occurs to me that saving the type and data in a wrapped tag isn't necessary, nor particularly desirable. C# static typing requires modders to provide the desired type when deserialising. This would allow for changes in type/namespace in mod code while retaining data compatiblity, as well as allow the future addition of classed based deserialisers to existing data tags.
Additionally item deserialisation to be compatible with the existing ItemIO.Save

Shockah:
I guess that's fine, except this way you won't know what should the data be deserialized into by just looking at the TagCompound. But yeah, I guess that's fine-ish.

JavidPack:
So, is this ready?

Shockah:
@JavidPack I just finished porting my mod into this new system and am trying to test, but CheatSheet doesn't work on this version of tML (from the tagserializable branch), so it's a bit of a pain.

Shockah:
Seems like it doesn't handle subclasses properly. For example, trying to save List<Affix> (where Affix is abstract) would end up in tML to find a DESERIALIZER in the Affix class, instead of the subclasses being saved - which ends up in an exception being thrown.

Shockah:
Here is the said mod that crashes: https://github.com/Shockah/tModLoader-Mods/tree/ItemAffix/wip (Shockah.ItemAffix, but it requires Shockah.Utils to be present too).

PR #109, opened 2017-01-19
Add PlaceInWorld and PostDrawTiles hooks.

FenixFyreX:
I've added two new hooks:
`PlaceInWorld`, a hook/callback for placing tiles into the world which passes the tile coordinates and Item stack used.
Additionally, I've added a `PostDrawTiles` hook for `ModWorld` which would allow users to draw their own systems similar to Terraria's wiring system without drawing above of items, npcs, and UIs.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalTile.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalTile.cs
@@ -178,5 +178,9 @@ namespace Terraria.ModLoader
 		{
 			return -1;
 		}
+
+		public virtual void PlaceInWorld(int i, int j, Item item)
+		{
+		}
 	}
 }
--- a/patches/tModLoader/Terraria.ModLoader/GlobalWall.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalWall.cs
@@ -1,4 +1,5 @@
 using Microsoft.Xna.Framework.Graphics;
+using Terraria;
 
 namespace Terraria.ModLoader
 {
@@ -67,6 +68,10 @@ namespace Terraria.ModLoader
 		}
 
 		public virtual void PostDraw(int i, int j, int type, SpriteBatch spriteBatch)
+		{
+		}
+
+		public virtual void PlaceInWorld(int i, int j, Item item)
 		{
 		}
 	}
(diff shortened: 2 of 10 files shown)


JavidPack:
Did you test these? The post draw tiles being right after spritebatch.close worries me

FenixFyreX:
Yes, these modifications are completely tested and should work fine. They do for me. The PostDrawTiles hook is not for the tiles themselves, but for starting a new spritebatch routine (Begin/End) for things like drawing layers between the map tiles, and the rest (items, npcs, ui, etc).

As for the PlaceInWorld hook, I could find no better alternative for saving the last used item without finding and replacing every instance of deincrementing a stack's `stack` value, which would be an entirely different ball game.

Chicken-Bones:
Firstly, please use tabs instead of spaces for indenting. You can see the discrepancy in the Files Changed tab.

I took a look at the difficulty of stack decrement and think you should take a closer look at the specific placement logic. Assuming that all placement code is done in `Player.PlaceThing`, I only get 7 stack decrements. Most of which are for painting and auto painting. One of which is for auto actuators, and the other couple seem to be valid wall/tile placement. 

I'll also note that you're relying on consume item to be true. I'd be looking around `flag32 && ItemLoader.ConsumeItem(item, this)` for this

FenixFyreX:
I changed every config option related to tabs in VS to not replace tabs with spaces. For some reason, this was the default (and I'm new to git, I'm sorry I didn't catch the whitespace). I think in `Player.cs` and `Main.cs` there are still start of line whitespace issues, but no matter what I tried, I could not avoid them? I recloned the repository, reset my fork, and rewrote everything, respecting the formatting every possible way, and checked the source with a hex editor to make sure it was, in fact, tabs.

As for the `PlaceInWorld` hooks, I've added support for walls, and under your recommendation, Chicken-Bones, I've corrected the callback code in `Player.cs` and got rid of the four uneccessary fields. As for the `PostDrawTiles` hook, are there any suggestions for a better name? I completely understand why JavidPack was thrown off, so suggestions are welcome.

Chicken-Bones:
Yep your tabs came through. assuming you've tested this, it's approved. If there's any edge cases we can fix them as they come

PR #110, opened 2017-01-19
Added comments to several example object

CXUtk:
I added some comments for guiding the beginners.


gorateron:
Seems good! Ah, jopo merged it :)

PR #116, opened 2017-01-29
Added a Markdown file to document build.txt

raydeejay:


Chicken-Bones:
hmm, do we not have enough on the documentation wiki? The problem with the markdown is that it's only really viewable in github, and few people use the repo

raydeejay:
The idea is to move the documentation which is now only in the wiki next to the code it documents, then generate HTML documentation from that

PR #121, opened 2017-02-04
Fixed the bug when player is killed by mod projectile

CXUtk:
I have noticed that when player is killed by mod projectile, it actually throw an IndexOutOfRange error which leads to no death text would be shown. I'd like to fix this bug by adding the hook for `Main.projName`.


JavidPack:
Thanks.

PR #130, opened 2017-02-26
Added multiplayer support for Example Laser

CXUtk:


Jofairden:
Uhm what is the issue? I'm pretty sure I fixed the laser for MP **[here](https://github.com/bluemagic123/tModLoader/commit/d2c0bfe88755769a392c024bd6b7be5506f3c9a2)**

ExampleMod/Projectiles/ExampleLaser.cs
```diff
@@ -111,21 +119,23 @@ public override void AI()
...
-				#endregion
+                projectile.velocity = diff;
+                projectile.netUpdate = true;
 			}
-
-
-			#region Charging process
-			// Kill the projectile if the player stops channeling
-			if (!player.channel)
+            projectile.position = player.Center + projectile.velocity * MOVE_DISTANCE;
```
Jofairden: Did you take this code out of the **if (projectile.owner == Main.myPlayer)** statement? That's exactly what should make it work properly in multiplayer, you only want the projectile pos and player rotation to update if the current player is the owner.

CXUtk: Yes, it should be take out of the statement. The `player` now is the projectile's owner. The key of my code is the` projectile.netUpdate = true;` inside of `if (projectile.owner == Main.myPlayer)`. Which updates the projectiles velocity(which is the unit vector of player's center to the local mouse position). In this way, the laser can shoot to the correct position.

Jofairden:
@CXUtk, I reviewed your changes and can confirm the code works in multiplayer, however, the drawing code and the projectile itself are botched. I revise you to review your code and make changes to that it works properly again.

Jofairden:
Code tested, works in multiplayer correctly. Merging.

Chicken-Bones:
@CXUtk In the off chance you're still subscribed to this thread, I'm just looking to get in touch for your approval on our license change, Can you drop on Discord or have a look at #705

Thanks

PR #151, opened 2017-03-27
Add object overload for ErrorLogger.Log

Trivaxy:
This overload allows the user to log objects, it simply retrieves methods, constructors, fields and properties in them.
(Fields and properties have their name AND value get retrieved)

I tested this and it should work, tell me if you get any errors/exceptions

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/ErrorLogger.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ErrorLogger.cs
@@ -222,31 +222,31 @@ namespace Terraria.ModLoader
 		{
 			Directory.CreateDirectory(LogPath);
 			using (StreamWriter writer = File.AppendText(LogPath + Path.DirectorySeparatorChar + "Logs.txt"))
-                        {
-                             writer.WriteLine("Object type: " + param.GetType());
-                             foreach (PropertyInfo property in param.GetType().GetProperties())
-                             {
-                                 writer.Write("PROPERTY " + property.Name + " = " + property.GetValue(param, null) + "\n");
-                             }
+			{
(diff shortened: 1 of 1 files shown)


Jofairden:
Tabs are wrong. Change them 

Also, can you show some sample output with usage?

Trivaxy:
Ok, fixed tabs.
Take note this does not retrieve private members, if you want me to make it retrieve private members just tell me.

Here's an example:
`public class TestObject
    {
	private int testIntProperty { get; set; }
        public TestObject testObject { get; set; }

        public bool no;

        public int Add(int i, int b, float c)
        {
            return (int)(i + b + c);
        } //also useless, but for sake of examples

        public TestObject(int a = 0) //useless constructor, for sake of example
        {
            Console.WriteLine("hi");
	    testIntProperty = 90;
        }

        public TestObject(bool b)
        {
            Console.WriteLine("hi");
        }
    }`

And then doing `ErrorLogger.Log(new TestObject());` (obviously the constructor i marked useless is called) produces a simple object output (assuming it's an empty file) like this:
reflectionwhy

Jofairden:
Everything seems good to me, but, just for verifiability, @Chicken-Bones / @JavidPack , everything ok?

JavidPack:
It seems to me that the more natural way to do this would be for objects to override the ToString method. If modders want this additional reflection thing, I think think that should be manually specified, not the default.

Maybe `ErrorLogger.Log(object param, bool alternateOutput = false)` so the typical use case of using the ToString method is the default, but if the modder additionally supplies true, it switches to reflection output.

Trivaxy:
Sounds good as well, once I get home I'll add these changes.

JavidPack:
Thanks.

PR #179, opened 2017-05-29
Cleanup the Linux installer

513G3:
Newer installations of Steam organize the installation
differently.  Now the Linux installer handles 3 scenarios:

1. XDG_DATA_HOME is set (unaffected by this patch)
2. HOME is set and Steam was installed with "old" paths
2. HOME is set and Steam was installed with "new" paths

### Description of the Change
The tModLoader Linux installer now handles 3 scenarios:

1. XDG_DATA_HOME is set (unaffected by this patch)
2. HOME is set and Steam was installed with "old" paths
2. HOME is set and Steam was installed with "new" paths

### Alternate designs
I considered more verbose code with comments, but felt that would not adhere to the style of this file.

Without this patch, users with newer versions of Steam will be unable to easily install tModLoader.

### Benefits
Easier installation for novice Linux users.

### Possible drawbacks
None that I can see.

### Applicable Issues
The XDG_DATA_HOME code path was untested and unchanged.

### Sample Usage
N/A

### DIFF
--- a/installer2/LinuxInfo.java
+++ b/installer2/LinuxInfo.java
@@ -40,16 +40,30 @@ public class LinuxInfo
 
     private static File getInstallDir(String homeDir, boolean normalHome)
     {
-        File file = new File(homeDir);
+        File installDir = new File(homeDir);
+        File newSteamDir;
         if (normalHome)
         {
-            file = new File(file, ".local");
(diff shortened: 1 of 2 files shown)


513G3:
Cleaner solution forthcoming...

PR #181, opened 2017-05-29
Added ModContainer

Itorius:
### Description of the Change

Adds the ModContainer class (which inherits ModTileEntity) which allows the modder to create containers.

### Alternate designs

I considered making an interface for it, but those aren't commonly spread in Terraria/tML.

The fact that this is an united system, allows for easy manipulation with the inventories (such as searching for items/making mods that function like Magic Storage but has IO)

### Benefits

Relatively easy way to add containers, saving/loading is supported without the need to do it yourself, containers automatically drop items upon being broken (optional), ModContainer class contains Send/Receive methods for syncing inventory.

### Possible drawbacks

I don't see any of this code breaking already existing systems, I had to redo some data saving for ModTileEntity, but I have tested it and it both keeps the data (Save method) and the storage data.

### Applicable Issues

Unsure if the networking works.

### Sample Usage

I plan on including an example container in the ExampleMod (sometimes soon when I have time)

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Default/MysteryTileEntity.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Default/MysteryTileEntity.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using Terraria.DataStructures;
 using Terraria.ModLoader.IO;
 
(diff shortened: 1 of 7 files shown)


Chicken-Bones:
The proposed implementation seems unnecessarily invasive. We suggest you work with one of us on the discord to create something that satisfies your needs, but doesn't require modification of the tModLoader API. If this becomes a thing that multiple modders want to make use of, and it would be more productive to have in tModLoader itself, we'll revise things at a later date.

PR #182, opened 2017-05-29
Account for newer Steam path in Linux installer

513G3:
### Description of the Change
Newer Steam installations under Linux change the installation path.  The tModLoader installer can now handle the new path (in addition to the two previous paths already handled).

### Alternate designs
At first, I made a patchset (now revoked) that mimic'd the file's format more closely.  I feel that adding support for the new path in the previous format made the code unwieldy.  I feel that this solution is superior.

Users with newer Steam installations should be able to install tModLoader as easily as possible.

### Benefits
This patch makes tModLoader easier to use for more people.

### Possible drawbacks
None that I can see.

### Applicable Issues
I did not test the XDG_DATA_HOME variant as I am unfamiliar with that setup.  But I am confident that the code flow for that variant remains unchanged.

### Sample Usage
N/A

### DIFF
--- a/installer2/LinuxInfo.java
+++ b/installer2/LinuxInfo.java
@@ -4,20 +4,6 @@ public class LinuxInfo
 {
     public static void main(String[] args)
     {
-        String path = System.getenv("XDG_DATA_HOME");
-        File directory = null;
-        if (path != null)
-        {
-            directory = getInstallDir(path, false);
-        }
-        if (directory == null || !directory.exists())
-        {
-            path = System.getenv("HOME");
-            if (path != null)
-            {
-                directory = getInstallDir(path, true);
-            }
-        }
         String[] files = new String[]
         {
             "Terraria.exe",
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Functionally it looks fine, but I suggest the following variant using another function and return to reduce nesting and improve readability https://pastebin.com/qimB8sX9

513G3:
Hi, thank you for the input; I made the edits you suggested.  I ended up deciding against the throw since the other *Info.java files don't do it... I just return null if nothing matches and let the caller handle that (the caller starts out with a null check).

I'm not savvy on the github interface.  I want to see that my amended commit landed in the right place... but there does not seem to be a timestamp or a history (??).  I do see that the "Files changed (2)" button displays the new set of files correctly.

Thanks!

Chicken-Bones:
Yeah good choice on the throw, I didn't actually look at the installer code, just reworking the control flow and added in lieu of actually knowing what to do.

With regards to git, it seems like you did an amend commit, and then forcefully overwrote your remote repo? The result is your commit has the right content, but wrong timestamp.

In generally it is extremely inadvisable to perform actions that rewrite a repository's history, once that history has been pushed and is no longer local (hence having to add the -force flag when pushing).
Fortunately, it was just an amend commit with no comments or other directly related stuff in the PR, so everything silently lined up.

What's recommended in this case, is that you just create another commit with the new changes, and when we merge it into the repo there's an option to squash everything down so it looks clean.

513G3:
Thanks for the insights.  You were both astute and correct: I force pushed an amended commit.

I'm used to the Gerrit frontend for Git (not GitHub).  I was trying to mimic my familiar workflow where committers are responsible for periodically amending their commits (either with rebases or fixes from comments) so that the committer's job is as easy as possible when he decides to pull the trigger.  I did not realize that it was preferable on GitHub for me to just make a second commit (as opposed to amending the first) and to let you take care of the squash.  Thanks for the patience and for the cool software product!

Chicken-Bones:
Yeah, using a git frontend is perfectly fine, I'm a fan of TortoiseGit myself, and have used a few others in the past. We also advocate rebasing over merging wherever possible, because the linear history is much cleaner in most cases

513G3:
Hmmm... please see attached.  Was this my change???  I sure hope not.  It
worked for me before pushing...

Thanks!

PR #191, opened 2017-06-09
Added Better MP3 Support

Rartrin:
MP3s are no longer converted to wavs and cached, now they only hold the
part of the MP3 that needs to be played but only for music. Normal non-music MP3s are still converted when needing to be played but no caching is done for any of them regardless.


Jofairden:
So why are music files not converted but the rest is? If we don't need conversion, why do all that expensive work?

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3009,7 +3014,7 @@
  						}
  						Microsoft.Xna.Framework.Color color5 = Lighting.GetColor((int)((double)dust.position.X + 4.0) / 16, (int)((double)dust.position.Y + 4.0) / 16);
 -						if (dust.type == 6 || dust.type == 15 || (dust.noLight && dust.type < 86 && dust.type > 91) || (dust.type >= 59 && dust.type <= 64))
```
Jofairden: I'm pretty sure this shouldn't be undone, a recent dust fix.

patches/tModLoader/Terraria.ModLoader/ModSound.cs
```diff
@@ -12,7 +13,7 @@ public class ModSound
 		/// <summary>
 		/// The SoundEffect instance of the sound that this ModSound controls. This is here so you can call CreateInstance on it.
 		/// </summary>
-		public SoundEffect sound
+		public SoundWrapper sound
```
JavidPack: This breaks all mods that call `sound.CreateInstance();` See ExampleMod's Wooo.cs.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -124,12 +125,27 @@
...
+ 		public static SoundEffectInstance[] soundInstanceNPCKilled = new SoundEffectInstance[63];
+ 		public static SoundEffectInstance soundInstanceMoonlordCry;
+ 		public static SoundEffect soundDoorOpen;
 @@ -1067,7 +_,7 @@
  		public static AudioEngine engine;
  		public static SoundBank soundBank;
  		public static WaveBank waveBank;
 -		public static Cue[] music = new Cue[42];
-+		public static MusicWrapper[] music = new MusicWrapper[42];
++		public static Music[] music = new Music[42];
```
JavidPack: This breaks all mods that do anything with the Main.music array. An example would be commented out code in ExampleMod.Load

PR #221, opened 2017-08-03
Examplemod improvements (do not merge)

Jofairden:
The following is a sample of improvements for Examplemod (Do not merge). Open for debate.


ExampleMod/RecipeHelper.cs
```diff
@@ -6,7 +6,37 @@ namespace ExampleMod
 {
 	public static class RecipeHelper
 	{
-		public static void AddBossRecipes(Mod mod)
+		private static readonly Mod mod = ExampleMod.instance;
```
JavidPack: This line doesn't work. With a reload, you are referencing an only instance of ExampleMod, causing the recipes to fail because the items do not exist.

Jofairden: Yeah I noticed that too

Jofairden:
Ok, blue is not against this. @JavidPack what about you? Mainly merging like ExampleYoyo, and adding more comments.

Chicken-Bones:
I recommend refactoring RecipeEditor to ExampleRecipes

Jofairden:
No time to overhaul more on the Example Mod right now.

- Undid some merges. 
- Added more comments to ExampleMod.cs
- Improved ExampleMod's RecipeHelper, and it's now loaded with comments for explanation
- Added more comments and improved some code for Example Spear and Example Javelin

PR #223, opened 2017-08-04
4 operators for Point16

Itorius:
### Description of the Change

Adds the +, -, * and / operators for Point16 as well as 2 new constructors

### Benefits

Allows to do sum, subtract, multiply and divide with Point16 without the need to do e.g. `new Point16(first.X + second.X, first.Y + second.Y)`


Jofairden:
I DUN GOOFED

PR #239, opened 2017-08-24
Bugfix LanguageManager.ProcessCopyCommandsInTexts crash (Resolves #230)

### LINKED ISSUE #230: AddMapEntry with an empty ModTranslation crashes
The following code in a ModTile crashes/stalls the game:

```
ModTranslation name = CreateMapEntryName();
AddMapEntry(new Color(200, 200, 200), name);
```

The crash happens in LanguageManager.ProcessCopyCommandsInTexts when it comes across a null value for the value in localizedtext.

We need to either ensure that AddMapEntry throws a load error when it has a null value in the ModTranslation's dictionary, have CreateMapEntryName initialize to "" rather than null, or other solutions.

Murfalo:
### Description of the Change

Resolves a crash in LanguageManager.ProcessCopyCommandsInTexts caused by the specific bit of code mentioned by @JavidPack in #230.

### Alternate designs

Chose version proposed in #230.  @Jofairden brought up the possibility of the programmer passing a null value to AddMapEntry, but I don't think this is a concern because either the compiler will not be able to resolve between the overloaded methods of AddMapEntry or the programmer is using an uninitialized variable, which in tModLoader's case aborts the compilation process.  I am very tired though so I am probably missing something.

### Applicable Issues

#230

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/ModTile.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ModTile.cs
@@ -150,7 +150,7 @@ namespace Terraria.ModLoader
 			{
 				key = Name;
 			}
-			return new ModTranslation(string.Format("MapObject.{0}.{1}", mod.Name, key), true);
+			return new ModTranslation($"MapObject.{mod.Name}.{key}", true);
 		}
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


Jofairden:
I think I meant being able to pass null to not use any text for the map entry and not having to create a ModTranslation object.

JavidPack:
I'll wait for blushie, she made this and probably had a reason for allowing null.

blushiemagic:
Issue resolved by 339fd6b5634cd4708011fa11ecb9fd498f3d6e51

PR #240, opened 2017-08-25
Tooltip draw hooks for ModItem & GlobalItem

Jofairden:
For review, but I think this should work fine. I tested it the entire evening with Example Sword.


Jofairden:
Feel free to improve the Example mod (Example  sword) tooltip draws, I have not much interest in that, modder will know what to do.

PR #249, opened 2017-09-14
Normalizes translation to one entry per sentence + Adds Chinese translation

Kimi-Arthur:
### Description of the Change

This PR is about translation of TML itself. It contains two parts.

1. Normalizes the translation structure so that each sentence will only use one text element. This will make it easier to translate because with this change, the location of the non-translated element won't make the sentense akward. They are always at the end of the sentence.
2. Adds the Chinese translation for TML.

### Alternate designs

For the first part, an alternative solution would be to use placeholders in translation. Like "Compiling {mod} for Mono...", or "Compiling {0} for Windows...". But since there is only one instance of combining two translated texts in TML, the proposed approach should be enough.

- It simplifies translation for other languages.
- It adds Chinese translation.

### Benefits

Same as section above.

### Possible drawbacks

Needs to update current translations (which I've done, hopefully correct).

### DIFF
--- a/patches/tModLoader/Terraria.Localization.Content.en-US.tModLoader.json
+++ b/patches/tModLoader/Terraria.Localization.Content.en-US.tModLoader.json
@@ -13,7 +13,7 @@
 		"ModsOpenModsFolder": "Open Mods Folder",
 		"ModsModPacks": "Mod Packs",
 		"ModsDisable": "Disable",
-		"ModsEnable": "Enable", 
+		"ModsEnable": "Enable",
 		"ModsMoreInfo": "More info",
 		"ModsOriginatedFromModBrowser": "This mod originated from the Mod Browser",
 		"ModsTypeToSearch": "Type to search",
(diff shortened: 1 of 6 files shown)


Jofairden:
I'm not sure what you mean with point 1,
Care to share a screenshot of the difference?

Chicken-Bones:
We should really be using string.Format forms "{0} {1}" etc, that way the location of the non-translated element can be changed by putting {0} at the appropriate location

Kimi-Arthur:
So if you choose to use "{0} {1}", it would be no problem for point 1.

Let me explain what I was thinking. Originally the sentence for compiling is like "Compiling " + modname + " for Mono", which won't read comfortably in Chinese as we don't put "for something" at the end.

So there are two solutions,

1. Change the format completely with placeholders
2. Restructure that sentence to put the mod name at end after ':'. That way the mod name won't be in the actual sentence and the sentence can be easier to translate.

I chose the second approach for simplicity (I don't need to change all the locations with translation needs). I know finally we should switch to the first approach.

I suggest I get this PR in (with Chinese translation + 1 element/sentence) and you can then create a new change to restructure the whole translation model to utilize the placeholder model? What do you say?

JavidPack:
Yeah, if you can just revert the changes to things so it's just the chinese json file that would be good (regardless of how awkward the sentence structure is). 

You can also include an extra/commented out version of the Chinese for the "MSCompiling" key with `{0}` in the place of the Mod name. Once that is in, I'll work on making sure things work with the `{#}` placeholders and then make the changes myself to the other languages in a later commit.

JavidPack:
Can you revert all the changes except the json file? I can't see to get github to do what I want.

PR #257, opened 2017-10-19
Reforge hooks updates

Jofairden:
**Added** a ReforgePrice hook available to adjust the reforge pricing. The `reforgePrice` can be changed to override the price, `canApplyDiscount` can be set to false to never apply discount. Returning false will stop the vanilla pricing to take place and the price will be set to `reforgePrice`. The passed `reforgePrice` is `item.value` (as Terraria uses that as the base price). Returns true by default so vanilla pricing takes place. This hook will be useful to adjust reforge pricing without touching the actual item value.

**Adjusted** the PreReforge hook to return a bool.  Returns whether the reforge will take place. If false is returned, the PostReforge hook is never called because the reforge will not happen. Returns true by default.

**Changed** the reforge code slightly so any modded data on the item being reforged will be preserved. This means modders no longer have to manually preserve their modded data with PreReforge and PostReforge, provided they've setup their Clone override properly.

Personally I'll be using the changes for the Even More Modifiers mod. The reforge price will be adjusted based on the modifier rarity. Possibly, I will let reforging interact with the modifiers.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
@@ -493,9 +493,16 @@ namespace Terraria.ModLoader
 		/// Returns whether the reforge will take place. If false is returned, the PostReforge hook is never called.
 		/// Reforging preserves modded data on the item. 
 		/// </summary>
-		public virtual bool PreReforge(Item item)
+		public virtual bool NewPreReforge(Item item)
 		{
 			return true;
+		}
+
+		// @todo: PreReforge marked obsolete until v0.11
+		[method: Obsolete("PreReforge now returns a bool to control whether the reforge takes place. For now, use NewPreReforge")]
+		public virtual void PreReforge(Item item)
+		{
+			NewPreReforge(item);
 		}
 
 		/// <summary>
--- a/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
@@ -479,6 +479,16 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
+		/// Returns if the normal reforge pricing is applied. 
+		/// If true or false is returned and the price is altered, the price will equal the altered price.
+		/// The passed reforge price equals the item.value. Vanilla pricing will apply 20% discount if applicable and then price the reforge at a third of that value.
+		/// </summary>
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
```diff
@@ -479,14 +479,28 @@ public virtual void OpenVanillaBag(string context, Player player, int arg)
 		}
 
 		/// <summary>
-		/// This hooks gets called immediately before an item gets reforged by the Goblin Tinkerer. Useful for storing custom data, since reforging erases custom data.
+		/// Returns if the normal reforge pricing is applied. 
+		/// If true or false is returned and the price is altered, the price will equal the altered price.
+		/// The passed reforge price equals the item.value. Vanilla pricing will apply 20% discount if applicable and then price the reforge at a third of that value.
 		/// </summary>
-		public virtual void PreReforge(Item item)
```
JavidPack: Do we know of any mods using PreReforge? Making a breaking change like this might be OK if no mods actually ever got around to using it. Otherwise we need a `bool NewPreReforge` that by default calls PreReforge, like what we did with when we had to change the Wing hooks. Also mark PreReforge as Obsolete until 0.11.

Jofairden: Woops, you're right. I didn't think about that. For sure there are plenty of mods using it because it had to be used to preserve modded data.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3239,13 +3239,44 @@
...
++
+ 						string text3 = "";
+ 						int num63 = 0;
+ 						int num64 = 0;
+@@ -36637,14 +_,17 @@
+ 							}
+ 							Main.mouseReforge = true;
+ 							Main.player[Main.myPlayer].mouseInterface = true;
+-							if (Main.mouseLeftRelease && Main.mouseLeft && Main.player[Main.myPlayer].BuyItem(num62, -1))
++							if (Main.mouseLeftRelease && Main.mouseLeft && Main.player[Main.myPlayer].BuyItem(num62, -1) && ItemLoader.PreReforge(Main.reforgeItem))
```
Jofairden: Woops. `Main.player[Main.myPlayer].BuyItem` will take away gold regardless if the PreReforge hook returns true or not.

Jofairden:
Should be fine, player.CanBuyItem is dirty. I blame vanilla

PR #259, opened 2017-10-31
Add Chinese translation.

Kimi-Arthur:
### Description of the Change

This PR is about translation of TML itself. It adds the Chinese translation for TML.

- It adds Chinese translation.

### Benefits

Same as section above.


JavidPack:
Thanks a lot!

PR #268, opened 2017-11-17
Added clarification on running script

teaberryy:
### Description of the Change

Added some clarification on how to run the tModLoaderServer script

### Alternate designs

It didn't contain the information

Because it would probably help one or two people, I guess.

### Benefits

Less confused people

### Possible drawbacks

 The readme file grows a few bytes, information might seem unnecessary 

### Applicable Issues

Ease of life when running the server software


JavidPack:
Hmm, I don't remember having to change anything when I ran it on linux, I'll try to check this.

PR #281, opened 2017-12-13
Fixes error when throwing torches

Leemyy:
### Description of the Change
Fixes an error that causes torches to duplicate and appear in the wrong position when thrown.
Also adds capability for other vanilla torches, aswell as modded torches, to be thrown the same way.

### Alternate designs
I first considered [this way](https://github.com/Leemyy/tModLoader/tree/torchThrowFix) of addressing the issue. But I decided that it was an inferior solution, since it would require each mod to manually add each of their torches to `ItemID.Sets.Torches`.

There is no way for a mod to realize the implemented behaviour.

### Benefits
Modded torches (and other vanilla torches) will be able to be thrown individually, just like the standard vanilla torch.

### Possible drawbacks
Negligibly worse performance of `Player.DropSelectedItem`.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -98,11 +98,23 @@
  					Main.item[num3].newAndShiny = false;
  					if (Main.netMode == 1)
  					{
-@@ -1760,6 +_,7 @@
+@@ -1758,8 +_,17 @@
+ 					flag2 = true;
+ 				}
  				int num = Item.NewItem((int)this.position.X, (int)this.position.Y, this.width, this.height, this.inventory[this.selectedItem].type, 1, false, 0, false, false);
- 				if (!flag2 && this.inventory[this.selectedItem].type == 8 && this.inventory[this.selectedItem].stack > 1)
- 				{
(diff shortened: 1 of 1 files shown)


JavidPack:
Can you show me this error with a gif or gfycat? I'm throwing torches and I see no difference in behavior between vanilla and modded torches.

Leemyy:
This is what throwing torches currently behaves like:
https://gfycat.com/SoftConsiderateAruanas
As you can see, normal torches cause errors when thrown, and don't appear in the right position.
Whereas all other torches are only thrown as a whole stack as opposed to just throwing a single like I would expect.

This is what the fixed code behaves like:
https://gfycat.com/DisloyalRaggedHapuku
You can see that normal torches no longer cause bugs, and all other torches can be thrown individually, too.

Leemyy:
I think the error message itself is caused by the torches being released at (0, 0) and then being tossed upwards and out of bounds.
This is further supported by the fact that dropping a stack of torches and then picking it up again will instead result in the torches appearing where the satack was before being picked up.

JavidPack:
Ok, we decided that despite the feature being an abandoned vanilla feature, we'd accept this pr.

PR #291, opened 2018-01-18
Allows thrown weapons to have prefixes

Joost8910:
### Description of the Change
As the title says, this change allows thrown weapons to have prefixes. This change only affects non-stacked thrown weapons as _GeneralPrefix(item)_ already checks for _item.maxStack_ being 1.
This makes it more like how vanilla handles prefixes for weapons like the Bone Glove, which receives ranged prefixes. 
### Benefits
The fact that it allows prefixes on thrown weapons.
### Possible drawbacks
Mods that add these weapons would have to rebalance them to accommodate the potential for increased stats.


Jofairden:
Thanks.

PR #296, opened 2018-02-14
Item chat tags now support modded data

Jofairden:
The following changes are what item tags should've been in the first place: tracking passed instances instead of tid bits of data (since the item instance is being used for hover anyway)

The solution is 'subpar', since this is what vanilla should've been, but it's now adding this 'system' on top of what vanilla already did. There is no feasible other way to do this right though, except rewriting the entire thing, but we tend to avoid that for easier times with patching, so here we are.

TODO: look at MysteryItem, it doesn't clone properly.
~~TODO: fix MP (needs sync across clients)~~ <- new approach doesnt have this issue

### DIFF
--- a/patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
+++ b/patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
@@ -1,0 +1,142 @@
+--- src/Terraria\Terraria.GameContent.UI.Chat\ItemTagHandler.cs
++++ src/tModLoader\Terraria.GameContent.UI.Chat\ItemTagHandler.cs
+@@ -2,6 +_,10 @@
+ using Microsoft.Xna.Framework.Graphics;
+ using ReLogic.Graphics;
+ using System;
++using System.Collections;
++using System.Collections.Generic;
++using System.Linq;
++using System.Text.RegularExpressions;
+ using Terraria.UI;
+ using Terraria.UI.Chat;
+ 
+@@ -9,7 +_,7 @@
+ {
+ 	public class ItemTagHandler : ITagHandler
+ 	{
+-		private class ItemSnippet : TextSnippet
++		internal class ItemSnippet : TextSnippet
+ 		{
+ 			private Item _item;
+ 
+@@ -22,10 +_,11 @@
+ 
+ 			public override void OnHover()
+ 			{
+-				Main.HoverItem = this._item.Clone();
++				Main.HoverItem = this._item;
+ 				Main.instance.MouseText(this._item.Name, this._item.rare, 0, -1, -1, -1, -1);
+ 			}
+ 
++			//@todo possibly allow modders to custom draw here
+ 			public override bool UniqueDraw(bool justCheckingString, out Vector2 size, SpriteBatch spriteBatch, Vector2 position = default(Vector2), Color color = default(Color), float scale = 1f)
+ 			{
+ 				float num = 1f;
+@@ -70,8 +_,50 @@
+ 			}
+ 		}
+ 
++		// stores item instances, used to properly display modded data
++		internal static readonly List<Item> Items = new List<Item>();
++
++		// attempts to remove stored item instances in Items
++		// that got pruned from the chat (i.e. they are no longer needed)
++		// in Main.cs at 60085 in NewText add this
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -24,6 +24,7 @@
 +		}
 +
 +		internal GlobalItem[] globalItems = new GlobalItem[0];
++		internal GlobalItem[] globalItemsInstanced {  get { return globalItems.Where(gI => gI.InstancePerEntity).ToArray(); } }
```
Chicken-Bones: this is redundant, all globalItems on an item are instanced

patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
```diff
@@ -0,0 +1,142 @@
...
+@@ -133,6 +_,13 @@
+ 			if (I.stack != 1)
+ 			{
+ 				text = text + "/s" + I.stack;
++			}
++			// assuming we have modded data
++			if (I.globalItemsInstanced.Any())
++			{
++				ushort mid = (ushort) Items.Count;
++				Items.Add(I.CloneWithModdedData(I));
```
Chicken-Bones: This is literally (read the code) 1:1 functionally equivalent to `I.Clone()`

patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
```diff
@@ -0,0 +1,142 @@
...
+ 		public static string GenerateTag(Item I)
+ 		{
+ 			string text = "[i";
+@@ -133,6 +_,13 @@
+ 			if (I.stack != 1)
+ 			{
+ 				text = text + "/s" + I.stack;
++			}
++			// assuming we have modded data
++			if (I.globalItemsInstanced.Any())
```
Chicken-Bones: Just use `item.modItem != null || item.globalItems.Any()`

patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
```diff
@@ -0,0 +1,142 @@
...
++				.ToArray();
++
++			if (itemSnippets.Any())
++			{
++				foreach (TextSnippet snippet in itemSnippets)
++				{
++					MatchCollection matchCollection = ChatManager.Regexes.Format.Matches(snippet.TextOriginal);
++					foreach (Match match in matchCollection)
++					{
++						string MIDGroup = match.Groups["options"].Value
```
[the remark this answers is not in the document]
Jofairden: @Chicken-Bones I don't see the point. These is how vanilla does it, why make something new if this can already do what we want? (it's already a regex that does it)

patches/tModLoader/Terraria/Item.cs.patch
Jofairden: @Chicken-Bones jopo said not when I asked

patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
```diff
@@ -0,0 +1,142 @@
...
++		internal static readonly List<Item> Items = new List<Item>();
++
++		// attempts to remove stored item instances in Items
++		// that got pruned from the chat (i.e. they are no longer needed)
++		// in Main.cs at 60085 in NewText add this
++		// in Main.cs at 60043 in NewText add this
++		internal static bool TryRemoveLastStoredItem()
++		{
++			TextSnippet[] itemSnippets = Main.chatLine[Main.numChatLines - 1].parsedText
++				.Where(x => x.GetType() == typeof(ItemSnippet))
```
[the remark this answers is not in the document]
Jofairden: Good catch.

patches/tModLoader/Terraria.GameContent.UI.Chat/ItemTagHandler.cs.patch
Jofairden: Why exactly does modItem need to be not null? Players can modify vanilla items.

Jofairden:
After discussion, revamped to use base64 functionality.

PR #298, opened 2018-02-16
SyncEquipment mod data support

Jofairden:
Implements IO Save/Load in SyncEquipment to support mod data.


patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -1,6 +1,10 @@
 --- src/Terraria\Terraria\Chest.cs
 +++ src/tModLoader\Terraria\Chest.cs
-@@ -5,6 +_,7 @@
+@@ -1,10 +_,10 @@
+ using Microsoft.Xna.Framework;
+-using System;
```
Chicken-Bones: please revert, I know the decompiler adds unnecessary usings, but we want to keep patches small. We don't want to have to deal with failed patches because a new using statement got added in 1.3.6

PR #326, opened 2018-04-05
Consume Item Additions

DarkLight66:
### Description of the Change

* Added missing ItemLoader.ConsumeItem calls to most instances of an item being consumed:
  * Snowballs when using the Snowball Launcher.
  * Lihzahrd Power Cells when spawning the Golem on the Lihzahrd Altar.
  * Temple Keys when unlocking the Temple Door. 
    * Also fixed a vanilla bug where 1 temple key is consumed from all stacks in the inventory rather than only consuming the first one.
  * Keys when unlocking vanilla chests.
  * Paint when painting a tile or wall manually or when using the Paint Sprayer or Achitect Gizmo Pack.
  * Actuators when placing them manually or when using the Presserator.
  * Walls when placed automatically to fill 1 block holes.
  * Wormhole Potions when used by clicking a teammate's icon on the map.
  * Fishing bait items.
  * Wires when used by a single colored Wrench.
  * Recall Potions.
  * Teleportation Potions.
  * Gender Change Potions.
  * Vanilla Grab Bags and any other right-clickable item added by mods.
  * Any item being consumed through Player.ConsumeItem. In vanilla this includes:
    * Eternia Crystals when being placed in the Eternia Crystal Stand.
    * Large Gems when placed in a Gem Lock.
    * Etherian Mana when used by the Sentry Summons from the DD2 event.
    * Golden Keys when used to unlock Golden Lock Boxes.
    * Wires when used by the Multicolor Wrench or The Grand Design.
    * Actuators when used by The Grand Design.

* Adds OnConsumeItem hook, which only runs if ConsumeItem returns true.
* Adds OnConsumeAmmo hook, which only runs if ConsumeAmmo returns true for both the weapon and the ammo.
  * Also fixed a bug were ItemLoader.ConsumeAmmo would use the GlobalItem instance of the weapon to call the GlobalItem.ConsumeAmmo method of the ammo.

* Without the ConsumeItem fixes, the consumption of some items that are only consumed in special circumstances couldn't be prevented (i.e paint, wires, bait, etc).

* Without the OnConsumeItem and OnConsumeAmmo hooks, there aren't many ways to know for certain if an item was consumed on use.

### Benefits

* The fixes allows modders to prevent pretty much any item from being consumed.

* The new hooks allow modders to know if any item was consumed on usage, and to act on it.

### Possible drawbacks

Hopefully none.

### Sample Usage

See the new Infinity accessory and edits to ExampleBullet in the included ExampleMod edits.


patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
```diff
@@ -467,16 +467,32 @@ public static bool ConsumeAmmo(Item item, Item ammo, Player player)
...
 			return true;
 		}
 
+		private static HookList HookOnConsumeAmmo = AddHook<Action<Item, Player>>(g => g.OnConsumeAmmo);
+		/// <summary>
+		/// Calls ModItem.OnConsumeAmmo for the weapon, ModItem.OnConsumeAmmo for the ammo, then each GlobalItem.OnConsumeAmmo hook for the weapon and ammo.
+		/// </summary>
+		public static void OnConsumeAmmo(Item item, Item ammo, Player player)
+		{
+			item.modItem?.OnConsumeAmmo(player);
```
Jofairden: If an `item.IsAir` check is present below, why not here?

patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
```diff
@@ -968,7 +1000,8 @@ public static void RightClick(Item item, Player player)
 			foreach (var g in HookRightClick.arr)
 				g.Instance(item).RightClick(item, player);
 
-			item.stack--;
+			if (ConsumeItem(item, player))
+				item.stack--;
 			if (item.stack == 0)
```
Jofairden: Can be made nested to avoid an unnecessary check from happening if `ConsumeItem` returns false

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4924,21 +5162,25 @@
  				{
  					flag2 = true;
  				}
-@@ -37645,6 +_,14 @@
+@@ -37648,8 +_,18 @@
+ 				{
  					flag2 = true;
  				}
- 				if ((shoot == 145 || shoot == 146 || shoot == 147 || shoot == 148 || shoot == 149) && this.itemAnimation < this.itemAnimationMax - 5)
++				if (!PlayerHooks.ConsumeAmmo(this, sItem, item))
```
Jofairden: The patch difference can be made smaller by using the bitwise operator
`flag2 = !PlayerHooks.ConsumeAmmo(this, sItem, item) & !ItemLoader.ConsumeAmmo(sItem, item, this);`

patches/tModLoader/Terraria.UI/ItemSlot.cs.patch
```diff
@@ -53,6 +53,75 @@
  				{
  					if (Main.mouseRightRelease)
  					{
+ 						player.OpenBossBag(inv[slot].type);
+-						inv[slot].stack--;
```
Jofairden: In regards to the earlier comment, do not turn these nested because it will increase the patch difference.

Chicken-Bones:
I haven't looked over the code but this is the kind of quality content we're looking for from contributors. Good work!

patches/tModLoader/Terraria/Player.cs.patch
Jofairden: Yes

Jofairden:
We just tend to forget. Though I believe cb wanted to review it, but he's a busy man.

Chicken-Bones:
Good work. Sorry we left it for so long.
Thankyou for your contribution

PR #327, opened 2018-04-06
skipselect now takes in player and world name

Itorius:
### Description of the Change

User will be able to specify which player and world they want to launch with -skipselect launch option. When either one of the fields is invalid/empty, it will choose the default player/world (meaning first in the list).

### Benefits

Adds the ability to specify which player/world is used for debugging.

### Sample Usage

`-skipselect "Itorius:TestWorld"`


JavidPack:
Thanks a bunch.

Issue #265, opened 2017-11-12
Preserving Item Mod Data Megathread


JavidPack:
This issue serves as a home for all issues pertaining to preserving mod data in Mod Items through all possible movement vectors. 

All items are tested with ExampleMod's ExampleInstancedGlobalItem.

### Single Player
- [ ] Inventory
  - [x] Save/Load
  - [ ] Quick stack to chests (see #235)
  - [x] Linking an item into chat (by holding LALT)
  - [x] Preserve through Unload (Data Lost as of Nov 11, 2017)
    - [x] Global Item data recovered from MysteryGlobalItem
    - [x] MysteryGlobalItem preserved through reloads
- [x] Chest
  - [x] Save/Load in Chest
  - [x] Preserve through Unload (Data Lost as of Nov 11, 2017)
- [x] Shop Sell/Buy back
- [x] Throw into world
- [x] Reforge ?
- [x] Tooltip (Clone)
- [x] Shift Click to Chest/Trash/Shop
- [ ] Right Click to/from Equipment ?
- [x] Tile Entities and Special Tiles
  - [x] Item Frames #207 

### 1.4 Update
- [ ] Void Storage
- [ ] ItemDropRules

### Multiplayer
TODO: Untested
- [ ] Quick Stack #235 

## ModItem
TODO: Untested

Chicken-Bones:
If you want to handle the testing for this, I'm quite happy to do the fixing and verify everything is going through the appropriate channels (since I did the Save/Load, Networking and advised on Mod/Global design)

Also are we talking about ModItem data or GlobalItem data in each case, and as a sub point for Save/Load/Unload, are we talking about when the mod providing the GlobalItem is unloaded, or when the mod providing the host item with GlobalItem data from another mod is unloaded?

JavidPack:
For the preserve through unload, what I tried so far was a vanilla item with GlobalItem data attached (ExampleInstancedGlobalItem, it attaches a string to the item). The mod supplying the GlobalItem was unloaded, the world entered and exited, the mod was re-enabled, and the world was re-entered. In this case, the data was still present in the .twld file after leaving the world while the mod was unloaded, but the MysteryGlobalItem was not restored to the Player or World once the mod was reenabled. Also, if the world is entered while the data is already in MysteryGlobalItem, the data is lost as well. (As in, Exit World/Unload/Enter/Exit/Enter/Exit)

Jofairden:
[...]

PR #329, opened 2018-04-21
Recover GlobalItem data through unload

Leemyy:
### Description of the Change

GlobalItem data will no longer be lost when an Item is loaded without the GlobalItem's Mod active.
Judging by how simple this change is, it seems like a typo of sorts. I did several tests to make sure I am not overlooking something, but all GlobalItems seem to get recovered properly. Even after several reloads without the Mods activated, all GlobalItems were recovered.

### Alternate designs

I first considered extracting the contents of `MysteryGlobalItem.data` in `ItemIO.SaveGlobals(Item item)` and to append them directly to the TagCompound. I opted for this instead, because it is a much simpler change.

It fixes part of a known issue.

### Applicable Issues

#265


Jofairden:
Huh, odd that it's such a small change. Might be an oversight indeed.

Chicken-Bones:
Blatant typo, good find.

PR #332, opened 2018-04-23
Offset buff tooltip properly, fixes #318

### LINKED ISSUE #318: Buffs with multiple lines of description not drawing properly BUG
### Description

Modded buffs descriptions are offset, creating extra unnecessary space for each new line in the description.

### Steps to Reproduce

These are the two codes I've tested.
`Description.SetDefault("DESCRIPTION\nSECOND LINE");`
`Description.SetDefault("DESCRIPTION\nSECOND LINE\nTHIRD LINE\nFOURTH LINE");`

### Actual behavior

First

Second

### Versions

tModLoader v0.10.1.3

### Mods

Just the mod used for testing.

Leemyy:
### Description of the Change

The buff tooltip needs to be offset by the height of the buff name, not by its own height.
This is technically a vanilla bug, but vanilla does not have multiline buff tooltips, so it never has a visual impact.

### Alternate designs

The buff name could be passed, instead of just its height, but this would just result in more code without any real benefit. It would also mean that the dimensions of the buff name will have to be calculated twice.

### Benefits

Multiline buff tooltips will get drawn properly. (See linked Issue for details)

### Possible drawbacks

None, as far as I can tell.

### Applicable Issues

#318

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -877,6 +877,26 @@
  		}
  
  		private void DoUpdate_AnimateWaterfalls()
+@@ -14484,6 +_,7 @@
+ 				num2 += 6;
+ 			}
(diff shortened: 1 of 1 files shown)


Jofairden:
Thank you for another contribution.

PR #337, opened 2018-05-03
Update Terraria.Localization.Content.fr-FR.tModLoader.json

orian34:
French translation update


orian34:
I haven't finished it, I'll comment when it's complete

orian34:
THERE IT IS!
Dunno if the lines should stay commented out or not. I only translated

Jofairden:
I don't think they should stay commented, not sure why they are anyway.

Jofairden:
@JavidPack made a python script for the translations. I believe the commented lines are placeholder and meant to be uncommented when implemented.

JavidPack:
Yeah, can you remove the "// " from all the lines that you translated?

PR #340, opened 2018-05-07
Ensure the world is fully loaded before updating entities

Leemyy:
### Description of the Change

Before I go into detail on this change, [here's a video](https://youtu.be/YE0Qr5y36A8) to illustrate it more clearly.
When connecting to a server, vanilla will eventually send message 8 (`MessageID.RequestTileData`). This will prompt the server to send tile information, aswell as various game objects to the client. This includes all projectiles that are considered pets or have `netImportant` set to true (MessageBuffer:720). However, when connecting to a server, the pets/summons of other players will not show up unless they are re-summoned.
This has to do with the way the world is loaded on the client. While the data is being transmitted to the client, it will still perform regular updates. Because of this, summons, pets, and some other entities will get updated before their associated player is loaded (players are loaded last). Almost all entities that are in some way bound to a player will just deactivate themselves, when said player is not present; which is what's happening here.
My solution to this problem is to skip updating any entities while the client is initially receiving world data. In order to achieve this I added a condition to `Main.DoUpdate`, which will skip updating all entities if `Main.syncingWorld` is true.
`Main.syncingWorld` gets set before the client sends the request for world data (packet 8) to the server (Netplay:354). And it gets reset as soon as the client receives packet 74 (`MessageID.AnglerQuest`). I chose this packet to reset `Main.syncingWorld`, because it gets sent by the server, right after all players have been sent to the client. (MessageBuffer:789 and NetMessage:2270 which gets called by `MessageBuffer.SyncConnectedPlayer` -> `NetMessage.SendAnglerQuest`) If there are no other players, packet 74 will still get sent.
In case there is a connection issue, or if the client disconnects, `Main.syncingWorld` will also get reset (Netplay:393).

### Alternate designs

I considered adding an extra packet to tell the client when syncing is done, aswell as the possibilty of just re-sending the projectiles after all players have been synced. However, both of those options would require more code changes and cause more network overhead.

It fixes an oversight in the vanilla code.

### Benefits

This will reduce the ammount of desyncing between clients.

### Possible drawbacks

There is a chance that a loss of connection at a specific point in the syncing process will leave the client permanently frozen, because `Main.syncingWorld` stays `true`. Though I tried to think of all possible outcomes of a connection loss while syncing, and I am farily certian that I covered all cases. My testing also never led to a frozen client.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -788,6 +788,14 @@
  				}
  				this._isDrawingOrUpdating = false;
  			}
+@@ -11968,6 +_,7 @@
+ 			}
+ 		}
+ 
++		internal static bool syncingWorld = false;
+ 		protected void DoUpdate(GameTime gameTime)
+ 		{
+ 			Main.ignoreErrors = true;
 @@ -11997,6 +_,7 @@
  			{
  				Main.InGameUI.Update(gameTime);
@@ -842,6 +850,15 @@
 +				PlayerHooks.SendClientChanges(Main.player[Main.myPlayer], Main.clientPlayer);
  			}
  			if (Main.netMode == 1)
+ 			{
+@@ -12624,6 +_,8 @@
+ 			PortalHelper.UpdatePortalPoints();
+ 			Main.tileSolid[379] = false;
+ 			Main.ActivePlayersCount = 0;
++			if (Main.syncingWorld)
++				goto IL_1B52;
+ 			int num12 = 0;
+ 			while (num12 < 255)
  			{
 @@ -12982,7 +_,7 @@
  
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -855,7 +855,7 @@
  			PortalHelper.UpdatePortalPoints();
  			Main.tileSolid[379] = false;
  			Main.ActivePlayersCount = 0;
-+			if (Main.syncingWorld)
++			if (Main.netMode == 1 && Main.syncingWorld)
 +				goto IL_1B52;
  			int num12 = 0;
  			while (num12 < 255)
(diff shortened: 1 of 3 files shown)


JavidPack:
Can you tell me if this is a vanilla bug as well? If so I'll consult with them as well on a fix.

Can we just use if(Main.netMod == 1 && Netplay.Connection.State != 10) instead of making a whole new state variable?

Leemyy:
Yes, this is a vanilla bug as can be seen in [this video](https://youtu.be/ohYWYjyWNkM).

As for your second question, I tried using `if (Main.netMode == 1 && Netplay.Connection.State != 10)`, but it doesn't work.
The connection state gets set to 10 (MessageBuffer:3607) before the player spawns (MessageBuffer:3610). `Player.Spawn` sends packet 12 to the server (Player:28605), which is what prompts the server to send the other players (MessageBuffer:787).

Checking for `Main.netMode == 1` is definitely a good idea, though. So I addded that for extra safety.

Chicken-Bones:
Good enough for me, last suggestion before I approve. Move `syncingWorld` to `Netplay` as it's kindof an extension to `ConnectionState` the `if Main.netMode == 1` makes it natural to then check `&& Netplay.syncingWorld`

Leemyy:
After giving this some more thought, I realized that it would be possible to change (MessageBuffer:3607) to `Netplay.Connection.State = 9;` (Nothing in vanilla relies upon this being connection state 10, I checked).
Then we could replace [`Main.syncingWorld = false;`](https://github.com/Leemyy/tModLoader/blob/1934c0a43ce080694564c05b3acdf79ad086ca7c/patches/tModLoader/Terraria/MessageBuffer.cs.patch#L407) with `Netplay.Connection.State = 10;` and use the check @JavidPack suggested.

Chicken-Bones:
I'd sooner add state 11

Leemyy:
I was recommending this, because some mods might rely on it being state 10.

Chicken-Bones:
Let's not touch it then, new variable is better, easier to find references to as well

Leemyy:
Okay, I moved `syncingWorld` to `Netplay`.

Jofairden:
Thanks !

PR #147, opened 2017-03-21
Add SetChatButtons and OnChatButtonClicked to GlobalNPC (Resolves #68)

Murfalo:
Finally got around to actually fixing this

Issue #68, opened 2016-12-11
SetChatButtons and OnChatButtonClicked not implemented in GlobalNPC


alecamaracm:
Both SetChatButtons and OnChatButtonClicked are not implemented in the GlobalNPC class.

Jofairden:
This was never closed, but it should actually be solved. See #345

PR #343, opened 2018-05-12
More NPC chat hooks and something extra

DarkLight66:
### Description of the Change

* **Added**:

  * **ModNPC / GlobalNPC CanChat hooks**: allows determining if an NPC can talk with the player, regardless of it being a town NPC or not.
  * **GlobalNPC SetChatButtons hook**: allows editing the text of vanilla and modded NPC chat buttons.
  * **GlobalNPC OnChatButtonClicked hook**: allows defining the functions of vanilla NPC chat buttons with no default behavior, as well as adding to the functions of modded NPC chat buttons.
  * **GlobalNPC PreChatButtonClicked hook**: allows overriding or adding to the functions of vanilla and modded NPC chat buttons.
  * **ModNPC / GlobalNPC ShowHoverInfo hooks**: allows hiding the name and health text when hovering over an NPC. 
  * **ModPlayer UpdateAutopause hook**: allows doing things in single player when the game is paused thanks to the autopause option.
  * **ModProjectile / GlobalProjectile PostDamage hooks**: allows doing things after a projectile finishes doing damage to tiles and NPCs / players. Only runs if the CanDamage hook returns true.
  * Usage examples of all the above hooks to Example Mod.

* **Changed**:

  * In Player.TileInteractionsCheckLongDistance, added a check if the tile is active to prevent showing hover info of unactive tiles. This was really only noticeable when using chests to spawn mimics from keys.
  * ExampleMod chest mimic summon check to match the vanilla mimic's checks.

### Alternate designs

* **PreChatButtonClicked**: i thought of including a shop parameter similar to the one in OnChatButtonClicked, but couldn't find a way to properly implement the case were the NPC already has a shop.
* **PostDamage**: The intention of this hook was to allow projectiles to apply effects to NPCs / players without them having a damage value, but it ended up being more of a general use hook.

* **Whitout**:

  * **CanChat hooks**: There is no proper way of allowing NPCs to chat with the player otehr than setting `npc.townNPC` to true, nor is there a way of preventing specific town NPCs from talking with the player.
  * **Global NPC chat button hooks**: There is no proper way of adding to / overriding the text or functions of vanilla and modded NPC chat buttons.
  * **ShowHoverInfo hooks**: There is no proper way of hiding the NPCs name and health when hovering over them.
  * **UpdateAutopause hook**: some bugs related to the autopause option being on can't be fixed properly.
  * **PostDamage hook**: applying effects to NPCs / players from projectiles with no damage value requires putting the code in one of the AI methods, where they ran regardless of the CanDamage hook returning false, and without using the damage hitbox after being modified by the ModifyDamageHitbox hook.

### Benefits

* **New NPC chat hooks**: allow creating custom NPCs that don't have any of the town NPC's limitations, as well as allowing them to globally modify how vanilla and modded NPC chat buttons function, or preventing certain NPCs from talking under special conditions.
* **ShowHoverInfo hook**: allows implementing full invisibility for NPCs.
* **UpdateAutopause hook**: allows updating field values that would cause bugs with autopause the auto pause option being on, like mimic spawning from keys in chests.
* **PostDamage hook**: is another general update hook for projectiles.

### Possible drawbacks

* The new NPC chat hooks allow modders to mess with NPC chats from other mods without the other mod being able to account for it.

### Applicable Issues

#68 

### Sample Usage

See the new Fallen Ghost NPC and edits to ExampleGlobalNPC and ExampleChestSummon in the included ExampleMod edits.

### Notes

Some of the new Global NPC chat buttons code was based on #147 made by @Murfalo.


JavidPack:
I haven't been able to read everything yet, but I can see a lot of these being controversial, it might be better to split something like this into several PRs.

Jofairden:
I think you should separate this into multiple PRs, one for NPCs, one for Players and one for Projectiles.
I feel like PostDamage is unnecessary bloat, there is already hooks for this. You could use the current CanDamage, or possible inside ModifyDamageHitbox (goes after CanDamage) most likely. If we were still to go for a functionality to ignore damage, we should have something like a NeedsDamageToHit property instead. Though, what is a good usecase? You're probably better off doing manual collision in one of the hooks mentioned or AI. For modifying vanilla chat we are concerned with mod compatibility, as there is no way to make it compatible between mods. To hide NPC names and healthbars you could modify npc.GivenOrTypeName to be empty or force npc.nameOver to be 0f, and to stop drawing the healthbar there already exists the DrawHealthBar hook. (simply return false) Though one could argue that it could use a DrawHoverName hook for consistency, although healthbars do not have such values to cancel out the drawing of the npc, so it's a conflicting matter. To stop the player health bar from drawing you could force Main.HealthBarDrawSettings to not be 1 or 2. You can see how this becomes a conflicting matter when there is already plenty of ways to satisfy possible usecases. I can't speak about UpdateAutopause, it seems like a useful enough hook.

DarkLight66:
I agree the PostDamage hook isn't really necessary, i just thought it would be more consistent with the other vanilla projectiles that do things like transform NPCs without doing damage.
On the ShowHoverInfo hook, i was talking about the text "CurrentLife / MaxLife" that is shown after the name, which can only be hidden if the NPC's max life is less or equal to 1, or if the `dontTakeDamage` flag is true. I didn't know about `npc.nameOver`, but that only seems to affect the drawing of the name when using a game-pad, and setting `npc.GivenOrTypeName` doesn't hide the health text, like i mentioned. I was already aware of the DrawHealthBar hook, as seen in the Fallen Ghost NPC example. 
On the problem of modifying vanilla chats, what if those are excluded, only allowing the modification of modded ones? Can you give more feedback about this?
And about hiding player health bars, i actually didn't really think about those at all, as i don't play in MP, let alone in PvP.
I'll be dividing this PR into one for UpdateAutopause, another one for all the NPC chat hooks, a third one for ShowHoverInfo, and PostDamage may have to be scraped.

PR #344, opened 2018-05-13
UpdateAutopause hook

DarkLight66:
### Description of the Change

* **Added**:

  * **ModPlayer UpdateAutopause hook**: allows doing things in single player when the game is paused due to the autopause option being on.
  * Usage example of the hook to Example Mod.

* **Changed**:

  * In Player.TileInteractionsCheckLongDistance, added a check if the tile is active.
  * ExampleMod ExampleChestSummon.ChestItemSummonCheck to match the vanilla mimic's checks.

* Without the UpdateAutopause hook, some bugs related to the autopause option being on can't be fixed properly.
* Without the check in Player.TileInteractionsCheckLongDistance, tiles could show hover info even if they weren't active. In vanilla, this was really only noticeable when using chests to spawn mimics from keys.
* Without the change in ChestItemSummonCheck, spawning the NPC from a modded chest would remove the chest entity but leave the chest tiles instead of removing them too.

### Benefits

* **UpdateAutopause hook**: allows updating field values that would cause bugs with autopause the auto pause option being on, like mimic spawning from keys in chests.

### Sample Usage

See edit to ExampleChestSummon in the included ExampleMod edit.

### Notes

The lack of a check for the tile being active in Player.TileInteractionsCheckLongDistance is probably a vanilla bug rather than a tModLoader one.

### DIFF
--- a/ExampleMod/Items/Infinity.cs
+++ b/ExampleMod/Items/Infinity.cs
@@ -32,7 +32,7 @@ namespace ExampleMod.Items
 
 		public override Color? GetAlpha(Color lightColor)
 		{
-			return Color.White; // So the item's sprite isn't effected by light
+			return Color.White; // So the item's sprite isn't affected by light
 		}
 
         // This gives the item an outline that constantly changes color
@@ -50,6 +50,7 @@ namespace ExampleMod.Items
 			return true;
 		}
 
+		// Same as above but for drawing inside the player's inventory
 		public override bool PreDrawInInventory(SpriteBatch spriteBatch, Vector2 position, Rectangle frame, Color drawColor, Color itemColor, Vector2 origin, float scale)
 		{
 			Texture2D texture = Main.itemTexture[item.type];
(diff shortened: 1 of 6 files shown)


Jofairden:
> Without the UpdateAutopause hook, some bugs related to the autopause option being on can't be fixed properly.

Could you give specific examples?

For the UpdateAutopause hook, wouldn't checking `Main.gamePaused` be sufficient?

DarkLight66:
Spawning biome mimics from chests with keys of light / night works with autopause on in vanilla because there's an update to the player's last chest index that only happens when the game is paused. The ExampleChestSummon code cannot account for autopause being on because none of the ModPlayer Update hooks run when the game is paused, as the player's Update method doesn't run at all when the game is paused. I know this probably could be fixed using a hook from another class, but in my opinion ModPlayer should be the central class for things of this nature. There's probably other problems that could be fixed with a hook that is sure to run only when the game is paused when using autopause, but the mimic spawning one is the only one i remember of the top of my head.

Issue #347, opened 2018-05-16
Carrot replaces item in hot bar from other mods with Collectors Edition


thakyZ:
### Description

When I have the creator's edition perks enabled on Terraria, specifically the carrot for the pet bunny. When creating a new character the item that is the fourth item in the hot bar placed by any mod gets replaced automatically by the carrot.

### Steps to Reproduce

1. Have the collector's edition enabled, or use RegEdit to enable carrot's for new characters (given this is tested on a Windows machine)
2. Create a new character with a mod like Thorium
3. Notice that the player doesn't spawn with an Heirloom and instead spawns with a carrot

### Expected behavior 

Spawn with a carrot and any and all items any mods add to the how bar when creating a new character.

### Actual behavior

Spawns with whatever item a mod added to the fourth hot bar slot replaced with a carrot

### Reproduction frequency

Always as long as the collector's edition is enabled

### Versions

tMod Loader 0.10.1.4

### Mods

BossChecklist
CalamityMod
DALib
Fargowiltas
FKBossHealthBar
imkSushisMod
ReforgeArmor
TerrariaOverhaul
ThoriumMod
VariablePotionSickness

### Additional Information

If the testers do not know how to edit regedit to enable the carrot, I can test for you, as I don't think it's allowed to repost the way to edit it.

PR #348, opened 2018-05-16
Collector's Edition Fix

DarkLight66:
### Description of the Change

* Prevents the carrot from the Collector's Edition version of Terraria from replacing the item on the fourth slot on the hotbar on character creation after the SetupStartInventory hook is called.

### Applicable Issues

#347

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/PlayerHooks.cs
+++ b/patches/tModLoader/Terraria.ModLoader/PlayerHooks.cs
@@ -112,7 +112,7 @@ namespace Terraria.ModLoader
 
 		private static HookList HookSetupStartInventory = AddHook<Action<List<Item>>>(p => p.SetupStartInventory);
 
-		public static IList<Item> SetupStartInventory(Player player, bool newPlayer = true)
+		public static IList<Item> SetupStartInventory(Player player, bool mediumcoreDeath = false)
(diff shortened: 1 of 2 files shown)


Jofairden:
What's the `newPlayer ` bool for? SetupStartInventory is meant for newly created characters, so it's redundant.

Leemyy:
SetupStartInventory also gets called when a mediumcore character dies, so I'd assume that's why it's needed.

Jofairden:
And why does it get called? Mediumcore characters should drop items upon death

JavidPack:
Don't they get a free copper pick and other items when they respawn?

Leemyy:
Yes, it's called to give you back the starting tools. Otherwise you could easily end up softlocked, because your stuff dropped in a place you can't reach without tools. Also, when a mediumcore character dies all items that are part of the starting inventory will get deleted, instead of getting dropped.

I'm assuming that DarkLight added this flag, because the pet is only given to a character upon creation, but not when they respawn.

Jofairden:
In that case, I think it is better to have the boolean suggest mediumcore death rather than character creation, as mediumcore death in this context is the special circumstance. If you do that, I'll accept it.

JavidPack:
Looks good.

Issue #349, opened 2018-05-19
Mouseover text on animated items bug


DarkLight66:
### Description

The hitbox that is used for showing the name of items in the world depends on texture size rather than frame size or item hitbox size. For animated items, this results in the name drawing when the cursor is far above the item.

### Steps to Reproduce

1. Drop an animated item in the world.
2. Move the cursor above the item, inside the area the actual texture of the item would cover.

### Expected behavior 

The name of the item isn't shown.

### Actual behavior

The name of the item is shown.

### Reproduction frequency

Always

### Versions

tModLoader v0.10.1.4

### Mods

None

### Additional Information

This is actually a vanilla bug, but was probably overlooked because not many items are animated in vanilla.

Jofairden:
In my opinion, it should be offset by the entity's size. It makes the most sense. Though, since vanilla does it by texture size, we should probably follow that convention (and offset it by frame size if animated)

PR #350, opened 2018-05-20
Fixes mouse over hitbox for animated items

DarkLight66:
### Description of the Change

* Changes the hitbox that is used for showing the name of items in the world to be dependant on frame size than texture size. This fixes the bug where the name was being drawn when the cursor was far above the item.

* Replaced the unused ItemLoader.AnimatedItemFrame method with ItemLoader.ItemHoverHitbox method, which returns a rectangle representing the item texture's current frame for animated items, or the texture's dimensions for normal items, with its location set to the item's position.

### Applicable Issues

#349

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
@@ -261,11 +261,10 @@ namespace Terraria.ModLoader
 			}
 		}
 
-		private static Rectangle AnimatedItemFrame(Item item)
+		internal static Rectangle ItemHoverHitbox(Item item)
 		{
-			int frameCount = Main.itemAnimations[item.type].FrameCount;
-			int frameDuration = Main.itemAnimations[item.type].TicksPerFrame;
-			return Main.itemAnimations[item.type].GetFrame(Main.itemTexture[item.type]);
+			Rectangle frame = Main.itemAnimations[item.type]?.GetFrame(Main.itemTexture[item.type]) ?? Main.itemTexture[item.type].Bounds;
+			return new Rectangle((int)(item.position.X + item.width * 0.5f - frame.Width * 0.5f), (int)(item.position.Y + item.height - frame.Height), frame.Width, frame.Height);
 		}
 
 		private static HookList HookChoosePrefix = AddHook<Func<Item, UnifiedRandom, int>>(g => g.ChoosePrefix);
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -3531,6 +3531,15 @@
  				Main.PlaySound(7, -1, -1, 1, 1f, 0f);
  			}
  		}
+@@ -37535,7 +_,7 @@
+ 				{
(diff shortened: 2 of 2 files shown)


Jofairden:
The `Applicable Issues` section is more meant for possible issues with the requested changes. Next time it's fine if you make the PR without making the issue. Oh well...

patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
```diff
@@ -261,11 +261,10 @@ internal static void DrawAnimatedItem(Item item, int whoAmI, Color color, Color
...
 		}
 
-		private static Rectangle AnimatedItemFrame(Item item)
+		internal static Rectangle ItemHoverHitbox(Item item)
 		{
-			int frameCount = Main.itemAnimations[item.type].FrameCount;
-			int frameDuration = Main.itemAnimations[item.type].TicksPerFrame;
-			return Main.itemAnimations[item.type].GetFrame(Main.itemTexture[item.type]);
+			Rectangle frame = Main.itemAnimations[item.type]?.GetFrame(Main.itemTexture[item.type]) ?? Main.itemTexture[item.type].Bounds;
+			return new Rectangle((int)(item.position.X + item.width * 0.5f - frame.Width * 0.5f), (int)(item.position.Y + item.height - frame.Height), frame.Width, frame.Height);
```
Jofairden: Potential problem is if a modder decides to use different frame sizes. However, it is highly unlikely someone would ever do that. Therefore I let it pass.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3531,6 +3531,15 @@
  				Main.PlaySound(7, -1, -1, 1, 1f, 0f);
  			}
  		}
+@@ -37535,7 +_,7 @@
+ 				{
+ 					if (Main.item[i].active)
+ 					{
+-						Microsoft.Xna.Framework.Rectangle value = new Microsoft.Xna.Framework.Rectangle((int)((double)Main.item[i].position.X + (double)Main.item[i].width * 0.5 - (double)Main.itemTexture[Main.item[i].type].Width * 0.5), (int)(Main.item[i].position.Y + (float)Main.item[i].height - (float)Main.itemTexture[Main.item[i].type].Height), Main.itemTexture[Main.item[i].type].Width, Main.itemTexture[Main.item[i].type].Height);
++						Microsoft.Xna.Framework.Rectangle value = ItemLoader.ItemHoverHitbox(Main.item[i]);
```
Jofairden: Our own method is potentially redundant if it is only used here. (can be considered sort of exasperate when redirecting to other classes when not needed) Oh well...

patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
DarkLight66: Yeah, most modders don't even make animated items, let alone override how they are drawn in the world. And if they really want different frame sizes, they could potentially create a custom class that inherits from DrawAnimation for it.

Jofairden:
@JavidPack  I approved this a while ago. If we intend to update soon, I think we should merge this.

PR #351, opened 2018-05-20
Implemented Luck for NPC Loot

HumanGamer:
### Description of the Change
This modifies the NPC Loot function to use a new class called LuckBasedRandom to allow modders to adjust the luck of NPC Loot drops. (incl. drop chance and quantity)

Without this change there is no way to globally change the chances/quantities of item drops.

### Alternate designs

One alternate idea could be to completely redesign how the loot works with a List where the mob ids have an associated struct/class object with all information about that mob's drops. Although this would be a much better alternative, doing so would take forever to implement properly. It could also be difficult to replicate the exact functionality of the loot calculations if not done perfectly and extensively tested.

Changing drop luck cannot be done easily from within a mod, if at all. Also it allows for the luck to be changed by multiple mods (otherwise having many different implementations per mod could easily cause problems, assuming there is any other way to do this).

### Benefits

This will allow modders to change the drop luck globally with potions, accessories etc. Modders have the option of using this in their custom NPCs (to allow for luck to apply)

### Possible drawbacks

If I made a mistake anywhere there may be issues with some drops. These could easily be corrected however.

Also, every time the game updates with changes related to the NPC loot, this will need to be updated to support the new loot/changed drop rates.

### Applicable Issues

N/A

### Sample Usage

To allow a custom NPC to have luck apply, one simply has to do
```cs
NPC.rand.NextBool(100)
```
instead of
```cs
Main.rand.Next(100) == 0
```
and
```cs
NPC.rand.NextInt(24, 30)
```
instead of
```cs
Main.rand.Next(24, 30)
```

To change the luck value (for use with a potion/accessory or other means), you can simply do something like this
```cs
NPC.Luck += 100
```

Keep in mind that the default is 100 (100%) and lowering the value below 100 will lower the drop chances and raising above 100 will increase the drop chances. 200 (200%) will double the drop chances.

### DIFF
--- a/patches/tModLoader/Terraria.Localization.Content.de-DE.tModLoader.json
+++ b/patches/tModLoader/Terraria.Localization.Content.de-DE.tModLoader.json
@@ -169,6 +169,7 @@
 		// "MBMyPublishedModsStats": "{0} downloads ({1} latest version)",
 		// "EnterSteamID": "Please Enter Your SteamID",
 		// "PasteSteamID": "Paste SteamID (ctrl-v)",
+		// "MBServerResonponse": "The Mod Browser server response:\n\n{0}",
 
 		// TML Options
 		"DownloadFromServersYes": "Lade mods vom Server herunter: An",
--- a/patches/tModLoader/Terraria.Localization.Content.en-US.tModLoader.json
+++ b/patches/tModLoader/Terraria.Localization.Content.en-US.tModLoader.json
@@ -169,6 +169,7 @@
 		"MBMyPublishedModsStats": "{0} downloads ({1} latest version)",
 		"EnterSteamID": "Please Enter Your SteamID",
 		"PasteSteamID": "Paste SteamID (ctrl-v)",
+		"MBServerResonponse": "The Mod Browser server response:\n\n{0}",
 
 		// TML Options
 		"DownloadFromServersYes": "Download Mods From Servers: On",
--- a/patches/tModLoader/Terraria.Localization.Content.es-ES.tModLoader.json
+++ b/patches/tModLoader/Terraria.Localization.Content.es-ES.tModLoader.json
@@ -169,6 +169,7 @@
 		// "MBMyPublishedModsStats": "{0} downloads ({1} latest version)",
 		// "EnterSteamID": "Please Enter Your SteamID",
 		// "PasteSteamID": "Paste SteamID (ctrl-v)",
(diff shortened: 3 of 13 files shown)


Jofairden:
So how is this any useful when `Luck` is global on the `NPC` class?

HumanGamer:
It would work for luck potions, accessories etc (all npcs at once). It wouldn't be that hard to change to a non-static variable though which could be used per-npc.

Chicken-Bones:
This thing has been discussed in the past and we'd like to see it, but the sheer patch quantity makes it basically unmaintainable when Terraria updates. We'll have to wait till Terraria makes more organised drop code

Chicken-Bones:
After reading, that's actually a rather innovative solution, I'll think about it but it's probably more suited for an individual mod with some method hooking

JavidPack:
We could easily do this by subclassing UnifiedRandom and swapping out Main.rand before PreNPCLoot and swapping it back after. That said, I also don't know if we want to add a luck stat to NPC itself. Ideally a luck stat would be a stat of the Player, but as such a thing is up to debate, doing this in tModLoader itself means we are designing gameplay elements into the mod loader. Is luck tied to the NPC that died, the player who hit it last, the closest player, the world? All of these are suitable designs, so the decision is hard to make. 

By the way, here is an example of subclassing UnifiedRandom HumanGamer: https://github.com/JavidPack/RecipeBrowser/blob/master/LootUnifiedRandom.cs

Chicken-Bones:
Ah very good point

HumanGamer:
That is true but because it uses Next for everything and there is no difference between NextInt/NextBool there is no way to change the drop rate without messing up the quantity. There are also random checks in there that can't be higher or lower than certain values. Subclassing would not work because of that.

Chicken-Bones:
Terraria's loot system is very poorly designed and not really worth the name.
At least this would massively reduce patches (which is very important for us)

HumanGamer:
I just noticed an issue where items with low drop quantities would only ever drop their minimum drop amount, and made another commit to fix that.

Chicken-Bones:
It was a good idea, but hard to implement well. Come back to it when 1.4 hits, since there'll be an actual loot table by the looks of things

Jofairden:
Luck will be possible with the 1.4 changes, if we edit things enough to allow it (according to Yorai)

PR #352, opened 2018-05-21
New icons for the Mod Browser to replace placeholders

Flashkirby:
### Description of the Change

New mod browser filter icons for all/client/server/nosync.

### Alternate designs

https://i.imgur.com/YMqjfd7.png
Selection down to vote; question mark for no sync picked over red outlines because there wasn't a real leader in votes for that, and that design came first.

Graphic update should make it easier to tell what the filter actually does at a glance. Also I like shiny new icons.

### Benefits

The mod browser filter icon will look slightly more fancy, maybe more intuitive.

### Possible drawbacks

The mod browser filter icon will be slightly less readable for some, maybe less intuitive.


Flashkirby:
Why is 6f700b4 in this pull request

JavidPack:
I think it is because you are still on the same branch of your fork and you didn't rebase. It doesn't matter though.

Chicken-Bones:
For this reason you should create a new branch for any PRs. If you commit to master, and then we update master, then they have different history. To fix your master (make it the same as ours) you'll need to reset your local branch to be a copy of remotes/upstream/master, and then push to remotes/origin/master with the --force flag

PR #367, opened 2018-06-24
Volcano Tremor Fix

Pokecreeper:
### Description of the Change

Volcano Tremor code updated to work with the new version of ModifyTransformMatrix. I also added a comment with some code showing how to do things with ModifyTransformMatrix that the Volcano Tremor doesn't implement. (Namely, zooming the screen and flipping it)

### Benefits

Modders can use the Example Mod to help them learn how to use the new version of ModifyTransformMatrix.

### Applicable Issues

Comment could appear off to some, due to most additional comments like that being one-line comments, not comment blocks.

### Comments

I'm honestly wondering why this hasn't been updated. It was very easy to do, as I just had to look at SpriteViewMatrix.cs in the source code and just make minor tweaks.

### DIFF
--- a/ExampleMod/ExampleMod.cs
+++ b/ExampleMod/ExampleMod.cs
@@ -231,9 +231,8 @@ namespace ExampleMod
 		float targetOffsetX = 0;
 		float targetOffsetY = 0;
 
-		// Volcano Tremor
-		/* To be fixed later.
-		public override Matrix ModifyTransformMatrix(Matrix Transform)
+		// Volcano Tremor (Fixed!)
+		public override void ModifyTransformMatrix(ref SpriteViewMatrix Transform)
 		{
 			if (!Main.gameMenu)
 			{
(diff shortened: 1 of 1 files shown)


JavidPack:
Did you build this or test this? Transform.TransformationMatrix is read only and that Matrix constructor doesn't exist. Also, you forgot the using statement.

Pokecreeper:
I thought it could be edited, my bad.

strongrex2001:
I've tried to create such effects in my mod now. I found that changing transform matrix in the new matrix class is originally diabled but if I use reflection to change it, the results are right. So maybe changing _transform into public can solve the problem?

JavidPack:
I kinda got stumped when I tried to fix this. Matrix transforms need to happen in certain orders, so trying to allow for special things like rotation happen while preserving mod compatibility is hard.

If we only allow z rotation centered on the center of the screen and also make an additionalTranslation field, I think we can allow the commonly desired functionality.

GoodPro712:
Can this be closed

Pokecreeper:
Closing

Issue #382, opened 2018-07-14
GetHealValue hook for changing healing values of potions conditionally


JavidPack:
Would be useful. Same for mana.

Currently the item.healLife can't adjust to conditions. An item like "heal half life" can't be made.

PR #383, opened 2018-07-15
New Healing Hooks

DarkLight66:
### Description of the Change

* **Added**:

  * **ModPlayer, ModItem and GlobalItem GetHealLife Hooks**: allows modifying the amount of health a life restoring item heals on use.

  * **ModPlayer, ModItem and GlobalItem GetHealMana Hooks**: allows modifying the amount of mana a mana restoring item restores on use.

  * **GetHealLife and GetHealMana methods for the vanilla Player class**: similar to the `GetWeaponDamage` and other similar methods, these return the healing value of an item after being modified by the respective hooks.

There is no proper way of modifying the amount of health/mana a healing item restores without either adding the amount on top, causing the item to show 2 combat texts for the healing on use, or modifying `item.healLife`/ `item.healMana` directly.

### Benefits

Allow modders to add healing items that restore heal/mana based on percentage of maximum life / mana, missing life / mana, buffs, equipment, etc.

### Applicable Issues

#382 

### Sample Usage

See the new Example Healing Potion item in the included ExampleMod edits.


JavidPack:
Looks great. Were you able to test these?

Jofairden:
These are great additions to tModLoader, thanks! Should we add the possibility to enable healing even without setting healLife above 0? What if GetHealLife returns something higher than 0? (What if I want to make items that normally don't heal, heal now?) I think since we are altering it anyway, we might as well enable that functionality. (same goes for mana)

Jofairden:
I guess that makes sense. Alright, I'll approve.

SheepishShepherd:
This would make my issue I mentioned about a week ago easier. I currently have some complicated code that allows me to do this, however its on an accessory, and the only way I managed to apply it to the accessory was to search through each accessory slot. My code doesn't account for modded accessory slots though, so I would actually love to have this.

Jofairden:
Thx again.

PR #401, opened 2018-08-06
Auto-download/update dependencies

Itorius:
### Description of the Change

Added the ability for users for download/update mods with their dependencies, thus eliminating the potential complaints about mods not working due to missing dependencies.

### Applicable Issues

I replaced downloading code `UIModDownloadItem` with a call to `UIModBrowser.DownloadMods` which *might* break some things, but from the testing which I have done it seems to work.


Itorius:
### Bug
When downloading mods without dependencies it shows 'Missing mods [nothing] required by [mod].'

Jofairden:
@Itorius In hindsight, do you have time to finish this sometime soon? We could then include it in next release.

Agrair:
I tried to fix this on my tML fork, try and see if that works

Jofairden:
@JavidPack  @Chicken-Bones Do we still like to have this feature?

Jofairden:
@Itorius I implemented this in cd84dfc 
Your current PR could not be accepted due to current code changes in the mod browser and download manager.

PR #407, opened 2018-09-04
Adding ability to offset trash slot in Vanilla UI.

VectorAdw:
### Description of the Change
Simple change that adds ability to offset trash slot when creating UI that mimics Chest Inventory by using (Point16) Main.trashSlotOffset variable. If it's set to anything else than Point16.Zero and there is no Vanilla chest open it will be taken into account.

### Alternate designs
N/A

### Why this should be merged into tModLoader / Benefits
I think it will be useful to mods that want to provide user with vanilla-like UI.

### Possible drawbacks
None I could think of.

### Applicable Issues
Change will not reset slot position to default, so mod author needs to keep in mind when closing UI.


JavidPack:
I don't see how this does anything useful. Any modder can make a trash slot in their own UI using the VanillaItemSlotWrapper and the TrashItem Context. Why would we want one specific inventory element to be positioned in this manner?

VectorAdw:
Yes you are right, you can create one without any problems.
But I did not see any possibility to hide Vanilla one (please correct me if I'm missing something obvious).
If you want to make UI that is just below Vanilla Inventory (like 1st row of the chest) then the UI will overlay (or be under) the Vanilla Trash slot (I can upload example screenshot when I get home).

VectorAdw:
This is example of something I made with this change: [offset](https://drive.google.com/file/d/1qCyOp76mfxvn5SNyQ4COkFXTFu5TGOha/view)
And example of existing mod I have downloaded from browser: [no offset](https://drive.google.com/file/d/1NWXI1isKp7i8ogmdsCxaqbXklDdEojMc/view)
Both have Vanilla-like storage UI modded in, but not being able to move/hide Trash slot is leaving a gap.

JavidPack:
K, that's what you mean. I get it now. I'll look at this and think about it some more before next release.

Jofairden:
I think something like this is a fine addition, but some of us will have to look at it thoroughly to review it and also make sure there maybe isn't an even better solution :)

Chicken-Bones:
I have no idea why I didn't look at this sooner, but it should definitely go in the next release, just gotta see the conflict. It's a clean implementation

PR #417, opened 2018-10-02
When compiling mods, check for {Dep}.Windows.dll and {Dep}.Mono.dll

0x0ade:
### Description of the Change

Allow compiling any given mod's `Windows.dll` and `Mono.dll` against the appropriate version of a given `dllReferences` by checking for `.Windows.dll` and `.Mono.dll` versions of the library.

### Alternate designs

I've initially thought about allowing modders to supply `dllWindowsReferences` and `dllMonoReferences`, but dropped the idea due to the additional work that needs to be done by the modder and the higher possibility for errors.

This is required to compile mods which use libraries using XNA / FNA. Otherwise, both `Windows.dll` and `Mono.dll` would use the XNA version of the library.

### Benefits

Mods will be able to use libraries which in themselves depend on XNA / FNA, as long as both versions are provided.

### Possible drawbacks

There could be conflicts if a library `CoolLibrary.dll` ships with `CoolLibrary.Windows.dll` and / or `CoolLibrary.Mono.dll`, which serve some internal purpose. Renaming the files should resolve the issue though, as the file name doesn't affect assembly resolution.

~~From what I can tell, tModLoader loads all libraries when the mod loads. This causes both versions to be loaded, but it shouldn't cause issues: only the version needed for the platform is actually used at runtime.~~  
Resolved: https://github.com/blushiemagic/tModLoader/pull/417/commits/74da2ff5bb932ffc489eee42b4cbc46c4c31f1ae

### Sample Usage

- In `build.txt`, `dllReferences = CoolLibrary`
- In `lib` folder,
    - `CoolLibrary.dll` (XNA)
    - `CoolLibrary.Mono.dll` (FNA)
- Compiled `Windows.dll` uses `CoolLibrary.dll`
- Compiled `Mono.dll` uses `CoolLibrary.Mono.dll`

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/AssemblyManager.cs
+++ b/patches/tModLoader/Terraria.ModLoader/AssemblyManager.cs
@@ -118,8 +118,14 @@ namespace Terraria.ModLoader
 				{
 					modFile.Read(TmodFile.LoadedState.Code);
 
+					string suffixPlat = ModLoader.windows ? ".Windows.dll" : ".Mono.dll";
 					foreach (var dll in properties.dllReferences)
-						LoadAssembly(EncapsulateReferences(modFile.GetFile("lib/" + dll + ".dll")));
+					{
+						LoadAssembly(EncapsulateReferences(
+							modFile.GetFile("lib/" + dll + suffixPlat) ??
+							modFile.GetFile("lib/" + dll + ".dll")
+						));
+					}
 
 					assembly = LoadAssembly(EncapsulateReferences(modFile.GetMainAssembly()), modFile.GetMainPDB());
 					NeedsReload = false;
--- a/patches/tModLoader/Terraria.ModLoader/ModCompile.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ModCompile.cs
@@ -427,7 +427,22 @@ namespace Terraria.ModLoader
 			refs.AddRange(GetTerrariaReferences(tempDir, forWindows));
 
 			//libs added by the mod
-			refs.AddRange(mod.properties.dllReferences.Select(refDll => Path.Combine(mod.path, "lib/" + refDll + ".dll")));
+			string suffixPlat = forWindows ? ".Windows.dll" : ".Mono.dll";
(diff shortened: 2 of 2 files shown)


Chicken-Bones:
Can you add a check using EndsWith when loading to save some memory? You're right that there wouldn't be any resolution issues but we may as well

0x0ade:
I've just noticed that `AssemblyManager.LoadAssemblies` required a change anyway, as it blindly appended `.dll` to the name in `dllReferences`.

0x0ade:
Just checked for anything else using `dllReferences` that might require special suffixing.
`EncapsulateName` would require it if the assembly's "internal" name differs from the name in `dllReferences`, which it shouldn't anyway.

Jofairden:
@Chicken-Bones do your recent changes address this issue? Let's close it if it's solved.

Chicken-Bones:
It's not solved, but the compile split has been refocused from Windows/Mono to XNA/FNA. I believe TerrariaHooks was the only one needing such a thing, so it's not high priority, though it wouldn't be too difficult to add now

tansheron:
I would like to voice my support for this change, especially the part in 74da2ff5bb932ffc489eee42b4cbc46c4c31f1ae.

I split my mod into multiple projects to manage complexity. I have to add all of them to the dllReferences line, regardless of which framework they need, because they won't be included in the mod package otherwise. The mod uses reflection to search all loaded assemblies for types which meet certain criteria. When it attempts to get the exported types for an assembly which references the wrong framework, a FileNotFoundException is thrown. Terraria seems to log the exception even if I catch and eat it, which pollutes the log and will make people think there's a problem when there actually isn't. If I could use the .XNA.dll/.FNA.dll naming scheme to prevent the wrong versions of the assemblies from being loaded in the first place, that would be the ideal solution for me.

Chicken-Bones:
Note that the current build split is XNA/FNA and not Windows/Mono. For standard tML, those currently align (because XNA is used on Windows, and FNA elsewhere) but for 64bit tML FNA is used on Windows as well, and we want to be open to future experimentation with .NET Core and FNA. If you still want this I'm happy to take a version with XNA/FNA split

PR #428, opened 2018-10-18
WIP: Add extra update hooks

Leemyy:
### Description of the Change
This PR adds several new mid-update hooks to allow for actions to be taken after a certain point in the update process. These hooks are specifically intended to be called on both client and server in both single- and multiplayer. There may be some pre-existing hooks which are similar to some of the ones I added, but none of them consistently get called in all of those cases.

### Alternate designs
Some of the new hooks may be better placed in other Mod Classes; in this regard I would be happy to hear suggestions and implement them. For now I placed them all in ModWorls since I felt like they are best left in one place, and ModWorld seemed like the most logical option for that.
`PreUpdateEntities` could also be caled `PreUpdatePlayers`, but I wanted the name to express that it comes before any type of Entity gets updated.

The naming scheme for these methods could also be changed to indicate what comes after them, instead of what came before (e.g. `PreUpdateNPCs` instead of `PostUpdatePlayers`). But I usually find myself wanting to perform actions after a certain type of game object had been updated, so I picked the names to reflect what I expect to be the most common useage.

These hooks add possibilities for functionality that was previously either impossible to achieve or only possible with very ugly/hacky measures.

### Benefits
I already mentioned most benefits in the sections above. In general this allows to run code on all Players/NPCs/Items/etc. after they have been updated, but before any other action is taken. This is especially useful when you want to do some processing of said objects, which can only be done after all of them have been updated. It also allows for unrelated logic to be run in between update steps.

### Possible drawbacks
None, as far as tML itself goes. I can see some people potentionally missunderstanding what these methods are for, and attempting to use them in unfit situations. To this end I am planning to add descriptive documentation comments, as soon as I have come up with a good descriptions. (Suggestions would be appreciated)

### Applicable Issues
N/A

### Sample Usage
I can't think of a simple one at the moment, all the stuff I want this for is either complex or very nieche.
I will hopefully be able to add a good Example later.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/ModWorld.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ModWorld.cs
@@ -110,7 +110,7 @@ namespace Terraria.ModLoader
 
 		/// <summary>
 		/// Use this if you want to do something before anything in the World gets updated.
-		/// Gets called after UI updates, but before anything in the World (Players, NPCs, Projectiles, Tiles) gets updated.
+		/// Called after UI updates, but before anything in the World (Players, NPCs, Projectiles, Tiles) gets updated.
 		/// <para />
 		/// When <see cref="Main.autoPause" /> is true or <see cref="Main.FrameSkipMode" /> is 0 or 2, the game may do a partial update. This means that it only updates menus and some animations, but not the World or Entities. This hook - and every hook after it - only gets called on frames with a full update.
 		/// <para />
@@ -121,7 +121,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after all Players are updated, but before any NPCs get updated.
+		/// Called after Players got updated, but before any NPCs get updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="PreUpdateAnything" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateNPCGore" />.
@@ -131,7 +131,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after all NPCs are updated, but before any Gores get updated.
+		/// Called after NPCs got updated, but before any Gores get updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdatePlayerNPC" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateGoreProjectile" />.
@@ -141,7 +141,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after all Gores are updated, but before any Projectiles get updated.
+		/// Called after Gores got updated, but before any Projectiles get updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateNPCGore" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateProjectileItem" />.
@@ -161,7 +161,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after all Items are updated, but before any Dust gets updated.
+		/// Called after Items got updated, but before any Dust gets updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateProjectileItem" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateDustTime" />.
@@ -171,7 +171,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after all Dust is updated, but before Time effects get updated.
+		/// Called after Dust got updated, but before Time (day/night, events, etc.) gets updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateItemDust" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateTimeWorld" />.
@@ -181,7 +181,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after Time effects are updated, but before the World gets updated.
+		/// Called after Time got updated, but before the World gets updated.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateDustTime" />.
 		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateInvasionNet" />.
@@ -205,7 +205,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after Invasions are updated. The only thing that is updated after this is the Network.
+		/// Called after Invasions got updated. The only thing that is updated after this is the Network.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateTimeWorld" />.
 		/// If you are looking to hook even after the Network is updated, see <see cref="PostUpdateEverything" />.
@@ -215,7 +215,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Gets called immediately after the Network is updated, this is the last thing that happens in an update.
+		/// Called after the Network got updated, this is the last hook that happens in an update.
 		/// <para />
 		/// If you are looking to hook an earlier part of the update process, see <see cref="MidUpdateInvasionNet" />.
 		/// </summary>
--- a/patches/tModLoader/Terraria.ModLoader/ModWorld.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ModWorld.cs
@@ -109,88 +109,6 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Use this if you want to do something before anything in the World gets updated.
-		/// Called after UI updates, but before anything in the World (Players, NPCs, Projectiles, Tiles) gets updated.
-		/// <para />
-		/// When <see cref="Main.autoPause" /> is true or <see cref="Main.FrameSkipMode" /> is 0 or 2, the game may do a partial update. This means that it only updates menus and some animations, but not the World or Entities. This hook - and every hook after it - only gets called on frames with a full update.
-		/// <para />
-		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdatePlayerNPC" />.
-		/// </summary>
-		public virtual void PreUpdateEntities()
-		{
-		}
-
-		/// <summary>
-		/// Called after Players got updated, but before any NPCs get updated.
-		/// <para />
-		/// If you are looking to hook an earlier part of the update process, see <see cref="PreUpdateEntities" />.
-		/// If you are looking to hook a later part of the update process, see <see cref="MidUpdateNPCGore" />.
-		/// </summary>
-		public virtual void MidUpdatePlayerNPC()
-		{
-		}
-
-		/// <summary>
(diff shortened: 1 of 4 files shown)


Leemyy:
I am neither expecting nor wanting this to get merged rigth away. I mostly submitted the pull request to get feedback and suggestions for possible improvements.

Jofairden:
Thank you for your contribution first and foremost.

I'd like to emphasize on the fact that, we'd really like to see use-cases for each and every hook you suggest. Because, and don't get me wrong, but everything we add to tML adds necessary maintenance work to it. We really appreciate your contribution, we really do, but once it is merged it's mostly up to us to keep things maintained.

You added a lot of hooks, and don't get me wrong because again we appreciate it, but if some of these hooks have really niche use-cases it might be better to use other hooks (or 'work-arounds') for them.

> The naming scheme for these methods could also be changed to indicate what comes after them, instead of what came before (e.g. PreUpdateNPCs instead of PostUpdatePlayers)

You're right, the problem here lies in the fact that a modder doesn't necessarily have any knowledge about the chain these hooks are called in. If you call something PreUpdateNPCs, they don't know that that's called after updating players. Therefore you might consider a different naming scheme, such as MidUpdatePlayerNPC, which would more clearly signify that the update happens between Players and NPCs. At least, that's just what popped in my mind.

I think we're happy to accept your hooks, as long as we can find good use-cases for them.

> I am neither expecting nor wanting this to get merged rigth away. I mostly submitted the pull request to get feedback and suggestions for possible improvements.

I'll tag your PR as `WIP:` for now, as is the usual.

Leemyy:
Thank you for the feedback.

First, I would like to say that I think the `MidUpdatePlayerNPC` namig scheme is a really good idea, because it tells people what came before aswell as what will come afterwards without resulting in terribly long or obscure names. It also allows people to easily infer the order in which these happen. I am definitely going to implement that.

Secondly, I can definitely appreciate the fact that this will add extra maintenance work, so I will try to come up with good justifications for all of these - and potetially discard the ones that don't seem necessary.
Three things that I definitely think to be useful are `PreUpdateEntities` and `PostUpdateInvasion` and to a lesser degree `PostUpdateNet`

`PreUpdateEntities` is very useful, not just because it gets called before anything (besides menus and some texture animations) gets updated. `Mod.UpdateUI` can actually be used for that, since it is also guaranteed to get called in every case. But the problem with `Mod.UpdateUI` is, that it gets called before the frame skip code which aborts the Update early. This means that if you have frame skip enabled, it may get called many times, before any Entities are actually Updated and without any Draws happening. This is where `PreUpdateEntities` comes in, I specifically placed it so that it would only get called if Entities are guaranteed to be Updated on the same frame. So it is a good way of doing some action on every actual game tick, before any updates happen.

`PostUpdateInvasion` is useful for the exact opposite. It gets called after everything has been updated, making it perfect for code that can only work after every Entity has been moved/updated.
I personally want this because I have a new kind of Entity that I want to introduce, which is supposed to be able to collide with all other Entities. The reason I can not do this at any earlier point is that otherwise Players/NPCs might move out of collision range in the same frame that a collision has been detected, making it look to the Player as if no collision should have occurred.
The only other way would be to hook into every type of Entity and to check for collision with my things after they get updated. However, this would greatly fragment my code, and also have a big impact on performance.

`PostUpdateNet` actually gets called even after `PostUpdateInvasion`, but I don't think it should get used for the same purpose that I assigned to `PostUpdateInvasion`. Everything in vanilla gets updated before the net update, and I think it'd probably be best to follow suit for consistency.
Having a hook immedeately after the net updates can really come in handy to either make modifications to what has been received, or to manage networking behaviour that would be too complicated to achieve through direct NetSend and NetReceive hooks. Though I cannot claim to have ever encountered the latter, so it might not be something people will have much need for.

Chicken-Bones:
This is actually a really good PR, nice and simple, easy to maintain. You've given us good descriptions for some of the events here, now they need to go in the `/// <summary>` sections in the code as documentation. The summary should say 1. when it's called in relation to other update hooks and terraria functions and 2. what it should be used for.

PreUpdateAnything, I notice it happens after UI updates, skymanager, liquidrenderer, backgrounds, emotes, animations, chat, input, weather, net syncing and inventory, but before stars and clouds. It seems like the most important thing is it happens after any opportunity to pause the game? maybe just relate it to `Main.autoPause` and `Main.FrameSkipMode`?

Perhaps say it happens after rendering updates which still function when the game is paused or frame skipping, but before Player, NPC and AI, Gore, Projectile, Item, World updates and importantly game time.

JavidPack:
I'm not sure I agree with these being in ModWorld, I think they should go in Mod. Thoughts anyone?

Chicken-Bones:
Given there's currently only one world per game and these hooks aren't UI related I'm fine with either

Leemyy:
I am undecided on that aswell @JavidPack.
The only hooks in `ModWorld` that are related to these are `PreUpdate` and `PostUpdate`, whereas the only ones in `Mod` are `UpdateUI`, `UpdateMusic`, and `PostUpdateInput`. I wouldn't consider either of those to mark any real precedence, so I guess we'll have to decide where they make more sense.
My rationale for putting them in `ModWorld` was that they only get called on frames on which the World also gets updated, whilst all the hooks in `Mod` also get called when autopause is active - `UpdateUI` even gets called when frame skip activates.

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
```diff
@@ -108,6 +108,88 @@ public virtual void ResetNearbyTileEffects()
 		{
 		}
 
+		/// <summary>
+		/// Use this if you want to do something before anything in the World gets updated.
+		/// Gets called after UI updates, but before anything in the World (Players, NPCs, Projectiles, Tiles) gets updated.
+		/// <para />
+		/// When <see cref="Main.autoPause" /> is true or <see cref="Main.FrameSkipMode" /> is 0 or 2, the game may do a partial update. This means that it only updates menus and some animations, but not the World or Entities. This hook - and every hook after it - only gets called on frames with a full update.
+		/// <para />
+		/// If you are looking too hook a later part of the update process, see <see cref="MidUpdatePlayerNPC" />.
```
Chicken-Bones: *to

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
```diff
@@ -108,6 +108,88 @@ public virtual void ResetNearbyTileEffects()
...
+		/// </summary>
+		public virtual void PreUpdateAnything()
+		{
+		}
+
+		/// <summary>
+		/// Gets called immedeately after all Players are updated, but before any NPCs get updated.
+		/// <para />
```
Chicken-Bones: *immediately

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
```diff
@@ -108,6 +108,88 @@ public virtual void ResetNearbyTileEffects()
...
+		/// If you are looking too hook an earlier part of the update process, see <see cref="PreUpdateAnything" />.
+		/// If you are looking too hook a later part of the update process, see <see cref="MidUpdateNPCGore" />.
+		/// </summary>
+		public virtual void MidUpdatePlayerNPC()
+		{
+		}
+
+		/// <summary>
+		/// Gets called immedeately after all NPCs are updated, but before any Gores get updated.
```
Chicken-Bones: Probably just "Called after X is updated and before Y is updated"

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
```diff
@@ -108,6 +108,88 @@ public virtual void ResetNearbyTileEffects()
...
+		{
+		}
+
+		/// <summary>
+		/// Gets called immedeately after Time effects are updated, but before the World gets updated.
+		/// <para />
+		/// If you are looking too hook an earlier part of the update process, see <see cref="MidUpdateDustTime" />.
+		/// If you are looking too hook a later part of the update process, see <see cref="MidUpdateInvasionNet" />.
+		/// </summary>
+		public virtual void MidUpdateTimeWorld()
```
Chicken-Bones: Perhaps, "UpdateBetweenXAndY" moght be better?

Leemyy: Yeah, I thought that the `MidUpdateXY` was nice and short whilst still conveying the meaning. And in cases where people aren't sure, they can always look at the documentation.
I was thinking that using plurals like `MidUpdatePlayersNPCs` might be better than `MidUpdatePlayerNPC`, but I'm not sure.

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
Leemyy: Sometimes I just can't even, the **i** is literally pronounced the same as the **e** right before it. Why do you have to do this, English?

Leemyy:
I was thinking about maybe adding a line to the effect of "Called in single- and multiplayer on the server and every client." This information about when a hook gets executed is actually something that a lot of hooks could use in their documentation. Maybe I should make a seperate PR adding that information to some other hooks aswell?

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
Jofairden: I think that is up for debate/personal preference. Someone can also think "between Player and NPC means between any Player and NPC" which is basically the same thing as using plurals.

patches/tModLoader/Terraria.ModLoader/ModWorld.cs
Leemyy: I made those lines more concise. I think they should be good now.

Leemyy:
One use case that applies to a lot of these hooks would be freezing some, or all, of any kind of Entity. This can easily be achieved by temporarily swapping the desired Entities, or even the whole array, for dummy ones. It might even be possible to have a time stop effect that works on both server and client with this. Though I am not 100% sure on that. I definitely know that I have seen people try to make something like this on multiple occasions. Most of the current approaches are not completely reliable, though because some parts of the Update still happen, even if you return false in `PreUpdate`.
I can make some additions to Example Mod and add them to this PR, if you think this would be a good example.

Jofairden:
> Is there anything besides adding an Example that still needs to be done?
I think naming and documentation are fine now. Unless you have any strong objections to the way they currently are. I can still move the hooks to Mod, if you would prefer that. But I'll wait for you to come to a concensus on that before I move everything around.

I think we all (devs) just need to agree on the naming, usefulness and placing of the methods.
For example I still think `PreUpdateAnything` is a bit controversial. As the docs state, you don't actually pre update before everything (or anything), because UI updates before this hook. Maybe `PreUpdateEntities` is better. (you can see how this sort of is a never ending discussion)

Also, you have  redundant < / Para > tags in every XMLdoc.

Leemyy:
Okay, I finally did that rename.
If you want to I can remove the `<para />` tags, but as I said they add line braks and, in my opinion, really add to the documentations' readability.
Do you think I should add something to Example Mod - like the ability to freeze Entities I mentioned above? Or is that better placed in a different PR?

JavidPack:
Thinking of an example is hard, since we want ExampleMod to be accessible, but these hooks are fairly specialized. Maybe just a simple example that counts active NPC in the before NPC and after NPC hooks, then finds the difference, to find how many npc were killed last update. Something simple like that. 

Freezing Entities is something we'd rather not have everyone abusing just because they saw it in ExampleMod

Leemyy:
Good point, I hadn't thought of it that way.
Freezing any kind of Entity will definitely require a load of special cases that need to be accounted for. And having a bunch of mods that freeze stuff all over the place would indeed be a recipe for chaos.

Leemyy:
Is there anything that still needs to be added to/changed about this PR?
Or are you just undecided on whether you want to merge it yet?

Jofairden:
>Or are you just undecided on whether you want to merge it yet?

We're undecided yet. We'll let you know, don't worry

PR #431, opened 2018-10-22
WIP: Adding Alt Ore Functionality

imkSushi:
### Description of the Change

New function:
`public static void WorldGen.AddAlt<T>(T alt)`
This takes in one of the new classes:
AltCopper, AltIron, AltSilver, AltGold, AltCobalt, AltMythril, AltAdamantite
When initialised, these all must take the ore tile to be placed and the bar item for use in certain treasure chests. The AddAlt function adds the new alt ores to a list of them, each contained within the relevant class. When the world is generated, before the 'Reset' task, the vanilla ores are added, and then in the 'Reset' task, the ores to use are chosen and assigned to the relevant variables.

In addition, the ore in the MiningExplosivesBiome is now specified in a compatible way with this.

### Alternate designs

One alternative was to have seperate `AddAltCopper()`, `AddAltIron()` etc.. I decided not to do this, as the method I decided on made it possible in the near future for mods to easily let other mods add alt ores to them.

DO NOT MERGE THIS YET - It isn't ready. I just want some feedback before I implement many other alts (i.e. HM ores, Hellstone, Evil Biomes, Other Biomes, more...)
I'm also not quite sure where the AddAlt function, and the AltCopper, AltIron, etc. classes should go.

Eventually, it should be merged, as this adds in functionality that cannot be added by a mod, even with TerrariaHooks (at least, not easily, and certainly not flexibly)

### Benefits

Mods will be able to add alt ores of their own, and hopefully (eventually) alt other stuff as well

### Possible drawbacks

Not applicable (as far as I can think)

### Applicable Issues

Any mod that replaces the 'reset' task in World Gen will break this feature.

### Sample Usage

In the ModWorld of ExampleMod:
```C
public override void ModifyWorldGenTasks(List<GenPass> tasks, ref float totalWeight)
{
	int resetTask = tasks.FindIndex(genpass => genpass.Name == "Reset");
	if (reset != -1)
	{
		tasks.Insert(reset, new PassLegacy("Adding Alt Ores", delegate
		{
			WorldGen.Add(new AltCopper(mod.TileType<ExampleCopperTierOre>(), mod.ItemType<ExampleCopperTierBar>()));
		}));
	}
}
```
This would give the following behaviour:
1/3 of the time, Copper Ore spawns
1/3 of the time, Tin Ore spawns
1/3 of the time, ExampleCopperTierOre spawns instead

Of course, if you add more copper tier ores, those probabilities will decrease.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
+++ b/patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
@@ -1,0 +1,58 @@
+using System;
+using System.Collections.Generic;
+using Terraria.ID;
+
+namespace Terraria.ModLoader
+{
+	public class AltAdamantite
+	{
+		public readonly int oreTileType;
+		public readonly string oreBlessingText;
+		public readonly float abundanceMultiplier;
+		private static readonly List<AltAdamantite> AltAdamantites = new List<AltAdamantite>();
+		internal static bool vanillaAltsAdded = false;
+		public static AltAdamantite ChosenAltAdamantite { get; private set; }
+
+		public AltAdamantite(int oreTileType, string oreBlessingText, float abundanceMultiplier = 1f)
+		{
+			this.oreTileType = TileLoader.TileCount > oreTileType ? oreTileType : throw new Exception("oreTileType is an invalid tile");
+			this.oreBlessingText = oreBlessingText;
+			this.abundanceMultiplier = abundanceMultiplier;
+		}
+
+		public static AltAdamantite GetAltAdamantite()
+		{
+			if (!vanillaAltsAdded)
+			{
+				WorldGen.AddAlt(new AltCobalt(TileID.Adamantite, Lang.misc[14].Value));
+				WorldGen.AddAlt(new AltCobalt(TileID.Titanium, Lang.misc[23].Value, 0.9f));
+			}
+			return ChosenAltAdamantite ?? (ChosenAltAdamantite = AltAdamantites[WorldGen.genRand.Next(AltAdamantites.Count)]);
+		}
+
+		internal static void UnchooseAltAdamantite()
+		{
+			ChosenAltAdamantite = null;
+		}
+
+		internal static void Add(AltAdamantite alt)
+		{
+			AltAdamantites.Add(alt);
+		}
+
+		internal static bool unchosen()
+		{
+			return ChosenAltAdamantite == null;
+		}
+
+		internal static void tryFind(int oreTileType)
+		{
+			foreach (var altAdamantite in AltAdamantites)
+			{
+				if (altAdamantite.oreTileType != oreTileType) continue;
+				ChosenAltAdamantite = altAdamantite;
+				break;
+			}
+		}
+	}
+}
--- a/patches/tModLoader/Terraria.GameContent.Biomes/MiningExplosivesBiome.cs.patch
+++ b/patches/tModLoader/Terraria.GameContent.Biomes/MiningExplosivesBiome.cs.patch
@@ -1,0 +1,26 @@
+--- src/Terraria\Terraria.GameContent.Biomes\MiningExplosivesBiome.cs
++++ src/tModLoader\Terraria.GameContent.Biomes\MiningExplosivesBiome.cs
+@@ -1,6 +_,7 @@
+ using Microsoft.Xna.Framework;
+ using System;
+ using Terraria.GameContent.Generation;
++using Terraria.ModLoader;
+ using Terraria.World.Generation;
+ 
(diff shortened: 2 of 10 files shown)


Jofairden:
I'm not sure if I follow the use of this? Can't you just use Worldgen.rand to randomly spawn differing ore types?

imkSushi:
You can't add alternatives to the vanilla ores (i.e. Copper, Tin and your ore, or maybe Adamantite, Titanium and Troxinium (from Avalon)) - those were hard-coded, but this now lets you add ores that could replace the vanilla ones.

Chicken-Bones:
It might be better to use a world gen hook to just go through the world and replace all the copper with your copper (if you roll the 1/3 chance)

I'm not sure many people would use this, or it woule scale well. You could also just retexture existing copper if you wanted

imkSushi:
That breaks any mod that gens copper ore, etc. More particularly, I want it for the HM ores, which you can't get the correct 'blessing' text if you use that method.

Chicken-Bones:
Ahh, because they're only for combat, it doesn't really matter if your world doesn't have them, and other mods probably won't depend on them

imkSushi:
Yeah, but aesthetics are everything. Seriously though, it makes the mod look much lower quality, plus there are plenty of tml features that only a couple mods use.

Chicken-Bones:
Better off with a single Alt class and an enum

imkSushi:
Yeah, but if you want to add in ALTs for different things, then you require different data to implement it, so having one class and an enum means you can't do this easily.

imkSushi:
Fair. I'll change that then when I have time. Do you want me to drop the alt phm ores?

Chicken-Bones:
They can stay, provided we have a way to recipe group them. Also rip gem staves

patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
```diff
@@ -0,0 +1,58 @@
...
+		public readonly int oreTileType;
+		public readonly string oreBlessingText;
+		public readonly float abundanceMultiplier;
+		private static readonly List<AltAdamantite> AltAdamantites = new List<AltAdamantite>();
+		internal static bool vanillaAltsAdded = false;
+		public static AltAdamantite ChosenAltAdamantite { get; private set; }
+
+		public AltAdamantite(int oreTileType, string oreBlessingText, float abundanceMultiplier = 1f)
+		{
+			this.oreTileType = TileLoader.TileCount > oreTileType ? oreTileType : throw new Exception("oreTileType is an invalid tile");
```
Jofairden: I personally think this should be abstracted

patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
```diff
@@ -0,0 +1,58 @@
...
+			this.oreBlessingText = oreBlessingText;
+			this.abundanceMultiplier = abundanceMultiplier;
+		}
+
+		public static AltAdamantite GetAltAdamantite()
```
Jofairden: Similarly to above, I feel like this should be abstracted. E.g: AltOre class, GetAltOre()

patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			ChosenAltAdamantite = null;
+		}
+
+		internal static void Add(AltAdamantite alt)
+		{
+			AltAdamantites.Add(alt);
+		}
+
+		internal static bool unchosen()
```
Jofairden: no camelCase, UpperCase in c#

patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			return ChosenAltAdamantite == null;
+		}
+
+		internal static void tryFind(int oreTileType)
```
Jofairden: idem dito.

Jofairden: Also TryFind is a bit vague, TryFind what? Could use a better name.

patches/tModLoader/Terraria.ModLoader/AltAdamantite.cs
Jofairden: Also, it's usually preferred to `IsChosen()` and then !IsChosen() for conciseness.

Chicken-Bones:
What is free time. Can always come back to it I guess, but probably needs a rethink/design

PR #458, opened 2018-12-10
Make the installer add the execute permission to tModLoaderServer on unix systems

### LINKED ISSUE #402: "host and play" causes game to crash
### Description

"Host and Play" Causes the game to crash (with and without mods)

### Steps to Reproduce

1. Click on Multiplayer
2. Click on "Host and Play"
3. Choose character and world
4. Server Settings: "Enabled, Friends, Off"
5. Type password(or not) 
6. Click "Accept" and the game crashes

### Expected behavior 

Game doesn't crash

### Actual behavior

Game crashes with no error log showing up or anything

### Reproduction frequency

Every time

### Versions

Ubuntu 18.04(?)
Terraria (1.3.5.2)
TmodLoader (0.10.1.5)

### Mods

Thorium
Overhaul

Both tested independently of each other, and tested when no mods are enabled

### Additional Information

Terraria installed via Steam

Silly511:
### Description of the Change

When tModLoader is installed on linux or mac using the jar installer, trying to use Host & Play will crash the game due to the tModLoaderServer file not having the right permissions. This PR makes the installer add the execute permission to tModLoaderServer on linux and mac, which fixes that issue.

### Alternate designs

None.

Because it fixes a bug that prevents people on unix systems from using tModLoaderServer.

### Benefits

Bugs will be fixed.

### Possible drawbacks

I haven't tested this on linux because I don't own a linux computer, so I don't know how well it works on linux.

### Applicable Issues

Possibly #402.

### Sample Usage

None.

### DIFF
--- a/installer2/LinuxInfo.java
+++ b/installer2/LinuxInfo.java
@@ -4,6 +4,8 @@ public class LinuxInfo
 {
     public static void main(String[] args)
     {
+        File directory = getInstallDir();
+        
         String[] files = new String[]
         {
             "Terraria.exe",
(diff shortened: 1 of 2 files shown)


JavidPack:
Question, does this match up with the permissions of the vanilla exe? I'm pretty sure the script is executable, just can't remember if the .bin files need to be executable as well.

PR #462, opened 2018-12-19
More town NPC interaction hooks.

Agrair:
### Description of the Change

New hooks in `ModPlayer`, PostSellItem, PostBuyItem, CanSellItem, CanBuyItem, PostNurseHeal, and ModifyNurseHeal.

### Alternative implementations
Hooks for GlobalNPC/ModNPC instead

This allows players to edit the way players exchange money with town NPCs.

### Benefits

Selling and buying is now changeable by modders, as is when the player can heal themselves with the nurse.

### Possible drawbacks

If many mods are enabled, and some of them are returning false to a `Can-` hook, then the player may have some problems.

### Applicable Issues
What is said above, and 

### Future Hopes
1. Something for Example Mod.
2. Something for the stylist.


Agrair:
I am currently looking for a way to edit the price of of a bought item.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3621,15 +3621,50 @@
...
- 												Main.player[Main.myPlayer].DelBuff(l);
- 												l = -1;
+-											{
+-												Main.player[Main.myPlayer].DelBuff(l);
+-												l = -1;
++										Main.player[Main.myPlayer].statLife += health;
++										if (removeDebuffs) {
++											for (int l = 0; l < 22; l++) {
++												int num24 = Main.player[Main.myPlayer].buffType[l];
++												if (Main.debuff[num24] && Main.player[Main.myPlayer].buffTime[l] > 0 && BuffLoader.CanBeCleared(num24)) {
```
Chicken-Bones: Please avoid changing indentation unless necessary. Breaking good formatting rules is acceptable to mimise patch size and improve maintainability.

Chicken-Bones:
We definitely need some use cases for these hooks so we can evaluate them. The implementation of the hooks looks quite neat and reasonable.

patches/tModLoader/Terraria/Main.cs.patch
Agrair: I don’t really know what happened there, VS started changing the tabulation and then reverted back to normal ten minutes later.

patches/tModLoader/Terraria.ModLoader/ModPlayer.cs
```diff
@@ -815,5 +815,57 @@ public class ModPlayer
...
+
+		}
+
+		/// <summary>
+		/// Return false to prevent a transaction. Called before the transaction.
+		/// </summary>
+		/// <param name="vendor"></param>
+		/// <param name="item"></param>
+		/// <returns></returns>
+		public virtual bool ModifyBuyItem(NPC vendor, Item item) => true;
```
Agrair: The item param should be ref.

Agrair: I wonder if I should add a manual “ref int price” param as well.

Chicken-Bones:
My main concern here is why you'd want to modify an item or price on purchase.  Maybe you want to do something akin to rolling modifiers? Maybe that's a use case.

You definitely shouldn't be changing the item or price, because then the user isn't going to get the same item they thought they were buying.

Agrair:
I understand what your concern, I'll change `Modify-` to `Can-`.

patches/tModLoader/Terraria.ModLoader/ModPlayer.cs
Jofairden: No, the Item parameter does not need the `ref` keyword, it is already a reference. For an int param, yes it would need `ref`, since the primitive type 'int' is copied, not referenced. But no, you should not add it (as mentioned by chicken-bones), this hook should not be for modifying prices directly.

Agrair:
The modifier rolling is an interesting idea, but I think that would fundamentally change the way shops work. I'll look into it though.

Agrair:
I have realized a mistake: now that I changed some of the Modify- hooks into Can- hooks, there is no more need for the extra code that allows for referenced parameters. I will fix this as soon as I can.

Chicken-Bones:
The code quality and implementation of this PR is very good. We just need examples of how these hooks will be used, so we know how to maintain them and what they should be used for

Agrair:
Sure thing!

JavidPack:
We still need examples. It isn't clear if some of these hooks couldn't be done with already existing hooks.

Agrair:
Added something really basic for `CanBuyItem`.

Agrair:
Now that I think about it, I still need to do some more thorough testing.

Agrair:
I know this is preferred, but should I add an example for each and every hook? I can't really think of too many use cases.

Chicken-Bones:
Examples on ExampleMod are great, and a few are needee, but when we asked for use cases, we meant you can just describe them here in a comment. And we need some for wach hook (otherwise why have the hook)?

Agrair:
Say the player wanted to count how many times the item is bought or add up the prices: I did that with stock and inflation. Another idea is that if a modder wants to have extra price than vanilla money, they could use the buying hooks to prevent a trade if they don’t meet the extra criteria as well as consume it. As for selling items to NPCs, players could trigger certain events (I was thinking this could be useful in storylines where the item is key to progression). Another is that modders could prevent the selling of certain items because the vendor does not want the item (the Wizard wouldn’t want Minishark, and the Arms Dealer wouldn’t want a Crystal Ball).

patches/tModLoader/Terraria.UI/ItemSlot.cs.patch
```diff
@@ -17,22 +17,59 @@
...
++					&& num == 3)
++				{
 +					Main.mouseItem = inv[slot].Clone();
 +					Main.mouseItem.stack = 1;
- 					if (inv[slot].buyOnce)
- 					{
--						Main.mouseItem.Prefix((int)inv[slot].prefix);
++					if (Main.mouseItem.buyOnce)
++					{
 +						Main.mouseItem.value *= 5; // preserve item value for items sold to the shop
```
Chicken-Bones: may need to be wary here, if we can change sell values and such, *5 may not be appropriate anymore

patches/tModLoader/Terraria.UI/ItemSlot.cs.patch
```diff
@@ -17,22 +17,59 @@
...
+@@ -556,12 +_,15 @@
+ 						}
+ 					}
  				}
- 				else if (num == 3)
- 				{
+-				else if (num == 3)
+-				{
 -					Main.mouseItem.netDefaults(inv[slot].netID);
+-					if (inv[slot].buyOnce)
```
Chicken-Bones: may-as well keep this as `inv[slot]` for better patch context/updating

patches/tModLoader/Terraria.UI/ItemSlot.cs.patch
Jofairden: this is how vanilla handles buy/sellprice , by *5 / 5
modfying price should modify item.price (afaik), so this should scale appropriately

Agrair:
One sec, I'll consider all the conversations, which are all good feedback, then resolve the conflict.

patches/tModLoader/Terraria.ObjectData/TileObjectData.cs.patch
```diff
@@ -105,7 +105,7 @@
 +				// legacy compatibility, mismatch isn't a problem here
 +				if (Height < _tileObjectCoords.heights.Length && ModContent.LoadingMod.tModLoaderVersion < new Version(0, 11))
 +					return;
-+
```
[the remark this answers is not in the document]
Agrair: Did [this](https://github.com/blushiemagic/tModLoader/pull/462/commits/dcc72bb1f7913a40b519e655e6c08212519396c0) fix it?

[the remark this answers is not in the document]
Agrair: Hmmm, probably shouldn't have done that using the online editor. *Probably*.

patches/tModLoader/Terraria.UI/ItemSlot.cs.patch
Agrair: wait, so does that mean I don't have to change anything?
and yes, the idea was that the modder should just modify item.price, which I've decided is the most optimal way.  If the modder wants to change the item itself, it's their fault if users complain, I guess.

patches/tModLoader/Terraria.ObjectData/TileObjectData.cs.patch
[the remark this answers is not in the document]
Jofairden: No, chicken bones is talking about the removal of that newline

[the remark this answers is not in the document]
Agrair: oh

Agrair:
Shit, I think I did break the setup. I can’t regen source anymore since I made that edit with the online editor. How can I delete a commit?

patches/tModLoader/Terraria.ObjectData/TileObjectData.cs.patch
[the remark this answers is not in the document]
Agrair: I think I did break the setup. I can’t regen source anymore since I made that edit with the online editor. I’ll need a few more days before I’m at my keyboard again and can fix it, sorry ‘bout that.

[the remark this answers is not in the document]
Agrair: I think I fixed it with [bb8051a](https://github.com/blushiemagic/tModLoader/pull/462/commits/bb8051aab4631cee296dd50ce8e2192813dafc33)

Chicken-Bones:
I don't actually have time to test the changes I pushed. Please test the hooks, including inflation save/load and multiplayer syncing (some netsync code is there, but I don't think the inflation will actually propagate if a different player does the buying)

Agrair:
Ok

patches/tModLoader/Terraria.ModLoader/ModContent.cs
```diff
@@ -24,7 +24,7 @@ public class ModContent
 
 		internal static readonly IDictionary<string, ModHotKey> modHotKeys = new Dictionary<string, ModHotKey>();
 
-		private static void SplitName(string name, out string domain, out string subName) {
+		public static void SplitName(string name, out string domain, out string subName) {
```
Agrair: weird but ok

patches/tModLoader/Terraria.ModLoader/ModPlayer.cs
```diff
@@ -815,5 +815,57 @@ public class ModPlayer
...
+		/// <summary>
+		/// Return false to prevent the player from healing.
+		/// </summary>
+		/// <param name="nurse"></param>
+		/// <param name="health">How much health the player gains.</param>
+		/// <returns></returns>
+		public virtual bool ModifyNurseHeal(NPC nurse, ref int health, ref int price, ref bool removeDebuffs)
+			=> true;
+
+		// TODO: Do the buy hooks.
```
Agrair: get rid of this

Chicken-Bones:
I added a server config. Modconfig has been merged into master

JavidPack:
K, there was a lot about inflation and stock that was causing me headaches thinking about all the ways it could go wrong. I replaced it with a simpler example that I was capable to properly testing.

Agrair:
It's cool, my fault anyway for implementing them like that.

PR #471, opened 2019-01-06
Update Chinese translation

Dierney:


Jofairden:
Thanks for the fast updates 👍

PR #483, opened 2019-01-20
Fixes a NRE crash if an item has a prefix > 83 but no ModPrefix

AJMJ2012:
Just a quick fix to prevent crashes on save.
However prefixes that would normally crash are instead ignored and discarded.

Example of crash:
1: Set a prefix on an item >83 and >ModPrefix defined max, such as 255.
2: Exit the world and you'll crash to desktop without a message when saving is done.
3: You can still load the world and character, but regardless if you have the offending mod loaded or not, it will still crash on exit save.

My mod, Item Qualities and Durability used to set a prefix of 255 if the prefix is 0 to trick the tooltips in displaying stat changes when the item's quality changes. However it seems since ModPrefix was added, it would just crash like this.
Either way, said mod will be using a different way of doing this rather than the invalid prefix method.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.IO/ItemIO.cs
+++ b/patches/tModLoader/Terraria.ModLoader.IO/ItemIO.cs
@@ -35,8 +35,10 @@ namespace Terraria.ModLoader.IO
 
 			if (item.prefix >= PrefixID.Count) {
 				ModPrefix modPrefix = ModPrefix.GetPrefix(item.prefix);
-				tag.Set("modPrefixMod", modPrefix.mod.Name);
-				tag.Set("modPrefixName", modPrefix.Name);
+				if (modPrefix != null) {
+					tag.Set("modPrefixMod", modPrefix.mod.Name);
+					tag.Set("modPrefixName", modPrefix.Name);
+				}
 			}
 
 			if (item.stack > 1)


AJMJ2012:
Example of crash:
1: Set a prefix on an item >83 and >ModPrefix defined max, such as 255.
2: Exit the world and you'll crash to desktop without a message when saving is done.
3: You can still load the world and character, but regardless if you have the offending mod loaded or not, it will still crash on exit save.

My mod, Item Qualities and Durability used to set a prefix of 255 if the prefix is 0 to trick the tooltips in displaying stat changes when the item's quality changes. However it seems since ModPrefix was added, it would just crash like this.
Either way, said mod will be using a different way of doing this rather than the invalid prefix method.

Jofairden:
I wonder why we are saving item.prefix and getting by that rather than assembly/mod name and prefix name. If loading changes, aren't we losing the right prefix anyway? I might be missing something here

Jofairden:
So, confirming that this is a 'bug'. We need to change how the prefix saves to ensure it will work correctly with different load orders. I will close this PR with a commit when it's fixed.

Jofairden:
@AJMJ2012 thanks and sorry for the delay. Your fix was totally fine.

Chicken-Bones:
This is a perfectly fine hotfix, but prefixes will still be lost or altered when mods are unloaded and reordered.

Jofairden:
@Chicken-Bones Why are they? They are being saved and loaded by their mod name and prefix name, so it will return the correct ID based on those.

Jofairden:
Related fix: https://github.com/blushiemagic/tModLoader/commit/be135d493b326b79f49d512648694db92ffa0e09

PR #484, opened 2019-01-20
Fixed filename-less compilation errors (like CS1704) crashing the game.

Mirsario:
Title.

This fixes this > https://i.imgur.com/cf5UfSI.png < crash I've mentioned in discord earlier.


Chicken-Bones:
Nono it's valid

Mirsario:
The NRE was being thrown in CompilerError.ToString() (originally called in Logger.Log method on the error as an object), which definitely does sound weird, since it's a .NET class iirc.

Jofairden:
Okay, I understand. It is indeed part of .NET itself. My problem with this is that we now only safeguard for filename. Apparently any of these fields can be null, so I'd rather see something like a 'SafeLog' method that safeguards us against any of them being null.

Issue #488, opened 2019-02-06
Way to force HiDef graphics profile.


chatrat12:
Working on Terraria3D. Once I made my shaders a little more complex, it wasn't hard to pass the maximum shader instructions in SM2, I had to upgrade to SM3. Unfortunately, Reach does not support SM3. More complex shaders require a way to force Terraria into HiDef. Unfortunately this can only be done before the base game starts loading its own content. 

Wrote the stuff below first but just came up with a better idea. Can we enable HiDef by default if the graphics device can handle it? There is no advantage to defaulting to reach. 

Been pondering ways to do this. I think the simplest and most straight forward way is to just enable HiDef if the graphics device can support it. There is no advantage to using Reach by default if the machine is capable of HiDef.

I can still let the user, use reach mode but with less shader features and I can do inform them of that if the game was not able to boot with HiDef enabled.

Chicken-Bones:
Seems sound, please submit a pull request.

PR #489, opened 2019-02-06
Forces HiDef if grahpics device supports it

chatrat12:
### Description of the Change
Forces the graphics profile into HiDef if the graphics device supports it.

### Alternate designs
Not applicable

Allows mods to utilize SM3 shaders and other HiDef features if graphics device supports it.

### Benefits
SM3 shaders!

### Possible drawbacks
None

### Applicable Issues
#488 

### Sample Usage
```cs
Effect shader = null;

if (graphics.GraphicsProfile == GraphicsProfile.HiDef)
	shader = new Effect(graphics.GraphicsDevice, sm3ByteCode);
else // Device could not start HiDef, fallback to sm2 shader
	shader = new Effect(graphics.GraphicsDevice, sm2ByteCode);
```

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -653,14 +653,16 @@
  			}
  			return result;
  		}
-@@ -7482,15 +_,11 @@
+@@ -7482,15 +_,17 @@
  			}
  			Main.Configuration.Get<bool>("Support4K", ref Main.Support4K);
  			bool flag = Main.Support4K && base.GraphicsDevice.Adapter.IsProfileSupported(GraphicsProfile.HiDef);
--			if (GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Width <= 1920 && GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Height <= 1200)
--			{
--				flag = false;
--			}
++			/* tModLoader will attempt to use HiDef even if the computer resolution doesn't need it to support newer shaders.
+ 			if (GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Width <= 1920 && GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Height <= 1200)
+ 			{
+ 				flag = false;
+ 			}
++			*/
  			if (Main.Support4K && flag)
  			{
  				Main.SetGraphicsProfile(GraphicsProfile.HiDef);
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -653,11 +653,23 @@
...
  		}
-@@ -7490,7 +_,7 @@
+@@ -7480,17 +_,9 @@
  			{
- 				Main.SetGraphicsProfile(GraphicsProfile.HiDef);
+ 				TexturePackSupport.Enabled = true;
  			}
+-			Main.Configuration.Get<bool>("Support4K", ref Main.Support4K);
+-			bool flag = Main.Support4K && base.GraphicsDevice.Adapter.IsProfileSupported(GraphicsProfile.HiDef);
+-			if (GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Width <= 1920 && GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Height <= 1200)
```
JavidPack: Rather than delete all this vanilla code, I think we should just comment out the check for display resolution, these 3 lines. Support4K defaults to true, so this would allow users to continue to use the Support4K config flag as designed to limit the HiDef profile.

chatrat12:
Whoops, totally overlooked that. I agree!

chatrat12:
So I made it so that if `Support4k` is false, it still uses HiDef (if supported) but max values are 1920, 1200, and 2048.

JavidPack:
I don't see why you added more changes. If someone sets Support4k to false, that's because they are having issues with HiDef profile and have manually attempted to prevent HiDef from being enabled.  

Isn't the idea that Hidef should have no drawbacks, so simply disabling the resolution checks that are in place should be all we need?

chatrat12:
So my thoughts behind it is that users don't know anything about graphics profiles and it's kind of confusing if disabling 4k also disables rendering capabilities (MRTs, shaders, ect)

As far as issues caused by HiDef, I'm not aware of any draw backs. If your graphics device can handle it, nothing in base Terraria changes other than max screen size.

This kind of makes the supports 4K flag irrelevant, but I didn't want to have it do nothing, that is also potentially confusing to the user. That's why I added the extra limitation to HiDef.

What are your thoughts?

Chicken-Bones:
I think the supports4K flag was originally a misnomer for 'enable hidef on large screens'

You should just keep the misnomer and ignore screensize. The only reason someone would turn it off is if there were issues.

Chicken-Bones:
If anyone makes an issue with confusion we can make separate flags

chatrat12:
Rodger, I'll get to it tonight. :D

chatrat12:
Yeah, I'll do a write up when I can :D

PR #496, opened 2019-02-23
 Update Chinese translation

Dierney:
Well, I'm glad I finally understand the "LoadErrorMajorVersionMismatch"......

### DIFF
--- a/patches/tModLoader/Terraria.Localization.Content.zh-Hans.tModLoader.json
+++ b/patches/tModLoader/Terraria.Localization.Content.zh-Hans.tModLoader.json
@@ -157,7 +157,7 @@
 		"MBOfflineWithReason": "- 离线({0})",
 		"MBBusy": "繁忙",
 		"MBUnknown": "未知",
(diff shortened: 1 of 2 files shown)


Jofairden:
Maybe it's helpful if you join our chat? I mean, if changes keep coming you keep updating the translation files which is kind of tedious. It would be much better to just to it at the end no?

Dierney:
> 也许如果你加入我们的聊天会有帮助吗？我的意思是，如果不断变化，你会不断更新翻译文件，这有点乏味。到最后没有它会好得多吗？

Yup, but I don't know what to do. Pull request is the quickest way I can think of.

Dierney:
> Maybe it's helpful if you join our chat? I mean, if changes keep coming you keep updating the translation files which is kind of tedious. It would be much better to just to it at the end no?

Yup, but I don't know what to do. Pull request is the quickest way I can think of.

JavidPack:
Thanks again for being very diligent with this.

Jofairden:
> Yup, but I don't know what to do. Pull request is the quickest way I can think of.

What I meant is, waiting before translating because we keep adding new strings

Dierney:
> What I meant is, waiting before translating because we keep adding new strings

Well, no problem. Just notify me before the new version is released.

PR #505, opened 2019-04-03
hiding boots using the robes

orian34:
Adding a method to hide the boots while wearing it.
I hope it's not the wrong place to put it.


Jofairden:
It sounds like you want to hide the boots layer when wearing the robe. We need to verify if changing player.shoe is the right edit to make, or that we should hide a PlayerLayer.

orian34:
Mh, you might know better.
That's what i used for a legless leggings, by reusing what vanilla does for vanity things similar to it.

PR #509, opened 2019-04-07
Change IDK to projectile id.

wooden-utensil:
Comment change.


JavidPack:
It's not though. 10 is purification powder. Since useammo is there, any positive number works. 10 is probably a relic of past versions of Terraria that needed it

wooden-utensil:
> It's not though. 10 is purification powder. Since useammo is there, any positive number works. 10 is probably a relic of past versions of Terraria that needed it

What? Never knew that. Great now I can delete this pull request and fix my current code.

PR #526, opened 2019-04-22
New Traveling Merchant Addition

SheepishShepherd:
### Description of the Change
The code provided adds a new NPC mimicking the traveling merchant. Most of its setup is done in ModWorld, which includes spawn conditions, shop item generating, setting a time for it to spawn in with variation, spawn chance, and chat announcements. Comments are provided for understanding.

### Alternate designs
Not Applicable

ExampleMod could use an alternative town NPC, and a traveling merchant could take its place. Of course the code is open to alterations.

### Benefits
Having new traveling merchants in mods would be neat, as it can bring more originality and potential. A night time only traveling NPC, that needs to leave before sunrise for one example. A flock of traveling merchants spawning in could be troublesome, but in the end it that comes to a mod's design.

### Possible drawbacks
Not Applicable, only adds code that isnt too hard to understand. Nothing is changed or made more complex.

### Applicable Issues
The code provided does not match the behaviors of the Vanilla traveling merchant. One thing that could be changed is resetting the conditions on a timer, instead of each day.

### Sample Usage
This addition can help modders understand how to create randomized shop items and spawning NPCs based on conditions. It can also be a good example to point to when someone asks about traveling merchants in general.


JavidPack:
K, I'll be looking at this. Forcing homeless sounds like a hacky solution, gonna see if we actually need a new field or hook to do this properly.

SheepishShepherd:
Found this in
private void AI_007_TownEntities()

```
if (this.type == 368)
				{
					this.homeless = true;
					if (!Main.dayTime)
					{
						this.homeTileX = (int)(base.Center.X / 16f);
						this.homeTileY = (int)(base.position.Y + (float)base.height + 2f) / 16;
						if (!flag3 && this.ai[0] == 0f)
						{
							this.ai[0] = 1f;
							this.ai[1] = 200f;
						}
						flag = false;
					}
				}
```
Not sure if this is actually part of the AI but I assumed it was

Chicken-Bones:
that diff is creeping me out, but I'm happy to merge this

Jofairden:
Agree with @Chicken-Bones , the diff is messed up somehow (doesn't look like it's an indent problem). I would love to see it solved and see the actual changes.

JavidPack:
If you click the "hide whitespace changes" it looks fine.

Jofairden:
Nice tip @JavidPack

ExampleMod/NPCs/ExampleTravelingMerchant.cs
```diff
@@ -0,0 +1,163 @@
...
+
+		public override void HitEffect(int hitDirection, double damage) {
+			int num = npc.life > 0 ? 1 : 5;
+			for (int k = 0; k < num; k++) {
+				Dust.NewDust(npc.position, npc.width, npc.height, mod.DustType("Sparkle"));
+			}
+		}
+
+		public override bool CanTownNPCSpawn(int numTownNPCs, int money) {
+			for (int i = 0; i < Main.maxNPCs; i++) {
```
Jofairden: Just `return Main.npc.Any(x => x.type == mod.NPCType("ExamplePerson"));`

ExampleMod/NPCs/ExampleTravelingMerchant.cs
```diff
@@ -0,0 +1,163 @@
...
+				if (Main.npc[i].type != mod.NPCType("ExamplePerson")) continue; // If the npc is not the Example Person, move onto the next npc
+				return true; // If one Example Person exists, Example traveling Merchant can spawn
+			}
+			return false;
+		}
+
+		public override string TownNPCName() {
+			switch (WorldGen.genRand.Next(4)) {
```
Jofairden: Why do you use WorldGen.genRand specifically here?

ExampleMod/NPCs/ExampleTravelingMerchant.cs
```diff
@@ -0,0 +1,163 @@
...
+			}
+			else if (ExampleWorld.travelerItems[2] == 3) {
+				shop.item[nextSlot].SetDefaults(mod.ItemType("ExamplePickaxe"));
+			}
+			nextSlot++;
+		}
+
+		public override void AI() {
+			npc.homeless = true; // Make sure it stays homeless
+			base.AI();
```
Jofairden: base call is unnecessary, ModNPC does nothing.

Jofairden:
Don't be scared by my review, your work is highly appreciated!

Chicken-Bones:
I don't like how the merchant shop is split over many places, with hardcoded lengths/chances and if chains.
Also, if you kill it currently it will instantly respawn (I might've just introduced that accidentally with my changes).
I think the shop should be saved with the NPC not the world. If it dies and comes back the next day it can get a new shop, etc.

SheepishShepherd:
I do think that would be better. I'll have to change how randomizing the shop works anyways because if a mod has a Traveling Merchant who sells items from other mods, if a player were to unload a mod that a shop supports while a merchant is active, it can either pull up the wrong items or even worse crash with an out of bounds error. I'll have to save Item instances themselves and possibly do a check for unloaded items.

Chicken-Bones:
Yeah pretty much my thoughts, I didn't have time to do it right now because there were considerations. You can get away with just storing the item identifier string though "Mod:name" I think we have a utility function for that

Chicken-Bones:
I think you just reverted all my changes

SheepishShepherd:
Oops, sorry I haven't fully grasped GitHub yet.

Jofairden:
I'm seeing spaces instead of tabs, please change to tabs (see https://github.com/blushiemagic/tModLoader/blob/master/CONTRIBUTING.md#your-files-use-tab-indendation-keep-tabs-of-size-4)

ExampleMod/ExampleWorld.cs
```diff
@@ -27,46 +27,51 @@ public class ExampleWorld : ModWorld
...
-			}
-
-			return new TagCompound {
-				["downed"] = downed,
-				["travelerShopItems"] = travelerItems
-			};
-		}
-
-		public override void Load(TagCompound tag) {
+            travelerItems = new int[3]; // Initializing to set 3 random items (Starts off with zeros)
```
Jofairden: I recommend doing `new int[] { 0, 0, 0 }` to reduce linecount

ExampleMod/ExampleWorld.cs
```diff
@@ -412,98 +417,99 @@ public class ExampleWorld : ModWorld
...
+            bool conditionsMet = false; // If no conditions, either remove or set true
+            for (int i = 0; i < Main.maxNPCs; i++)
+            {
+                if (Main.npc[i].type != mod.NPCType("ExampleTravelingMerchant") || !Main.npc[i].active) continue; // We dont care about NPCs that arent the Example Traveling Merchant or not active
+                travelerExists = true; // We found one, so set it to true
+            }
+
+            if (Main.dayTime && randomTime == 0) // Each day we will set a new SpawnChance and TimeVariation
+            {
+                canSpawn = Main.rand.Next(4) == 0; // 4 = 25% Chance
```
Jofairden: `Main.rand.NextBool(4)` is still preferred (1 in 4 chance)

ExampleMod/ExampleWorld.cs
```diff
@@ -102,400 +104,437 @@ public override TagCompound Save()
...
+
+                // Annouce that the traveler has spawned in!
+                if (Main.netMode == NetmodeID.SinglePlayer) Main.NewText(Language.GetTextValue("Announcement.HasArrived", traveller.FullName), 50, 125, 255);
+                else NetMessage.BroadcastChatMessage(NetworkText.FromKey("Announcement.HasArrived", traveller.GetFullNetName()), new Color(50, 125, 255));
+            }
+        }
+
+        public static NPC FindNPC(int npcType)
+        {
+            foreach (NPC npc in Main.npc)
```
Jofairden: I prefer changing it to a oneliner: `return Main.npc.FirstOrDefault(npc.type == npcType);`

ExampleMod/ExampleWorld.cs
```diff
@@ -1,501 +1,703 @@
...
+        }
+
+        public List<Item> GetTravelerShop()
+        {
+            List<Item> RandomizedItems = new List<Item>();
+
+            // For each slot we add a switch case to determine what should go in that slot
+            Item nextItem = new Item();
+
+            switch (WorldGen.genRand.Next(2))
```
Jofairden: Why `WorldGen.genRand` again? Use `Main.rand`

Jofairden:
Hmm I see things like canSpawn are removed. Those are better than in-line Main.rand quality checks. (It is clearer)

Chicken-Bones:
I see you like american spelling.
I'd like to see some arrays or lists or something for the traveller item selection rather than switch statements, but it's good, we've hassled you enough.
I'd also like to see as much as possible moved into the traveller class, but it's not your fault that ExampleMod didn't have good design there to start with. There's too many things in ExampleWorld simultaneously, and it needs a good refactor.

The only thing I want you to check before it goes in, is currently I believe if you kill the npc he'll instantly respawn. May need to reset the spawn time once he does spawn for the day. Also the despawn time is set to 6pm, but it's not checked during spawn time. If an event runs all day, the traveller could theoretically spawn and then despawn instantly.

SheepishShepherd:
The spelling was very inconsistent, just wanted to fix that up.
As for moving the code, I could just throw everything into a function and have it all within the ModNPC class and reference it from there. Would certainly reduce the amount of space in ModWorld.

I'll be sure to test everything again. Haven't tested the changes much since I added the pull request,

Chicken-Bones:
Definitely, toss everything, except the spawn time field in there. Save the shop with the NPC too.

Solid work👌

SheepishShepherd:
After a bit of testing, your suspicion is correct. The traveling merchant respawns nearly instantly after dying. I suppose a "hadSpawned" check would be necessary, though not sure if 1 traveling merchant per day is the way we want to go.
On a side note, the items are broken with the current switch case. They appear to all get the defaults of the last chosen item, so I added each Item instance separately.

`Item FirstSlotItem = new Item();
Item SecondSlotItem = new Item();
Item ThirdSlotItem = new Item();`

If there is another approach though, like you were mentioning with lists, im open to it.

Chicken-Bones:
Just set the `travelerSpawnTime` to `double.MaxValue` when it spawns. You can also just re-assign `item` with a `new Item()` after each random pick. The alternative would be to have something like `ExampleTravelingMerchant.SetupShopPool()` which is called from `ExampleMod.Load()` and creates a couple of static lists, and then you just call `rand.Next(list.Count).clone()`

SheepishShepherd:
Everything Ive tested seems to work. If anyone else wants to test go ahead, Im still listening in for issues.

Jofairden:
It looks like you changed a lot of styling like braces unnecessarily. It makes it very hard to view all the actual changes.
Sure, I think GetTravelerShop could be better, but for the small amount of items in it now _it is not a problem_. I personally like seeing SetDefaults and SetStaticDefaults as the first methods though, it's usually the most common thing to start with.

Anyways, enough nitpicking. Looks kinda fine.

SheepishShepherd:
Bumping this for questionable issue. The new traveling merchant addition shows up on the TownNPC list, unlike the vanilla traveling merchant. Should this be taken care of?

PR #528, opened 2019-04-24
Add GoG support

HTishev:
### Description of the Change

Modify Setup App to handle differences between Steam and GoG executable. Provide corresponding patch files for initial platform patching. Existing patches for "merged", "terraria" and "tModLoader" have been modified to handle the additional patching step.
### Alternate designs

N/A
### Benefits

GoG users can enjoy tModLoader
### Possible drawbacks

On specific occasions changes in terraria project file may lead to merge conflicts. 
### Applicable Issues

N/A
### Sample Usage

N/A


Chicken-Bones:
We want to keep things as consistent and reliable as possible we only want to decompile a single version. I think the purpose of your setup changes is to allow developing tML when decompiling the GoG version? If so we'd rather not incur the maintenance cost. 

Is the only thing required to disable the steam check? If so I have a good plan

HTishev:
I think I understand your idea better now. You are correct the main difference is the Steam Social API. The other issue is some sort of UID in projectiles but your patches are removing that. 
Based on that input all this probably can be handled with just project build settings to compile GoG exe.

PR #533, opened 2019-04-29
Small Fix to GlobalItem/ModItem PickAmmo

DarkLight66:
### Description of the Change

Made it so GlobalItem/ModItem PickAmmo receives the item that's using the ammo as an argument, and made the old methods obsolete.

Given that the item using the ammo is important when deciding which projectile to shoot from the weapon in some cases, like flame throwers and rocket launchers in vanilla, not having that as an argument for the aforementioned methods looks like an oversight. Although most of the time the item using the ammo is the item being held by the player, it is still better to have said item as an argument for a better understanding of the method's usage.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
@@ -206,17 +206,17 @@ namespace Terraria.ModLoader
 		/// <summary>
 		/// Allows you to modify the projectile created by a weapon based on the ammo it is using.
 		/// </summary>
-		/// <param name="item">The ammo item</param>
 		/// <param name="shooter">The item that is using this ammo</param>
+		/// <param name="item">The ammo item</param>
 		/// <param name="player">The player using the item</param>
 		/// <param name="type">The ID of the projectile shot</param>
 		/// <param name="speed">The speed of the projectile shot</param>
 		/// <param name="damage">The damage of the projectile shot</param>
 		/// <param name="knockback">The speed of the projectile shot</param>
-		public virtual void PickAmmo(Item item, Item shooter, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
+		public virtual void PickAmmo(Item shooter, Item item, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
 		}
 
-		[Obsolete("PickAmmo now has a shooter parameter that represents the item using the ammo.", true)]
+		[Obsolete("PickAmmo now has a shooter parameter that represents the item using the ammo.")]
 		public virtual void PickAmmo(Item item, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
 		}
 
--- a/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
@@ -207,13 +207,13 @@ namespace Terraria.ModLoader
 		/// Allows you to modify the projectile created by a weapon based on the ammo it is using.
 		/// </summary>
 		/// <param name="shooter">The item that is using this ammo</param>
-		/// <param name="item">The ammo item</param>
+		/// <param name="ammo">The ammo item</param>
 		/// <param name="player">The player using the item</param>
 		/// <param name="type">The ID of the projectile shot</param>
 		/// <param name="speed">The speed of the projectile shot</param>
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Haven't checked code but rationale is solid

Jofairden:
Can't you just use Main.LocalPlayer.selectedItem ? (Or whatever the equivalent is)
This would avoid breaking API changes.

DarkLight66:
> 
> 
> Can't you just use Main.LocalPlayer.selectedItem ? (Or whatever the equivalent is)

Yes, but this isn't something immediately obvious to someone new to modding, and this way allow mods to call `PickAmmo` using a different item from the selected one.

> If it's still a warranted change, we should postpone it to the update after the next since we were trying to avoid breaking changes.

I don't have a problem with this, the next update is already big as it is.

> Also I would like the shooting item parameter to be the first in that case, so it is coherent with all the other Global class' methods.

Will do.

patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
```diff
@@ -206,12 +206,17 @@ public class GlobalItem
...
-		/// <param name="type"></param>
-		/// <param name="speed"></param>
-		/// <param name="damage"></param>
-		/// <param name="knockback"></param>
+		/// <param name="player">The player using the item</param>
+		/// <param name="type">The ID of the projectile shot</param>
+		/// <param name="speed">The speed of the projectile shot</param>
+		/// <param name="damage">The damage of the projectile shot</param>
+		/// <param name="knockback">The speed of the projectile shot</param>
+		public virtual void PickAmmo(Item shooter, Item item, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
```
Chicken-Bones: change `item` to `ammo`

patches/tModLoader/Terraria.ModLoader/ItemLoader.cs
```diff
@@ -439,11 +439,14 @@ private class HookList
...
-		public static void PickAmmo(Item item, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
-			item.modItem?.PickAmmo(player, ref type, ref speed, ref damage, ref knockback);
+		public static void PickAmmo(Item shooter, Item item, Player player, ref int type, ref float speed, ref int damage, ref float knockback) {
+			item.modItem?.PickAmmo(shooter, player, ref type, ref speed, ref damage, ref knockback);
+			item.modItem?.PickAmmo(player, ref type, ref speed, ref damage, ref knockback); // deprecated
 
-			foreach (var g in HookPickAmmo.arr)
-				g.Instance(item).PickAmmo(item, player, ref type, ref speed, ref damage, ref knockback);
+			foreach (var g in HookPickAmmo.arr) {
+				g.Instance(item).PickAmmo(shooter, item, player, ref type, ref speed, ref damage, ref knockback);
```
Chicken-Bones: Unfortunately, HookPickAmmo is built to only include those items which actually override the hook from the base class, so you'll probably need 2 separate hook lists

DarkLight66: Oh, i didn't even think about that when i was making this, thanks.

DarkLight66: Wait, not sure if i understood what you meant, but i realized that i needed two hooklists, one for the new and one for the deprecated PickAmmo methods, because each hooklist only grabs one overload.

patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
DarkLight66: Done.

Jofairden: And then change back `shooter` to `item`? Shooter sounds like the player (the shooter) but is later down the line (Discussion please)

Chicken-Bones:
Yeah, just label it `weapon`

patches/tModLoader/Terraria.ModLoader/GlobalItem.cs
DarkLight66: Calling it `item` may be confusing since normally `item` would be the item the current GlobalItem instance belongs to, but that would be `ammo` in this case

DarkLight66:
I wanted to avoid calling it weapon since not all items in vanilla which use ammo deal damage, like the clentaminator for example. However, that may be the best call since most modders are probably thinking of ammo for a weapon when using this hook.

PR #538, opened 2019-04-30
King and Queen statues compatibility

DarkLight66:
### Description of the Change
* Adds `GoToKingStatue` and `GoToQueenStatue` methods to `ModNPC` and `GlobalNPC`, which lets modders allow NPCs to be teleported to the King or Queen statues when triggered.
* Adds `OnGoToStatue` method to `ModNPC` and `GlobalNPC`, which allows adding effects when an NPC teleports to an statue.
* Also removes the limit to the amount of town NPCs each statue can support (used to be 10). Normally, if there are more than 10 of the NPCs a King/Queen statue can teleport, only the first 10, ordered by their slot in the NPCs array, can teleport to the statue.

### Sample Usage
Check `ExamplePerson` for a sample usage of this change.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
@@ -501,6 +501,31 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
+		/// Wether this NPC can be telported to the King statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
+		/// </summary>
+		/// <param name="npc">The NPC</param>
+		public virtual bool? CanGoToKingStatue(NPC npc) {
+			return null;
+		}
+
+		/// <summary>
+		/// Wether this NPC can be telported to the Queen statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
+		/// </summary>
+		/// <param name="npc">The NPC</param>
+		public virtual bool? CanGoToQueenStatue(NPC npc) {
+			return null;
+		}
+
+		/// <summary>
+		/// Allows you to make things happen when this NPC teleports to a King or Queen statue.
+		/// This method is only called server side.
+		/// </summary>
+		/// <param name="npc">The NPC</param>
+		/// <param name="toKingStatue">Whether the NPC was teleported to a King or Queen statue.</param>
+		public virtual void OnGoToStatue(NPC npc, bool toKingStatue) {
+		}
+
+		/// <summary>
 		/// Allows you to modify the stats of town NPCs. Useful for buffing town NPCs when certain bosses are defeated, etc.
 		/// </summary>
 		/// <param name="damageMult"></param>
--- a/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
@@ -501,18 +501,11 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Wether this NPC can be telported to the King statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
+		/// Whether this NPC can be telported a King or Queen statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
 		/// </summary>
 		/// <param name="npc">The NPC</param>
-		public virtual bool? CanGoToKingStatue(NPC npc) {
-			return null;
-		}
-
-		/// <summary>
-		/// Wether this NPC can be telported to the Queen statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
-		/// </summary>
-		/// <param name="npc">The NPC</param>
-		public virtual bool? CanGoToQueenStatue(NPC npc) {
+		/// <param name="toKingStatue">Whether the NPC is being teleported to a King or Queen statue.</param>
+		public virtual bool? CanGoToStatue(NPC npc, bool toKingStatue) {
 			return null;
 		}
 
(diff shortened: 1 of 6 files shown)


Jofairden:
Hmm would methods be favorable? It would be nice if could make additional things happen on or during the teleport, such as visual effects and sounds. Also you **forgot GlobalNPC** :)

wooden-utensil:
Well then I could make an enemy with 99999 damage and health, pop out of the statue, and kill all the players, when they thought that it would summon a queen/king! :trollface: 🤣

DarkLight66:
Yes, that's how those statues work.

wooden-utensil:
> Yes, that's how those statues work.

Aahhhhh! Of course.

patches/tModLoader/Terraria.ModLoader/ModNPC.cs
```diff
@@ -636,6 +628,28 @@ public class ModNPC
...
+		/// Wether this NPC can be telported to the King statue. Returns false by default.
+		/// </summary>
+		public virtual bool CanGoToKingStatue() {
+			return false;
+		}
+
+		/// <summary>
+		/// Wether this NPC can be telported to the Queen statue. Returns false by default.
+		/// </summary>
+		public virtual bool CanGoToQueenStatue() {
```
Chicken-Bones: merge these like `OnGotoStatue`?

patches/tModLoader/Terraria.ModLoader/NPCLoader.cs
```diff
@@ -946,6 +946,84 @@ private class HookList
...
+				case NPCID.DyeTrader:
+				case NPCID.Painter:
+				case NPCID.WitchDoctor:
+				case NPCID.Pirate:
+				case NPCID.LightningBug: // A literal bug
+				case NPCID.Angler:
+				case NPCID.DD2Bartender:
+					defaultCanGo = true;
+					break;
+			}
```
Chicken-Bones: as much as I like this, couldn't you get the default value from the vanilla statement? rather than replacing it just assign it to a flag, that would make it more update compatible

patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
```diff
@@ -500,6 +500,31 @@ public class GlobalNPC
 		public virtual void SetupTravelShop(int[] shop, ref int nextSlot) {
 		}
 
+		/// <summary>
+		/// Wether this NPC can be telported to the King statue. Return true to allow the NPC to teleport to the statue, return false to block this NPC from teleporting to the statue, and return null to use the vanilla code for whether the NPC can teleport to the statue. Returns null by default.
+		/// </summary>
+		/// <param name="npc">The NPC</param>
+		public virtual bool? CanGoToKingStatue(NPC npc) {
+			return null;
```
Jofairden: We could potentially mitigate nullability by making the default behaviour 
`return npc.modNpc?.CanGoToKingStatue() ?? vanillaTrueFalse`
which is your defaultCanGo thing
(nulls may be unnecessary evil here)

patches/tModLoader/Terraria.ModLoader/ModNPC.cs
Jofairden: Agreed

patches/tModLoader/Terraria.ModLoader/NPCLoader.cs
```diff
@@ -946,6 +946,84 @@ private class HookList
...
+
+			foreach (GlobalNPC g in HookCanGoToQueenStatue.arr) {
+				bool? canGo = g.Instance(npc).CanGoToQueenStatue(npc);
+				if (canGo.HasValue) {
+					if (!canGo.Value) {
+						return false;
+					}
+					defaultCanGo = true;
```
Jofairden: Just making sure why you set this to true. Is any NPC supposed to teleport if no Global class returns false? Earlier you are setting defaultCanGo to false if not a mod NPC and not any of the vanilla types.

patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -123,7 +123,7 @@
...
-+																	if (Main.npc[num140].active && (Main.npc[num140].type == 17 || Main.npc[num140].type == 19 || Main.npc[num140].type == 22 || Main.npc[num140].type == 38 || Main.npc[num140].type == 54 || Main.npc[num140].type == 107 || Main.npc[num140].type == 108 || Main.npc[num140].type == 142 || Main.npc[num140].type == 160 || Main.npc[num140].type == 207 || Main.npc[num140].type == 209 || Main.npc[num140].type == 227 || Main.npc[num140].type == 228 || Main.npc[num140].type == 229 || Main.npc[num140].type == 358 || Main.npc[num140].type == 369 || Main.npc[num140].type == 550 || (Main.npc[num140].townNPC && Main.npc[num140].modNPC?.AddToKingStatue == true)))
++																	if (Main.npc[num140].active && NPCLoader.CanGoToKingStatue(Main.npc[num140]))
  																	{
 -																		array[num139] = num140;
 +																		array.Add(num140);
```
Jofairden: Is it safe to do array.Add in these cases? If it does not make any difference I'd rather see it without change to minimize the diff patch

patches/tModLoader/Terraria.ModLoader/NPCLoader.cs
```diff
@@ -946,6 +946,84 @@ private class HookList
...
+
+			return defaultCanGo;
+		}
+
+		private static HookList HookCanGoToQueenStatue = AddHook<Func<NPC, bool?>>(g => g.CanGoToQueenStatue);
+
+		public static bool CanGoToQueenStatue(NPC npc) {
+			bool defaultCanGo = npc.modNPC?.CanGoToQueenStatue() ?? false;
+
+			switch (npc.type) {
```
Jofairden: What about NPCID.Sets.TeleportationImmune ? (Not 100% sure if these are related)

patches/tModLoader/Terraria.ModLoader/NPCLoader.cs
DarkLight66: Any NPC is supposed to teleport if no Global class returns false and at least one of them returns true, and that's what the code is doing.

Jofairden: Right, I forgot you return null otherwise.

patches/tModLoader/Terraria/Wiring.cs.patch
DarkLight66: I'm doing array.Add here because array isn't an `Array` anymore but a `List`. I changed it because before the limit was of 10 NPCs and modded terraria can get bigger than that, hell, vanilla already has more than 10 that can teleport to the king statue.

patches/tModLoader/Terraria.ModLoader/NPCLoader.cs
DarkLight66: That seems to be only used for teleporters in vanilla. Besides, if a modder doesn't want to let their NPC teleport to a statue at all, they can just return false in GlobalNPC.CanGoToStatue.

Jofairden: Approved

patches/tModLoader/Terraria/Wiring.cs.patch
Jofairden: Alright, good change 👍

PR #539, opened 2019-05-05
Add Hooks for mana consumption

DarkLight66:
### Description of the Change

* Adds `GetManaCost` method to `ModItem`, `GlobalItem` and `ModPlayer`, which allows modders to temporarily modify the mana cost of an item that uses mana.
* Adds `OnMissingMana` method to `ModItem`, `GlobalItem` and `ModPlayer`, which allows modders to make things happen when a player doesn't have enough mana for the item they are using.
* Adds `OnConsumeMana` method to `ModItem`, `GlobalItem` and `ModPlayer`, which allows modders to make things happen whenever mana is consumed after using an item.
* Adds `GetManaCost` method to the `Player` class, which let modders check the mana cost of an item while maintaining mod compatibility.
* Adds an overload to `Player.CheckMana` that also takes an item, which is where all of the previous methods are used.
* Adds a sample usage of each of the previous ModLoader methods to ExampleMod, check Example Magic Missile, Mana Heart and Infinity items.
* Small fix to ExamplePerson teleport to statue effect not showing.

* Right now there's no easy way to change the mana cost of an item without messing with its mana value, and doing so breaks tooltips of items with modifiers that affect mana cost.
* There's also no way to reproduce a similar effect as the mana flower without IL editing.

### Sample Usage
In ExampleMod, check the new Example Magic Missile and Mana Heart items, as well as changes to the Infinity item.


Jofairden:
Hmm, I fail to see the use of the GetManaCost hook in ModPlayer. There is nothing you can do with that that you can't do with the others, let's not cause avoidable method bloat. Since an item in question is held by the player, you access the player (and thus your ModPlayer) through the hooks in GlobalItem or ModItem.

DarkLight66:
If it's about usage, there's other hooks in tmodloader that are like this too, which have ModPlayer versions that are kind of unnecessary since there's GlobalItem and ModItem versions that do the job already, like GetWeaponCrit, GetWeaponDamage, GetweaponKnockback, etc. I just added a ModPlayer version of GetManaCost for completion. I can always remove it but i'd like to hear other dev's opinion on this.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -315,6 +315,25 @@
...
  					{
  						if (num == 27 && (this.buffType[j] == num || this.buffType[j] == 101 || this.buffType[j] == 102))
+@@ -2173,6 +_,10 @@
+ 					}
+ 					if (this.inventory[i].mana > 0 && flag)
+ 					{
++						if (this.CheckMana(this.inventory[i], this.GetManaCost(this.inventory[i]), true, true)) {
++							this.manaRegenDelay = (int)this.maxRegenDelay;
++						}
++						goto SkipManaCode;
```
Chicken-Bones: nice goto, I should consider using that to keep patches down

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -6125,4 +6176,46 @@
...
++			}
++			ItemLoader.OnMissingMana(item, this, amount);
++			PlayerHooks.OnMissingMana(this, item, amount);
++			if (this.statMana < amount && this.manaFlower) {
++				this.QuickMana();
++			}
++			if (this.statMana >= amount) {
++				if (pay) {
++					ItemLoader.OnConsumeMana(item, this, amount);
++					PlayerHooks.OnConsumeMana(this, item, amount);
```
Chicken-Bones: I very recently added the `CombinedHooks` class, where I'd like to put any hooks of this style that bridge two modX types, could you just put a fowarder function there?

patches/tModLoader/Terraria/Player.cs.patch
DarkLight66: Thanks, and yeah, to me gotos seem to be the way to go when wanting to reduce patch size

patches/tModLoader/Terraria/Player.cs.patch
DarkLight66: Done.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -6125,4 +6176,46 @@
...
++		}
++
++		public int GetManaCost(Item item) {
++			int mana = (item.type != 127 || !this.spaceGun) ? (int)(item.mana * this.manaCost) : 0;
++			ItemLoader.GetManaCost(item, this, ref mana);
++			PlayerHooks.GetManaCost(this, item, ref mana);
++			return mana >= 0 ? mana : 0;
++		}
++
++		public bool CheckMana(Item item, int amount, bool pay = false, bool blockQuickMana = false) {
```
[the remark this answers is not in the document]
DarkLight66: Also done.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -598,6 +607,33 @@
  							item.newAndShiny = true;
  							Item item2 = Main.player[this.owner].GetItem(this.owner, item, false, false);
  							if (item2.stack > 0)
+@@ -36636,7 +_,7 @@
+ 				}
+ 				if (flag && Main.myPlayer == this.owner)
+ 				{
+-					bool flag2 = player.channel && player.CheckMana(player.inventory[player.selectedItem].mana, true, false) && !player.noItems && !player.CCed;
```
Chicken-Bones: now those are nice diffs

PR #564, opened 2019-06-22
Changed GetManaCost to ModifyManaCost

DarkLight66:
### Description of the Change
Changed `ModItem`, `GlobalItem` and `ModPlayer` `GetManaCost` hooks to `ModifyManaCost`. The new hooks closely resemble the new `ModifyWeaponDamage` hooks, which allow modifying the multipliers that affect the value in the end, rather than the value itself.
The change was done after realizing that the previous hook not only didn't took the mana cost multiplier into account correctly, but also didn't allow to temporarily modify it.

### Sample Usage
See the modified version of the ExampleMagicMissile item.

### DIFF
--- a/ExampleMod/Items/Weapons/ExampleMagicMissile.cs
+++ b/ExampleMod/Items/Weapons/ExampleMagicMissile.cs
@@ -1,3 +1,4 @@
+using System;
 using Terraria;
 using Terraria.ID;
 using Terraria.ModLoader;
@@ -33,13 +34,27 @@ namespace ExampleMod.Items.Weapons
 		// This item's mana usage changes through the day, peaking at 1.5x mana usage at noon, and 0.5x mana usage at midnight.
 		public override void ModifyManaCost(Player player, ref float reduce, ref float mult) {
 			float currentTime = (float)Main.time;
+			// The time at which it changes from day to night and vice versa.
 			int maxTime = Main.dayTime ? 54000 : 32400;
+			// The time at which it is 12pm or 12am, which isn't exactly half the values above.
 			int time12 = Main.dayTime ? 28800 : 18000;
+			// If the time it after 12, recalculate the current time to account for the length of the second half of the day instead of the first.
 			if (currentTime > time12) {
+				currentTime -= time12;
 				time12 = maxTime - time12;
-				currentTime = maxTime - currentTime;
+				currentTime += time12;
 			}
-			float timeMult = 1 + currentTime / time12 * 0.5f * (Main.dayTime ? 1 : -1);
+			// If it is night time, add twice the duration of the current half of the night.
+			if (!Main.dayTime) {
+				currentTime += time12 * 2;
+			}
+			// We divide the current time by the current half of the day or night's time, then by 2.
+			// The result would gradually go from 0 at 4:30am, 0.5 at 12pm, 1 at 7:30pm and 1.5f at 12 am.
+			float timeMult = currentTime / time12 / 2;
+			// Then we multiply the result above by PI, and calculate the sine of that, then multiply it by 0.5f and add 1.
+			// The result would go from 1 at 4:30am, 1.5 at 12pm, 1 at 7:30pm and 0.5f at 12 am, in a sine wave pattern.
+			timeMult = 1 + (float)Math.Sin(timeMult * Math.PI) * 0.5f;
+			// Last, we multiply the current mana cost multiplier of the item by our multiplier.
 			mult *= timeMult;
 		}
 
--- a/ExampleMod/Items/Weapons/ExampleMagicMissile.cs
+++ b/ExampleMod/Items/Weapons/ExampleMagicMissile.cs
@@ -32,28 +32,17 @@ namespace ExampleMod.Items.Weapons
 		}
 
 		// This item's mana usage changes through the day, peaking at 1.5x mana usage at noon, and 0.5x mana usage at midnight.
+		// Thanks to chikenbones for the help in the calculations
 		public override void ModifyManaCost(Player player, ref float reduce, ref float mult) {
-			float currentTime = (float)Main.time;
+			double currentTime = Main.time;
 			// The time at which it changes from day to night and vice versa.
-			int maxTime = Main.dayTime ? 54000 : 32400;
-			// The time at which it is 12pm or 12am, which isn't exactly half the values above.
-			int time12 = Main.dayTime ? 28800 : 18000;
-			// If the time it after 12, recalculate the current time to account for the length of the second half of the day instead of the first.
-			if (currentTime > time12) {
-				currentTime -= time12;
-				time12 = maxTime - time12;
-				currentTime += time12;
-			}
-			// If it is night time, add twice the duration of the current half of the night.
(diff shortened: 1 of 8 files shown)


ExampleMod/Items/Weapons/ExampleMagicMissile.cs
```diff
@@ -31,16 +31,16 @@ public class ExampleMagicMissile : ModItem
...
 			float currentTime = (float)Main.time;
 			int maxTime = Main.dayTime ? 54000 : 32400;
 			int time12 = Main.dayTime ? 28800 : 18000;
 			if (currentTime > time12) {
 				time12 = maxTime - time12;
 				currentTime = maxTime - currentTime;
 			}
-			float multi = 1 + currentTime / time12 * 0.5f * Main.dayTime.ToDirectionInt();
-			mana = (int)(mana * multi);
+			float timeMult = 1 + currentTime / time12 * 0.5f * (Main.dayTime ? 1 : -1);
```
Chicken-Bones: ever heard of a sine wave?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -6288,15 +6288,17 @@
...
  					NetMessage.SendData(21, -1, -1, null, number, 1f, 0f, 0f, 0, 0, 0);
  				}
  			}
 +		}
 +
 +		public int GetManaCost(Item item) {
-+			int mana = (item.type != 127 || !this.spaceGun) ? (int)(item.mana * this.manaCost) : 0;
-+			CombinedHooks.GetManaCost(this, item, ref mana);
++			float reduce = manaCost;
++			float mult = item.type != 127 || !spaceGun ? 1 : 0;
```
Chicken-Bones: split this into `mult = 1` and then an `if` statement for clarity?

ExampleMod/Items/Weapons/ExampleMagicMissile.cs
DarkLight66: Did ithat, and also added a lot of comments, since the calculations may not be too obvious.

ExampleMod/Items/Weapons/ExampleMagicMissile.cs
```diff
@@ -31,16 +32,30 @@ public class ExampleMagicMissile : ModItem
...
+			// If it is night time, add twice the duration of the current half of the night.
+			if (!Main.dayTime) {
+				currentTime += time12 * 2;
+			}
+			// We divide the current time by the current half of the day or night's time, then by 2.
+			// The result would gradually go from 0 at 4:30am, 0.5 at 12pm, 1 at 7:30pm and 1.5f at 12 am.
+			float timeMult = currentTime / time12 / 2;
+			// Then we multiply the result above by PI, and calculate the sine of that, then multiply it by 0.5f and add 1.
+			// The result would go from 1 at 4:30am, 1.5 at 12pm, 1 at 7:30pm and 0.5f at 12 am, in a sine wave pattern.
+			timeMult = 1 + (float)Math.Sin(timeMult * Math.PI) * 0.5f;
```
Chicken-Bones:
```
// goal, 1.5x mana usage at noon, 0.5 at midnight
float time = Main.time;
if (!Main.dayTime) time += 54000; //Main.time resets at night, so add the length of the day (15 hours)
// 86400 ticks (ingame minutes, real seconds) per day. 3600 ticks per ingame hour
// day starts at 4:30am so add the time between midnight and 4:30
time += 16200; //time is now measured as time since midnight
// period of sin is 2pi, period of day is 86400 ticks
// cos peaks at 0, but peak mana usage should be at midday. Either shift 12 hours, or invert (we invert here)
// cos goes between -1 and 1, so multiply by 0.5 and add 0.5 to make it go between 0.5 and 1.5
double timeMult = 0.5 - 0.5 * Math.Cos(time / 86400 * 2 * Math.PI);
```

ExampleMod/Items/Weapons/ExampleMagicMissile.cs
```diff
@@ -31,16 +32,30 @@ public class ExampleMagicMissile : ModItem
...
+				currentTime += time12;
 			}
-			float multi = 1 + currentTime / time12 * 0.5f * Main.dayTime.ToDirectionInt();
-			mana = (int)(mana * multi);
+			// If it is night time, add twice the duration of the current half of the night.
+			if (!Main.dayTime) {
+				currentTime += time12 * 2;
+			}
+			// We divide the current time by the current half of the day or night's time, then by 2.
+			// The result would gradually go from 0 at 4:30am, 0.5 at 12pm, 1 at 7:30pm and 1.5f at 12 am.
```
Chicken-Bones: you are correct here, but it's probably better to just have it be 1 at 6am and 6pm, the scaling to length of day/night is pretty confusing

ExampleMod/Items/Weapons/ExampleMagicMissile.cs
DarkLight66: Neither of these really do what i wanted to to show: 
* Normal mana usage at dawn / dusk, which are at 4:30 am and 7:30 pm in Terraria for some reason.
* And, coupled with the above, 1.5x mana usage at noon, 0.5 at midnight.

Because 12pm isn't the midpoint between dawn and dusk, nor is 12am the midpoint between dusk and dawn, each 4th of a day period differs in length.

* Your first example only matches the mana usage multiplier at noon and midnight, then it becomes normal at 6am/pm. This would match what would be expected but i wanted to show exactly when it changes between day and night and vice versa.
* Your second example only matches the mana usage multiplier at dusk and dawn, but differs at noon and midnight.

Chicken-Bones: Looks like I did miss that, but 4.5+7.5 = 12
Are you sure midnight isn't right inbetween 4:30 and 7:30 (and hence midday too)

DarkLight66: What i mean is that 12 / 2 == 6, but 4.5+6 != 12, so mana usage would peak at 10:30 in the day, and at 1:30 at night. It would be a different story if dawn and dusk were at 6.

Chicken-Bones: I think you'll find the second example works. Neither 12, 6 or 2 are used. Just the length of day/night becomes 0-pi in the sin

Chicken-Bones: "Because 12pm isn't the midpoint between dawn and dusk, nor is 12am the midpoint between dusk and dawn, each 4th of a day period differs in length."

I think you might be wrong here. 12 is halfway between 4.5 and 17.5, likewise there are 4.5 hours between 7:30 and midnight, and 4.5 hours between midnight and dawn

DarkLight66: Damn, you are completely right, i got so confused with the day's length, i somehow thought that the time between dawn and noon was different to the time from non to dusk, same for night time.

Jofairden:
Thanks for the contribution!

Issue #558, opened 2019-06-19
Straightforward ExamplePotion needed.


JavidPack:
We need a straightforward example of a potion that gives a buff. Potion.cs also needs to be renamed since it usually is confusing for modders looking to make a potion. 

On a similar note, health regen debuffs/buffs in ExampleMod need documentation.

PR #598, opened 2019-07-09
The addition of ExampleBuffPotion at the request of issue #558.

NuovaPrime:
### Description of the Change

Adding in a extremely basic and straightforward example of a potion item that grants a small defense buff to the player. I've also renamed Potion.cs to GlobalPotion.cs to make it less confusing.

It should help beginner users understand how buff application through items works in tModloader.

### Benefits

More examples for users to learn through and benefit from.


JavidPack:
Great, thanks, fixes #558

PR #607, opened 2019-07-11
Added `Search` properties to BuffID, ProjectileID, TileID, WallID and added `GetUniqueId()` to those and ItemID, NPCID

hamstar0:
### Description of the Change
* Added `Search` properties to BuffID, ProjectileID, TileID, WallID
* Added 2 `GetUniqueId(...)` methods (each) to BuffID, ProjectileID, TileID, WallID, ItemID, NPCID

### Alternate designs
Not applicable.

Styled after ModConfig's current method of saving items in configs ("Terraria IronPickaxe", "Stamina EnergyPotionItem", etc.), this branch adds equivalents for NPCs, projectiles, buffs, walls, and tiles (and maybe more later).

### Benefits
This will help adding future ModConfig support for the above listed data types.

### Possible drawbacks
Unsubstantial added memory use from the added `Search` properties (`IdDictionary` data structures).

### Applicable Issues
None foreseen.

### Sample Usage
```cs
Main.NewText( "Projectile 123: "+ProjectileID.Search.GetName(123) );
Main.NewText( "Projectile 123 UID: "+ProjectileID.GetUniqueId(123) );
```

### DIFF
--- a/patches/tModLoader/Terraria.ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/BuffID.cs.patch
@@ -7,22 +7,27 @@
  
  namespace Terraria.ID
  {
-@@ -210,5 +_,19 @@
+@@ -210,5 +_,24 @@
  		public const int Oiled = 204;
  		public const int BallistaPanic = 205;
  		public const int Count = 206;
 +
 +		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, short>();
 +
-+		public static string GetUniqueId(int buffType) {
++		public static string GetReadableUniqueKey(int buffType) {
 +			if ( BuffID.Search.ContainsId(buffType) ) {
 +				return "Terraria " + BuffID.Search.GetName(buffType);
 +			}
 +
-+			return "" + buffType;
++			ModBuff modBuff = BuffLoader.GetBuff(buffType);
++			if( modBuff != null ) {
++				return BuffID.GetReadableUniqueKey(modBuff);
++			}
++
++			throw new ArgumentOutOfRangeException("Invalid buff type value.");
 +		}
 +
-+		public static string GetUniqueId(ModBuff buff) {
++		public static string GetReadableUniqueKey(ModBuff buff) {
 +			return buff.mod.Name + " " + buff.Name;
 +		}
  	}
--- a/patches/tModLoader/Terraria.ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/BuffID.cs.patch
@@ -1,0 +1,30 @@
+--- src/Terraria\Terraria.ID\BuffID.cs
++++ src/tModLoader\Terraria.ID\BuffID.cs
+@@ -1,4 +_,6 @@
++using ReLogic.Reflection;
+ using System;
++using Terraria.ModLoader;
+ 
+ namespace Terraria.ID
+ {
+@@ -210,5 +_,19 @@
+ 		public const int Oiled = 204;
+ 		public const int BallistaPanic = 205;
+ 		public const int Count = 206;
++
++		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, short>();
++
++		public static string GetUniqueId(int buffType) {
++			if ( BuffID.Search.ContainsId(buffType) ) {
++				return "Terraria " + BuffID.Search.GetName(buffType);
++			}
++
++			return "" + buffType;
++		}
++
++		public static string GetUniqueId(ModBuff buff) {
++			return buff.mod.Name + " " + buff.Name;
++		}
+ 	}
+ }
+
--- a/patches/tModLoader/Terraria.ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/BuffID.cs.patch
@@ -7,28 +7,33 @@
  
  namespace Terraria.ID
  {
-@@ -210,5 +_,24 @@
+@@ -210,5 +_,29 @@
  		public const int Oiled = 204;
  		public const int BallistaPanic = 205;
  		public const int Count = 206;
 +
-+		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, short>();
++		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, int>();
 +
-+		public static string GetReadableUniqueKey(int buffType) {
-+			if ( BuffID.Search.ContainsId(buffType) ) {
-+				return "Terraria " + BuffID.Search.GetName(buffType);
++		public static string GetUniqueKey(int type) {
++			if (type < 1 || type >= BuffLoader.BuffCount)
++				throw new ArgumentOutOfRangeException("Invalid type: " + type);
++			if (type < Count) {
++				return "Terraria " + BuffID.Search.GetName(type);
 +			}
-+
-+			ModBuff modBuff = BuffLoader.GetBuff(buffType);
-+			if( modBuff != null ) {
-+				return BuffID.GetReadableUniqueKey(modBuff);
-+			}
-+
-+			throw new ArgumentOutOfRangeException("Invalid buff type value.");
++			var modBuff = BuffLoader.GetBuff(type);
++			return $"{modBuff.mod.Name} {modBuff.Name}";
 +		}
 +
-+		public static string GetReadableUniqueKey(ModBuff buff) {
-+			return buff.mod.Name + " " + buff.Name;
++		public static int TypeFromUniqueKey(string key) {
++			string[] parts = key.Split(new char[] { ' ' }, 2);
++			if (parts.Length != 2)
++				return 0;
++			if (parts[0] == "Terraria") {
++				if (!BuffID.Search.ContainsName(parts[1]))
++					return 0;
++				return BuffID.Search.GetId(parts[1]);
++			}
++			return ModLoader.ModLoader.GetMod(parts[0])?.BuffType(parts[1]) ?? 0;
 +		}
  	}
  }
--- a/patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
@@ -5,12 +5,4 @@
  using System;
  
  namespace Terraria.ID
-@@ -258,5 +_,7 @@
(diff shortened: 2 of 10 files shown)


Jofairden:
@JavidPack isn't this close to what you wanted with the cache stuff?
@hamstar0 for reference, we wanted to create a cache system where you can get vanilla stats for items/npcs/projectiles etc.

hamstar0:
> @JavidPack isn't this close to what you wanted with the cache stuff?
> @hamstar0 for reference, we wanted to create a cache system where you can get vanilla stats for items/npcs/projectiles etc.

Oddly, that sounds like a mod I was working on called [Game Changer](https://github.com/hamstar0/tml-gamechanger-mod).

No, this is just simple unique IDing of entities by type.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
```diff
@@ -0,0 +1,41 @@
...
++using ReLogic.Reflection;
+ using System;
+ 
+ namespace Terraria.ID
+@@ -1199,5 +_,32 @@
+ 		public const short BookStaffShot = 712;
+ 		public const short DD2Win = 713;
+ 		public const short Count = 714;
++
++		public static readonly IdDictionary Search = IdDictionary.Create<ProjectileID, short>();
```
Jofairden: In terms of naming I much prefer:
`ProjectileID.Find.Name(123)` or `ProjectileID.FindName(123)` 
or the search variants: `ProjectileID.Search.Name(123)` / `ProjectileID.SearchName(123)`
An alternative design is hosting these methods in a specific cache/query object:
`GameCache.GetProjectileName(123)` or `GameCache.Projectile.GetName(123)`
Which would help against polluting the vanilla class.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
```diff
@@ -0,0 +1,41 @@
...
++
++		public static string GetUniqueId(int projType) {
```
Jofairden: I think `GetUniqueId` is ambigious, people tend to associate numbers with Ids. I prefer it being renamed to `GetStringRepresentation`

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
```diff
@@ -0,0 +1,41 @@
...
++			else {
++				var proj = new Projectile();
++				proj.SetDefaults(projType);
++
++				if ( proj.modProjectile != null ) {
++					return proj.modProjectile.mod.Name + " " + proj.modProjectile.Name;
++				}
++			}
++
++			return "" + projType;
```
Jofairden: Please use `thething.ToString()`

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
```diff
@@ -0,0 +1,41 @@
...
+ 		public const short DD2Win = 713;
+ 		public const short Count = 714;
++
++		public static readonly IdDictionary Search = IdDictionary.Create<ProjectileID, short>();
++
++		public static string GetUniqueId(int projType) {
++			if ( ProjectileID.Search.ContainsId(projType) ) {
++				return "Terraria " + ProjectileID.Search.GetName(projType);
++			}
++			else {
```
Jofairden: else chain can be omitted as if-case contains an early return; this reduces indents and is preferred

patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
```diff
@@ -0,0 +1,16 @@
...
++using ReLogic.Reflection;
+ using System;
+ 
+ namespace Terraria.ID
+@@ -258,5 +_,7 @@
+ 		public const short ArkhalisPants_Female = 250;
+ 		public const short ArkhalisWings = 251;
+ 		public const short Count = 252;
++
++		//public static readonly IdDictionary Search = IdDictionary.Create<GlowMaskID, short>();	Can't use with `static` classes!
```
Jofairden: Is there a reason why we wouldn't want these functionalities in the commented classes?

JavidPack:
I'm also wondering if the ID classes are the right home for these methods. And no, the cache idea is separate. I'll give this more thought to decide if the methods belong here, but the Search fields definitely do.

Chicken-Bones:
`GetSaveKey` is definitely something we should have

patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
hamstar0: They're `static` classes, and create an error. Changing them to non-`static` classes would throw exceptions with mods that use them. This is a change for a hypothetical future tModLoader major versioned release.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
hamstar0: In theory, the fallthrough should never occur. Maybe I'll just throw an `Exception` instead.

hamstar0: Or maybe I'll return `null` instead? I hate doing that with functions that expect a string, in particular.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
hamstar0: They are "ID" relevant for the singular purpose of identifying types for human readable serialization, since the English names of items and NPCs and stuff are occasionally untrustworthy (even vanilla has duplicates; Seaweed for example).

hamstar0:
`GetSaveKey` seems specific for saving, and `GetStringRepresentation` doesn't indicate human readability. `GetReadableRepresentativeKey` would be the most "correct" form, but that's a mouthful.

@JavidPack If `Search` already exists in `ItemID` and `NPCID`, does that risk breaking compat?

hamstar0:
Newer [commit](https://github.com/tModLoader/tModLoader/pull/607/commits/6b435b9c57e3792b24644129daf61eb52ebc4841) added.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
Jofairden: It doesn't matter,it's not about style. .ToString() is conveying and clear, moreso than adding it to an empty string. In c# you replace "" with string.Empty (in Java you don't). If it's a flow that shouldn't happen, then yes, an exception may be preferred.

patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
Jofairden: Ah, copy that. In that case I recommend none of these features are part of the ID classes, since that'd be incompatible for these. In that case they should be present in a new custom class from tModLoader.

Jofairden:
> and GetStringRepresentation doesn't indicate human readability

A string representation doesn't have to indicate human readability anyways, it's a representation of the subject as a string. That can be in any way shape or form. A good example is representing some data in say, base64 format (we do this in chat tags). The base64 contains all kinds of data, for example an item's full data including mod data. It isn't exactly human readable, since we can't read any of it through the base64 unless we decode it. It still represents it in the form of a string though. Also "Terraria #ID" is very human readable.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
Jofairden: Yeah since we wanted to implement such cache (soon™️) a class like it would exist anyways, hence my suggestion to maybe make such class already 👍

hamstar0:
> > and GetStringRepresentation doesn't indicate human readability
> 
> A string representation doesn't have to indicate human readability anyways, it's a representation of the subject as a string. That can be in any way shape or form. A good example is representing some data in say, base64 format (we do this in chat tags). The base64 contains all kinds of data, for example an item's full data including mod data. It isn't exactly human readable, since we can't read any of it through the base64 unless we decode it. It still represents it in the form of a string though. Also "Terraria #ID" is very human readable.

Not in a JSON config file so much. Granted ModConfig uses a UI to make the matter of readability a non-issue, but prior to that I've have tons of trouble with people editing configs incorrectly even with human-readable entries for entities (including the previously-mentioned non-uniqueness issue). The current approach (ModConfig already uses this for items, I should re-mention) solves both uniqueness and human readability.

patches/tModLoader/Terraria.ID/GlowMaskID.cs.patch
hamstar0: Is there a way to get an `IdDictionary` (or equivalent) with a `static` class? Otherwise, it won't matter to make a custom class as a layer on top of these. Keep in mind their only use is to map the `public const short` field names to their `type`s in a dictionary data structure; maybe other means are available?

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
hamstar0: I'm currently using `GetReadableUniqueKey(...)`, based on Chicken Bones suggestion. Feel free to suggest an alternative.

patches/tModLoader/Terraria.ID/ProjectileID.cs.patch
hamstar0: Keep in mind classes to cache each field of each entity type will use up a fair chunk of memory.

JavidPack:
Thanks a bunch for pushing this feature along.

patches/tModLoader/Terraria.ID/BuffID.cs.patch
```diff
@@ -7,28 +7,33 @@
...
++				throw new ArgumentOutOfRangeException("Invalid type: " + type);
++			if (type < Count) {
++				return "Terraria " + BuffID.Search.GetName(type);
 +			}
-+
-+			ModBuff modBuff = BuffLoader.GetBuff(buffType);
-+			if( modBuff != null ) {
-+				return BuffID.GetReadableUniqueKey(modBuff);
-+			}
-+
```
hamstar0: Why remove the ModBuff check? An out-of-range `buffType` should just mean to check if it's a ModBuff instead, right?

PR #615, opened 2019-07-17
Updated french translations

orian34:
adjusted a few old lines and added the new ones.


Jofairden:
JSON verified and correct. Thanks!

PR #620, opened 2019-07-22
Add exclusive accessory example to ExampleMod

direwolf420:
## Description of the Change
EDIT: Bunch of things have changed, read the comment chain below
Adds two accessories to ExampleMod demonstrating how to implement a simple "only one of this kind can be equipped at a time" accessory (similar to wings in vanilla). In addition to that, it also shows a very basic usage of abstract/sealed/virtual to make this easier organization wise. The structure is similar to what [ExampleDamageItem](https://github.com/tModLoader/tModLoader/blob/master/ExampleMod/Items/ExampleDamageClass/ExampleDamageItem.cs) does.

### Alternate designs
This example of "exclusive" accessories is realized through current tModLoader hooks and is kept as simple as possible. I couldn't think of an even easier approach.

There aren't alot of accessories in ExampleMod, plus a bunch of people in the #help-with-code channel on the official discord asked for this kind of behavior for accessories, and a basic example to point them towards makes it easier on everyone.

### Benefits
An additional example of inheritance, of which there aren't many in ExampleMod.

### Possible drawbacks
None since this is just an example.

### Applicable Issues
One thing that doesn't work exactly like vanilla is swapping accessories of the same type in the same slot. With vanilla wings, you can hold one on your mouse, click on already equipped wings, and it will swap them. With this example, CanEquipAccessory forbids that, requiring the player to remove the equipped accessory first before equipping another one.

### Sample Usage
Implementing exclusive accessories similar to wings, for example "baubles" or "emblems".


DarkLight66:
There's already an example of such an accessory, and it doesn't have the issues your example does: https://github.com/tModLoader/tModLoader/blob/master/ExampleMod/Items/Abomination/SixColorShield.cs
The use of inheritance could however be beneficial when checking if the accessory you are trying to equip can replace the accessory on the slot.

direwolf420:
Thanks for pointing that out, I experienced that exact issue that the code in CanEquipAccessory from SixColorShield is solving. Any chance the PR can be adjusted/expanded to still be justified?

DarkLight66:
Maybe doing something like this for the `CanEquipAccessory` part ```cs
public override bool CanEquipAccessory(Player player, int slot) {
	if (slot < 10) // This allows the accessory to equip in Vanity slots with no reservations.
	{
		int maxAccessoryIndex = 5 + player.extraAccessorySlots;
		for (int i = 3; i < 3 + maxAccessoryIndex; i++) {
			// We need "slot != i" because we don't care what is currently in the slot we will be replacing.
			if (slot != i && player.armor[i].modItem is ExclusiveAccessory) {
				return false;
			}
		}
	}
	return true;
}``` and removing the `exclusiveAccessory` bool from ModPlayer and everything related to it (including the sealed override of `UpdateAccessory`) since it isn't needed anymore.

direwolf420:
Exactly my train of thought, you were quicker to write it up :stuck_out_tongue:. Will get to work then and add some comments explaining the whole thing

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+// PR NOTE, REMOVE THIS IF IT'S MERGED: some of the comments in here are taken from ExampleDamageItem and SixColorShield
```
Jofairden: Please remove that yourself

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -0,0 +1,94 @@
...
+			return true;
+		}
+	}
+
+	// Here we add our accessories, note that they inherit from ExclusiveAccessory, and not ModItem
+
+	public class GreenExclusiveAccessory : ExclusiveAccessory
+	{
+		public override void SetStaticDefaults() {
+			Tooltip.SetDefault("You can't equip this when 'Yellow Exclusive Accessory' is already equipped!"
```
Jofairden: It would be great if you used ModifyTooltips to show the  "can't equip" line dynamically if one of those exclusives is equipped.

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -0,0 +1,94 @@
...
+			// 50% melee and ranged damage increase
+			player.meleeDamage += 0.5f;
+			player.rangedDamage += 0.5f;
+		}
+	}
+
+	public class YellowExclusiveAccessory : ExclusiveAccessory
+	{
+		public override void SetStaticDefaults() {
+			Tooltip.SetDefault("You can't equip this when 'Green Exclusive Accessory' is already equipped!"
```
Jofairden: Idem dito

direwolf420:
Thanks, just to clarify, you mean right clicking one accessory while the other one is equipped automatically swaps them?

Jofairden:
> 
> 
> Thanks, just to clarify, you mean right clicking one accessory while the other one is equipped automatically swaps them?

Yep.

direwolf420:
I found this code related to the swap feature in ItemSlot.RightClick(Item[] inv, int context = 0, int slot = 0): 
```if (flag2 && ItemLoader.CanEquipAccessory(inv[slot], slot - 10))
{
	Utils.Swap<Item>(ref inv[slot], ref inv[slot - 10]);
	Main.PlaySound(7, -1, -1, 1, 1f, 0f);
	Recipe.FindRecipes();
	if (inv[slot].stack > 0)
	{
		if (context != 0)
		{
			switch (context)
			{
				case 8:
				case 9:
				case 10:
				case 11:
				case 12:
				case 16:
				case 17:
					AchievementsHelper.HandleOnEquip(player, inv[slot], context);
					break;
			}
		}
		else
		{
			AchievementsHelper.NotifyItemPickup(player, inv[slot]);
		}
	}
}```
Will I need to hook into it with IL?

**Comment by Jofairden** [comment]
Maybe use the RightClick hook?

**Comment by direwolf420** [comment]
* The right clicking should work as expected now, it spawns in a new item when you swap items though which might be annoying
* I went all out on this one, introduced delegates aswell, cause why not
* I couldn't think of any uses of the "virtual Safe..." methods but would like to showcase their use

**Comment by direwolf420** [comment]
Allow me to explain: I thought about doing a bit more than just the exclusive accessory feature, by showing how to interact with methods in an inheritance setting in two different ways. One of them is via `base.Method(args)`, which is covered in `YellowExclusiveAccessory` (explains what happens if you do and don't use it). The other one is the concept of sealed overrides that [ExampleDamageItem](https://github.com/tModLoader/tModLoader/blob/master/ExampleMod/Items/ExampleDamageClass/ExampleDamageItem.cs?ts=4#L18) uses, which creates a guaranteed behavior that is wanted (in that case, removing all the damage types on a weapon so it counts as a custom damage type). The modder would expect the items inherited from `ExclusiveAccessory` to behave properly, even when overriding. making the overrides sealed but not provide a `Safe` method will restrict the modders ability to further expand on the features of the accessory, and not making it sealed in the first place will force the modder to call `base.Method(args)` in every instance to preserve "expected" behavior. 
Here is an example of how it can go wrong. Assuming `CanEquipAccessory` isn't sealed and there is no `Safe` method: If it's overriden and `base.CanEquipAccessory` isn't called inside of it, this accessory will lose it's property of being "exclusive", causing it to be equipped even if it should't. But since `ModifyTooltips` isn't overriden yet, its tooltip will still say that it can't be equipped.
Basically, introducing `Safe` methods makes sure that the modder doesn't have to worry about messing up the default behavior of the `ExclusiveAccessory`.
Basically expanding on the concept ExampleDamageItem introduced.

**Comment by Chicken-Bones** [comment]
Sealing an inheritance chain and providing a new sub-member for further overrides is a valid design practice. I don't think ExampleMod is really the place for presenting competing high-level design examples.  The majority of modders aren't proficient developers and the pros and cons of different approaches here may just add confusion.

In many cases I'm happy to have separate classes showcase different approaches, some more sophisticated than others, but since the arguments for Safe vs base only have real implications when you get to API design, I think it's better to keep it simple here.

**Comment by direwolf420** [comment]
So should I just get rid of all the `Safe` methods and put a disclaimer that if you forget to use base. it will override the default behavior?

**Comment by Chicken-Bones** [comment]
Absolutely, good 'how to code' advice. If you want the functionality that's in the base method, then call it.

**Comment by Jofairden** [comment]
> Basically, introducing Safe methods makes sure that the modder doesn't have to worry about messing up the default behavior of the ExclusiveAccessory.

I get what you mean, but it makes no sense not calling base if you inherit this class (because then why do you inherit it?)

But yeah also what cb said. It's not like it's invalid or something, but keep it simple.

**Comment by Chicken-Bones** [comment]
Funnily enough, I remember the same argument being made several years ago for some of our Mod/GlobalItem hooks. The reason that `AutoSetStaticDefaults` exists is because we didn't want modders to have to call `base.SetStaticDefaults()`. Though at that time, it would've been a poorly documented API change because previously `SetStaticDefaults` was empty and we wanted to add some 'default implementation' to it.

I think almost all tML virtual methods have empty default implementations, because there's a lot of modders new to programming, but I don't think it's necessarily a good thing to /learn/.

**Comment by direwolf420** [comment]
I found an issue with the swapping code, looks like vanilla takes care of it sometimes (even when CanRightClick returns false). Investigating, will take me a few days

**Comment by direwolf420** [comment]
So I THINK I fully restored vanilla behavior, so these items now work like wings in terms of swapping.

File: ExampleMod/Items/Accessories/ExclusiveAccessory.cs [hunk 4]
```diff
@@ -0,0 +1,185 @@
...
+				// IsAir makes sure we don't check for "empty" slots
+				// IsTheSameAs() compares two items and returns true if their types match
+				// "is ExclusiveAccessory" is a way of performing pattern matching
+				// Here, inheritance helps us determine if the given item is indeed one of our ExclusiveAccessory ones
+				if (!accessory.IsAir &&
+					!item.IsTheSameAs(accessory) &&
+					accessory.modItem is ExclusiveAccessory) {
+					// If we find an item that matches these criteria, execute the code inside it
+					// The second argument is just for convenience, technically we don't need it since we can get the item from just i
+					whenFound.Invoke(i, accessory);
```
Jofairden: Change this to return i, or alternatively change it to return a value tuple as (i, accessory) that can be deconstructed. It keeps the calling code simpler and removes the nesting level.

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -0,0 +1,185 @@
...
+					break;
+				}
+			}
```
Jofairden: Simply return -1 or (-1, null) here.

direwolf420:
ChickenBones recommended to redo the method aswell:
```
Chicken Bones: FindDifferentEquippedExclusiveAccessory doesn't actually need to take a delegate, 
and the code would be cleaner if it just returned the slot
Chicken Bones: you could return -1 for not found, you could return an int? or even a bool with out params for slot and accessory
```

I think the reason I decided to use delegates is to not do the usual "if item is not found" check every time, let the method handle it instead:

```cs
int slot = FindDifferentEquippedExclusiveAccessory();
if (slot != -1) {
//actual code here operating on player.armor[slot]
}
```

So which way should I go? Tuple? int?

direwolf420:
First draft, doing this for the first time so I have no idea how to make it look good. It works the same as before though

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -74,44 +74,47 @@ public abstract class ExclusiveAccessory : ModItem
 
 			bool canRightClick = false;
 			// Only allow right clicking if there is a different ExclusiveAccessory equipped
-			FindDifferentEquippedExclusiveAccessory((int i, Item accessory) => {
+			if (FindDifferentEquippedExclusiveAccessory().accessory != null) {
 				canRightClick = true;
```
Jofairden: please just change this to `return true;` for simpler control flow

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -74,44 +74,47 @@ public abstract class ExclusiveAccessory : ModItem
...
-			});
+			}
 			// If this hook returns true, the item is consumed (just like crates and boss bags)
 			return canRightClick;
```
Jofairden: the most correct from here would be to `return base.CanRightClick()`

ExampleMod/Items/Accessories/ExclusiveAccessory.cs
```diff
@@ -74,44 +74,47 @@ public abstract class ExclusiveAccessory : ModItem
...
-					whenFound.Invoke(i, accessory);
-					break;
+					return (i, otherAccessory);
 				}
 			}
+			// If no item is found, we return default values for index and item, always check one of them with this default when you call this method!
+			return (-1, null);
 		}
-	}
+    }
```
Jofairden: I see spaces

direwolf420:
Done

PR #629, opened 2019-07-26
Installer: Unix installer written in bash

Kreyren:
Draft issue: Not finished.

This issue improves provided scripts in tModLoader, fixes spellcheck and adds alternative to `tModLoader.Installer.jar` so that end-user doesn't have to use java (which seems insane to prioritize java over bash on linux).

TODO: Force GUI since steam doesn't output text output.
TODO: Test on darwin
TODO: Provide installer

### DIFF
--- a/installer/Terraria
+++ b/installer/Terraria
@@ -1,0 +1,38 @@
+elevant https://mywiki.wooledge.org/BashFAQ/028
+
+# Error handling
+if ! command -v einfo > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
+if ! command -v warn > /dev/null; then	warn()	{	printf "WARN: %s\n" "$1"	1>&2	;	} fi
+if ! command -v die > /dev/null; then	die()	{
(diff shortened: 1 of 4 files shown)


Jofairden:
Hey, thanks for starting on this and keep us updated! May I recommend renaming the project to "installer-bash". We will/should rename installer2 to "installer-java".  (Feel free to do that in this PR so that those are updated when it merges)

Kreyren:
> Hey, thanks for starting on this and keep us updated! May I recommend renaming the project to "installer-bash". We will/should rename installer2 to "installer-java". (Feel free to do that in this PR so that those are updated when it merges)

Will do :)

Kreyren:
Relevant: Winetricks is using `zenity` (https://en.wikipedia.org/wiki/Zenity) for the GUI. We might be able to use the same thing assuming that steam doesn't show any dialogs nor output by default.

Jofairden:
I think I will close this due to inactivity, plus installation on Steam (most users) will be automatic.

PR #634, opened 2019-07-27
Modify MonoModHooks Method Signature

TechnoJo4:
### Description of the Change
Currently, when creating a MonoMod Hook, tModLoader logs the target method's signature with the parameter types' real names, but most people don't use these (I've never seen `Single` or `Boolean` used in real code before). This PR also fixes types not showing up when they were ref/out.

Old:
`On.Terraria.Player::PickAmmo(Item, ref , ref , ref , ref , ref , Boolean)`

New:
`On.Terraria.Player::PickAmmo(Item, ref int, ref float, ref bool, ref int, ref float, bool)`

### Alternate designs
Use something other than a Dictionary for aliases? I don't know if it's really needed since people who use MonoMod probably know the types' real names anyways.

N/A

### Benefits
Fixes a problem.
Clearer and easier to understand.

### Possible drawbacks
I don't think there are any, really.

### Applicable Issues
N/A

### Sample Usage
N/A

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/MonoModHooks.cs
+++ b/patches/tModLoader/Terraria.ModLoader/MonoModHooks.cs
@@ -15,6 +15,24 @@ namespace Terraria.ModLoader
 {
 	public static class MonoModHooks
 	{
+		private static Dictionary<Type, string> defaultAliases = new Dictionary<Type, string> {
+			{ typeof(object), "object" },
+			{ typeof(bool), "bool" },
+			{ typeof(float), "float" },
+			{ typeof(double), "double" },
+			{ typeof(decimal), "decimal" },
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria.ModLoader/MonoModHooks.cs
```diff
@@ -15,6 +15,24 @@ namespace Terraria.ModLoader
 {
 	public static class MonoModHooks
 	{
+		private static Dictionary<Type, string> defaultAliases = new Dictionary<Type, string> {
```
Jofairden: Huh, interesting how this is much simpler than what I can find on Google to get names of primitives.. Nice job

PR #644, opened 2019-08-07
Added Runnable mods in main menu and TModLoader compile access (dev "ONLY" mod hack)

Lanboost:
### Description of the Change
A small modification to accomplish 3 things
- Added a public access to the field "selectedMenu" in the Main.cs
- Added a method to run TModLoader compile all directly from Main.cs
- Added a ugly event hook to enable global updates (so a mod get updates even in main menu)

These three things allows mods to do stuff in main menu. Main use will be to autoLogin, easyRecompile, by simulating "mouse presses" till will ensure consistency with normal compile and setup flow

### Alternate designs
Make a real Update hook for mods instead of the event based one, make the new methods non static

It could probably be done by a lt mod, and it breaks the "sandbox" of the mods a bit, however it will make mod development much easier imo

### Benefits
Modable control in main menu

### Possible drawbacks
Ugly hack (a pretty small one tho)

### Applicable Issues
Feel free to come with feedback with the static properties and how a nicer update hook could be made without making to big of a change

This is intended as a dev only hack and not really for a general mod stuff

### Sample Usage
Added ExampleEasyRecompile to the ExampleMod (automatic login+recompile with Keys.P)
A standalone mod version can be found @ https://github.com/Lanboost/TModloader_ExampleEasyRecompile

### DIFF
--- a/ExampleMod/ExampleEasyRecompile.cs
+++ b/ExampleMod/ExampleEasyRecompile.cs
@@ -1,0 +1,133 @@
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Input;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using Terraria;
+using Terraria.GameContent.UI.States;
+using Terraria.ModLoader;
+using Terraria.ModLoader.Config;
+using Terraria.Social;
+using Terraria.UI;
+
+
+
+
+
+namespace ExampleMod
+{
+    public class ExampleEasyRecompile
+	{
+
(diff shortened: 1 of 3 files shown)


Chicken-Bones:
What's your use case here? Rebuilding a mod each time you open the game? Why not use visual studio?

Lanboost:
Yea, I managed to do it as a PURE mod with reflection and MonoMod

Closing

And it is so I can easily test without having to restart the client or start the server etc, the "RecompileMod" will take care of all that for me so I can test multiplayer easily

Issue #582, opened 2019-07-02
More player.allDamage problems


DarkLight66:
### Description
* Some special damage dealing effects that are affected by certain vanilla damage bonuses that don't rely on `player.GetWeaponDamage` to calculate damage dealt, like some vanilla dashes and mounts like minecarts, aren't affected by the new `allDamage` field or any of the new `damageMult` fields.
* Items that had no damage type listed but still dealt damage, like the flare gun and   holy water, are affected by `allDamage` and `allDamageMult`, which differs from vanilla.
* Items that had more than somehow had more than one damage type set are now affected by all damage bonuses related to those types, rather than only being affected by the one with the highest priority, with the priority chain shown in tooltips and used by projectiles being being melee => ranged => thrown => magic => summon. As far as i know there's no example of an item with multiple damage types set in vanilla though.

### Steps to Reproduce
Equip items that increase all damage dealt

### Expected behavior 
Damage from things like dashes and mounts is increased, while damage of items like the flare gun and holy water remains the same

### Actual behavior
Damage from things like dashes and mounts isn't affected, while damage of items like the flare gun and holy water is increased

### Reproduction frequency
Always

### Versions
`tModLoader v0.11`

### Mods
None

### Additional Information
List of the special vanilla damage dealing effects that are affected by some vanilla damage bonuses:
* Shield of Cthulhu and Solar Flare armor dashes (melee damage)
* Forbidden armor ancient storm (magic and minion damage)
* Slime mount while jumping on top of enemies (minion damage)
* Unicorn and Basilisk mounts while charging (minion damage)
* ~~Minecarts while moving fast (highest damage multiplier out of melee, magic and ranged)~~ This was actually referring to crit change rather than damage, oops.
* Stardust Guardian's explosion attack (minion damage)

List of items with no listed damage type that deal damage in vanilla that i have found:
* Flare Gun (ID 930)
* Holy Water (ID 422)
* Unholy Water (ID 423)
* Blood Water (ID 3477)

A note on the damage type priority chain, for weapons that do damage on melee contact, the chain would be thrown => summon => magic => ranged => melee for some reason, though since all melee weapons in vanilla do melee damage, it's safe to assume that it'd be better to use the priority chain used for everything else.

Chicken-Bones:
Good list, we definitely want to fix the accidental nerfs, but I'm less concerned about buffs

DarkLight66:
[...]

PR #654, opened 2019-08-17
Added Player.GetOtherDamage and ModPlayer.ModifyOtherDamage

DarkLight66:
### Description of the Change
* Added `Player.GetOtherDamage`, which is now used to calculate damage from most special damage dealing effects that can't rely on `Player.GetWeaponDamage`, due to the damage not being directly tied to specific instance of an item. This includes:
  * Shield of Cthulhu and Solar Flare Armor dashes
  * Forbidden Armor ancient storm
  * Unicorn, Basilisk and Slime mount's attacks
  * Minecart collision damage
  * Stardust Guardian's explosion attack
  * Inferno Potion's ring of fire
* Added `ModPlayer.ModifyOtherDamage`, which is called from `Player.GetOtherDamage` to let modders modify damage from special damage sources
* Prevented vanilla items that deal damage with no type from being affected by `Player.allDamage` and `Player.allDamageMult`, to properly mimic vanilla behavior
* Changed how ranged ammo damage is calculated so it calls `Player.GetWeaponDamage` on the ammo
* Implementing the change above fixed a vanilla bug where arrow ammo tooltip wouldn't reflect damage bonus from archery potion
* Fixed a bug where global flat damage bonuses would be applied to items with no damage value, which only affected projectiles shot by these items
* Fixed a bug where Last Prism projectile damage wasn't affected by `Player.allDamage` and `Player.allDamageMult`, as the damage wasn't being calculated using `Player.GetWeaponDamage`

This PR ultimately fixes a bug introduced in 0.11, where damage from special damage sources that are affected by specific class damage bonuses wasn't affected by global damage bonuses

### Applicable Issues
#582


Chicken-Bones:
Rolling Melee/Magic/Ranged/Minion in here seems incongruous, because normal damage doesn't run through these code paths. I'd also prefer if you left the flat/add/mult parameters which apply to each damage subtype as parameters, so they don't mysteriously disappear from the original damage calculation only to appear in an separate new method

PR #659, opened 2019-08-23
Check for Grass instead of Dirt tiles in MakeWells

DarkLight66:
Right now MakeWells checks for dirt tiles, but worlds hardly ever generate with dirt on the surface touching air. Checking for grass should be more effective.


JavidPack:
No, there is no grass at that point in world generation. Spreading grass comes much later

DarkLight66:
Hmm this explains it. There was someone who was having problems with getting the wells to generate, it turns out they had changed the worldgen pass to "Micro Biomes".

PR #362, opened 2018-06-12
Added new DrawMenu hooks

BlueTRaven:
Added 3 new hooks: PreMenuPress, ModifyMenus, ModifyMenuTextsColors.

### Description of the Change
I have added 3 new hooks with the intention of adding functionality to be able to change the Main Menu's text buttons and their functionality.
These hooks are: 

PreMenuPress(int selectedMenu)
A hook created to detect what menu text button has been pressed before the button does anything.

selectedMenu parameter: The menu that has been clicked. 

This hook is not called if no button has been clicked (Main.selectedMenu == -1)

ModifyMenus(int focusMenu, int selectedMenu, string[] buttonNames, float[] buttonScales, ref int offY, ref int spacing, ref int numButtons, ref bool backButtonDown)
This is the main hook that allows changing/adding Main Menu text buttons.

focusMenu parameter: The currently focused (hovered) menu.
selectedMenu parameter: The menu that has been clicked. 
buttonNames parameter: An array of the button names for the current menuMode ID.
buttonScales parameter: An array of the button scales for the current menuMode ID.
buttonVerticalSpacing parameter: An array of the spacings for the current menuMode ID.
numButtons parameter: The total number of buttons on the screen.

ModifyMenuTextsColors (int currentTextIndex, bool isOutlineColor, ref Color oldColor)
Allows you to modify the colors of a menu text button. 
This is called in a for loop 5 times for each existing menu text button; the first 4 being the outline draw pass (isOutlineColor is true,) and the last being the inner draw pass.

currentTextIndex: The current index of the menu text button.
isOutlineColor: Whether or not this is running in an outline color draw pass.
oldColor: The current color before modification.

### Alternate designs
N/A

### Why this should be merged into tModLoader
Modifying the the main menu is (afaik) impossible in current versions in mods. If this change were added it would allow much greater freedom to modders, as the main menu is no longer off-limits.

### Benefits
The ability to change the main menu to the modder's discretion, something that is not currently possible (afaik) without modifying vanilla in some way.

### Possible drawbacks
None that I can think of.

### Sample Usage
Code added to a mod's main Mod class:
`public override void PreMenuPress(int selectedMenu)
{
    if (Main.menuMode == -7)
    {
		Main.expertMode = false;    //may not be necessary; I think vanilla resets this automatically. But better safe than sorry.
    }
}

[...]

PR #662, opened 2019-08-24
WIP feat: Make max (de)buff limit modifiable

AnimeCatgirl:
# Description of the Change
`Player.maxBuffs` is a property defining the (de)buff limit within Terraria, it is a constant property, meaning that it is unchangeable and set at compile time.

This PR aims to make the buff limit changeable from a single property rather than hooking **28** (see below) different methods that affect the buff limit.
The way this is accomplished is by
- Changing `Player.maxBuffs` from `const` to `static`
- Modifying all points where `22` (vanilla buff limit) became hard-coded to reference `Player.maxBuffs`
### Alternate designs
- There could be an optional `maxBuffs` parameter within the Player constructor which defines the limit, defaulting to 22
- Another option could be adding a `maxBuffs` property to `ModLoader.ModPlayer` enabling the same functionality while leaving `Player.maxBuffs` completely vanilla
This PR brings more mod functionality to tModLoader, as of the time of this PR, in order to modify the buff limit, you'd have to modify multiple methods throughout multiple classes, both public and non public.
### Benefits
Merging these patches would enable the average mod creator to implement more content such as buffs, debuffs, potions, and summons, without the need to hook into multiple methods.
### Possible drawbacks
As these changes currently stand,
~~- There is an issue within the `Player.SavePlayer` method when the buff limit is changed. The way player files are created and read from, if a character is created with a buff limit set to 22, afterwards reloading with a new buff limit of any other number, the file will appear corrupted in-game because of the way `BinaryReader` functions.~~

~~This could be remedied by including a marker of some kind that states the buff limit within the file, then reading that marker in order to on-the-fly switch the buff limit to load that player file.~~ Fixed in https://github.com/tModLoader/tModLoader/pull/662/commits/1a74ba76483e9ea242dea01400c1d3dd46d06458
~~- I have not implemented any form of auto modification of the Y value when [drawing buff icons](https://github.com/AnimeCatgirl/tModLoader/blob/StaticBuffLimit/patches/tModLoader/Terraria/Main.cs.patch#L4010)~~

~~This is a simple fix only requiring some basic logic and math adding onto the Y value when it hits 22 or more.~~ Fixed in https://github.com/tModLoader/tModLoader/pull/662/commits/4bb5281aaaf4826ae3c758bd8c7859b5623c883d
~~- What happens when two mods change the buff limit?~~

~~We could implement some type of mod selector for choosing what mod change tModLoader should follow, similar to what @Agrair suggested in #362.~~ Fixed in https://github.com/tModLoader/tModLoader/pull/662/commits/3fc4a687e4f744f9770236dd3bdd74caf8966c39
### Sample Usage

Originally in order to change the buff limit, you would need to modify [these methods](https://paste.mod.gg/rolifoqura.cs).
With this PR a mod creator can simply add a new override for `ExtraBuffSlots` in their main mod file.
```c#
public override uint ExtraBuffSlots => 100;
```

### DIFF
--- a/patches/tModLoader/Terraria.ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/BuffID.cs.patch
@@ -7,10 +7,12 @@
  
  namespace Terraria.ID
  {
-@@ -210,5 +_,29 @@
+@@ -209,6 +_,30 @@
+ 		public const int BetsysCurse = 203;
  		public const int Oiled = 204;
  		public const int BallistaPanic = 205;
- 		public const int Count = 206;
+-		public const int Count = 206;
++		public static int Count = Player.maxBuffs + 206;
 +
 +		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, int>();
 +
--- a/patches/tModLoader/Terraria.ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/BuffID.cs.patch
@@ -7,12 +7,10 @@
  
  namespace Terraria.ID
  {
-@@ -209,6 +_,30 @@
- 		public const int BetsysCurse = 203;
+@@ -210,5 +_,29 @@
  		public const int Oiled = 204;
  		public const int BallistaPanic = 205;
--		public const int Count = 206;
-+		public static int Count = Player.maxBuffs + 206;
+ 		public const int Count = 206;
 +
 +		public static readonly IdDictionary Search = IdDictionary.Create<BuffID, int>();
 +
--- a/patches/tModLoader/Terraria.ModLoader.IO/PlayerIO.cs
+++ b/patches/tModLoader/Terraria.ModLoader.IO/PlayerIO.cs
@@ -149,7 +149,6 @@ namespace Terraria.ModLoader.IO
 
 		internal static List<TagCompound> SaveModBuffs(Player player) {
 			var list = new List<TagCompound>();
-			byte vanillaIndex = 0;
 			for (int k = 0; k < Player.maxBuffs; k++) {
 				int buff = player.buffType[k];
 				if (buff == 0 || Main.buffNoSave[buff])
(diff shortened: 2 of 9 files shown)


Chicken-Bones:
I suggest changing the buff limit to something we set, maybe 50 or 100. Also can you just disable vanilla buff saving? Or only save the first 22 buffs in vanilla?

AnimeCatgirl:
It would be possible to keep vanilla buff saving at a constant 22 buff limit, but as it currently stands removing vanilla buff saving seems out of the question because people would lose the ability to use old player files on newer tML versions.

Chicken-Bones:
Keep the loading, just remove the saving.

Chicken-Bones:
On rereading, it's actually a really good idea to save only vanilla buffs in the vanilla plr file, and save all other modded buffs in the tplr using a list. It shouldn't matter if buffs get reordered on load

AnimeCatgirl:
What if we save no buffs within the vanilla plr file, but still write the 22 required 0 bytes on quit.
We still load any buffs a player has from the plr file with a constant 22, keeping plr compatibility.
Instead we would save all buffs into the tplr file.

If a player loads a character with buffs from before the update, the buffs WILL get loaded properly with no issues,
When they quit playing, we then save ALL buffs, vanilla or modded, into the tplr file.

The old characters would get updated to the "new" format without any issues.

Chicken-Bones:
I'm also on board with that

AnimeCatgirl:
Now that the main issues have been addressed, do we plan on setting our own custom value? or leaving it at 22 and letting mod creators set the value.

Jofairden:
> Instead we would save all buffs into the tplr file.

But if you revert to non-modded you lose any (vanilla) buffs you otherwise had saved ;')

patches/tModLoader/Terraria.ID/BuffID.cs.patch
```diff
@@ -7,10 +7,12 @@
...
  namespace Terraria.ID
  {
-@@ -210,5 +_,29 @@
+@@ -209,6 +_,30 @@
+ 		public const int BetsysCurse = 203;
  		public const int Oiled = 204;
  		public const int BallistaPanic = 205;
- 		public const int Count = 206;
+-		public const int Count = 206;
++		public static int Count = Player.maxBuffs + 206;
```
Chicken-Bones: Not sure on this line

AnimeCatgirl: This may have been a leftover remnant from when I was running a bunch of tests, thanks for pointing this out!

Issue #664, opened 2019-08-25
[Hook Request] OnCatchNPC


direwolf420:
### Description
Request to add a new hook that runs after an NPC is successfully catched when `NPC.CatchNPC(int i, int who = -1)` is called. This hook would be in the GlobalNPC and ModNPC classes.

### Alternatives
The Modder could do IL/On the CatchNPC function instead

### Example usage
* Create visuals
* Track "catch progress/count"
* Spawn items

### Code
```cs
//public static void CatchNPC(int i, int who = -1)
//...
Item item = new Item();
item.SetDefaults((int)Main.npc[i].catchItem, false);
Item.NewItem((int)Main.player[who].Center.X, (int)Main.player[who].Center.Y, 0, 0, (int)Main.npc[i].catchItem, 1, false, 0, true, false);
//Hook called here
//NPCLoader.OnCatchNPC(Main.npc[i], who);
Main.npc[i].active = false;
NetMessage.SendData(23, -1, -1, null, i, 0f, 0f, 0f, 0, 0, 0);```

PR #671, opened 2019-09-01
New OnCatchNPC hook

direwolf420:
### Description of the Change
A new hook that runs after an NPC is successfully catched when `NPC.CatchNPC(int i, int who = -1)` is called. See: #664 

### Alternate designs
Using `On.` and replace the method, but a hook is more elegant.

There was no hook before where you could make things happen when an NPC is catched.

### Benefits
Modders can now make things happen when an NPC is catched, such as:
* Create visuals
* Track "catch progress/count"
* Spawn items

### Possible drawbacks
Dunno

### Applicable Issues
It's inserted right after the `npc.catchItem` is spawned. Maybe make the hook allow the modder to decide if the item should be spawned in or not.

### Sample Usage
```cs
//public class MyNPC : ModNPC
public override void OnCatchNPC(Player player)
{
    Main.NewText(player.name + " just catched " + npc.GivenOrTypeName);
}

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalNPC.cs
@@ -192,6 +192,14 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
+		/// Allows you to make things happen when an NPC is catched. Ran Serverside.
+		/// </summary>
+		/// <param name="npc">The catched NPC</param>
+		/// <param name="player">The player catching the NPC</param>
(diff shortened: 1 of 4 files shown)


direwolf420:
Added item as a parameter. If you currently wish to despawn the item properly (for example, to spawn a different one) you need to write this in the implementing hook:
```cs
item.TurnToAir();
NetMessage.SendData(MessageID.SyncItem, -1, -1, null, item.whoAmI, 1f, 0f, 0f, 0, 0, 0);

PR #682, opened 2019-09-20
Add ToFromStringConverter example to ExampleConfig

direwolf420:
### Description of the Change
If you work with ModConfigs you most likely will come across Dictionaries and want to use them with non-primitive data types aswell. An example of such are the `EntityDefinition` ones provided by tModLoader. Those use the `TypeConverter` attribute so they are able to be saved as a string representation inside the config file, and loaded back as an object. Since `ToFromStringConverter` is only used internally by tml, I want to provide an example that is easily understood and explain why it is needed if you plan to use your custom class as a key type in a Dictionary.

### Alternate designs
To be in line with ExampleConfigs examples, I used simple to understand variable names and comments where necessary. The way I did the `ToString` and `FromString` can be changed, and I will be happy for suggestions on how to improve those.

Just an example showcasing something that is also internally used by tml.

### Benefits
Give modders more possibilities to use ModConfig.

### Possible drawbacks
not applicable.

### Applicable Issues
There is no try/catch inside the `FromString`, which causes illegal manipulations to the file to fail loading the key, and the mod will crash. Awaiting suggestions, see Alternate Designs.

### Sample Usage
See commit

### DIFF
--- a/ExampleMod/ExampleConfig.cs
+++ b/ExampleMod/ExampleConfig.cs
@@ -155,8 +155,8 @@ namespace ExampleMod
 			[new PrefixDefinition("ExampleMod", "ReallyAwesome")] = 0.8f
 		};
 
-        // Using a custom class as a key in a Dictionary (see the ClassUsedAsKey definition by Ctrl+Leftclick on the class name)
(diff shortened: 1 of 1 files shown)


Jofairden:
I like this PR, but configs are jopo's domain atm so we'll let him decide.

PR #691, opened 2019-10-08
Update issue template with more relevant info and less caps

Chicken-Bones:


ISSUE_TEMPLATE.md
```diff
@@ -37,18 +37,5 @@ Please keep this as concise as possible, preferably a single line describing the
 ### Reproduction frequency
 <!-- How often are you reliably able to reproduce this issue? -->
 
-### Versions
-<!-- Which versions are applicable for this issue?
-The tModLoader version should be notified in the bottom left corner when in the main menu. Follow this format: `vX.X.X.X tModLoader` or `tModLoader vX.X.X.X` (MAJOR.MINOR.BUILD.REVISION) -->
```
Jofairden: Since you omit this, maybe add something like this at the end of the template:
```
<!-- If you do not use the latest tModLoader version, please tell us the version you are using -->
```

Chicken-Bones: tML version is in the logs now. If they can't give logs because they're on an old version I expect it'll be pretty apparent

Issue #388, opened 2018-07-21
Prevent ModItem from accepting a prefix


JavidPack:
Attempting to prevent any prefix from applying to an item via
`		public override int ChoosePrefix(UnifiedRandom rand)
		{
			return 0;
		}`
causes some issues. It correctly prevents a prefix on crafting the item, but it will crash when placed in Goblin Tinkerer. I think the issue is the item shouldn't return true when item.prefix(-3) is called, but since we don't the code assumes it needs to continue to try for a non 0 prefix when item.prefix(-2) is called. Since ChoosePrefix is random, the computer can't know if it is always suppose to return 0 or if it is just random. (returning 0 some of the time would be a normal thing to do, but the -3 call forces the process to repeat until a valid non-0 prefix is rolled.) I think the solution is just a property or hook for CanHavePrefix or something and make sure to return false if prefix(-3) is called.

Reminder: 
Prefix(-1): 75% chance roll prefix. OnCraft
Prefix(-2): Roll non-0 prefix. Goblin Tinkerer
Prefix(-3): return true if item can be prefixed
Prefix(0+): Assign specified prefix

Jofairden:
Well, shouldn't we just check if it returns 0? If it does, do not roll. (0 if empty prefix)
And -1 is vanilla behaviour.

JavidPack:
The problem is I think ChoosePrefix returning 0 sometimes is a valid thing to do in a mod. Imagine you want your item to be 10x more likely to be crafted (-1) without a prefix, you can return 0 90% of the time and -1 the other 10%. This would still work with tinkerer (-2) but not reliably with the tinkerers slot (-3). I guess you can say this is an example of the halting problem.

We don't know if the 0 return represents "never give this item a prefix" or "this roll for prefix resulted in 0....but it could still roll something else if you try again."

I think we also discussed before allowing ModItems to manually specify which Prefix set they want to use. Currently we use some item fields to guess, but that might not work for all mods.

direwolf420:
Bumping because of frequent requests in the discord. More example uses:
preserving the rarity color on an item
wiping all other modded prefixes on an item without any vanilla damage types

GoodPro712:
could this get closed now?

PR #704, opened 2019-10-31
Prefix hook fixes and use examples for ExampleMod

direwolf420:
### Description of the Change
As requested by Chicken Bones, here are examples of usages for all three prefix related hooks `ChoosePrefix`, `AllowPrefix` and `PrefixChance`. With them come a few small fixes required for all the three hooks to work accordingly so their intended usage doesn't overlap. This is especially related to #388. ChoosePrefix hook now ignores 0 as an input (treats it as -1), since the previously planned function for allowing or not allowing is now handled by `AllowPrefix`, and chances by `PrefixChance`.

* Control reforgability or prefixes an item receives dynamically
* Preserving the rarity color on an item
* Wiping all other modded prefixes on an item without any vanilla damage types

### Sample Usage
ExampleMod code provided

### DIFF
--- a/ExampleMod/Items/Accessories/ManaHeart.cs
+++ b/ExampleMod/Items/Accessories/ManaHeart.cs
@@ -1,6 +1,7 @@
 ﻿using Terraria;
 using Terraria.ID;
 using Terraria.ModLoader;
+using Terraria.Utilities;
 
 namespace ExampleMod.Items.Accessories
 {
@@ -20,6 +21,11 @@ namespace ExampleMod.Items.Accessories
 
 		public override void UpdateAccessory(Player player, bool hideVisual) {
 			player.GetModPlayer<ExamplePlayer>().manaHeart = true;
+		}
+
+		public override int ChoosePrefix(UnifiedRandom rand) {
+			// When the item is given a prefix, only roll the best modifiers for accessories
+			return rand.Next(new int[] { PrefixID.Arcane, PrefixID.Lucky, PrefixID.Menacing, PrefixID.Quick, PrefixID.Violent, PrefixID.Warding });
 		}
 
 		public override void AddRecipes() {
--- a/ExampleMod/Items/PrefixChanceGlobalItem.cs
+++ b/ExampleMod/Items/PrefixChanceGlobalItem.cs
@@ -15,7 +15,7 @@ namespace ExampleMod.Items
 	{
 		public override bool? PrefixChance(Item item, int pre, UnifiedRandom rand) {
 			// pre: The prefix being applied to the item, or the roll mode
-			// -1 is when an item is bought from a shop, crafted or generated in a chest
+			// -1 is when an item is naturally generated in a chest, crafted, purchased from an NPC, looted from a grab bag (excluding presents), or dropped by a slain enemy
 			// -2 is when an item is rolled in the tinkerer
 			// -3 determines if an item can be placed in the tinkerer slot
 
@@ -35,7 +35,7 @@ namespace ExampleMod.Items
 
 			// To prevent rolling of a prefix on spawn, return false when pre is -1
 			if (pre == -1) {
-				if (item.melee && item.modItem != null && item.modItem.mod.Name == mod.Name) {
+				if (item.melee && item.modItem?.mod == mod) {
 					// All melee weapons from ExampleMod won't have a prefix when they are crafted, bought, taken from a generated chest, opened, or dropped by an enemy
 					return false;
 				}
--- a/ExampleMod/Items/PrefixChanceGlobalItem.cs
+++ b/ExampleMod/Items/PrefixChanceGlobalItem.cs
@@ -1,0 +1,66 @@
+﻿using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+using Terraria.Utilities;
+
+namespace ExampleMod.Items
+{
+	/// <summary>
+	/// This class demonstrates how to manipulate the chances of prefixes given to items.
+	/// For other prefix related hooks and their usage, see also:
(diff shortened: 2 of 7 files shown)


ExampleMod/Items/Accessories/ManaHeart.cs
```diff
@@ -22,6 +23,11 @@ public class ManaHeart : ModItem
 			player.GetModPlayer<ExamplePlayer>().manaHeart = true;
 		}
 
+		public override int ChoosePrefix(UnifiedRandom rand) {
+			// When the item is given a prefix, only roll the best modifiers for accessories
```
Chicken-Bones: very nice

ExampleMod/Items/PrefixChanceGlobalItem.cs
```diff
@@ -0,0 +1,66 @@
...
+	{
+		public override bool? PrefixChance(Item item, int pre, UnifiedRandom rand) {
+			// pre: The prefix being applied to the item, or the roll mode
+			// -1 is when an item is bought from a shop, crafted or generated in a chest
+			// -2 is when an item is rolled in the tinkerer
+			// -3 determines if an item can be placed in the tinkerer slot
+
+			// To prevent putting an item in the tinkerer slot, return false when pre is -3
+			if (pre == -3 && item.type == ItemID.LaserRifle) {
+				// This will make the Laser Rifle not be reforgeable at all (useful if you want your item to preserve its custom name color)
```
Chicken-Bones: good example, using vanilla items

ExampleMod/Items/PrefixChanceGlobalItem.cs
```diff
@@ -0,0 +1,66 @@
...
+				return false;
+			}
+
+			// To make an item reset its prefix when reforging
+			if (pre == -2) {
+				if (Main.LocalPlayer.HasBuff(BuffID.Cursed)) {
+					// If the player is cursed, make it remove the prefix
```
Chicken-Bones: cool idea

ExampleMod/Items/PrefixChanceGlobalItem.cs
```diff
@@ -0,0 +1,66 @@
...
+					return false;
+				}
+			}
+
+			// To prevent rolling of a prefix on spawn, return false when pre is -1
+			if (pre == -1) {
+				if (item.melee && item.modItem != null && item.modItem.mod.Name == mod.Name) {
+					// All melee weapons from ExampleMod won't have a prefix when they are crafted, bought, taken from a generated chest, opened, or dropped by an enemy
```
Chicken-Bones: also neat, just change this to `item.melee && item.modItem?.mod == mod`

ExampleMod/Items/Weapons/ExampleYoyo.cs
```diff
@@ -38,6 +39,25 @@ public class ExampleYoyo : ModItem
 			item.shoot = ProjectileType<ExampleYoyoProjectile>();
 		}
 
+		// Make sure that your item can even receive these prefixes (check the vanilla wiki on prefixes)
```
Chicken-Bones: s/even/never?

ExampleMod/Items/Weapons/ExampleYoyo.cs
```diff
@@ -38,6 +39,25 @@ public class ExampleYoyo : ModItem
...
+		// These are the ones that reduce damage of a melee weapon
+		private static readonly int[] unwantedPrefixes = new int[] { PrefixID.Terrible, PrefixID.Dull, PrefixID.Shameful, PrefixID.Annoying, PrefixID.Broken, PrefixID.Damaged, PrefixID.Shoddy};
+
+		public override bool AllowPrefix(int pre) {
+			// return false to make the game reroll the prefix
+
+			// DON'T DO THIS BY ITSELF:
```
Chicken-Bones: good warning

ExampleMod/Items/Weapons/ExampleYoyo.cs
```diff
@@ -38,6 +39,25 @@ public class ExampleYoyo : ModItem
...
+			// return false;
+			// This will get the game stuck because it will try to reroll every time. Instead, make it have a chance to return true
+
+			if (Array.IndexOf(unwantedPrefixes, pre) > -1) {
```
Chicken-Bones: `return Array.,IndexOf(unwantedPrefixes, pre) == -1`
With comment "return true (allow prefix) if not in the array of unwantedPrefixes (IndexOf returns -1 if item not found)"

Chicken-Bones (commented):
Very minor changes requested. Excellent examples. Many thanks.

ExampleMod/Items/Weapons/ExampleYoyo.cs
direwolf420: Basically, why putting prefixes in that array, if this item can't even get that prefix assigned by the game anyway? Certain item types can get different prefixes (weapons vs accessories, knockback vs no knockback etc.)

ExampleMod/Items/Weapons/ExampleYoyo.cs
direwolf420: I usually don't like "compound" statements like these in a "showcase" setting, the logic is less easy to follow. I'll just explain the method more instead

direwolf420: plus, IndexOf doesn't always return -1 when not found :p 
`Because most arrays have a lower bound of zero, this method generally returns -1 if value isn't found. In the rare case that the lower bound of the array is equal to Int32.MinValue(0x80000000) and value isn't found, this method returns Int32.MaxValue (0x7FFFFFFF).`

Issue #522, opened 2019-04-21
ModPlayer.PreHurt and playSound example


JavidPack:
We need a sound to play while the player gets hurt while wearing ExampleCostume. This will show off how to accomplish FrostArmor type hurt sound replacements.

PR #711, opened 2019-11-08
FrostArmor type hurt sound replacements.

KryptonIon:
### Description of the Change
When the player has a `Charm of Example` or`ExampleCostume` worn, this will make the `blockyAccessory` field in ExamplePlayer true, I have added 2 lines of code that check if this field is true, one disables the default sound([Line 452](https://github.com/KryptonIon/tModLoader/blob/5e0b7d544f145dd20f8c55bb2fe943239e5272ef/ExampleMod/ExamplePlayer.cs#L452)) and the other plays a the frog sound([Line 460](https://github.com/KryptonIon/tModLoader/blob/5e0b7d544f145dd20f8c55bb2fe943239e5272ef/ExampleMod/ExamplePlayer.cs#L460))

### Alternate designs
N/A

Quoted from the issue this is trying to fix,
> This will show off how to accomplish FrostArmor type hurt sound replacements.

### Benefits
Modders can simply look at ExampleMod instead of asking on Discord.

### Possible drawbacks
N/A

### Applicable Issues
#522

### Sample Usage
`if (blockyAccessory) playSound = false;` In PreHurt
`if (blockyAccessory) Main.PlaySound(SoundID.Frog, player.position);` In Hurt

### DIFF
--- a/ExampleMod/ExamplePlayer.cs
+++ b/ExampleMod/ExamplePlayer.cs
@@ -114,7 +114,6 @@ namespace ExampleMod
 			packet.Write((byte)ExampleModMessageType.ExamplePlayerSyncPlayer);
 			packet.Write((byte)player.whoAmI);
 			packet.Write(exampleLifeFruits);
-			packet.Write(blockyAccessory);
 			packet.Write(nonStopParty); // While we sync nonStopParty in SendClientChanges, we still need to send it here as well so newly joining players will receive the correct value.
 			packet.Send(toWho, fromWho);
 		}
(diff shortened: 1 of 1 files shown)


KryptonIon:
I still need to prove the sound can be heard by other players in Multiplayer.

JavidPack:
Adding to SendClientChanges was incorrect, blockyAccessory is known on all clients automatically. You also only added it to the Send and not the HandlePacket, so it was breaking the nonStopParty data sync.

Also, SoundID.Frog is an ambiance sound (preventing it from playing while the game is out of focus or ambient sound turned off), so I switched it to Zombie 13, which is the same sound, but not ambiance.

PR #714, opened 2019-11-12
Add ExampleSimpleMinion

direwolf420:
### Description of the Change
Adds a simple minion for ExampleMod, a guide on minions will follow very soon on the wiki.
In additon, fixes a small issue with the other ExampleMod minion summon weapon.

### Alternate designs
All necessary files for a minion to work were put in the same file, due to beginners getting confused about the previous minion example having its files split across the whole mod. They might've been put in their proper locations aswell.

### Benefits
Very basic, straight to the point minion "template" with the necessary classes around it. Making it easier for beginners to understand what is going on and how multiple classes interact with eachother directly, or through the game itself.

### Possible drawbacks
Bad habits because of everything being in the same file. But a warning has been put in the header comment.


JavidPack:
Thanks a bunch, this will be very helpful.

PR #730, opened 2019-11-30
Can burn in lava hook take 2

GoodPro712:
### Description of the Change
Added CanBurnInLava hook. Return true in a ModItem to allow that item to burn in lava

### Alternate designs
Not applicable.

Because tModLoader currently does not have a way to make items burn in lava without having a white rarity

### Benefits
People can make certain items burn in lava

### Possible drawbacks
Not applicable.

### Applicable Issues
Possibly add a global CanBurnInLava if more people request it

### Sample Usage
There is an example of the code in ExampleMod\Items\ExampleBurnableItem.cs


JavidPack:
Thanks.

Jofairden:
Thanks for your contribution, but you forgot about GlobalItem. @JavidPack ?

PR #742, opened 2019-12-20
Created internal Wave Timer Field in DD2Event

GoodPro712:
Added this so that modders can change what the timer says via reflection. 
I am wanting to do this, that's why I contributed it.
I don't know what else to say here..


GoodPro712:
made public

Mirsario:
It **would** be a bad addition. There's zero chance anyone else would want to do this, and from CB's words, after MonoMod has been added, tML's hooks' main role is improving mod compatibility. No correlation with that. Use IL edits to do what you want.

GoodPro712:
Not many people can perform IL edits, but I do get your point

Chicken-Bones:
Firstly, thanks for taking the time to contribute. We're always looking for more community involvement. In the current form, I don't think this addition fits what we're aiming for with tModLoader. It would be good to have that clarified somewhere, maybe in the GitHub first post. 

If you're looking for a way to get your mod to work, or think it'd be usefully for others, please come chat to us in the Discord. We can help you refine it or consider alternative approaches.

PR #743, opened 2019-12-22
Fix hotbar and and recipe list moving when scrolling a UI scrollbars.

TehCupcakes:
### Description of the Change

- `Main.InGameUI.Update()` -> Technically this happens first, but because the player's input hasn't been read yet, it's actually one step behind for picking up scrolls.
- `PlayerInput.UpdateInput()` -> Sets scroll wheel values.
- `Player.Update()` -> Updates the hotbar/recipes position based on scroll value.

This creates a problem, since we don't want the hotbar to move if a scrollbar has been scrolled, but we haven't reached that point yet. It's also not possible (or at least quite difficult) to identify if a scrollbar _will_ update at the point of `Player.Update()`, so the only option is to rearrange the order a bit. My change moves the hotbar update portion outside the player's normal `Update` method, and instead places it just after the UI updates. That way the UI can set a variable if it was handled and let the player instance know to skip cycling the hotbar.

This may not the cleanest way to do this. But it was a bug that bothered me and I hope this PR gives some ideas if somebody wants to implement it differently.

### Alternate designs
I tried identifying in the player update method whether the user was hovering on a UI element with a scrollbar. I did not get this working, hence why I went with the design I did.

### Benefits
Better user experience.

### Possible drawbacks
Some mods may have relied on the old behavior. For instance, BossChecklist would calculate how much the hotbar moved and reverse it. [(See here.)](https://github.com/JavidPack/BossChecklist/blob/master/BossChecklistUI.cs#L205) This fix breaks the workaround, but obviously it's better to have a proper fix in tModLoader that prevents stutter.

### Applicable Issues
N/A

### Sample Usage
N/A


Chicken-Bones:
I haven't had time to look at this yet, but just reading the description, have you considered the alternate design of just delaying hotbar scroll processing by 1 tick, and saving the 'hotbar scroll from last tick' in the UI update?

It wouldn't cause any noticeable lag and might be more maintainable

TehCupcakes:
That's not a bad idea, but I think it would be a bit more drastic, since in vanilla it is managed by the Player class rather than UserInterface. And since the UserInterface update runs for each UIElement, you would still need to find a way to place the hotbar update after it has finished all the UIElements. The end result would be not much different than what I did implement, it would just relocate the hotbar/recipe logic outside the player. (Which I think would be appropriate, but that's not how the code was originally written. ¯\_(ツ)_/¯)

TehCupcakes:
I did what you suggested and I think this is working nicely now. Using the built-in UI elements, anything with a scrollbar should reset the value to 0. For any mods that override the ScrollWheel event, they can set this value themselves if they want different behavior.

TehCupcakes:
Replaced with #749.

PR #749, opened 2019-12-26
Fix scrolling a UI scrollbar also adjusting hotbar position.

TehCupcakes:
**NOTE:** This replaces #743. My other branch got a little screwy so I made a fresh one.

### Description of the Change

- `Main.InGameUI.Update()` -> Technically this happens first, but because the player's input hasn't been read yet, it's actually one step behind for picking up scrolls.
- `PlayerInput.UpdateInput()` -> Sets scroll wheel values.
- `Player.Update()` -> Updates the hotbar/recipes position based on scroll value.

This creates a problem, since we don't want the hotbar to move if a scrollbar has been scrolled, but we haven't reached that point yet. It's also not possible (or at least quite difficult) to identify if a scrollbar _will_ update at the point of `Player.Update()`. My change records the input scroll value when attempting to move the hotbar and delays processing by one cycle. This gives the UI elements and mods an opportunity to reset the value if they handle the scrolling and want to cancel out the hotbar offset.

### Alternate designs
I originally relocated the hotbar handling so it would happen just after the UI update, and added a condition that it would only update if a scrollbar event had not been handled. However, this was a bit of a more significant change to the original code, so it was suggested that I try delaying the processing a frame instead of changing the order altogether.

I also tried identifying in the player update method whether the user was hovering on a UI element with a scrollbar. I did not get this working, and decided there must be a better method.

### Benefits
Better user experience.

### Possible drawbacks
Some mods may have relied on the old behavior. For instance, BossChecklist (and many of @JavidPack's other mods) would calculate how much the hotbar moved and reverse it. [(See here.)](https://github.com/JavidPack/BossChecklist/blob/master/BossChecklistUI.cs#L205) This fix breaks the workaround, but obviously it's better to have a proper fix in tModLoader that prevents stutter.

### Applicable Issues
N/A

### Sample Usage
N/A


Chicken-Bones:
Looks good to me. @JavidPack please sign off

Solxanich:
@TehCupcakes I've ported this PR to tML's active development branch [1.4], and given tML isn't going to be updating 1.3 in the foreseeable future, we are hoping to close this PR targeting that branch.

If you are happy with the port, please consider closing this PR. Thank you for your time.

TehCupcakes:
I'm cool with that! Just would like to see a fix get merged one way or another. :)

PR #755, opened 2019-12-29
Allow for internal loops in .ogg files

jjohnsnaill:
### Description of the Change
Allows for music in the .ogg format to have internal loops. If the file has the tag `LOOPSTART`, the song will jump to the tag's value instead of the beginning once it reaches the end (or the value of `LOOPEND`, if specified).

### Alternate designs
N/A

Mods cannot do this on their own. At least, not without code injection.

### Benefits
Songs with intros can use `LOOPSTART` to stop the intro from playing more than once, and both `LOOPSTART` and `LOOPEND` can be used to give songs proper loops if they lack them.

### Possible drawbacks
N/A

### Applicable Issues
Doesn't account for `LOOPSTART` being below 0 or higher than the song's length in samples, but 1. an exception is thrown if it is, and 2. why would it ever be set to either values?

### Sample Usage
N/A. The tags cannot be added via code.

### Comments
This is my first pull request, so I hope I did everything correctly!

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/MusicWrappers.cs
+++ b/patches/tModLoader/Terraria.ModLoader/MusicWrappers.cs
@@ -211,6 +211,8 @@ namespace Terraria.ModLoader.Audio
 	{
 		private VorbisReader reader;
 		private float[] floatBuf;
+		private int loopStart;
+		private int loopEnd;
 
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Looks great, I can't see any flaws.

PR #782, opened 2020-01-31
ExampleMod editorconfig

GoodPro712:
### Description of the Change

Added the .editorconfig file to ExampleMod

### Alternate designs
N/A

To help contributors format ExampleMod appropriately

### Benefits
Contributors would now have an easier way to format contributions to ExampleMod 

### Possible drawbacks
I don't see any...

### Applicable Issues
I don't see any...

### Sample Usage
N/A

### DIFF
--- a/ExampleMod/.editorconfig
+++ b/ExampleMod/.editorconfig
@@ -1,0 +1,47 @@
+root = true
+
+[*.cs]
+
+#Core editorconfig formatting - indentation
+
+#use hard tabs for indentation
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
As long as this is a copy of the one in the repo root. Normally you only need one but running ExampleMod from the Mod Sources folder means it's not part of the heirachy

GoodPro712:
yes, it is

PR #788, opened 2020-02-06
ExampleDashAccessory and ExampleDashPlayer

absoluteAquarian:
### Description of the Change
Added an ExampleDashAccessory accessory and ExampleDashPlayer player.
Both of these classes work in conjunction with each other to show how to implement custom dashes.

### Alternate designs
n/a

How dashes work in vanilla is fairly obscure, since only 3 items and one armor set (the Solar Armor set) implement a dash.
This could potentially be integrated into the tModLoader framework via PlayerHooks, but it should suffice for now to have them in ExampleMod as a reference for all modders to use.

### Benefits
See the previous section.
Having a readily available example for custom dashing would be beneficial to all modders who want to do so in their own mods.

### Possible drawbacks
n/a

### Applicable Issues
Multiplayer compatability?  Still needs to be tested.

### Sample Usage
n/a


ExampleMod/Items/ExampleDashAccessory.cs
```diff
@@ -0,0 +1,208 @@
...
+			item.rare = ItemRarityID.Blue;
+			item.value = Item.sellPrice(silver: 60);
+		}
+
+		public override void UpdateAccessory(Player player, bool hideVisual)
+		{
+			ExampleDashPlayer mp = player.GetModPlayer<ExampleDashPlayer>();
+
+			//If the dash is active, decrement the timers and do other things
+			if(mp.DashActive)
```
Chicken-Bones: `if (!mp.DashActive) return;`

ExampleMod/Items/ExampleDashAccessory.cs
```diff
@@ -0,0 +1,208 @@
...
+		//This array will be used to keep track of the previous values of "player.doubleTapCardinalTimer"
+		private int[] oldCardinalTimers = new int[4];
+
+		//These indicate what direction is what in the timer arrays used
+		private readonly int Down = 0;
+		private readonly int Up = 1;
+		private readonly int Right = 2;
+		private readonly int Left = 3;
+
+		//Bools for if a direction was pressed in any of the cardinal directions
```
Chicken-Bones: You could use an array here, it'd be nicer to see you use Up/Down/Right/Left for all `oldCardinalTimers` `Pressed` and `DoubleTap`

ExampleMod/Items/ExampleDashAccessory.cs
```diff
@@ -0,0 +1,208 @@
...
+
+			//Check if the ExampleDashAccessory is equipped and also check against this priority:
+			// If the Shield of Cthulhu, Master Ninja Gear, Tabi and/or Solar Armour set is equipped, prevent this accessory from doing its dash effect
+
+			//The priority is used to prevent undesirable effects.
+			//Without it, the player is able to use the ExampleDashAccessory's dash as well as the vanilla ones
+			bool dashAccessoryEquipped = false;
+			bool shieldOfCthulhuEquipped = false;
+			bool masterNinjaGearEquipped = false;
+			bool tabiEquipped = false;
```
Chicken-Bones: While this is explanatory, you use all of these in the same function. You could have a single variable, or better yet, just change all your `xAccessoryEquipped = true;` to `return;` and keep the same comments about what you're doing

ExampleMod/Items/ExampleDashAccessory.cs
```diff
@@ -0,0 +1,208 @@
...
+			}
+		}
+
+		public override void PostUpdate()
+		{
+			//This hook will be used to assign the values in our oldCardinalTimers array to the player.doubleTapCardinalTimer array
+
+			//Check if each direction was pressed.  If it was, store the old value
+			//Using a helper method to shorten repetition
+			UpdateTimer(Down,  PressedDown);
```
Chicken-Bones: Is there a reason this is in PostUpdate rather than ResetEffects?

ExampleMod/Items/ExampleDashAccessory.cs
```diff
@@ -0,0 +1,208 @@
...
+			UpdateTimer(Up,    PressedUp);
+			UpdateTimer(Right, PressedRight);
+			UpdateTimer(Left,  PressedLeft);
+		}
+
+		private void UpdateTimer(int timerIndex, bool condition)
+		{
+			if(condition)
```
Chicken-Bones: The way you detect double taps isn't very well explained, I had to consult the terraria source to understand how it's supposed to work. By my read:

`ResetEffects` is called after the vanilla `doubleTapCardinalTimer` logic for a tick. On the first press-release, the timer is set to 15. At this point your `condition` will be true here, and you'll copy the value `15` into `oldCardinalTimers`.

Vanilla gives 15 ticks for the second tap (1/4 of a second) by counting down this variable. It doesn't reset either, so it looks to me like you could just use the same `doubleTapCardinalTimer[l] > 0` check directly.

Your approach of checking if the current doubletap timer is > then the one when the key was pressed the first time (`oldCardinalTimers`), and then clearing it when it reaches zero works, but you're effectively just doing `player.doubleTapCardinalTimer < 15 && player.doubleTapCardinalTimer > 0`
the first of which will always be true

PR #793, opened 2020-02-14
ExampleFlail

GoodPro712:
Example Flail (finally)
I don't know what variable is for the length of the flail, I copied of redcode.


GoodPro712:
sorry for the crappy job

PR #798, opened 2020-02-21
When compiling and loading mods, check for {Dep}.XNA.dll and {Dep}.FNA.dll

tansheron:
This change adapts #417 to work with XNA/FNA instead of Mono/Windows. A bit of extra code is required to support noCompile mods because tML needs to include both dlls if they exist, rather than just picking one as is done during compilation.

Updated sample usage:

* In build.txt, dllReferences = CoolLibrary
* In lib folder,
  * CoolLibrary.dll (XNA)
  * CoolLibrary.FNA.dll
* Compiled {modName}.XNA.dll uses CoolLibrary.dll
* Compiled {modName}.FNA.dll uses CoolLibrary.FNA.dll

Since the purpose of this request is the same as that of #417, please refer to the request description there.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
@@ -376,7 +376,7 @@ namespace Terraria.ModLoader.Core
 
 			// add dll references from the bin folder
 			var libFolder = Path.Combine(mod.path, "lib");
-			foreach (var dllPath in mod.properties.dllReferences.Select(dllName => DllRefPath(mod, dllName)))
+			foreach (var dllPath in mod.properties.dllReferences.SelectMany(dllName => DllRefPathsForPackaging(mod, dllName)))
 				if (!dllPath.StartsWith(libFolder))
 					mod.modFile.AddFile("lib/"+Path.GetFileName(dllPath), File.ReadAllBytes(dllPath));
 		}
@@ -554,7 +554,7 @@ namespace Terraria.ModLoader.Core
 				.Where(path => !path.EndsWith("Thunk.dll") && !path.EndsWith("Wrapper.dll")));
 
 			//libs added by the mod
-			refs.AddRange(mod.properties.dllReferences.Select(dllName => DllRefPath(mod, dllName)));
+			refs.AddRange(mod.properties.dllReferences.Select(dllName => DllRefPathForCompilation(mod, dllName, xna)));
 
 			//all dlls included in all referenced mods
 			foreach (var refMod in refMods) {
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
```diff
@@ -376,7 +376,7 @@ private class BuildingMod : LocalMod
 
 			// add dll references from the bin folder
 			var libFolder = Path.Combine(mod.path, "lib");
-			foreach (var dllPath in mod.properties.dllReferences.Select(dllName => DllRefPath(mod, dllName)))
+			foreach (var dllPath in mod.properties.dllReferences.SelectMany(dllName => DllRefPathsForPackaging(mod, dllName)))
```
Chicken-Bones: The only purpose of this code branch is to copy dll references from the directory provided by the `-eac` flag. This allows people building with vs to ignore the lib folder. Platform dependent dll paths are only in the `lib` folder so they'll just get ignored in the next `if` statement. The entire `lib` folder is packaged wholesale regardless. 

This means you can probably remove `DllRefPathsForPackaging` and combine `DllRefPathForCompilation` with `DllRefPath`

Chicken-Bones:
the purpose of searching the -eac path for libs was mostly to support nuget packages. Platform specific dllReferences should just be copied to the lib folder. There's no need to seek them in the -eac path as well.

This could be revised but needs more discussion. Feature is functional with current implementation

patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
```diff
@@ -595,38 +595,14 @@ private class BuildingMod : LocalMod
...
-		private IEnumerable<string> DllRefPathsForPackaging(BuildingMod mod, string dllName)
-		{
-			string path = Path.Combine(mod.path, $"lib/{dllName}");
-			string xnaRefPath = DllRefPath(path, dllName, true);
-			string fnaRefPath = DllRefPath(path, dllName, false);
-
-			string[] refPaths = new[] {xnaRefPath, fnaRefPath}.Where(p => p != null)
-				.Distinct().ToArray();
+		private string DllRefPath(BuildingMod mod, string dllName, bool? xna) {
+			var path = Path.Combine(mod.path, "lib", dllName + ".dll");
```
tansheron: The code below already appends the extension, so this results in file names like "DllName.dll.dll" or "DllName.dll.XNA.dll".

patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
```diff
@@ -595,38 +595,14 @@ private class BuildingMod : LocalMod
...
-		}
-
-		private string DllRefPath(string path, string dllName, bool xna)
-		{
-			//check if .XNA.dll /.FNA.dll exists when compiling for that platform
-			string suffixPlat = xna ? ".XNA.dll" : ".FNA.dll";
-			string pathPlat = path + suffixPlat;
-			if (File.Exists(pathPlat))
-				return pathPlat;
+			if (xna.HasValue) { //check for platform specific dll
```
tansheron: In this version of PackageMod, xna will always be null, so the check for platform-specific paths will never happen for noCompile mods. This functionality was the primary reason why I made this pull request and why I used SelectMany in PackageMod.

PR #803, opened 2020-02-25
Example Last Prism

Ozzatron:
### Description of the Change
Adds an example item to ExampleMod called *Example Last Prism*.

Example Last Prism is a laser-firing prism item which is very similar to its vanilla counterpart, but has some notable differences:
- It is colored blue to visually distinguish it from the vanilla item
- All of the laser beams it fires are blue to purple in color
- It fires 10 laser beams instead of 6
- It deals less damage, but uses local NPC immunity frames so all the laser beams strike simultaneously when focused together
- The beams deal 150% damage instead of 300% when fully focused
- Mana cost and damage smoothly scale as the prism focuses
- Beams determine the prism they are based on via UUID when spawned, making them display properly in multiplayer.

### Alternate designs
It was discussed by @Mirsario that having the laser use local NPC immunity frames is not a feature that anyone would want. I kept it as is because I believe Last Prism is a great weapon to demonstrate that you can have many projectiles striking one NPC at once using local immunity frames.

This is an example item for Example Mod which modders can reference to see how to make holdout projectiles, lasers, and how to combine them to make an extremely complex vanilla item.

The *Calamity Mod* contains an example of this sort of item ([Yharim's Crystal](http://calamitymod.gamepedia.com/Yharim's_Crystal)), but is closed source. @Chicken-Bones has personally expressed that he wants a Last Prism-like weapon in Example Mod.

### Benefits
- Provides an example of a holdout projectile
- Provides an example laser beam which can be swung around
- Demonstrates that Prism weapons can display correctly in multiplayer
- Provides an example of a projectile which uses Terraria's Projectile UUID system

### Possible drawbacks
**N/A**

### Applicable Issues
There are a few variables or other pieces inside the code which I do not fully understand. I can amend the pull request with additional commits which meaningfully comment those variables if another user could explain to me what they do.

### Sample Usage
This item is in Example Mod. It is an example item, and it is its own sample usage.

**::edit:: 05/08/2020** Added a few commits. One fixes a minor graphical bug and the other fixes multiplayer display issues. These issues exist with the vanilla Last Prism.


ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+		private float PerformBeamHitscan(Projectile prism, bool fullCharge)
+		{
+			// By default, the hitscan interpolation starts at the projectile's center.
+			// If the host Prism is fully charged, the interpolation starts at the Prism's center instead.
+			// Overriding that, if the player shoves the Prism into or through a wall, the interpolation starts at the player's center.
+			// This last part prevents the player from projecting beams through walls under any circumstances.
+			Vector2 samplingPoint = projectile.Center;
+			if (fullCharge)
```
direwolf420: This is the first occurence of doing "one line" if statements on the next line. There are countless ones in your code. It's not something examplemod does, afaik it just does 
```cs
if (fullCharge) {
    samplingPoint = prism.Center;
}

File: ExampleMod/Projectiles/ExampleLastPrismBeam.cs [hunk 2]
```diff
@@ -0,0 +1,353 @@
...
+				samplingPoint = prism.Center;
+			if (!Collision.CanHitLine(Main.player[projectile.owner].Center, 0, 0, prism.Center, 0, 0))
```
direwolf420: You could create a local for `Main.player[projectile.owner]` since you use it twice

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,353 @@
...
+			DelegateMethods.f_1 = 1f; // f_1 is an unnamed decompiled variable whose function is unknown. Leave it at 1.
+			Vector2 beamStartPos = centerFloored - Main.screenPosition;
+			Vector2 beamEndPos = beamStartPos + projectile.velocity * visualBeamLength;
+			Utils.LaserLineFraming llf = new Utils.LaserLineFraming(DelegateMethods.RainbowLaserDraw);
+
+			// Draw the outer beam.
+			// c_1 is an unnamed decompiled variable which is the render color of the beam drawn by DelegateMethods.RainbowLaserDraw.
+			Color outerBeamColor = GetBeamColor();
+			DelegateMethods.c_1 = outerBeamColor * OuterBeamOpacityMultiplier * projectile.Opacity;
+			Utils.DrawLaser(Main.spriteBatch, tex, beamStartPos, beamEndPos, scaleVec, llf);
```
direwolf420: use just `spriteBatch`, not `Main.spriteBatch` (even tho its the same object 99% of the time, just better using the passed one instead of Main.)

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			// Update the Prism's position in the world and relevant variables of the player holding it.
+			UpdatePlayerVisuals(player, rrp);
+
+			// Update the Prism's behavior: project beams on frame 1, consume mana, and despawn if out of mana.
+			if (projectile.owner == Main.myPlayer)
+			{
+				// Slightly re-aim the Prism every frame so that it gradually sweeps to point towards the mouse.
+				UpdateAim(rrp, player.inventory[player.selectedItem].shootSpeed);
```
direwolf420: `player.inventory[player.selectedItem]` is `player.HeldItem`, same a few lines below

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			// This ensures that the Prism never times out while in use.
+			projectile.timeLeft = 2;
+		}
+
+		private void UpdateDamageForManaSickness(Player player)
+		{
+			float ownerCurrentMagicDamage = player.allDamage + (player.magicDamage - 1f);
+			projectile.damage = (int)((player.HeldItem?.damage ?? 0) * ownerCurrentMagicDamage);
```
direwolf420: does `player.HeldItem?.damage ?? 0` have any reason to exist? afaik HeldItem just gets inventory[selectedItem]

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,216 @@
...
+			player.itemTime = 2;
+			player.itemAnimation = 2;
+			player.itemRotation = (float)Math.Atan2(projectile.velocity.Y * projectile.direction, projectile.velocity.X * projectile.direction);
+		}
+
+		// Completely custom drawcode because the Prism is a holdout projectile.
+		public override bool PreDraw(SpriteBatch spriteBatch, Color lightColor)
+		{
+			SpriteEffects eff = projectile.spriteDirection == -1 ? SpriteEffects.FlipHorizontally : SpriteEffects.None;
+			Texture2D tex = Main.projectileTexture[projectile.type];
```
direwolf420: use `tex` in the next line since you declared it

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,216 @@
...
+			int frameHeight = Main.projectileTexture[projectile.type].Height / Main.projFrames[projectile.type];
+			int texYOffset = frameHeight * projectile.frame;
+			Vector2 sheetInsertVec = (projectile.Center + Vector2.UnitY * projectile.gfxOffY - Main.screenPosition).Floor();
+
+			// The Prism is always at full brightness, regardless of the surrounding light. This is equivalent to it being its own glowmask.
+			// It is drawn in a non-white color to distinguish it from the vanilla Last Prism.
+			Color drawColor = ExampleLastPrism.OverrideColor;
+			Main.spriteBatch.Draw(tex, sheetInsertVec, new Rectangle?(new Rectangle(0, texYOffset, tex.Width, frameHeight)), drawColor, projectile.rotation, new Vector2(tex.Width / 2f, frameHeight / 2f), projectile.scale, eff, 0f);
```
direwolf420: same as before with `spriteBatch`

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
Chicken-Bones: I use them in tML, and there'll be a lot in the Terraria source soon, but I tend to agree, keeping ExampleMod all braces helps newbies avoid pitfalls. Braceless is contentious as is

Ozzatron: Agreed. While it is necessary for people to learn to read braceless code (so I'm okay with TML internals and the TML formatting of Terraria source having it for example), ExampleMod is meant to be a starting point and shouldn't have potentially confusing choices in it.

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
Ozzatron: I forgot that there was a passed in sprite batch. Whoops.

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
Ozzatron: This was added with the intent of making sure that even if the player's held item is null for some reason it wouldn't throw an exception. However the rest of this projectile's code doesn't have the same protection and I am pretty sure inventory slots aren't ever null, they just become air.
```suggestion
			projectile.damage = (int)(player.HeldItem.damage * ownerCurrentMagicDamage);
```

Jofairden: It actually is still valid, because _they can_ become null (if a modder does it)
But yeah you shouldn't, you should always use `item.TurnToAir()` 👍

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,370 @@
...
+			Vector2 dustVel = dustAngle.ToRotationVector2() * dustStartDist;
+			Dust d = Dust.NewDustDirect(laserEndPos, 0, 0, dustType, dustVel.X, dustVel.Y, 0, beamColor, scale);
+			d.color = beamColor;
+			d.noGravity = true;
+
+			// If the beam is currently large, make the dust faster and larger to match.
+			if (projectile.scale > 1f) {
+				d.velocity *= projectile.scale;
+				d.scale *= projectile.scale;
+			}
```
Jofairden, line 339: In general I recommend consistent variable names like stated before etc.
You can omit things like 'dust' prefixes because this context is only about dusts etc.
You might have to explain the (Main.rand.NextBool() ? 1f : -1f) part, boolean converted to multiplier
And maybe use DustID for type
```suggestion
			const int type = 15;
			Vector2 endPosition = projectile.Center + projectile.velocity * (BeamLength - 14.5f * projectile.scale);

			// 50% chance for the dust to come off on either side of the beam.
			float angle = projectile.rotation + (Main.rand.NextBool() ? 1f : -1f) * MathHelper.PiOver2;
			float startDistance = Main.rand.NextFloat(1f, 1.8f);
			float scale = Main.rand.NextFloat(0.7f, 1.1f);
			Vector2 velocity = angle.ToRotationVector2() * startDistance;
			Dust dust = Dust.NewDustDirect(endPosition, 0, 0, type, velocity.X, velocity.Y, 0, beamColor, scale);
			dust.color = beamColor;
			dust.noGravity = true;

			// If the beam is currently large, make the dust faster and larger to match.
			if (projectile.scale > 1f) {
				dust.velocity *= projectile.scale;
				dust.scale *= projectile.scale;
			}
```

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		private void ProduceWaterRipples(Vector2 beamDims)
+		{
+			WaterShaderData wsd = (WaterShaderData)Filters.Scene["WaterDistortion"].GetShader();
```
Jofairden, line 344: I'd personally `var` this, but considering it's ExampleMod we don't. (Type is clear on other end of the statement)
```suggestion
			WaterShaderData shaderData = (WaterShaderData)Filters.Scene["WaterDistortion"].GetShader();
```

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,370 @@
...
+
+			// A universal time-based sinusoid which updates extremely rapidly. GlobalTime is 0 to 3600, measured in seconds.
+			float waveSine = 0.1f * (float)Math.Sin(Main.GlobalTime * 20f);
+			Vector2 ripplePos = projectile.position + new Vector2(beamDims.X * 0.5f, 0f).RotatedBy(projectile.rotation);
+
+			// WaveData is encoded as a Color. Not really sure why.
+			Color waveData = new Color(0.5f, 0.1f * Math.Sign(waveSine) + 0.5f, 0f, 1f) * Math.Abs(waveSine);
+			wsd.QueueRipple(ripplePos, waveData, beamDims, RippleShape.Square, projectile.rotation);
```
Jofairden, line 352:
```suggestion
			shaderData.QueueRipple(ripplePos, waveData, beamDims, RippleShape.Square, projectile.rotation);
```

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,224 @@
...
+using Microsoft.Xna.Framework.Graphics;
+using System;
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+using static Terraria.ModLoader.ModContent;
+
+namespace ExampleMod.Projectiles
+{
+	public class ExampleLastPrismHoldout : ModProjectile
```
Jofairden, line 12: All the same suggestions count for this class. Just see what you can/feel like improve.

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
```diff
@@ -0,0 +1,224 @@
...
+			DisplayName.SetDefault("Example Last Prism");
+			Main.projFrames[projectile.type] = 5;
+		}
+
+		public override void SetDefaults()
+		{
+			projectile.CloneDefaults(ProjectileID.LastPrism);
+		}
+
+		public override void AI()
```
Jofairden, line 69: You should subdivide the AI method so it's easier to read/manage
e.g.
```suggestion
		public override void AI() {
			UpdateFrame();
			PlaySound();
			UpdateVisuals();
			//... etc.
		}
```

ExampleMod/Projectiles/ExampleLastPrismHoldout.cs
Jofairden: No, not really. I get the idea, but the only way I see someone wanting to null the field is if they do something really custom. Otherwise, there is no reason not to follow vanilla convention, which is doing TurnToAir(). It's better to teach modders the correct habit rather than checking for every possible modder/user error everywhere, that way you may not even find certain errors anymore.

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,370 @@
...
+			float visualBeamLength = BeamLength - 14.5f * projectile.scale * projectile.scale;
+
+			DelegateMethods.f_1 = 1f; // f_1 is an unnamed decompiled variable whose function is unknown. Leave it at 1.
+			Vector2 beamStartPos = centerFloored - Main.screenPosition;
+			Vector2 beamEndPos = beamStartPos + projectile.velocity * visualBeamLength;
+			Utils.LaserLineFraming llf = new Utils.LaserLineFraming(DelegateMethods.RainbowLaserDraw);
+
+			// Draw the outer beam.
+			// c_1 is an unnamed decompiled variable which is the render color of the beam drawn by DelegateMethods.RainbowLaserDraw.
+			Color outerBeamColor = GetBeamColor();
```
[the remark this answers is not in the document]
Ozzatron: Whoops. That function was supposed to be renamed to `GetOuterBeamColor` when I added `GetInnerBeamColor`. Originally the latter function did not exist and its invocation was simply a color literal.

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
```diff
@@ -0,0 +1,370 @@
...
+				return true;
+			}
+			
+			// Otherwise, perform an AABB line collision check to check the whole beam.
+			float _ = float.NaN;
+			Vector2 beamEndPos = projectile.Center + projectile.velocity * BeamLength;
+			return Collision.CheckAABBvLineCollision(targetHitbox.TopLeft(), targetHitbox.Size(), projectile.Center, beamEndPos, BeamHitboxCollisionWidth * projectile.scale, ref _);
+		}
+
+		public override bool PreDraw(SpriteBatch spriteBatch, Color lightColor)
```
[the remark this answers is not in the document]
Ozzatron: Drawing each beam is only three lines. Most of the function is setup, so I instead wrote a `DrawBeam` function which is reused.

ExampleMod/Projectiles/ExampleLastPrismBeam.cs
Ozzatron: `DustID` does not contain a nice name for dust 15. In fact, most dusts aren't included.

PR #807, opened 2020-03-06
Doublejump

direwolf420:
### Description of the Change
Adds a way to add custom double jumps to the game, with respect to vanilla double jumps and other mods' double jumps so they don't clash. Therefore a `ModDoubleJump` class was created that is tied to the `Player` in the same way a `ModPlayer` is, aswell as a `DoubleJumpHooks` class executing the necessary logic for the jumps, guaranteeing mod compatibility. For this, most of the patching was done in the `Player.JumpMovement`. The rest of the patches are spread across various places: `ResetEffects`, `CarpetMovement`, `GrappleMovement`, `Update` and `DoubleJumpVisuals`.

### Alternate designs
* Not make a separate class for double jumps but rather have everything be `ModPlayer` hooks instead, called during the proper timings outlined above.
* Jump timing is set after all vanilla double jumps have completed, can do it another way (see **Future improvements**).

Double Jumps are a basically a hardcoded thing in vanilla, making it expandable via tml provides great benefits for modders wanting to add their own double jumps instead of falling back to vanilla ones.

### Benefits
* Enabled like you would any other effect (`player.doubleJumpCloud = true;` => `player.EnableDoubleJump<MyDoubleJump>();`
* Customize what happens when the jump occurs and while the jump key is pressed, aswell as allowing the same jump to trigger again the next time (Useful for configurable amount double jumps without creating a new class each time)
* Mod compatibility by executing each jump one after another

### Possible drawbacks
System may be too limiting for some modders, a more "free" approach like explained in Alternate Designs perhaps better.

### Applicable Issues
None yet

### Sample Usage
An accessory and two (four really) `ModDoubleJump`s provided, the one used for the accessory being slightly advanced, using inheritance, and another simple one, unused though.
Video of that in action: [Video](https://cdn.discordapp.com/attachments/103110554649894912/685464342107390004/doublejumps.webm)

### Future improvements
* A way for a `ModDoubleJump` to register the timing when to execute (before/after vanilla).
* A `GlobalDoubleJump` class hooking into each double jump, even vanilla

### Notes for reviewers
I left some comments in from my coding noting my observations/decisionmaking, maybe remove them if and before merging.


ExampleMod/Items/Accessories/ExampleDoubleJump.cs
```diff
@@ -0,0 +1,65 @@
...
+	/// A simpler ModDoubleJump is <see cref="SimpleDoubleJump"/>
+	/// </summary>
+	public abstract class ExampleDoubleJump : ModDoubleJump
+	{
+		// Optional field we use in our inheritance below, to not override the same hooks again just to adjust one number.
+		// It represents the jump height increase and the amount of dust we spawn
+		public abstract int Power { get; }
+
+		public override float Jump(ref bool playSound) {
+			CombatText.NewText(player.getRect(), Color.Lime, "'" + Power + "!'", true);
```
Jofairden: What's this for?

ExampleMod/Items/Accessories/ExampleDoubleJumpAccessory.cs
```diff
@@ -0,0 +1,41 @@
...
+			item.width = 20;
+			item.height = 26;
+			item.accessory = true;
+			item.value = Item.sellPrice(gold: 1);
+			item.rare = ItemRarityID.Orange;
+		}
+
+		public override void UpdateAccessory(Player player, bool hideVisual) {
+			// Jumps are executed in alphabetic order
+			player.EnableDoubleJump<ExampleDoubleJump1>();
```
Jofairden: Cool!

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
```diff
@@ -0,0 +1,195 @@
...
+
+			public HookList(MethodInfo method)
+			{
+				this.method = method;
+			}
+		}
+
+		private static List<HookList> hooks = new List<HookList>();
+
+		private static HookList AddHook<F>(Expression<Func<ModDoubleJump, F>> func)
```
Jofairden: I think we prefer a concrete delegate here

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
```diff
@@ -0,0 +1,195 @@
...
+
+		/// <summary>
+		/// Do the jump if a valid <see cref="nextJump"/> is found in SetNextJump
+		/// </summary>
+		/// <param name="player">Jumping player</param>
+		/// <param name="playSound">if left on true, will play the default double jump sound</param>
+		/// <returns>Jump height multiplier</returns>
+		internal static float Jump(Player player, ref bool playSound)
+		{
+			if (nextJump != -1) {
```
Jofairden: Yeah not sure if I like this design. It makes more sense to just store an `activeJump` with ModJump type, and have the changing logic happen during keypress switches/changes. Then in the rest of the code, you just call `activeJump.logic()` etc. Reduces 'magic numbers' happening

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
```diff
@@ -0,0 +1,195 @@
...
+			}
+		}
+
+		/// <summary>
+		/// Returns true if atleast one double jump is able to be used
+		/// </summary>
+		/// <param name="player">Jumping player</param>
+		internal static bool AnyJumpAgain(Player player)
+		{
+			foreach (ModDoubleJump doubleJump in player.modDoubleJumps) {
```
Jofairden: This should be simplified to `return player.modDoubleJumps.Any(jump => jump.again);`

patches/tModLoader/Terraria.ModLoader/ModDoubleJump.cs
```diff
@@ -0,0 +1,97 @@
+using System;
+
+namespace Terraria.ModLoader
+{
+	/// <summary>
+	/// A ModDoubleJump instance represents an addition to a Player instance that handles jumping logic. A ModDoubleJump is tied to the player instance, use <see cref="Player.EnableDoubleJump"/> to enable a jump like you would enable any effect through an accessory or armor
+	/// </summary>
+	public class ModDoubleJump
```
Jofairden: Yeah not sure if it's a good idea to do this in a separate class, but it seems like a big enough feature for it to be desired. My biggest concern is that it may be confusing to modders as nothing else is really done outside the current ModPlayer class, but since we want to redesign going forward this may not necessarily be an issue.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3683,6 +3691,93 @@
...
+ 							flag6 = true;
+ 						}
+-						else
++						else if (this.jumpAgainCloud)
+ 						{
+ 							this.jumpAgainCloud = false;
++							cloud = true;
++						}
++						// Mod jumps are AFTER vanilla jumps here, maybe modder wants to specify before/after?
++						else if(DoubleJumpHooks.SetNextJump(this))
```
Jofairden: Before/after may be necessary. It could be as simple as `myJump.beforeVanilla = true`

ExampleMod/Items/Accessories/ExampleDoubleJump.cs
JavidPack: I think this is just to demonstrate to the modder that it is working

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
direwolf420: this is the exact same syntax as in ModPlayer https://github.com/tModLoader/tModLoader/blob/master/patches/tModLoader/Terraria.ModLoader/PlayerHooks.cs#L34

patches/tModLoader/Terraria.ModLoader/ModDoubleJump.cs
direwolf420: The reasoning behind a separate class is to move all the "control" that `DoubleJumpHooks` does (toggling states of jumping and effects at the appropriate timings which are hardcoded into vanilla) away from the modder so he doesn't have to worry about the specifics. It would take more than 6 `ModPlayer` hooks in proper locations to achieve the same (and they would be redundant for anything other than double jumping), and then the modder would have to handle all the vanilla + mod compatibility himself

patches/tModLoader/Terraria/Player.cs.patch
direwolf420: Should this happen via a virtual property that is evaluated when `AddDoubleJump` is called? Or dynamic (modder can change it any time)?

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
direwolf420: Not sure I'm following, please check the latest commit

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
direwolf420: This has been removed in the latest commit

ExampleMod/Items/Accessories/ExampleDoubleJump.cs
direwolf420: Yeah it's for representing which of the three jumps is currently triggered ingame

direwolf420:
Added a new hook to customize horizontal movement during the jump, I somehow missed this in the initial PR

patches/tModLoader/Terraria.ModLoader/DoubleJumpHooks.cs
Jofairden: Yup. See how much simpler it is and how much code you're able to throw away? :-)

direwolf420:
I will either continue or redo this PR after tml starts working on 1.4.
Plans: Rename ModDoubleJump to ModExtraJump, and add a before/after vanilla jumps property

PR #810, opened 2020-03-09
Clouds

aberna01:
### Description of the Change
Made an addition to give the player some options to mess with the clouds color-wise. Also converts the cloud's data into a drawdata format.

### Alternate designs
First tried three files, then two, I didn't want to make a nuscence for some cloud methods so I tried to keep them in the most reasonable place I could find.

For the same reasons listed below these questions feel like the same thing.

### Benefits
May give an insight on custom weather events or ideas. I always thought the background really set the mood for terraria, at least it did for me. Even the clouds which we sometimes never think about. Clouds do so much to keep the world going, and never ask anything in return, they just come and go and are often forgotten only to be seen again. Well that's in real life anyways, in terraria clouds are pngs that move slowly across the screen, but surely there can be a reason for them.

### Possible drawbacks
The only thing I can come up with is reliance on the vanilla code so when a new patch hits it's likely gonna break too, or be another thing to patch.

### Applicable Issues
I was running out of ideas/time/patience/focus trying to make SOMETHING that could at least be a tiny part of TML since nothing else I've ever made was good enough.

### Sample Usage
I'm sorry I refuse to do this. The only time I've been aware of people even looking at something I made was to point out the flaws or the bad parts about it. If I can't even give one thing to a community I've been trying to give to for a long time them I'm going to have to stop programming and just be forgotten because the internet will go on without me.


JavidPack:
Looks like you made a PR from our ModConfig branch to our Master branch, rather than your Fork of our Master branch

aberna01:
My bad, but when I try to do pull requests from my end it just says nothing
different and gives me no options.

Jofairden:
image
You need to set base to master if you want to merge into master.
Comparing across forks is the same thing, so compare master against your fork/branch

Jofairden:
See https://github.com/tModLoader/tModLoader/pull/811.

PR #811, opened 2020-03-09
Added ExampleCloud

ghost:
### Description of the Change
Made an addition to give the player some options to mess with the clouds color-wise. Also converts the cloud's data into a drawdata format.

### Alternate designs
First tried three files, then two, I didn't want to make a nuscence for some cloud methods so I tried to keep them in the most reasonable place I could find.

For the same reasons listed below these questions feel like the same thing.

### Benefits
May give an insight on custom weather events or ideas. I always thought the background really set the mood for terraria, at least it did for me. Even the clouds which we sometimes never think about. Clouds do so much to keep the world going, and never ask anything in return, they just come and go and are often forgotten only to be seen again. Well that's in real life anyways, in terraria clouds are pngs that move slowly across the screen, but surely there can be a reason for them.

### Possible drawbacks
The only thing I can come up with is reliance on the vanilla code so when a new patch hits it's likely gonna break too, or be another thing to patch.

### Applicable Issues
I was running out of ideas/time/patience/focus trying to make SOMETHING that could at least be a tiny part of TML since nothing else I've ever made was good enough.

### Sample Usage
I'm sorry I refuse to do this. The only time I've been aware of people even looking at something I made was to point out the flaws or the bad parts about it. If I can't even give one thing to a community I've been trying to give to for a long time them I'm going to have to stop programming and just be forgotten because the internet will go on without me.

### DIFF
--- a/ExampleMod/Backgrounds/ExampleCloud.cs
+++ b/ExampleMod/Backgrounds/ExampleCloud.cs
@@ -1,0 +1,59 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using Terraria;
+using Terraria.DataStructures;
+using Terraria.Utilities;
+using static Terraria.Cloud;
+using static Terraria.WorldGen;
+
+namespace ExampleMod
+{
+	class ExampleCloud
+	{
+		public static UnifiedRandom Rand = new UnifiedRandom();
+
+		/// <summary>
+		/// As the name implies, clears all the draw data from any currently active clouds.
+		/// </summary>
+		public static void Reset()
+			=> resetClouds();
+
+		/// <summary>
+		/// This encapsulates all the data needed to draw the clouds.
+		/// </summary>
+		/// <returns>A DrawData object that can be modified or drawn immediately.</returns>
+		public static DrawData CloudToDrawData(Cloud C)
+			=> new DrawData(Main.cloudTexture[C.type], C.position, null, C.cloudColor(Main.bgColor), C.rotation, Vector2.Zero, C.scale, C.spriteDir, 1);
+
+		/// <summary>
+		/// A simple command to let you add a bunch of vanilla clouds.
+		/// </summary>
+		/// <param name="Amount"></param>
+		public static void AddClouds(int Amount)
+		{
+			for(int Index = 0; Index < Amount; Index++)
+			{
+				addCloud();
+			}
+		}
+
+
+		
+
+		/////  BONUS INFORMATION	//////////
+		///
+		///
+		///   .Clouds are stored like entities and other objects in terraria in an array.
+		///   
+		///   .The vanilla limit for clouds in terraria is 200.
+		///   .They have bool fields that when marked false can have their value overwritten.
+		///   .There is also a field bool for a flag of whether or not the cloud is going to be removed this update cycle.
+		///   
+		///   .These concepts are important if you go further in video game design as it's the basic system of iterating
+		///      through limited objects to conserve resources.
+		///
+		///   .That being said, lets make our own clouds.
+		///   .Wind Speed is like the cloud's velocity for an entity.
+
+	}
+}
--- a/ExampleMod/Backgrounds/ExampleCloudConfig.cs
+++ b/ExampleMod/Backgrounds/ExampleCloudConfig.cs
@@ -1,0 +1,54 @@
+﻿using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.ModLoader.Config;
+
+namespace ExampleMod
+{
+	class ExampleCloudConfig : ModConfig
+	{
+		public override ConfigScope Mode => ConfigScope.ClientSide;
+
+
+		[Label("Click this to enable custom wind speeds.")]
+		public bool UseCustomWindSpeed;
+
+
+		[Range(-1f, 1f)]
+		[Increment(0.1f)]
+		[Label("Use this to change the directiono the clouds in the sky are moving and the rate at which they move.")]
+		public static float SetWindSpeed;
+
+		[Label("Enable Clouds to take on custom color values below")]
+		public static bool UseCustomCloudColor;
+
+		[Range(0f, 255f)]
+		[Increment(1f)]
+		[Label("Shift the red value of the clouds.")]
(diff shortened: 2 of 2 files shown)


ExampleMod/Backgrounds/ExampleCloud.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using Terraria;
+using Terraria.DataStructures;
+using Terraria.Utilities;
+using static Terraria.Cloud;
```
GoodPro712, line 6: using `using static`s makes the code quite confusing

ExampleMod/Backgrounds/ExampleCloud.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using Terraria;
+using Terraria.DataStructures;
+using Terraria.Utilities;
+using static Terraria.Cloud;
+using static Terraria.WorldGen;
+
+namespace ExampleMod
+{
+	class ExampleCloud
```
GoodPro712, line 11: use `public class` for consistency reasons.

ExampleMod/Backgrounds/ExampleCloudConfig.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.ModLoader.Config;
+
+namespace ExampleMod
+{
+	class ExampleCloudConfig : ModConfig
```
GoodPro712, line 7: use `public class` for consistency reasons

ExampleMod/Backgrounds/ExampleCloudConfig.cs
```diff
@@ -0,0 +1,54 @@
...
+		public static void SetAllCloudColor()
+		{
+			if (UseCustomCloudColor) {
+				for (int Index = 0; Index < Main.cloudLimit; Index++) {
+					Main.cloud[Index].cloudColor(new Color(RedHueValue, GreenHueValue, BlueHueValue, CloudOpacity));
+				}
+			}
+		}
+	}
+
```
GoodPro712, line 53: remove random newline

ExampleMod/Backgrounds/ExampleCloudConfig.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		[Label("Click this to enable custom wind speeds.")]
+		public bool UseCustomWindSpeed;
+
+
+		[Range(-1f, 1f)]
+		[Increment(0.1f)]
+		[Label("Use this to change the directiono the clouds in the sky are moving and the rate at which they move.")]
+		public static float SetWindSpeed;
```
GoodPro712, line 19: Configs shouldn't be static

ExampleMod/Backgrounds/ExampleCloud.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using Terraria;
+using Terraria.DataStructures;
+using Terraria.Utilities;
+using static Terraria.Cloud;
+using static Terraria.WorldGen;
```
GoodPro712, line 7: `using static` makes the code quite confusing

ExampleMod/Backgrounds/ExampleCloud.cs
GoodPro712, line 11: Use `public class` for consistency reasons

ExampleMod/Backgrounds/ExampleCloudConfig.cs
GoodPro712, line 7: Use `public class` for consistency reasons

Metacinnabar (changes requested):

ExampleMod/Backgrounds/ExampleCloud.cs
Jofairden: I do agree, in some scenarios it's fine though. This file is so small it really doesn't matter.

ExampleMod/Backgrounds/ExampleCloud.cs
```diff
@@ -0,0 +1,59 @@
...
+using Terraria.DataStructures;
+using Terraria.Utilities;
+using static Terraria.Cloud;
+using static Terraria.WorldGen;
+
+namespace ExampleMod
+{
+	class ExampleCloud
+	{
+		public static UnifiedRandom Rand = new UnifiedRandom();
```
Jofairden, line 13: Any reason you're making a new random instead of using an existing one? (eg. Main.rand)

ExampleMod/Backgrounds/ExampleCloud.cs
```diff
@@ -0,0 +1,59 @@
...
+		///   
+		///   .The vanilla limit for clouds in terraria is 200.
+		///   .They have bool fields that when marked false can have their value overwritten.
+		///   .There is also a field bool for a flag of whether or not the cloud is going to be removed this update cycle.
+		///   
+		///   .These concepts are important if you go further in video game design as it's the basic system of iterating
+		///      through limited objects to conserve resources.
+		///
+		///   .That being said, lets make our own clouds.
+		///   .Wind Speed is like the cloud's velocity for an entity.
```
Jofairden, line 56: Please put this above the class definition.

Jofairden (changes requested):

ExampleMod/Backgrounds/ExampleCloud.cs
aberna01: cause I was stupid at the time and didn't know how it worked cause I was always making my own, I know now, I honestly made this a little fast and should've worked on it more than I did, sorry. I'll try to get back to it some time

ExampleMod/Backgrounds/ExampleCloud.cs
aberna01: for consistency with example mod I'm fine with no using statics, because it's not a necessary thing to the function of the code and something that was introduced in like version 6 I believe, and a lot of programmers don't really use it.

PR #819, opened 2020-03-31
Example armor glowmask

direwolf420:
### Description of the Change
Easy to understand and follow PlayerLayer example, since this is often an issue with newer modders who want to get into PlayerLayers, as it is often requested on the discord.

### Alternate designs
Could've made the glowmask look better, but I'm not a spriter.

N/A

### Benefits
Easy to understand and self-learn PlayerLayers based on an example. Also explains what to do if it's not a breastplate glowmask.

### Possible drawbacks
N/A

### Applicable Issues
N/A

### Sample Usage
N/A


ExampleMod/ExamplePlayer.cs
```diff
@@ -634,6 +635,103 @@ public class ExamplePlayer : ModPlayer
...
+			// To figure out when we should draw our ExampleBreastplate glowmask, we make use of GetEquipSlot.
+			// It takes in the name of the item that comes with an equip texture, and then its EquipType.
+			// By comparing that to drawPlayer.body, which represents the currently visible body armor, we make sure it's safe to draw ours
+			// Here we abort if the currently equipped body armor isn't what we need for our glowmask
+			if (drawPlayer.body != mod.GetEquipSlot("ExampleBreastplate", EquipType.Body)) {
+				return;
+			}
+
+			// Because we have a breastplate glowmask, and breastplates have different textures depending on if the player is male or female, we have to use two textures here
+			string gender = "";
```
Jofairden, line 663: Personally prefer:
`Texture2D texture = drawPlayer.Male ? mod.GetTexture("Items/Armor/ExampleBreastplate_Body_Glowmask") : mod.GetTexture("Items/Armor/ExampleBreastplate_Female_Body_Glowmask");`
You don't use gender elsewhere, so with this it's easier and faster to see the effect of the gender

ExampleMod/ExamplePlayer.cs
```diff
@@ -634,6 +635,103 @@ public class ExamplePlayer : ModPlayer
...
+			// The position this texture will be drawn at on the screen
+			// Because we are drawing a breastplate and it has the same alignment as the players' body, we use bodyPosition
+			Vector2 position = new Vector2(drawX, drawY) + drawPlayer.bodyPosition - Main.screenPosition;
+
+			// The alpha of the texture when the player is blinking (immune)
+			// Should stay like that
+			float alpha = (255 - drawPlayer.immuneAlpha) / 255f;
+
+			// The color of the glowmask, pick what you want
+			Color color = Color.White;
```
Jofairden, line 690: Not 'pick what you want' , most/all glowmasks will use a fully opaque white color because that's what glowmasks are supposed to do: they should be drawn at full brightness regular color regardless of light. It may be beneficial to put this kind of information in the comments.

ExampleMod/ExamplePlayer.cs
```diff
@@ -634,6 +635,103 @@ public class ExamplePlayer : ModPlayer
...
+			// Should stay like that
+			SpriteEffects spriteEffects = drawInfo.spriteEffects;
+
+			DrawData drawData = new DrawData(texture, position, frame, color * alpha, rotation, origin, 1f, spriteEffects, 0);
+
+			// The shader that's applied to the texture
+			// Depending on what EquipType this glowmask is representing, you need to assign the proper shader to it
+			// (This references the dye)
+			// Because this is a breastplate, we want to use the dye that is in the breastplate slot
+			drawData.shader = drawInfo.bodyArmorShader;
```
Jofairden, line 710: Good to note you can also apply different shaders if you want, giving special glowmask effects. Sky is the limit.

ExampleMod/ExamplePlayer.cs
```diff
@@ -634,6 +635,103 @@ public class ExamplePlayer : ModPlayer
...
+			// Here we generate visual dust while our glowmask is visible
+			if (Main.rand.NextBool(10)) {
+				int index = Dust.NewDust(drawPlayer.position, drawPlayer.width, drawPlayer.height, 135);
+				Dust dust = Main.dust[index];
+				dust.noGravity = true;
+				dust.noLight = true;
+				dust.fadeIn = Main.rand.NextFloat(0.5f, 0.8f);
+
+				// Same deal as before with drawData.shader
+				dust.shader = GameShaders.Armor.GetSecondaryShader(drawInfo.bodyArmorShader, drawPlayer);
```
Jofairden, line 724: Very cool, please add: `// Gives the dust the same shader effect that is equipped for the armor`

Jofairden (approved):
Great job!

ExampleMod/ExamplePlayer.cs
```diff
@@ -650,25 +650,25 @@ public class ExamplePlayer : ModPlayer
...
 				return;
 			}
 
+			// The texture we want to display on our player
+			Texture2D texture;
 			// Because we have a breastplate glowmask, and breastplates have different textures depending on if the player is male or female, we have to use two textures here
-			string gender = "";
-			if (!drawPlayer.Male) {
-				gender = "Female";
+			if (drawPlayer.Male) {
```
Jofairden, line 664: Still we really prefer the elvis operator for assignments: `texture = drawplayerMale ? maleoption : femaleoption`

Jofairden (changes requested):

direwolf420:
I was under the impression that ternary operator is not very newbie friendly since this is a very basic example/walkthrough. I very rarely see them in Examplemod too.

ExampleMod/ExamplePlayer.cs
direwolf420: I was under the impression that ternary operator is not very newbie friendly since this is a very basic example/walkthrough. I very rarely see them in Examplemod too.

direwolf420: Plus with the length of the single line you would have to line break after the "?" which is always confusing for people new to code (the fact that you can break the same line/line without a ";" at the end)

Jofairden: I think using the ternary operator is fine here because Player Layers are already decently advanced topics, I don't think newbies should be attempting this.

Jofairden (approved):

Jofairden (changes requested):
image
It's not working for the other arm ^Can you fix this? Then I'll merge.

direwolf420:
I know this is bad from a programming perspective (reusing code), but I've done it like this to not confuse beginners with more complexity by adding some method that returns all the relevant drawdata besides the texture. So by itself, the ExampleBreastplateBodyGlowmask fulfills a "showcase/walkthrough" type purpose, while ExampleBreastplateArmsGlowmask is optional, and doesn't contain any comments.

The reason those are two separate glowmasks is because of how terraria draws body and arms on separate layers. Wanting to simply reuse the local variables in ExampleBreastplateBodyGlowmask would lead to me changing where the layer is inserted (after PlayerLayer.Arms, not after PlayerLayer.Body), but it would make the code very compact and in just one layer

Solxanich:
@direwolf420 
Mini- review:
Update to target 1.4. If you plan on changing how how glowmasks are done in tML side and doing EM in that change, consider repurposing this PR or closing it.

direwolf420:
Needs port to 1.4, which still requires draw layers and some publicization for legs. Body does not require a draw layer but requires code to set the color

Issue #804, opened 2020-03-01
Idea -Try Reloading Again/Try Compiling Again button


aberna01:
On the mod sources screen when you get an error, it would be nice to have a button that immediately try to load/compile again, would save 1-2 clicks

Jofairden:
Closed because implemented in #822

PR #822, opened 2020-04-05
Add retry build button, fix tiny error in compile failed message

Trivaxy:
### What is the new feature?
This PR adds a new button that allows the modder to retry building and loading their mod. It only appears if the error is that of a build failure, and it doesn't appear otherwise (e.g due to mod load failure). Useful for saving 1-2 clicks and implements the suggestion in #804 
(This PR also fixes a very small issue with the CompileError message. The bug was that the string was formatted with the error message in place of the number of warnings. I feel like the fix was too small to warrant a separate PR 😛 )

A small quality-of-life addition to make mod dev lives a little easier. It's also helpful in the _extreme_ case a modder's mod sources list takes a long time to repopulate (for whatever odd reason).

Perhaps separate the button into two, where one retries the build and the other retries the build and reloads as well, or it can stay as one button that builds and infers whether to reload or not based on what the modder clicked (`Build` or `Build + Reload` in the mod sources UI).
Optionally the retry button can be extended to retry mod loading alone, but often this yields the same result and it's better if the root issue that causes mod loading failure gets fixed by the mod dev.
The only other point is to probably add localization for `Retry`.

https://streamable.com/ude1mn

### ExampleMod updates
No changes to ExampleMod.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
@@ -591,7 +591,7 @@ namespace Terraria.ModLoader.Core
 
 			if (results.HasErrors) {
 				var firstError = results.Cast<CompilerError>().First(e => !e.IsWarning);
-				throw new BuildException(Language.GetTextValue("tModLoader.CompileError", Path.GetFileName(outputPath), numErrors, firstError));
+				throw new BuildException(Language.GetTextValue("tModLoader.CompileError", Path.GetFileName(outputPath), numErrors, numWarnings) + $"\nError: {firstError}");
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Looks pretty good to go. I'll just have to double check something at my computer

Chicken-Bones:
Super on board with this but lets make it a little less specific.  Change `Show` to accept an optional `retryAction`

We should also change `showRetry` to `continueIsRetry` to avoid confusion.

Trivaxy:
Thanks for the feedback. Made the changes (hopefully this is what you meant), and tested again - everything works normally. If there's anything wrong please let me know 😄 

(p.s to clear up any confusion, the old `showRetry` that was used for the continue button got renamed to `continueIsRetry`, and the new `showRetry` has been repurposed to only serve as an indicator as whether to show the new Retry button or not)

Jofairden:
Nice, I'll let someone else do the final merge. Thanks 👍

Issue #613, opened 2019-07-16
Hover over unloaded Tiles for info


JavidPack:
A suggestion has been made to allow for hovering over the pink unloaded tiles to view the source mod and tile name. This could be useful.

Jofairden:
Looks like the mystery tile saving/loading will need to be changed to make this happen, as the current format does not support this feature.

GoodPro712:
Contributions Welcome tag?

Jofairden:
Solved by #824. Closed.

PR #824, opened 2020-04-08
Added hover text to MysteryTile

Rartrin:
### What is the new feature?
Mousing over a mystery tile will now show the mod and tile name of the unloaded tile.
From #613 

Helpful for knowing what specific mod an unloaded tile was from.

No. It is intended to be a rather simple, basic, and functional. As it stands, it requires no changes to the base game.

Somewhat niche but still, having a way to discern between multiple unloaded tiles that otherwise look completely identical could be useful at times.

### ExampleMod updates
None


JavidPack:
Thanks

Issue #812, opened 2020-03-09
World Generation should crash back to World selection


Plantterror:
### Description
When a world is generated and fails to generate (puts "an error occurred during world generation" in logs) currently the world bar will be frozen in place and a restart of tmodloader is entirely needed. I propose that tmodloader should exit back to the world selection screen (with an error message, possibly) to resolve some confusion.

### Log File
Not Applicable

### Steps to Reproduce

1. Generate a world with a problematic mod (currently reliable method would be generating Obsidian biome from Enigma with a world greater than large)
2. Watch as the bar gets stuck at "Obsidification" forever

### Expected behavior/Actual behavior
Not Applicable.

### Reproduction frequency
Not Applicable yet again.

### Additional Information
Most of the notable world generation errors for common worlds seem to have been fixed, so most of the time this will mostly affect those on FNA/Mac/Linux.
Also paging @DradonHunter11 for this.

Plantterror:
Wow. I can't believe I typed all that on mobile and actually missed my main point. Forgive my stupidity.

PR #825, opened 2020-04-09
Added a menu for when world gen crashes.

Rartrin:
### What is the new feature?
A menu for when world gen crashes which shows you the name of the world gen pass that caused the issue and the exception message. Also includes both an "Open Logs" and a "Back to Main Menu" button. (See #812 )

1. So that the user knows if whether it is simply world gen taking a long time or if it actually crashed.
2. Generally, even when an exception was thrown during world gen, the game is still completely playable. Some larger mod packs have extremely long load times (10-15 minutes) and much higher chances of crashing worldgen, so having a way to restart world gen without restarting the game, can save a lot of time, 

My original implementation involved changing the vanilla world loading screen but I changed it to be its own separate screen which means less changes to vanilla. YAY!

Trying to play with a mod pack that takes 10 minutes to load but only 10 seconds to get softlocked because a world gen pass threw an IndexOutOfRangeException.

### ExampleMod updates
N/A


Rartrin:
This mod will make it crash on pass 15, which is named "CRASH" and throw an exception with the message "Crashed".
[Crash on Pass 15.zip](https://github.com/tModLoader/tModLoader/files/4453756/Crash.on.Pass.15.zip)

Chicken-Bones:
We should consider forcing mods to call a 'keep alive' method in long running passes, allowing the user to kill worldgen if it doesn't look like progress is being made.

If a pass doesn't call the keepalive once per 100ms in dev mode, we can crash the game with a log message and popup

Rartrin:
For me it is always softlocks due to something throwing an exception like a mod not finding a tile it is looking for and it winds up going outside of an array.
That said, I could try to look into it later, though it is somewhat of a separate issue from what I was doing here. Probably the best way to do it would be make the keep alive be any update to the Progress instance, that way it is somewhat backwards compatible with some mods.
Maybe after 5-10 seconds of no progress updates, pop up a message on the screen saying "mod doesn't seem to be responding" it can show an abort/stop button which take them to the crash screen, otherwise, if an update happens or the pass finishes, it hides the message and button. This way, if a mod actually does work but just doesn't update the progress bar, it will still be usable, while also providing an easy way out if it is actually softlocked. Also, we just tell modders to make sure they update the progress bar, they don't need to use or implement anything new.

patches/tModLoader/Terraria.World.Generation/WorldGenerator.cs.patch
```diff
@@ -13,4 +13,28 @@
...
++				catch(Exception e)
++				{
++					string message = string.Join
++					(
++						"\n",
++						Localization.Language.GetTextValue("tModLoader.WorldGenError"),
++						current2.Name,
++						e.Message
++					);
++					Main.MenuUI.SetState(new ModLoader.UI.UIWorldLoadCrash(message));
```
Chicken-Bones, line 34: Is this threadsafe? I do much prefer constructing UI's each time they're used, but it's not the Terraria paradigm.

Chicken-Bones:
Is there a reason you can't use UIErrorMessage?

patches/tModLoader/Terraria.World.Generation/WorldGenerator.cs.patch
Jofairden: Shouldn't it be setting Main.menuMode to a custom ID? That is how we do this for all menus. Check Interface.cs
( I do want to rewrite this, still )

Rartrin:
Okay, I brought it up to date with the master and converted it over to use UIErrorMessage. I also noticed that the Open Logs button wasn't working. I found the issue and included it. It is a one line fix is all.

Issue #815, opened 2020-03-17
Disable All button in Mods menu disables all mods, even when they are not enabled.


Plantterror:
### Description
Mods that aren't enabled will still get disabled when pressing the "Disable All" button in the Mods menu. This can end up causing a lag spike if you have more than 100 mods in the folder.

### Log File
https://cdn.discordapp.com/attachments/534215632795729922/689539123954188512/client.log
(Example log file to showcase)

### Steps to Reproduce

1. Go into the mods menu while some mods are not enabled.
2. Press Disable All 
3. Check the log file.

### Expected behavior 
Mods that are enabled should only be disabled.

### Actual behavior
All mods are being force disabled.

### Reproduction frequency
Guaranteed.

### Additional Information
N/A for now

Jofairden:
What kind of lag spike are you experiencing specifically? That _should_ not happen

Plantterror:
It was unresponsive for about the time it took to disable every mod.

Chicken-Bones:
Do you still get the lag spike with no mods enabled?
Can you show us a profiling report?

PR #826, opened 2020-04-09
Fix enable/disable mods lag/freeze with many mods

Rartrin:
### What is the bug?
From #815 .
When there are a large number of mods installed, using Enable All or Disable All will always cause a lag spike, even if only a few mods would change. This is because it allows it to Enable/Disable a mod which has already been Enabled/Disabled respectively. When it is enabled, it tells it to save the set of enabled mods which it has to do for every mod.

### How did you fix the bug?
Added a quick check to return when told to enable/disable while already in that state.
Also added a property to pause saving EnabledMods list. Once it is unpaused, if any changes were made, it will automatically save the new list. With this, I implemented it around all cases where multiple mods would be enabled/disabled in a loop (like for mod dependencies and for the enable/disable all buttons).

Not likely without some significant reworking of how the enabled mods list is handled.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.UI/UIModItem.cs
+++ b/patches/tModLoader/Terraria.ModLoader.UI/UIModItem.cs
@@ -247,18 +247,20 @@ namespace Terraria.ModLoader.UI
 		}
 
 		private void ToggleEnabled(UIMouseEvent evt, UIElement listeningElement) {
-			Main.PlaySound(12);
+			Main.PlaySound(SoundID.MenuTick);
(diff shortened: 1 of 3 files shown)


Rartrin:
Oh also, I noticed a few sounds next where I was making changes were using hardcoded values so I swapped them to SoundIDs while I was there.

Rartrin:
Hold on, never mind, I found why it is lagging. It is saving enabled.json after each mod is enabled/disable rather than all at once. That is bad.

PR #830, opened 2020-04-15
fix whoami/active inconsistency with entities on unload

direwolf420:
### What is the bug?
Due to the order of things, so called "dummy entities" have empty arrays that should contain instances of modded classes. In the case of NPCs, that entry is also active, which combined with bad code from some mods, causes IndexOutOfBounds exceptions.

### How did you fix the bug?
What this PR basically does is bring `ModContent.CleanupModReferences` in line with `Main.Initialize`, so there are no "missed spots".
A similar bug/fix was done a while ago for projectiles (and noticed also by me): https://github.com/tModLoader/tModLoader/commit/acec154a75e8b3bc9410c2e664eaff2268aa0aa0#diff-8fbc78a2f416d0b475dd3624dff42fbbL555
This PR is done in the same way. 
Also, while investigating `whoAmI` for each class, I noticed that `item.whoAmI` is never written to nor read in vanilla code, which leads to the line
```cs
Main.item[index].whoAmI = index;
```
in `Player.QuickSpawnClonedItem` (which is a method added by tml) to be obsolete. I have commented it but not removed it, so you can decide to keep it or not.

Dunno, this is mostly mirroring `Main.Initialize`.

### Backstory
Terraria uses dummy entities in the last entry of any `Entity[]` (Such as `Main.npc`). The process goes like this:
`Main.Initialize` creates new entity objects and assigns them to each element in the array
Then, ModLoader loads, and then calls `ModContent.CleanupModReferences` which clears the arrays (this is there the bug is). It does it by excluding the dummy entity for NPCs and Items, and calling SetDefaults(0), which sets all the affected entities' active state to true.
Then, when the player joins the world, `Worldgen.clearWorld` is called and creates new entity objects excluding the dummy entity.
This leads to for example the 200th (which is Main.maxNPCs) NPC being active without having any reasonable data attached to it.
Since vanilla doesn't touch that dummy during the game loop, it stays active, which causes mods with faulty code to throw exceptions (example below). Faulty because foreach should't be used for the entity arrays, but some modders aren't aware of that.
Reason (explained on NPC): during `ModContent.CleanupModReferences`, which happens after modded data is cleared from the static arrays, `npc.SetDefaults(0)` is called which should create instances of GlobalNPCs and add them to the `npc.globalNPCs` array. Since they are unloaded, this array is never filled. During the game though, when NPCs are spawned, this modded data is present and that array is filled normally. If a mod were to accidently call a method that accesses GlobalNPC data on the dummy NPC, it will throw an exception because it tries to access an array element in an array that is empty (Screenshots below):
[Empty array](https://cdn.discordapp.com/attachments/539214017172865025/699742606418247800/Screenshot_12.png)

[Accessing an instance (3 GlobalNPC instances were present)](https://cdn.discordapp.com/attachments/539214017172865025/699742608246702090/Screenshot_13.png)

Example code that will throw this exception:
```csharp
foreach(NPC npc in Main.npc) {
	if (npc.active) {
		Main.NewText(npc.GetBossHeadTextureIndex());
	}
}
```

TLDR: PR fixes 200th NPC being active


Jofairden:
Great detailed report, but I'm gonna ask one of the others to have a thorough look at the underlying cause and if the fix is correct.

Chicken-Bones (approved):

PR #834, opened 2020-04-20
Sparkling ExampleHerb code

GoodPro712:
Cleaned up ExampleHerb quite a bit. The original code looked really ugly so I tried to make it better

### DIFF
--- a/ExampleMod/Tiles/ExampleHerb.cs
+++ b/ExampleMod/Tiles/ExampleHerb.cs
@@ -21,7 +21,7 @@ namespace ExampleMod.Tiles
 	//Sadly, modded plants are unable to be grown by the flower boots
 	public class ExampleHerb : ModTile
 	{
-		private const byte FrameWidth = 18; //a field for readibilty and to kick out those magic numbers
+		private const int FrameWidth = 18; //a field for readibilty and to kick out those magic numbers
 
 		public override void SetDefaults()
 		{
--- a/ExampleMod/Tiles/ExampleHerb.cs
+++ b/ExampleMod/Tiles/ExampleHerb.cs
@@ -45,8 +45,9 @@ namespace ExampleMod.Tiles
 
 		public override bool Drop(int i, int j)
 		{
-			Stage stage = GetStage(i, j);
+			Stage stage = GetStage(i, j); //The current stage of the herb
 
+			//Only drop items if the herb is grown
 			if (stage == Stage.Grown)
 				Item.NewItem(new Vector2(i, j).ToWorldCoordinates(), ItemType<ExampleHerbSeeds>());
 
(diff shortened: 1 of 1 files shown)


ExampleMod/Tiles/ExampleHerb.cs
```diff
@@ -7,78 +9,82 @@
 
 namespace ExampleMod.Tiles
 {
+	//A plant with 3 stages, planted, growing and grown.
+	//Sadly, modded plants are unable to be grown by the flower boots
 	public class ExampleHerb : ModTile
 	{
-		public override void SetDefaults() {
+		private static readonly byte pixelsPerStage = 18; //a field for readibilty and to kick out those magic numbers
```
Jofairden, line 16: Just make this `const`

ExampleMod/Tiles/ExampleHerb.cs
```diff
@@ -7,78 +9,82 @@
...
+
+		//A method to quickly get the current stage of the herb
+		private Stage GetStage(int i, int j)
+		{
+			Tile tile = Framing.GetTileSafely(i, j); //Always use Framing.GetTileSafely instead of Main.tile as it prevents any errors caused from other mods
+			return (Stage)(tile.frameX / pixelsPerStage);
+		}
+
+		//An enum on the 3 stages of herb growth.
+		private enum Stage : byte
```
Jofairden, line 83: This should be above the const, preferably even outside the class as `PlantStage` when used for multiple plants.

ExampleMod/Tiles/ExampleHerb.cs
Jofairden: Also name it FrameHeight instead

ExampleMod/Tiles/ExampleHerb.cs
```diff
@@ -6,79 +8,83 @@
...
+		Growing,
+		Grown
+	}
+
+	//A plant with 3 stages, planted, growing and grown.
+	//Sadly, modded plants are unable to be grown by the flower boots
 	public class ExampleHerb : ModTile
 	{
-		public override void SetDefaults() {
+		private const byte FrameWidth = 18; //a field for readibilty and to kick out those magic numbers
```
Jofairden, line 24: Should be int, but alright

GoodPro712: What's your reason why?

Jofairden: Because it's just a number, the height of a frame. It has nothing to do with bytes.

GoodPro712: as far as I know it being a byte would reduce size

Jofairden: You're not dealing with any networking or saving data here, so that doesn't matter.

PR #848, opened 2020-04-29
Implemented new autoloading system

Rartrin:
### What is the new feature?
IAutoloadable and two base classes that most modded types derive from.

The base classes reduced down a lot of boiler plate code.
These base classes as well as IAutoloadable are also public which can allow modders to leverage tML's autoloading system in there own mods.

The one potential alternative would be a simple interface with Load and Unload. If a type has the interface, it autoloads, otherwise it won't.
However, the way I choose seemed like the most elegant approach in that in unifies all loading and also won't break existing code compiling. Plus it allows for having things which may need autoloading but not unloading.

ModLiquid

### ExampleMod updates
None.


patches/tModLoader/Terraria.ModLoader/ModTexturedType.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+
+namespace Terraria.ModLoader
+{
+	/// <summary>
+	/// The base type for most modded things with textures.
+	/// </summary>
+	public abstract class ModTexturedType:IAutoloadable
```
Jofairden, line 8: Why does this not extend ModType ?

Rartrin: It would lead to some weird method wrapping. It is very doable, but then it would have 2 AddInstance methods, one of which wouldn't do anything. If we wanted to add a base Texture property which would be assigned some time that might help... Currently the Name property doesn't autoassign either unless the AddInstance method calls a method which does it.

Actually, now with the major rework we will have for 1.4, maybe it would be a perfect time to just change the system entirely. Instead of autoload methods, what if we had autoload properties and made the ref name and texture fields just rely on overriding Name and Texture properties. If someone wants to have multiple instances of the same class with different name/textures, then they can override them with auto properties and set them in the constructor.

Jofairden: That's not a bad idea, but what are the usecases (can we come up with a good example?)

Rartrin: It is definitely more of an organizational change. In terms of functionality, it does allow for a single `ModType` class rather than two, like mentioned above.
In terms of organization though, it would help diversify the information into more meaningful locations. From a conceptual stand point, having to override Autoload in order to change the internal name or the texture doesn't make much sense. Also, if modders want to manually add a class, they wouldn't need to pass in the name or texture. Instead, they can keep this information within the class it relates to.

Jofairden: `having to override Autoload in order to change the internal name or the texture doesn't make much sense`
Yeah that's a good statement. I'm definitely in favor of the change, will ask others.

patches/tModLoader/Terraria/ModLoader/GlobalBuff.cs
```diff
@@ -7,29 +7,12 @@ namespace Terraria.ModLoader
...
-			get;
-			internal set;
-		}
-
-		/// <summary>
-		/// Allows you to automatically load a GlobalBuff instead of using Mod.AddGlobalBuff. Return true to allow autoloading; by default returns the mod's autoload property. Name is initialized to the overriding class name. Use this method to either force or stop an autoload or to control the internal name.
-		/// </summary>
-		public virtual bool Autoload(ref string name) {
-			return mod.Properties.Autoload;
+		internal sealed override void AddInstance() {
```
Chicken-Bones, line 12: Suggest changing AddInstance to Register

patches/tModLoader/Terraria/ModLoader/ModType.cs
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		public virtual void Load(){}
+
+		public virtual void Unload(){}
+
+		/// <summary>
+		/// DO NOT CALL THIS! This is called automatically in ILoadable.Load
+		/// </summary>
+		internal virtual void AddInstance(){}
```
Chicken-Bones, line 31: shouldn't this be protected?

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -418,190 +357,10 @@ internal void PrepareAssets()
...
-			ModWorld modWorld = (ModWorld)Activator.CreateInstance(type);
-			modWorld.mod = this;
-			string name = type.Name;
-			if (modWorld.Autoload(ref name)) {
-				AddModWorld(name, modWorld);
-			}
+		private void AutoloadInstance(Type type) {
+			var loadable = (ILoadable)Activator.CreateInstance(type);
+			loadable.Load(this);
+			loadables.Add(loadable);
```
Chicken-Bones, line 363: nuke this method, just call `AddContent` with the new instance above.

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -278,91 +284,24 @@ partial class Mod
...
-					AutoloadWaterfallStyle(type);
-				}
-				else if (type.IsSubclassOf(typeof(GlobalRecipe))) {
-					AutoloadGlobalRecipe(type);
-				}
-				else if (type.IsSubclassOf(typeof(ModCommand))) {
-					AutoloadCommand(type);
+				else if (typeof(ILoadable).IsAssignableFrom(type)) {
+					bool? autoload = AutoloadAttribute.GetValue(type);
+					if (autoload ?? Properties.Autoload) {
```
Chicken-Bones, line 302: Can we remove `Properties` and just give mod an overridable property?

patches/tModLoader/Terraria/ModLoader/ModType.cs
```diff
@@ -0,0 +1,34 @@
...
+		/// <summary>
+		/// The internal name of this.
+		/// </summary>
+		public virtual string Name => GetType().Name;
+
+		void ILoadable.Load(Mod mod) {
+			Mod = mod;
+			Load();
+			AddInstance();
+		}
```
Chicken-Bones, line 22: This separation seems to lack justification. From what I can see in this PR, `Load` is only further overridden once in `ModDust` and the code there could easily go in `AddInstance`

I suggest removing both `ILoadable.Load` and `AddInstance` and just using the single `Load(Mod)` override be inherited and passed up and down. 

~~You should also make `Mod` `private set;`~~ so people can't forget to call `base.Load(mod)` for long. Not that many people will be making custom `ModType` subclasses

Still no great way to do 'cloning'/construction, but that's another day's problem.

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -198,23 +199,28 @@ partial class Mod
...
 
 			foreach (ModWaterfallStyle waterfallStyle in waterfallStyles.Values) {
-				Main.instance.waterfallManager.waterfallTexture[waterfallStyle.Type] = ModContent.GetTexture(waterfallStyle.texture);
+				Main.instance.waterfallManager.waterfallTexture[waterfallStyle.Type] = ModContent.GetTexture(waterfallStyle.Texture);
 			}
 		}
 
 		internal void UnloadContent() {
 			Unload();
+			foreach(var loadable in loadables.Reverse()) {
```
Chicken-Bones, line 219: good

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -278,91 +284,24 @@ partial class Mod
...
-				}
-				else if (type.IsSubclassOf(typeof(ModWaterfallStyle))) {
-					AutoloadWaterfallStyle(type);
-				}
-				else if (type.IsSubclassOf(typeof(GlobalRecipe))) {
-					AutoloadGlobalRecipe(type);
-				}
-				else if (type.IsSubclassOf(typeof(ModCommand))) {
-					AutoloadCommand(type);
+				else if (typeof(ILoadable).IsAssignableFrom(type)) {
```
Chicken-Bones, line 300: heh, the rest of this chain uses `IsSubclassOf`

patches/tModLoader/Terraria/ModLoader/ModBuff.cs
```diff
@@ -1,61 +1,43 @@
...
 		public bool canBeCleared = true;
 
-		/// <summary>
-		/// Allows you to automatically load a buff instead of using Mod.AddBuff. Return true to allow autoloading; by default returns the mod's autoload property. Name is initialized to the overriding class name and texture is initialized to the namespace and overriding class name with periods replaced with slashes. Use this method to either force or stop an autoload, and to change the default display name and texture path.
-		/// </summary>
-		public virtual bool Autoload(ref string name, ref string texture) {
-			return mod.Properties.Autoload;
+		internal sealed override void AddInstance() {
+			if (Mod.buffs.ContainsKey(Name))
+				throw new Exception("You tried to add 2 ModBuff with the same name: " + Name + ". Maybe 2 classes share a classname but in different namespaces while autoloading or you manually called AddBuff with 2 buffs of the same name.");
```
Chicken-Bones, line 32: it'd be nice if we could genericise this check

patches/tModLoader/Terraria/ModLoader/ModTileEntity.cs
```diff
@@ -192,13 +182,25 @@ public abstract class ModTileEntity : TileEntity
...
 		}
 
-		/// <summary>
-		/// Allows you to automatically load a tile entity instead of using Mod.AddTileEntity. Return true to allow autoloading; by default returns the mod's autoload property. Name is initialized to the overriding class name. Use this method to either force or stop an autoload, or change the default display name.
-		/// </summary>
-		public virtual bool Autoload(ref string name) {
-			return mod.Properties.Autoload;
+		void ILoadable.Load(Mod mod) {
+			Mod = mod;
+			Load();
```
Chicken-Bones, line 187: as before, on inheritance and `base.Load`

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
Rartrin: `IsSubclassOf` doesn't work the same for interfaces. IIRC, when used on an interface it will determine if one interface is derived from another. When you need to check if a class implements an interface, you need to use `IsAssignableFrom`.

patches/tModLoader/Terraria/ModLoader/ModType.cs
Rartrin: I admit I went back and forth on this method entirely. In a perfect world, the new Load() method handle it and then anyone wanting to override it would call base.Load(). Given that most tML hooks don't require this, I decided to have a separate method for this with the build-in types to ensure that they get loaded. Anyone wanting to extend this base class can simply just add their stuff in the Load method. By making it internal, no modders can accidentally call it or replace it.

patches/tModLoader/Terraria/ModLoader/ModType.cs
Rartrin: Load was left for modders to overload for any code they need since Autoload() is no longer available. If you'd rather it use a single Load, I can do that. I wanted Mod to have a private set, but the cloning stuff needs it. Perhaps we could generalize the cloning into ModType.

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
Jofairden: Yup I remember that too

PR #850, opened 2020-05-01
Made ExampleExplosive count towards Demolitionist Spawn

chloenaut:
**Description of the Change**

Adds an IL Hook that allows ExampleExplosive to count towards Demolitionist spawn.

**Why this should be merged into tModLoader**

This is an example of how to edit vanilla NPC spawn conditions using IL edits.

**Possible drawbacks**

Could break in the upcoming update,  however should be fixable

**Applicable Issues**

N/A

**Sample Usage**

This item is already existent within Example Mod this just adds an extra feature.

**Are there alternative designs?**

Potentially we could add a hook or some other method to alter vanilla spawn conditions.
Unless there is another method that I am not aware of. 

Feel free to give feedback on the implementation. I don't exactly know if this is the best way to implement this kind of edit.


ExampleMod/Items/Weapons/ExampleExplosive.cs
```diff
@@ -1,20 +1,51 @@
...
+		//Trying to find the if statement that checks the players inventory for explosives and add Example explosive as one of them
+		private void HookUpdateTime_SpawnTownNPCs(ILContext il)
+		{
+			ILCursor cursor = new ILCursor(il);
+			//Try to find where the item id for the Bomb is pushed onto stack
+			if (!cursor.TryGotoNext(MoveType.Before,
+				i => i.MatchLdcI4(166)))
+				return;
+			//Move to the start of the if statement 
+			cursor.Index -= 7;
```
Jofairden, line 27: This seems pretty scary

chloenaut:
added a revision to make it more accurate

Chicken-Bones:
A good idea, I'll review the IL for it when I have time

JavidPack:
K, I just brought ItemID.Sets.ItemsThatCountAsBombsForDemolitionistToSpawn from 1.4 here instead of IL editing.

Chicken-Bones:
Just want to say that I'm sorry I didn't review this earlier. But the IL code is actually really good!
I'd try and find a way to not harcode the local variable index (37) and instead derive it from somewhere it's used.

Issue #847, opened 2020-04-29
'Fluent API' Recipe Builder syntax


webmilio:
### Description
An extension coupled with a RecipeBuilder class, allowing for easier recipe building. The current implementation is in my library and works fine, although I assume it would need some adjusting to fit tML standards.

### What does this proposal attempt to solve or improve?

The way recipes are added right now is not exactly the most pleasant, albeit not atrocious. This syntax aims to solve the issue by providing some kind of 'natural language' implementation to recipes added in AddRecipes.

### Which (other) solutions should be considered?

I am unaware of any other solutions such as this one, other than Mirsario's.

### Examples
Here are two examples that show what the syntax could be.
The first example instantiates a new RecipeBuilder manually.

![image](https://user-images.githubusercontent.com/836345/80650836-b9eccd00-8a42-11ea-886a-3c9d2efeec88.png)

The second example uses both the extension method and multiple crafting stations

![image](https://user-images.githubusercontent.com/836345/80650927-eb659880-8a42-11ea-9508-773ea5331c87.png)

The current implementation has several other methods which I've not included in this issue, but are as follows:

- `SetResult`: calls `SetResult` for the recipe
- `SetResultStack`: changes only the amount produced, useful in case of multiple recipes producing the same item but not the same quantity
- Several instances of `New`, which take different parameters. `New` will automatically call `Finish` on an old recipe if its not been added to the game
- `Clear`: effectively calls `New` with the existing information. Not named `New` since it only works if you already have an existing product in the builder
- The different constructors for `RecipeBuilder` (which all follow the different `SetResult` for `ModRecipe`)

Chicken-Bones:
I really like the syntax and concept.
My primary feedback is to make sure the system isn't too "overloaded". As per the zen of python "There should be one-- and preferably only one --obvious way to do it."

We want this to be _the one_ and _the obvious_ way
so `BuildRecipe()` would be the way to do it from within `AddRecipes()` for an item.

"New will automatically call Finish on an old recipe if its not been added to the game"
I don't like this, it's hidden, and could cause bugs with forgetting the last recipe in the chain. "Explicit is better than implicit."

I suggest replacing `RequiresSinglesOf` with just `Requires` and having a few overloads

perhaps the best way is to construct a recipebuilder with the target item and qty, and always have the last call be `At`. If no station is required just call `At(null)` or `Anywhere()`

Jofairden:
First of all, thanks for your contribution / idea!
[...]

PR #856, opened 2020-05-05
Fluent API RecipeBuilder syntax

webmilio:
### What is the new feature?
Recipe building using a fluent api style syntax.

Because the current way recipes are added is a bit lacking.

For more information, check out [this issue](https://github.com/tModLoader/tModLoader/issues/847).


patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+			_recipe = new ModRecipe(mod);
+			_recipe.SetResult(mod, itemName, stack);
+		}
+
+		/// <summary>Creates a new instance with no ingredients with the given item and stack as a result.</summary>
+		/// <param name="item"></param>
+		/// <param name="stack"></param>
+		/// <seealso cref="Requires(int,int)"/>
+		/// <seealso cref="Build"/>
+		public RecipeBuilder(ModItem item, int stack = 1)
```
Jofairden, line 49: I would get rid of this, it's a bit dangerous to just 'inherit' the mod of the item. You will be adding a recipe 'as another mod' if the item comes from another mod.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+		/// Adds the specified ingredients to this recipe with the given item types.
+		/// Ex.: 
+		/// <example>recipe.AddIngredient(ItemID.IronAxe)</example>
+		/// </summary>
+		/// <param name="type1">The first item type.</param>
+		/// <param name="type2">The second item type.</param>
+		/// <param name="type3">The third item type.</param>
+		/// <param name="types">The remaining item types.</param>
+		/// <returns></returns>
+		public RecipeBuilder Requires(int type1, int type2, int type3, params int[] types)
```
Jofairden, line 92: Why type1 through 3 and then params types? Just do params types.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+
+		/// <summary>Adds an ingredient to this recipe with the given item name from the given mod, and with the given stack stack. If the mod parameter is null, then it will automatically use an item from the mod creating this recipe.</summary>
+		/// <param name="mod">The mod.</param>
+		/// <param name="itemName">Name of the item.</param>
+		/// <param name="stack">The stack.</param>
+		/// <returns></returns>
+		/// <exception cref="RecipeException">The item " + itemName + " does not exist in mod " + mod.Name + ". If you are trying to use a vanilla item, try removing the first argument.</exception>
+		public RecipeBuilder Requires(Mod mod, string itemName, int stack = 1)
+		{
+			_recipe.AddIngredient(mod ?? _recipe.mod, itemName, stack);
```
Jofairden, line 115: Why the null coalescing? The player should pass mod, if it is incorrect there should be an error

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+			need(_recipe);
+
+			return this;
+		}
+
+
+		/// <summary>Adds this recipe to the game. Call this after you have finished setting the result, ingredients, etc.</summary>
+		/// <returns></returns>
+		/// <exception cref="RecipeException">A recipe without any result has been added.</exception>
+		public void Build() => _recipe.AddRecipe();
```
Jofairden, line 182: You should have a 'finalized' or 'added' bool on this builder and check against it in any of the method calls, the builder is 'done' once Build() is called.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+{
+	public class RecipeBuilder
+	{
+		private readonly ModRecipe _recipe;
+
+
+		/// <summary>Creates a new instance with no ingredients and no result.</summary>
+		/// <seealso cref="Requires(int,int)"/>
+		/// <seealso cref="Build"/>
+		public RecipeBuilder()
```
Jofairden, line 16: It also doesn't initialize the recipe, remove this.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: As per the TML documentation, if the passed mod is null, it'll use the one who created the recipe.

See http://tmodloader.github.io/tModLoader/html/class_terraria_1_1_mod_loader_1_1_mod_recipe.html#a7c390feb8cf0a3212de31b53ad5bcfc2.

I could let `AddIngredient` do the work if thats what it already does I suppose.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: This is the equivalent of `ModRecipe.SetResult(ModItem, int)`, which is pretty standard in most mods adding ModItems with recipes. I don't think this should be removed, but maybe fixed ? Adding a mod parameter perhaps ?

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: Thats a lot of hand-holding. Do we need to throw an error every time as well ?

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: Will do.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: I did as was discussed on this [comment](https://github.com/tModLoader/tModLoader/issues/847#issuecomment-621730542), but I will change it back to `Requires(params int[])`.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
Mirsario: Strings should only be usable for accessing other mods' content in such a manual context. Why is there even a string overload if there isn't one with a generic ModItem parameter?

webmilio: I'm not entirely sure what you mean. I followed the existing structure that ModRecipe provides. Do you mean to say we should have a `Require<T>(...) where T : ModItem` ?

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
Jofairden: Ok if it's the same as that one, then it's fine

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+		public RecipeBuilder RequiresHoney() => NeedVar(recipe => recipe.needHoney = true);
+
+		public RecipeBuilder RequiresWater() => NeedVar(recipe => recipe.needWater = true);
+
+		public RecipeBuilder RequiresSnowBiome() => NeedVar(recipe => recipe.needSnowBiome = true);
+
+
+		private RecipeBuilder NeedVar(Action<ModRecipe> need)
+		{
+			need(_recipe);
```
Jofairden, line 179: What is need ?

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
Jofairden: You put this back, by accident?

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -100,11 +112,17 @@ public RecipeBuilder Requires(params int[] types)
...
+			_recipe.AddIngredient(mod, itemName, stack);
 
 			return this;
 		}
 
+		/// <summary>Adds an ingredient to this recipe of the given type and stack size.</summary>
+		/// <typeparam name="T">The type.</typeparam>
+		/// <param name="stack">The stack.</param>
+		/// <returns></returns>
+		public RecipeBuilder Requires<T>(int stack = 1) where T : ModItem => Requires(ModContent.ItemType<T>(), stack);
```
Jofairden, line 124: If we're on this route, you could make a `Requires(params Type[] types)` so you can do `Requires(typeof(MyItem))`
But it gets pretty verbose at that point, `Requires<T>` is fine

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: Simple method I made that sets a flag to true and returns RecipeBuilder.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
webmilio: Yea, I'll remove it again. Seems I forgot to fetch before pulling.

patches/tModLoader/Terraria.ModLoader.Recipes/RecipeBuilder.cs
```diff
@@ -0,0 +1,182 @@
...
+		public RecipeBuilder At(params int[] tileTypes)
+		{
+			for (int i = 0; i < tileTypes.Length; i++)
+				_recipe.AddTile(tileTypes[i]);
+
+			return this;
+		}
+
+
+		public RecipeBuilder RequiresLava() => NeedVar(recipe => recipe.needLava = true);
```
Chicken-Bones, line 160: no need for an Action and a common method here. Yes it lets you use expression methods. No it makes it harder to read.
```
public RecipeBuilder RequiresLava() {
    _recipe.needLave = true;
    return this;
}
```

Mirsario:
I've realized that this is completely unfit for custom ModRecipe types. We'll need to look into redesigning recipes, because extending the ModRecipe class currently looks pretty ugly, and still will with RecipeBuilder. Have a look: https://paste.mod.gg/rupotepaqi.cs

Chicken-Bones:
ModRecipe should be a builder itself. There's really no need for a separate builder class. Just make all the existing functions `return this;`

webmilio:
> I've realized that this is completely unfit for custom ModRecipe types. We'll need to look into redesigning recipes for OOP, because extending the ModRecipe class currently looks pretty ugly, and still will with RecipeBuilder. Have a look: https://paste.mod.gg/esetilitug.cs

What exactly does this show ? I don't see you using the RecipeBuilder for comparaison.

<br>

> ModRecipe should be a builder itself. There's really no need for a separate builder class. Just make all the existing functions `return this;`

I agree, I wasn't too sure if I could modify ModRecipe. I'll be unable to change anything for the next week and a half sadly.

Chicken-Bones:
This might've been too hasty. I think there was some good syntax discussion we could use for improvement, and changing signatures of current `ModRecipe` methods is a breaking change.

Jofairden:
It was, I kind of assumed it'd work because all I was thinking of was "if you don't use the return value right now that doesn't matter", though that doesn't matter for already compiled mods >.>

Itorius:
Fluent API was already merged into 1.4.

PR #860, opened 2020-05-06
Added filling of textures' Name and Tag properties to ease debugging of spritebatch issues

Mirsario:
This PR modifiers Main.OurLoad so that it fills GraphicsResource.Name & Tag properties with texture path and Terraria strings. Additionally, mod textures will now have their Tag property set to the mod's name.

This should help modders easily identify faulty textures passed to SpriteBatch methods during, for example, an ObjectDisposedException, which previously frequently forced modders to guess what the texture was only based on its width & height properties, if the texture used was vanilla.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/ModInternals.cs
+++ b/patches/tModLoader/Terraria.ModLoader/ModInternals.cs
@@ -69,7 +69,10 @@ namespace Terraria.ModLoader
 							Language.GetTextValue("tModLoader.LoadErrorTextureFailedToLoad", path), t.Exception);
 
 					var tex = t.Result;
+					
 					tex.Name = Name + "/" + path;
+					tex.Tag = Name;
+
 					lock (textures)
 						textures[path] = tex;
 				}));
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -741,7 +741,7 @@
  		}
  
  		private void ClientInitialize() {
-@@ -6674,41 +_,48 @@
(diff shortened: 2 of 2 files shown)


JavidPack:
Are you sure this will help? My memory was that this won't help with ObjectDisposedException since the Texture.Dispose method clears out Name and Tag

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -760,16 +760,30 @@
  
  		public T OurLoad<T>(string path) {
```
Chicken-Bones, line 761: Can you reduce patch size by making a wrapper method, and just renaming this to `OurLoad_`?

patches/tModLoader/Terraria.ModLoader/ModInternals.cs
```diff
@@ -69,7 +69,10 @@ partial class Mod
 							Language.GetTextValue("tModLoader.LoadErrorTextureFailedToLoad", path), t.Exception);
 
 					var tex = t.Result;
+					
 					tex.Name = Name + "/" + path;
+					tex.Tag = Name;
```
Chicken-Bones, line 74: Tag seems redundant here, why can't we just use `Name`? I'd like to leave `Tag` for other modders to use

Mirsario:
> Are you sure this will help? My memory was that this won't help with ObjectDisposedException since the Texture.Dispose method clears out Name and Tag

Yes, seems to be the case, my bad. I'll close this until I come up with a solution.

Chicken-Bones:
can always hook the `Dispose` method to not do that.

PR #861, opened 2020-05-07
Fixed UnauthorizedAccessExceptions in the setup tool by replacing directory deletion with wiping.

Mirsario:
There's also protection against file attribute issues.

It's good enough for now? Won't clash with VS anymore.


Jofairden:
Thx!

PR #868, opened 2020-05-08
add additional IsAir checks to PlayerHooks

direwolf420:
### Not-Really-Bugfix
`PlayerHooks` methods with`Item item` in them usually have an `item.IsAir` check at the start to validate. Without these checks, it can happen that an "empty" item is passed in and then the GlobalItem hooks will error (same manifestation as https://github.com/tModLoader/tModLoader/pull/830). This PR just adds this check to other hooks where such a scenario can occur. Someone needs to doublecheck if the currently returned value for non-void methods I specified is valid depending on the context.

Example of such an error because an empty item has a `globalNPCs` array length of 0, and there is atleast one mod adding an instanced `GlobalNPC` class:
```cs
[10:30:36] [1/WARN] [tML]: Silently Caught Exception: 
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Terraria.ModLoader.GlobalItem.Instance(Item item)
   at Terraria.ModLoader.ItemLoader.OnHitNPC(Item item, Player player, NPC target, Int32 damage, Single knockBack, Boolean crit)

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/PlayerHooks.cs
+++ b/patches/tModLoader/Terraria.ModLoader/PlayerHooks.cs
@@ -660,7 +660,9 @@ namespace Terraria.ModLoader
 		private static HookList HookGetWeaponCrit = AddHook<DelegateGetWeaponCrit>(p => p.GetWeaponCrit);
 
 		public static void GetWeaponCrit(Player player, Item item, ref int crit) {
-			if (item.IsAir) return;
+			if (item.IsAir)
+				return;
+
 			foreach (int index in HookGetWeaponCrit.arr) {
 				player.modPlayers[index].GetWeaponCrit(item, ref crit);
 			}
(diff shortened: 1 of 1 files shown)


JavidPack:
Some of the returns are wrong.

But what I want to know is how empty items are triggering these methods, that doesn't make sense. OnHitNPC, for example, only runs as an actual item is swinging. The only way I can see it not being correct is if someone forces player.selectedItem to change at a time when it shouldn't. I'd rather the mod causing this issue know that they are doing it wrong rather than make the game seem to not complain about their code. 

Can you reproduce this particular error? I'd like to see the mod causing the issue.

direwolf420:
I was very hesitant in pushing this PR because it has no reason to exist in tml, since only bad mods can cause this error

Jofairden:
Agreed with Jopo. Checks aren't bad, but in this case if it's in hooks where air items are triggering the hooks something else is wrong, they should not trigger those hooks at all as it is 'nothing'.

Jofairden:
Adding to 1.4 milestone since I wanna make sure we check this again for 1.4

Solxanich:
@direwolf420 
Mini-review - 
Update to target 1.4. 
Investigate a bit further into why hooks are running on IsAir items per earlier comments; basically aim to fix the root cause first instead of masking the issue with checks

JavidPack:
This issue relates to 1.3 tModLoader, which is now legacy and will not receive any further non-critical updates or bug fixes. Please give 1.4 tModLoader a try, there is a good chance this issue might be resolved there already. If you find that 1.4 tModLoader has a similar issue, feel free to open a new issue and fill out the issue template so we can once again track this issue. This issue will now be closed.

PR #871, opened 2020-05-10
Add improved MessageBoxShow error messages

Trivaxy:
### What is the new feature?
Improving upon the error messages that are shown by `ModLoader.UI.Interface.MessageBoxShow` to point users towards their client.log file and, depending on the exception, provides a unique tip that may fix it.
(If tModLoader does not recognize the exception, no tip is shown, but it will still point you towards client.log)

It may help alleviate constantly needing to provide users with the same solutions for frequent errors/exceptions by pointing them in the right direction immediately, allowing them to give it a shot at fixing the error themselves (or at least letting them know the importance of your client.log file 😛 ).

Changing error messages, adding/removing exception tips, or just modifying what is shown in general.

These are examples of some of the new error messages (all exceptions shown here are manually thrown as Terraria is launching). There are other exception messages not shown here.
image
image
image
image

(P.S: The changes git detected are a little weird... not sure why. Sorry about that 😅 )

### DIFF
--- a/patches/tModLoader/Terraria/Program.cs.patch
+++ b/patches/tModLoader/Terraria/Program.cs.patch
@@ -213,7 +213,7 @@
  
  #if CLIENT
 -				MessageBox.Show(e.ToString(), "Terraria: Error");
-+				ModLoader.UI.Interface.MessageBoxShow(e.ToString());
++				ModLoader.UI.Interface.MessageBoxShow(e, generateTip: true);
 +
 +				if (e is FolderCreationFailedException)
 +					Process.Start(e.HelpLink);
--- a/patches/tModLoader/Terraria.ModLoader.UI/Interface.cs
+++ b/patches/tModLoader/Terraria.ModLoader.UI/Interface.cs
@@ -378,12 +378,36 @@ namespace Terraria.ModLoader.UI
 		internal static void MessageBoxShow(string text, string caption = null) {
 			// MessageBox.Show fails on Mac, this method will open a text file to show a message.
 			caption = caption ?? "Terraria: Error" + $" ({ModLoader.versionedName})";
+			string message = $"{text}\n\nA client.log file containing error information has been generated in\n{Path.Combine(Main.SavePath, "Logs")}\n(You will need to share this file if asking for help)";
 #if !MAC
-			System.Windows.Forms.MessageBox.Show(text, caption);
+			System.Windows.Forms.MessageBox.Show(message, caption);
 #else
 			File.WriteAllText("fake-messagebox.txt", $"{caption}\n\n{text}");
 			Process.Start("fake-messagebox.txt");
 #endif
+		}
+
+		internal static void MessageBoxShow(Exception e, string caption = null, bool generateTip = false) {
+			string tip = "";
+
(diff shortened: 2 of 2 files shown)


Jofairden:
This could be a really neat addition, though the inherent problem with the errors is that players don't even read them to begin with :P

Chicken-Bones:
I like all of this, and am happy to click that big green button, with 1 caveat. Lets not tell people to disable their antivirus when we're installing mods, it just might protect them from a malicious mod. Are we sure there's no other solution here?

Trivaxy:
> I like all of this, and am happy to click that big green button, with 1 caveat. Lets not tell people to disable their antivirus when we're installing mods, it just might protect them from a malicious mod. Are we sure there's no other solution here?

I completely agree with you. Perhaps I should change the message to simply let the user know that the crashing might be caused by their antivirus so they can take action on their own? (rather then telling them to allow tModLoader through). Of course depending on the situation the only solution might be to simply add tML to your whitelist.

JavidPack:
Let's wait for feedback from 0.11.7 to implement this. We fixed a cause of the synchronization lock exception, we should be able to get a better solution. We also will probably be falling back to a different save directory if Onedrive issues persist, but that will also have to wait for more data from the steam release.

patches/tModLoader/Terraria/Program.cs.patch
```diff
@@ -109,12 +109,16 @@
  			}
  		}
  
-@@ -202,9 +_,18 @@
+@@ -202,13 +_,43 @@
  				}
  
  #if CLIENT
 -				MessageBox.Show(e.ToString(), "Terraria: Error");
-+				ModLoader.UI.Interface.MessageBoxShow(e.ToString());
```
Jofairden, line 117: I suggest the following:
`ModLoader.UI.Interface.MessageBoxShow(e);` 
and have it enhance the exception automatically.
Making a method outside the class in one of our own reduces the patch, and possible margin of error
A method that enhances the error is the case here, with a simple change of calling our own MessageBox instead of vanillas'

patches/tModLoader/Terraria/Program.cs.patch
```diff
@@ -130,3 +134,24 @@
  #endif
  			}
  			catch {
+ 			}
++		}
++
++		private static string GetTip(Exception e) {
```
Jofairden, line 140: It might be preferred to not have this method in Program.cs, but one of our own files.
To reduce the patch, do what I suggested.

patches/tModLoader/Terraria/Program.cs.patch
Trivaxy: Done - I created a `MessageBoxShow` overload that takes an `Exception` and an optional `generateTip` parameter. By default, it calls the original `MessageBoxShow` with the exception message, but if `generateTip` is true then it'll also show a tip. Hopefully this is what you meant 👍

Jofairden: It's probably better if MessageBoxShow accepts a string parameter, then make a 'GenerateTip()' method that returns the standard tips. This allows custom strings when needed.

PR #877, opened 2020-05-14
Fixed DirectoryNotFoundException in setup's DeleteEmptyDirs

Mirsario:


setup/Setup/SetupOperation.cs
```diff
@@ -129,16 +129,25 @@ protected class WorkItem
...
+
+				if(!allEmpty || Directory.EnumerateFiles(dir).Any())
+					return false;
+
+				Directory.Delete(dir);
+
+				return true;
+			}
+
+			return Recursion(directory);
```
Jofairden, line 150: Can't this be resolved a bit simpler?
```suggestion
		public static bool DeleteEmptyDirs(string directory) {
			if (!Directory.Exists(directory)) return true;
			
			bool allEmpty = true;
			
			foreach(string subDir in Directory.EnumerateDirectories(directory))
				allEmpty &= DeleteEmptyDirs(subDir);
			
			if (!allEmpty || Directory.EnumerateFiles(dir).Any()) return false;
			
			Directory.Delete(directory);
			return true;
		}
```

Mirsario: You're asking to revert my second commit, which just got rid of useless Directory.Exists checks past the first invocation. I'll just rewrite this method, sec.

PR #878, opened 2020-05-16
Improved stability of OutOfMemory handling

Mirsario:
### What is the bug?
OutOfMemory may occur inside FirstChanceException when, for example, trying to allocate enough memory to show the message box.

### How did you fix the bug?
I added code to unload Main.tile array and call GC.Collect() as early in the method as possible, and also disabled any ways of blocking the handling of the exception.

This is the most practical solution, but not the most good looking.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/Logging.cs
+++ b/patches/tModLoader/Terraria.ModLoader/Logging.cs
@@ -188,22 +188,22 @@ namespace Terraria.ModLoader
 			if (handlerActive.Value)
 				return;
 
-			bool ignoreable = true;
+			bool oom = args.Exception is OutOfMemoryException;
 
 			//In case of OOM, unload the Main.tile array and do immediate garbage collection.
 			//If we don't do this, there will be a big chance that this method will fail to even quit the game, due to another OOM exception being thrown.
-			if (args.Exception is OutOfMemoryException) {
+			if (oom) {
 				Main.tile = null;
 
 				GC.Collect();
 
-				ignoreable = false;
+				oom = false;
 			}
 
 			try {
 				handlerActive.Value = true;
 
-				if (ignoreable)
+				if (oom)
 					if (args.Exception == previousException ||
 						args.Exception is ThreadAbortException ||
 						ignoreSources.Contains(args.Exception.Source) ||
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria.ModLoader/Logging.cs
```diff
@@ -188,27 +188,38 @@ public static class Logging
...
 			try {
 				handlerActive.Value = true;
 
-				if (args.Exception == previousException ||
-					args.Exception is ThreadAbortException ||
-					ignoreSources.Contains(args.Exception.Source) ||
-					ignoreMessages.Any(str => args.Exception.Message?.Contains(str) ?? false) ||
-					ignoreThrowingMethods.Any(str => args.Exception.StackTrace?.Contains(str) ?? false))
-					return;
+				if (!oom)
```
Chicken-Bones, line 204: braces please

patches/tModLoader/Terraria.ModLoader/Logging.cs
```diff
@@ -188,27 +188,38 @@ public static class Logging
...
 
-				if (ignoreContents.Any(traceString.Contains))
+				if (!oom && ignoreContents.Any(traceString.Contains))
 					return;
 
 				traceString = traceString.Substring(traceString.IndexOf('\n'));
 				var exString = args.Exception.GetType() + ": " + args.Exception.Message + traceString;
 				lock (pastExceptions) {
-					if (!pastExceptions.Add(exString))
+					if (!pastExceptions.Add(exString) && !oom)
```
Chicken-Bones, line 222: redundant, can't have 2 ooms in a row

Mirsario: We indeed can and did in the past. That exact return had to be the *second* reason why the game would continue to look semi-fine after encountering OOM and failing to show the error message. But sure, I'll remove the check, hoping that Main.tile unloading will always free enough memory.

patches/tModLoader/Terraria.ModLoader/Logging.cs
```diff
@@ -188,21 +188,33 @@ public static class Logging
 			if (handlerActive.Value)
 				return;
 
+			bool oom = args.Exception is OutOfMemoryException;
+
+			//In case of OOM, unload the Main.tile array and do immediate garbage collection.
+			//If we don't do this, there will be a big chance that this method will fail to even quit the game, due to another OOM exception being thrown.
+			if (oom) {
+				Main.tile = null;
```
Jofairden, line 196: Are there other kinds of unloading we can do here to free up more memory? (We should look into unloading all textures from items and such, since they are now loaded on demand)

Issue #894, opened 2020-05-17
Terraria: Error (tModLoader v0.11.7) steam version


alexandrage:
Terraria: Error (tModLoader v0.11.7) steam version

System.0verf|owException:

a
MonoMod.RuntimeDetour.Platforms.DetourRuntimeNETPlatform.GetFu
nctionPointermlethodBase method, RuntimeMethodHandle handle)

5
MonoMod.RuntimeDetour.Platforms.DetourRuntimelLPlatform.GetNativ
eStart[MethodBase method)

a MonoMod.RuntimeDetour.Detour._TopApply[)

5 MonoMod.RuntimeDetour.Detour.-RefreshChain[MethodBase
method)

5 MonoMod.RuntimeDetour.Detour.Applyl)

a MonoMod.RuntimeDetour.Hook.Apply0

5 MonoMod.RuntimeDetour.Hook..ctor[MethodBase from,
Methodlnfo to, Object target, HookConfigBL config)

5Terraria.ModLoader.Core.LoggingHooks.HookProcessStart[)

aTerraria.ModLoader.Logging.lnit[)

BTerraria.Program.LaunchGame[Stringfl args, Boolean monoArgs)

GoodPro712:
If you are getting the `Arithmetic operation resulted in an overflow.` error and would like to help test a fix, download this zip, and place the files from within this zip into your `C:\Program Files (x86)\Steam\steamapps\common\tModLoader` folder, then try to launch tModLoader.
[Test.zip](https://cdn.discordapp.com/attachments/436261997416546305/711513703287095317/20051401Test.zip)

Jofairden:
You can try what GoodPro said (I checked the files, they should be safe), fixed in next ver

PR #899, opened 2020-05-17
Updated MonoMod to 20.05.14.01 to fix 'Arithmetic operation resulted in an overflow' startup crashes.

Mirsario:
aka this thing:
image

All tested this time, (but only on Windows).


Mirsario:
(Fixes #894)

PR #910, opened 2020-05-18
Proposal: Use multiple cores to pre-JIT methods

tansheron:
Terraria uses only a single core to pre-JIT its methods during startup. Using multiple cores can speed this up. The difference is quite noticeable on my i7-4800MQ (8 cores) with tML 1.3 or vanilla 1.4. I expect that the time spent loading mods will dwarf the pre-JIT time on most end-user machines, but every little bit helps, especially when you're developing or testing code.

Topics for discussion:

- Is this of interest to tML at all? If all developers are assumed to be using the QuickLaunch startup option to bypass the pre-JIT step entirely, there won't be as much point to having it.
- The number of cores used is at the TPL's discretion. There could be configuration for this.
- Projectile.VanillaAI and (especially) NPC.VanillaAI take much longer to JIT than anything else does. They could be started before the rest of the methods, given priority, and be configured to not block the loading process from finishing; they would probably be done by the time anyone needs to call them. When I tested this with tML on my machine, Terraria went straight from the splash screen to loading tML without having to show the "Loading: x%" message in between. This would add complexity, though; tasks would need to be created individually since AsParallel doesn't allow TaskCreationOptions.PreferFairness to be used or await some tasks but not others.

### DIFF
--- a/patches/tModLoader/Terraria/Program.cs.patch
+++ b/patches/tModLoader/Terraria/Program.cs.patch
@@ -1,6 +1,6 @@
 --- src/Terraria/Terraria/Program.cs
 +++ src/tModLoader/Terraria/Program.cs
-@@ -2,6 +_,7 @@
+@@ -2,20 +_,24 @@
  using ReLogic.OS;
  using System;
  using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


JavidPack:
I just did a quick test, from 35 to 10 seconds. By the time the splash screen goes away the loading is already at 90% and finishes right up. Using QuickLaunch isn't really a reason not to pull this in, the end users also would benefit. Unless there is some undefined behavior that we aren't aware of, either due to the order of the Types or some Mac/Linux issues, this looks good to pursue.

Chicken-Bones:
Awesome

Jofairden:
Thanks a ton!

PR #930, opened 2020-05-22
Added CanActuate hooks

OliHeamon:
### What is the new feature?

Hooks for `ModTile`s/`GlobalTile`s that allow control over whether or not specific types of tile can be actuated.

Currently, the only way of achieving this is through a Runtime Detour, and adding the feature to tML will make it easier for more modders to grasp. Additionally, it may be useful for modders wishing to create more sophisticated structures that cannot be actuated into, reminiscent of the Jungle Temple.

### Possible drawbacks
N/A

### Applicable Issues
N/A

### ExampleMod updates
Will PR ExampleNonActuableTile shortly.


patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -50,6 +50,27 @@
  		private static int CurrentUser = 254;
  
  		public static void SetCurrentUser(int plr = -1) {
+@@ -226,7 +_,8 @@
+ 
+ 		public static bool Actuate(int i, int j) {
+ 			Tile tile = Main.tile[i, j];
+-			if (!tile.actuator())
++
```
Jofairden, line 58: This patch can be reduced in size by not having this additional space

patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -50,6 +50,27 @@
...
++			if (!tile.actuator() || !TileLoader.CanActuate(tile))
+ 				return false;
+ 
+ 			if ((tile.type != 226 || !((double)j > Main.worldSurface) || NPC.downedPlantBoss) && (!((double)j > Main.worldSurface) || NPC.downedGolemBoss || Main.tile[i, j - 1].type != 237)) {
+@@ -241,6 +_,10 @@
+ 
+ 		public static void ActuateForced(int i, int j) {
+ 			Tile tile = Main.tile[i, j];
++
++			if (!TileLoader.CanActuate(tile))
```
Jofairden, line 68: Are you sure this should be the same hook, or should it be a new hook? `CanActuateForcibly`, you can have that by default call `CanActuate`, which gives a bit finer control.

PR #931, opened 2020-05-22
Added extra hooks representing ActuateForced

oli-x64:
### What is the new feature?
Hooks for `ModTile`s/`GlobalTile`s that allow control over whether or not specific types of tile can be actuated. Also includes a second hook, `CanActuateForcibly`, so that modders can choose whether they want to also prevent their tiles from being force actuated using WorldGen, etc.

The title refers to the fact that this is the second PR concerning this, as the former one didn't have the correct hooks or an EM use case.

Currently, the only way of achieving this is through a Runtime Detour, and adding the feature to tML will make it easier for more modders to grasp. Additionally, it may be useful for modders wishing to create more sophisticated structures that cannot be actuated into, reminiscent of the Jungle Temple.

### Possible drawbacks
N/A

### Applicable Issues
N/A

### ExampleMod
Added use cases to EM.


Jofairden:
You could've just left the other PR open, but okay.

patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -50,6 +50,31 @@
  		private static int CurrentUser = 254;
  
  		public static void SetCurrentUser(int plr = -1) {
+@@ -226,21 +_,21 @@
+ 
+ 		public static bool Actuate(int i, int j) {
+ 			Tile tile = Main.tile[i, j];
+-			if (!tile.actuator())
++			if (!tile.actuator() || !TileLoader.CanActuate(tile))
```
Jofairden, line 58: Shouldn't CanActuate override tile.actuator? I thought the intention was to provide the possibly for actuation without vanilla logic.

Jofairden (changes requested):

ExampleMod/ExampleMod.csproj
```diff
@@ -13,4 +13,72 @@
...
+    <Reference Include="NVorbis">
+      <HintPath>..\..\..\tML\test\bin\Debug\net45\NVorbis.dll</HintPath>
+    </Reference>
+    <Reference Include="Terraria">
+      <HintPath>..\..\..\tML\test\bin\Debug\net45\Terraria.exe</HintPath>
+    </Reference>
+    <Reference Include="tModLoaderTests">
+      <HintPath>..\..\..\tML\test\bin\Debug\net45\tModLoaderTests.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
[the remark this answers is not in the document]
OliHeamon: Oops, will remove this.

patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -50,6 +50,31 @@
  		private static int CurrentUser = 254;
  
  		public static void SetCurrentUser(int plr = -1) {
+@@ -226,21 +_,21 @@
+ 
+ 		public static bool Actuate(int i, int j) {
+ 			Tile tile = Main.tile[i, j];
+-			if (!tile.actuator())
++			if (!TileLoader.CanActuate(tile) || !tile.actuator())
```
Jofairden, line 58: Isn't this still wrong, wasn't your intent to override vanilla? You'd need different logic:
```suggestion
+			if (!TileLoader.CanActuate(tile) && !tile.actuator())
```

Jofairden (changes requested):
Good, you fixed almost all of it. See comment.

Solxanich:
Hello @OliHeamon  ; thank you for drafting this PR.
As the tModLoader project continues to progress, I'd like to ask if you intend to return to complete this PR and address Jofairden's review comments?

If not, could you please let us know, so we can have someone else pick up the branch and bring it to completion?

Thank you for your time.

PR #937, opened 2020-05-26
Add PaintID class

direwolf420:
### What is the new feature?
Adds a class containing magic numbers associated with paints for tiles, such as:
`public const byte Violet = 11;`

Heavily improves access to paints and their readability.

How I did it before this to sort of "humanize" paints is create an item cache of valid paint items, and then access their paint through a dictionary so it would look like this: 
`byte paint = PaintCache[ItemID.RedPaint];`

```cs
byte paint = PaintID.Red;
WorldGen.paintTile(i, j, paint);
WorldGen.paintWall(i, j, paint);
```

### ExampleMod updates
Doesn't seem like something needing a showcase as it's a simple replacement for magic numbers


direwolf420:
I just noticed I borked the branch, redoing PR

direwolf420:
Duplicate of #938

PR #958, opened 2020-06-02
Make MessageBox texts localizable

### LINKED ISSUE #955: [Enhancement] Make MessageBox texts localizable
Some of the hardcoded texts should be localizable for translation.
Samples:
https://github.com/tModLoader/tModLoader/blob/4b1c029242e41ae822bf657db68edc7bdc75a877/patches/tModLoader/Terraria/Main.cs.patch#L802
https://github.com/tModLoader/tModLoader/blob/bdcd2e936692f36baf765c3dd752794b1d986824/patches/tModLoader/Terraria/Program.cs.patch#L225
Find more by searching `MessageBox`/`MessageBoxShow`/etc..

Add resources for those strings.
`LanguageManager` is loaded pretty early (at least before `SocialAPI`, and this sample is loaded after it), so we can simply add them to `Terraria.Localization.Content.[language].tModLoader.json`

sgkoishi:
Make `MessageBox` texts localizable, resolves #955

I'm not sure if the logging part should be localizable (which might break some format), so only `MessageBox` related strings were added right now.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Engine/InstallVerifier.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Engine/InstallVerifier.cs
@@ -48,7 +48,7 @@ namespace Terraria.ModLoader.Engine
 			else {
 				string message = "Unknown OS platform: unable to verify installation.";
 				Logging.tML.Fatal(message);
-				Exit(message, string.Empty);
+				Exit(Language.GetTextValue("tModLoader.UnknownVerificationOS"), string.Empty);
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria.ModLoader.Engine/InstallVerifier.cs
```diff
@@ -48,7 +49,7 @@ static InstallVerifier()
 			else {
 				string message = "Unknown OS platform: unable to verify installation.";
```
Jofairden, line 50: You forgot the logging message to also use the localized message

patches/tModLoader/Terraria.ModLoader/Logging.cs
```diff
@@ -240,7 +240,7 @@ public static class Logging
 				if (oom) {
 					const string error = "Game ran out of memory. You'll have to find which mod is consuming lots of memory, and contact the devs or remove it.";
```
Jofairden, line 241: Same thing here

Issue #482, opened 2019-01-19
Add piracy checking


Shimogawa:
### Description

It is believed that tML needs Steam to run, but recently I found that a modified `steam_api.dll` can be used to bypass the Steam check. Many players here do not want to pay for the game, so they use a pirate version of Terraria, and also tML if replacing the executable. Thus, I think a check on `steam_api.dll` should be performed.

### Versions

All versions.

### Additional Information

I suggest to perform a SHA1 or any other Hash check on the library before running `SocialAPI` to check the integrity of official version of tML including any other libraries it uses. 

The SHA1 hash of vanilla `steam_api.dll` is `4ee43374520904fa6d80c12c273d67eb7b5c984e`. I wrote my own solution as follows:
```cs
byte[] b = new byte[] { 0x4e, 0xe4, 0x33, 0x74, 0x52, 0x09, 0x04, 0xfa, 0x6d, 0x80, 0xc1, 0x2c, 0x27, 0x3d, 0x67, 0xeb, 0x7b, 0x5c, 0x98, 0x4e };
var data = File.ReadAllBytes("steam_api.dll");
var hash = SHA1.Create().ComputeHash(data);
if (hash.SequenceEqual(b)) {...}    // using System.Linq
```

~~Also, the pirate version of `steam_api.dll` can be downloaded here: [modified_steam_api.zip]~~

Chicken-Bones:
See, you're missing the part where anyone can modify and build tModLoader to remove the check.

Shimogawa:
Not everyone knows how to modify the client, and also modifying the dll is currently the only variant of piracy as far as I know. Also, I think adding this checker can at least block most of the users who use pirate version and also the piracy problem in China is so severe that we should take action.

Chicken-Bones:
Yeah but it only takes one person to modify it and release it. It's just like steam_api.dll but /easier/. Also tModLoader is small in comparison to Terraria.

Dierney:
To be honest, because of Chinese government, Chinese players do not have access to the websites outside China. We (SBMW Group), as the biggest mod localization and transferring group in China, are the administrator in the Chinese Terraria community. We really want to improve the game environment so that the severity of piracy can diminish, and we are providing the mods only to Steam and GOG players. 

If you feel it is not necessary, please give us the permission to add the checking code to the tML distribution to Chinese players by our own.

Chicken-Bones:
[...]

PR #964, opened 2020-06-04
Remove hash check, serves no useful purpose

flarn2006:
Simply put, it is not our job to make pirates' lives difficult. That is Re-Logic's job, if they choose to do so. This hash check has a zero chance of helping the end user, and a nonzero chance of causing frustration, even if the game is not pirated. What if the user bought it on Steam and wants to replace it with a cracked DLL to avoid difficulties with offline mode? What if there's an update that changes the DLL, but wouldn't otherwise break compatibility with tModLoader? Having it exit with an error in these cases is bad enough, and if it happens for a reason like this, that just adds insult to injury.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Engine/InstallVerifier.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Engine/InstallVerifier.cs
@@ -104,11 +104,6 @@ namespace Terraria.ModLoader.Engine
 				return false;
 			}
 #endif
-			if (!HashMatchesFile(steamAPIPath, steamAPIHash)) {
-				Process.Start(@"https://terraria.org");
-				Exit("Steam API hash mismatch, assumed pirated.\n\ntModLoader requires a legitimate Terraria install to work.", string.Empty);
-				return false;
-			}
 
 			Logging.tML.Info("Steam installation OK.");
 			return true;


sgkoishi:
> What if the user bought it on Steam and wants to replace it with a cracked DLL to avoid difficulties with offline mode?   

I guess this is a violation against EULA?

> What if there's an update that changes the DLL, but wouldn't otherwise break compatibility with tModLoader?

Updates are not for `compatibility` but for fixes and more contents. It's expected to have tModLoader update if the game itself updates.


You may also check [Issue #482](https://github.com/tModLoader/tModLoader/issues/482#issuecomment-455870013) about the `steam_api` hash check.

flarn2006:
If it's a violation of the EULA, that's probably just a technicality, and
not something the developers actually have an issue with. And either way,
it isn't tModLoader's responsibility.

What if tModLoader doesn't update properly?

sgkoishi:
If you install both Terraria and tModLoader from Steam, it is checking the one in `tModLoader` folder: the update of `Terraria`'s `steam_api` doesn't make any difference, you can still start and play tModLoader from Steam.

Chicken-Bones:
Yeah, read #482, every few months we get a complaint but I'm reasonably convinced that avenues of discussion are exhausted.

PR #966, opened 2020-06-05
Changed Localization Caps & UIManagePublished Error Text

GoodPro712:
### Description
Changed from when in the manage publish menu without internet to show "Mod Browser Offline (Unknown)" instead of "Mod Browser OFFLINE ()"

### Why should this be accepted?
Having offline all in caps isn't pleasant, in my opinion, and doesn't look nice with the other normal capitalized text. Brackets without anything in them doesn't seem correct, either.

### Notes
This capitalization change may affect other languages.
I don't feel like this is a bug or a feature so I hope it's fine if I write without a template
My first PR since 0.11.7.X so I hope I did well 😅 ..also my first PR using a proper git service

image


Jofairden:
Thanks.

PR #967, opened 2020-06-05
Lowered main menu button Y position

GoodPro712:
### What is the new feature?
Lowered the Y position of the buttons on the main screen of the main menu by 20 pixels.

New:
image

Currently, the buttons clash with the new large tModLoader logo and it looks improper.

Reduce how much the logo can scale up to.

N/A

### ExampleMod updates

N/A

### Notes
Also "fixed" a tiny mistake with `out var _` creating a variable when it doesn't need to :p

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.UI/Interface.cs
+++ b/patches/tModLoader/Terraria.ModLoader.UI/Interface.cs
@@ -74,7 +74,7 @@ namespace Terraria.ModLoader.UI
 				buttonNames[buttonIndex] = Language.GetTextValue("tModLoader.MenuModSources");
 				if (selectedMenu == buttonIndex) {
 					Main.PlaySound(10, -1, -1, 1);
-					Main.menuMode = ModCompile.DeveloperModeReady(out var _) ? modSourcesID : developerModeHelpID;
+					Main.menuMode = ModCompile.DeveloperModeReady(out _) ? modSourcesID : developerModeHelpID;
 				}
 				buttonIndex++;
 				numButtons++;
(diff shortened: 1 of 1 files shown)


Jofairden:
Does this also work for other menus like player creation?

patches/tModLoader/Terraria.ModLoader.UI/Interface.cs
```diff
@@ -74,7 +74,7 @@ internal static class Interface
 				buttonNames[buttonIndex] = Language.GetTextValue("tModLoader.MenuModSources");
 				if (selectedMenu == buttonIndex) {
 					Main.PlaySound(10, -1, -1, 1);
-					Main.menuMode = ModCompile.DeveloperModeReady(out var _) ? modSourcesID : developerModeHelpID;
```
Jofairden, line 77: What's exactly the issue with this? This is completely fine.

GoodPro712: Not really an issue. VS just underlines it and says it's unused

GoodPro712:
> Does this also work for other menus like player creation?
99% sure it doesn't, but I was thinking about doing that too.

Jofairden:
> > Does this also work for other menus like player creation?
> > 99% sure it doesn't, but I was thinking about doing that too.

Please do, since those menus have issues as well. Would be neat if this PR fixes all menus.

GoodPro712:
I'll give it a shot

Jofairden:
You still have time for this? Would be a good addition

GoodPro712:
Ill see if I can find time later in the week

Mirsario:
This is a bad idea, because these legacy UIs already have issues with some buttons going off-screen on minimal resolutions. Scaling down the logo seemed enough in 1.4.

Jofairden:
I agree with mirsario and I think I'll close this, I think the logo is a little too big and can be scaled down (it definitely really just is the cog being in the way)

PR #968, opened 2020-06-05
Public ModCompile

GoodPro712:
### What is the new feature?
Changed `ModCompile`, `modCompileDir` and `modCompileVersionPath` to `public`. I don't think paths can hurt anyone! They just wouldn't have to copy-paste the fields themselves :p

Currently, a modder has to use reflection to see if a user is in developer mode or not. I know most people will never need to know stuff like that, but this caters towards the people who do.

Making some of the fields/methods inside `ModCompile` `internal`

```cs
if (ModCompile.DeveloperMode)
    Main.NewText("The user is in developer mode!");
```

### ExampleMod updates

N/A

### Notes
Removed the unused `using Terraria.ModLoader.UI;` because VS had it underlined
Changed `DeveloperMode` to be a one-liner since VS had it underlined (I assume that means it's preferable?)

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
+++ b/patches/tModLoader/Terraria.ModLoader.Core/ModCompile.cs
@@ -16,14 +16,13 @@ using System.Threading;
 using System.Threading.Tasks;
 using Terraria.Localization;
 using Terraria.ModLoader.Exceptions;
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Could move, wouldn't want to break mods. Not really meant to be part of the "public API" `Debugger.IsAttached` is preferred.

PR #970, opened 2020-06-05
No problems found message instead of blank

GoodPro712:
### What is the new feature?
Changes the message box text to have "No Problems Found. If a problem occurs it will be shown here" instead of being empty

New:
image

Without this, the blank box makes users wonder what it's for and now they will know.

Different text?

N/A

### ExampleMod updates

N/A

### Notes
Weird file changed stuff - Jof said it looked fine
I actually did this in a mod first ;p
Could also apply to mods' configs


patches/tModLoader/Terraria.Localization.Content/zh-Hans.tModLoader.json
```diff
@@ -334,23 +334,12 @@
 
 		// Misc
 		"OK": "好的",
+		// "NoProblemsFound": "No Problems Found. If a problem occurs it will be shown here",
 		"Continue": "继续",
 		"PatreonSetTooltip": "感谢您的支持!",
 		"WorldGenError": "生成世界时遇到问题",
 		"WorldIODataException": "游戏遇到一个关于地图的模组数据问题.",
 		"ServerCrash": "服务器崩溃: {0}\n{1}\n\n请检查[server.log]以确定是否由模组导致,否则请转到tModLoader Discord中的#support"
-
```
Jofairden, line 342: You removed this commit:  https://github.com/tModLoader/tModLoader/commit/4356f94b797da276065c43b423c449fa5add310f
You'll have to put it back

GoodPro712:
Oh I didn't notice that, apologies. That was a waste of an hour, I should communicate next time :p

patches/tModLoader/Terraria.Localization.Content/zh-Hans.tModLoader.json
GoodPro712: ended up deleting the file, then downloading from Github and placing it back edited, lol

JavidPack:
So, was a similar solution already merged? should this be closed?

PR #979, opened 2020-06-10
Move more hardcoded strings to resource file

sgkoishi:
Addition to #958 
* Add dedicated server mod list
* Add dedicated server mod browser
* Add `MessageBoxShow` generated tip


Jofairden:
Thanks a bunch, this is definitely good.

PR #980, opened 2020-06-10
Update Chinese translation

sgkoishi:
Add missing comma
Add Chinese translations for #979
Use LF instead of CRLF line breaks


Jofairden:
Thanks a bunch! ^^

PR #984, opened 2020-06-14
Grapple target point hook + aiType

gamrguy:
### What is the new feature?
Adds a new hook, called `GrappleTargetPoint`, which allows changing the position a grappling hook pulls the player towards.
Also adds the ability to use `ModProjectile.aiType` to use the vanilla implementations of `TrackHook`, `AntiGravityHook`, and `StaticHook`.
Currently it is not possible to create grappling hooks similar to the Anti-Gravity Hook or Static Hook without IL editing or completely overriding the regular grapple code. This would allow it.
Using IL patches or completely overriding the `Player.GrappleMovement` functionality with custom code
See included ExampleMod example
### ExampleMod updates

Updated `ExampleHook.cs` to include a sample usage of this feature, which causes ExampleHook to suspend the player 50px away from the hook's position rather than pulling them all the way.
hoooook

### DIFF
--- a/ExampleMod/Items/ExampleHook.cs
+++ b/ExampleMod/Items/ExampleHook.cs
@@ -122,6 +122,13 @@ namespace ExampleMod.Items
 			speed = 4;
 		}
 
+		public override void GrappleTargetPoint(Player player, ref float grappleX, ref float grappleY) {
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
Excellent. Nice work. Especially the `useAiType`

Feedback: space between `if` and opening bracket.

Issue #987, opened 2020-06-15
Create Mod, upon second enter, doesn't clear the BasicSword text field


direwolf420:
`_basicSword.SetText("");` is missing from OnInitialize here: https://github.com/tModLoader/tModLoader/blob/f913a7ec6f7e9f68fb72f121dc72dfe246789325/patches/tModLoader/Terraria.ModLoader.UI/UICreateMod.cs#L123

PR #992, opened 2020-06-21
Fix issue with BasicSword not clearing during Mod Creation

patterson-joseph:
Resolves tModLoader/tModLoader#987

### What is the bug?
The BasicSword input does not clear properly from the Mod Sources > Create Mod screen

### How did you fix the bug?
Added a SetText("") reset to the OnActivate() function that matches how we are resetting the other inputs on this screen.

None known. This is a simple change that matches existing patterns.


Jofairden:
Thanks.

PR #1010, opened 2020-07-14
Replace SetupStartInventory

Trivaxy:
### Overview

The Forbidden set consumes mana but never fires any of the mana-related hooks in tML. This PR fixes that.
Main purpose of this PR is to deprecate `ModPlayer.SetupStartInventory` in favor of two new hooks: `ModPlayer.AddStartingItems` and `ModPlayer.ModifyStartingInventory`.

The former is used by mods to return an `IEnumerable` containing a collection of items the modder wants to add to the player inventory either upon their creation or if the player dies in mediumcore.
The latter is used to modify the "pool" of items that vanilla and other mods want to add, by accessing a dictionary where the key is the mod's internal name and the value is the list of items it wants to add. Vanilla also exists in this pool under the key "Terraria".


JavidPack:
A locked door is a door that requires a key, not just any closed door. We'll need to implement methods similar to the locked chest methods to actually implement this

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3692,6 +3692,15 @@
  					if (buffType[i] >= 170 && buffType[i] <= 172)
  						DelBuff(i);
  				}
+@@ -38234,7 +_,7 @@
+ 			flag &= (list.Count <= 1);
+ 			flag = (flag && flag2);
+ 			if (flag) {
+-				flag = CheckMana(20, pay: true);
++				flag = CheckMana(new Item(), 20, pay: true);
```
Chicken-Bones, line 3700: We should do something similar to what we did with `GetOtherDamage` in #654, where it doesn't necessarily come from an item. At the least, a string should be used to identify the 'source' of the mana usage.

patches/tModLoader/Terraria/ModLoader/PlayerHooks.cs
```diff
@@ -76,22 +76,54 @@ private class HookList
...
+				items.Add(item);
+				item = new Item();
+				item.SetDefaults(ItemID.CopperPickaxe);
+				item.Prefix(-1);
+				items.Add(item);
+				item = new Item();
+				item.SetDefaults(ItemID.CopperAxe);
+				item.Prefix(-1);
+				items.Add(item);
+			}
```
Jofairden: Did you copy this from vanilla? Isn't it nicer to just have the vanilla method populate this list or???

Trivaxy: the way tML does it is to discard vanilla's inventory setup anyway, and doing it on its own (ie copper tools for non-journey). I'm all for manually converting the array to a list if needed, just thought i shouldn't deviate from the way tML did it before

Trivaxy: P.S I have no idea why this change ended up in this PR :/

Jofairden: `the way tML does it is to discard vanilla's inventory setup anyway`
I'm not sure if we should continue doing that though ,it sounds dumb

patches/tModLoader/Terraria/GameContent/UI/States/UICharacterCreation.cs.patch
```diff
@@ -8,11 +8,14 @@
...
-@@ -1053,6 +_,7 @@
+@@ -1053,6 +_,10 @@
  
  			_player.savedPerPlayerFieldsThatArentInThePlayerClass = new Player.SavedPlayerDataWithAnnoyingRules();
  			CreativePowerManager.Instance.ResetDataForNewPlayer(_player);
-+			PlayerHooks.SetStartInventory(_player, PlayerHooks.SetupStartInventory(_player));
++
++			var inventoryAdditions = PlayerHooks.AddStartingItems(_player);
++			PlayerHooks.ModifyStartingInventory(_player, inventoryAdditions);
++			PlayerHooks.SetStartInventory(_player, inventoryAdditions.Values.SelectMany(i => i).ToList());
```
Jofairden, line 18: Can't you just do `inventoryAdditions.Values.ToList()`?
Oh wait, I guess SelectMany is flattening it...

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -927,5 +916,23 @@ public class ModPlayer
 		/// <param name="price">The price the player paid in copper coins.</param>
 		public virtual void PostNurseHeal(NPC nurse, int health, bool removeDebuffs, int price) {
 		}
+
+		/// <summary>
+		/// Called when the player is created in the menu.
+		/// You can use this method to add items to the player's starting inventory, as well as their inventory when they respawn in mediumcore.
+		/// </summary>
+		/// <param name="mediumCoreDeath">Whether you are setting up a mediumcore player's inventory after their death.</param>
+		/// <returns>An enumerable of the items you want to add. If you want to add nothing, return an empty enumerable.</returns>
```
Jofairden, line 925: Maybe you can add `return an Enumerable.Empty<Item>()`

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -927,5 +916,23 @@ public class ModPlayer
...
+		public virtual IEnumerable<Item> AddStartingItems(bool mediumCoreDeath) {
+			return Enumerable.Empty<Item>();
+		}
+
+		/// <summary>
+		/// Allows you to modify the items that will be added to the player's inventory. Useful if you want to stop vanilla or other mods from adding an item.
```
Jofairden, line 931: Maybe you can add: You can access a mod's items by using the mod's internal name as the indexer: `additions["ModName"]`. To access vanilla items you can use "Terraria" as the indexer.

ExampleMod/ExampleInventoryPlayer.cs
```diff
@@ -0,0 +1,46 @@
...
+using Terraria.ModLoader;
+using static Terraria.ModLoader.ModContent;
+
+namespace ExampleMod
+{
+	public class ExampleInventoryPlayer : ModPlayer
+	{
+		// AddStartingItems is a method you can use to add items to the player's starting inventory.
+		// It is also called when the player dies a mediumcore death
+		// Simply create an enumerable (think a collection) of the items you want to add.
```
Jofairden, line 15:
```suggestion
		// Return an enumerable with the items you want to add to the inventory.
```

patches/tModLoader/Terraria/GameContent/UI/States/UICharacterCreation.cs.patch
Trivaxy: Yes, it flattens the list

Mirsario (approved):

ExampleMod/Common/Players/ExampleInventoryPlayer.cs
```diff
@@ -0,0 +1,46 @@
...
+			if (mediumCoreDeath) {
+				return new[] {
+					new Item(ItemID.HealingPotion)
+				};
+			}
+
+			return new[] {
+				testItem,
+				otherItem
+			};
```
Chicken-Bones, line 33: Please combine unnecessary variables, especially to declare them and then return without using them. Just `return new[] {new Item(...), new Item(...) {stack = 256 }}`

Chicken-Bones (changes requested):
Looks great, just one more little code cleanup and then green button for you

PR #1011, opened 2020-07-14
tooltips but again

ThomasSpeedrunner:
I also fixed the issue where you would always buy two of something

there's a really dumb thing here where Journey dupe menu tooltips just decide to perish, I spent most of today trying to diagnose the issue but eventually I got tired of staring literally the entire tooltip method down the barrel so hey maybe someone smarter than me can diagnose this issue

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -1010,348 +1010,160 @@
  			if (diff == 1)
  				baseColor = new Microsoft.Xna.Framework.Color((byte)((float)(int)mcColor.R * num2), (byte)((float)(int)mcColor.G * num2), (byte)((float)(int)mcColor.B * num2), mouseTextColor);
  
-@@ -14837,107 +_,114 @@
- 
- 			int num2 = 30;
- 			int numLines = 1;
--			string[] array = new string[num2];
-+			string[] tooltipLines = new string[num2];
- 			bool[] array2 = new bool[num2];
- 			bool[] array3 = new bool[num2];
- 			for (int i = 0; i < num2; i++) {
+@@ -14844,8 +_,9 @@
  				array2[i] = false;
  				array3[i] = false;
  			}
 +			string[] tooltipNames = new string[num2];
  
 -			MouseText_DrawItemTooltip_GetLinesInfo(HoverItem, ref yoyoLogo, ref researchLine, knockBack, ref numLines, array, array2, array3);
-+			MouseText_DrawItemTooltip_GetLinesInfo(HoverItem, ref yoyoLogo, ref researchLine, knockBack, ref numLines, tooltipLines, array2, array3, tooltipNames);
++			MouseText_DrawItemTooltip_GetLinesInfo(HoverItem, ref yoyoLogo, ref researchLine, knockBack, ref numLines, array, array2, array3, tooltipNames);
  			float num3 = (float)(int)mouseTextColor / 255f;
--			float num4 = num3;
-+			float colorThing = num3;
+ 			float num4 = num3;
  			int a = mouseTextColor;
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1010,73 +1010,348 @@
...
- 						array[numLines] = Lang.tip[49].Value + " " + text;
+-					CustomCurrencyManager.GetPriceText(hoverItem.shopSpecialCurrency, array, ref numLines, num5);
++					CustomCurrencyManager.GetPriceText(hoverItem.shopSpecialCurrency, tooltipLines, ref numLines, num5);
+-					color = new Microsoft.Xna.Framework.Color((byte)(255f * num4), (byte)(255f * num4), (byte)(255f * num4), a);
++					color = new Microsoft.Xna.Framework.Color((byte)(255f * colorThing), (byte)(255f * colorThing), (byte)(255f * colorThing), a);
+ 				}
+ 				else if (num5 > 0) {
+ 					string text = "";
+-					int num6 = 0;
++					int costPlatinum = 0;
```
Jofairden, line 1046: You're gonna have to undo these because they make the patch necessarily big
We generally want to keep patches small, and renaming num variables is a bit irrelevant since they pretty much can change every update, this counts for literally all renames you did (it's also impossible for me to find now what exactly you did)

ThomasSpeedrunner: it has been done now, should make what I actually did much more apparent
does make the method a bit less readable but w/e

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1029,12 +1029,10 @@
  					CustomCurrencyManager.GetPriceText(hoverItem.shopSpecialCurrency, array, ref numLines, num5);
  					color = new Microsoft.Xna.Framework.Color((byte)(255f * num4), (byte)(255f * num4), (byte)(255f * num4), a);
  				}
-@@ -14912,7 +_,8 @@
- 						array[numLines] = Lang.tip[49].Value + " " + text;
+@@ -14913,6 +_,7 @@
  					else
  						array[numLines] = Lang.tip[50].Value + " " + text;
--
-+
```
Jofairden, line 1037: I still see some redundant patch changes here.
Just in general we try to make sure to only edit the necessary stuff and keeping patches small is a necessity

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1050,23 +1048,55 @@
  			}
  
  			Vector2 zero = Vector2.Zero;
-+			Color?[] overrideColor;
++			Microsoft.Xna.Framework.Color?[] overrideColor;
```
Jofairden, line 1051: Idem

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1076,17 +1106,43 @@
...
  			}
  		}
  
 -		public static void MouseText_DrawItemTooltip_GetLinesInfo(Item item, ref int yoyoLogo, ref int researchLine, float oldKB, ref int numLines, string[] toolTipLine, bool[] preFixLine, bool[] badPreFixLine) {
 +		public static void MouseText_DrawItemTooltip_GetLinesInfo(Item item, ref int yoyoLogo, ref int researchLine, float oldKB, ref int numLines, string[] toolTipLine, bool[] preFixLine, bool[] badPreFixLine, string[] toolTipNames) {
  			toolTipLine[0] = item.HoverName;
 +			toolTipNames[0] = "ItemName";
--			if (item.favorited) {
-+			if (item.favorited)
-+			{
```
Jofairden, line 1089: We use/enforce braces on the same line

PR #1018, opened 2020-07-16
Merge Recipe and ModRecipe

Itorius:


Jofairden:
@Mirsario  @Chicken-Bones  opinions requested
I'm personally not a fan of 'Register', odd wording for me. Also not a fan of arrow notation for non one-liners.
Overall though, I like the syntax.

Jofairden:
`AddIngredient<ExampleItem>(10)` would be even nicer

Mirsario:
> 
> 
> @Mirsario @Chicken-Bones opinions requested
> I'm personally not a fan of 'Register', odd wording for me. Also not a fan of arrow notation for non one-liners.
> Overall though, I like the syntax.

Registering is what it does.

> 
> 
> `AddIngredient<ExampleItem>(10)` would be even nicer

I already added it yesterday or so, and all basic recipe changes were merged into 1.4 branch. This PR is about merging ModRecipe into Recipe, and there's plenty to question here.

Mirsario:
I am extremely all-around unhappy with pretty much everything here, **especially** Recipe.cs.patch. All it needed to have is access modifier changes. You could easily set up data for new members from tML in a single method, based on vanilla data. But instead, there's this mess where nearly every line is rewritten, and there are even some methods removed for no reason, even though making them private or internal could effectively achieve the same goal without any ugly and unneeded patches that will have to be maintained, mind you.

Also, *why* are we even merging ModRecipe into Recipe again? It's all too hasty.

Itorius:
> I am extremely all-around unhappy with pretty much everything here, **especially** Recipe.cs.patch. All it needed to have is access modifier changes. You could easily set up data for new members from tML in a single method, based on vanilla data. But instead, there's this mess where nearly every line is rewritten, and there are even some methods removed for no reason, even though making them private or internal could effectively achieve the same goal without any ugly and unneeded patches (that will have to be maintained, mind you).
> 
> Also, _why_ are we even merging ModRecipe into Recipe again? It's all too hasty.

Recipe.cs.patch has minimal amount of changes to make it work. CB agreed the useX methods can be removed as they're completely obsolete because they got replaced by the RecipeGroup system. And I would argue this is easier to maintain than having _essentially_ 2 separate systems (vanilla & tML).

tl;dr - it is cleaner, more powerful, and more easily maintainable

Jofairden:
_I don't really know why vanilla has to be changed_
We have RecipeFinder and RecipeEditor to change any recipe, including vanilla iirc.
I think we should revise those instead if we're doing this anyways.
We've always been extending vanilla by "ModRecipe", which has worked fine
I get the incentive, I'm just still wondering if it's 'worth' changing. Everything has the Mod notation, but we'd be changing recipes 'just because'

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -62,16 +66,277 @@ public sealed class Condition : ICondition
...
+			RequireGroup(recipeGroupId);
+
+			return this;
+		}
+
+		/// <summary>
+		/// Adds a recipe group ingredient to this recipe with the given RecipeGroup.
+		/// </summary>
+		/// <param name="recipeGroup">The RecipeGroup.</param>
+		public Recipe AddRecipeGroup(RecipeGroup recipeGroup, int stack = 1) {
```
Chicken-Bones, line 207: I see this parameter version slipped in here, but is it necessary? Is there a use case for it?

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -62,16 +66,277 @@ public sealed class Condition : ICondition
...
 			}
 
-			public bool RecipeAvailable(ModRecipe recipe) => Predicate(recipe);
+			public bool RecipeAvailable(Recipe recipe) => Predicate(recipe);
+		}
+
+		public static class ConsumptionRules
+		{
+			/// <summary> Gives 1/3 chance for every ingredient to not be consumed, if used at an alchemy table. (!) This behavior is already automatically given to all items that can be made at a placed bottle tile. </summary>
+			public static int Alchemy(Recipe _, int type, int amount) {
```
Chicken-Bones, line 84: good

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -9,11 +9,78 @@
...
++		public Recipe(Mod mod) {
++			Mod = mod;
++			Conditions = new List<Condition>();
++			
++			for (int i = 0; i < maxRequirements; i++) {
++				requiredItem[i] = new Item();
++				requiredTile[i] = -1;
++				acceptedGroups[i] = -1;
++			}
++		}
```
Chicken-Bones: reordering makes patches hard

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -31,104 +105,343 @@
...
+-			for (int i = 0; i < maxRequirements; i++) {
+-				requiredItem[i] = new Item();
+-				requiredTile[i] = -1;
+-				acceptedGroups[i] = -1;
+-			}
+-		}
+-
++	
+ 		public void Create() {
+-			Item[] array = null;
```
Chicken-Bones: this is unmaintainable, keep as close to the vanilla code as possible

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -148,3 +473,72 @@
...
++			if (currentRecipe.needSnowBiome) currentRecipe.AddCondition(Condition.InSnow);
++			if (currentRecipe.needWater) currentRecipe.AddCondition(Condition.NearWater);
++			if (currentRecipe.needLava) currentRecipe.AddCondition(Condition.NearLava);
++			if (currentRecipe.needHoney) currentRecipe.AddCondition(Condition.NearHoney);
++
++			if (currentRecipe.anyFragment) currentRecipe.RequireGroup(RecipeGroupID.Fragment);
++			if (currentRecipe.anySand) currentRecipe.RequireGroup(RecipeGroupID.Sand);
++			if (currentRecipe.anyWood) currentRecipe.RequireGroup(RecipeGroupID.Wood);
++			if (currentRecipe.anyIronBar) currentRecipe.RequireGroup(RecipeGroupID.IronBar);
++			if (currentRecipe.anyPressurePlate) currentRecipe.RequireGroup(RecipeGroupID.PressurePlate);
```
Chicken-Bones, line 309: this is the good stuff

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3700,3 +3700,24 @@
  				if (buffType[i] >= type && buffType[i] < type + 3)
  					DelBuff(i);
  			}
+@@ -38411,6 +_,20 @@
+ 
+ 			if (list != null)
+ 				TouchedTiles = list;
++		}
++
++		public Item[] GetPlayerChest() {
```
Mirsario, line 3709: This should either be in Player.TML.cs, an extension method, or in an util class.

patches/tModLoader/Terraria/RecipeGroup.cs.patch
```diff
@@ -1,24 +1,33 @@
 --- src/Terraria/Terraria/RecipeGroup.cs
 +++ src/tModLoader/Terraria/RecipeGroup.cs
-@@ -8,6 +_,7 @@
+@@ -5,13 +_,15 @@
+ {
+ 	public class RecipeGroup
```
Mirsario, line 5: Better to have the new members moved to RecipeGroup.TML.cs, but you can do this later.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -148,3 +220,80 @@
...
+ 		private static void AddSolarFurniture() {
+ 			currentRecipe.createItem.SetDefaults(4229);
+ 			currentRecipe.createItem.stack = 10;
+@@ -13962,8 +_,9 @@
+ 				}
+ 			}
+ 		}
+-
+-		public void SetIngridients(params int[] ingridients) {
++
```
Mirsario, line 246: Whitespace changes again.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -148,3 +220,80 @@
...
+-					return 0;
+-
+-				return 1;
+-			}
+-
+-			return 0;
+-		}
++		#endregion
++
++		public static int GetRequiredTileStyle(int type) => type == TileID.DemonAltar && WorldGen.crimson ? 1 : 0;
```
Mirsario, line 297: Unnecessary cleanup.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -31,104 +95,100 @@
...
+ 					}
+@@ -401,8 +_,8 @@
+ 						if (!(num6 && flag3 && flag4 && flag5 && flag6))
  							flag = false;
  					}
- 
+-
 -					if (flag) {
++					
 +					if (flag && RecipeHooks.RecipeAvailable(Main.recipe[n])) {
```
Mirsario, line 187: Whitespace changes is a tiresome phrase to repeat.

patches/tModLoader/Terraria/ModLoader/RecipeHooks.cs
```diff
@@ -53,17 +53,7 @@ public static class RecipeHooks
...
-			}
-
-			foreach (GlobalRecipe globalRecipe in globalRecipes) {
-				if (!globalRecipe.RecipeAvailable(recipe)) {
-					return false;
-				}
-			}
-
-			return true;
+			return recipe.Conditions.All(c => c.RecipeAvailable(recipe)) && globalRecipes.All(globalRecipe => globalRecipe.RecipeAvailable(recipe));
```
Mirsario, line 56: Use expression method body.

patches/tModLoader/Terraria/RecipeGroup.cs.patch
Chicken-Bones: not enough in this case, happy for them to be here.

patches/tModLoader/Terraria/Player.cs.patch
Chicken-Bones: entirely agree, just gonna remove it for now.

PR #1023, opened 2020-07-20
Better Autoloading 2

Rartrin:
Changes to my last autoloading system:
- `AutoloadAttribute` now allows you to specify a `ModSide` which will control when a type can be autoloaded (client side, server side, or both).
- The base tML ModTypes now expose the implemented `Load(Mod)` method. This unifies the loading and removes the `Load()` and `AddInstance()` methods. Anyone wanting to overload `Load(Mod)` will need to call `base.Load(mod)` to ensure their ModType is loaded.

Remaining things todo/discuss before merging:
- [ ] Make ModConfig use ILoadable?
- [ ] Removing mod.Properties (currently `Properties.Autoload` does nothing, the other autoloads still do though).
  - [ ] Should ModGore and ModSound use ILoadable?
- [ ] Ideas for unifying duplicate-name checking. Perhaps using the IModType interface.
- [ ] Moving `ModOrganizer.LoadSide` to a more generalize location as it is now needed during autoloading. There may be other places that could use it as well.

### DIFF
--- a/ExampleMod/Content/Items/Accessories/WaspNest.cs
+++ b/ExampleMod/Content/Items/Accessories/WaspNest.cs
@@ -11,8 +11,7 @@ namespace ExampleMod.Content.Items.Accessories
 	public class WaspNest : ModItem
 	{
 		// Only gets run once per type
-		public override void Load(Mod mod) {
-			base.Load(mod);
+		public override void Load() {
 			IL.Terraria.Player.beeType += HookBeeType;
 		}
 
--- a/ExampleMod/Content/Items/Accessories/WaspNest.cs
+++ b/ExampleMod/Content/Items/Accessories/WaspNest.cs
@@ -11,7 +11,8 @@ namespace ExampleMod.Content.Items.Accessories
 	public class WaspNest : ModItem
 	{
 		// Only gets run once per type
-		public override void Load() {
+		public override void Load(Mod mod) {
+			base.Load(mod);
 			IL.Terraria.Player.beeType += HookBeeType;
 		}
 
--- a/ExampleMod/Content/Items/Tools/ExampleHook.cs
+++ b/ExampleMod/Content/Items/Tools/ExampleHook.cs
@@ -1,6 +1,7 @@
 ﻿using ExampleMod.Content.Tiles.Furniture;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
+using ReLogic.Content;
 using Terraria;
 using Terraria.ID;
 using Terraria.ModLoader;
@@ -100,7 +101,7 @@ namespace ExampleMod.Content.Items.Tools
 			speed = 10;
 		}
 
-		private Texture2D chainTexture = GetTexture("ExampleMod/Content/Items/Tools/ExampleHookChain").Value;
+		private Asset<Texture2D> chainTexture = GetTexture("ExampleMod/Content/Items/Tools/ExampleHookChain");
 
 		public override bool PreDraw(SpriteBatch spriteBatch, Color lightColor) {
 			Vector2 playerCenter = Main.player[projectile.owner].MountedCenter;
@@ -117,8 +118,8 @@ namespace ExampleMod.Content.Items.Tools
 				Color drawColor = lightColor;
 
 				//Draw chain
-				spriteBatch.Draw(chainTexture, new Vector2(center.X - Main.screenPosition.X, center.Y - Main.screenPosition.Y),
-					new Rectangle(0, 0, chainTexture.Width, chainTexture.Height), drawColor, projRotation,
+				spriteBatch.Draw(chainTexture.Value, new Vector2(center.X - Main.screenPosition.X, center.Y - Main.screenPosition.Y),
+					new Rectangle(0, 0, chainTexture.Width(), chainTexture.Height()), drawColor, projRotation,
 					chainTexture.Size() * 0.5f, 1f, SpriteEffects.None, 0f);
 			}
 
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -134,6 +134,20 @@
  		public static string worldName = "";
  		public static int worldID;
  		public static int background;
+@@ -820,7 +_,12 @@
+ 		public static float invAlpha = 1f;
+ 		public static float invDir = 1f;
+ 		[ThreadStatic]
++		private static UnifiedRandom _rand;
+-		public static UnifiedRandom rand;
++		public static UnifiedRandom rand
++		{
++			get => _rand ??= new UnifiedRandom((int)DateTime.Now.Ticks);
++			set => _rand = value;
++		}
+ 		public static bool allChestStackHover;
+ 		public static bool inventorySortMouseOver;
+ 		public static float GraveyardVisualIntensity;
 @@ -837,7 +_,7 @@
  		public static AudioEngine engine;
  		public static SoundBank soundBank;
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -134,20 +134,6 @@
  		public static string worldName = "";
  		public static int worldID;
  		public static int background;
-@@ -820,7 +_,12 @@
- 		public static float invAlpha = 1f;
- 		public static float invDir = 1f;
- 		[ThreadStatic]
-+		private static UnifiedRandom _rand;
--		public static UnifiedRandom rand;
-+		public static UnifiedRandom rand
-+		{
-+			get => _rand ??= new UnifiedRandom((int)DateTime.Now.Ticks);
-+			set => _rand = value;
-+		}
- 		public static bool allChestStackHover;
- 		public static bool inventorySortMouseOver;
- 		public static float GraveyardVisualIntensity;
 @@ -837,7 +_,7 @@
  		public static AudioEngine engine;
  		public static SoundBank soundBank;
--- a/patches/tModLoader/Terraria/ModLoader/AutoloadAttribute.cs
+++ b/patches/tModLoader/Terraria/ModLoader/AutoloadAttribute.cs
@@ -8,19 +8,26 @@ namespace Terraria.ModLoader
 	/// True to always autoload, false to never autoload, null to use mod default.
 	/// </summary>
 	[AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = true, Inherited = true)]
-	public class AutoloadAttribute : Attribute
+	public sealed class AutoloadAttribute : Attribute
 	{
+		private static readonly AutoloadAttribute Default = new AutoloadAttribute();
+
 		public readonly bool Value;
+		public ModSide Side{get;set;} = ModSide.Both;
 
-		public AutoloadAttribute(bool value = true) { Value = value; }
+		public bool NeedsAutoloading => Value && Core.ModOrganizer.LoadSide(Side);
 
-		public static bool? GetValue(Type type) {
+		public AutoloadAttribute(bool value = true) {
+			Value = value;
+		}
+
+		public static AutoloadAttribute GetValue(Type type) {
 			//Get all AutoloadAttributes on the type.
 			object[] all = type.GetCustomAttributes(typeof(AutoloadAttribute), true);
 			//The first should be the most derived attribute.
 			var mostDerived = (AutoloadAttribute)all.FirstOrDefault();
 			//If there were no declarations, then return null.
-			return mostDerived?.Value;
+			return mostDerived??Default;
 		}
 	}
 }
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -70,7 +70,8 @@ namespace Terraria.ModLoader.Core
 				.Union(Directory.GetFiles(ModLoader.ModPath, "temporaryDownload.tmod", SearchOption.TopDirectoryOnly)); // Old tML remnant
 		}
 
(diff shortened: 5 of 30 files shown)


ExampleMod/Content/Items/Accessories/WaspNest.cs
```diff
@@ -11,7 +11,8 @@ namespace ExampleMod.Content.Items.Accessories
 	public class WaspNest : ModItem
 	{
 		// Only gets run once per type
-		public override void Load() {
+		public override void Load(Mod mod) {
+			base.Load(mod);
 			IL.Terraria.Player.beeType += HookBeeType;
```
Chicken-Bones, line 16: This just feels kinda good

ExampleMod/Content/Items/Tools/ExampleHook.cs
```diff
@@ -100,7 +101,7 @@ internal class ExampleHookProjectile : ModProjectile
 			speed = 10;
 		}
 
-		private Texture2D chainTexture = GetTexture("ExampleMod/Content/Items/Tools/ExampleHookChain").Value;
+		private Asset<Texture2D> chainTexture = GetTexture("ExampleMod/Content/Items/Tools/ExampleHookChain");
```
Chicken-Bones, line 104: someone teach me how to use this asset system properly someday

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -134,6 +134,20 @@
...
+ 		public static float invAlpha = 1f;
+ 		public static float invDir = 1f;
+ 		[ThreadStatic]
++		private static UnifiedRandom _rand;
+-		public static UnifiedRandom rand;
++		public static UnifiedRandom rand
++		{
++			get => _rand ??= new UnifiedRandom((int)DateTime.Now.Ticks);
++			set => _rand = value;
++		}
```
Chicken-Bones, line 147: This isn't autoloading. Also why are we changing this? Synchronisation?

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -70,7 +70,8 @@ internal static class ModOrganizer
 				.Union(Directory.GetFiles(ModLoader.ModPath, "temporaryDownload.tmod", SearchOption.TopDirectoryOnly)); // Old tML remnant
 		}
 
-		private static bool LoadSide(ModSide side) => side != (Main.dedServ ? ModSide.Client : ModSide.Server);
+		//TODO: Find a good common place for this code.
+		internal static bool LoadSide(ModSide side) => side != (Main.dedServ ? ModSide.Client : ModSide.Server);
```
Chicken-Bones, line 74: yeah, lets move `LoadSide` to a public class, how about `AutoloadAttribute`?? or like, top level loading, maybe `Mod`?

patches/tModLoader/Terraria/ModLoader/ModTileEntity.cs
```diff
@@ -50,7 +50,7 @@ public abstract class ModTileEntity : TileEntity, IModType
 			return type >= numVanilla && type < nextTileEntity ? tileEntities[type - numVanilla] : null;
 		}
 
-		internal static void Unload() {
+		internal static void UnloadAll() {
```
Chicken-Bones, line 53: Sure I guess

patches/tModLoader/Terraria/ModLoader/AutoloadAttribute.cs
```diff
@@ -8,19 +8,26 @@ namespace Terraria.ModLoader
 	/// True to always autoload, false to never autoload, null to use mod default.
 	/// </summary>
 	[AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = true, Inherited = true)]
-	public class AutoloadAttribute : Attribute
+	public sealed class AutoloadAttribute : Attribute
```
Jofairden, line 11: Are we sure we want this sealed?

ExampleMod/Content/Items/Accessories/WaspNest.cs
Jofairden: Does every modder have to do this in every Load hook? I mean I guess it's fine, but we gotta make sure to document that, we haven't /really/ had the modders use proper inheritance calls yet

ExampleMod/Content/Items/Tools/ExampleHook.cs
Jofairden: So much better

ExampleMod/Content/Items/Tools/ExampleHook.cs
```diff
@@ -117,8 +118,8 @@ internal class ExampleHookProjectile : ModProjectile
 				Color drawColor = lightColor;
 
 				//Draw chain
-				spriteBatch.Draw(chainTexture, new Vector2(center.X - Main.screenPosition.X, center.Y - Main.screenPosition.Y),
-					new Rectangle(0, 0, chainTexture.Width, chainTexture.Height), drawColor, projRotation,
+				spriteBatch.Draw(chainTexture.Value, new Vector2(center.X - Main.screenPosition.X, center.Y - Main.screenPosition.Y),
```
Jofairden, line 121: Is this intended? (Is Asset from TML now or Vanilla?) I'd kind of expect draw to be able to accept an Asset.. huh..

ExampleMod/Content/Items/Tools/ExampleHook.cs
```diff
@@ -117,8 +118,8 @@ internal class ExampleHookProjectile : ModProjectile
...
+					new Rectangle(0, 0, chainTexture.Width(), chainTexture.Height()), drawColor, projRotation,
```
Jofairden, line 122: Are these calculated when called? I remember them not being methods on Texture2D

patches/tModLoader/Terraria/Main.cs.patch
Jofairden: I guess it can help ensuring the random generator is always there?

patches/tModLoader/Terraria/ModLoader/ModDust.cs
```diff
@@ -92,18 +93,14 @@ public class ModDust:ModTexturedType
...
 
-		public sealed override void Load()
-		{
-			base.Load();
-			Texture2D = !string.IsNullOrEmpty(Texture) ? ModContent.GetTexture(Texture).Value : TextureAssets.Dust.Value;
-		}
-
-		internal sealed override void AddInstance() {
+		public override void Load(Mod mod) {
+			base.Load(mod);
```
Jofairden, line 97: Imagine not calling base.Load now on your overriding class...... oof (this is pretty vulnerable)

patches/tModLoader/Terraria/Main.cs.patch
Mirsario: This is a fix that should've been on 1.4 branch, and I'll recommit it there right now because everything is essentially broken without this.

ExampleMod/Content/Items/Accessories/WaspNest.cs
Chicken-Bones: Yes, but only if they want to put their IL hooks there. The use cases for overriding are limited to modders who will know C# anyway

patches/tModLoader/Terraria/ModLoader/AutoloadAttribute.cs
Chicken-Bones: 99%, it's an attribute, it doesn't make sense to inherit from

Jofairden: What if you want some custom logic to follow the auto loading?

ExampleMod/Content/Items/Accessories/WaspNest.cs
Mirsario: Base calls definitely have to go away. 

> It's a noob trap, and makes it impossible to not call base method in a class that derives from something that derives from a ModX.

ExampleMod/Content/Items/Tools/ExampleHook.cs
Mirsario: Should've just commited this directly outside the PR, but whatever.

patches/tModLoader/Terraria/ModLoader/ModDust.cs
Mirsario: Yeah, it may result in a null ref inside tML's code, which even intermediate modders won't be able to figure out or guess the reason of.

patches/tModLoader/Terraria/ModLoader/ModType.cs
```diff
@@ -15,20 +15,10 @@ public abstract class ModType : IModType
...
 
-		void ILoadable.Load(Mod mod) {
-			Mod = mod;
-			Load();
-			AddInstance();
-		}
-
-		public virtual void Load(){}
+		/// <summary>Make sure to call base.Mod(mod) to load in your type.</summary>
+		public virtual void Load(Mod mod) => Mod = mod;
```
Mirsario, line 19: Too basic of a thing to require base calls for.

Chicken-Bones:
Summarising the debate over `base.Load(mod);` We have 2 options for `ModType` when implementing `ILoadable.Load` (which is how it gets the `Mod` reference to store in its property. This affects the implementation for subclasses `ModX` and also sub-subclasses (actual content instances).

Option 1a
```cs
public abstract class ModType : IModType
{
	void ILoadable.Load(Mod mod) {
		Mod = mod;
		Load();
		AddInstance();
	}

	protected virtual void Load(){}
	protected virtual void AddInstance(){}
}

public class ModItem:ModTexturedType
{
	protected sealed override void AddInstance() {
		...
		Mod.items[Name] = this;
		ItemLoader.items.Add(this);
		ContentInstance.Register(this);
		...
	}
}

public class WaspNest:ModItem
{
	protected override void Load() {
		IL.Terraria.Player.beeType += HookBeeType;
	}
}
```

Option 1b
```cs
public abstract class ModType : IModType
{
	void ILoadable.Load(Mod mod) {
		Mod = mod;
		Load();
	}

	protected virtual void Load(){}
}

public class ModItem:ModTexturedType
{
	protected sealed override void Load() {
		...
		Mod.items[Name] = this;
		ItemLoader.items.Add(this);
		ContentInstance.Register(this);
		...
	}
}

public class WaspNest:ModItem
{
		IL.Terraria.Player.beeType += HookBeeType;
	}
}
```

Option 2a
```cs
public abstract class ModType : IModType
{
	public override void Load(Mod mod) {
		Mod = mod;
	}
}

public class ModItem:ModTexturedType
{
	public override void Load(Mod mod) {
		base.Load(mod);
		...
		Mod.items[Name] = this;
		ItemLoader.items.Add(this);
		ContentInstance.Register(this);
		...
	}
}

public class WaspNest:ModItem
{
	public override void Load(Mod mod) {
		base.Load(mod);
		IL.Terraria.Player.beeType += HookBeeType;
	}
}
```

Option 2b
```cs
public abstract class ModType : IModType
{
	public override void Load(Mod mod) {
		Mod = mod;
	}
}

public class ModItem:ModTexturedType
{
	public override void Load(Mod mod) {
		base.Load(mod);
		...
		Mod.items[Name] = this;
		ItemLoader.items.Add(this);
		ContentInstance.Register(this);
		...
	}
}

public class WaspNest:ModItem
{
		IL.Terraria.Player.beeType += HookBeeType;
	}
}
```


Prior discussion has been heavily skewed by the use of `Load` to add an ILHook in `WaspNest`. 
While this is a valid use case, and it feels natural to put ILHooks in `Load`, this is a new option, not previously available.
Previously, modders could use `SetStaticDefaults` to add ILHooks for items, and they still can. I've included this variant as options labelled b. 

Comparison of pros and cons:
1a
+ no base calls, impossible to stuff up
+ content instances (WaspNest) can override `Load` if they want to
- actual `Load` method changed to `AddInstance`, conflicting with general `ILoadable` design
1b
+ no base calls, impossible to stuff up
+ load method still called `Load`, unlikely to cause confusion, but slightly weird code path
- content instances cannot override `Load`
2
- base calls required in `ModX.Load`, only affects tML, or people adding their own subclasses of `ModType`
+ single OOP consistent `Load` method, `ILoadable.Load`
+ content instances _can_ override `Load` if they want to (but shouldn't need to)
- content instances must call `base.Load` if they override `Load`

Opinion below:

Which of these you prefer depends on how big of a pro you think the pros are, and how big of a con the cons are.
I personally vote for 2b, because it's clearer and more maintainable for us as experienced tML developers, and provides a good standard model for other developers to implement their own ModX classes, or similar systems as they grow with C#.

The biggest argument against #2 is the requirement to call `base.Load` is a noob trap. I agree, it is a noob trap, and I'm in favour of staying away from base calls in all hooks used by new and inexperienced modders. My only argument here is that there is no 'noob' use case for overriding Load in content classes. The only implementors should be people making their own ModX types, or people who have some other system to which they're binding content during Loading (or maybe ILHooks if you really want them in Load for some reason). All 3 of these are 'advanced' use cases, and unlikely to trap noobs.

Jofairden:
Option 1a feels right to me because that's how I do it in EMM but I get your points for 2 and they are valid
My opinion right now on this is that if we can make sure the API does _at least_ the things that are required (e.g. setting the Mod field) without the modder being able to screw up a mandatory process just by using one of the methods that's great, but yes this does result in 'magic code' that is sometimes heavily frowned upon. I think the bigger question is: what people use TML? Most modders, _are noobs_ (in terms of programming). But they can learn, even noobs, that they have to call the base method.

As for Load vs SetStaticDefaults, I feel like Load makes more sense to me for IL hooks since I'd normally be adding these while my mod loads as well. SetStaticDefaults really just makes me think about stats for the item.

Mirsario:
1a all the way. @Chicken-Bones you're saying that overriding Load is something 'advanced' on ModX classes. But it's not. It's perfectly usable as a static constructor without their issues. And if I implement ModSystems, Load() there will be the most important hook. Like it is on the Mod class, because it's the same use case. None of you want that to be called Init().

PR #1028, opened 2020-07-27
Added steam cloud quota check

direwolf420:
**This is an alternate version of #1045**

### What is the bug?
On Steam, because Terraria and tModLoader share the same cloud space, people (usually with a full Terraria steam cloud) want to move their tModLoader files to the cloud, which leads to them being deleted instead (usually after restarting the game)

### How did you fix the bug?
Prevent pressing the "move to cloud" button when the quota exceeds 90%

Currently this recalculates the quota on mouseover, aka every tick. This could be solved by caching the last value that was calculated (once when the game starts, and then once every "move to/off cloud" press)

### DIFF
--- a/patches/tModLoader/Terraria.GameContent.UI.Elements/UICharacterListItem.cs.patch
+++ b/patches/tModLoader/Terraria.GameContent.UI.Elements/UICharacterListItem.cs.patch
@@ -114,7 +114,15 @@
  			_deleteButtonLabel.Top.Set(-3f, 0f);
  			Append(_deleteButtonLabel);
  			uIImageButton.SetSnapPoint("Play", snapPointIndex);
-@@ -113,13 +_,22 @@
+@@ -109,17 +_,29 @@
+ 			if (_data.IsCloudSave)
+ 				_buttonLabel.SetText(Language.GetTextValue("UI.MoveOffCloud"));
+ 			else
++				if (!ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient())
++					_buttonLabel.SetText(Language.GetTextValue("tModLoader.CloudWarning"));
++				else
+-				_buttonLabel.SetText(Language.GetTextValue("UI.MoveToCloud"));
++					_buttonLabel.SetText(Language.GetTextValue("UI.MoveToCloud"));
  		}
  
  		private void PlayMouseOver(UIMouseEvent evt, UIElement listeningElement) {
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria.ModLoader.Engine/Steam.cs
```diff
@@ -49,5 +51,14 @@ public static bool EnsureSteamAppIdTerrariaFile()
 			File.WriteAllText("steam_appid.txt", TerrariaAppID.ToString());
 			return exists;
 		}
+
+		public static bool CheckSteamCloudStorageSufficient()
+		{
+			if (IsSteamApp)
```
Jofairden, line 57: if (IsSteamApp && other && other)

patches/tModLoader/Terraria.ModLoader.Engine/Steam.cs
```diff
@@ -14,6 +14,8 @@ internal class Steam
 		public const uint TMLAppID = 1281930;
 		public const uint TerrariaAppID = 105600;
 
+		public const float SteamCloudQuotaLimitRatio = 0.9f;
```
Jofairden, line 17: Why 90%? You should be able to fill until 100%

patches/tModLoader/Terraria.GameContent.UI.Elements/UIWorldListItem.cs.patch
```diff
@@ -67,6 +75,19 @@
  		private void DeleteMouseOver(UIMouseEvent evt, UIElement listeningElement) {
  			_deleteButtonLabel.SetText(Language.GetTextValue("UI.Delete"));
  		}
+@@ -164,8 +_,11 @@
+ 		private void CloudButtonClick(UIMouseEvent evt, UIElement listeningElement) {
+ 			if (_data.IsCloudSave)
+ 				_data.MoveToLocal();
+-			else
++			else {
```
Jofairden, line 83: else if (ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient()) 
   _data.MoveToCloud();

patches/tModLoader/Terraria.GameContent.UI.Elements/UICharacterListItem.cs.patch
```diff
@@ -138,6 +146,19 @@
  		private void DeleteMouseOut(UIMouseEvent evt, UIElement listeningElement) {
  			_deleteButtonLabel.SetText("");
  		}
+@@ -131,8 +_,11 @@
+ 		private void CloudButtonClick(UIMouseEvent evt, UIElement listeningElement) {
+ 			if (_data.IsCloudSave)
+ 				_data.MoveToLocal();
+-			else
++			else {
++				if (!ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient())
```
Jofairden, line 155: else if (ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient())
_data.MoveToCloud();

Jofairden (changes requested):
Yeah, we need to do this but this PR needs changes. On hover is not the right location for this, the check should be done when the player file loads or maybe when the menu loads.

patches/tModLoader/Terraria.ModLoader.Engine/Steam.cs
direwolf420: should I still do line breaks after each &&?

patches/tModLoader/Terraria.ModLoader.Engine/Steam.cs
direwolf420: since steam cloud doesn't compress, I could figure out the size of the file and add it up to available, and compare to total?

direwolf420:
> On hover is not the right location for this, the check should be done when the player file loads or maybe when the menu loads.
On hover just displays the "CloudWarning" text, but yeah I agree this is a bad solution. You'd somehow have to calculate each ones size on initialize and then display the text based on the sum of filesize + available, most certainly creating bigger patches

PR #1029, opened 2020-07-28
Added ModRarity

Steviegt6:
### What is the new feature?
Modded Rarities (ModRarity). Acts like regular vanilla rarities (changes the item's name color when hovering over it, picking it up, etc.)

Modders have been adding their own modded rarities for a while now, but they've had to do it in hacky ways such as through ModifyTooltips(), which doesn't cover stuff like pick-up text. ModRarity aims to solve that issue by introducing an official implementation of modded rarities.

Not that I know of.

Giving modded items & vanilla items modded rarities.
Demonstration: https://i.gyazo.com/890592ce48e5eb0f6cccac1945ed5e11.png https://i.gyazo.com/82ca6eacaf00d047ee6dc51b8e273bfe.gif

### DIFF
--- a/patches/tModLoader/Terraria.ID/ItemRarityID.cs
+++ b/patches/tModLoader/Terraria.ID/ItemRarityID.cs
@@ -1,7 +1,11 @@
+using ReLogic.Reflection;
+using System;
+using Terraria.ModLoader;
+
 namespace Terraria.ID
 {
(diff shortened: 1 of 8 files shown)


Steviegt6:
I was dumb and made this for the 1.3 branch. Will make a 1.4 PR in a bit.

Issue #775, opened 2020-01-23
Add an upload speed to publishing


GoodPro712:
### Description
Add the upload speed as a text under the loading bar when publishing a mod.

### Expected behavior 
Text

### Actual behavior
No text

### Reproduction frequency
100%

Mirsario:
Implemented with #1032.

PR #1032, opened 2020-07-29
Added download/upload telemetry to mod downloads and uploads

OliHeamon:
### What is the new feature?
Shows the total download/upload progress in KB or MB based on size, as well as the average speed in the appropriate unit per second. Additionally, moved some code from UILoadMods to UIProgress so that the small text below the bar exists on all UIProgress instances.

Because of https://github.com/tModLoader/tModLoader/issues/775 (can now be marked as complete).
No.
https://cdn.discordapp.com/attachments/445276626352209920/738089021133160578/unknown.png
https://cdn.discordapp.com/attachments/445276626352209920/738094672039248023/unknown.png

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/DownloadManager/DownloadFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/DownloadManager/DownloadFile.cs
@@ -16,7 +16,7 @@ namespace Terraria.ModLoader.UI.DownloadManager
 
 		public HttpWebRequest Request { get; private set; }
 
-		public event Action<float, float, float> OnUpdateProgress;
+		public event ProgressUpdated OnUpdateProgress;
 		public event Action OnComplete;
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria/ModLoader/UI/DownloadManager/DownloadFile.cs
```diff
@@ -42,7 +42,7 @@ internal class DownloadFile
 			return true;
 		}
 
-		public Task<DownloadFile> Download(CancellationToken token, Action<float> updateProgressAction = null) {
+		public Task<DownloadFile> Download(CancellationToken token, Action<float, float, float> updateProgressAction = null) {
```
Mirsario, line 45: Why is this a <float, float, float>, if you're passing float, int and long to it? Just make it <float, long, long>

PR #1033, opened 2020-07-30
Fix/Change location of ItemSlot.ShiftClicked to be in a more appropri…

Lanboost:
### What is the bug?
The hook for inventory shift seems to not be in the correct location, it does not fire when using basic shift operations when no sell window (or only fires when holding down control (trashkey) otherwise).

### How did you fix the bug?
I moved the hook to from what I can understand is a more suitable location, but was mostly a "hack"

Probably, seems to be a lot of shift functions

### DIFF
--- a/patches/tModLoader/Terraria/UI/ItemSlot.cs.patch
+++ b/patches/tModLoader/Terraria/UI/ItemSlot.cs.patch
@@ -44,14 +44,61 @@
  						}
  
  						Recipe.FindRecipes();
-@@ -712,20 +_,24 @@
- 			if (inv[slot].type <= 0)
- 				return;
+@@ -683,28 +_,34 @@
+ 		}
  
+ 		private static bool LeftClick_SellOrTrash(Item[] inv, int context, int slot) {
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/UI/ItemSlot.cs.patch
```diff
@@ -44,14 +44,61 @@
...
+-				if ((uint)context <= 4u || context == 7)
+-					flag = true;
+-
+-				if (ControlInUse && flag) {
+-					SellOrTrash(inv, context, slot);
+-					result = true;
+-				}
++			Player player = Main.player[Main.myPlayer];
 +			if (PlayerHooks.ShiftClickSlot(player, inv, context, slot)) {
-+			}
```
Mirsario, line 52: This is an unnecessary large patch. Avoid moving and deleting vanilla code, and, especially, indenting it. Here you could've just used a short-circuit return.

Mirsario (changes requested):

PR #1034, opened 2020-07-30
Replace all the content dictionaries in Mod.cs with a generalised sys…

Chicken-Bones:
The pile of `IDictionary<string, ModX>`, `Mod.GetX` code, `Mod.XType`, `Mod.SetupContent` and all the overloads, are really a lot of duplication, plenty of inconsistency, and not really compatible with people adding their own ModTypes

With the power of generics, we can overcome! (also we can de-duplicate the name collision checking) The main idea is to channel everything through ModContent.Get/TryGet, and maybe ModContent.XType. @JavidPack has previously mentioned that returning 0 or -1 from the type methods is an opportunity for bugs and inconsistencies, so currently it's either TryGet or or Get with a KeyNotFound exception (or whatever dictionary is).

Lastly... the current implementation doesn't really favour iterating all ModX from a mod, which we use in a few places. I could make a multi-level dictionary, and even some strongly typed generic getters like `ModContent.Get<ModClass, ModX>(string name)` but that's probably overkill. I think it's better to just do something like `ModContent.GetAll<ModX>().Where(x => x.Mod == myMod);` etc, this can be a helper method, like `ModContent.GetAllFrom<ModX>(mod);`


Chicken-Bones:
Future todo notes
`IAddRecipes` on any content for adding recipes instead of current hardcode
translations need some cleanup

patches/tModLoader/Terraria/ModLoader/IModType.cs
```diff
@@ -5,11 +5,16 @@ public interface IModType : ILoadable
...
 		/// <summary>
 		/// The internal name of this instance.
 		/// </summary>
-		public string Name{get;}
+		public string Name { get; }
+
+		/// <summary>
+		/// => $"{Mod.Name}/{Name}"
+		/// </summary>
+		public string FullName { get; } // default Properties when??
```
Rartrin, line 18: Having this as part of the interface give the possibility for wrong implementations. If we need to use this a bunch, just make an extension methods for it, that way it is effectively "sealed" (just another case where extension properties would be nice).

patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
```diff
@@ -0,0 +1,36 @@
...
+	{
+		private static Dictionary<string, T> dict = new Dictionary<string, T>();
+
+		static ModTypeLookup() {
+			ModTypeLookup.OnClear += () => dict.Clear();
+		}
+
+		public static void Register(T instance) {
+			if (!instance.Mod.loading)
+				throw new Exception("AddBackgroundTexture can only be called from Mod.Load or IModContent.Load");
```
Rartrin, line 24: "AddBackgroundTexture" :P

patches/tModLoader/Terraria/ModLoader/Mod.cs
```diff
@@ -111,38 +111,15 @@ public void AddConfig(string name, ModConfig mc)
...
-		/// <returns></returns>
-		public int ItemType(string name) => GetItem(name)?.item.type ?? 0;
-
-		/// <summary>
-		/// Gets the GlobalItem instance with the given name from this mod.
-		/// </summary>
-		/// <param name="name">The name.</param>
-		/// <returns></returns>
-		public GlobalItem GetGlobalItem(string name) => globalItems.TryGetValue(name, out var globalItem) ? globalItem : null;
+		public IEnumerable<ILoadable> GetContent() => content;
```
Mirsario, line 120: May wanna make a readonly list field wrapping around content to return it here (still as IEnumerable.)

patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
```diff
@@ -0,0 +1,36 @@
...
+	internal class ModTypeLookup
+	{
+		public static event Action OnClear;
+
+		public static void Clear() => OnClear?.Invoke();
+	}
+
+	internal class ModTypeLookup<T> where T : IModType
+	{
+		private static Dictionary<string, T> dict = new Dictionary<string, T>();
```
Mirsario, line 16: I have some performance concerns about this has-it-all dictionary being used via Mod.Get/TryGet. Perhaps a secondary `Dictionary<Mod, Dictionary<string, T>>` would be better for those cases?

Mirsario (commented):
Great overall.

patches/tModLoader/Terraria/ModLoader/IModType.cs
Chicken-Bones: If we had either extension properties or default interface implementations (which we might get before tML 1.4 if we get mods building with C#8) then we could solve this nicely. It's unlikely any modders will implement the interface and not extend the type.

patches/tModLoader/Terraria/ModLoader/Mod.cs
Chicken-Bones: I normally choose performance over safety in tML. I don't think this is a high traffic method so it's unlikely to matter and in general I like the idea of returning read only collections, but anyone re-casting it could also get the impl via reflection anyway.

patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
Chicken-Bones: Dict lookup is O(1) right :'), Honestly you'd have to profile the benefits of tiered lookup but I don't expect it to be more than 30% faster.

patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
Chicken-Bones: good catch

patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
Chicken-Bones:
```
Preparing 8 mods with 10000 content each.
Single Dictionary. Retrieved 80000 items in 25ms. Avg: 324ns
Tiered Dictionary. Retrieved 80000 items in 79ms. Avg: 994ns

Preparing 8 mods with 1000 content each.
Single Dictionary. Retrieved 8000 items in 2ms. Avg: 332ns
Tiered Dictionary. Retrieved 8000 items in 4ms. Avg: 605ns

Preparing 50 mods with 1000 content each.
Single Dictionary. Retrieved 50000 items in 20ms. Avg: 404ns
Tiered Dictionary. Retrieved 50000 items in 37ms. Avg: 750ns

Preparing 50 mods with 100 content each.
Single Dictionary. Retrieved 5000 items in 1ms. Avg: 394ns
Tiered Dictionary. Retrieved 5000 items in 2ms. Avg: 532ns
```

https://gist.github.com/Chicken-Bones/b1ee052214521f87b4c0ca9d4fcb1d33

Mirsario: @Chicken-Bones well of course they're skewed by that. I was suggesting only using the tiered dictionary in cases when there's already separate Mod and content name parameters. I did a test like that, with non-string mod keys too just in case. Looks like Dictionary<string, Dictionary<string, T>> would be ~x2.5 faster at average.

```
Running...
Preparing 8 mods with 10000 content each.
Single Dictionary. Retrieved 80000 items in 18ms. Avg: 228ns
Tiered Dictionary (object mod keys). Retrieved 80000 items in 8ms. Avg: 110ns
Tiered Dictionary (string mod keys). Retrieved 80000 items in 7ms. Avg: 90ns

Preparing 8 mods with 1000 content each.
Single Dictionary. Retrieved 8000 items in 0ms. Avg: 95ns
Tiered Dictionary (object mod keys). Retrieved 8000 items in 0ms. Avg: 93ns
Tiered Dictionary (string mod keys). Retrieved 8000 items in 0ms. Avg: 79ns

Preparing 50 mods with 1000 content each.
Single Dictionary. Retrieved 50000 items in 11ms. Avg: 224ns
Tiered Dictionary (object mod keys). Retrieved 50000 items in 5ms. Avg: 104ns
Tiered Dictionary (string mod keys). Retrieved 50000 items in 4ms. Avg: 86ns

Preparing 50 mods with 100 content each.
Single Dictionary. Retrieved 5000 items in 0ms. Avg: 87ns
Tiered Dictionary (object mod keys). Retrieved 5000 items in 0ms. Avg: 97ns
Tiered Dictionary (string mod keys). Retrieved 5000 items in 0ms. Avg: 82ns
```

https://gist.github.com/Mirsario/ec340a2a94138dc7098656eca0997eaf

Mirsario: Added that secondary dictionary in https://github.com/tModLoader/tModLoader/commit/d5431444418cc67fd6b83b49a07008b971ec03e3

PR #1043, opened 2020-08-07
Added ModMenu and re-implemented UpdateMusic patches

OliHeamon:
### What is the new feature?
A ModMenu allows a modder to customize the aesthetics of the main menu by facilitating things like new logos and adding hooks for updating and drawing. It also provides a clean system of switching between menus to avoid mod conflicts caused by multiple mods editing the menu. Additionally, it allows modders to append UIElements to the main menu if they so choose.

Furthermore, the UpdateMusic code has been re-implemented, although currently music asset loading does not function properly, so that can be fixed in future PRs.

Because it is the easiest way to allow modders to edit previously difficult things such as main menu music.

No.

Turns the Journey's End and Terraria 1.3.5.3 menus into ModMenus, which lets the user choose between them:
https://media.discordapp.net/attachments/686646291685703680/741326110578442261/unknown.png?width=1204&height=677
https://cdn.discordapp.com/attachments/686646291685703680/741323251686441000/unknown.png

### ExampleMod updates
ExampleModMenu added to showcase features.


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2910,14 +3108,37 @@
...
++			ModMenu menu = ModLoader.ModLoader.AvailableMenus.Count > 0 ? ModLoader.ModLoader.AvailableMenus[ModLoader.ModLoader.currentModMenuIndex] : null;
+ 			int num = moonType;
++			if (gameMenu && menu != null) {
++				value = menu.SunTexture.Value;
++				value2 = menu.MoonTexture.Value;
++			}
++			else {
++				value = TextureAssets.Sun.Value;
+-			if (!TextureAssets.Moon.IndexInRange(num))
++				if (!TextureAssets.Moon.IndexInRange(num))
```
Mirsario, line 3132: Use gotos to avoid changing indentation.

patches/tModLoader/Terraria/ModLoader/ModMenu.cs
```diff
@@ -0,0 +1,88 @@
...
+	{
+		internal UserInterface userInterface;
+
+		/// <summary>
+		/// The amount of time (in ticks) that this ModMenu has been active for. Resets when the menu is not selected.
+		/// </summary>
+		public int timeActive;
+
+		protected override void Register() {
+		}
```
Mirsario, line 22: Why is this not used? Please move all your hardcode from ModLoader.cs to a loader class that functions through this override.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2274,19 +2409,82 @@
...
- 				spriteBatch.Draw(TextureAssets.Logo4.Value, new Vector2(screenWidth / 2, 100f), new Microsoft.Xna.Framework.Rectangle(0, 0, TextureAssets.Logo.Width(), TextureAssets.Logo.Height()), color3, logoRotation, new Vector2(TextureAssets.Logo.Width() / 2, TextureAssets.Logo.Height() / 2), logoScale, SpriteEffects.None, 0f);
  			}
+ 			else {
+-				spriteBatch.Draw(TextureAssets.Logo.Value, new Vector2(screenWidth / 2, 100f), new Microsoft.Xna.Framework.Rectangle(0, 0, TextureAssets.Logo.Width(), TextureAssets.Logo.Height()), color2, logoRotation, new Vector2(TextureAssets.Logo.Width() / 2, TextureAssets.Logo.Height() / 2), logoScale, SpriteEffects.None, 0f);
+-				spriteBatch.Draw(TextureAssets.Logo2.Value, new Vector2(screenWidth / 2, 100f), new Microsoft.Xna.Framework.Rectangle(0, 0, TextureAssets.Logo.Width(), TextureAssets.Logo.Height()), color3, logoRotation, new Vector2(TextureAssets.Logo.Width() / 2, TextureAssets.Logo.Height() / 2), logoScale, SpriteEffects.None, 0f);
++				if (ModLoader.ModLoader.lastUsedModMenuName != currentMenu.GetType().Name) {
```
Mirsario, line 2446: I would prefer if the following block was moved to a method in a loader, since it's pretty big.

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -130,6 +129,8 @@ internal static void PrepareAssets()
 		private static void Load(CancellationToken token = default)
 		{
 			try {
+				MenuLoader.moddedMenus.Clear();
```
Mirsario, line 132: Shouldn't this be called through ModContent.Unload()?

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -0,0 +1,105 @@
...
+			index = Utils.Repeat(index + 1, menus.Count);
+			currentModMenu = menus[index];
+		}
+
+		private static void GotoPreviousModMenu() {
+			List<ModMenu> menus = AvailableMenus;
+			int index = menus.IndexOf(currentModMenu);
+			index = Utils.Repeat(index - 1, menus.Count);
+			currentModMenu = menus[index];
+		}
```
Mirsario, line 40: Just replace this with this **local** method:
```cs
		static void OffsetModMenu(int offset) {
			var menus = AvailableMenus;
			
			currentModMenu = menus[Utils.Repeat(menus.IndexOf(currentModMenu) + offset, menus.Count)];
		}
```

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -0,0 +1,105 @@
...
+
+		internal static void GotoSavedModMenu() {
+			if (lastUsedModMenuName == nameof(MenuOldVanilla)) {
+				Main.instance.playOldTile = true; // If the previous menu was the 1.3.5.3 one, automatically reactivate it.
+			}
```
Mirsario, line 45: This can be moved to MenuOldVanilla.OnSelected

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+		internal static List<ModMenu> AvailableMenus => moddedMenus.Where(m => m.IsAvailable).ToList();
+
+		internal static string lastUsedModMenuName = $"ModLoader/{nameof(MenutML)}";
+
+		internal static void Add(ModMenu modMenu) {
+			moddedMenus.Add(modMenu);
+		}
+
+		private static void OffsetModMenu(int offset) {
+			var menus = AvailableMenus;
```
Chicken-Bones, line 31: what happens if a menu goes unavailable? eg a holiday menu as the clock hits midnight. Crash?

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+				Main.instance.playOldTile = true; // If the previous menu was the 1.3.5.3 one, automatically reactivate it.
+			}
+			List<ModMenu> menus = AvailableMenus;
+			int index = menus.FindIndex(m => GetModMenuName(m) == lastUsedModMenuName);
+			if (index == -1) {
+				index = 0;
+			}
+			currentModMenu = menus[index];
+			currentModMenu.SelectionInit();
+			lastUsedModMenuName = GetModMenuName(currentModMenu);
```
Chicken-Bones, line 47: the 2 lines above look redundant, should be handled by the start of `UpdateAndDrawModMenu`

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+			lastAvailableMenus = AvailableMenus;
+		}
+
+		private static string GetModMenuName(ModMenu menu) => $"{menu.Mod.Name}/{menu.Name}";
+
+		internal static void UpdateAndDrawModMenu(ModMenu currentMenu, SpriteBatch spriteBatch, GameTime gameTime, Color color, float logoRotation, float logoScale) {
```
Chicken-Bones, line 53: okay, so you almost had me here, `currentMenu` doesn't shadow `currentModMenu` but it looks awful close, just change this to `menu` or `modMenu`

patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -429,6 +429,7 @@ internal static bool TryGetTexture(string name, out Asset<Texture2D> texture)
 		}
 
 		internal static void UnloadModContent() {
+			MenuLoader.currentModMenu = null; // This is necessary to do before anything else because the main menu drawing is on a separate thread and it can cause issues with trying to use disposed textures
```
Chicken-Bones, line 432: ~~Seems like it could work, but you need a better comment here and where you check `currentModMenu` for null. Noting that it's _only_ null during mod reloading, and that none of the ModMenu code runs during reloading, it's all just back to vanilla~~

Letting `currentMenu` be null seems like a recipe for disaster. I think the vanilla and tML menus should just be always active, and this code should just switch back to the tML menu _if_ they were on a mod one

patches/tModLoader/Terraria/ModLoader/ModMenu.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		/// <summary>
+		/// Controls whether this ModMenu will be available to switch to. Useful if you want this menu to only be available at specific times.
+		/// </summary>
+		public virtual bool IsAvailable => true;
+
+		/// <summary>
+		/// Controls the name that shows up at the base of the screen when this ModMenu is active. If not overridden, it will use this mod's display name.
+		/// </summary>
+		public virtual string NameOnMenu => null;
```
Chicken-Bones, line 61: `DisplayName`, and default it to `Mod.DisplayName` here

patches/tModLoader/Terraria/ModLoader/ModMenu.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		internal void SelectionInit() {
+			userInterface = new UserInterface();
+			OnSelected(userInterface);
+		}
+
+		/// <summary>
+		/// The logo texture shown when this ModMenu is active. If not overridden, it will use the tModLoader logo.
+		/// </summary>
+		public virtual Asset<Texture2D> Logo => ModLoader.ManifestAssets.Request<Texture2D>("Terraria.ModLoader.Logo");
```
Chicken-Bones, line 31: Cache this pls and ty

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
OliHeamon: If the ModMenu is unavailable while selected, it will remain selected until the player selects a different menu (at which point it will no longer be selectable). In addition, crashes while reloading are preventing by setting the menu to null when an unload occurs.

patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
OliHeamon: those exist for when the game opens to a modded menu, because `lastUsedModMenuName` will be equal to the current ModMenu, so SelectionInit wouldn't fire. I can remove the assignment to lastUsedModMenuName but probably not the SelectionInit call

patches/tModLoader/Terraria/ModLoader/ModContent.cs
OliHeamon: there's a check in the Main patch for the menu being null - it just resorts to vanilla drawcode if that's the case. I've checked and it should all be null safe

patches/tModLoader/Terraria/ModLoader/ModMenu.cs
OliHeamon: I copied what the drawcode used in Main, so I'm not sure of any alternative

patches/tModLoader/Terraria/ModLoader/Default/ModLoaderMod.cs
```diff
@@ -60,6 +60,9 @@ public override void SetupAssetRepository(IList<IContentSource> sources, AssetRe
 			AddContent<UnloadedTilesWorld>();
 			AddContent<HelpCommand>();
 			AddContent<ModlistCommand>();
+			AddContent<MenutML>();
+			AddContent<MenuJourneysEnd>();
+			AddContent<MenuOldVanilla>();
```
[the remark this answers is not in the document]
OliHeamon: added them to the list in the static constructor

PR #1047, opened 2020-08-10
Repaired and overhauled player draw layers.

Mirsario:
### Overview

This PR repairs modded equips and player draw layers, additionally redesigning the latter. There can be holes in the design that I'm not aware of, so a discussion is welcome.
The following changes in layers are the most noteworthy:

- PlayerDrawLayer is now an autoloaded abstract ModType without any constructors, which should encourage OOP in mods.
- There is now a LegacyPlayerDrawLayer, which is very similar to the old PlayerLayer, as it has delegate-accepting constructors.
- **Head-only draw layers are no more**, and so layers in general should be twice less painful to deal with. PlayerDrawLayer now has an 'IsHeadLayer' virtual boolean property, which will determine whether or not the layer will be drawn for head-only draw contexts.
- ~~Layers are now sorted based on a 'depth' float field, in ascending order. I believe this, again, makes them way easier to deal with. Most notably, because of this, ModifyDrawLayers will now have readonly lists (inside a modname -> list dictionary), since there's no need to deal with indices. This also completely eliminates issues with some beginner modders relying on IndexOf for their insert code.~~
- ~~ModPlayer.ModifyDrawLayers has been split into AddDrawLayers and ModifyDrawLayers, very similar to the SetupStartInventory PR.~~

### My remarks

- ~~I think I would like to make LegacyPlayerDrawLayer internal to force OOP, and therefore remove ModPlayer.AddDrawLayers. Does anyone have objections?~~ (Applied.)
- ~~Perhaps PlayerDrawLayer.Setup's return value should determine whether the layer is visible, and not whether it should be added to the list? All vanilla layers are usually present even if they won't do anything. This would be consistent with that idea.~~ (Applied.)

### Sample usage for the new feature / ExampleMod updates

See [ExampleMod/Common/ExamplePlayerDrawLayer.cs](https://github.com/tModLoader/tModLoader/blob/1.4_playerrendering/ExampleMod/Common/ExamplePlayerDrawLayer.cs)


ExampleMod/Common/ExamplePlayerDrawLayer.cs
```diff
@@ -0,0 +1,48 @@
...
+			visible = drawPlayer.HeldItem?.type == ModContent.ItemType<ExampleItem>();
+			//The layer will be drawn right before the vanilla 'Head' layer.
+			depth = Head.depth - 0.1f;
+
+			//If you'd like to reference another PlayerDrawLayer's defaults,
+			//you can do so by getting its instance via ModContent.GetInstance<OtherDrawLayer>(), and calling GetDefaults on it.
+		}
+
+		public override void Draw(ref PlayerDrawSet drawInfo) {
+			//The following code draws ExampleItem's texture behind the player's head/
```
Ozzatron, line 30:
```suggestion
			//The following code draws ExampleItem's texture behind the player's head
```

Ozzatron, line 30:
```suggestion
			//The following code draws ExampleItem's texture behind the player's head
```

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -14,7 +16,7 @@ namespace Terraria.ModLoader
 	public static class EquipLoader
 	{
 		//in Terraria.Main.DrawPlayer and Terraria.Main.DrawPlayerHead get rid of checks for slot too high (not necessary for loading)
```
Ozzatron, line 18: These checks were removed, so this comment should also be removed.

patches/tModLoader/Terraria/ModLoader/PlayerHooks.cs
```diff
@@ -926,96 +926,51 @@ private class HookList
...
+
+				if (newLayers != null) {
+					string modName = modPlayer.Mod.Name;
+
+					if (!layers.TryGetValue(modName, out var list)) {
+						layers[modName] = list = new List<PlayerDrawLayer>();
+					}
+
+					list.AddRange(newLayers);
+				}
```
Chicken-Bones, line 964: Good patterns. I'd probably use a continue, but fine either way.

patches/tModLoader/Terraria/Graphics/Renderers/LegacyPlayerRenderer.cs.patch
```diff
@@ -0,0 +1,219 @@
...
++			//Actually make layer lists readonly
++			var readonlyLayersBackingDictionary = new Dictionary<string, IReadOnlyList<PlayerDrawLayer>>();
++			var readonlyLayers = new ReadOnlyDictionary<string, IReadOnlyList<PlayerDrawLayer>>(readonlyLayersBackingDictionary);
++
++			foreach (var pair in layers) {
++				readonlyLayersBackingDictionary[pair.Key] = pair.Value.AsReadOnly();
++			}
++
++			//Modify draw layers, but not the collections.
++			PlayerHooks.ModifyDrawLayers(drawPlayer, readonlyLayers);
```
Chicken-Bones, line 209: should there be a flag on a layer to disable it?

Chicken-Bones, line 209: I'd prefer if this whole block of code went in a separate method, `GetDrawLayers`, in `PlayerHooks` or `PlayerLayerHooks` or `CombinedHooks` and then the patch is just `foreach(var layer in GetDrawLayers(drawPlayer, vanillaLayers))`

patches/tModLoader/Terraria/ModLoader/Default/Developer/Jofairden/AndromedonEffect.cs
```diff
@@ -70,54 +70,43 @@ internal sealed class AndromedonEffect : ICloneable
...
+				yield break;
+			}
+
 			_headSlot ??= mod.GetEquipSlot($"PowerRanger_{EquipType.Head}", EquipType.Head);
 			_bodySlot ??= mod.GetEquipSlot($"PowerRanger_{EquipType.Body}", EquipType.Body);
 			_legSlot ??= mod.GetEquipSlot($"PowerRanger_{EquipType.Legs}", EquipType.Legs);
 
-			if (LayerStrength >= 0f) {
-				int i;
+			static PlayerDrawLayer Layer(PlayerDrawLayer layer, float depth) {
```
Chicken-Bones, line 82: ~~definitely hate this, surely there's a way to not have to reset visible every frame, or modify depth.~~
Shouldn't the Andromedon draw layers just be changed to extend `PlayerDrawLayer` and set their visible flags themselves?

patches/tModLoader/Terraria/ModLoader/DrawLayer.cs
```diff
@@ -1,280 +1,48 @@
...
-		/// <param name="name"></param>
-		/// <param name="layer"></param>
-		protected DrawLayer(string mod, string name, Action<InfoType> layer) {
-			this.mod = mod;
-			this.Name = name;
-			this.parent = null;
-			this.layer = layer;
-		}
+		/// <summary> This layer's depth. Determines the order in which layers will appear. </summary>
+		public float depth;
```
Chicken-Bones, line 19: Please, if we can replace this with a `Before/After` system. Depths changing each frame is meh, but using fractional depths to arbitrarily inb4 is even more sad.

patches/tModLoader/Terraria/DataStructures/PlayerDrawSet.cs.patch
```diff
@@ -0,0 +1,46 @@
...
+ 		public List<DrawData> DrawDataCache;
+ 		public List<int> DustCache;
+ 		public List<int> GoreCache;
+@@ -118,7 +_,9 @@
+ 
+ 		public Vector2 Center => new Vector2(Position.X + (float)(drawPlayer.width / 2), Position.Y + (float)(drawPlayer.height / 2));
+ 
+-		public void BoringSetup(Player player, List<DrawData> drawData, List<int> dust, List<int> gore, Vector2 drawPosition, float shadowOpacity, float rotation, Vector2 rotationOrigin) {
++		public void BoringSetup(Player player, List<DrawData> drawData, List<int> dust, List<int> gore, Vector2 drawPosition, float shadowOpacity, float rotation, Vector2 rotationOrigin, bool useLighting = true) {
++			Lighting.use = useLighting;
```
Chicken-Bones, line 24: could reduce a couple lines of patches/maintenance by setting this at the call site

patches/tModLoader/Terraria/ModLoader/Default/Developer/Jofairden/AndromedonItem.cs
```diff
@@ -35,59 +37,59 @@ internal abstract class AndromedonItem : DeveloperItem
...
+				Origin = drawInfo.legVect,
 				Rotation = drawPlayer.legRotation,
 				Texture = texture
 			};
 		}
 
-		public static PlayerLayer CreateShaderLayer(string name, PlayerLayer parent, Func<PlayerDrawInfo, DrawDataInfo> getDataFunc) {
-			return new PlayerLayer("ModLoaderMod", name, parent, (drawInfo) => {
+		public static PlayerDrawLayer CreateShaderLayer(string name, bool isHeadLayer, PlayerDrawLayer parent, Func<PlayerDrawSet, DrawDataInfo> getDataFunc) {
+			return new LegacyPlayerDrawLayer(ModLoader.GetMod("ModLoaderMod"), name, isHeadLayer, parent, (ref PlayerDrawSet drawInfo) => {
```
Chicken-Bones, line 92: `GetInstance<ModLoaderMod>`?

patches/tModLoader/Terraria/ModLoader/LegacyPlayerDrawLayer.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+using Terraria.DataStructures;
+
+namespace Terraria.ModLoader
+{
+	/// <summary> This class represents a DrawLayer for the player, and uses PlayerDrawInfo as its InfoType. Drawing should be done by adding Terraria.DataStructures.DrawData objects to Main.playerDrawData. </summary>
+	[Autoload(false)]
+	public sealed class LegacyPlayerDrawLayer : PlayerDrawLayer
```
Chicken-Bones, line 8: should this be private? I see it's used by `ModLoaderMod` but should it be? I get that it might be useful to provide a lambda for the draw func, maybe there's a way to roll the inherit vs functional constructor versions together? or at least rename this to not include `Legacy`.

Passing `Mod` to the constructor also feels wrong, since layers are allocated in the dict via what mod the hook adding them belongs to. I know we've used ModType in this inheritance chain, but did we need to? Is anyone going to enumerate layers and want to know what mod they came from/what their name is and not have that info? For debugging it'll be in the stack trace

patches/tModLoader/Terraria/ModLoader/PlayerLayerHooks.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
+
+namespace Terraria.ModLoader
+{
+	public static class PlayerLayerHooks
```
Chicken-Bones, line 5: wait so, `PlayerDrawLayer`s register themselves and add themselves to the layer list, unless they're the special subclass which isn't autoloaded. This is confusing. I think this design here is good, but `LegacyPlayerDrawLayer` is out of line...

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -733,28 +734,27 @@ public class ModPlayer:ModType
...
 
 		/// <summary>
-		/// Allows you to modify the drawing of the player. This is done by removing from, adding to, or rearranging the list, by setting some of the layers' visible field to false, etc.
+		/// Allows you to add your own player draw layers.
+		/// Return an enumerable with the layers you want to add.
 		/// </summary>
-		/// <param name="layers"></param>
-		public virtual void ModifyDrawLayers(List<PlayerLayer> layers) {
-		}
+		public virtual IEnumerable<PlayerDrawLayer> AddDrawLayers() => Enumerable.Empty<PlayerDrawLayer>();
```
Chicken-Bones, line 751: Maybe this should disappear? All layers should be added by default, and `ModifyDrawInfo` used to control visibility?

patches/tModLoader/Terraria/Graphics/Renderers/LegacyPlayerRenderer.cs.patch
```diff
@@ -0,0 +1,219 @@
...
++				PlayerDrawLayer.BalloonAcc,
++			};
++
+ 			if (drawinfo.weaponDrawOrder == WeaponDrawOrder.BehindBackArm)
+-				PlayerDrawLayers.DrawPlayer_27_HeldItem(ref drawinfo);
++				vanillaLayers.Add(PlayerDrawLayer.HeldItem);
+ 
+-			PlayerDrawLayers.DrawPlayer_12_Skin(ref drawinfo);
++			vanillaLayers.Add(PlayerDrawLayer.Skin);
+ 			if (drawinfo.drawPlayer.wearsRobe && drawinfo.drawPlayer.body != 166) {
```
Chicken-Bones, line 99: maybe all vanilla layers could be automatically added all the time, and we just set visibility in here?

patches/tModLoader/Terraria/DataStructures/PlayerDrawSet.cs.patch
Mirsario: This is public though, and is usable by modders.

patches/tModLoader/Terraria/ModLoader/Default/Developer/Jofairden/AndromedonEffect.cs
Mirsario: Ye, I have an outdated version of that in a stash, will update it.

patches/tModLoader/Terraria/Graphics/Renderers/LegacyPlayerRenderer.cs.patch
Mirsario: Ye, all the collections are readonly, but visibility and position can still be modified there.

Mirsario: That's how it was before, but I found it difficult to compare changes with vanilla layers and their orders. You sure about this?

patches/tModLoader/Terraria/DataStructures/PlayerDrawSet.cs.patch
Chicken-Bones: Can you give me a use case? it's only called in `DrawPlayer` as far as I can see

Mirsario: It could be useful if someone wants to implement their own IPlayerRenderer, I guess.

Chicken-Bones: in that case they could also set `Lighting.use` to true, if you make the field public

Mirsario: (Resolved in discord.)

patches/tModLoader/Terraria/Graphics/Renderers/LegacyPlayerRenderer.cs.patch
Mirsario: They're already always added. The conditions only affect their ordering.

Chicken-Bones:
- Layers can no longer move during the game. This will remove all the mind bending possibilities for conflicts where two mods move a layer, one orders after a layer and another mod moves it etc.
- Vanilla layers which previously 'moved' have been split and the movement conditions replaced with default visibilities. There are now separate Leggings and Robe layers, 3 HeldItem layers, and 2 FrontAcc layers. Think of layers as 'slots' for drawing.
- Layer ordering is done after mod loading, using a topographical sort. Currently to position a layer you use code like:
```cs
public override IEnumerable<LayerConstraint> GetConstraints() {
	yield return LayerConstraint.After(Leggings);
	yield return LayerConstraint.Before(Shoes);
}
```
- There's a `ModifyDrawLayerOrdering` hook in `ModPlayer`. I don't think there's an actual use case for this, but I added it in case... use with caution??
- Added `DrawLayer.PhysicalGroup` with `PreDraw` and `PostDraw` methods. Used for the torso and mount offsets in vanilla. 


### Food for thought:
- Should `DrawLayer.PhysicalGroup` be autoloaded/IModType?
- Should the code for inserting a layer directly before/after be simplified?
    - No because wider constraints and thinking about the layer orderings which actually matter make for better compatibility
    - Yes because that's the dominant use case, and we're pretty much banning reordering of layers anyway. (unless someone thinks of why we need `ModifyDrawLayerOrdering`)
- Can/should we somehow weave the declaration of vanilla layers with a commented version of the vanilla draw code to increase patch maintainability and make verifying the correct translation of our layers easier.

Chicken-Bones:
`PhysicalGroup` is now called `Transformation`
Mods cannot set `Visible` directly. They can only call `Hide`
`PlayerDrawLayerHooks.Layers` contains the list of all registered layers.
`PlayerDrawLayerHooks.DrawOrder` contains the sorted layer list, including auto-generated 'slots' for mobile layers, and _not including_ child layers.
`PlayerDrawLayer.ChildrenBefore/ChildrenAfter` contain the child layers. Children are drawn inside the parent `Transform` and only if the parent is `Visible`.

constraints are no more, now there's just one override. `Position GetDefaultPosition()`
A `Position` is one of:
- `Between(Layer1, Layer2)` creating a new static top level layer (most vanilla layers are this)
- `Before(parent)/After(parent)`
- `Multiple() { { Between(a, b), condition }, ... }` for mobile layers (held item and front acc).

It's expected that most use cases will be either top level layers or child layers. Creating your own mobile layers is not difficult, just see `PlayerDrawLayers.HeldItem` for an example.

Moving vanilla layers is tricky, but possible, as is making vanilla layers mobile.

I removed the generic `DrawLayer<LayerType, InfoType>` class because this is too specific to be useful in its exact form and we really just need to trial this and get it running.

`PlayerDrawLayer` is currently _fat_. Maybe some of the inner classes should be moved. 

I made `PlayerDrawLayers` public again, so I could put the vanilla layers in that class. That may not have been the best decision, but I don't think it will hurt, and modders might have some use for those methods.

Jofairden:
> Mods cannot set Visible directly. They can only call Hide
Alright, and what about Show? why take this functionality away?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
```diff
@@ -0,0 +1,57 @@
...
+			return drawInfo.drawPlayer.HeldItem?.type == ModContent.ItemType<ExampleItem>();
+
+			// If you'd like to reference another PlayerDrawLayer's visibility,
+			// you can do so by getting its instance via ModContent.GetInstance<OtherDrawLayer>(), and calling GetDefaultVisiblity on it
+		}
+
+		// This layer will be a 'child' of the head layer, and draw before (beneath) it.
+		// If the Head layer is hidden, this layer will also be hidden.
+		// If the Head layer is moved, this layer will move with it.
+		public override Position GetDefaultPosition() => new Before(PlayerDrawLayers.Head);
```
Jofairden, line 29: This honestly is very confusing when I Read this. The comment says it's a child of the head layer, so personally I would expect something like an .Add on the head layer. Except... is says it goes before the head layer?

Jofairden: Also is there are reason this is a method and not a get only property ?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
```diff
@@ -0,0 +1,57 @@
...
+		// If you want to make a layer which isn't a child of another layer, use `new Between(Layer1, Layer2)` to specify the position.
+		// If you want to make a 'mobile' layer which can render in different locations depending on the drawInfo, use a `Multiple` position.
+
+		protected override void Draw(ref PlayerDrawSet drawInfo) {
+			//The following code draws ExampleItem's texture behind the player's head.
+
+			if (exampleItemTexture == null) {
+				exampleItemTexture = ModContent.GetTexture("ExampleMod/Content/Items/ExampleItem");
```
Jofairden, line 37: It would be awesome if assets could be lazy loaded, just define the path in the constructor when you define the asset.

patches/tModLoader/Terraria/DataStructures/LegacyPlayerDrawLayer.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Terraria.ModLoader;
+
+namespace Terraria.DataStructures
+{
+	[Autoload(false)]
+	internal sealed class LegacyPlayerDrawLayer : PlayerDrawLayer
```
Jofairden, line 6: Not sure what the reason of this is? Is it so players can use it? (If so, do we want that?)

patches/tModLoader/Terraria/DataStructures/PlayerDrawSet.cs.patch
```diff
@@ -0,0 +1,47 @@
+--- src/Terraria/Terraria/DataStructures/PlayerDrawSet.cs
++++ src/tModLoader/Terraria/DataStructures/PlayerDrawSet.cs
+@@ -10,6 +_,15 @@
+ {
+ 	public struct PlayerDrawSet
+ 	{
++		//An absolutely utterly epic trick to avoid big patches for disabling lighting below.
++		private class Lighting
++		{
++			public static bool use;
```
Jofairden, line 10: Just name enabled please

patches/tModLoader/Terraria/ModLoader/PlayerDrawLayer.cs
```diff
@@ -0,0 +1,150 @@
...
+		public class After : Position
+		{
+			public PlayerDrawLayer Layer { get; }
+
+			public After(PlayerDrawLayer layer) {
+				Layer = layer;
+			}
+		}
+
+		public abstract class Transformation {
```
Jofairden, line 59: I like this!

patches/tModLoader/Terraria/ModLoader/PlayerDrawLayerHooks.cs
```diff
@@ -0,0 +1,99 @@
...
+					case After a:
+						a.Layer.AddChildAfter(layer);
+						break;
+					case Multiple m:
+						int slot = 0;
+						foreach (var (pos, cond) in m.Positions)
+							positions.Add(new MobilePlayerDrawLayerSlot(layer, cond, slot++), pos);
+						break;
+					default:
+						throw new ArgumentException($"PlayerDrawLayer {layer} has unknown Position type {kv.Value}");
```
Jofairden, line 50: Can't we use expressive switch cases yet?

Jofairden (commented):
I only have some questions why the legacy class exists

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Mirsario: They will be autoloaded, and this code doesn't conflict with that, since it returns an asset, which doesn't have to be loaded immediately.

patches/tModLoader/Terraria/DataStructures/LegacyPlayerDrawLayer.cs
Mirsario: This class accepts delegates, and is used with vanilla draw methods. It's internal because we want to force modders to use OOP.

patches/tModLoader/Terraria/ModLoader/PlayerDrawLayerHooks.cs
Chicken-Bones: Maybe? Link?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Chicken-Bones: Can make it a property.
Can you suggest better comnents?
It's a child so it inherits visibility, transform, and will move with the parent if someone reorders it.

It renders _Before_ (behind) the head layer

Jofairden: Well my comment is mostly regarding how, the way it is used, it isn't apparent (to me) that it's a child of Head, only that it goes before it. How can you see it is a child, or are you just supposed to know that's 'how it works'?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Jofairden: I don't mean autoloaded, I mean lazy loaded. That means, you would do something like:
`Asset myAsset = new Asset<Texture2D>("My/Path/To/Asset");`
The null check and loading GetTexture I commented on essentially is the lazy loading inside the asset you define. 
This just gets rids of null checks and calling GetTexture all the time for your assets.

patches/tModLoader/Terraria/ModLoader/PlayerDrawLayerHooks.cs
Jofairden: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression
See first example

patches/tModLoader/Terraria/DataStructures/LegacyPlayerDrawLayer.cs
Jofairden: So why isn't it called VanillaPlayerDrawLayer then ?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Mirsario: In any context, X being a child of Y shouldn't mean that it's rendered after it. It just means that it's attached to it, while still being able to have relative (local) offsets.

Jofairden: Yeah, but where can I see this layer is a child of the Head layer?

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Jofairden: See also: https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization

Mirsario: I did a typo. I meant that we plan to make assets load their .Values on-demand.

ExampleMod/Common/ExamplePlayerDrawLayer.cs
Chicken-Bones: programmatically, you can see it in the head layer's `Children` list. If your suggestion is that we name this something like `BeforeParent` and `AfterParent` that's completely reasonable.

patches/tModLoader/Terraria/ModLoader/PlayerDrawLayerHooks.cs
Chicken-Bones: We can, but these are statements, not expressions.

PR #1059, opened 2020-08-22
[Draft] OOPing double/extra jumps

direwolf420:
### Description
I originally made a #807 Double jump PR for 1.3.5.2 to explore the possibility for modders to easily add double jumps that are compatible with each other, Due to the way they were implemented to work around vanilla ones, you could only order them after all vanilla jumps (before is also possible but not implemented in that PR). In 1.4, a few things changed on the vanilla side which might encourage OOPing double jumps as a whole for tml. This PR explores this in an attempt to make double jumps have a set ordering. By no means this implementation is ideal, as it is done purely to showcase what OOPing will bring for modders.

### Rudimentary Implementation

Basically, in extension to `ModExtraJump`, a `VanillaExtraJump` class exists that is just used as a reference for the ordering aspect. They aren't included in any jumping calculations/decisions yet. This allows the modder to insert their jump at any point in the jumping "progression" like this:
```cs
public override VanillaExtraJump JumpAfter => VanillaExtraJump.Sandstorm;
```

Because the vanilla order is hardcoded, a goto is used to circumvent the part of the code responsible for deciding which jump to do next, and the logic is copied into a Loader method that takes into account the ordering. This [patch](https://github.com/tModLoader/tModLoader/compare/tModLoader:1.4...absoluteAquarian:1.4_extrajump_extended?ts=4#diff-319c7d2075ca4ff585a98979ba6237c0R1492) around `jumpDecision` is the crucial one. It "wraps" around the existing bools for jump execution by passing them as ref.

In the PR the example adds 3 new jumps, where two have custom ordering, shown in [this video](https://cdn.discordapp.com/attachments/634778013531373580/746017938812960798/jump2.webm).

### Outlook

As for OOPing all jumps, this would make way for a `GlobalExtraJump` where you can modify vanilla jumps too, for example wearing a certain armor would give you this effect: `Increases jump height from 'Cloud in a bottle' by 100%`.

As for implementing/rewriting vanilla jumps to support this, I don't quite know how to go about this in regards to the existing fields for the jumps (which there are 27 of (9 jumps, each with 3 variables)).

Again, this is by far not the complete product, it was just made to showcase what can be done with the ordering.
And the most important question: `DoubleJump` or `ExtraJump`? :smile:


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3485,6 +3608,22 @@
...
++		/// </summary>
++		internal ModExtraJump activeJump;
++
++		private ModExtraJump GetModExtraJump(int? index) => index.HasValue ? modExtraJumps[index.Value] : null;
++
++		public ModExtraJump GetModExtraJump(Mod mod, string name) => GetModExtraJump(mod.GetExtraJump(name)?.index);
++
++		public T GetModExtraJump<T>() where T : ModExtraJump => (T)GetModExtraJump(ModContent.GetInstance<T>()?.index);
++
++		public void EnableExtraJump<T>() where T : ModExtraJump => GetModExtraJump<T>().hasJumpOption = true;
```
Chicken-Bones, line 3626: @Mirsario maybe all of this should be moved to Player_tML.cs?

patches/tModLoader/Terraria/ModLoader/ModExtraJumpLoader.cs
```diff
@@ -0,0 +1,218 @@
...
+		/// <param name="player">Jumping player</param>
+		public static bool AnyJumpAgain(Player player) => player.modExtraJumps.Any(j => !(j is VanillaExtraJump) && j.canJumpAgain);
+
+		/// <summary>
+		/// Allows the ability to double jump if it is enabled.
+		/// </summary>
+		/// <param name="player">Jumping player</param>
+		internal static void SetJumpState(Player player) {
+			foreach (ModExtraJump extraJump in player.modExtraJumps.Where(j => !(j is VanillaExtraJump))) {
+				if (extraJump.hasJumpOption) {
```
Chicken-Bones, line 200: this should just set `extraJump.canJumpAgain = extraJumpOption.hasJumpOption`

patches/tModLoader/Terraria/ModLoader/ModExtraJumpLoader.cs
```diff
@@ -0,0 +1,218 @@
...
+					extraJump.canJumpAgain = true;
+				}
+			}
+		}
+
+		/// <summary>
+		/// Toggles the ability to double jump if it's active, otherwise forbid it.
+		/// </summary>
+		/// <param name="player">Jumping player</param>
+		internal static void ToggleJumpState(Player player) {
```
Chicken-Bones, line 210: This should be called `DisableLostJumps` or something to that effect. Its sole purpose is to disable jumps when their benefactor is lost during airtime (also update the comment with that)

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1468,6 +1476,78 @@
...
++					}
+ 				}
+ 
+ 				releaseJump = false;
+@@ -14655,7 +_,7 @@
+ 
+ 		public void CarpetMovement() {
+ 			bool flag = false;
+-			if (grappling[0] == -1 && carpet && !canJumpAgain_Cloud && !canJumpAgain_Sandstorm && !canJumpAgain_Blizzard && !canJumpAgain_Fart && !canJumpAgain_Sail && !canJumpAgain_Unicorn && !canJumpAgain_Santank && !canJumpAgain_WallOfFleshGoat && !canJumpAgain_Basilisk && jump == 0 && velocity.Y != 0f && rocketTime == 0 && wingTime == 0f && !mount.Active) {
++			if (grappling[0] == -1 && carpet && !canJumpAgain_Cloud && !canJumpAgain_Sandstorm && !canJumpAgain_Blizzard && !canJumpAgain_Fart && !canJumpAgain_Sail && !canJumpAgain_Unicorn && !canJumpAgain_Santank && !canJumpAgain_WallOfFleshGoat && !canJumpAgain_Basilisk && !ModExtraJumpLoader.AnyJumpAgain(this) && jump == 0 && velocity.Y != 0f && rocketTime == 0 && wingTime == 0f && !mount.Active) {
```
Chicken-Bones, line 1536: it'd be nice if at least the `canJumpAgain` logic was all handled by `ModExtraJumpLoader` via a ref property in `VanillaDoubleJump`

would make this easier to maintain

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1646,6 +1735,32 @@
  			}
  
  			lastTileRangeX = tileRangeX;
+@@ -17727,6 +_,7 @@
+ 				canJumpAgain_Santank = false;
+ 				canJumpAgain_WallOfFleshGoat = false;
+ 				canJumpAgain_Basilisk = false;
++				ModExtraJumpLoader.ResetAllJumps(this);
```
Chicken-Bones, line 1742: Just above this you need to handle 
```
			if (mount.Active)
				movementAbilitiesCache.CopyFrom(this);
			else
				movementAbilitiesCache.PasteInto(this);
```

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -2105,6 +2220,14 @@
  			else if (shieldRaised) {
  				bodyFrame.Y = bodyFrame.Height * 10;
  			}
+@@ -25178,6 +_,7 @@
+ 			isPerformingJump_Sail = false;
+ 			isPerformingJump_Unicorn = false;
+ 			isPerformingJump_Santank = false;
++			ModExtraJumpLoader.ResetAllVisualEffects(this);
+ 		}
```
Chicken-Bones, line 2228: Below this, rocket boots currently check `canJumpAgain_Cloud` to know when they should trigger. You'll need to change that to support after cloud double jumps

Chicken-Bones (commented):
Overall a solid PR, I don't see any way to implement this with better overall design. See what you can do with my comments.

patches/tModLoader/Terraria/Player.cs.patch
direwolf420: Elaborate please, how would such an implementation look like?

patches/tModLoader/Terraria/ModLoader/ModExtraJumpLoader.cs
direwolf420: Wouldn't follow vanilla logic then

patches/tModLoader/Terraria/Player.cs.patch
direwolf420: Judging by the wiki, the rocket boots only interact with the cloud, not other jumps: `If the Cloud in a Bottle is also equipped, its double-jump effect will engage first; pressing and holding the jump key afterward while still in midair will then activate the Rocket Boots.`

So idk if we should touch this behavior

patches/tModLoader/Terraria/ModLoader/ModExtraJumpLoader.cs
```diff
@@ -0,0 +1,220 @@
...
+			// This method is only called if atleast one modded jump exists
+			foreach (var extraJump in player.modExtraJumps) {
+				if (extraJump.index == VanillaExtraJump.Mounts.index) {
+					if (player.canJumpAgain_Basilisk) {
+						basilisk = true;
+						player.canJumpAgain_Basilisk = false;
+					}
+
+					if (player.canJumpAgain_WallOfFleshGoat) {
+						goat = true;
```
Mirsario, line 26: This all should be made OOP in the actual non-lazy implementation. If we won't have hardcoded messes like this, then it'll be fine.

direwolf420: how would you deal with the 27 fields related to jumps? Turn them into properties pointing at the associated `DoubleJump` and the field?

direwolf420:
Closed as completed in <https://github.com/tModLoader/tModLoader/commit/24366adb8f4ec04d4fb91e5dc1038189959e222d>

PR #1064, opened 2020-08-28
Support Item Research & the Duplication Menu

kittenchilly:
### What is the new feature?
Supports item research as well as the duplication menu from Journey mode natively. You just have to set the research amount of an item in SetStaticDefaults and it'll be available to research and duplicate!
Also adds a ModifyResearchSorting hook in ModItem and GlobalItem. This allows you to group existing items together separately from the automatic system, as well as add new groups for modded items (modded damage classes and the like). This hook is optional; if you don't use it, items will just get automatically assigned based on the vanilla algorithm.

### ExampleMod updates
Added research amounts to literally every ExampleMod item based on their adjacent item from vanilla. Also added an example GlobalItem sorting code using ModifyResearchSorting which sorts all the ExampleMod items together in the Duplication menu, as well as modifying the existing sort of a vanilla item, and a ModItem example in ExampleTorch that manually assigns it into the Torch sorting group since vanilla doesn't assign torches (and a few other items) automatically.


ExampleMod/Content/Items/Accessories/ExampleShield.cs
```diff
@@ -12,6 +13,7 @@ public class ExampleShield : ModItem
 	{
 		public override void SetStaticDefaults() {
 			Tooltip.SetDefault("This is a modded shield.");
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[item.type] = 1;
```
Mirsario, line 16: What's the default?

ExampleMod/ExampleMod.cs
```diff
@@ -43,11 +40,22 @@ public class ExampleMod : Mod
...
+		public override void AddResearchSorting(Item item, ref ContentSamples.CreativeHelper.ItemGroup itemGroup) {
+		//Lets you add your own sorting methods in Journey mode's Duplication menu, or change the sorting methods of existing items.
+		//This is useful for items with custom damage classes, as well as specific themed items.
+			if (item.modItem?.Mod == this) { 
+				itemGroup = (ContentSamples.CreativeHelper.ItemGroup)1337; //This number is where the item sort is in relation to any other sorts added by vanilla or mods. To see the vanilla sorting numbers, refer to (insert wiki page here).
+			};
+			if (item.type == ModContent.ItemType<Content.Items.Placeable.ExampleTorch>()) {
+				itemGroup = ContentSamples.CreativeHelper.ItemGroup.Torches; //Vanilla usually matches sorting methods with the right type of item, but sometimes, like with torches, it doesn't. Make sure to set whichever items manually if need be. 
+			}
+		}
```
Mirsario, line 53: Why is this hook in Mod, and not GlobalItem?

Mirsario: Additionally, its name is better off being 'ModifyResearchSorting'. And there should be a ModItem version.

Mirsario (changes requested):
The added hook is most definitely out of place.

ExampleMod/ExampleMod.cs
Chicken-Bones: This hook in general needs more development and discussion. Free-for-all id based sorting seems unnecessarily chaotic and error prone.

ExampleMod/Content/Items/Accessories/ExampleShield.cs
Chicken-Bones: Also use `Type` instead of `item.type` in this and 40 other places.

ExampleMod/ExampleMod.cs
kittenchilly: This is practically the way vanilla did it, and I think it didn't matter as much for such a small thing as this.

kittenchilly: I agree that putting it in GlobalItem would make sense, but you barely use it for items that aren't in your mod (which is what most people use GlobalItem for). Its more of an organizational thing, like how AddRecipe is in Mod as well. I guess I could add a ModItem and GlobalItem version like AddRecipe though.

ExampleMod/Content/Items/Accessories/ExampleShield.cs
kittenchilly: What do you mean "whats the default?" There is no default, so you have to set it. I wanted to keep the option of people not wanting their items to be researchable, which you can easily do by just not setting it. Doesn't crash or anything. You set a static variable and it makes sense in load times, so I just put it there. Theoretically you can put it anywhere, I guess.

ExampleMod/Common/GlobalItems/ExampleResearchSorting.cs
```diff
@@ -0,0 +1,24 @@
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.GlobalItems
+{
+		//This file shows examples of creating and setting your own sorting group in Journey mode's Duplication menu, as well as changing the sorting groups of existing items.
+		//Creating your own research sorting group can be useful if your mod has a specific custom item type, or the vanilla sorting method doesn't assign the right group to your item.
+		//While you can do this in ModItem, there are benefits to adding all modded items to sorting groups in bulk using GlobalItem, as shown here.
+	public class ExampleResearchSorting : GlobalItem
```
Mirsario, line 10:
```suggestion
	//This file shows examples of creating and setting your own sorting group in Journey mode's Duplication menu, as well as changing the sorting groups of existing items.
	//Creating your own research sorting group can be useful if your mod has a specific custom item type, or the vanilla sorting method doesn't assign the right group to your item.
	//While you can do this in ModItem, there are benefits to adding all modded items to sorting groups in bulk using GlobalItem, as shown here.
	public class ExampleResearchSorting : GlobalItem
```

ExampleMod/Common/GlobalItems/ExampleResearchSorting.cs
```diff
@@ -0,0 +1,24 @@
...
+	{
+		//Here we add both every item in this mod to a single custom sorting group, as well as add an existing item, the copper shortsword, to a vanilla sorting group.
+		//These can be interchanged, modded items can go in vanilla sorting groups and vice versa.
+		public override void ModifyResearchSorting(Item item, ref ContentSamples.CreativeHelper.ItemGroup itemGroup) {
+			if (item.modItem?.Mod == Mod) {
+				itemGroup = (ContentSamples.CreativeHelper.ItemGroup)1337; //This number is where the item sort is in relation to any other sorts added by vanilla or mods. To know where your custom group relates to the vanilla sorting numbers, refer to (insert wiki page here).
+			};
+			if (item.type == ItemID.CopperShortsword) {
```
Mirsario, line 18:
```suggestion
			}

			if (item.type == ItemID.CopperShortsword) {
```

ExampleMod/Content/Items/ExampleSoul.cs
```diff
@@ -11,11 +12,13 @@ public class ExampleSoul : ModItem
...
 			DisplayName.SetDefault("Soul of Exampleness");
 			Tooltip.SetDefault("'The essence of example creatures'");
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 25;
 			// Registers a vertical animation with 4 frames and each one will last 5 ticks (1/12 second)
 			Main.RegisterItemAnimation(item.type, new DrawAnimationVertical(5, 4)); // Reminder, (4, 6) is an example of an item that draws a new frame every 6 ticks
 			ItemID.Sets.AnimatesAsSoul[item.type] = true; // Makes the item have 4 animation frames by default.
 			ItemID.Sets.ItemIconPulse[item.type] = true; // The item pulses while in the player's inventory
 			ItemID.Sets.ItemNoGravity[item.type] = true; // Makes the item have no gravity
+
 		}
```
Mirsario, line 22:
```suggestion
			ItemID.Sets.ItemNoGravity[item.type] = true; // Makes the item have no gravity
		}
```

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -208,6 +209,15 @@ public class GlobalItem:ModType
...
+		/// <summary>
+		/// Allows you to set an item's sorting group in Journey Mode's duplication menu. This is useful for setting custom item types that group well together, or whenever the default vanilla sorting doesn't sort the way you want it.
+		/// </summary>
+		/// <param name="item">The item being used</param>
+		/// <param name="itemGroup">The item group this item is being assigned to</param>
+		public virtual void ModifyResearchSorting(Item item, ref ContentSamples.CreativeHelper.ItemGroup itemGroup) {
+		}
+
+
 		/// <summary>
```
Mirsario, line 221:
```suggestion
		}

		/// <summary>
```

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -413,6 +413,19 @@ private class HookList
...
+		public static void ModifyResearchSorting(Item item, ref ContentSamples.CreativeHelper.ItemGroup itemGroup) {
+			if (item.IsAir)
+				return;
+
+			item.modItem?.ModifyResearchSorting(ref itemGroup);
+
+			foreach (var g in HookModifyResearchSorting.arr)
+				g.Instance(item).ModifyResearchSorting(item, ref itemGroup);
+
+		}
```
Mirsario, line 427:
```suggestion
				g.Instance(item).ModifyResearchSorting(item, ref itemGroup);
		}
```

Mirsario (approved):

Chicken-Bones (approved):

Issue #1050, opened 2020-08-14
Force mods to pass a reload test before upload to mod browser.


Chicken-Bones:
### Description
Force mods to be enabled, and reload without crashing before uploading to mod browser.

### What does this proposal attempt to solve or improve?
In 1.3, we were wary of force unloading textures and sounds because mods would fail to reload properly. Even so, many mods still fail. With this in place for 1.4 we could be stricter on mods properly unloading.

### Which (other) solutions should be considered?
This isn't foolproof, many mods will only fail after actually loading a world, using certain items, or interacting with certain content but it's a step in the right direction.

We could also consider displaying a warning if the mod doesn't GC properly, saying "may cause memory leaks, follow this guide"

Chicken-Bones:
Actual TODO is to fix mod failed to unload properly warnings. In my time messing around with ALCs I couldn't get them to unload properly all the time, but I had made _some_ progress in tracking it down.

PR #1067, opened 2020-08-31
Enforce reload before publishing

Agrair:
### What is it?
Does what it says. Solves #1050, with very little code :^)

### Other implementations
To be honest, I'm not sure. This implementation definitely works but I dunno if it fits your guys' idea of good practice. In the future, maybe a new translation key that reads "Reload Before Publishing" or just "Reload" (which would fit a lot better actually) that can be used instead just reusing the "Reload Mods" one might be nice, but I think any modder who reads it should probably be able to figure it out based on context anyway.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
@@ -162,6 +162,16 @@ namespace Terraria.ModLoader.UI
 			}
 			SoundEngine.PlaySound(10);
 			try {
+				if (ModLoader.GetMod(_builtMod.Name) == null) {
+					if (!_builtMod.Enabled)
+						_builtMod.Enabled = true;
+					Main.menuMode = Interface.reloadModsID;
+					ModLoader.OnSuccessfulLoad += () => {
+						PublishMod(null, null);
+					};
+					return;
+				}
+
 				var modFile = _builtMod.modFile;
 				var bp = _builtMod.properties;
 


patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
```diff
@@ -40,7 +40,15 @@ internal class UIModSourceItem : UIPanel
...
 
-			string addendum = Path.GetFileName(mod).Contains(" ") ? $"  [c/FF0000:{Language.GetTextValue("tModLoader.MSModSourcesCantHaveSpaces")}]" : "";
+			bool isLoaded = ModLoader.GetMod(mod) != null;
+
+			string addendum = "";
+			if (Path.GetFileName(mod).Contains(" ")) {
+				addendum = $"  [c/FF0000:{Language.GetTextValue("tModLoader.MSModSourcesCantHaveSpaces")}]";
+			}
+			else if (builtMod != null && !isLoaded) {
+				addendum = $"  [c/FF0000:{Language.GetTextValue("tModLoader.ModReloadRequired")}]";
```
Agrair, line 50: If you're reviewing my PR and aren't sure what I meant by reusing the "Reload Required" translation, here it is. But I think it should be fine.

Agrair:
Actually maybe I should have made an isolated "Reload" button, and just scrap the "Reload Required' header.

Jofairden:
From my personal understanding, the idea was just that the mod would cause a reload once you press publish, and after a successful reload process the publishing process can begin. The reload process will throw an exception in case of failure either way, which should stop the publish process from triggering.

Agrair:
Hmm ok. Looking it over it's bugged anyway because if the mod is already enabled, it'll crash when you press Reload. (Enabled should be set only if it's false because it calls .Add on Mods)

Agrair:
So I did Jof's suggestion. It looks a bit like a workaround, but it works like a charm. And sorry for all the force-pushing, now that I think about it that was probably unnecessary, but I just *really* wanted the PR to display a single commit. Feel free to look at the other ones though.

Chicken-Bones:
Okay, this is sweet. Now we just need to enable a flag here that turns memory leak warnings into errors.
Happy to accept as is if you'd like

Mirsario:
> Now we just need to enable a flag here that turns memory leak warnings into errors.

You mean throwing errors on every remaining static field value? That's gonna get ugly way too quick. Better just assume that we'll finally move our asses to .NET Core eventually, for that assembly unloading. Or, just come up with a way of knowing what static constructors were called and which weren't. That way we'll be able to null out everything on unload, and reinitialize things on reload.

Agrair:
Hmm. Obviously there’s a bunch to be desired with enforcing not only successful loads but also proper unload mechanics, but I agree it’ll probably be messy. If you’re asking me if I’m ok with it getting merged, then yeah I am. I’m gonna take a peek at the GC stuff and see if I can put what Mirsar said into practice, but in all likelihood that’s either not going to end up being committed at all or just long after this PR should have been merged in the first place.

PR #1063, opened 2020-08-27
Support the Bestiary + new NPC Loot system

kittenchilly:
### What is the new feature?
This PR adds support to the new Bestiary introduced in 1.4. It automatically adds any modded NPCs into the Bestiary list (unless specified otherwise in SetStaticDefaults), as well as adding a new SetBestiary hook in ModNPC & GlobalNPC that allows you to set Bestiary tags, spawning conditions, and descriptions.
Alongside that, this PR adds support for the extensive new NPC Loot system with a new ModifyNPCLoot hook (the old NPCLoot hooks were renamed to OnDeath)..

### Why should this be part of tModLoader?
Currently opening the Bestiary crashes in the current build, technically making this a bug fix. I did both the Bestiary and the new NPC Loot system in one PR as they are closely linked to eachother functionally.

### ExampleMod updates
Re-added the ExampleWorm NPC, mostly as an example of Bestiary formatting for any multi-part NPCs.
Updated both the PartyZombie and ExamplePerson NPCs as well to have new Bestiary and NPC Loot info.
Added extensive GlobalNPC documentation for practically everything you can do with the new NPC Loot system.

PR #1078, opened 2020-09-10
Bestiary UI Bug

ModernMAK:
### What is the bug?
When viewing the Bestiary, the Bestiary is blank and tells the user that an error has occurred.

This is because `ContentSamples.NpcBestiarySortingId` is cleared when `ModContent` calls  `ContentSamples.Initialize()`.

### How did you fix the bug?
Rebuilding `NpcBestiarySortingId` after every call to `Initialize()` fixes the issue.

`Initialize()` could be patched to not clear `NpcBestiarySortingId`.


direwolf420:
#1063 potential duplicate, this branch also fixed what you described afaik

ModernMAK:
Confirmed duplicate; looks like its fixed in #1063

PR #1080, opened 2020-09-12
Fix for water CTD

ModernMAK:
### What is the bug?
A missing water asset causes a `NullReferenceException` resulting in a CTD. Specifically, `Water_12`.

### How did you fix the bug?
By changing the # of water textures to load from 12 to 13.
On 1.4.0.5; I can confirm there are 13 water files, so I'm not sure if this was a mistake, or if a version of 1.4 only has 12 assets.

The draw code for water could be patched to fail safely when missing a water asset.


JavidPack:
Thanks for finding this.

PR #1081, opened 2020-09-12
Mod Accessory BugFix

ModernMAK:
### What is the bug?
Certain `EquipTypes` that `ModItem`s load do not display their equip sprites on the character.
Back accessories specifically did not show up. (Tested with ExampleMod's Hive Nest item on raw 1.4 branch)

### How did you fix the bug?
Altered the checks for whether an equip accessory could be drawn.
Previously, it used the hard-coded vanilla value as an upper limit (except for wings, balloons, and shields, which had none)
Now, it uses the appropriate Texture list to determine an upper limit. For those with a missing upper limit, I re-added it for consistency.

The upper limits could be removed completely.


Mirsario:
This is all fixed in the playerrendering branch.

Mirsario:
If you want to search for bugs yourself, I recommend doing that on 1.4_mergedtesting. That has playerrendering, contenttags and modsystems merged together. I'll update it to your previous fix in few minutes.

ModernMAK:
Alright, Ill start doing that; trying to get a stable-ish version of 1.4 mod loader. A quick peek at the other branch, and this doesn't seem to add anything of value.

Issue #1004, opened 2020-07-04
Handful of `Dictionary`s somewhere to centralize "entity group" storage (one for each `Entity` type)


hamstar0:
### Description
See title. That would literally be as simple as something like:
```cs
public class ModContent {
	private static IDictionary<string, IReadOnlyList<int>> _ItemGroups;
	public static IReadOnlyDictionary<string, IReadOnlyList<int>> ItemGroups { get; private set; }

	static ModContent() {
		_ItemGroups = new ConcurrentDictionary<string, IReadOnlyList<int>>();
		ItemGroups = new ReadOnlyDictionary<string, IReadOnlyList<int>>( _ItemGroups );
	}

	public static void AddToItemGroup( string groupName, IEnumerable<int> group ) {
		if( !_ItemGroups.ContainsKey( groupName ) ) {
			_ItemGroups[groupName] = group.ToList().AsReadOnly();
		} else {
			_ItemGroups[groupName] = _ItemGroups[groupName].Union( group ).ToList().AsReadOnly();
		}
	}

	...
}
```

### What does this proposal attempt to solve or improve?
A place for mods to look for, and contribute to, listings of entities (by their `type` field's value) by a given name (e.g. any 'spooky' NPC, any explosive item, etc.).

Chicken-Bones:
Basically ore dictionary from minecraft forge. Yes

Mirsario:
From my experience with overhaul, constant string dictionary access on entities will slow things down incredibly. I would prefer just making a proper tag system (that I could use in overhaul) with OOP tag definitions that will actually just be shortcuts to generic static data, so that strings don't need to be used in contexts where performance matters. (Your usage would still be possible.)

hamstar0:
> From my experience with overhaul, constant string dictionary access on entities will slow things down incredibly. I would prefer just making a proper tag system (that I could use in overhaul) with OOP tag definitions that will actually just be shortcuts to generic static data, so that strings don't need to be used in contexts where performance matters. (Your usage would still be possible.)

Could you give a code example? I'm unclear on the meaning of tags in this context.

Keep in mind part of this idea is that modders won't always know to assign their items, NPCs, etc. (entities) to groups in their mods. This method lets anyone assign any given entity (by its `type`) to any given group, and thus crowd sourced group definitions (I'm going to assume there won't be issues with this concept, since mods are already practically naked to one another to begin with).

Mirsario:
[...]

PR #1083, opened 2020-09-12
[Draft] Content tags

Mirsario:
### What is the new feature?
This PR solves #1004, by introducing content tags. Content tags are basically booleans associated with content types (NPC, Item, etc.) and integer IDs. 

It lets content creators describe their content for each other. Content tags were the basis of much of functionality of Terraria Overhaul. For example, it used tags to determine whether or not a tile is flammable, or what kind of footstep sound should things have. A much improved system being in tML would help improve compatibility between many mods. Off the top of my head, I could guess that something like Vein Miner could use being able to determine whether something's an ore or not.

I'm not sure. The designs here have been made with performance as the highest focus. If you have ideas for improving their usage - let me know.

You first cache a TagData like this:
```cs
public static class MyTileTags
{
	/// <summary> Affects footstep sounds. </summary>
	public static readonly TagData Dirt = ContentTags.Get<TileTags>(nameof(Dirt));
}
```
(It's totally safe to call ContentTags.Get in a static constructor. I would like to have TagData instances persist between reloads, although their contents will get cleared.)

Setting a tag's value:
```cs
Dirt.Set(TileID.Dirt, true);
```

Getting a tag's value:
```cs
if (Dirt.Has(TileID.Dirt)) {
	//...
}
```

Getting a list of IDs associated with the tag:
```cs
IReadOnlyList<int> list = Dirt.GetEntries();
```

The 1.4 rebuild of Overhaul already uses this PR through 1.4_mergedtesting:

[OverhaulTileTags.cs](https://github.com/Mirsario/TerrariaOverhaul/blob/1.4/Common/Tags/OverhaulTileTags.cs) (Note that that currently uses extensions.)
[FootstepSystem.cs](https://github.com/Mirsario/TerrariaOverhaul/blob/c559da6e8f06b2cbf098b49e3b6504d343798a25/Common/Systems/Footsteps/FootstepSystem.cs#L41)

### ExampleMod updates
No examples were added yet, but I'll touch on many files, just to have newcomers even accidentally copypaste appropriate tagging into their mods.


Mirsario:
I should definitely make tag definition be per-type, actually. Fields really don't scale well when you think of default definitions accompanying them 99% of the time.

JavidPack:
Closing since #4381 should cover most of the desired functionality.

PR #1084, opened 2020-09-12
The Grand Sound Playback Reimplementation

Mirsario:
## Summary
This is a re-implementation and redesign of both vanilla and modded sound playback.

Originally this was also meant to fix sounds in general, but the branch went stale for a while, resulting in #1920 to be made and merged way sooner. Should've been done so from the start, to be honest. Now this PR's goals are only that of redesigns and bettering.

### Sound playback unification
Previously, vanilla's legacy sound styles, vanilla's "modern" (tracked) sound styles, and TML's mod sound styles, all used to play in completely different ways, barely sharing any code. This PR makes all 3 styles use the "modern" sound playback, and aims to heavily improve that system, so that everything is possible under it.

With approval of all asked contributors & content mod developers, I have gone ahead and converted all legacy sound styles and their logic to the new modern playback system, via basically redoing all of SoundID. There are some mistakes that are noted as comments in SoundID.TML.cs, but this actually turned out not as scary as originally thought -- `LegacySoundPlayer.PlaySound`'s convoluted logic (which is now never used), was perfectly reproducible in simplistic data-driven form.

### Sound Styles modernization
The `LegacySoundStyle`, `CustomSoundStyle`, and the abstract `SoundStyle` classes have all been replaced with the `SoundStyle` struct record, which uses sound effects from `AssetRepository`s.

The `SoundStyle` structure provides a lot more functions and control over sound playback than anything before.

### ModSound
`ModSound` has been removed for the following reasons:
- Had barely any functionality.
- Was too tied to specifically modded sounds and nothing else.

Modernized sound styles have enough customization in them to replace most common usages of ModSound.
A replacement for implementing sound logic (which, again, `ModSound` *barely* allowed to define) is pending.

### Other things
There have been plans for:
- Making the SoundStyle structure extendable. Maybe via linked list styled pointers? Idk yet, tired.
- Making ActiveSound instances programmable.

# Sample Usage

```cs
// Definition of a mod SoundStyle.

public static readonly SoundStyle DodgerollSound = new($"{nameof(TerrariaOverhaul)}/Assets/Sounds/Player/Armor", 3) {
	Volume = 0.65f,
	Pitch = -0.1f,
	PitchVariance = 0.2f
};

// Definition of custom vanilla sound styles.

public static readonly SoundStyle HighPitchedExplosionSound = ItemID.Item14 with {
	Volume = 0.9f,
	PitchRange = (1.5f, 1.9f),
};
public static readonly SoundStyle PotionBoilSound = new("Terraria/Sounds/Item_", 85, 3) {
	Volume = 0.5f,
	MaxInstances = 3,
	RestartIfPlaying = false,
};
public static readonly SoundStyle MegaTink = new("Terraria/Sounds/Tink", 0, 3) {
	Volume = 0.5f,
	PitchRange = (-0.5f, -0.25f),
};

// Playback

SoundEngine.PlaySound(DodgerollSound, Main.LocalPlayer.position); // Heard at a specific position

SoundEngine.PlaySound(HighPitchedExplosionSound); // Heard everywhere

// Making a vanilla item play that sound when it's used.

item.UseSound = PotionBoilSound;

// Setting a ModTile's hit sound

HitSound = MegaTink;
```


fakelag:
What is the status of this currently?

Mirsario:
> 
> 
> What is the status of this currently?

Basic implementation, in need of feedback.

fakelag:
> > What is the status of this currently?
> 
> Basic implementation, in need of feedback.

Does this add some kind of api so that it can be tested from ExampleMod (or is this just internal)? If It does, do you think you could include some usage examples?

Also, about music / MusicLoader, is that being worked on in some separate branch?

Mirsario:
> Does this add some kind of api so that it can be tested from ExampleMod (or is this just internal)? If It does, do you think you could include some usage examples?

Yes, but so far I'm testing this branch with overhaul instead - https://github.com/Mirsario/TerrariaOverhaul/tree/1.4

> Also, about music / MusicLoader, is that being worked on in some separate branch?
No, no one started work on that yet.

patches/tModLoader/Terraria/ModLoader/ModSoundStyle.cs
```diff
@@ -0,0 +1,45 @@
+using Microsoft.Xna.Framework.Audio;
+using System.IO;
+using Terraria.Audio;
+
+namespace Terraria.ModLoader
+{
+	public class ModSoundStyle : SoundStyle
```
Solxanich, line 7: Just a minor comment: implement existing IModType?
Both ModTileEntity and This derive a vanilla base, so would be good to see consistency across the two in use of that interface.

Mirsario: This is a struct-like class, unrelated to ModTileEntity.

patches/tModLoader/Terraria/Audio/SoundEngine.cs.patch
```diff
@@ -9,6 +18,75 @@
  		}
  
  		public static void Load(IServiceProvider services) {
+@@ -60,13 +_,20 @@
+ 			PlaySound(type, (int)position.X, (int)position.Y, style);
+ 		}
+ 
+-		public static SoundEffectInstance PlaySound(LegacySoundStyle type, Vector2 position) => PlaySound(type, (int)position.X, (int)position.Y);
++		public static SoundEffectInstance PlaySound(ISoundStyle type, Vector2 position) => PlaySound(type, (int)position.X, (int)position.Y);
```
DarioDaF, line 26: Proposal to add an parameter `bool? farAway` specified between for panning in a static way (far away target located in the direction of normalized position difference on the frame this is called) and have `null` be the default used for vanilla effects, so that a mod can inject here and detect if the mod author forced on or off dynamic panning or didn't care (null is the same as true in this case to preserve vanilla behaviour, but can be easily be detoured to be false)

PR #1090, opened 2020-09-15
1.4 Potential Fix #1087

### LINKED ISSUE #1087: ModWorld.Initialize isn't called prior to NetReceive in multiplayer
### Description
is it a design flaw of tml? ModWorld, in it:
`public static Dictionary<string, bool> toolEnabled;`

in the `Initialize` hook:
`toolEnabled = new Dictionary<string, bool>();`

In multiplayer, if the client just goes straight into multiplayer and joins a server, he will get stuck forever because `toolEnabled` is null during `NetReceive` (where it reads the values from the server), because `Initialize` isn't called prior. You can fix it by initializing the variable before receiving yourself though, but shouldn't `Initialize` be called atleast once before starting to receive things, as `Initialize` implies that it initializes all variables for the ModWorld to be fully usable?

Code that would freeze:
```cs
toolEnabled["first"] = reader.ReadBoolean();
```

Code that wouldn't freeze:
```cs
toolEnabled = new Dictionary<string, bool>();
toolEnabled["first"] = reader.ReadBoolean();
```

### Expected behavior 
`Initialize` hook being ran before receiving data from the server

### Actual behavior
`NetReceive` is ran before `Initialize`

ModernMAK:
### What is the bug?
See #1087 for more details
`ModWorld.NetReceive` is ran before `ModWorld.Initialize`

### How did you fix the bug?
When `Connection.State` is 4, `Initialize` is called. Specifically, `NetPlay.InnerClientLoop` calls `WorldGen.clearWorld()` which eventually calls `Initialize`.

`NetReceive` seems to be called as early as step 3. Specifically inside `MessageBuffer.GetData`, calling `WorldIO.ReceiveModData` which eventually calls `NetRecieve`.

To remedy this, I added a check that the client has reached step 4 before `WorldIO.ReceiveModData`. This should make it so that Initialize will always run before NetRecieve.

I experimented with ensuring `SetupWorld` was called before the `TCPClientLoop` thread spun up, but that was a band aid fix. When `Initialize` is called in step 4, any changes made between the start of the thread and step 4 would be lost.

There may be other solutions, but this seemed to be the smallest fix which directly fixes the issue.

### Notes
My tests showed that `NetSend` doesn't seem to have this issue, so I didn't add a check for it.


Mirsario:
That fix looks super weird. Isn't not reading data going to completely screw up the reads after that? It would be fine if mod data was the last thing in the packet.

Mirsario:
..Yes, you should ensure in NetMessage.cs that SendModData is only called for clients with the right state.

ModernMAK:
Drafted to continue discussion but avoid a proper pr

ModernMAK:
> That fix looks super weird. Isn't not reading data going to completely screw up the reads after that? It would be fine if mod data was the last thing in the packet.

The first read (status 3) corrupts lobbyid, and Sandstorm intended severity; but when the info is sent again (status 10) it is corrected. I believe both steps happen before the player actually enters the world, so the player wouldn't notice any changes outside of errors caused by the corruption, it's such a small window to debug that I can't really test it by deubbing.

But it looks like if a friend is extremely fast, they can join the wrong lobby id; don't know what that would do, but it probably wouldn't be good.

Even with a check in NetMessage; It may still be possible for a server to send a message to a client that isn't ready yet. Due to the server having completed the step but the client not. 

The proposed solution shouldn't have any crashes, and will eliminate the associated issue; BUT, I can imagine a case where a mod sends data the first send, and doesn't resend it expecting it to have been sent, when the value was discarded due to client not being ready.

Shifting the modworld data to the end of the packet fixes corruption, but requiring it be the end of the packet might cause issues down the road. This wont fix mod data being discarded.

A hack which could work around the issue; writing a length for mod world data, and skipping that number of bytes. This would allow it to be placed anywhere in the packet and safely be ignored; but it would require writing to a separate buffer, then copying from the separate buffer to the main one. Not to mention this doesn't fix the issue of data being sent and then discarded.

So I guess ill keep digging into the netcode to find some better alternatives...

Mirsario (approved):

Chicken-Bones:
This fix is a correct implementation because packet 7 (to which ModWorld data is appended) is sent twice during the connection setup, once in response to packet 6 (on entering connection state 4) and once in response to packet 8 (on entering connection state 6). State 4 is "requesting world data" and state 6 is "world cleared"

I would like to see some comments on the `&& Netplay.Connection.State > 4)` part of the check, and preferably, we could avoid sending the data by setting the `int number = 0` param to `1` when calling `TrySendData(7` in response to packet 6

PR #1096, opened 2020-09-20
Fix CraftItem not creating new instances of items

Itorius:
Main.CraftItem now creates a new instance of an item. Also adds several ModItem Clone and NewInstance changes.

Previously the only way for an instanced item to work with crafting was to override `OnCraft` and re-implement your constructor. This way crafting respects the `CloneNewInstances` flag

### DIFF
--- a/ExampleMod/Content/Items/ExampleInstancedItem.cs
+++ b/ExampleMod/Content/Items/ExampleInstancedItem.cs
@@ -1,0 +1,48 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.ModLoader;
+using Terraria.ModLoader.IO;
+
+namespace ExampleMod.Content.Items
+{
+	public class ExampleInstancedItem : ModItem
+	{
+		public override string Texture => "ExampleMod/Content/Items/ExampleItem";
+
+		public override bool CloneNewInstances => false;
+
+		public override ModItem Clone(Item item) {
+			ExampleInstancedItem clone = (ExampleInstancedItem)base.Clone(item);
+			clone.colors = (Color[])colors.Clone();
+			return clone;
+		}
+
+		public Color[] colors;
+
+		public ExampleInstancedItem() {
+			colors = new Color[5];
+			for (int i = 0; i < 5; i++) {
+				colors[i] = Main.hslToRgb(Main.rand.NextFloat(), 1f, 0.7f);
+			}
+		}
+
+		public override void ModifyTooltips(List<TooltipLine> tooltips) {
+			for (int i = 0; i < colors.Length; i++) {
+				TooltipLine tooltipLine = new TooltipLine(Mod, "EM" + i, "Example " + i);
+				tooltipLine.overrideColor = colors[i];
+				tooltips.Add(tooltipLine);
+			}
+		}
+
+		public override TagCompound Save() =>
+			new TagCompound { ["Colors"] = colors.ToList() };
+
+		public override void Load(TagCompound tag) {
+			colors = tag.GetList<Color>("Colors").ToArray();
+		}
+
+		public override void AddRecipes() => CreateRecipe().AddIngredient<ExampleItem>(10).Register();
+	}
+}
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2438,9 +2438,25 @@
  
  			new Microsoft.Xna.Framework.Color(150, 150, 150, 150);
  			if (mouseX >= num && (float)mouseX <= (float)num + (float)TextureAssets.InventoryBack.Width() * inventoryScale && mouseY >= num2 && (float)mouseY <= (float)num2 + (float)TextureAssets.InventoryBack.Height() * inventoryScale && !PlayerInput.IgnoreMouseInterface) {
-@@ -31247,8 +_,11 @@
+@@ -31236,19 +_,26 @@
+ 			UILinkPointNavigator.SetPosition(6000 + gamepadPointOffset, vector + rectangle.Size() * 0.65f);
+ 		}
+ 
++		// while in SetDefaults it should be a clone of the base modItem or a new instance
++
+ 		public static void CraftItem(Recipe r) {
+ 			int stack = mouseItem.stack;
+-			mouseItem = r.createItem.Clone();
++			mouseItem = (Item)mouseItem.ShallowClone();
++			mouseItem.modItem = r.createItem.modItem?.NewInstance(mouseItem);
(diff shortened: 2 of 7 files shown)


patches/tModLoader/Terraria/ModLoader/ModItem.cs
```diff
@@ -102,22 +100,24 @@ public class ModItem:ModTexturedType
 		/// </summary>
 		/// <param name="item">The item being cloned</param>
 		/// <param name="itemClone">The new item</param>
-		public virtual ModItem Clone(Item item) => NewInstance(item);
+		public virtual ModItem Clone(Item item) {
+			ModItem clone = Clone();
+			clone.item = item;
+			return clone;
+		}
```
Chicken-Bones, line 107: This looks like a correct change. However I'd like you to replace `Clone()` there with `MemberwiseClone` and remove the no-args clone method (it's not needed)

Chicken-Bones (commented):
There's also a method on `Item` called `DeepClone` which looks new in 1.4. We'll need to make sure ModItem supports this

Chicken-Bones:
While you're at it, can we fix the mess which is reforging?
I suggest adding a method on `Item` called `ResetStats` which generates a new `Item` with `SetDefaults` and then copies over the `damage, useAnimation, useTime, reuseDelay, mana, knockBack, scale, shootSpeed, crit, value` fields. Then add the same hook in Mod/GlobalItem, and pass the 'default item' for reference

patches/tModLoader/Terraria/ModLoader/ModItem.cs
Itorius: Yeah, I was wondering why both methods exist. Will Do!

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2438,6 +2438,18 @@
...
  			if (mouseX >= num && (float)mouseX <= (float)num + (float)TextureAssets.InventoryBack.Width() * inventoryScale && mouseY >= num2 && (float)mouseY <= (float)num2 + (float)TextureAssets.InventoryBack.Height() * inventoryScale && !PlayerInput.IgnoreMouseInterface) {
+@@ -31238,7 +_,10 @@
+ 
+ 		public static void CraftItem(Recipe r) {
+ 			int stack = mouseItem.stack;
+-			mouseItem = r.createItem.Clone();
++			int origType = r.createItem.type;
++			int origStack = r.createItem.stack;
++			mouseItem = r.createItem;
++			r.createItem = new Item(origType) { stack = origStack };
```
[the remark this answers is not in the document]
Itorius: Prefixes should get applied after afaik. (See mouseItem.Prefix(-1)). And also if the item is e.g. colored you'd want to get the color you see.

Itorius:
> While you're at it, can we fix the mess which is reforging?
> I suggest adding a method on `Item` called `ResetStats` which generates a new `Item` with `SetDefaults` and then copies over the `damage, useAnimation, useTime, reuseDelay, mana, knockBack, scale, shootSpeed, crit, value` fields. Then add the same hook in Mod/GlobalItem, and pass the 'default item' for reference

Sure, but this should probably go into its separate PR.

Itorius:
> There's also a method on `Item` called `DeepClone` which looks new in 1.4. We'll need to make sure ModItem supports this

DeepClone returns a shallow copy (haha redcode). So it should probably return Item.Clone (shallow copy + mod and global item clone).

Itorius:
> > While you're at it, can we fix the mess which is reforging?
> > I suggest adding a method on `Item` called `ResetStats` which generates a new `Item` with `SetDefaults` and then copies over the `damage, useAnimation, useTime, reuseDelay, mana, knockBack, scale, shootSpeed, crit, value` fields. Then add the same hook in Mod/GlobalItem, and pass the 'default item' for reference
> 
> Sure, but this should probably go into its separate PR.

Opened a new issue so we don't forget #1101

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2438,6 +2438,16 @@
  
  			new Microsoft.Xna.Framework.Color(150, 150, 150, 150);
  			if (mouseX >= num && (float)mouseX <= (float)num + (float)TextureAssets.InventoryBack.Width() * inventoryScale && mouseY >= num2 && (float)mouseY <= (float)num2 + (float)TextureAssets.InventoryBack.Height() * inventoryScale && !PlayerInput.IgnoreMouseInterface) {
+@@ -31238,7 +_,8 @@
+ 
+ 		public static void CraftItem(Recipe r) {
+ 			int stack = mouseItem.stack;
+-			mouseItem = r.createItem.Clone();
++			mouseItem = r.createItem;
++			r.createItem = new Item(mouseItem.type) { stack = mouseItem.stack };
```
Chicken-Bones, line 2447: I still think this should be the other way
`mouseItem = new Item(mouseItem.type) {stack = r.createItem.stack}`
and leave `r.createItem` alone. 

The default behaviour should be for an item which changes colour on craft, to not change colour in the recipe viewer. People can implement this if they want using OnCraft (move the recipe colour into the newItem, and recolour the recipe item) but it should be opt_in_

Itorius: I suppose, I can add it to the ExampleInstanceItem.

ExampleMod/Content/Items/ExampleInstancedItem.cs
```diff
@@ -42,6 +42,12 @@ public class ExampleInstancedItem : ModItem
 		public override void AddRecipes() {
 			CreateRecipe()
 				.AddIngredient<ExampleItem>(10)
+				.AddOnCraftCallback((recipe, item) => {
+					ExampleInstancedItem craftedItem = (ExampleInstancedItem)item.modItem;
+					ExampleInstancedItem recipeItem = (ExampleInstancedItem)recipe.createItem.modItem;
+
+					Utils.Swap(ref craftedItem.color, ref recipeItem.color);
```
Chicken-Bones, line 49: give this a comment!

 ```
// By default, the recipe's colour doesn't change, and you just get a new item with a new colour whenever you craft it.
// This swap gives the crafted item the colour of the recipe, and the recipe a new colour for the next craft
```

File: patches/tModLoader/Terraria/Main.cs.patch [hunk 5]
```diff
@@ -2438,16 +2438,17 @@
...
  
  		public static void CraftItem(Recipe r) {
  			int stack = mouseItem.stack;
 -			mouseItem = r.createItem.Clone();
-+			mouseItem = r.createItem;
-+			r.createItem = new Item(mouseItem.type) { stack = mouseItem.stack };
- 			mouseItem.stack += stack;
+-			mouseItem.stack += stack;
++			mouseItem = new Item(r.createItem.type);
++			mouseItem.stack = r.createItem.stack + stack;
```
Chicken-Bones, line 2448: use the initializer syntax! It was cool

PR #1098, opened 2020-09-20
NPC and Projectile AI Style IDs

pollen00:
### What is the new feature?
NPC and Projectile AI Style IDs

AI Style IDs should be part of tModLoader because it eliminates magic numbers and makes code more readable. Same reason that TileIDs and ItemIDs exist.

Not that I can think of.

projectile.aiStyle = ProjAIStyleID.GolfBall;
npc.aiStyle = NPCAIStyleID.TownNPC;

### ExampleMod updates

I haven't updated example mod yet, but if this feature is added I will.

### Extra Notes
pbone wrote the NPC AI Style IDs and I wrote the Projectile AI Style IDs. Also note that Mello did not contribute to this, he contributed to ExampleGolfBall and I accidentally added them in my fork, Example Golf Ball is no longer in here.
If you have a suggestion to shorten the classnames please let me know. I feel like the current class names are a bit long.


patches/tModLoader/Terraria/ID/NPCAIStyleID.cs
```diff
@@ -0,0 +1,184 @@
...
+		public const short Slime = 1;
+		public const short DemonEye = 2;
+		public const short Fighter = 3;
+		public const short EyeOfCthulhu = 4;
+		/// <summary>
+		/// Includes things such as Eaters of Souls
+		/// </summary>
+		public const short Flying = 5;
+		public const short Worm = 6;
+		public const short TownNPC = 7;
```
Mirsario, line 15: I would rename this to Friendly or something, since bunnies use this too.

Mirsario: Wiki names this "Passive".

Mirsario (approved):
It seems good overall, and this has been needed for a while. Though, @Chicken-Bones, maybe constants in all new Id classes should be swapped out for static readonly fields? There also could be a special patch for that on vanilla ones.

Chicken-Bones:
Only `const` fields can be used as `case` labels

RyanMakesMods:
At first it seemed redundant, but it seems like a nice patch now

Chicken-Bones:
@pollen00 can you search ExampleMod for any aiStyle references (also in commented/unimplemented code) and replace all the values with references to the ID class?

PR #1099, opened 2020-09-20
LiquidID.cs

RyanMakesMods:
### What is the new feature?
LiquidID
To make reading Liquid Code and such easier on the eye.
No.
`liquidType == LiquidID.Water;`
### ExampleMod updates
N/A


Mirsario:
Wrong source branch.

PR #1100, opened 2020-09-20
Create LiquidID.cs

RighteousRyan1:
### What is the new feature?
LiquidID

To make reading Liquid Code and such easier on the eye.

No.

`liquidType == LiquidID.Water;`

### ExampleMod updates
N/A

### DIFF
--- a/patches/tModLoader/Terraria/ID/LiquidID.cs
+++ b/patches/tModLoader/Terraria/ID/LiquidID.cs
@@ -1,10 +1,10 @@
 namespace Terraria.ID
 {
-  /// <summary> Enumerates liquid values to make them easier to see/read </summary>
-  public static class LiquidID
-  {
-    public const int Water = 0;
-    public const int Lava = 1;
-    public const int Honey = 2;
-  }
+	/// <summary> Enumerates liquid values to make them easier to see/read </summary>
+	public static class LiquidID
+	{
+		public const int Water = 0;
+		public const int Lava = 1;
+		public const int Honey = 2;
+	}
 }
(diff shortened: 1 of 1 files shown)


direwolf420:
This already exists in the game:
```cs
public class Tile
{
    public const int Type_Solid = 0;
    public const int Liquid_Honey = 2;
    public const int Liquid_Lava = 1;
    public const int Liquid_Water = 0;

    //...
}
```

Used like: `if (tile.liquidType() == Tile.Liquid_Honey)`

Chicken-Bones:
There you have it, but also oof to determining tile ids in reverse

patches/tModLoader/Terraria/ID/LiquidID.cs
```diff
@@ -0,0 +1,10 @@
+namespace Terraria.ID
+{
+  /// <summary> Enumerates liquid values to make them easier to see/read </summary>
+  public static class LiquidID
```
Mirsario, line 4: Tabs pls

RyanMakesMods: done.

Chicken-Bones:
are we accepting this? given that the fields exist in Tile?

JavidPack:
I don't really see the point, I'm already using the Tile class consts in my mod, having two places for the same info isn't ideal.

Mirsario:
Well, the consts in Tile can be made internal.

Chicken-Bones:
/polls "Accept (make Tile.Liquid_* internal)" "Reject (modders just use Tile.Liquid_)"

Chicken-Bones:
Deferred to #1301
I think there's a stronger case for it with the rest of the property rework. It's not something I actually have an opinion on

PR #1108, opened 2020-09-28
Fixed item frames' & all other tile entities' interaction with mod content.

Mirsario:
### What is the bug?
Using mod content with item frames, weapon racks, mannequins, and 1.4 alternatives of those may break worlds. You know about this. This could be fixed years ago.

### How did you fix the bug?
- Pulled ModTileEntity.Save/Load up to TileEntity, and made vanilla tile entities' TagCompound data get saved in .twlds alongside modded ones,
- Implemented TagCompound saving & loading for all vanilla tile entities that needed that.
- Modified vanilla tile entities' Read/WriteExtraData methods to use ItemIO.Read/Write (made out of Send/Receive), which uses VarInts instead of Int16 in modded contexts.
- Non-mod TileEntities are now either sealed, or have an internal constructor, just in case.

I could keep ModTileEntity.Save/Load where it is and make new internal virtual that would call that in case of ModTileEntity, but that's dumb. I think we may want to merge ModTileEntity into TileEntity at some point.


patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.cs.patch
```diff
@@ -0,0 +1,60 @@
...
+ 		private static byte _myEntityID;
+ 		private const int MyTileID = 470;
+@@ -105,18 +_,14 @@
+ 			for (int i = 0; i < 8; i++) {
+ 				Item item = _items[i];
+ 				if (!item.IsAir) {
+-					writer.Write((short)item.netID);
+-					writer.Write(item.prefix);
+-					writer.Write((short)item.stack);
++					ItemIO.Write(item, writer, !networkSend || ModNet.AllowVanillaClients, writeStack: true);
```
Chicken-Bones, line 25: See comments in Discord about increasing the item limit. The correct thing to do here is just `if (networkSend) ItemIO.SendModData(item, writer)`

Chicken-Bones (changes requested):
Your description is great, and the methodology sound, fix up the data sending patches, and I'll give it another pass through.

I assume the code in `LoadTileEntities` is correct, vanilla tile entities load first, with their vanilla copy of the data, and it just gets ovewritten if they have valid tmod data

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.cs.patch
Mirsario: This is already correct. The write method writes mod data only if vanilla write boolean is false, and in this case it's only false if this is a network send with no vanilla compatibility.

Mirsario: Reverted these patches and made OOP overrides for modded contexts.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -550,56 +550,68 @@ internal struct ContainerTables
...
-					newEntity.type = (byte)tileEntity.Type;
-					newEntity.Position = new Point16(tag.GetShort("X"), tag.GetShort("Y"));
-					((UnloadedTileEntity)newEntity).SetData(tag);
-				}
-				if (tileEntity.ValidTile(newEntity.Position.X, newEntity.Position.Y)) {
-					newEntity.ID = TileEntity.AssignNewID();
-					TileEntity.ByID[newEntity.ID] = newEntity;
-					if (TileEntity.ByPosition.TryGetValue(newEntity.Position, out TileEntity other)) {
+
+				if (baseModTileEntity != null && baseModTileEntity.ValidTile(tileEntity.Position.X, tileEntity.Position.Y)) {
```
Chicken-Bones, line 606: Maybe this does do what's requested above?

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -550,56 +550,68 @@ internal struct ContainerTables
...
+					tileEntity.ID = TileEntity.AssignNewID();
+					TileEntity.ByID[tileEntity.ID] = tileEntity;
+
+					if (TileEntity.ByPosition.TryGetValue(tileEntity.Position, out TileEntity other)) {
```
Chicken-Bones, line 610: ..how?

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEHatRack.TML.cs
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		public override void Load(TagCompound tag) {
+			PlayerIO.LoadInventory(_items, tag.GetList<TagCompound>("items"));
+			PlayerIO.LoadInventory(_dyes, tag.GetList<TagCompound>("dyes"));
+		}
+
+		public override void NetSend(BinaryWriter writer, bool lightSend) {
+			writer.Write(BitsByte.ComposeBitsBytesChain(false, _items.Select(i => !i.IsAir).ToArray())[0]);
+			writer.Write(BitsByte.ComposeBitsBytesChain(false, _dyes.Select(i => !i.IsAir).ToArray())[0]);
```
Chicken-Bones, line 23: Ouch, I know vanilla does this (much faster without the lambdas and enumerables) but dang. I guess it's worth it? `ItemIO.Send` should (probably doesn't) send only the one VarInt if the id is 0.

Chicken-Bones: hmm, yeah lets keep this, but I'd rather use something slightly more performant, 2 for loops even.

Chicken-Bones (commented):
Looks good overall.

patches/tModLoader/Terraria/ModLoader/ModTileEntity.cs
```diff
@@ -191,53 +161,25 @@ public abstract class ModTileEntity : TileEntity, IModType
 			NetReceive(reader, networkSend);
 		}
 
+		// The base implementations of these methods call Read/WriteExtraData, and those in ModTileEntity's case now call NetSend/Receive.
+		public override void NetSend(BinaryWriter writer, bool lightSend) { }
+		public override void NetReceive(BinaryReader reader, bool lightReceive) { }
```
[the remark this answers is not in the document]
Mirsario: That was there in 1.3, and this is an override. I believe these booleans are only false on the first send/receive for that specific TE.

patches/tModLoader/Terraria/ModLoader/ModTileEntity.cs
```diff
@@ -101,14 +74,11 @@ public abstract class ModTileEntity : TileEntity, IModType
 		/// You should never use this. It is only included here for completion's sake.
 		/// </summary>
 		public override void NetPlaceEntityAttempt(int i, int j) {
```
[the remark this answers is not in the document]
Mirsario: Ye, this is repeated a bunch of times. Making TEs not crap is another PR's job. I think Itorius was keen on doing something like that.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
Mirsario: image
Yes

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
Mirsario: How what? I think I didn't write this part.

Chicken-Bones: yeah... weird... I guess it can stay

Chicken-Bones (approved):

Issue #1110, opened 2020-09-30
Add new instance creation hook to ModItem to support declarative spawn sources and tables


Chicken-Bones:
## Summary
- Remove `CloneNewInstances`
- All new instances are cloned (either from the root `ModItem`, `Recipe.createItem` or a drop table, etc)
- Replace `OnCraft` with more generic `OnCreated(source, ref bool rollPrefix)`

## Details
When a player 'creates' an item by crafting, buying, opening a bag, or an item enters the world in another way such as loot drop or chest spawn, it might be time to roll some unique attributes. In vanilla, this just means prefixes, but mod items may want to randomise the colour, or add other properties. 

The tML pattern for modded content is to register a 'template' instance, normally autoloaded, and then to clone the template and attach it to the matching in world vanilla class. Modders familiar with OOP and struggling to deal with item identity, were frustrated with the restrictions of cloning (particularly the constructor being effectively ignored) and thus `CloneNewInstances` was born, defaulting to `false` the constructor and `SetDefaults` could handle the vast majority of content.

However, when creating recipes, it's sometimes desirable to make recipes which create a specific variant of an item, one with a specific prefix, or a unique name, or better stats. Recipes are declarative, with the result stored as a unique template instance which is shown in the crafting UI, and to support this declarative style, crafted items are always cloned from the recipe `createItem`, ignoring the `CloneNewInstances` flag. 

As seen in #1096 this clashes with the intended constructor pattern and requires modders to override `OnCraft` to point to their 'new variant' code. 

With tML moving to more declarative spawning mechanics (loot tables for NPCs, chests, and boss bags) we need a better solution.

I propose a new hook, `OnCreated(object source, ref bool rollPrefix)` where `source` can be one of `Recipe` `NPC` `LootBagOpen` `ChestSpawn` `PurchaseEvent` `BlockMined`. We can then remove the `CloneNewInstances` and `OnCraft` hooks, and make cloning from a template followed by a call to `OnCreated` be the default behaviour. 

Things to be discussed:
Naming
Common interface for creation sources?
How this might affect modders and their content

Ozzatron:
There is nothing inherently special about crafting if all item creation events will be unified to cloning ModItem templates. However modders will want to know through which method the item is being created, and perhaps a lot of contextual information.

[...]

PR #1118, opened 2020-10-07
Add new instance creation hook to ModItem to support declarative spawn sources and tables

Itorius:
See #1110 

Future creation contexts:
Drop from entity (depends on loot table PR #1063)
Boss bag opening
NPC Shops (depends on NPC shop rewrite [Issue + PR pending])
Populating chests on worldgen


ExampleMod/Content/Items/ExampleInstancedItem.cs
```diff
@@ -0,0 +1,52 @@
...
+			for (int i = 0; i < 5; i++) {
+				colors[i] = Main.hslToRgb(Main.rand.NextFloat(), 1f, 0.7f);
+			}
+		}
+
+		public ExampleInstancedItem() {
+			colors = new Color[5];
+			for (int i = 0; i < 5; i++) {
+				colors[i] = Main.hslToRgb(Main.rand.NextFloat(), 1f, 0.7f);
+			}
```
Chicken-Bones, line 33: this is the kind of duplication we want to avoid. We could call `OnCreate(null)` here, but ideally we wouldn't use the constructor at all. The only places where `OnCreate` isn't called is in declarative templates like recipes or loot tables. Maybe we should make a creation context for this? Unsure

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -1719,9 +1721,7 @@ public static bool CanBurnInLava(Item item)
 				if (!item.InstancePerEntity)
 					throw new Exception(type + " has instance fields but does not set InstancePerEntity to true. Either use static fields, or per instance globals");
 
-				if (!item.CloneNewInstances &&
-						!HasMethod(type, "NewInstance", typeof(Item)) &&
-						!HasMethod(type, "Clone", typeof(Item), typeof(Item)))
+				if (!HasMethod(type, "Clone", typeof(Item), typeof(Item)))
 					throw new Exception(type + " has InstancePerEntity but must either set CloneNewInstances to true, or override NewInstance(Item) or Clone(Item, Item)");
```
Chicken-Bones, line 1725: Update this message

Chicken-Bones (commented):
Looks like the right direction, need to add the other creation contexts

PR #1119, opened 2020-10-07
Added a regex plural method

pollen00:
### What is the new feature?
I added a method that can get the correct plural/singular noun.

There is currently no support for plurals and singulars, and this would fix that issue.

Possibly, but this was the best one I found.

The method returns a string, so you just need to give it the string and value, I might want to change the parameter name.
LanguageManager.Instance.RegexPlural("string", int))

### ExampleMod updates

I didn't update example mod since this was meant to be used with Mod Menu and Mod Browser, but it would still function in ExampleMod.


Chicken-Bones:
Honestly this looks great. Did you create this yourself or get inspiration from somewhere? If there's an original source (blog post or library) that you've improved on, can you put it in a comment on your `pluralizationPatternRegex`, or at least reference it here?

Mirsario:
This is nice now with jopo's rewrite. There may still be a need for optimizations, though.

Though, I now wonder if there's a reason why we shouldn't make languages somewhat OOP (and therefore redo that language switch with virtuals). I saw many people wanting to localize the game into languages that aren't supported, and there's just no simple way of doing that right now.

@Chicken-Bones jopo's rewrite has this linked in comments. https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals

pollen00:
Unless there is another place where plurals need to be fixed, this should be done.

JavidPack:
Thanks everyone.

Jofairden:
Very nice! Always a fan of localization improvements 👍

PR #1120, opened 2020-10-07
New ModifyTreasureFairySearch hook

Agrair:
### What is the new feature?
A new feature to change how treasure fairies will react to tiles.

It allows modders to increase or decrease the amount of tiles for a fairy to pick a chunk of ore, and also to discount a chunk of ore entirely.

If you look closely, this actually only works for ores at the moment. There could probably be some more hooks to change how it decides between different tiles (atm, the fairy will almost always pick a chest over a chunk of Example Ore).

If a modder wants their ore to seem "extra special", or dynamically available to fairies depending on various conditions, this hook lets them do that.

### ExampleMod updates
I just added some stuff to Example Ore to make it work for fairies.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/NewRecipeFinder.cs
+++ b/patches/tModLoader/Terraria/ModLoader/NewRecipeFinder.cs
@@ -1,0 +1,291 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Terraria.Localization;
+using Terraria.ModLoader.Exceptions;
+
+namespace Terraria.ModLoader
+{
+	public class NewRecipeFinder : IEnumerable<Recipe>
+	{
+		private List<Item> items = new List<Item>();
+		private List<int> tiles = new List<int>();
+		private List<RecipeGroup> groups = new List<RecipeGroup>();
+		private List<Recipe.Condition> conditions = new List<Recipe.Condition>();
+		private Item result;
+
+		private NewRecipeFinder() {
+		}
+
+		public static NewRecipeFinder Create(int itemID, int stack = 0) {
+			if (itemID <= 0 || itemID >= ItemLoader.ItemCount)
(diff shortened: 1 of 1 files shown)


Agrair:
I think if you wanted to override how it reacted to other tiles in relation to the one specified the hook could just be given a "previousTile" parameter or something.

Chicken-Bones:
Current implementation allows modders to make fairies search further for their ore, and alter how much is required for it to be considered. I assume your alternate design is in addition to the current one, because the current one is good. I don't know much about how fairies work, but I think the switching hooks should probably be merged, because you really just want to define a priority system, and to do that you need to know both the current and the new possibility. Then switch if any mod votes yes, unless there's a disagreement.

Agrair:
Well the way the fairy search works is that it just iterates over all nearby tiles around the player, checking if each one has the FriendlyFairyCanLureTo and MetalDetectorValue flags set. If it’s an ore, it will check a space around the tile to see if it’s an ore chunk or a lonely ore tile. The current implementation lets you change the cutoff for the ore chunk for be considered valid. Then, the fairy will ask itself if the tile it’s looking at has a higher value than the one it’s already decided on. For ore chunks, size doesn’t actually matter, only to a point to get it considered like I’ve said. Problem is for a priority system is that the fairy won’t create a list of options and pick one; it’ll just procedurally check every single candidate. So if Example Ore decided to make sometimes make itself more valuable than Palladium Ore, it could make it so that the fairy always picks Example Ore over Palladium, but if it selected a chest and *then* the Palladium, there would be no way to tell Example Ore was even there. Tl;dr I’m not sure how to do this. I also think some function for normal Metal Detectors should be added, but ahhh.

Agrair:
Ok I wracked my head and I couldn’t come up with a solution. I’ll fix the conflicts and then the PR should be merged.

Chicken-Bones:
There's several places where a general purpose priority system is needed. Not sure if we should merge before we get that up and running

PR #1121, opened 2020-10-07
update Example Player Draw Layer

Agrair:
For some reason this wasn’t done already. This just updates the file to use the new OOP model.


Agrair:
Actually I should probably readd that comment I deleted.

PR #1122, opened 2020-10-07
Removed "using static" from ExampleMod (1.4)

Chik3r:
Removed `using static` from ExampleMod code, so that people don't get confused when `ItemType<>()` isn't working.

No


Chicken-Bones (approved):
This has been voted for by many in Discord. Let me know when it's ready and I'll merge

Chik3r:
It should be ready, I already removed all of the `using static`s. And it builds and runs.
Also, the PRs that are adding things to ExampleMod should stop using `using static` if we are going to remove them.

Issue #1105, opened 2020-09-26
Add a NaturalManaRegen hook to ModPlayer


IDGCaptainRussia94:
### Description
I noticed lately that ManaRegen is quite hardcoded into the UpdateManaRegen method, and I was wondering if we could get a ModPlayer hook similar to NaturalLifeRegen for it

### What does this proposal attempt to solve or improve?
Grant more control over mana regen, like say if you want a debuff to halve regen rates
### Which (other) solutions should be considered?
player.manaRegenBonus can be used to boost regen by a set number, but this is rather limited, and the only means rn is to IL Edit or method swap the method with your own code, which replaces the standard mana regen code and isn't very effective, especially if other mods want to do the same thing.

PR #1123, opened 2020-10-08
Adding natural mana regen hooks to tmodloader

Jeremy-Laier:
### What is the new feature?
referencing this feature request https://github.com/tModLoader/tModLoader/issues/1105

Adding a player hook for natural mana regen, just like natural life regen.

Give modders a more intricate way of messing around with mana.

ATM player.manaRegenBonus exists although it is not as in depth as this hook would provide 

```cs
public class ExamplePlayer: ModPlayer {
                // for example this would double natural player regen 
		public override void NaturalManaRegen(ref float flatRegen, ref float multiRegen) {
			multiRegen *= 2;
                        flatRegen = 0; // doesn't change mana rate at all as value defaults to 0 in Player.cs
		}
}
```

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -1505,15 +1505,30 @@
  			float num4 = (float)statLifeMax2 / 400f * 0.85f + 0.15f;
  			num2 *= num4;
  			lifeRegen += (int)Math.Round(num2);
-@@ -13071,6 +_,8 @@
+@@ -13069,8 +_,9 @@
+ 				manaRegen = statManaMax2 / 7 + 1 + manaRegenBonus;
+ 				if ((velocity.X == 0f && velocity.Y == 0f) || grappling[0] >= 0 || manaRegenBuff)
  					manaRegen += statManaMax2 / 2;
- 
+-
++				
  				float num2 = (float)statMana / (float)statManaMax2 * 0.8f + 0.2f;
-+				PlayerHooks.NaturalManaRegen(this, ref num2);
 +
  				if (manaRegenBuff)
  					num2 = 1f;
  
+@@ -13080,7 +_,11 @@
+ 				manaRegen = 0;
+ 			}
+ 
+-			manaRegenCount += manaRegen;
++			float flatManaRegenNum = 0;
++			float multiManaRegenNum = 1;
++			PlayerHooks.NaturalManaRegen(this, ref flatManaRegenNum, ref multiManaRegenNum);
++
++			manaRegenCount += (int) ((manaRegen + flatManaRegenNum) * multiManaRegenNum);
+ 			while (manaRegenCount >= 120) {
+ 				bool flag = false;
+ 				manaRegenCount -= 120;
 @@ -13108,8 +_,8 @@
  
  		public void UpdateJumpHeight() {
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Intended usage should always use a `*=` so that multiple mods with the same effect stack, and don't override eachother. I also suggest changing the hook to include both the base and the multiplier. So that mods can additive stack on the base `manaRegen` value, and multiply on `num2`

Jeremy-Laier:
> Intended usage should always use a `*=` so that multiple mods with the same effect stack, and don't override eachother. I also suggest changing the hook to include both the base and the multiplier. So that mods can additive stack on the base `manaRegen` value, and multiply on `num2`

I think I placed the mana changes in the wrong spot in the mana equation. I decided to add them at the VERY end

https://terraria.gamepedia.com/Mana 

I suppose the multiplier could be seen as `1.15` in the wiki formula

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1505,6 +1505,30 @@
...
+-
++				
+ 				float num2 = (float)statMana / (float)statManaMax2 * 0.8f + 0.2f;
++
+ 				if (manaRegenBuff)
+ 					num2 = 1f;
+ 
+@@ -13080,7 +_,11 @@
+ 				manaRegen = 0;
+ 			}
```
Jeremy-Laier, line 1521: Note this is what was there before, diff just shows bc I originally had changes there, but took em out

Jeremy-Laier:
@Chicken-Bones

patches/tModLoader/Terraria/Player.cs.patch
Mirsario: It's shown there because you or your editor added tabs to an empty line. Remove those and rediff.

Mirsario:
> ATM player.manaRegenBonus exists although it is not as in depth as this hook would provide

In my opinion, we're better off upgrading fields than making hooks, as it would be a pain to use ModPlayer hooks for, say, effects of a buff (that already has its own class). We're trying to improve designs, after all. @Chicken-Bones What about just using DamageModifiers here?

Jeremy-Laier:
> > ATM player.manaRegenBonus exists although it is not as in depth as this hook would provide
> 
> In my opinion, we're better off upgrading fields than making hooks, as it would be a pain to use ModPlayer hooks for, say, effects of a buff (that already has its own class). We're trying to improve designs, after all. @Chicken-Bones What do you think about just using DamageModifiers here and in life regen?

It seems that chicken thinks at least the hook addition is a good idea, he added a label to it on the issue here https://github.com/tModLoader/tModLoader/issues/1105

patches/tModLoader/Terraria/Player.cs.patch
Jeremy-Laier: ty will do

Chicken-Bones:
DamageModifier sounds like a great idea here. While we're at it I think we should just change that class name to Modifier

Solxanich:
@Jeremy-Laier - are you intending to return to this in the future?
If not, I'd like to pick this up and wrap it up on your behalf, 
Although I would note that you may require to close this as I can't seem to pull from upsteam in to this fork, and I'll have to start from a new branch...

Jeremy-Laier:
> @Jeremy-Laier - are you intending to return to this in the future?
> If not, I'd like to pick this up and wrap it up on your behalf,
> Although I would note that you may require to close this as I can't seem to pull from upstream in to this fork, and I'll have to start from a new branch...

All yours, thanks for being polite and asking. I'm going to close this PR as well

Issue #1125, opened 2020-10-08
[1.3] Impose ModRecipe limitations so its methods are only called during recipe setup phase


direwolf420:
### Description
Lately I've seen more and more people mistakenly put code that usually goes into the `ModItem.AddRecipes` hook into `ModItem.SetDefaults` instead, which leads to a stack overflow when the mod is loaded because `ModRecipe.SetResult` invokes `ModItem.SetDefaults`. That makes it hard for people not familiar with the API (and modding in general) to figure out what's wrong as theres no errors visible in the logs, nor a crash window that contains any useful info. I suggest locking certain methods (`ModRecipe.SetResult` for example) from being called in a context that isn't recipe related (limit to around where `RecipeHooks.AddRecipes()` is called for example). And if they are still called detect this when the mod is loaded, and give a descriptive error message.

### Reproduction
recipe.SetResult(this); in SetDefaults of a ModItem
Code example: https://pastebin.com/ZzzPLHEA
```cs
using Terraria.ID;
using Terraria.ModLoader;
namespace TEST.Items {
    class TESTBOW : ModItem {
        public override void SetDefaults() {
            ModRecipe recipe = new ModRecipe(mod);
            recipe.AddIngredient(ItemID.WoodenBow, 1);
            recipe.SetResult(this);
            recipe.AddRecipe();
        }
    }
}
```

### 1.4
Could do something similar with calling `CreateRecipe` perhaps

Chicken-Bones:
Approved, PR, lets go

PR #1129, opened 2020-10-13
Add setupRecipes bool to detect ModRecipe misuse

direwolf420:
### What is the bug?
"fixes" #1125 

### How did you fix the bug?
By introducing an additional check inside the `ModRecipe` constructor

Do it in `ModRecipe.SetResult` instead, which is the origin of the stack overflow


Chicken-Bones:
1.4 edition? And please localise the error message

direwolf420:
Will do once I get around to make my 1.4 fork work again

PR #1131, opened 2020-10-14
[1.4] OnSpawn hooks

Itorius:
*@Mirsario speaking:*

(PR changed quite much since its inception)

(**The amount of patches in this PR will be reduced with most changes applied on vanilla's side...** Hopefully?)

## OnSpawn hooks
Added the `OnSpawn` hooks for items, NPCs and projectiles (called when X.NewX gets called). Aside from being usable for basic initialization of these entity types, the hooks provide access to `IEntitySource` arguments that have been inaccessible for quite too much time.

## Modernization
ModGore and ModDust have been tweaked to streamline the API.

Redundant `CloneNewInstances` properties have been removed on `Global(Item|NPC|Projectile)` classes. A property like these has already been removed from `GlobalItem` prior.

Implementations of cloning hooks have been updated in `(Mod|Global)(Item|NPC|Projectile)` classes.

## Gore Sources

`IEntitySource`s are now used for instantiation of gores too. Previously there had been an attempt to make them be used for dusts, but collaborators found this to be way too tiresome to deal with.

A new system was made for making vanilla's `NewGore` calls use fallback sources, since those couldn't have been specified by ReLogic. It's pretty basic and abuses the game's single-threadedness by using a simple `Stack<IEntitySource>`. Currently only used in HitEffect to give sources to vanilla's spawns of gores that happen during the method's execution.

## Entity Source Contexts

`ItemSourceID` and `ProjectileSourceID` static classes were removed in favor of a new `string? Context { get; }` property, which is available on every `IEntitySource` implemetation. This was done to heavily reduce and simplify hierarchy branching, as well as to have less ties between sources and the types of the entities that are being instantiated. Initially, merging the IDs was attempted, and that was a successful improvement and simplification of the vanilla designs. Looking at the merged class 10,000 times, however, made it clear, that all of the entries (except for `PlayerDeath`, which has been replaced with `EntitySource_Death`) appear more like additional details than standalone source/context descriptions. Vanilla had 4 cases of hierarchy subbranching that were for the sake of adding the IDs into the picture, 2 with the aforementioned merge. But it's clear that it's likely to want to be able to specify something like the related set bonus or a tie to golf in more than technical 2 entity source types (one group was non descriptive, and another was tied to entities hitting each other.) This is why it seemed necessary to add the context property to the interface. Now, anyone can specify additional hints with any already present entity source - "the real source is player death, but it was that set bonus that triggered this"; "the source is the enemy AI, but specifically the first phase of the boss battle", etc.

Plus, like, the IDs were hardcode, y' know. Not at all extendable by modders. And they had no way to communicate in sources without using the type system.

Present and removed Contexts / IDs:
```diff
-TorchGod
-PlayerDeath
SetBonus_Nebula
LuckyCoin
PlayerDropItemCheck
GrandDesignOrMultiColorWrench
SetBonus_SolarExplosion_WhenTakingDamage
SetBonus_SolarExplosion_WhenDashing
SetBonus_ForbiddenStorm
SetBonus_Titanium
SetBonus_Orichalcum
SetBonus_Chlorophyte
SetBonus_Stardust
WeaponEnchantment_Confetti
PlayerDeath_TombStone
FallingStar
PlayerHurt_DropFootball
StormTigerTierSwap
AbigailTierSwap
SetBonus_GhostHeal
SetBonus_GhostHurt
VampireKnives
```

## Entity Source Helpers

The vanilla helper methods for Players, NPCs, Projectiles, ~~Items~~ have been redone to remove entity-type ties as much as possible. The original methods have been internalized and turned into redirects to new ones, most of which are defined in [Entity.TML.Sources.cs](https://github.com/tModLoader/tModLoader/blob/1.4_onspawn/patches/tModLoader/Terraria/Entity.TML.Sources.cs?ts=4) (have a look!), with only `GetSource_Buff` methods defined in `Player.TML.Sources.cs` and `NPC.TML.Sources.cs`.

Summary:
```diff
Entity:
+GetSource_FromThis
+GetSource_FromAI
+GetSource_DropAsItem
+GetSource_Loot
+GetSource_GiftOrReward
+GetSource_Accessory
+GetSource_OpenItem
+GetSource_ItemUse
+GetSource_ItemUse_WithPotentialAmmo
+GetSource_OnHit
+GetSource_OnHurt
+GetSource_Death
+GetSource_Misc
+GetSource_TileInteraction
+GetSource_ReleaseEntity
+GetSource_CatchEntity
+GetSource_None (static)
+InheritSource (static)
+GetSource_NaturalSpawn (static)
+GetSource_TownSpawn (static)

Player:
+GetSource_Buff
-GetProjectileSource_Buff
-GetProjectileSource_Item
-GetItemSource_OpenItem
-GetItemSource_Death
-GetItemSource_Misc
-GetProjectileSource_Item_WithPotentialAmmo
-GetProjectileSource_SetBonus
-GetProjectileSource_OnHit
-GetProjectileSource_OnHurt
-GetProjectileSource_Accessory
-GetProjectileSource_Misc
-GetProjectileSource_TileInteraction
-GetItemSource_TileInteraction
-GetNPCSource_TileInteraction
-GetItemSource_OnHit

NPC:
+GetSource_Buff
-GetSpawnSource_NPCRelease (static)
-GetSpawnSource_NPCCatch (static)
-GetSpawnSource_NPCHurt
-GetSpawnSource_ForProjectile
-GetSpawnSourceForProjectileNPC
-GetSpawnSourceForNaturalSpawn (static)
-GetSpawnSourceForTownSpawn (static)
-GetSpawnSourceForNPCFromNPCAI
-GetItemSource_Loot

Projectile:
-GetNoneSource (static)
-GetNPCSource_FromThis
-GetProjectileSource_FromThis
-GetItemSource_DropAsItem
-GetItemSource_FromThis
-GetProjectileSource_OnHit
-GetItemSource_OnHit
-InheritSource (static)
```

## Changes to entity source classes

Some new entity sources and helper methods were added. Some were removed.
Have a look:
```diff
+ Context arguments on everything

 AEntitySource_Tile(int tileCoordsX, int tileCoordsY) (abstract)
 EntitySource_BossSpawn(Entity entity)
 EntitySource_Buff(Entity entity, int buffId, int buffIndex)
-EntitySource_ByItemSourceId(Entity entity, int itemSourceId)
-EntitySource_ByProjectileSourceId(int projectileSourceId)
+EntitySource_Death(Entity entity)
 EntitySource_DebugCommand()
 EntitySource_DropAsItem(Entity entity)
 EntitySource_Film()
 EntitySource_FishedOut(Entity entity)
 EntitySource_Gift(Entity entity)
+EntitySource_HitEffect(Entity entity)
 EntitySource_ItemOpen(Entity entity, int itemType)
 EntitySource_ItemUse(Entity entity, Item item)
 EntitySource_ItemUse_WithAmmo(Entity entity, Item item, int ammoItemIdUsed)
 EntitySource_Loot(Entity entity)
+EntitySource_Misc() (Always requires context)
 EntitySource_Mount(Entity entity, int mountId)
 EntitySource_OldOnesArmy()
-AEntitySource_OnHit(Entity entityStriking, Entity entityStruck) (abstract)
+EntitySource_OnHit(Entity entityStriking, Entity entityStruck)
-EntitySource_OnHit_ByItemSourceID(Entity entityStriking, Entity entityStruck, int itemSourceId)
-EntitySource_OnHit_ByProjectileSourceID(Entity entityStriking, Entity entityStruck, int projectileSourceId)
 EntitySource_Parent(Entity entity)
 EntitySource_RevengeSystem()
 EntitySource_ShakeTree(int tileCoordsX, int tileCoordsY)
 EntitySource_SpawnNPC()
 EntitySource_Sync()
 EntitySource_TileBreak(int tileCoordsX, int tileCoordsY)
 EntitySource_TileEntity(TileEntity tileEntity)
 EntitySource_TileInteraction(Entity entity, int tileCoordsX, int tileCoordsY)
+EntitySource_TileUpdate(int tileCoordsX, int tileCoordsY) (Might remove??)
 EntitySource_Wiring(int tileCoordsX, int tileCoordsY)
 EntitySource_WorldEvent()
 EntitySource_WorldGen()
```

## Etc

Some random parts of ExampleMod might've been improved in addition to the changes related to the above.

Not yet reviewed, what else was changed?


Chicken-Bones:
Perfectly happy with this idea, and the hook names/locations. I think the `userdata` parameter should be removed.

fakelag:
This seems good, how close is this to being ready?

Also, what is the purpose of `EntitySpawnData`? If there is relevant info to be passed from the hook locations then couldn't those be added as arguments to the hook directly?

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -162,14 +162,14 @@
...
  					projectile.projUUID = projectile.identity;
  
-@@ -8042,6 +_,8 @@
- 				}
- 			}
+@@ -8158,6 +_,8 @@
+ 			if (Type == 249)
+ 				projectile.frame = Main.rand.Next(5);
  
 +			ProjectileLoader.OnSpawn(projectile, spawnSource);
```
DarkLight66, line 169: Why move this to the end of the method? Having it before the `NetMessage.SendData` call would be better as modders wouldn't have to wait an update tick for their changes to sync. If the issue is that the code section after the `NetMessage.SendData` call would override whatever changes were made in `OnSpawn`, that section could be made to run before it using gotos.

Itorius: Indeed, I think the easiest solution is to move the `NetMessage.SendData` block to the end of the method, which I did.

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -706,6 +706,24 @@
...
+-		public static int NewNPC(int X, int Y, int Type, int Start = 0, float ai0 = 0f, float ai1 = 0f, float ai2 = 0f, float ai3 = 0f, int Target = 255) {
++		public static int NewNPC(int X, int Y, int Type, int Start = 0, float ai0 = 0f, float ai1 = 0f, float ai2 = 0f, float ai3 = 0f, int Target = 255, INPCSource source = null) {
+ 			if (Main.getGoodWorld) {
+ 				if (Type == 46)
+ 					Type = 614;
+@@ -62517,6 +_,8 @@
+ 						ChatHelper.BroadcastChatMessage(NetworkText.FromKey("Announcement.HasAwoken", Main.npc[num].GetTypeNetName()), new Color(175, 75, 255));
+ 				}
+ 
++				NPCLoader.OnSpawn(Main.npc[num], source);
```
DarkLight66, line 722: I just found one annoying vanilla code detail: The code that spawns NPCs does something special for NPCs with negative `netID`, which are just variants of other NPCs, where it does something like:
```cs
int index = NPC.NewNPC(spawnTileX * 16 + 8, spawnTileY * 16, NPCID.BlueSlime); // ID 1
Main.npc[index].SetDefaults(NPCID.GreenSlime); // ID -3
```
Now, because `NPC.SetDefaults` is being called again after `NPC.NewNPC`, it will probably wipe whatever the `NPCLoader.OnSpawn` call may have done. 
On a similar note, a problem with tModLoader currently is that the call to `NPCLoader.SetDefaults` is at a place were it doesn't receive the correct `netID` for negative NPCs, due to how default values of negative `netID` NPCs are set. Idk if fixing that may be within the scope of this PR though.

ExampleMod/Content/Items/Weapons/ExampleGun.cs
```diff
@@ -50,6 +52,18 @@ public class ExampleGun : ModItem
...
+		public override bool Shoot(Player player, ProjectileSource_Item_WithAmmo source, Vector2 position, Vector2 velocity, int type, int damage, float knockback) {
+			if (Main.rand.NextBool(3))
+			{
+				Projectile.NewProjectile(source, position, velocity, ModContent.ProjectileType<ExampleInstancedProjectile>(), damage, knockback, player.whoAmI);
+
+				return false;
+			}
+			
+			return true;
+		}
```
DarkLight66, line 65: Since you are only changing the type of the projectile being shot with a chance, this is also achievable by changing `type` in `ModifyShootStats`, like so:
```suggestion
		public override void ModifyShootStats(Player player, ref Vector2 position, ref Vector2 velocity, ref int type, ref int damage, ref float knockback) {
			if (Main.rand.NextBool(3)) {
				type = ModContent.ProjectileType<ExampleInstancedProjectile>();
			}
		}
```

Mirsario (approved):
Updated the PR and its description, and all seems good enough. I've put the branch into `1.4_mergedtesting` and will merge it soon if I don't find issues.

Mirsario (changes requested):
In the previous 'Approve' I completely forgot about actually making entity source types and then making vanilla use them where possible.

direwolf420:
Please rework the current code around `initialized` in [ExampleBobber](https://github.com/tModLoader/tModLoader/blob/1.4_onspawn/ExampleMod/Content/Projectiles/ExampleBobber.cs#L42) to use `OnSpawn`. (Also frankly the `Main.dedServ` check for `AddLight` is unnecessary)

EDIT: done

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -29,39 +30,22 @@ public abstract class GlobalNPC : GlobalType<NPC>
...
+		/// Allows you to set the properties of any and every NPC that gets created.
 		/// </summary>
-		public virtual GlobalNPC NewInstance(NPC npc) {
-			if (CloneNewInstances) {
-				return Clone();
-			}
-			GlobalNPC copy = (GlobalNPC)Activator.CreateInstance(GetType());
-			copy.Mod = Mod;
-			copy.index = index;
-			return copy;
```
Chicken-Bones, line 57: I'm unhappy with the removal of `NewInstance` primarily because `CloneNewInstances` previously defaulted to false. Requiring people to override `Clone` whenever they add something which can't be memberwised cloned is undesirable. I'm not, in general a fan of 'cloning' being the default method of instantiating an instance of a global. 

This change will silently break any mods which relied on reconstruction being the default behaviour

ExampleMod/Common/GlobalProjectiles/ExampleProjectileModifications.cs
```diff
@@ -3,14 +3,14 @@
 using Terraria.ModLoader;
 using Microsoft.Xna.Framework;
 using ExampleMod.Common.GlobalNPCs;
+using Terraria.DataStructures;
```
Chicken-Bones: A lot of these need cleanup

ExampleMod/Content/Items/Tools/ExampleMagicMirror.cs
```diff
@@ -61,15 +69,14 @@ internal class ExampleMagicMirror : ExampleItem
...
-
 		public override void ModifyTooltips(List<TooltipLine> tooltips) {
 			// This code shows using Color.Lerp,  Main.GameUpdateCount, and the modulo operator (%) to do a neat effect cycling between 4 custom colors.
 			foreach (TooltipLine line2 in tooltips) {
 				if (line2.Mod == "Terraria" && line2.Name == "ItemName") {
 					float fade = (Main.GameUpdateCount % 60) / 60f;
 					int index = (int)((Main.GameUpdateCount / 60) % 4);
-					line2.OverrideColor = Color.Lerp(itemNameCycleColors[index], itemNameCycleColors[(index + 1) % 4], fade);
+
+					line2.OverrideColor = Color.Lerp(itemNameCycleColors[index], itemNameCycleColors[(index + 1) % itemNameCycleColors.Length], fade);
```
Chicken-Bones: changes in this class look to be an unrelated refactor? Not particularly a problem but... PR is less clean. I see no real reason to change this, or insert an extra blank line. I see you changed the `% 4`, which is sensible, but didn't change it above, may as well just put the `+ 1` in the `index` above, or remove it entirely...

ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -217,9 +217,10 @@ public class ExamplePerson : ModNPC
 					Main.npcChatText = $"I upgraded your {Lang.GetItemNameValue(ItemID.HiveBackpack)} to a {Lang.GetItemNameValue(ModContent.ItemType<WaspNest>())}";
 
 					int hiveBackpackItemIndex = Main.LocalPlayer.FindItem(ItemID.HiveBackpack);
+					var entitySource = NPC.GetSource_GiftOrReward();
 
 					Main.LocalPlayer.inventory[hiveBackpackItemIndex].TurnToAir();
-					Main.LocalPlayer.QuickSpawnItem(new EntitySource_Gift(NPC), ModContent.ItemType<WaspNest>());
+					Main.LocalPlayer.QuickSpawnItem(entitySource, ModContent.ItemType<WaspNest>());
```
Chicken-Bones, line 222: preferred 1 line, not have extra code between source and spawn call.

ExampleMod/Content/NPCs/MinionBoss/MinionBossBody.cs
```diff
@@ -270,9 +270,11 @@ public class MinionBossBody : ModNPC
 				int backGoreType = Mod.Find<ModGore>("MinionBossBody_Back").Type;
 				int frontGoreType = Mod.Find<ModGore>("MinionBossBody_Front").Type;
 
+				var entitySource = NPC.GetSource_Death();
```
Chicken-Bones, line 273: I definitely intend to implement a `DamageSource` style system on the back of this. It would be good to have actual 'strikes' and those could be context to the death source.

patches/tModLoader/Terraria/Entity.TML.Sources.cs
```diff
@@ -0,0 +1,82 @@
...
+namespace Terraria
+{
+	partial class Entity
+	{
+		// Common
+
+		public IEntitySource GetSource_FromThis(string? context = null)
+			=> new EntitySource_Parent(this, context);
+
+		public IEntitySource GetSource_FromAI(string? context = null)
```
Chicken-Bones, line 15: If there's no differentiation between `FromAI` and `FromThis`, then I'd remove this method. It doesn't seem like there's use cases for this.

ExampleMod/Content/NPCs/MinionBoss/MinionBossMinion.cs
```diff
@@ -108,9 +108,11 @@ public class MinionBossMinion : ModNPC
 				// If this NPC dies, spawn some visuals
 
 				int dustType = 59; // Some blue dust, read the dust guide on the wiki for how to find the perfect dust
+
 				for (int i = 0; i < 20; i++) {
 					Vector2 velocity = NPC.velocity + new Vector2(Main.rand.NextFloat(-2f, 2f), Main.rand.NextFloat(-2f, 2f));
 					Dust dust = Dust.NewDustPerfect(NPC.Center, dustType, velocity, 26, Color.White, Main.rand.NextFloat(1.5f, 2.4f));
+
```
Chicken-Bones, line 115: unnecessary formatting change? It can stay... just something to skim over when reviewing

patches/tModLoader/Terraria/ID/ProjectileSourceID.cs.patch
```diff
@@ -0,0 +1,40 @@
...
++	internal static class ProjectileSourceID
+ 	{
+ 		public const int None = 0;
+ 		public const int SetBonus_SolarExplosion_WhenTakingDamage = 1;
+@@ -21,5 +_,26 @@
+ 		public const int SetBonus_GhostHurt = 16;
+ 		public const int VampireKnives = 18;
+ 		public const int Count = 19;
++
++		public static string? ToContextString(int itemSourceId) => itemSourceId switch {
```
Chicken-Bones, line 21: I'm mostly fine with this, hopefully declarative set bonuses take care of some of this.

Chicken-Bones (commented):
Overall fairly happy with this. I'm very much not sure how I feel about context for sources other than `misc` can you make a list of these for me?

ExampleMod/Content/NPCs/ExamplePerson.cs
Mirsario: Separating these things to local variables gives people a magical hint of the possibility of 'caching' source instances in the context of single methods. Real mods can be written however they like it.

ExampleMod/Content/NPCs/MinionBoss/MinionBossMinion.cs
Mirsario: I can't not accidentally spam readability improvements when.. trying to read. Honestly just plain too lazy to move to direct commits...

Mirsario (approved):

patches/tModLoader/Terraria/Entity.TML.Sources.cs
Mirsario: There's many methods with repeating returns both here and in vanilla still. Yorai's idea with this was to first - ease search for fitting sources, and second - ease the potential future rewrite processes by making contexts a little bit more explicit, even if they're not going to be passed anywhere for now.

ilija04:
Oh no! This broke the Cheat Sheet mod. I guess there will be created a new Cheat Sheet mod exclusive to the preview branch.

PR #1138, opened 2020-10-16
Fixed a long-standing use sound desync issue

ThomasSpeedrunner:
Non-(broadsword/shortsword) items which fire projectiles should no longer have their usage sound(s) desynced from their actual usage animation(s) as you continue to fire them.

Allow me to elaborate.
Whenever an item is used in Terraria, generally it makes a sound. It also possesses an animation for usin' it, and the important thing here is the animation variable, `useAnimation`. To explain why the fix is programmed the way it is, it's because of the way broadswords (and shortswords, presuming my research from pre-JE still holds up) work.
When a broadsword or shortsword is...swung, let's say, to keep it simple...it uses the `useAnimation` variable to determine how long it should swing for. Now, this would be all fine and dandy, except for one thing: whenever a broadsword is swung, it skips the last frame of its animation. This means an autoswing sword actually gets used one tick faster than its `useAnimation` value would imply.
(You may think `useTime` is used for this, but it's not --- it is instead used to dictate the cooldown on certain weapons' projectiles, such as the Enchanted Sword.)
For any other type of weapon, this isn't the case. Their usage animation is exactly the number of frames long that the `useAnimation` value dictates. HOWEVER, because the sound system still relies on the idea that everything skips the last frame, the sound is played again one tick before the animation restarts. This causes things such as staves and guns to become incredibly desynced between their usage animations and usage sounds, while broadswords and shortswords remain fine.
The question of whether or not the sound desyncs is also dependent upon other factors, most of which I simply can't explain --- but I wanted to make it a point regardless to cover all the bases.
This is why the fix is programmed the way it is; a specific exception has to be put in place for broadsword- and shortsword-style weapons without potentially changing the game on a major balance level; because of how broadswords in particular work, it's possible that this could majorly affect the damage output of multiple weapons if just left as a one-and-done fix.

I know hardcode isn't my thing, but it's important here to retain the stability of the game.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3113,7 +3113,7 @@
  			if (Main.dedServ)
  				return;
  
-@@ -35342,8 +_,14 @@
+@@ -35342,8 +_,13 @@
  			attackCD = 0;
  			ApplyItemAnimation(sItem);
  			bool flag2 = ItemID.Sets.SkipsInitialUseSound[sItem.type];
@@ -3123,7 +3123,6 @@
 +				// If the projectile shoots no projectiles or is a broadsword or pre-JE shortsword, play the sound here.
 +				if (sItem.shoot == ProjectileID.None)
 +					SoundEngine.PlaySound(sItem.UseSound, base.Center);
-+
 +				else if ((sItem.useStyle == 1 || sItem.useStyle == 3) && !sItem.noMelee && !sItem.noUseGraphic)
 +					SoundEngine.PlaySound(sItem.UseSound, base.Center);
 +			}
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3069,11 +3069,13 @@
  					int num178 = Projectile.NewProjectile(pointPoisition.X, pointPoisition.Y, num2, num3, projToShoot, Damage, KnockBack, i);
  					if (sItem.type == 726)
  						Main.projectile[num178].magic = true;
-@@ -33910,6 +_,9 @@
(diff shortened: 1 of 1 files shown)


Ozzatron (approved):

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3101,6 +3111,22 @@
...
+ 			attackCD = 0;
+ 			ApplyItemAnimation(sItem);
+ 			bool flag2 = ItemID.Sets.SkipsInitialUseSound[sItem.type];
+-			if (sItem.UseSound != null && !flag2)
++			if (sItem.UseSound != null && !flag2) {
++				if (sItem.shoot == ProjectileID.None)
+-				SoundEngine.PlaySound(sItem.UseSound, base.Center);
++					SoundEngine.PlaySound(sItem.UseSound, base.Center);
++
++				else if ((sItem.useStyle == 1 || sItem.useStyle == 3) && !sItem.noMelee && !sItem.noUseGraphic)
```
Mirsario, line 3124: Weird line break

Mirsario (commented):
Please add comments to the first 2 patches, I'm confused from this even though I somewhat know what they do.

Mirsario:
We need to fix the underlying issue instead, or have yorai fix it. From 10 seconds of debugging, that issue apparently has nothing at all to do with swords and audio (as you may already know). It happens with all items in the game that have autoReuse set to true. With every use, player.itemTime and itemAnimation desync by 1 tick.
image

Reproduction on a shotgun (with double itemTime & itemAnimation, and autoReuse forced on):
https://cdn.discordapp.com/attachments/445276626352209920/777965788601516062/2020-11-16_21-37-20.mp4
(Whenever the cursor is on screen, I'm using a speedhack)

PR #1139, opened 2020-10-16
Research + Fishing  Modding APIs

JPANv2:
### What is the new feature?
This Pull request creates two new hooks to the Research System:

One "CanResearch",  in order to allow modders to allow or not an item to be researched, regardless of it was already fully researched in the past or even if it has a research value;

One "OnResearched", called before the item is modified, that allows for custom behaviour to be run if an item was either partially or fully researched.

These hooks allow for many features to be implemented by modders that are otherwise pratically impossible to achieve, including:

- Allow Modders to prevent an item from being researched before a certain stage in game is reached (ex- Hardmode Only)
- Allow Modders to force an item to be researched even if its researched value is already full, like researching a weapon a second time to unlock its prefix, or create Random Research items that add research for a random item when researched.
- Allows for code to run when the item is researched, to add features such as my mod's "Research unlocks everything crafted from that item".

It would be possible to make the design more complex, by adding new parameters to the functions, but they would add nothing new.

`class ChestOfRandomResearch : ModItem
{

public override bool? CanResearch(){
 return true;
}

public override void OnResearched(bool fullyResearched){
int stack = item.stack;
for(int i = 0; i < stack; i++){
Item newItm = new Item();
newItm.SetDefaults(Main.rand.Next(1, ItemLoader.Count));
newItm.stack = 1;
int amount = 0;
Main.CreativeMenu.SetItembyIndex(newItm,0);
Main.CreativeMenu.SacrificeItem(amount);
}
}
}
`
### ExampleMod updates
The above example could be added to the ExampleMod, after tested for bugs.


patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -425,6 +425,38 @@ private class HookList
...
+
+			if (canResearch != null)
+				return canResearch;
+
+			foreach (var g in HookCanResearch.arr) {
+				canResearch = g.Instance(item).CanResearch(item);
+				
+				if (canResearch != null)
+					return canResearch;
+			}
```
Chicken-Bones, line 444: all hooks like this need to decide on a precedence behaviour. Returning the result of the first hook to give a non-null (vanilla) response will be weird when one mod says yes and another says no.

You need to decide, and describe in the documentation of the hook, which should take priority.

Do this by suggesting use cases for these hooks.

Solxanich:
Hello @JPANv2 ; thank you for drafting this PR.
As the tModLoader project continues to progress, I'd like to ask if you intend to return to complete this PR and address Chicken-Bones's review comments?

If not, could you please let us know, so we can have someone else pick up the branch and bring it to completion?

Thank you for your time.

JPANv2:
Created the requested changes and updated my branch to the latest TMod 1.4. However, it now says that my branch can't build, even though checking it out and clean/build in VS works just fine.

JPANv2:
Updated version of both Fishing changes and Research Hooks. Unfortunately, could not separate them into two forks, but both are building, using the latest patch and tested working. If needed, I'll also add an ExampleMod example for both APIs.

ExampleMod/Content/Items/ExampleResearchPresent.cs
```diff
@@ -0,0 +1,61 @@
...
+        public override bool? CanResearch() {
+			return true;
+        }
+
+        public override void OnResearched(bool fullyResearched) {
+			/*If you managed to be lucky enough to get all presents, it will give you all
+			 accessories*/
+			if (fullyResearched) {
+				for (int i = 1; i < ItemLoader.ItemCount; i++) {
+					Item tr = new Item(i);
```
Chicken-Bones: Use `ContentSamples.ItemsByType` to query item properties, rather than making a new item. And also below on line 47

ExampleMod/Content/Items/ExampleResearchPresent.cs
```diff
@@ -0,0 +1,61 @@
...
+                    if (tr.accessory) {
+						CreativeUI.ResearchItem(tr);
+					}
+                }
+				Main.NewText("You got all accessories!");
+				return;
+			} else {
+				int j = this.Item.stack;
```
Chicken-Bones: It's not clear to me exactly what the intended behaviour is here. Needs a comment.
From what I can see, if you research multiple at once, they all get consumed, but you might run out of rolls on the item table (fixed 1000) before that.
If you run out of rolls, you may have actually researched some accessories, but the text will still say "no new accessory was inside"
The `ResearchItem` call doesn't check if it's already been fully researched either...

ExampleMod/Content/Items/ExampleResearchPresent.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace ExampleMod.Content.Items
+{
+	public class ExampleResearchPresent : ModItem
+	{
+
+		public override void SetStaticDefaults() {
+			DisplayName.SetDefault("Example Present");
+			Tooltip.SetDefault("Contains a random accessory! Try researching it for infinite of it!");
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = ItemLoader.ItemCount;
```
Chicken-Bones: Does sacrifice count changing when you add/remove mods cause a problem? What if you sacrifice 5k of a thing and then unload a mod bringing the count needed down to 4k. Does the item become fully researched then and then become unresearched when you add more mods?

Chicken-Bones (changes requested):
Partial Example review

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -191,6 +191,34 @@ public abstract class GlobalItem : GlobalType<Item>
 		public virtual void ModifyResearchSorting(Item item, ref ContentSamples.CreativeHelper.ItemGroup itemGroup) {
 		}
 
+		/// <summary>
+		/// Allows you to choose if a given bait will be consumed by a given player
```
Chicken-Bones, line 196: Need to specify precedence here.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4180,6 +4180,15 @@
  			int whoAmI = base.whoAmI;
  			bool flag = true;
  			int num = (int)((float)Main.mouseX + Main.screenPosition.X) / 16;
+@@ -36421,6 +_,8 @@
+ 			if (baitTypeUsed == 2673)
+ 				flag = true;
+ 
++			bool? forced = PlayerLoader.CanConsumeBait(this, item) ?? ItemLoader.CanConsumeBait(this, item);
```
Chicken-Bones: You might need to use `CombinedHooks` here, because you need to combine the precedence of the two results.

Chicken-Bones (changes requested):

ExampleMod/Content/Items/ExampleResearchPresent.cs
JPANv2: I changed the Research value from a random integer (1-100) to ItemCount and back, and if the player already researched fully the present, would still be able to research it further because of CanResearch returning always true. Also, when changing the values, the item went from fully researched to only partially researched whenever it went from lower to higher numbers, but researching one new item will reduce the amount researched to the maximum at the time.

patches/tModLoader/Terraria/Player.cs.patch
JPANv2: This one I had decided that player-based decisions take precedence over item decisions, for scope reasons. But I can change it to always have true as precedence for every single case.

ExampleMod/Content/Items/ExampleResearchPresent.cs
JPANv2: didn't know about that function. Will apply it where it's relevant.

ExampleMod/Content/Items/ExampleResearchPresent.cs
JPANv2: I changed ResearchItem to have a out boolean alreadyResearched, that will indicate if something was already researched before being sent out. only place that will not have it is the original Sacrifice function.

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
JPANv2: done False trumps true, trumps null. Placed in the comments of both here and player hooks.

Chicken-Bones (approved):

PR #1143, opened 2020-10-16
Fishing hooks Pt 1

JPANv2:
### What is the new feature?
This pull request adds a couple new Fishing Hooks, and fixes the main one (CatchFish) to use the new Terraria FishingAttempt.
- CatchFish now takes a FishingAttempt, an optional text and color parameters to allow overriding the Sonar text.
- OnFishingNPCSummon Hook that allows to run extra code after an npc is summoned through fishing (including Duke Fishron)
- A ConsumeBait hook that allows modders to control how their bait is spent.

Fishing hooks are something tMod is sorely lacking.
The old system did not allow for players to access all information about their fishing conditions, and could not spawn npcs just like the blood moon fishing allows.
The hooks allow for new implementations, like more Duke-like bossses, or simply surprising the player with different items.

The old method for using CatchFish was still usable, but lacked QoL features the new design presents. As for the new hooks, ConsumeBait could be divided in a canConsume/ modifyUse pair. Also, consumeBait could be accessed from both bait and Fishing Rod, but the current implementation only uses Bait.

### ExampleMod updates
Adding a new ExampleFishingPlayer could showcase the new features of the FishingAttempt implementation. Could also add a ExampleBait with reduced consumption chance.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
@@ -246,6 +246,7 @@ namespace Terraria.ModLoader
 
 		/// <summary>
 		/// Allows you to change an item bait's rate of consumption, being able to force it to be consumed or stop it from being consumed or return null for unchanged vanilla behaviour.
+		/// Forcing to be consumed takes precedence over not being consumed, over vanilla usage.
 		/// </summary>
 		/// <param name="bait">The bait item being targeted</param>
 		/// <param name="player">The player using this bait</param>
--- a/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
@@ -245,6 +245,15 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
+		/// Allows you to change an item bait's rate of consumption, being able to force it to be consumed or stop it from being consumed or return null for unchanged vanilla behaviour.
+		/// </summary>
+		/// <param name="bait">The bait item being targeted</param>
+		/// <param name="player">The player using this bait</param>
+		public virtual bool? ConsumeBait(Item bait, Player player) {
+			return null;
+		}
+
+		/// <summary>
 		/// Allows you to modify the projectile created by a weapon based on the ammo it is using.
 		/// </summary>
 		/// <param name="weapon">The item that is using this ammo</param>
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -517,15 +517,21 @@ namespace Terraria.ModLoader
 			if (bait.IsAir)
 				return null;
 
-			bool? ans = bait.modItem?.ConsumeBait(player);
-			if (ans != null)
-				return ans;
-			foreach (var g in HookConsumeBait.arr) {
-				ans = g.ConsumeBait(bait, player);
-				if (ans != null)
-					return ans;
+			bool? result = null;
+
+			foreach (GlobalItem g in HookConsumeBait.arr) {
+				bool? consumeBait = g.Instance(bait).ConsumeBait(bait, player);
+
+				if (consumeBait.HasValue) {
+					if (!consumeBait.Value) {
+						return false;
+					}
+
+					result = true;
+				}
 			}
-			return null;
+
+			return result ?? bait.modItem?.ConsumeBait(player);
 		}
 
 		private static HookList HookOnConsumeAmmo = AddHook<Action<Item, Player>>(g => g.OnConsumeAmmo);
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -513,6 +513,10 @@ namespace Terraria.ModLoader
 
 		private delegate bool? DelegateConsumeBait(Item bait, Player player);
 		private static HookList HookConsumeBait = AddHook<DelegateConsumeBait>(p => p.ConsumeBait);
+
+		/// <summary>
+		/// Hook that decides if bait is consumed or not. True stops the code and immediately forces consumption. False is only valid if no True is found.
+		/// </summary>
 		public static bool? ConsumeBait(Player player, Item bait) {
 			if (bait.IsAir)
 				return null;
@@ -523,11 +527,11 @@ namespace Terraria.ModLoader
 				bool? consumeBait = g.Instance(bait).ConsumeBait(bait, player);
 
 				if (consumeBait.HasValue) {
-					if (!consumeBait.Value) {
-						return false;
+					if (consumeBait.Value) {
+						return true;
 					}
 
-					result = true;
+					result = false;
 				}
 			}
 
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -511,6 +511,23 @@ namespace Terraria.ModLoader
 			return true;
 		}
 
+		private delegate bool? DelegateConsumeBait(Player player);
+		private static HookList HookConsumeBait = AddHook<DelegateConsumeBait>(p => p.ConsumeBait);
+		public static bool? ConsumeBait(Player player, Item bait) {
+			if (bait.IsAir)
+				return null;
+
+			bool? ans = bait.modItem?.ConsumeBait(player);
+			if (ans != null)
+				return ans;
+			foreach (var g in HookConsumeBait.arr) {
+				ans = g.ConsumeBait(bait, player);
+				if (ans != null)
+					return ans;
+			}
+			return null;
+		}
+
 		private static HookList HookOnConsumeAmmo = AddHook<Action<Item, Player>>(g => g.OnConsumeAmmo);
 		/// <summary>
 		/// Calls ModItem.OnConsumeAmmo for the weapon, ModItem.OnConsumeAmmo for the ammo, then each GlobalItem.OnConsumeAmmo hook for the weapon and ammo.
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -511,7 +511,7 @@ namespace Terraria.ModLoader
 			return true;
 		}
 
-		private delegate bool? DelegateConsumeBait(Player player);
+		private delegate bool? DelegateConsumeBait(Item bait, Player player);
 		private static HookList HookConsumeBait = AddHook<DelegateConsumeBait>(p => p.ConsumeBait);
 		public static bool? ConsumeBait(Player player, Item bait) {
 			if (bait.IsAir)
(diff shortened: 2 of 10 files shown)


patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -511,6 +511,23 @@ private class HookList
...
+
+			bool? ans = bait.modItem?.ConsumeBait(player);
+			if (ans != null)
+				return ans;
+			foreach (var g in HookConsumeBait.arr) {
+				ans = g.ConsumeBait(bait, player);
+				if (ans != null)
+					return ans;
+			}
+			return null;
```
Rartrin, line 528: Global hooks should take priority over mod ones. In other words, loop through the global hooks first, then run the mod hook.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3111,6 +3111,43 @@
...
+@@ -35787,8 +_,11 @@
+ 			baitTypeUsed = item.type;
+ 			if (baitTypeUsed == 2673)
+ 				flag = true;
++			bool? baitOverride = PlayerHooks.ConsumeBait(this, item);
++			if (baitOverride == null)
++				baitOverride = ItemLoader.ConsumeBait(this, item);
+ 
+-			if (flag) {
++			if (flag && baitOverride == null || baitOverride.Value) {
```
Rartrin, line 3147: Use null coalescence
```
bool? baitOverride = PlayerHooks.ConsumeBait(this, item) ?? ItemLoader.ConsumeBait(this, item);
```
Also, add parentheses to the `if`. `&&` evaluates before `||` and it isn't obvious if that was intentional. If you did intend for `baitOverride == null || baitOverride.Value` to be evaluated first, replace it with `(baitOverride??true)`.

patches/tModLoader/Terraria/PopupText.cs.patch
```diff
@@ -17,6 +17,48 @@
  		public bool sonar;
  		public PopupTextContext context;
  		public int npcNetID;
+@@ -40,6 +_,41 @@
+ 			}
+ 		}
+
```
direwolf420, line 23: Is this new NewText overload necessary? If it really is, make PopupText partial, add a new file called PopupText.tML.cs and add the method there

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -414,16 +414,66 @@
...
- 				if (Main.player[owner].sonarPotion) {
- 					Item item = new Item();
++				if (Main.player[owner].sonarPotion && (sonarText == null || sonarText.Trim().Length > 0)) {
++
+-				if (Main.player[owner].sonarPotion) {
++					if (sonarText == null) {
+-					Item item = new Item();
++						Item item = new Item();
+-					item.SetDefaults(fisher.rolledItemDrop);
++						item.SetDefaults(fisher.rolledItemDrop);
```
direwolf420, line 441: These indentation changes in patches are not welcome, figure out some code change that won't require these

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -414,16 +414,66 @@
...
++
+ 					if (PopupText.sonarText > -1)
+ 						Main.popupText[PopupText.sonarText].sonar = true;
+ 				}
+@@ -13093,13 +_,20 @@
+ 				float num3 = fisher.fishingLevel;
+ 				ai[1] = (float)Main.rand.Next(-240, -90) - num3;
+ 				localAI[1] = fisher.rolledItemDrop;
++				localAI[0] = fisher.rolledItemStack;
+ 				netUpdate = true;
```
direwolf420, line 459: Is this change/extension to the AI necessary? It's not mentioned in the PR but is definitely worth mentioning if this goes through (on the wiki later)

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -501,10 +551,90 @@
...
++			if (localAI[0] > 0) {
++				item.stack = (int)localAI[0];
++			}
++			else {
+-			if (itemType == 3196) {
++				if (itemType == 3196) {
+-				int finalFishingLevel = thePlayer.GetFishingConditions().FinalFishingLevel;
++					int finalFishingLevel = thePlayer.GetFishingConditions().FinalFishingLevel;
+-				int minValue = (finalFishingLevel / 20 + 3) / 2;
++					int minValue = (finalFishingLevel / 20 + 3) / 2;
```
direwolf420, line 577: Same thing with indentation, maybe `else if (itemType == 3196)`?

direwolf420 (commented):
Some easy to fix stuff and general comments from me

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -414,16 +414,66 @@
...
-@@ -13080,6 +_,9 @@
+@@ -13006,6 +_,7 @@
+ 
+ 			fisher.playerFishingConditions = Main.player[owner].GetFishingConditions();
+ 			int baitItemType = fisher.playerFishingConditions.BaitItemType;
++
+ 			if (baitItemType == 2673) {
+ 				Main.player[owner].displayedFishingInfo = Language.GetTextValue("GameUI.FishingWarning");
+ 				if ((fisher.X < 380 || fisher.X > Main.maxTilesX - 380) && fisher.waterTilesCount > 1000 && !NPC.AnyNPCs(370)) {
+@@ -13080,12 +_,23 @@
```
direwolf420, line 425: Forgot this, unnecessary newline

direwolf420 (commented):
Review Part 2/?

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
Mirsario: Copy this: https://github.com/tModLoader/tModLoader/blob/21166e09b6d0c526957f3e50cc8f01d6de6c5bc7/patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs#L300

patches/tModLoader/Terraria/Projectile.cs.patch
Mirsario: Gotos are good for this.

patches/tModLoader/Terraria/Player.cs.patch
JPANv2: That was indeed a bug. The parentheses were missing.

patches/tModLoader/Terraria/PopupText.cs.patch
JPANv2: PopupText.NewText with a string allows for modders to include any text they want instead of the Item or NPC name text options.

patches/tModLoader/Terraria/Projectile.cs.patch
JPANv2: I needed some space to allow for item stack manipulation to be possible with this system. Normal Vanilla behaviour is to always use stack of 1, and fix that in the player catch method.

localAI[0] was only being written to, but never read, so I used that to transmit the stack. Not my ideal solution, but there was no more space to use.

JPANv2:
Think I got all of them. Please review again and let me know if something's up.

patches/tModLoader/Terraria/PopupText.TML.cs
```diff
@@ -0,0 +1,48 @@
...
+			popupText.context = context;
+			if (stay5TimesLonger)
+				popupText.lifeTime *= 5;
+			popupText.coinValue = 0;
+			popupText.coinText = false;
+			popupText.color = color;
+		}
+
+		return num;
+	}
```
Chicken-Bones, line 46: This is a lot of duplication of vanilla code, and you've forgotten things like resetting  `npcNetID`

I think instead you should modify the `npcNetID` version of the vanilla method to take a string, as well, and create some overloads which call it

patches/tModLoader/Terraria/Projectile.cs.patch
Chicken-Bones: Acceptable, but please put a comment in the code

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -414,16 +414,54 @@
...
-+			PlayerHooks.CatchFish(Main.player[owner], Main.player[owner].inventory[Main.player[owner].selectedItem],
-+				fisher.waterNeededToFish, fisher.inLava ? 1 : fisher.inHoney ? 2 : 0, fisher.fishingLevel, fisher.heightLevel, fisher.questFish, ref fisher.rolledItemDrop);
++			string sonarText = null;
++			Color color = Color.White;
++			PlayerHooks.CatchFish(Main.player[owner], fisher, ref sonarText, ref color);
 +
  			if (fisher.rolledItemDrop > 0) {
- 				if (Main.player[owner].sonarPotion) {
+-				if (Main.player[owner].sonarPotion) {
++				if (Main.player[owner].sonarPotion && (sonarText == null || sonarText.Trim().Length > 0)) {
```
Chicken-Bones, line 427: I'd prefer just `sonarText != ""`

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -501,10 +539,42 @@
...
  							obj.velocity.Y -= 3f + (float)num3 * 1.5f;
-@@ -33196,6 +_,7 @@
- 				int num4 = item.stack = Main.rand.Next(minValue2, num3 + 1);
+@@ -32992,9 +_,6 @@
+ 						AI_061_FishingBobber_DoASplash();
+ 				}
+ 
+-				if (localAI[0] < 100f)
+-					localAI[0] += 1f;
+-
```
Chicken-Bones, line 548: needs comment

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -501,10 +539,42 @@
...
+ 					ReduceRemainingChumsInPool();
+@@ -33158,7 +_,10 @@
+ 		private void AI_061_FishingBobber_GiveItemToPlayer(Player thePlayer, int itemType) {
+ 			Item item = new Item();
+ 			item.SetDefaults(itemType);
++			if (localAI[0] > 0) {
++				item.stack = (int)localAI[0];
++			}
+-			if (itemType == 3196) {
++			else if (itemType == 3196) {
```
Chicken-Bones, line 560: no need to change this `if` to `else if`

Chicken-Bones (changes requested):
Good PR, got some questions

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -244,6 +244,15 @@ public class GlobalItem:ModType
 		public virtual void GetWeaponCrit(Item item, Player player, ref int crit) {
 		}
 
+		/// <summary>
+		/// Allows you to change an item bait's rate of consumption, being able to force it to be consumed or stop it from being consumed or return null for unchanged vanilla behaviour.
+		/// </summary>
```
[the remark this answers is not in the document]
JPANv2: Clarified with a comment - True (force consume) takes precedence, in both consumeBait hooks (Player and Items), then false, then null (vanilla behaviour is to consume)

JPANv2:
I'll close this pull request, as it is vastly outdated and I could only start from scratch from tMod 1.4 branch. It is now part of issue #1139

Issue #1135, opened 2020-10-16
DamageClasses aren't properly distinguished, leading to stat boosts across all classes


ThomasSpeedrunner:
### Description
 When attempting to boost the damage or critical strike chance of any one specific class, the game instead gives that boost to all classes.

### Steps to reproduce
 Literally just equip any class-specific armor or accessory and examine a weapon of a different class.

### Additional Information
 After doing some hands-on experimentation with the DamageClass system, I've deduced that the most likely cause of the problem is due to vanilla DamageClasses not being loaded the same way as modded DamageClasses. There is an "index" variable which tells the system where the class in question is in the damage class data array --- however, this isn't set properly for the vanilla classes, causing them all to reference and modify the first entry --- which also means any class-specific bonuses get re-routed to what ends up being a "universal" entry in the array. My attempt at correcting this was to have the different vanilla classes get their IDs set manually within Load in the DamageClass class, like so
![image](https://user-images.githubusercontent.com/8083133/96200937-46c01000-0f29-11eb-849d-e42937fecd0f.png)
As the continued existence of this issue may have implied, this didn't work.

PR #1146, opened 2020-10-17
DamageClass additions and fixes.

ThomasSpeedrunner:
you may now use DamageClass.Throwing to define a throwing item/projectile, this is for modded convenience since everyone and their dog does a throwing class
you may now assign a DamageClass to projectiles through projectile.DamageType
also added an attempt at a fix for a problem where all DamageClasses are boosted by all stat bonuses for no good reason, and by that I mean I just made it so their indexes are actually set ever, it doesn't work 100% but it might help. see the issue I made on that for more info


Mirsario (approved):
Looks fine. I'm not sure why the fix here didn't work for #1135, will look into it later.

PR #1147, opened 2020-10-17
1.4 fairystuff

Agrair:
### What is the new feature?
A new feature to change how treasure fairies will react to tiles.

It allows modders to increase or decrease the amount of tiles for a fairy to pick a chunk of ore, and also to discount a chunk of ore entirely.

If you look closely, this actually only works for ores at the moment. There could probably be some more hooks to change how it decides between different tiles (atm, the fairy will almost always pick a chest over a chunk of Example Ore).

If a modder wants their ore to seem "extra special", or dynamically available to fairies depending on various conditions, this hook lets them do that.

### ExampleMod updates
I just added some stuff to Example Ore to make it work for fairies.


Agrair:
So this is just the updated version of the old PR. I dunno when I'm going to work on it, but I intend to eventually add support for other treasure finding mechanisms, and not just fairies.

PR #1148, opened 2020-10-17
Modified ProjectileLoader.CanDamage call.

ThomasSpeedrunner:
<s>CanDamage if set to true now skips the massive fuckfest of an individual projectile ID blacklist at the start of Damage()
added showCrit, showSpeed, and showMana to simplify making certain weapons show or not show crit chance, use speed, and mana cost</s>

CanDamage is now a nullable wrapper (thank you to Ozzatron for teaching me this terminology)
setting it to true allows a projectile to do damage no matter what, setting it to false disallows a projectile from doing damage no matter what, and leaving it unset lets vanilla call the shots


patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -112,19 +112,24 @@
...
  				return;
-+			
+ 
++			SkipWhateverTheHellThisHardcodedGroceryListIs:
 +			if (Main.projPet[type] && !ProjectileLoader.MinionContactDamage(this))
 +				return;
 +			
-+			if (!ProjectileLoader.CanDamage(this))
-+				return;
-
```
Mirsario, line 125: You do realize that you just removed vanilla logic in its entirety?

Mirsario (commented):
Use PR templates and save memeing for discord.

patches/tModLoader/Terraria/Projectile.cs.patch
Mirsario: And you made it actually do the damage when it returns false.

Mirsario (changes requested):
Marking my previous comments as change requests. That should be rolled back. We can talk about tooltips though.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -343,6 +342,27 @@
...
...pe == 600 || type == 601 || type == 602 || type == 535 || (type == 631 && localAI[1] == 0f) || type == 651 || (type == 188 && localAI[0] < 5f) || (aiStyle == 137 && ai[0] != 0f) || aiStyle == 138 || (type == 261 && velocity.Length() < 1.5f) || (type == 818 && ai[0] < 1f) || type == 831 || (type == 833 && ai[0] == 4f) || (type == 834 && ai[0] == 4f) || (type == 835 && ai[0] == 4f) || (type == 281 && ai[0] == -3f) || ((type == 598 || type == 636 || type == 614) && ai[0] == 1f) || (type == 923 && localAI[0] <= 60f) || (type == 919 && localAI[0] <= 60f) || (aiStyle == 15 && ai[0] == 0f && localAI[1] <= 12f) || type == 861 || (aiStyle == 93 && ai[0] != 0f && ai[0] != 2f) || (aiStyle == 10 && localAI[1] == -1f) || (Main.projPet[type] && type != 266 && type != 407 && type != 317 && (type != 388 || ai[0] != 2f) && (type < 390 || type > 392) && (type < 393 || type > 395) && (type != 533 || !(ai[0] >= 6f) || !(ai[0] <= 8f)) && (type < 625 || type > 628) && (type != 755 || ai[0] == 0f) && (type != 946 || ai[0] == 0f) && type != 758 && type != 951 && (type != 759 || frame == Main.projFrames[type] - 1) && type != 833 && type != 834 && type != 835 && type != 864 && (type != 623 || ai[0] != 2f)))
```
Chicken-Bones, line 354: This kind of patch is to be avoided if possible (which it is). I'm actually not quite sure why we needed a separate method here, couldn't the vanilla code be kept, and just move the ProjectlieLoader.CanDamage call above it?

Mirsario: Yes, this is super unnecessary and unmaintainable. Don't move or rewrite vanilla code, work around it.

Mirsario (changes requested):
I had a small discussion in discord, and everyone there at the time seemed to agree with the following - I think there's a way better implementation for tooltip toggling to be done. After tooltips receive somewhat similar treatment to player layers (1.4_playerlayers), we can most likely force all vanilla tooltips be registered at all times, but be hidden based on original logic. That way, you'll be able to enable whatever tooltips you would like by just finding them in the ModifyTooltips' provided dictionary under "Terraria" index, and make them visible by just setting a boolean to true. So, remove the tooltip patches. I'll merge the bool? CanDamage change, it seems good now.

ThomasSpeedrunner:
should be good now

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1523,6 +1524,15 @@
...
 +						toolTipLine[numLines] = crit + Lang.tip[5].Value;
++						toolTipNames[numLines] = "CritChance";
++						numLines++;
++					}
++					*/
++					if (!item.summon) {
++						int crit = player[myPlayer].GetWeaponCrit(item);
++						ItemLoader.GetWeaponCrit(item, player[myPlayer], ref crit);
++						PlayerHooks.GetWeaponCrit(player[myPlayer], item, ref crit);
++						toolTipLine[numLines] = crit + Lang.tip[5].Value;
```
Chicken-Bones, line 1535: You need to revert this change too. I'm also happy to merge when you do. This is an example of why it's best to split features into separate PRs because one might be approved and the other need to wait for a redesign after discussion

ThomasSpeedrunner: fixed up, should be good for real this time now

ThomasSpeedrunner:
blep. pretty sure it's fine now

patches/tModLoader/Terraria/ModLoader/GlobalProjectile.cs
```diff
@@ -158,12 +158,15 @@ public class GlobalProjectile:ModType
 		}
 
 		/// <summary>
-		/// Whether or not the given projectile is capable of killing tiles (such as grass) and damaging NPCs/players. Return false to prevent it from doing any sort of damage. Returns true by default.
+		/// Whether or not the given projectile is capable of killing tiles (such as grass) and damaging NPCs/players.
+		/// Return false to prevent it from doing any sort of damage.
+		/// Return true if you want the projectile to do damage regardless of the default blacklist.
+		/// Return nothing to let the projectile follow vanilla can-damage-anything rules. This is what happens by default.
```
Mirsario, line 164:
```suggestion
		/// Return null to let the projectile follow vanilla can-damage-anything rules. This is what happens by default.
```

patches/tModLoader/Terraria/ModLoader/ModProjectile.cs
```diff
@@ -234,10 +234,13 @@ public class ModProjectile : ModTexturedType
 		}
 
 		/// <summary>
-		/// Whether or not this projectile is capable of killing tiles (such as grass) and damaging NPCs/players. Return false to prevent it from doing any sort of damage. Returns true by default.
+		/// Whether or not this projectile is capable of killing tiles (such as grass) and damaging NPCs/players.
+		/// Return false to prevent it from doing any sort of damage.
+		/// Return true if you want the projectile to do damage regardless of the default blacklist.
+		/// Return nothing to let the projectile follow vanilla can-damage-anything rules. This is what happens by default.
```
Mirsario, line 240:
```suggestion
		/// Return null to let the projectile follow vanilla can-damage-anything rules. This is what happens by default.
```

patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
```diff
@@ -295,18 +295,17 @@ private class HookList
...
 			foreach (GlobalProjectile g in HookCanDamage.arr) {
-				if (!g.Instance(projectile).CanDamage(projectile)) {
-					return false;
+				bool? canDamage = g.Instance(projectile).CanDamage(projectile);
+				if (canDamage.HasValue) {
+					return canDamage.Value;
 				}
 			}
-			return true;
+			return projectile.modProjectile?.CanDamage();
```
Mirsario, line 308: As this uses nullable bools, this method should probably not short-circuit so simply. I think it should prefer returning false, if any hooks returned it? Then true, then null.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -146,19 +146,27 @@
...
  					projectile.projUUID = projectile.identity;
  
-@@ -8831,6 +_,12 @@
+@@ -8829,9 +_,20 @@
+ 		}
+ 
  		public void Damage() {
++			bool? shouldHitAnything = ProjectileLoader.CanDamage(this);
++			if (shouldHitAnything.HasValue) {
++				if (shouldHitAnything.Value == true)
```
Mirsario, line 155: herbert.png
```suggestion
+				if (shouldHitAnything.Value)
```

Mirsario (changes requested):

Mirsario (approved):
@Chicken-Bones what do you think about the current ProjectileLoader.CanDamage?

patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
```diff
@@ -295,18 +295,24 @@ private class HookList
...
-					return false;
+				bool? canDamage = g.Instance(projectile).CanDamage(projectile);
+
+				if (canDamage.HasValue) {
+					if (!canDamage.Value) {
+						return false;
+					}
+
+					result &= canDamage.Value;
 				}
```
Chicken-Bones, line 312: This doesn't work, because the starting value is null. This should just be = true

Chicken-Bones (changes requested):
Good PR. Please fix last thing then merge.

Also @Mirsario I think we could've done some of the final cleanup for him, rather than having a lot of small back and forth. Poor guy's got a lot of commits here

patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
ThomasSpeedrunner: what even is &= as an operator :thinkies:

Mirsario: a = a & b.

PR #1150, opened 2020-10-17
Gave KillTile an easier to understand summary

RyanMakesMods:
### What is the bug?
Not necessarily a bug, but made KillTile easier to read
### How did you fix the bug?
change from "fail determines whether the tile is mined" to that *and* "fail determines whether the tile is mined. Being true means not mined"
Nope.

### DIFF
--- a/patches/tModLoader/Terraria.ModLoader/GlobalTile.cs
+++ b/patches/tModLoader/Terraria.ModLoader/GlobalTile.cs
@@ -143,7 +143,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Allows you to determine what happens when the tile at the given coordinates is killed or hit with a pickaxe. Fail determines whether the tile is mined, effectOnly makes it so that only dust is created, and noItem stops items from dropping.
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria.ModLoader/GlobalTile.cs
```diff
@@ -143,7 +143,7 @@ public class GlobalTile
 		}
 
 		/// <summary>
-		/// Allows you to determine what happens when the tile at the given coordinates is killed or hit with a pickaxe. Fail determines whether the tile is mined, effectOnly makes it so that only dust is created, and noItem stops items from dropping.
+		/// Allows you to determine what happens when the tile at the given coordinates is killed or hit with a pickaxe. fail determines whether the tile is mined. Being true means not mined, effectOnly makes it so that only dust is created, and noItem stops items from dropping.
```
Mirsario, line 146:
```suggestion
		/// Allows you to determine what happens when the tile at the given coordinates is killed or hit with a pickaxe. If <paramref name="fail"/> is true, the tile will not be mined; <paramref name="effectsOnly"/> makes it so that only dust is created; <paramref name="noItem"/> stops items from dropping.
```

PR #1154, opened 2020-10-21
[1.4] safety aiStyle >= 0 for NoMultiplayerSmoothingByAI

direwolf420:
Adds a safety check for a new 1.4 added array that is indexed by aiStyle. Modded NPCs that have a custom ai and don't wish to use default ai (0) (which does targeting, spriteDirection setting and horizontal movement reduction), will error on that line in multiplayer.

### DIFF
--- a/patches/tModLoader/Terraria/NPC.cs.patch
+++ b/patches/tModLoader/Terraria/NPC.cs.patch
@@ -746,6 +746,15 @@
  					if (buffTime[i] < time)
  						buffTime[i] = time;
  
+@@ -71162,7 +_,7 @@
+ 				if (NPCID.Sets.NoMultiplayerSmoothingByType[type]) {
+ 					netOffset *= 0f;
+ 				}
+-				else if (NPCID.Sets.NoMultiplayerSmoothingByAI[aiStyle]) {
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -746,6 +746,15 @@
  					if (buffTime[i] < time)
  						buffTime[i] = time;
  
+@@ -71162,7 +_,7 @@
+ 				if (NPCID.Sets.NoMultiplayerSmoothingByType[type]) {
+ 					netOffset *= 0f;
+ 				}
+-				else if (NPCID.Sets.NoMultiplayerSmoothingByAI[aiStyle]) {
++				else if (aiStyle >= 0 && NPCID.Sets.NoMultiplayerSmoothingByAI[aiStyle]) {
```
Mirsario, line 754: Check for upper bounds too. There are zero conventions on how to use aiStyle.

Mirsario (changes requested):

patches/tModLoader/Terraria/NPC.cs.patch
direwolf420: done, baffling how they use this `Set` by indexing with ai styles anyway

PR #1155, opened 2020-10-21
Remove recipe ingredient and tile limits

Itorius:
Replaced recipe's required item and tile arrays with list => allowing for "infinite" number of ingredients/req. tiles.

There might be some uninteded behavior changes which need to be tested.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2429,6 +2429,34 @@
  				UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeBig = -1;
  				UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeSmall = -1;
  				if (numAvailableRecipes > 0)
+@@ -30626,7 +_,7 @@
+ 				if (numAvailableRecipes > 0) {
+ 					UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeBig = -1;
+ 					UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeSmall = -1;
+-					for (int num72 = 0; num72 < Recipe.maxRequirements; num72++) {
++					for (int num72 = 0; num72 < recipe[availableRecipe[focusRecipe]].requiredItem.Count; num72++) {
+ 						if (recipe[availableRecipe[focusRecipe]].requiredItem[num72].type == 0) {
+ 							UILinkPointNavigator.Shortcuts.CRAFT_CurrentIngridientsCount = num72 + 1;
+ 							break;
+@@ -30671,7 +_,8 @@
+ 						UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeSmall = 1 + num72;
+ 						Microsoft.Xna.Framework.Color color5 = inventoryBack;
+ 						inventoryBack = new Microsoft.Xna.Framework.Color((byte)num75, (byte)num75, (byte)num75, (byte)num75);
+-						ItemSlot.Draw(spriteBatch, ref recipe[availableRecipe[focusRecipe]].requiredItem[num72], 22, new Vector2(num73, num74));
++						Item tempItem = recipe[availableRecipe[focusRecipe]].requiredItem[num72];
++						ItemSlot.Draw(spriteBatch, ref tempItem, 22, new Vector2(num73, num74));
+ 						inventoryBack = color5;
+ 					}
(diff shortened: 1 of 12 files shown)


patches/tModLoader/Terraria/Recipe.Extensions.cs
```diff
@@ -47,17 +37,7 @@ public partial class Recipe
...
-				if (groupID == -1) break;
-
-				if (groupID == id) {
-					return true;
-				}
-			}
-
-			return false;
-		}
+		public bool HasRecipeGroup(int id) => acceptedGroups.Any(group => group == id);
```
Chicken-Bones, line 40: `Contains`. Please

Chicken-Bones (commented):
The code is good, but I honestly don't feel good about the maintenance cost here. Needs a second and third opinion.

Mirsario:
There are 2 reasons to do this. The first one is obvious, mods like calamity and maybe fargo's love making giant recipes with every item in the game. The second one is that this will help people directly modify recipes, since I've heard that we're removing RecipeFinder & Editor. I agree on their removal.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2429,6 +2429,34 @@
...
+ 							UILinkPointNavigator.Shortcuts.CRAFT_CurrentIngridientsCount = num72 + 1;
+ 							break;
+@@ -30671,7 +_,8 @@
+ 						UILinkPointNavigator.Shortcuts.CRAFT_CurrentRecipeSmall = 1 + num72;
+ 						Microsoft.Xna.Framework.Color color5 = inventoryBack;
+ 						inventoryBack = new Microsoft.Xna.Framework.Color((byte)num75, (byte)num75, (byte)num75, (byte)num75);
+-						ItemSlot.Draw(spriteBatch, ref recipe[availableRecipe[focusRecipe]].requiredItem[num72], 22, new Vector2(num73, num74));
++						Item tempItem = recipe[availableRecipe[focusRecipe]].requiredItem[num72];
++						ItemSlot.Draw(spriteBatch, ref tempItem, 22, new Vector2(num73, num74));
+ 						inventoryBack = color5;
```
Mirsario, line 2448: Shouldn't you set the list entry's value to tempitem here?

patches/tModLoader/Terraria/ModLoader/RecipeEditor.cs
```diff
@@ -28,17 +28,17 @@ public class RecipeEditor
...
-					return;
-				}
-				if (recipe.requiredItem[k].type == itemID) {
-					recipe.requiredItem[k].stack += stack;
-					return;
-				}
-			}
+			// for (int k = 0; k < Recipe.maxRequirements; k++) {
+			// 	if (recipe.requiredItem[k].type == 0) {
+			// 		recipe.requiredItem[k].SetDefaults(itemID, false);
```
Mirsario, line 33: Use comment blocks kthx

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -43,61 +53,102 @@
...
++			foreach (int num in acceptedGroups) {
 +				if (RecipeGroup.recipeGroups[num].ContainsItem(type)) {
  					theText = RecipeGroup.recipeGroups[num].GetText();
  					return true;
  				}
-@@ -82,20 +_,22 @@
+ 			}
+-
++			
+ 			theText = "";
```
Mirsario, line 100: whitespace

Mirsario (changes requested):

patches/tModLoader/Terraria/Main.cs.patch
Itorius: this is just for drawing the crafting menu, the actual item in the slot won't change

patches/tModLoader/Terraria/ModLoader/RecipeEditor.cs
Itorius: RecipeEditor will get removed so I'm not worried about this

patches/tModLoader/Terraria/Recipe.Extensions.cs
Itorius: Woops, will fix in the main branch once this PR gets merged

Mirsario: Just fixed this.

Mirsario (approved):

PR #1157, opened 2020-10-21
Make AmmoIDs constant

ExterminatorX99:
### What is the bug?
AmmoIDs aren't constant making them unusable in switch cases
### How did you fix the bug?
Made them constant
I don't think so

### DIFF
--- a/patches/tModLoader/Terraria.ID/AmmoID.cs.patch
+++ b/patches/tModLoader/Terraria.ID/AmmoID.cs.patch
@@ -1,0 +1,42 @@
+--- src/Terraria/Terraria.ID/AmmoID.cs
++++ src/tModLoader/Terraria.ID/AmmoID.cs
+@@ -2,22 +_,22 @@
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Hmm, would rather do this in the future via Roslyn. Does anyone actually try to switch on AmmoID?

I could accept it as it's unlikely to be hard to maintain. Will do so if approved by another dev

Mirsario (changes requested):
This is targetting the 1.3 master branch. If you want this in - target the `1.4` branch, as we're only accepting breaking changes there.

ExterminatorX99:
I don't see how this is a breaking change

Mirsario:
I have just tested it. Constants and static fields are not interchangeable. The previously build mod binaries will look for fields they won't be able to find.
image

ExterminatorX99:
Ah. You meant mod breaking changes. Got it.

PR #1158, opened 2020-10-22
SlopeIDs

pollen00:
### What is the new feature?
SlopeIDs
There are currently no Slope IDs in tModLoader and they will be useful to people who don't remember them.
Nope.
tile.slope(SlopeID.Solid);
### ExampleMod updates

This is a fairly straightforward feature, so ExampleMod examples aren't really needed.

### DIFF
--- a/patches/tModLoader/Terraria/ID/SlopeID.cs
+++ b/patches/tModLoader/Terraria/ID/SlopeID.cs
@@ -2,10 +2,25 @@
 {
 	public static class SlopeID
 	{
+		/// <summary>
+		/// A full tile.
+		/// </summary>
 		public static byte None = 0;
+		/// <summary>
+		/// A tile with the bottom left corner solid.
+		/// </summary>
 		public static byte DownLeft = 1;
+		/// <summary>
+		/// A tile with the bottom right corner solid.
+		/// </summary>
 		public static byte DownRight = 2;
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
`const` instead of `static` please. I'm fine with this. @JavidPack @Mirsario @Jofairden any input?

Jofairden (changes requested):
Same as chicken bones, constants are preferred for this
However, I'm fine with SlopeID . It's not really an ID though. It's more like a SlopeType

Jofairden (approved):

Mirsario:
I would prefer if slope ids were abstracted away with an enum property in Tile that includes half-blocks. The separation is uncalled for. @Itorius has been cleaning up Tile, I believe.

Itorius:
Indeed, see [this commit](https://github.com/tModLoader/tModLoader/blob/95abcc47ee7e08e9143551434f0a50709150b629/patches/tModLoader/Terraria/ID/SlopeID.cs).

Usage would be like this `tile.Slope = SlopeID.Solid` for setting or `tile.Slope == SlopeID.HalfBrick`

Mirsario:
@Itorius looks good. Make a PR of the branch.

@pollen0 I'll close this in favor of that, surely you don't mind.

PR #1162, opened 2020-10-23
[1.3] fix prechatbuttonclicked crashing on editing a grave

direwolf420:
### What is the bug?
When the player clicks the edit button on a grave, tml silently crashes because it tries to access an NPC with whoami -1.

### How did you fix the bug?
tml breaks up the if/elseif chain by inserting two hooks, what it doesn't account for is talkNPC being possibly -1. Vanilla logic prevents that by not continuing the elseif chain when a sign (or grave) is edited. The fix is actually copied from the code that handles the second NPC UI button, where this problem doesn't occur.

NA


Mirsario (approved):
Tested and discussed this. going to merge it. Port to 1.4 too, kthx.

PR #1170, opened 2020-10-28
[1.4] Fixed and reworked ModGore

Mirsario:
This PR does the following:
- Fixes ModGore, as it just hasn't been working for some reason I don't even know.
- Splits ModGore into GoreLoader, improve some code quality in general.
- Turns ModGore into a ModTexturedType. Reduced the amount of hardcode in autoloading.
- Changes how Texture-only mod gores load. Now, for every such gore, a dummy ModGore is registered with ModName/TextureName FullName. This means that gores can now be gotten using Mod/ModContent.Find/TryFind, which do not use texture paths.
- Mod.AddGore is replaced with GoreLoader.AddGoreFromTexture, which does what you would think, i.e. what I described above.
- Removes by-texture-path GetGoreSlot methods, for less hardcode and way less confusion. Content devs seem to support the switch from texture paths to ModName/ContentName.
- Renamed Gore.modGore to Gore.ModGore. We did plenty of naming convention fixes like this, should just do it for everything on the main 1.4 branch.
- Splits tML's added fields of Gore.cs into Gore.TML.cs.
- Fixes tML's added fields not being reset in Gore.NewGore.
- Temporarily added ModContent.GoreType<T>(). I would like to replace all such methods with a generalized one.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
@@ -78,8 +78,8 @@ namespace Terraria.ModLoader
 			GoreCount = GoreID.Count;
 		}
 
-		internal static ModGore SetupGore(Gore gore) {
-			gores.TryGetValue(gore.type, out var modGore);
+		internal static ModGore GetModGore(int type) {
+			gores.TryGetValue(type, out var modGore);
 			
 			return modGore;
 		}
--- a/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
@@ -112,6 +112,7 @@ namespace Terraria.ModLoader
 				return gore.ModGore.DrawBehind(gore);
 			}
 
+			//TODO: Whatever calls this is a very bad patch. Don't move vanilla code, reuse it where it is instead.
 			return (((gore.type >= 706 && gore.type <= 717) || gore.type == 943 || gore.type == 1147 || (gore.type >= 1160 && gore.type <= 1162)) && (gore.frame < 7 || gore.frame > 9));
 		}
 	}
--- a/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
@@ -78,16 +78,10 @@ namespace Terraria.ModLoader
 			GoreCount = GoreID.Count;
 		}
 
-		//in Terraria.Gore add modGore property (internal set)
-		//in Terraria.Gore.NewGore after resetting properties call ModGore.SetupGore(Main.gore[num]);
-		internal static void SetupGore(Gore gore) {
-			if (gores.ContainsKey(gore.type)) {
-				gore.ModGore = gores[gore.type];
-				gore.ModGore.OnSpawn(gore);
-			}
-			else {
-				gore.ModGore = null;
-			}
+		internal static ModGore SetupGore(Gore gore) {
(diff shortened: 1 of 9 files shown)


patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		//in Terraria.Main.DrawGore and DrawGoreBehind replace type checks with this
+		internal static bool DrawBackGore(Gore gore) {
+			if (gore.ModGore != null) {
+				return gore.ModGore.DrawBehind(gore);
+			}
+
+			return (((gore.type >= 706 && gore.type <= 717) || gore.type == 943 || gore.type == 1147 || (gore.type >= 1160 && gore.type <= 1162)) && (gore.frame < 7 || gore.frame > 9));
+		}
```
Mirsario, line 116: Bad patch I didn't make, will need to fix it later.

patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			foreach (var pair in gores) {
+				TextureAssets.Gore[pair.Key] = ModContent.GetTexture(pair.Value.Texture);
+			}
+		}
+
+		internal static void Unload() {
+			gores.Clear();
+			gores.Clear();
```
Chicken-Bones, line 77: hmm

patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+			GoreCount = GoreID.Count;
+		}
+
+		//in Terraria.Gore add modGore property (internal set)
+		//in Terraria.Gore.NewGore after resetting properties call ModGore.SetupGore(Main.gore[num]);
+		internal static void SetupGore(Gore gore) {
+			gore.ResetNewFields();
```
Chicken-Bones, line 84: I'd prefer this be called from `Gore` and be private.

patches/tModLoader/Terraria/ModLoader/GoreLoader.cs
```diff
@@ -0,0 +1,118 @@
...
+
+			if (gores.ContainsKey(gore.type)) {
```
Chicken-Bones, line 86: maybe this whole method should just return `ModGore` and use a `TryGetValue`? Then in `Gore` itself you can call `gore.ModGore?.OnSpawn(this);`

patches/tModLoader/Terraria/ModLoader/ModGore.cs
```diff
@@ -1,117 +1,39 @@
...
-		}
+		public int Type { get; internal set; }
 
-		internal static void SetupUpdateType(Gore gore) {
-			if (gore.modGore != null && gore.modGore.updateType > 0) {
-				gore.realType = gore.type;
-				gore.type = gore.modGore.updateType;
-			}
-		}
+		public override string Name => nameOverride ?? base.Name;
```
Chicken-Bones, line 18: can we do a `ModGoreSimple` subclass instead?

Chicken-Bones (commented):
fantastic

patches/tModLoader/Terraria/ModLoader/ModGore.cs
Mirsario: GoreLoader.AddGoreFromTexture<T> is currently usable with custom ModGores too.

Chicken-Bones: good point

PR #1172, opened 2020-10-29
Fixed multiple accessory-related issues.

ThomasSpeedrunner:
- VanillaUpdateAccessory no longer erroneously uses player.whoAmI as an input.
- VanillaUpdateAccessory and ApplyEquipsVanity now account for extra acc slots properly, preventing the game from hanging upon having one.
- extraAccessorySlots now properly updates itself in ResetEffects.
- skyStoneEffects (Sun/Moon Stone statsheet) is no longer applied for every slot starting with the one where the stone in question is equipped.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Default/Patreon/Orian.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Default/Patreon/Orian.cs
@@ -45,7 +45,7 @@ namespace Terraria.ModLoader.Default.Patreon
 			item.height = 18;
 		}
 
-		public override void UpdateVanity(Player player, EquipType type)
+		public override void EquipFrameEffects(Player player, EquipType type)
 		{
 			player.shoe = 0;
 		}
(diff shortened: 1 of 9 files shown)


Chicken-Bones:
`extraAccessorySlots` is never read by vanilla, only written to. The number of patches required to make it actually work is undesirable. Especially with the 'visible but not active' effects and `IsAValidEquipmentSlotForIteration`. It's a mess all round. Mods increasing this number is a bad idea, they wouldn't save and load properly, which slot belongs to which mod, just... no can do.

I've fixed up all the hooks and patches (kept them real small) related to accessory updating. I _haven't_ actually playtested this, so please do some testing for me.

`skyStoneEffects` problem should still be fixed.

ThomasSpeedrunner:
Noticed and fixed up a small problem where "k" in VanillaUpdateEquip wasn't actually defined for the fake loop, but past that everything seems to be in working order. My testing char no longer crashes when having the Demon Heart's extra acc slot, which is a good sign. :hahayes:

Issue #1142, opened 2020-10-16
[1.4] Allow for choosing whether or not a GlobalX instance should be added onto an entity.


Mirsario:
To properly encourage good OOP without god-classing, it's important to let users have their GlobalX classes only be added to some of the entities. For example, in Overhaul, I had cases where I would need to change animations for 1 specific NPC (Raven), or do a feature that only applies to content with a specific tags (#1083) (Fire collision checks, npc bleeding, item physics improvements, pretty much everything.) It is, of course, not ideal for modders to do checks in literally every hook override, as that is both annoying and may extremely hinder performance in a well object-oriented mod that follows the single responsibility principles.

A boolean getter property would be perfect for this. The problem with implementing this, however, is again, optimization related, because the current hooklists will no longer be sufficient, and will need to be stored on per-entity basis.  I think I'll probably only get on implementing this myself when we get source generators after #1086 is merged and updated to *by then hopefully stable* .NET 5.

PR #1177, opened 2020-11-01
[1.4] Conditional GlobalX instancing

Mirsario:
Implements #1142 

### What is the new feature?
This PR adds conditional instancing for GlobalX classes, which lets people control whether or not they want their GlobalX class to affect a specific entity. It does so by introducing a `bool InstanceForEntity(Item/NPC/Projectile entity)` hook.

Because without this, good object-oriented programming is quite annoying to do in many contexts, and could heavily slow down the game if all mods do it.

In terms of the changes to the hooklists, I would say that the current implementation is a bit temporary. I would like to redo it later when we get source generators, with .NET Core.

### ExampleMod updates
See https://github.com/tModLoader/tModLoader/blob/1.4_conditionalinstancing/ExampleMod/Common/GlobalNPCs/GuideGlobalNPC.cs?ts=4


patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -35,12 +35,32 @@ public static class NPCLoader
...
 		{
-			public GlobalNPC[] arr = new GlobalNPC[0];
+			public int[] arr = new int[0];
 			public readonly MethodInfo method;
 
 			public HookList(MethodInfo method) {
 				this.method = method;
 			}
+
+			public IEnumerable<GlobalNPC> Enumerate(NPC npc) {
```
Chicken-Bones, line 45: I've long since banned anything except array foreach from tML in hooks. Because of allocations and the inability for jit to optimise it. 

Specifically, if we're at the point where we want to avoid empty function call overhead (the whole point of building hooklists) then we're on the same order of magnitude as non-optimised loops. Seriously, instruction pre-fetch is mandatory for modern day performance.

That said, C# _recently_ added support for structs implementing IEnumerator. There's a good chance that a simple enough one could be optimised by the JIT and it won't allocate, so, try it!

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -35,12 +35,32 @@ public static class NPCLoader
...
+				for (int i = 0; i < arr.Length; i++) {
+					var g = npc.globalNPCs[arr[i]];
```
Chicken-Bones, line 47: honestly it might be better for memory (and pretty similar or better for performance) to swap the loop indexing here. Have an `NPC` store the `globalNPCs` that are relevant for it, and then the hooklist have a `bool[]` for returning it in the iterator here.

Agrair:
I can't really tell, but will this allow for modders to "attach" a GlobalX to an already existing entity? For example, if I were to implement a system that transforms items by adding "augments" at a special crafting station with a UI and everything, I would like to be able to do something like `item.AddGlobal(new AugmentSpecification())` where `AugmentSpecification` is a `GlobalItem` that does something when the item is used/equipped, etc. And then ofc, a function I can use to remove it. Looking at the code, I'm not sure if that's possible with the given system, but I personally can't really tell what you did anyway from these diffs.

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
Agrair: Erm, if you're so willing, could you elaborate as to what this means? I'm just really curious. It doesn't really pertain to me otherwise, but what makes non-array foreach not JIT optimizable, why is that related to allocation, and what's instruction pre-fetch?

Mirsario:
No, that won't be possible, and I don't really see why it should be. It would complicate things for sure.

Agrair:
I guess I could implement a `GlobalItem` that manages that for me.

Chicken-Bones:
My idea for performance:
`InstancePerEntity` stays as a `bool`. `GlobalX` are either instanced or not.
Add `ApplyToEntity(Item/NPC/Projectile entity)`
`HookList`s generate an `int[]` of the indices into the `GlobalX[]` for which the method is overridden
when a hook is called, we loop over the indices, but rather than our old strategy of calling `GetInstance(entity)` which checked `InstancePerEntity` and a fixed "instanced index", we loop through the entity's instances to find this one.
The trick is we don't need a double loop, because the order of global items in both arrays (instanced and hook list) is the same, so we can just pick up from where we left off last iteration.

Then we can also do things like:
Change the instanced `GlobalX[]` to a `GlobalInstance<T>[]` where `GlobalInstance` is a struct containing a `GlobalX (T)` and a `short` with the `index` of the global, same as the indices used in the hook lists. That way when looping through an entities instanced globals for each hook call, we don't actually have to retrieve globals from the heap to check if they implement a hook (if their id is in hook list).

And later on, we could `Emit` the hook calling methods completely, baking the hook id array into the IL and calling `ApplyToEntity` automatically

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -34,18 +33,35 @@ public abstract class GlobalItem : ModType
...
+		/// <summary>
+		/// Use this to control whether or not a GlobalItem instance should be created for the provided Item instance.
+		/// </summary>
+		/// <param name="item"> The item for which the global instantion is being checked. </param>
+		/// <param name="lateInstantiation">
+		/// Whether this check occurs before or after the ModItem.SetDefaults call.
+		/// <br/> If you're relying on Item values that can be changed by that call, you should likely prefix your return value with the following:
+		/// <code> lateInstantiation &amp;&amp; ... </code>
+		/// </param>
+		public virtual bool InstanceForEntity(Item item, bool lateInstantiation) => true;
```
Chicken-Bones, line 45: This needs to be named `AppliesToEntity`

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -34,18 +33,35 @@ public abstract class GlobalItem : ModType
...
+
+		public GlobalItem Instance(Item item) {
```
Chicken-Bones, line 47: forward to a common `Instance<T>(Instanced<T>[] instances, short index)`

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -14,15 +14,14 @@ namespace Terraria.ModLoader
...
-		internal int instanceIndex;
+		internal short index;
 
 		protected sealed override void Register() {
 			ItemLoader.VerifyGlobalItem(this);
 
 			ModTypeLookup<GlobalItem>.Register(this);
 			
-			index = ItemLoader.globalItems.Count;
+			index = (short)ItemLoader.globalItems.Count;
```
Chicken-Bones, line 24: need an exception if item count ever passes a `short`
Also use `ushort` everywhere for indices

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -34,18 +33,35 @@ public abstract class GlobalItem : ModType
 		/// </summary>
 		public virtual bool InstancePerEntity => false;
```
Chicken-Bones, line 34: We should move this and `InstanceForEntity` into a common `GlobalType` class

patches/tModLoader/Terraria/ModLoader/Instanced.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Terraria.ModLoader
+{
+	/// <summary> The purpose of this struct is to micro-optimize lookups of GlobalX indices by providing these associations without additional retrievals from the heap. </summary>
+	[StructLayout(LayoutKind.Sequential, Pack = 2)]
+	internal readonly struct Instanced<T>
```
Chicken-Bones, line 7: consider a common interface for 'indexables' or use `GlobalType`

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -15,6 +15,7 @@
 using Terraria.ModLoader.IO;
 using Terraria.UI;
 using Terraria.Utilities;
+using HookList = Terraria.ModLoader.Core.HookList<Terraria.ModLoader.GlobalItem>;
```
Chicken-Bones, line 18: yes!

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -153,19 +144,33 @@ private class HookList
...
-				g.Instance(item).SetDefaults(item);
+			//Could potentially be sped up.
+			var lateInitGlobals = globalItems
+				.Where(g => !item.globalItems.Any(i => i.index == g.index) && g.InstanceForEntity(item, true))
+				.Select(g => new Instanced<GlobalItem>(g.index, g.InstancePerEntity ? g.Clone(item, item) : g));
+
+			item.globalItems = item.globalItems
+				.Union(lateInitGlobals)
+				.OrderBy(i => i.index)
+				.ToArray();
```
Chicken-Bones, line 163: Move to common, say.. `LoaderUtils`? Either 2 methods, or one with a lambda passed?

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -22,21 +22,46 @@ public partial class Projectile
 		/// <summary> Gets the instance of the specified GlobalProjectile type. This will throw exceptions on failure. </summary>
 		/// <exception cref="KeyNotFoundException"/>
 		/// <exception cref="IndexOutOfRangeException"/>
-		public T GetGlobalProjectile<T>() where T : GlobalProjectile
-			=> GetGlobalProjectile(ModContent.GetInstance<T>());
+		public T GetGlobalProjectile<T>(bool exactType = true) where T : GlobalProjectile {
```
Chicken-Bones, line 25: neat, I see the use for this now. As with the others, _generalise_ please

patches/tModLoader/Terraria/NPC.TML.cs
```diff
@@ -26,26 +26,51 @@ public partial class NPC
...
+			}
+
+			result = default;
+
+			return false;
+		}
 
 		/// <summary> Safely attempts to get the local instance of the type of the specified GlobalNPC instance. </summary>
 		/// <returns> Whether or not the requested instance has been found. </returns>
 		public bool TryGetGlobalNPC<T>(T baseInstance, out T result) where T : GlobalNPC {
```
Chicken-Bones, line 72: generalise?

Chicken-Bones (changes requested):
Design is approved, I'm happy. Changes requested are all about de-duplicating code (and one naming qualm).

Would be nice to see some profiling, on construction and some high traffic hooks (eg draw), but we may only get that when mods start abusing them.

patches/tModLoader/Terraria/ModLoader/GlobalType.cs
```diff
@@ -0,0 +1,91 @@
...
+		/// <summary>
+		/// Whether to create a new instance of this Global for every entity that exists. 
+		/// Useful for storing information on an entity. Defaults to false. 
+		/// Return true if you need to store information (have non-static fields).
+		/// </summary>
+		public virtual bool InstancePerEntity => false;
+
+		internal GlobalType() { }
+
+		internal static T Instance<T>(Instanced<T>[] globals, ushort index) where T : GlobalType {
```
Chicken-Bones, line 18: `public` so mods can use it? (or define their own globals)?

patches/tModLoader/Terraria/ModLoader/GlobalType.cs
```diff
@@ -0,0 +1,91 @@
...
+			for (int i = 0; i < globals.Length; i++) {
+				if (globals[i].instance is TResult result) {
+					return result;
+				}
+			}
+
+			throw new KeyNotFoundException($"Instance of '{typeof(TResult).Name}' does not exist on the current {typeof(TEntity).Name}.");
+		}
+
+		internal static bool TryGetGlobal<TEntity, TGlobal, TResult>(Instanced<TGlobal>[] globals, TResult baseInstance, out TResult result) where TGlobal : GlobalType<TEntity> where TResult : TGlobal {
```
Chicken-Bones, line 47: This doesn't actually need the `TEntity` param here, as you can do everything with the non-generic `GlobalType` constraint. And similar elsewhere

Chicken-Bones (approved):

PR #1179, opened 2020-11-05
Added ModWorld.ModifyTimeRate.

ThomasSpeedrunner:
it's about time, as they say
edit the `timeRate` parameter to modify the speed at which time progresses, edit the `tileUpdateRate` parameter to modify the speed at which tiles in the world update

can be useful for items or events which are meant to speed up or slow down time


Chicken-Bones:
This is an unnecessary feature, where modders could just use an on hook

```
On.Terraria.Main.UpdateTimeRate += orig => {
    orig();
    if (sonic)
        Main.dayRate *= 2;
}
```

But since it's a simple hook and probably desired by a few modders, I'll approve it

Chicken-Bones (approved):

Mirsario (approved):

PR #1180, opened 2020-11-05
Add API for custom containers

Itorius:
**To be filled in**

> From Ozzatron:
> Calamity's weapon chargers are containers with two slots. The two slots both have different arbitrary criteria for what can be placed in them. The container can be opened by all players simultaneously and has a function where it automatically updates without player interaction
> The contents of the container are synced across the network every time they change
> ( It's not currently a container but it COULD be )

This could be easily done using ItemHandler. Filtering of items can be done with `IsItemValid` and syncing is possible with `OnContentsChanged`


patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -304,7 +304,7 @@ public abstract class ModTile : ModBlockType
 		/// <param name="i">The x position in tile coordinates.</param>
 		/// <param name="j">The y position in tile coordinates.</param>
 		/// <param name="nextSpecialDrawIndex">The special draw count. Use with Main.specX and Main.specY and then increment to draw special things after the main tile drawing loop is complete via DrawSpecial.</param>
-		public virtual void DrawEffects(int i, int j, SpriteBatch spriteBatch, ref Color drawColor, ref int nextSpecialDrawIndex) {
+		public virtual void DrawEffects(int i, int j) {
 		}
```
Mirsario, line 308: Reasoning? + The docs are outdated.

PR #1182, opened 2020-11-07
Actually fix ExampleShield (best practice)

Chicken-Bones:
The dash accessory was implemented by enumerating the player's accessory slots. This approach is contrary to the vanilla and tML design where accessories enable capabilities on the player, and prevents all accessory related hooks or custom slots from working with it.

The new design also puts the movement code in the `PreUpdateMovement` hook, consistent with other movement affecting code.

Overall operation should be easier to understand. Posting here for feedback/thoughts.

### DIFF
--- a/ExampleMod/Content/Items/Accessories/ExampleShield.cs
+++ b/ExampleMod/Content/Items/Accessories/ExampleShield.cs
@@ -29,57 +29,7 @@ namespace ExampleMod.Content.Items.Accessories
 		public override void UpdateAccessory(Player player, bool hideVisual) {
 			player.allDamage += 1f; // Increase ALL player damage by 100%
 			player.endurance = 1f - (0.1f * (1f - player.endurance));  // The percentage of damage reduction
-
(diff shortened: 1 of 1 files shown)


Itorius (approved):
Looks fine to me

Mirsario (approved):
This is definitely better, but 1 class per file please.

Chicken-Bones:
> This is definitely better, but 1 class per file please.

What's our ExampleMod policy? should multi-class examples go in a separate namespace/folder?

Mirsario:
> What's our ExampleMod policy? should multi-class examples go in a separate namespace/folder?

I would make a namespace if a piece of content requires additional new data classes/structs, but if it's a ModPlayer, I would probably prefer putting it in Common.Players (or was it Common.ModEntities.Players?), separate from the content piece that uses it. The point is, things like items and tiles they place/projectiles they shoot are already separate, even if that tile/projectile is only used by that item.

Chicken-Bones:
Going to deal with the one class per file problem later.

PR #1185, opened 2020-11-08
Removed redundant code in FrameworkVersion.cs

ChrisDirkis:
Pretty much exactly what it says on the tin, tiny change to clean up some code in FrameworkVersion.cs


ChrisDirkis:
Talked with CB, the Framework.Unknown check was probably supposed to return early instead. Disregard this :)

Chicken-Bones:
As above, I'd rather the missing `return;` be added, but this code path will be made redundant during the .NET Core port so there's no urgency here.

PR #1188, opened 2020-11-10
added missing ItemLoader.ConsumeItem

Agrair:
Updated quick buff function


Agrair:
Force push is because itor decided to NOT use a branch :ech:

Mirsario (approved):

PR #1197, opened 2020-11-12
[1.4] Some 1.4 reimplementations from patching_todo.txt

direwolf420:
### Description
Reimplemented a few calls (ItemLoader.DrawAnimatedItem, PlayerHooks.UpdateAutopause, PlayerHooks.SendClientChanges and ModHooks.MidUpdateProjectileItem) while staying true to its surrounding 1.3 logic

### Remarks
Projectile drawing got a new structure, with `if/return` instead of `if/else if`, making implementing PreDraw, PreDrawExtras and PostDraw more complicated.


direwolf420:
Also tested them all @Mirsario

PR #1202, opened 2020-11-14
made TagSerializer implement ILoadable

Agrair:
### What is the new feature?
TagSerializer implements ILoadable.

It simply grants peace of mind to the average modder. I think that whenever possible with these kind of OOP objects that only get instantiated once, at load time, they should always be automatically handled. Unless they have to explicitly defined, which TagSerializers usually don't have to be, the API should just do it for the user.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
@@ -72,7 +72,7 @@ namespace Terraria.ModLoader.IO
 			return null;
 		}
 
-		protected override void Register() {
+		protected sealed override void Register() {
 			AddSerializer(this);
 		}
 	}
--- a/patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
@@ -7,7 +7,7 @@ using Terraria.DataStructures;
 
 namespace Terraria.ModLoader.IO
 {
-	public abstract class TagSerializer : ILoadable
+	public abstract class TagSerializer : ModType
 	{
 		public abstract Type Type { get; }
 		public abstract Type TagType { get; }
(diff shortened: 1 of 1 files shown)


BipPip:
poggers

patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
```diff
@@ -71,6 +71,13 @@ public abstract class TagSerializer
 
 			return null;
 		}
+
+		public void Load(Mod mod) {
+			AddSerializer(this);
```
Mirsario, line 76: AddSerializer should be made internal, since you made it be superseded by mod.AddContent.

Mirsario (changes requested):
Cool and good

patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Terraria.ModLoader.IO
 {
-	public abstract class TagSerializer
+	public abstract class TagSerializer : ILoadable
```
Mirsario, line 10: Actually, shouldn't it be ModType here? That has virtual Load and Unload.

Agrair:
Internalized AddSerializer and changed the ILoadable.Load to be ModType.Register.

Mirsario (approved):

patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
```diff
@@ -51,7 +51,7 @@ public abstract class TagSerializer
 			return false;
 		}
 
-		public static void AddSerializer(TagSerializer serializer) {
+		internal static void AddSerializer(TagSerializer serializer) {
 			serializers.Add(serializer.Type, serializer);
```
Chicken-Bones, line 55: So um, should we search tML itself for `ILoadable` ? I'm seeing more and more use for such a thing.

Chicken-Bones (approved):
Cool and also good

patches/tModLoader/Terraria/ModLoader/IO/TagSerializer.cs
Mirsario: I most definitely wanted autoloading in the default mod for at least the content, yes. Make an issue.

Issue #876, opened 2020-05-14
Camera mode ignores modded backgrounds


direwolf420:
### Description
Being in a place where a modded background shows, then pressing F1 for camera mode, making sure "Capture background" is on, and doing a snapshot still shows the vanilla background.

### Steps to reproduce
Load Examplemod, place as many "Example Block" tiles as needed for the background to appear, press F1, select enough area to see the difference, press snapshot. Then compare against what you see ingame.

### Reproduction frequency
100%

### Additional Information
This happens with just `public override bool ChooseBgStyle() => true;` too, so not related to examplemod per say.

Screens (had to up contrast cause I took it during ingame night): https://prnt.sc/sgzmpk

Jofairden:
I remember this being an issue I thought it was fixed, thx for report!

PR #1205, opened 2020-11-15
ModBiome! (and some reorganization of various fx)

Agrair:
### What is the new feature?
`ModBiome`. Specifically, all biome related hooks from `ModPlayer` have been moved into a new type.
Not only that, but quite a few other hooks relating to "atmospheric" functionality have been centralized and shuffled around. Please see solxan's more in depth comment below on this thread.

It further encourages the reduction of godclassing first of all, but now there are no more ugly `ZoneMyNewFancyBiome` `bool`s (at least, they were ugly in my opinion, and I hope others agree with me). There is now just `player.Active(ModContent.GetInstance<MyNewFancyBiome>()`, and tML will handle the rest.
As for the atmospheric functionality changes, a lot of priority management has been coalesced in to what will hopefully be a more manageable state.

Open to suggestions.

N/A (Not going to put anything here because I hope it's really obvious what the usage is)

### ExampleMod updates
None, yet. If enough people approve this PR , then someone could port the old Example Biome stuff (I think kittenchilly's mod generation PR may interfere with this as he also did some bits with Example Biome)


patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -0,0 +1,68 @@
...
+		/// Whether or not the player is in this specific biome.
+		/// </summary>
+		public bool Active { get; internal set; }
+
+		protected override void Register() {
+			ModTypeLookup<ModBiome>.Register(this);
+			BiomeLoader.Add(this);
+		}
+
+		internal void Update() {
```
Agrair, line 28: This concept of `Active` and the internalized `Update` method differ from I think all current hook types if I'm not mistaken, but it was the best I could come up with. I'm open to suggestions for changes.

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
```diff
@@ -0,0 +1,142 @@
...
+				data = stream.ToArray();
+			}
+			if (data.Length > 0) {
+				writer.Write(modBiome.Mod.Name);
+				writer.Write(modBiome.Name);
+				writer.Write((byte)data.Length);
+				writer.Write(data);
+				return true;
+			}
+			return false;
```
Agrair, line 102: I just realized, this is probably creating a little bit of overhead. This could be replaced with
```
if (modBiome.Active)
{
    writer.Write(modBiome.Mod.Name);
    writer.Write(modBiome.Name);
    writer.Write(sizeof(bool));
    writer.Write(modBiome.Active);
}
```
Maybe not even that. I can't remember exactly, but do `false` values ever get sent?

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
```diff
@@ -0,0 +1,142 @@
...
+
+				if (ModContent.TryFind<ModBiome>(modName, name, out var modBiomeBase)) {
+					var modBiome = player.GetModBiome(modBiomeBase);
+
+					using MemoryStream stream = new MemoryStream(data);
+					using BinaryReader customReader = new BinaryReader(stream);
+
+					modBiome.Active = customReader.ReadBoolean();
+					//try { modPlayer.ReceiveCustomBiomes(reader) }
+				}
```
Agrair, line 121: Again, this stuff with `BinaryWriter(MemoryStream)` is now going to cause overhead. I'll fix that in a bit.

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
Agrair: Actually, I'm a little confused here. There seems to be a few more `BinaryWriter` instances than there has to be, collectively in this method and the one super-method that calls it. Could someone please look at that? Since the `bool`s are now explicitly defined, I think all of them except the one passed in could just be removed.

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -0,0 +1,68 @@
...
+		/// <summary>
+		/// Allows you to create special visual effects in the area around the player. For example, the blood moon's red filter on the screen or the slime rain's falling slime in the background. You must create classes that override Terraria.Graphics.Shaders.ScreenShaderData or Terraria.Graphics.Effects.CustomSky, add them in your mod's Load hook, then call Player.ManageSpecialBiomeVisuals. See the ExampleMod if you do not have access to the source code.
+		/// </summary>
+		public virtual void UpdateBiomeVisuals() {
+		}
+
+		/// <summary>
+		/// Exactly the same as ModWorld.ModifyWorldGenTasks except it's tied to this ModBiome.
+		/// </summary>
+		public virtual void ModifyWorldGenTasks(List<GenPass> tasks, ref float totalWeight) {
```
Agrair, line 65: Uhhh, I literally just copied this from `ModWorld`. I think it's getting phased out anyway, but I wasn't sure how to replace it. I was thinking of maybe autoloading nested types as `GenPass` objects?

Agrair:
Whoops, I wasn't sure what that button did. I was hoping it meant "ask team member to review" 😆

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -24,6 +24,7 @@ public class ModPlayer:ModType
 		internal int index;
 
 		internal ModPlayer CreateFor(Player newPlayer) {
+			BiomeLoader.CopyBiomesTo(newPlayer);
```
Agrair, line 27: Uhh, whoops.

Agrair:
Ok so I wrote this as if ModBiomes are responsible for data, but that’s absurd. Player obejcts should be passed in as parameters to the hooks and ModPlayer.modBiomes replaces with a simple Boolean array of biomes that registered as active.

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
Mirsario: I don't think that biomes should deal with generation. Not in terraria.

Agrair: So no generation hooks at all? Just let ModWorlds keep it?

kittenchilly: Yeah, leave generation to ModWorld since vanilla doesn't really couple them with biomes in the first place

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -69,5 +69,18 @@ public partial class Player
 
 		/// <summary> Gets the reference to the damage modifier for this damage type on this player. Since this returns a reference, you can freely modify this method's return value with operators. </summary>
 		public ref Modifier GetDamage(DamageClass damageClass) => ref damageData[damageClass.index].damage;
+
+		internal bool[] modBiomeFlags = new bool[0];
+
+		/// <summary> Gets the instance of the specified ModBiome type. This will throw exceptions on failure. </summary>
+		/// <exception cref="IndexOutOfRangeException"/>
+		public bool IsModBiomeActive<T>() where T : ModBiome
```
Agrair, line 77: I might change this to "InModBiome", just to make more sense to modders.

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -0,0 +1,52 @@
...
+	public abstract class ModBiome : ModType
+	{
+		internal int index;
+
+		protected override void Register() {
+			ModTypeLookup<ModBiome>.Register(this);
+			BiomeLoader.Add(this);
+		}
+
+		internal void Update(Player player, ref bool value) {
```
Chicken-Bones, line 18: I think it makes more sense to move this method into the loop of `BiomeLoader.Update`

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -0,0 +1,52 @@
...
+		/// <summary>
+		/// Override this hook to make things happen when the player leaves the biome.
+		/// </summary>
+		protected virtual void OnLeave(Player player) {
+		}
+
+		/// <summary>
+		/// Allows you to create special visual effects in the area around the player. For example, the blood moon's red filter on the screen or the slime rain's falling slime in the background. You must create classes that override Terraria.Graphics.Shaders.ScreenShaderData or Terraria.Graphics.Effects.CustomSky, add them in your mod's Load hook, then call Player.ManageSpecialBiomeVisuals. See the ExampleMod if you do not have access to the source code.
+		/// </summary>
+		public virtual void UpdateBiomeVisuals(Player player) {
```
Chicken-Bones, line 49: multiple biomes may be simultaneously active. we should provide a priority mechanism to handle background drawing or music. 

Unlike player layers, here at may make sense to continually update the ordering, at least to find the 'most active' biome. I can imagine a system with several levels or categories, and then within categories a resolution via a weight hook (normally block count).

Eg, ice biome vs hallowed biome could be whichever block count is higher, but celestial pillar (event biome) takes priority

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
```diff
@@ -0,0 +1,92 @@
...
+
+		public static void CopyCustomBiomesTo(Player player, Player other) {
+			for (int i = 0; i < player.modBiomeFlags.Length; i++) {
+				other.modBiomeFlags[i] = player.modBiomeFlags[i];
+			}
+		}
+
+		public static void SendCustomBiomes(Player player, BinaryWriter writer) {
+			for (int i = 0; i < player.modBiomeFlags.Length; i++) {
+				writer.Write(player.modBiomeFlags[i]);
```
Chicken-Bones, line 72: bit packing please. Honestly the whole `modBiomeFlags` should probably be https://docs.microsoft.com/en-us/dotnet/api/system.collections.bitarray?view=netframework-4.5 instead of `bool[]`

Note that `BitArray` is optimised for performance. One day we might consider (or get yorai to consider) them for ID.Sets as well

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
```diff
@@ -0,0 +1,92 @@
...
+				if (player.modBiomeFlags[i] ^ other.modBiomeFlags[i])
+					return false;
+			}
+			return true;
+		}
+
+		public static void CopyCustomBiomesTo(Player player, Player other) {
+			for (int i = 0; i < player.modBiomeFlags.Length; i++) {
+				other.modBiomeFlags[i] = player.modBiomeFlags[i];
+			}
```
Chicken-Bones, line 67: `BitArray.Clone`?

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
```diff
@@ -0,0 +1,92 @@
...
+				biomes[i].Update(player, ref player.modBiomeFlags[i]);
+			}
+		}
+
+		public static bool CustomBiomesMatch(Player player, Player other) {
+			for (int i = 0; i < player.modBiomeFlags.Length; i++) {
+				if (player.modBiomeFlags[i] ^ other.modBiomeFlags[i])
+					return false;
+			}
+			return true;
```
Chicken-Bones, line 61: `BitArray.Equals`?

Chicken-Bones (commented):
Good stuff

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
Agrair: Hmm, well for `BitArray`, the internal backing field isn't exposed at all (`m_array`). When I use `BitArray`, I just use reflection to retrieve it, but I don't think that's the right way to serialize it, is it?

patches/tModLoader/Terraria/ModLoader/BiomeLoader.cs
Agrair: Tested it, and it doesn't look like `BitArray` implements a value based override for `Equals`, unfortunately.

Chicken-Bones: huh, I checked the reference source and you're right. That's mega lame... Guess we can make our own?

Solxanich:
To Dos:
- [ ]  GlobalBiome neccessary? Move UpdateBiome there?
- [ ]  Complete :bool ModType.TryFind or something similar (returns true if player is inside the biome, false if not) - https://github.com/tModLoader/tModLoader/pull/1205#discussion_r524393010
- [ ]  Hook Bool BiomeConditions, return true if the player is in the biome, false if not (false default). Shows music, spawn conditions, etc
- [ ]  Hook OnInBiome, make stuff happen when a player is in the biome outside of background and music.
- [ ]  Set Background when in the biome
- [ ]  Set Map Background
- [ ]  Set Music (music should be able to change based on conditions, for stuff like night tracks, etc)
- [ ]  Set Localizable Name (that will be used anywhere where the biome name is shown)
- [ ]  Set Bestiary Tab icon, every biome in the game has one so ModBiome will automatically set it with a placeholder icon if not set here. (ModEvent or something similar should have this too). Should create a SpawnConditionBestiaryInfoElement as well.
- [ ]  Specify an ordering with respect to other biomes (including vanilla ones) similar to PlayerDrawLayers. A player can be in multiple biomes simultaneously, and they should all apply for drop rules and spawns, but things like background and music need to prioritise based on order.
- [ ] Convert modBiomeFlags to BitArray
- [ ] https://github.com/tModLoader/tModLoader/pull/1205#discussion_r525138467
- [ ] Check: https://github.com/tModLoader/tModLoader/pull/1205#discussion_r523705981
- [ ] Also fix CaptureBiome.cs - GetCaptureBiome -> #876
- [ ] Important vanilla methods to check: 
  - [ ] Player.GetPrimaryBiome. Extremely important, biome ids should be assigned automatically here and there should be a ModBiome.type or something that gets what this sets. See ordering above.
  - [ ] GameContent.Personalities.HelperInfo
  - [ ] GameContent.ShopHelper
  - [ ] GameContent.Bestiary.SpawnConditionBestiaryInfoElement (vanilla sets in BestiaryDatabaseNPCsPopulator, maybe add onto NPCSpawnHelper to do this?)

patches/tModLoader/Terraria/ModLoader/AtmosphericLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+			}
+
+			if (!(shortList.Count == 0)) {
+				result.anyActive = true;
+				int weight = 0;
+
+				shortList.Sort((s1, s2) => s1.weight.CompareTo(s2.weight));
+
+				foreach (var atmos in shortList) {
+					if (atmos.weight > weight) {
```
Solxanich, line 106: I need to remove this. It may cause issues and no benefit to having this check.
Thanks for pointing out Agrair

patches/tModLoader/Terraria/ModLoader/AtmosphericLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+						weight = atmos.weight;
+
+						result.priority = atmos.type.Priority;
```
Solxanich, line 109: Agrair found bug:
biome: priority 1: music X1, background X2
event: priority 4: background X3.
Result: priority 4: background X3, music X1
Which means that either would have to wrap vanilla and deviate from how ModSystem did the ChooseMusic originally OR assume all derived classes have music. Notably, this only applies to music selection against vanilla today.
Given that the priority result is just cached, no reason we couldn't set it to match music at this point in time?

patches/tModLoader/Terraria/ModLoader/AtmosphericLoader.cs
Agrair: As a follow up, I think I fixed it, but it would be good if someone could [input on it](https://github.com/tModLoader/tModLoader/pull/1205/files#diff-9a1f12f29160d4c2b7c6a77b3ff9a11217e584089609fb35582ec0778bccfae3R113-R144)

patches/tModLoader/Terraria/ModLoader/AtmosphericLoader.cs
Agrair: As a hotfix and possibly a permanent change, I did just [that](https://github.com/tModLoader/tModLoader/pull/1205/files#diff-9a1f12f29160d4c2b7c6a77b3ff9a11217e584089609fb35582ec0778bccfae3R140). However, we should leave this thread open as it is liable to change.

ExampleMod/Content/Biomes/ExampleBiome.cs
```diff
@@ -0,0 +1,51 @@
...
+		public override CaptureBiome.TileColorStyle tileColorStyle => CaptureBiome.TileColorStyle.Mushroom;
+		public override int Music => base.Music;
+
+		public override bool IsBiomeActive(Player player) {
+			return Math.Abs(player.position.X - Main.maxTilesX / 2) < 40;
+		}
+
+		public override void ModifyShopPrices(HelperInfo helperInfo, ShopHelper shopHelperInstance) {
+			switch (helperInfo.npc.type) {
+				case 17:
```
Mirsario, line 25: NPCID please.

ExampleMod/Content/Biomes/ExampleBiome.cs
```diff
@@ -0,0 +1,51 @@
...
+					shopHelperInstance.DislikeBiome(helperInfo.PrimaryPlayerBiome);
+					break;
+				case 209:
+					shopHelperInstance.LoveBiome(helperInfo.PrimaryPlayerBiome);
+					break;
+			}
+		}
+	}
+
+	public class ExampleUndergroundBiome : ModBiome
```
Mirsario, line 35: No reason for both to be in the same file. There's a separate folder for biomes after all.

patches/tModLoader/Terraria/ModLoader/ModAVFX.cs
```diff
@@ -0,0 +1,46 @@
...
+	{
+		// AVFX properties
+		public virtual ModWaterStyle WaterStyle => null;
+		public virtual ModSurfaceBgStyle SurfaceBackgroundStyle => null;
+		public virtual ModUgBgStyle UndergroundBackgroundStyle => null;
+		public virtual int Music => -1;
+		public virtual AVFXPriority Priority => AVFXPriority.None;
+
+		// Capture Camera property
+		public virtual CaptureBiome.TileColorStyle tileColorStyle => CaptureBiome.TileColorStyle.Normal;
```
Mirsario, line 19: What's with the lower case?

patches/tModLoader/Terraria/ModLoader/AtmosphericLoader.cs
Solxanich: Fairly clear what it did, no worries. Looks good

patches/tModLoader/Terraria/ModLoader/AVFXLoader.cs
```diff
@@ -0,0 +1,161 @@
...
+
+			if (shortList.Count == 0)
+				return;
+
+			result.anyActive = true;
+
+			shortList.Sort(AtmosWeight.InvertedCompare);
+			int avfxFields = 0;
+
+			for (int i = 0; avfxFields < 7 && i < shortList.Count; i++) {
```
Solxanich, line 116: like the idea behind picking from the top in the shortList and exit early, but need to think about that surfaceBG and undergroundBG are in general mutually exclusive and may never be both set

Solxanich:
ModBiome PR is now tested and Complete.
- re-adds Mod Surface Backgrounds
- Fixes autoloading backgrounds from a subdirectory
- Fixes CaptureCamera for ModBiomes
- Adds and Cleans-up prioritization of backgrounds, water-styles to match music, including better class management of those loaders to accomplish this
- Adds bestiary support for Biomes
- Adds a ModBiome singleton to decrease god-classing in ModPlayer
- Ports ExampleBiomes to 1.4

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -0,0 +1,90 @@
...
+		/// This is where you can set values for DisplayName.
+		/// </summary>
+		public virtual void SetStaticDefaults() {
+		}
+
+		/// <summary>
+		/// Return true if the player is in the biome.
+		/// </summary>
+		/// <returns></returns>
+		public virtual bool IsBiomeActive(Player player) => false;
```
Chicken-Bones, line 65: should be distinguished from `IsActive` more easily. Perhaps this could be protected, and have an internal accessor for tML only? Something like `CalculateIsPlayerInBiome`?

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -98,5 +99,29 @@ public partial class Player {
...
+
+		/// <summary> 
+		/// Determines if the player is in specified ModBiome. This will throw exceptions on failure. 
+		/// </summary>
+		/// <exception cref="IndexOutOfRangeException"/>
+		/// <exception cref="NullReferenceException"/>
+		public bool InModBiome(ModBiome baseInstance)=> modBiomeFlags[baseInstance.ZeroIndexType];
+
+		public bool zonePurity = false;
+		public bool InZonePurity() {
```
Chicken-Bones, line 119: needs a summary at least. Maybe it should be `NoBiomesActive`?

Chicken-Bones (commented):
Looks quite good!

Extra question: How are biomes used in NPC loot tables? Can we add an example?

ExampleMod/Content/Biomes/ExampleSurfaceBiome.cs
```diff
@@ -0,0 +1,39 @@
...
+		public override string BackgroundPath => base.BackgroundPath;
+		public override Color? BackgroundColor => base.BackgroundColor;
+
+		public override void SetStaticDefaults() {
+			DisplayName.SetDefault("Example Surface");
+		}
+
+		public override bool IsBiomeActive(Player player) {
+			bool b1 = ModContent.GetInstance<ExampleBiomeTileCount>().exampleBlockCount >= 40;
+			bool b2 = Math.Abs(player.position.ToTileCoordinates().X - Main.maxTilesX / 2) < 40;
```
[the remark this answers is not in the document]
Solxanich: Changed to middle third of the world to give more room. 
Should give plenty of space; and in terms of if 40 was enough, it was roughly one screen length either direction of spawn, but doesn't hurt to make it bigger.

patches/tModLoader/Terraria/ModLoader/ModAVFX.cs
```diff
@@ -0,0 +1,48 @@
...
+		/// Does NOT cache the return type
+		/// </summary>
+		internal void RegisterAVFX(ModAVFX modAVFX) {
+			Loaders.AVFXs.Register(this);
+		}
+
+		/// <summary>
+		/// Returns the strength to which the AVFXEffects are active. 0% < val < 200%. Defaults to 100% (ie 100) 
+		/// </summary>
+		public virtual byte GetWeight(Player player) => 100;
```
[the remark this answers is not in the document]
Solxanich: I've added much more documentation to hopefully make it clear what it does.
Not sure why would want to return int, when GetCorrWeight is going to bound it anyways between 0 and 200, so any additional bytes are waste.

Solxanich:
> How are biomes used in NPC loot tables? Can we add an example

I have zero idea. If we want to look in to that, wouldn't that be better to do as a separate PR afterwards for an additional feature/investigation? As is, this contains several fixes for broken 1.3 features such as capture camera & surface backgrounds, so'd prefer not to defer for independently patchable enhancements.
But, can look in to if its a need under this PR.

patches/tModLoader/Terraria/ModLoader/ModAVFX.cs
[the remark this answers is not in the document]
Chicken-Bones: `int` is the preferred way to express 'this is just a number'. `byte` should only be used when the value is being stored or transmitted, in a place where the memory cost matters. It provides more to think about, is slightly harder to refactor, and makes general math harder (eg averaging). 

Similarly, it's accepted as best practice to use `int` in for loops or array indexers instead of `uint` even though the value cannot go below 0.

Chicken-Bones:
> > How are biomes used in NPC loot tables? Can we add an example
> 
> I have zero idea, haven't changed anything afaik related to it. If we want to look in to that, wouldn't that be better to do as a separate PR afterwards for an additional feature/investigation? As is, this contains several fixes for broken 1.3 features such as capture camera & surface backgrounds and feels feature bloated as a PR, so'd prefer to defer for independently patchable enhancements.
> But, can look in to if its a need under this PR.

Happy for that to be a followup PR. I'm pretty happy with getting this merged to get modder feedback and testing.

PR #1210, opened 2020-11-16
NPCLoader.CanChat for Smart Interaction (1.4)

Steviegt6:
Added back NPCLoader.CanChat for smart interaction.

See `PartyZombie.cs`.

### ExampleMod updates
Added an example to `PartyZombie`.


Rartrin (approved):

Steviegt6:
HoverNPC (Hover) still needs to be cleaned up and documented.

ExampleMod/Content/NPCs/FallenSoul.cs
```diff
@@ -0,0 +1,137 @@
...
+
+				if (distance > 100) {
+					// Make the NPC fade out the farther away the NPC is.
+					npc.alpha += (int)(155 * ((distance - 100) / 150));
+				}
+
+				return;
+			}
+
+			npc.alpha = 255;
```
Chicken-Bones, line 115: same as above, just use `100 + Utils.SmoothStep((250, 100, distance) * 155`

Seriously, maybe tML should have a `Utils.Lerp(float x1, float y1, float x2, float y2, float x)`

Also yeah, lets not use `Utils.SmoothStep` because it's definition is really confusing when there's also `MathHelper.SmoothStep` which does something completely different

ExampleMod/Content/NPCs/FallenSoul.cs
```diff
@@ -0,0 +1,137 @@
...
+		// Allows us to do something when the NPC is hit. Here, we spawn dust.
+		public override void HitEffect(int hitDirection, double damage) {
+			if (npc.life > 0) {
+				for (int i = 0; i < damage / npc.lifeMax * 100; i++) {
+					Dust dust = Dust.NewDustDirect(npc.position, npc.width, npc.height, 192, hitDirection, -1f, 100, new Color(100, 100, 100, 100), 1f);
+					dust.noGravity = true;
+				}
+
+				return;
+			}
```
Chicken-Bones, line 46: needs more comments. I'd use an `if/else` here, but also, the only difference betewen the two loops is the number of dust, and the SpeedY, so you'd be better to merge them with some local variables declared at the top.

ExampleMod/Common/GlobalProjectiles/ExampleGlobalProjectile.cs
```diff
@@ -0,0 +1,24 @@
...
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.GlobalProjectiles
+{
+	// This file shows an example of a GlobalProjectile class. GlobalProjectile hooks are called on all projectiles in the game and are suitable for sweeping
+	// changes like adding additional data to all projectiles in the game. Here we simply override PostAI in order to make Purification Power loop through the
+	// npc array to affect Wraiths specifically, transforming them into Fallen Souls, as it is simple to understand.
+	public class ExampleGlobalProjectile : GlobalProjectile
```
Chicken-Bones, line 11: this should probably be renamed to `WraithPurificationGlobalProjectile` and put in the same folder as the `ExampleFallenSoul` NPC.

I'm not sure what we decided with ExampleMod, but we wanted more self-contained content examples in their own subfolders. If we just want to showcase how to use GlobalProjectile, we should use a more standalone example.

patches/tModLoader/Terraria/GameContent/ObjectInteractions/NPCSmartInteractCandidateProvider.cs.patch
```diff
@@ -0,0 +1,17 @@
...
++using Terraria.ModLoader;
+ 
+ namespace Terraria.GameContent.ObjectInteractions
+ {
+@@ -43,7 +_,7 @@
+ 			float npcDistanceFromCursor = -1f;
+ 			for (int i = 0; i < 200; i++) {
+ 				NPC nPC = Main.npc[i];
+-				if (nPC.active && nPC.townNPC && nPC.Hitbox.Intersects(value) && !flag) {
++				if (nPC.active && NPCLoader.CanChat(nPC, nPC.townNPC) && nPC.Hitbox.Intersects(value) && !flag) {
```
Chicken-Bones, line 14: The design of this hook is inconsistent with other hooks which return nullables. We should make `NPCLoader.CanChat` return `bool?` and right here just do `(NPCLoader.CanChat(nPC) ?? nPC.townNPC)`

ExampleMod/Content/NPCs/FallenSoul.cs
```diff
@@ -0,0 +1,137 @@
...
+
+			for (int i = 0; i < 50; i++) {
+				Dust dust = Dust.NewDustDirect(npc.position, npc.width, npc.height, 192, 2 * hitDirection, -2f, 100, new Color(100, 100, 100, 100), 1f);
+				dust.noGravity = true;
+			}
+		}
+
+		// Allows the NPC to talk with the player, even if it isn't a town NPC.
+		public override bool CanChat() => true;
```
Chicken-Bones, line 55: Should this be a property? @Mirsario ?

Chicken-Bones (changes requested):
Good stuff, just a few minor discussion points

PR #1211, opened 2020-11-16
Allow abnormal world sizes + various other world gen improvements

kittenchilly:
1. Gives the ability for abnormal worlds (aka large than large worlds or smaller than small worlds) to be loaded. The method I'm using to do this should also improve performance on normal worlds as well.
2. Fixes many crashing issues if you try to generate a world that isn't one of the main three sizes.
3. Adds SetWorldGenDefaults that comes even before PreWorldGen to allow you to set world size or anything else yourself.
4. Makes literally everything in WorldGen.cs public. All of these variables that were previously private or local (like dungeonSide) are useful for anyone working with world gen.
5. Adds ModifyMeteorTasks and ModifyAltarTasks. They work the same way as ModifyWorldGenTasks or ModifyHardmodeTasks but with the specified stuff they do to the world.

This is a pretty radical pull request and it changes a lot about world gen and how terraria handle's the world in general. I would appreciate if there was extensive testing on this part of tModLoader due to this. All of this is 100% vanilla compatible.


patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -68,7 +437,220 @@
...
+ 
+@@ -2163,101 +_,110 @@
+ 				}
+ 			}
+ 
++			meteorTasks.Add(new PassLegacy("Meteor Spawn", (progress, config) => {
++				num = genRand.Next(17, 23);
++				for (int num2 = i - num; num2 < i + num; num2++) {
++					for (int num3 = j - num; num3 < j + num; num3++) {
++						if (num3 <= j + Main.rand.Next(-2, 3) - 5)
```
Mirsario, line 457: This is a terrible patch. Split the method into 3. Use the middle one as a delegate.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -170,6 +753,49 @@
...
+-					float num = (float)j / (float)lastMaxTilesX;
+-					Main.statusText = Lang.gen[46].Value + " " + (int)(num * 100f + 1f) + "%";
+-					for (int k = 0; k < lastMaxTilesY; k++) {
+-						Main.tile[j, k] = null;
+-					}
+-				}
+-			}
++			//Ok, so with this new system this is just not needed anymore. This is pretty radical to get rid of, so please look for bugs related to this closely.
++			//Everything should still be vanilla compatible.
++			//if (Main.netMode == 1 || lastMaxTilesX > Main.maxTilesX || lastMaxTilesY > Main.maxTilesY) {
```
Mirsario, line 787: Use comment blocks.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -840,6 +1910,278 @@
...
+ 			}
+ 
++			altarTasks.Add(new PassLegacy("Altar Decide Ore", (progress, config) => {
+-			switch (num) {
++				switch (num) {
+-				case 0: {
++					case 0: {
+-						if (SavedOreTiers.Cobalt == -1) {
++							if (SavedOreTiers.Cobalt == -1) {
+-							flag = true;
```
Mirsario, line 1932: Again, a horrendous patch, and many more like that below. Split the methods.

Mirsario (changes requested):
The amount of different patches does not feel nice at all.

patches/tModLoader/Terraria/WorldGen.cs.patch
TheStachelfisch: That shouldn't be a problem, using // is actually preferred since it's "safer"
Here would be a bit more detailed answer https://stackoverflow.com/questions/1745179/c-sharp-coding-style-comments

Mirsario: I asked for comment blocks because this is a patch. What you linked has nothing to do with diffs.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1475,7 +2672,7 @@
 +			return result;
 +		}
 -		public static int KillTile_GetTileDustAmount(bool fail, Tile tileCache) {
-+		private static int KillTile_GetTileDustAmount_Vanilla(bool fail, Tile tileCache) {
++		public static int KillTile_GetTileDustAmount_Vanilla(bool fail, Tile tileCache) {
```
Mirsario, line 2675: No need for making a split public.

kittenchilly: I just regexed all of this. I can go through and modify stuff that doesn't necessarily need to be public.

PR #1212, opened 2020-11-16
Added armor penetration as a statistic for items and projectiles.

ThomasThePencil:
Armor penetration is now saved on items and projectiles as a statistic that can be modified by any given effect, similar to how it's handled for the player. This greatly eases the process of granting armor penetration to projectiles and items, a feature that's now given precedent in vanilla.


Chicken-Bones:
Looks pretty good, I'll have to double check the patches on individual items, but nice and clean

Mirsario:
@Chicken-Bones I think we're better off renaming DamageModifier and using it for every stat value that ends up being an int. Like the one added here. Lines like `stat = (int)(stat * 204.394093848f)` don't stack well.

PR #1218, opened 2020-11-22
Enable autoloading on ModLoaderMod for default content

Chicken-Bones:
Also restores the dev/patreon sets.

This cleans up the code fairly well in quite a few places. Haven't tested past loading into the world, due to lack of cheat sheet.

The texture paths are annoying because we have to use '.' instead of '/' with `AssemblyResourcesContentSource`

`DamageClass` static instance properties are now getters for `ModContent.GetInstance`. The JIT should compile it down to just a direct fetch from the correct address, same as a static variable access. Pattern should work the same as any other damage class.

Had to make every ModX class abstract (they should be anyway), and add `SimpleModGore`

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
+++ b/patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
@@ -4,26 +4,26 @@ namespace Terraria.ModLoader
 {
 	public class Melee : DamageClass
 	{
-		internal protected override string DisplayNameInternal => Language.GetTextValue("LegacyTooltip.2");
+		internal protected override string DisplayNameInternal => " "+Language.GetTextValue("LegacyTooltip.2");
 	}
 
 	public class Ranged : DamageClass
 	{
-		internal protected override string DisplayNameInternal => Language.GetTextValue("LegacyTooltip.3");
+		internal protected override string DisplayNameInternal => " " + Language.GetTextValue("LegacyTooltip.3");
 	}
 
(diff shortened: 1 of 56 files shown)


Jofairden (approved):
Nice, I approve of this

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -4,26 +4,26 @@ namespace Terraria.ModLoader
 {
 	public class Melee : DamageClass
 	{
-		internal protected override string DisplayNameInternal => Language.GetTextValue("LegacyTooltip.2");
+		internal protected override string DisplayNameInternal => " "+Language.GetTextValue("LegacyTooltip.2");
```
Mirsario, line 7: These locals already have spaces in them for some reason. It would actually make more sense to remove them here, to not force modders to add them.
image

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
Chicken-Bones: I totally misread the `$"{(classIsVanilla ? string.Empty : " ")` logic. That's exactly what the default logic does, gives modders a space. Lets restore that and instead do `.Substring(0)` inside the vanilla `Language.GetTextValue`

That is, assuming all languages have that translation start with a space

Mirsario (approved):

PR #1220, opened 2020-11-23
[1.4] ModSystems

Mirsario:
### What is the new feature?
This PR introduces ModSystems. An autoloadable ModType class filled with hooks that previously existed on the Mod class. The purpose of this is to allow for proper object-oriented usage of those hooks, as there can only be 1 Mod instance in a mod, and there can be an infinite amount of ModSystems. 

**Update:** Now also consumes and replaces ModWorld, because it wasn't tied to any world, and was basically the same thing.

Update and similar hooks being on Mod is anti-OOP and encourages bloated god Mod classes. This fixes that.

~~(!) I wouldn't mind hearing what people think of merging ModWorld into ModSystems too. In the context of, for example, Overhaul's FireSystem, we would need ModSystems' update hooks, as well as ModWorld's save & load, and PostDrawTiles. The same would be true for Decal (blood on tiles) and TileOverlay (snow and leaves on tiles) systems.~~
Implemented.

See https://github.com/Mirsario/TerrariaOverhaul/tree/1.4/Common/Systems

### ExampleMod updates
There aren't any yet, besides the generation one that was previously a ModWorld.
I need ideas for what can be a good system someone would need in a basic content mod.


patches/tModLoader/Terraria/ModLoader/Mod.Hooks.cs
```diff
@@ -114,158 +99,5 @@ partial class Mod
 		public virtual bool HijackSendData(int whoAmI, int msgType, int remoteClient, int ignoreClient, NetworkText text, int number, float number2, float number3, float number4, int number5, int number6, int number7) {
```
Chicken-Bones, line 99: move this too!

Chicken-Bones (approved):

PR #1221, opened 2020-11-24
[1.4] Projectile Events

JasperDawg:
Hi, I'm extremely new to working with others and PRs and stuff. Don't be hard on me.
### What is the new feature?
I've added events for the purpose of reducing boilerplate code associated with projectile weapons and accessories.
The current solution for similar stuff is to manually spawn and flag projectiles to do something as simple incrementing a value on an item or on hit effects in your GlobalItem. To avoid God Classing and headaches with organization, I've come up with the solution of using events in order to attach code to any projectile. 

Mirsario had recommended hooks with basically identical features to this.

I have a gun. If I want every new ammo to increase the damage on this gun by one temporarily when it hits an enemy, I can simply subscribe to an event that increments a value on the gun's end by one when it hits an NPC. Previously this would require a lot of boilerplate - spawning a new projectile, getting it's ID, getting it's projectile instance from Main.projectile, and then flagging it. I'd then go to my general purpose GlobalProjectile to gather the player's helditem, which I can't reliably ensure is going to be the item I want it to when the projectile hits, cast it to the type I want, and then increment the value.

With this, I can simply make a method that increments said value and subscribe it to the event where I'd want it to go.
I also have an accessory. With this accessory, I want every projectile to explode on kill when I wear it. Same deal as the above one, less boilerplate needed. 

### ExampleMod updates
ExampleGlobalEvents - shows off subscribing in a globalitem context.
ExampleEventItem - shows off subscribing in a moditem context.

This commit is feature incomplete and will be updated as needed/requested.

### DIFF
--- a/ExampleMod/Common/GlobalItems/ExampleGlobalEvents.cs
+++ b/ExampleMod/Common/GlobalItems/ExampleGlobalEvents.cs
@@ -24,10 +24,8 @@ namespace ExampleMod.Common.GlobalItems
 			}
 		}
 
-		private void SayHi_OnProjectileKill() {
-		 	Main.NewText("Hi!", Main.DiscoColor);
-		}
+		private void SayHi_OnProjectileKill(Projectile projectile) => Main.NewText("Hi!", Main.DiscoColor);
 		
-		private void AddADebuff_OnProjectileHitNPC(NPC npc) => npc.AddBuff(BuffID.OnFire, 50);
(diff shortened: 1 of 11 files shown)


SpencerDawg:
Will write code to sync it across everything eventually.

Chicken-Bones:
I actually feel strangely good about this PR, Would it work in multiplayer though....

PR #1222, opened 2020-11-24
MenuButton basics

Steviegt6:
(Draft)

### What is the new feature?
`ModifyMenuButtons` (a hook in `Mod`) and`MenuButton`, which works similarly to `TooltipLine`.

This should be part of tML as it allows modders to add more to menus without the use of IL editing or other more hacky solutions. (TODO: ModMenuModes)

To achieve the same thing? Not that I can think of.

TODO

### ExampleMod updates
N/A (Potential update to come if I can think of anything)

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2841,14 +2841,33 @@
  
  			if (menuMode == 888) {
  				if (!_blockFancyUIWhileLoading)
-@@ -38839,6 +_,7 @@
+@@ -38838,7 +_,9 @@
+ 			}
  
  			bool flag11 = false;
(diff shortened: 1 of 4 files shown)


Steviegt6:
I need a better way to name vanilla menu buttons. The current way is temporary.

Jofairden:
Appreciate the concept, the feature however warrants discussion.
As we've discussed before, all the mods wanting to modify the menu is waiting for problems, as it's not really multi-mod compatible. (If you can move stuff around, delete buttons etc.)

One concept we talked about was _mod specific_ menus. So imagine being able to click an arrow or something, and you can traverse through menus. This way, mods add their entire own separate menu.

@Chicken-Bones  @JavidPack @Mirsario  Need input.

Mirsario:
My input in discord was just that I think that adding main menu items should remain a pain. We already have the menu clipping below limits of 800x600 windows.

PR #1224, opened 2020-11-25
[1.4] ExampleMod Commit showcasing projectile modification.

SpencerDawg:
Adds some stuff showcasing how to modify shot projectiles. Related to the issue I recently made.
Probably not, but add some if you need to.


Chicken-Bones:
Fantastic. Love it. The 'every third hit on the same enemy' one really hurts me.

PR #1226, opened 2020-11-28
Converted Modifier's & overload to method

Rartrin:
Converted Modifier's `&` overload to `CombineWith` method.

Also cleaned up the `Equals` method.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Modifier.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Modifier.cs
@@ -12,6 +12,9 @@
 			this.multiplicative = multiplicative;
 		}
 
+		public Modifier CombineWith(Modifier that) =>
+			new Modifier(additive + that.additive - 1, multiplicative * that.multiplicative);
+
 		public override bool Equals(object obj) {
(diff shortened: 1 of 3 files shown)


Chicken-Bones (changes requested):
*`CombinedWith` to make it clear it's not in place.

Rartrin:
> *`CombinedWith` to make it clear it's not in place.

"Combined" is past tense though.
I could make the method `readonly`.

Otherwise I can just make a static `Combine` method instead. Actually, that will probably work out the best when it comes time to add in crit and knockback, because then the methods will be `CombineDamages`, `CombineCrits`, `CombineKnockbacks`.

Chicken-Bones:
I believe the convention is to use past tense, as in "the result when this is combined with that" or at least, it clarifies.

I'm also happy with `Modifier.Combine(m1, m2)` but it can't be chained as easily. (unlike the `&` syntax).

`CritModifier` and `KnockbackModifier` should be separate classes if they have separate combine calculations. They shouldn't be assignable to each other. That's a type error.

Chicken-Bones:
Completed in the `StatModifier` PR

Issue #1219, opened 2020-11-22
Selling back items to NPCs multiplies price by 5


Itorius:
See https://cdn.discordapp.com/attachments/445276626352209920/780078250167894046/fb0EUw9OR9.mp4

Itorius:
Fixed in https://github.com/tModLoader/tModLoader/commit/124fa19301e7d7f134e31b667beac14b19dbec38

PR #1228, opened 2020-11-28
[1.4] Fix selling back items to NPCs multiplies price by 5 issue

TheVoxcraft:
Issue #1219 

### What is the bug?
Selling back items to NPCs multiplies price by 5. Pointed out by Itorius.
See https://cdn.discordapp.com/attachments/445276626352209920/780078250167894046/fb0EUw9OR9.mp4
(Itorius)

### How did you fix the bug?
I removed a line of code in _ItemSlot.cs_ where the value of an item was multiplied by 5 when sold.
This made a visual bug where the item after being sold and bought once had 1/5 the price in inventory (still sold for the correct price). This might be what the line from _ItemSlot.cs_ was trying to fix. 
I corrected this visual bug in _Main.cs_ by multiplying the item by 5 in the tooltip.

There must be a better way to fix this bug without this extra line of code. But this seems to work good.


Itorius:
Thanks for your contribution but I implemented a proper fix in https://github.com/tModLoader/tModLoader/commit/124fa19301e7d7f134e31b667beac14b19dbec38.

PR #1231, opened 2020-12-03
Melee CanHitNPC

Agrair:
Added back the CanHitNPC hooks. I also removed some redundant TODO entries. At least I'm pretty confident they were redundant. The giant patch blob looks to me like it's outdated anyway: those tML functions that got inserted are now called properly in a separate method that vanilla made.


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3034,6 +3034,25 @@
...
++
++						modCanHit = ItemLoader.CanHitNPC(sItem, this, Main.npc[i]);
++						if (modCanHit.HasValue && !modCanHit.Value)
++							continue;
++
++						modCanHit = PlayerHooks.CanHitNPC(this, sItem, Main.npc[i]);
++						if (modCanHit.HasValue && !modCanHit.Value)
++							continue;
++
+ 						if (itemRectangle.Intersects(value) && (Main.npc[i].noTileCollide || CanHit(Main.npc[i]))) {
```
Mirsario, line 3053: Good, but this needs some re-imagining.
First of all, all these hooks have nullable returns, but nothing is different between returning null and true. True would usually skip vanilla checks, and null would run them. So we need that, and perhaps the calls should also be pulled up in the loop. I'm not sure if they should control the intersection check, though.

Second -- can you move these 3 calls into a CombinedHooks method? Something like 'CanPlayerHitNPCWithItem'. The order of calls should probably be Item -> Player -> NPC.

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -276,10 +276,10 @@ public abstract class GlobalItem : ModType
 		}
 
 		/// <summary>
-		/// Allows you to determine whether a melee weapon can hit the given NPC when swung. Return true to allow hitting the target, return false to block the weapon from hitting the target, and return null to use the vanilla code for whether the target can be hit. Returns null by default.
+		/// Allows you to determine whether a melee weapon can hit the given NPC when swung. Return true to allow hitting the target, return false to block the weapon from hitting the target. Returns true by default.
 		/// </summary>
-		public virtual bool? CanHitNPC(Item item, Player player, NPC target) {
-			return null;
+		public virtual bool CanHitNPC(Item item, Player player, NPC target) {
+			return true;
```
Mirsario, line 282: Why is this?

Agrair: Why should they not be normal booleans? If all of them return true, nothing happens. If one of them returns false, the NPC can't be hit.

Agrair: I'm actually not sure how to do this. In vanilla, the first things that are checked are things like if the player has the no critter harm item, if the npc is friendly, if it's the clothier, if it's invincible. Not if it is colliding. So I think I would have to use a goto  to skip the ifs? If null is returned, let the vanilla checks run, if false is returned, just skip over the npc, if true is returned, skip immediately to the collision check. Does that sound right?

Jofairden: Isn't that how it was in the first place? Null is usually so you can tell to let vanilla do it's thing

PR #1236, opened 2020-12-10
Add MusicID's from 42-89

RighteousRyan1:
It will be necessary for modders who want to access music easily.
Probably, if naming needs changing.
`ModMenu.Music =>MusicID.Credits`
`ModMenu.Music => MusicID.OtherworldlyDay`


RighteousRyan1:
Changed the `MusicID` for `Hell` to `Underworld` so the naming stayed the same overall.

Steviegt6:
Shouldn't these added IDs go in a partial class?

Mirsario:
Wrong target branch, and cannot be changed. Stop using web tools.

PR #1245, opened 2020-12-23
Fix multiplayer crash by Bestiary PR

Chik3r:
### What is the bug?
On `1.4_mergedtesting`, starting a multiplayer server makes the server crash with the following error: https://paste.mod.gg/jidabapewa.css
If we follow the stacktrace, we can find that it's crashing because the `SetText` method of `UITextPanel` tries to access `FontAssets.DeathText.Value` and `FontAssets.MouseText.Value`, but because it's a server they are null since they aren't loaded.

### How did you fix the bug?
I changed:
```cs
Main.BestiaryUI = new UIBestiaryTest(Main.BestiaryDB);
```
to:
```cs
if (!Main.dedServ)
	Main.BestiaryUI = new UIBestiaryTest(Main.BestiaryDB);
```

maybe

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModContent.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModContent.cs
@@ -460,7 +460,8 @@ namespace Terraria.ModLoader
 			bestiaryDatabase.Merge(Main.ItemDropsDB);
 			
 			//Etc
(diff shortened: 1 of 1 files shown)


Mirsario:
I commited this on `1.4` with 90e302d. mergedtesting is a throwaway branch, it is never merged into anything.

PR #1246, opened 2020-12-23
 Update zh-Hans translation

sgkoishi:
Change List:
* Translate `ModDependencyClickTooltipMore`, `CloudWarning` and `VanillaGOGNotFound`
* Use half-width colon `:` instead of full-width `：`.
The font in-game does not display `：`(and also `，。`) as a full-width character and makes the alignment confusing.
This doesn't apply to the `MessageBox` or Console (Server) part that uses system font.
* Add space after comma and period.
So it looks closer to full-width.


Jofairden:
Thanks!

PR #1250, opened 2020-12-26
Add call to "GlobalBgStyle.ChooseSurfaceBgStyle"

Chik3r:
### What is the bug?
`GlobalBgStyle.ChooseSurfaceBgStyle` is never called so whatever mods do when overriding it doesn't happen.

### How did you fix the bug?
By adding a call to `SurfaceBgStyleLoader.ChooseStyle()`, which calls the `ChooseSurfaceBgStyle` hook of all mods.
In 1.3, this method is called in the `DrawBG` method of `Main`, after a bunch of nested ternary operators. In 1.4, that line was moved to a method called `GetPreferredBGStyleForPlayer`, so instead of calling `ChooseStyle` inside `DrawBG`, I'm calling it inside the new method right before the `return num;`.
1.3 code: 
image
1.4 code: DrawBG:
image
GetPreferredBGStyleForPlayer:
image

Another option would be to call `SurfaceBgStyleLoader.ChooseStyle()` after `GetPreferredBGStyleForPlayer` in `DrawBG` instead of inside `GetPreferredBGStyleForPlayer`.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -3330,7 +3330,11 @@
  			try {
  				DrawSurfaceBG();
  				if (BackgroundEnabled)
-@@ -48903,6 +_,7 @@
+@@ -48900,9 +_,11 @@
+ 						num = 11;
+ 					else if (WorldGen.treeBG1 != WorldGen.treeBG4)
+ 						num = 12;
++					//Patch context.
  				}
  			}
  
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3330,6 +3330,18 @@
...
+ 						num = 11;
+ 					else if (WorldGen.treeBG1 != WorldGen.treeBG4)
+ 						num = 12;
++					//Patch context.
+ 				}
+ 			}
+ 
++			SurfaceBgStyleLoader.ChooseStyle(ref num);
+ 			return num;
+ 		}
```
Mirsario, line 3343: There's sometimes a need to make the differ add extra patch context, because the area I selected isn't unique enough for this patch to not break in the future in something as big as Main.cs.

PR #1251, opened 2020-12-27
Add call to "GlobalNPC.PostDraw"

Chik3r:
### What is the bug?
`GlobalNPC.PostDraw` is only called when `PreDraw` returns false

### How did you fix the bug?
By adding a call to `NPCLoader.PostDraw` inside the `DrawNPCDirect` method. The call was added in the same position as it was in 1.3

no

Also, can whoever accepts this merge this PR and #1250 to `1.4_mergedtesting` so that I can finish making `ModdersToolkit` work for that branch.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2059,18 +2059,15 @@
  			NPC rCurrentNPC = npc[iNPCIndex];
  			Vector2 screenPos = screenPosition;
  			DrawNPCDirect(spriteBatch, rCurrentNPC, behindTiles, screenPos);
-@@ -16791,6 +_,20 @@
+@@ -16791,6 +_,17 @@
  				}
  			}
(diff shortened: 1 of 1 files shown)


Mirsario (changes requested):
There seem to be a lot of returns and gotos in-between the PreDraw and that PostDraw you readded, and the location of PreDraw seems random to me. The method those are in may be better off just being split into 2 and wrapped around with PreDraw and PostDraw, like this:

image

Chicken-Bones:
Excellent job adding patch context, but Mirsario is right, to capture PostDraw on all NPCs with all those returns you'll need to wrap the method

Chik3r:
So I did what you said, but I still left some code above the wrap, since the code for `PreDraw` uses `npcColor` which is calculated above, and I didn't want to move 50 lines of code somewhere else.
image

Mirsario (approved):

PR #1254, opened 2020-12-28
[1.4] Fix crash when opening chests in mutliplayer

staticfox:
### What is the bug?
The client crashes when opening a chest in multiplayer.

### How did you fix the bug?
`Main.chest[num261]` was null which seemed like a typo from f33571a. I changed it to 260 then opened a chest successfully.

This project is still kind of magic to me, so I don't know.


direwolf420:
Fix looks fine to me, Main.chest[num260] is the correct thing

Issue #1244, opened 2020-12-23
[1.4] Hovering over memory estimates crashes the game


Chik3r:
### Description
On branch `1.4_mergedtesting`, when hovering over the memory estimates bar, the game crashes:
Video (imgur killed the quality): https://imgur.com/a/nPZjekI

### Log file
[client.log](https://github.com/tModLoader/tModLoader/files/5732877/client.log)

### Steps to reproduce
1. Go to settings.
2. Enable the memory estimates bar.
3. Go to the `Mods` menu.
4. Hover over the memory estimates bar.

### Expected behavior 
There should be no crash and the memory estimates should appear

### Actual behavior
It's crashing

### Reproduction frequency
Everytime I hover it crashes

### Additional Information
Make sure you're on `1.4_mergedtesting`

Chicken-Bones:
Fixed in #1255

PR #1255, opened 2020-12-28
[1.4] Fix crash with memory statistics tooltip

fakelag:
### What is the bug?

Fixes crash when hovering over memory tooltip on the "Mods" page as described in the issue https://github.com/tModLoader/tModLoader/issues/1244

### How did you fix the bug?
Crash seems to be caused by editing render hierarchy during drawing, which raises an exception somewhere down the line. Fix simply removes the part of code that appends `_hoverPanel` into `UIMemoryBar`'s parent element's render tree. This also means that the relative offsetting to `Parent.GetDimensions()` can be removed since `_hoverPanel` is now rendered relative to 0 0. I also changed some of the magic offsets to preserve the original look of the tooltip.

I'm pretty new to this project so I can't say that this is the best or correct way to fix the issue

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIMemoryBar.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIMemoryBar.cs
@@ -82,19 +82,14 @@ namespace Terraria.ModLoader.UI
 			}
 
 			if (drawHover && hoverData != null) {
-				if(_hoverPanel.Parent!=Parent) {
-					_hoverPanel.Remove();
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
@fakelag would you like to come join us at https://discord.com/invite/RMZCqq6? I'll give you access to the collaborators channel.

fakelag:
> @fakelag would you like to come join us at https://discord.com/invite/RMZCqq6? I'll give you access to the collaborators channel.

I'll drop by, fakelag#2360

Issue #389, opened 2018-07-21
Shield of Cthulhu doesnt work properly with CanHitNPC() hooks


SheepishShepherd:
### Description

The Shield of Cthulhu appears to be when applied to CanHitNPC hooks. I have tested it in ModPlayer, GlobalItem and GlobalNPC classes, and all produce the same result.

### Steps to Reproduce

1.) Add CanHitNPC hooks to any or all of the above classes (In my case I tried all three and made it return false if the target was a Bunny)
2.) Load into a world with the Shield of Cthulhu equipped
3.) Dash into a bunny to see if it works, and the first time it does
4.) Dash into the same bunny again and watch it die

### Expected behavior 

Shield of Cthulhu will not hit ANY npcs while the CanHitNPC hooks are false

### Actual behavior

Shield of Cthulhu will not hit an npc ONCE, and then always hit npcs until you load up the world again.

### Reproduction frequency

Happens every time with all CanHitNPC hooks from each class

### Versions

v0.10.1.5

### Mods

No mods needed to reproduce

### Additional Information

For some reason it works 100% properly with modded npcs. I have tested the hooks on my own npcs and the shield of cthulhu never hits them, but will hit normal npcs afterwards.

JavidPack:
Turns out the 1st hit not working is probably a vanilla bug. I just reproduced that on 1.3.5.3.

Chicken-Bones:
We'd need a more general hook for preventing NPC damage, awaiting feedback

Solxanich:
Flagged as Fixed in #1231

JavidPack:
Just tested, not happening anymore. The unusual missing of the 1st attack is still happening, but now all subsequent dashes result in the player colliding and bouncing off but not doing any damage. Not sure if bouncing off should still occur, but that can be another discussion later if it comes up.

PR #1257, opened 2020-12-29
[1.4] Fix ModPlayer CanHitNPC hook not getting called ever

fakelag:
### What is this?
`ModPlayer::CanHitNPC` hook doesn't ever get called from anywhere (there are no calls to `PlayerHooks.CanHitNPC` in the 1.4 code by the looks of it). Add the call into `ItemCheck_MeleeHitNPCs` so that mods can use the hook.

### Considerations
- This implementation adds the call to the very root of `ItemCheck_MeleeHitNPCs` (so it can override all the relevant game logic), which means that the hook will be called very frequently. Do we expect modders to implement heavy logic into this hook?
- This implementation appends the checks into existing if-statements, leading to clutter. Since the change is done via patching, this seems like the way to go. Should refactoring & cleaner code be considered instead?
- This doesn't fix https://github.com/tModLoader/tModLoader/issues/389 , however the hook will now work as stated in the documentation:
```C#
/// <summary>
/// Allows you to determine whether a player can hit the given NPC by swinging a melee weapon. Return true to allow hitting the target, return false to block this player from hitting the target, and return null to use the vanilla code for whether the target can be hit. Returns null by default.
/// </summary>
/// <param name="item"></param>
/// <param name="target"></param>
/// <returns></returns>
public virtual bool? CanHitNPC(Item item, NPC target) {
	return null;
}
```

### Alternatives
There might be alternative places to place the call, but `ItemCheck_MeleeHitNPCs` appeared the most logical place

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3034,6 +3034,27 @@
  				if (Main.netMode != 0)
  					NetMessage.SendPlayerHurt(i, playerDeathReason, num, direction, flag, pvp: true, -1);
  
+@@ -30601,11 +_,17 @@
+ 				if (!Main.npc[i].active || Main.npc[i].immune[whoAmI] != 0 || attackCD != 0)
+ 					continue;
+ 
++				bool? modCanHit = PlayerHooks.CanHitNPC(this, sItem, Main.npc[i]);
++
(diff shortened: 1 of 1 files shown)


Mirsario:
Duplicate of #1231, unfortunately. I'll go fix and merge it right now.

Issue #1248, opened 2020-12-26
[1.4] Leaving the game when modded NPCs still exist in it, then unloading mods, "purple-screens" the game


direwolf420:
### Description
Leaving the game when modded NPCs still exist in it, then unloading mods, "purple-screens" the game

### Log file
```cs
[14:14:19] [14/INFO] [tML]: Unloading: Example Mod
[14:14:19] [14/DEBUG] [tML]: Unloaded 0 hooks, 1 detours and 1 native detours from ExampleMod
[14:14:19] [14/INFO] [tML]: Unloading: Cheat Sheet (for 1.4)
[14:14:19] [14/INFO] [tML]: Unloading: tModLoader
[14:14:19] [Main Thread/WARN] [tML]: Silently Caught Exception: 
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Terraria.NPC.GetBossHeadTextureIndex() in tModLoader\Terraria\NPC.cs:line 2039
   at Terraria.GameContent.UI.BigProgressBar.CommonBossBigProgressBar.ValidateAndCollectNecessaryInfo(BigProgressBarInfo& info) in tModLoader\Terraria\GameContent\UI\BigProgressBar\CommonBossBigProgressBar.cs:line 19
   at Terraria.GameContent.UI.BigProgressBar.BigProgressBarSystem.TryTracking(Int32 npcIndex) in tModLoader\Terraria\GameContent\UI\BigProgressBar\BigProgressBarSystem.cs:line 92
   at Terraria.GameContent.UI.BigProgressBar.BigProgressBarSystem.TryFindingNPCToTrack() in tModLoader\Terraria\GameContent\UI\BigProgressBar\BigProgressBarSystem.cs:line 71
   at Terraria.GameContent.UI.BigProgressBar.BigProgressBarSystem.Update() in tModLoader\Terraria\GameContent\UI\BigProgressBar\BigProgressBarSystem.cs:line 52
   at Terraria.Main.UpdateUIStates(GameTime gameTime) in tModLoader\Terraria\Main.cs:line 12997
```

### Steps to reproduce
Have only ExampleMod, enter world, get "Zombie" (the multicolored one from ExampleMod) to spawn, leave world, go to Mods, and unload ExampleMod, and reload mods.

### Expected behavior 
No purple screen

### Actual behavior
purple screen

### Additional Information
The issue is caused by two things:
1. BigProgressBarSystem.Update() runs regardless of main menu or not
[...]

PR #1260, opened 2020-12-30
[1.4] Fix IOOB on BigBossProgressBar.Update during unload when modded NPCs were alive

direwolf420:
Fixes Issue #1248 by adding a gameMenu check. I've checked code of all bars and they have no menu-reliant logic in them.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -1076,7 +1076,7 @@
 +			PlayerInput.ScrollWheelDeltaForUI = 0; //TODO: Should this be before PlayerInput.ResetInputsOnActiveStateChange()?
  			CreativeMenu.Update(gameTime);
 +
-+			if (gameMenu) //Don't update IBigProgressBars in the main menu to avoid IOOB during unload when modded NPCs were alive --direwolf420
++			if (!gameMenu) //Don't update IBigProgressBars in the main menu to avoid IOOB during unload when modded NPCs were alive --direwolf420
 -			BigBossProgressBar.Update();
 +				BigBossProgressBar.Update();
  		}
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1068,15 +1068,23 @@
...
+@@ -12806,8 +_,16 @@
  			if (InGameUI != null)
  				InGameUI.Update(gameTime);
  
 +			ModHooks.UpdateUI(gameTime);
 +			PlayerInput.ScrollWheelDeltaForUI = 0; //TODO: Should this be before PlayerInput.ResetInputsOnActiveStateChange()?
  			CreativeMenu.Update(gameTime);
++
++			if (gameMenu) //Don't update IBigProgressBars in the main menu to avoid IOOB during unload when modded NPCs were alive --direwolf420
++				goto afterBigBossProgressBarUpdate; //Avoids indentation change
```
fakelag, line 1080: Why avoid changing the indentation in original code? Does it mess with the patching in some way?

direwolf420: In this case not really actually, its more used for larger chunks of code to prevent exposing vanilla code too much (indentation change is picked up by the patcher).

fakelag: Sure, I can understand that. Generally speaking, aren't the original developers in favor of modding in this game?

Chicken-Bones: The primary reason for reducing patch size is for maintainability. When the base game code changes (or is decompiled differently) patches break and become hard to apply if there's lots of changed lines which don't actually have meaningful changes in them.

See <https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide> for some guidelines

Chicken-Bones: Honestly though @direwolf420 as this is a _one line_ indentation change. I'd rather you keep the indentation change. It's about keeping patch size _small_ primarily. I'll merge it after that adjustment

direwolf420: done

fakelag: I think you still need to invert the condition? Or am I missing something?

direwolf420: I did the indentation skip thing after having the correct condition, fixed the condition, then undid the indentation, and forgot to fix the condition again lol

PR #1261, opened 2020-12-30
[1.4] Basic support for the "boss health bars" feature

direwolf420:
### What is the new feature?
Exposes a few things required for mods to make custom "boss health bars" (called `BigProgressBar`s in code).

### Explanation
This PR adds surface-level access to the "boss health bar" feature in 1.4. It makes the `IBigProgressBar` interface public, aswell as adding a way to register custom instances for npcs. 
Here are some notes on how this system works in vanilla:
* The game has a "boss health bar system" using an interface to track them
    * Each NPC has by default a `CommonBossBigProgressBar` that only works if the NPC also has a boss head texture
    * Only if the NPC has special behavior regarding displaying its bar (e.g. multisegment bosses), or it should not display a bar at all (e.g. Lunatic Cultist clone), the game assigns a custom bar
* It scans through all NPCs and picks the closest trackable one
* If one custom bar is found, do its logic and draw instead of the default one
* This means the common bar is opt-out for all bosses (usually have a boss head). There exists a custom `NeverValidProgressBar` bar
    * That can be assigned to an npc so that it never displays one, even if has a boss head. It has been made public for this purpose

To interact with a vanilla feature.

Yes, a custom ModType that implements `IBigProgressBar` for example, that takes care of loading and instantiating them as ModContent (so you don't have to create new objects for AddBar yourself) and can also add more hooks ontop.

Currently the `AddBar` method works with an entirely separate dictionary, not touching the vanilla one, only tracking additions. It supports vanilla NPC IDs aswell, making overrides possible. It prioritizes these in the `TryGetBar` method.

Overall this feature is very basic in its nature, and so is the PR. It doesn't "add" new capabilities to this system, such as controlling on/off status for each bar (ala `GlobalX`), but it does leave room for fully custom drawing (which mods like `YetAnotherBossHealthBar` could utilize). Sadly because there are no control options (vanilla or this PR) it would be hard to define a proper order/priority system for these bars (the last loaded mod could replace all bars with `NothingValid` leaving all other mods' additions/changes obsolete)

```cs
class PartyZombieBigProgressBar : IBigProgressBar {
//implement these
    void Draw(ref BigProgressBarInfo info, SpriteBatch spriteBatch);
    bool ValidateAndCollectNecessaryInfo(ref BigProgressBarInfo info);
}
```

In PartyZombie.SetStaticDefaults:
```cs
Main.BigBossProgressBar.AddBar(npc.type, new PartyZombieBigProgressBar());
```

### ExampleMod updates
`Content/NPCs/PartyZombie.cs` has an example of registering a health bar to a modded NPC. `ExampleMod/Common/GlobalNPCs/BlueJellyfishBigProgressBarGlobalNPC.cs` has one for a vanilla NPC. `Content/BigProgressBars` contains three bars, one for general use. A prime example for showcasing a custom health bar would be a multisegment boss (old ExampleMods Abomination boss).

### DIFF
--- a/ExampleMod/Common/GlobalNPCs/BlueJellyfishBigProgressBarGlobalNPC.cs
+++ b/ExampleMod/Common/GlobalNPCs/BlueJellyfishBigProgressBarGlobalNPC.cs
@@ -1,0 +1,17 @@
+﻿using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.GlobalNPCs
+{
+	//Showcases how to assign something to a vanilla NPC like you would do for a modded NPC using the SetStaticDefaults hook
+	//In this case, setting a big progress bar
+	public class BlueJellyfishBigProgressBarGlobalNPC : GlobalNPC
+	{
+		//SetupContent hook runs once per class, right after all content has loaded in and is functional
+		public override void SetupContent() {
+			//Custom "boss health bar", see the Content/BigProgressBars folder
+			Main.BigBossProgressBar.AddBar(NPCID.BlueJellyfish, new Content.BigProgressBars.BlueJellyfishBigProgressBar());
+		}
+	}
+}
--- a/ExampleMod/Content/BigProgressBars/BigProgressBarTemplate.cs
+++ b/ExampleMod/Content/BigProgressBars/BigProgressBarTemplate.cs
@@ -1,0 +1,47 @@
+﻿using Microsoft.Xna.Framework.Graphics;
+using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.GameContent;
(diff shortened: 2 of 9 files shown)


direwolf420:
Upon further discussions with @Flashkirby (Creator of `YetAnotherBossHealthBar`) I have decided to give the ModType approach a try to eliminate the possibility of adding bars outside of mod loading, and perhaps make way for a `GlobalX` version that will include a setting ingame for mods that add their own versions of bars (YABHB, Calamity), kind of like the `ModMenu` toggle that already exists. If anyone has input im more than welcome.

Mirsario:
Autoloaded ModType bars do sound nicer than having to call Main.BigBossProgressBar.AddBar in a random ILoadable.

direwolf420:
Closed because experimental nature, real PR mentioned above

PR #1263, opened 2021-01-01
Port AnimatedLoom example to 1.4

direwolf420:
Port of the "AnimatedLoom" example from 1.3 to 1.4. The example shows how to replace a vanilla texture and add animation to it, and then how to unload it properly.

With the new `Asset` system the process is a bit more complicated, but it essentially works the same way. It also uses the new `ModType` methods so the code doesn't have to be in ExampleMod.cs anymore.

I documented almost every line in the example so I think it should be enough.


ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
```diff
@@ -0,0 +1,79 @@
...
+
+		//We have to cache the textures we work with so we can reference them easier
+		private static Asset<Texture2D> originalLoom;
+
+		private static Asset<Texture2D> animatedLoom;
+
+		private static bool replacedTexture = false;
+
+		public override void SetupContent() {
+			//This hook is ran once on mod load, after all content has been loaded into the game, to be sure that everything exists
```
Jofairden, line 24:
```suggestion
			//This hook is ran once on mod load, after all content has been loaded into the game, to be sure that everything exists
		public override void SetupContent() {
```

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
```diff
@@ -0,0 +1,79 @@
...
+			if (!Main.dedServ) {
+				//Do not try to do texture stuff on the server!
```
Jofairden, line 26:
```suggestion
			//Do not try to do texture stuff on the server!
			if (!Main.dedServ) {
```

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
```diff
@@ -0,0 +1,79 @@
...
+				int type = TileID.Loom;
+
+				Main.instance.LoadTiles(type); //Loads the tile texture
+				originalLoom = TextureAssets.Tile[type]; //Cache the tile texture so we can restore it later on unload
+
+				animatedLoom = ModContent.GetTexture("ExampleMod/Common/GlobalTiles/AnimatedLoom"); //Fetch our own texture
+				TextureAssets.Tile[type] = animatedLoom; //Replace the vanilla texture with ours
+				replacedTexture = true; //Flag it so we can be sure we replaced it
```
Jofairden, line 34: Seems a bit unneeded? Can't you just make it a Get and check if the texture matches the EM path? (If it is needed at all)

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
```diff
@@ -0,0 +1,79 @@
...
+				//Do not try to do texture stuff on the server!
+				int type = TileID.Loom;
+
+				if (replacedTexture) {
+					//We previously set this to true if our replacement was successful
+
+					Main.tileFrame[type] = 0; //Reset the frame of the loom tile
+					TextureAssets.Tile[type] = originalLoom; //Restore the original texture
+					replacedTexture = false; //Unset the flag
+				}
```
Jofairden, line 65:
```suggestion
					Main.tileFrame[type] = 0; //Reset the frame of the loom tile
					TextureAssets.Tile[type] = originalLoom; //Restore the original texture
```

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
```diff
@@ -0,0 +1,79 @@
...
+
+				//We hold a reference to the now reverted texture, so simply null it for us; calling Dispose() would also dispose it in TextureAssets.Tile!
+				originalLoom = null;
+
+				if (animatedLoom != null) {
+					//Dispose and null our own texture
+					animatedLoom.Dispose();
```
Jofairden, line 72: Isn't nulling it the same as disposing ?

Jofairden (changes requested):

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
direwolf420: sadly an ".rawimg" is appended to the assets Name property, and i don't know if it will stay like that implementation wise. I'd have to also concider the way the name is sanitized (don't know exactly off the top of my head) to then do a match with the path.

ExampleMod/Common/GlobalTiles/AnimatedLoomGlobalTile.cs
direwolf420: Just following the current examplemod code, found in ExampleHookProjectile, ExamplePlayerDrawLayer, etc

ExampleMod/Old/ExampleMod.cs
```diff
@@ -144,12 +140,6 @@ public class ExampleMod : Mod
 		}
 
 		public override void Unload() {
-			// All code below runs only if we're not loading on a server
-			if (!Main.dedServ) {
-				Main.tileFrame[TileID.Loom] = 0; // Reset the frame of the loom tile
-				Main.tileSetsLoaded[TileID.Loom] = false; // Causes the loom tile to reload its vanilla texture
```
Chicken-Bones, line 150: Is there a reason you abandoned this approach? It seems superior to me.

direwolf420: 1.3:
```cs
public void LoadTiles(int i) {
	if (!tileSetsLoaded[i]) {
		tileTexture[i] = OurLoad<Texture2D>("Images" + Path.DirectorySeparatorChar.ToString() + "Tiles_" + i);
		tileSetsLoaded[i] = true;
	}
}```

1.4:
```cs
public void LoadTiles(int i) {
	if (TextureAssets.Tile[i].State == AssetState.NotLoaded)
		Assets.Request<Texture2D>(TextureAssets.Tile[i].Name);
}```

Yes, the 1.3 design is superior, I agree. `tileSetsLoaded` doesn't exist in 1.4 anymore, the new loading is more of a "request", the actual loading happens much earlier in AssetInitializer:
`TextureAssets.Tile[n] = LoadAsset<Texture2D>("Images/Tiles_" + n, AssetRequestMode.DoNotLoad);`

So yeah swapping made the most sense to me at the time of writing the PR because it has to go by the assets name to actually "finalize" the texture used. There's also no way to force an asset to be unloaded so it can enter the NotLoaded state (setter private, some methods internal)

direwolf420: Also the intention of this PR is to make tml devs think about the Asset system more :P

Issue #302, opened 2018-02-25
Delete Unloaded Mod World Data


JavidPack:
Sometimes, a user might want to completely clear mod data from a World from a particular mod.

One reason might be the mod data is preventing the world from loading properly. Another might be that the user no longer wants the unloaded pink tiles left behind in a world when a mod is unloaded.

I propose a menu to be available from the World Select menu, this menu, called "More" or "Tools" or "Advanced", could list many buttons that would do specific things. In terms of the above suggestions, one could be "Delete unloaded data". This tool would delete all ModWorld data that belongs to currently unloaded mods. 

Another tool could be "Delete unloaded tiles", this tool would simply delete pink/unloaded tiles.

A more advanced implementation might allow selecting which mod to delete mod data from.

Jofairden:
Seems useful. We could generify this enough, buttons to remove any type of modded data: items, projectiles, npcs.. anything instanced basically. Unless we can think of arguments against that

Jofairden:
@JavidPack Coming back to this. Would it be useful if modders can specify a 'fallback' type that the tiles will be replaced with if deleted? It can be stored in MysteryTile data.

Chicken-Bones:
I'd really like to do a more thorough investigation on this. Leftover data should never corrupt worlds. Maybe people want to delete all the unloaded blocks/items etc, which is a useful feature.

I suppose if a mod updates with poor backwards compatibility, you might want to unload said mod, wipe it from the world, and then reload. I still think, this can be done just with a 'wipe unloaded data' button.

GoodPro712:
I have a command that removed unloaded **tiles**

Jofairden:
I think for 1.4 it would be good to have a feature so you can just completely remove unloaded data from player and world files.

Solxanich:
Hmm. I'm thinking of tackling this next, as its related with the placeholders walls and chests request I worked on. 
To confirm the request proper: 
"For Unloaded World, aim for a method like destroyUnloadedThisModData which removes all placeholders belonging to the mod, and all related unloaded mod data.   Nice to haves for this includes a way to set 'fallbacks' for destroyed placeholders, and being able to specify particular subsets of data to wipe (tiles, items, etc.). Method can be called from World Selection UI via appropriate popup display. Method should not corrupt the save.
Likewise implement for UnloadedPlayer, being called from Character Selection UI."

bvcbvc:
[...]

Issue #37, opened 2016-06-26
Extra placeholders for missing modded content


Mach12:
I know that very, VERY important parts of modded content are already placeholder'd when missing, and restored when the mod comes back. I just thought I'd throw out there a note of things one still loses upon disabling and reenabling a mod:
- Prefixes on items, except (for some reason) accessories in 1-6th slots. Accessories in modded slots still lose their prefixes
- Walls
- Modded armor on mannequins
- Townfolk (yes they come back slowly, but when you test things on vanilla and come back to a whole modpack, it can take several in-game days to get them all back)

So... yeah. Here's a list for reference. I guess it's not major, but the more we keep content when temporarily removing a mod, the better amirite?

Chicken-Bones:
I agree with the first three, though I haven't looked into how difficult they will be. 
As for Townfolk, it may be difficult to manage things such as their rooms getting taken or destroyed. Perhaps it's better to add a console command to fill all available houses with valid town NPCs

Mach12:
How about making them swarm back in when resuming the game? If it's hard to make them seem to always have been there, maybe we could dial such command or similar when joining a world that has lost town NPCs...

bluemagic123:
Hm, I could think of a way to make prefixes, modded armor on mannequins, and townfolk stay. However, walls would be pretty difficult to preserve. I actually wanted to make that happen in the most recent update, but because literally the only thing saved about walls is their type, there isn't any way to preserve them without storing extra per-tile data externally.

Chicken-Bones:
On second thoughts I'm not so sure about retaining prefixes and walls, due to save bloat. Perhaps there should instead be a confirmation message when loading a player/world when such things would be lost

Mach12:
I'm not sure about save bloating for prefixes, but maybe the wall data would be compressible?
I must admit that I'm in half-unknown territory, but travelling around in a laggy server showed me that the map is scattered into chunks.
If it's also the case in the save, maybe we could just not save chunks that do not contain modded walls?
Supposing that no mod replaces every wall in the world, this could spare us most/all of the sky, same for hell, probably also the jungle and dungeon... actually, most chunks would probably be just builded areas and randomly generated underground structures...

bluemagic123:
[...]

PR #1266, opened 2021-01-03
[1.4] TileIO rewrite

Solxanich:
### What is the new feature?
Upfront, this ended up being a fairly large amount of changes.

Resolve feature request made in issue #37 :
1) Add placeholder walls to preserve modded walls through accidental world loading without mod (implementation drafted by and credited to Direwolf420)
2) Add placeholder chests to preserve chest contents through accidental world loading without mod

Further, this also required a redesign in order to minimize impact on file size and inconsistent saving/loading methods for placeholders. This redesign encompasses the following changes:
1) Switching tile placeholders location-specific data from being indexed by modifying tile frames to instead being indexed by a combination of its position (simplified to a single integer) and an integer to integer dictionary. This facilitated for chests and walls to be implemented consistent with tiles, and is readily expandable to other modded instances such as liquids and wires through a dedicated class (PosIndexing). 
2) With the above switch made, per Direwolf420's recommendation, then refactored the unloadedTile class to be unloadedInfo with multiple subclasses for walls, tiles, and chests, so that unique data could be stored with respect to each without affecting each other. Further, with the switch we no longer need to record tile frame data, leading to some minimal file save size reduction.
3) With the above two redesigns complete, one can then reduce the presence of the previous placeholder tile implementation, as it becomes unnecessarily redundant as direct changes in tile types can be readily made without re-indexing.
4) To resolve file size increase concerns raised by Direwolf420, leverage that the dictionary is 
created as ordered integers for the keys and the existing re-occurence code in TileIO so that future calls to the dictionary by position can round down to get the same effective entry instead of creating a new entry for each. Also required implementing a custom dictionary clearing after all tiles that can be restored are restored. Overall, the file size when working with ExampleMod appears to be within control after adding all changes.
5) Adding additional flags save space to the TileIO ReadModTile/WriteModTile to facilitate for location-based chest style preservation, and possible future flags for modded liquid amount or mod-specific custom data. This also includes adding a legacy compat flag to allow for loading previous worlds into the new system without breaking the save. 

This PR resolves the longstanding (5 yrs) issue of placeholders for walls, and preserving chest contents. Mods like Calamity (abyss), Mod of Redemption (underground lab), and Ancients Awakened heavily add modded chests and modded walls, and consequently these objects become quite important to the mod's integrity of play. 
By implementing these placeholders, players who load the world without the requisite mods will not be inadvertently punished by losing access to modded chests contents, or functionality dependent upon the mod's walls.

Not aware of alternative designs that would meet needs, however am open to discussion.

On a similar note, this is a not a complete comprehensive PR. There are additional improvements/deficiencies that could be dealt with:
1) The legacy compat flag depends on at least one modded tile existing in the previous version of the save. This would mean that any world deficient of modded tiles but possessing modded walls would break on loading into new system, however *my opinion* is that this is a edge case to ignore.
2) Further re-factor boiler plate code in UnloadedTilesWorld, in particular in the load function.
3) Possible refactor containers and tile entities to take advantage of feature, where applicable
3) Add modded liquids, modded wires, mod-specific custom data etc. as required based on modded walls implementation

n/A at this time. Feature is intended to be independent to mod implementation.

### ExampleMod updates
Updated example PlayerInventory modifications in ExampleMod to include additional items to facilitate testing, and improve demonstration of adding different types of objects. Not required as part of feature implementation.


direwolf420:
Didn't expect a PR so soon. Concider setting it as a draft instead as in its current state its not really safe to merge.

As I pointed out in the linked issue, I think an important milestone towards this PR is figuring out an efficient and unified way of handling per-tile data so that it is also usable in other mods easily without them having to write hundreds of lines of boilerplate

Solxanich:
Thanks Direwolf - to confirm, the particular save-breaking cases of concern (as far as safety goes) are:
1) Loading a save with pre-existing unloadedTiles, as this won't work with the changes to how tile info is saved 
2) TileIO has the following lines that would lead to breaking changes if unloaded placeholders are created on a save during first-time loading of this change (ie 1.3.5 -> 1.4 with mods missing would break) due to new save data: Line 88 & 89
`["IsChest"] = TileID.Sets.BasicChest[type], // Flag Chests for preserving Chest Tile data 
["IsSolid"] = Main.tileSolid[type], // Flag non-solid blocks for liquids, player movement, etc.` 

And the change dataflags was removed from this PR, so that isn't a concern at this time

To confirm your point on an additional improvement on handling per-tile data such that mod writers can leverage it, would that not be more appropriate to address as a separate PR that improves on the current system? As is, this PR already notes it is not a comprehensive PR as there are outstanding items that can be improved on that have additional impacts and should be separated out (changing how mods are written or breaking all saves by modifying flags being examples)

Solxanich:
Nvm the latter, apologies for my abrasion; I see a few possible opportunities to explore further:
1) I'll look into making the PosIndexer more general so that mods could call it to store data. Were you thinking more of attaching data to the tileInfo or custom objects like Calamity's recharge station? I also guess I'll take a look at the containers and tile entities that have otherwise been untouched through all this if I'm tackling a more generic method.
2) I figured out the backwards compat - the added field IsSolid in the tile overhead tagcompound would not exist on previous saves, so thus a quick ContainsKey is sufficient to if statement around any flags based issues. I'll look to implement.

direwolf420:
1: No, that's tile entities. I meant a way to attach any data to a set of coordinates (most often tile related), e.g. a mod would want to flag a tile as "protected" for mods that add multiplayer protection on tiles (Minecraft has mods like this, albeit for regions), or flags in general (There is a mod called worldtilesflags but it doesnt save tiles efficiently, doesn't support cross mod compatibility, and makes worlds load/save 3-10 seconds longer). And yes, if that were to be a code feature of tml, it would probably be a separate PR is my bet. But I don't know if there are any API devs wanting to tackle this.

2: Nice

Solxanich:
Hmm. Just some discussion on my thoughts on what I would do with the custom data implementation for what you are describing:
1) I would break up and/or refactor UnloadedPosIndexing to better handle the info types (it would be best to generalize it for any datatype of 'info') and make it possible for mod's to call SaveInfoToPos, each referencing their own subclass of UnloadedInfo, and own appropriate Infos list, and matching mapping Dictionary. I would assume writers would populate the 'info' themselves
2) I would refactor the info/canRestore groups in UnloadedTileWorld to make it generic for any info type, similar to what would be done with UnloadedPosIndexing so that it could be called to handle their Infos group on reload
3) Again, similar refactor on data cleanup code in the Load function.
4) Leave any restoration/usage of the infos group to the mod writer to decide. 

This would expose all the elements needed to store by a PosID to a mod author imo; if you are content with that description, I can create an issue requesting exposing Position Indexing to authors, with the above summary of how *I* might implement it, and let that independent feature prioritize itself.

Solxanich:
I'm thinking this should be reviewed soon;
Given it's been a bit, I've also added some stuff from what was going to be part 2 of the rework: implementing structural lines for fallback types and placeholder purges as part #302; refactoring the load function in UnloadedTilesWorld.

Assuming this doesn't become conflicting with modsystems or other major changes, this should be mergeable.

Solxanich:
Kittenchilly + Direwolf comments:
1) Expand to include semi-solid, planters, dressers, and other containers

Will implement dressers and semi-solid, plus structure code for others.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedChestInfo.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Terraria.ModLoader.IO;
+using System.Collections.Generic;
+using Terraria.ID;
+
+namespace Terraria.ModLoader.Default
+{
+	internal class UnloadedChestInfo : UnloadedInfo 
+	{
+		internal byte ChestStyle;
```
Mirsario, line 9: This and many other fields should be in camelCase.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedDresser.cs
```diff
@@ -0,0 +1,59 @@
+using Terraria.ID;
+namespace Terraria.ModLoader.Default
+
+{
+	[Autoload(false)] // need two named versions
+	public class UnloadedDresser : ModTile
+	{
+		public override string Name{get;}
```
Mirsario, line 8:
```suggestion
		public override string Name { get; }
```

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
```diff
@@ -7,111 +7,180 @@ namespace Terraria.ModLoader.Default
...
 
-			if (canRestoreFlag) {
-				RestoreTiles(canRestore);
+			// Cleanup data remnants
+			tileUpdater.cleanupMaps();
+			tileUpdater.cleanupInfos();
+			wallUpdater.cleanupMaps();
+			wallUpdater.cleanupInfos();
+			chestUpdater.cleanupMaps();
+			chestUpdater.cleanupInfos();
```
Mirsario, line 121: All methods must be PascalCase, read the style guide - https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide

patches/tModLoader/Terraria/ModLoader/Default/UnloadedWallInfo.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Terraria.ModLoader.IO;
+
+namespace Terraria.ModLoader.Default
+{
+	internal class UnloadedWallInfo : UnloadedInfo 
+	{
+		
+		public UnloadedWallInfo(string modName, string name, ushort fallbackType = 0) :base(modName, name,fallbackType) {
```
Mirsario, line 8:
```suggestion
	{
		public UnloadedWallInfo(string modName, string name, ushort fallbackType = 0) :base(modName, name,fallbackType) {
```

patches/tModLoader/Terraria/Terraria.csproj.patch
```diff
@@ -82,6 +82,12 @@
      <EmbeddedResource Include="Localization/Content/**" />
      <EmbeddedResource Include="Microsoft/**" />
    </ItemGroup>
++  <ItemGroup>
++    <None Remove="ModLoader\Default\UnloadedDresser.png" />
++  </ItemGroup>
+   <Target Name="EditBin" AfterTargets="Build">
```
Mirsario, line 88: Uh.

Mirsario (changes requested):
Style review mostly.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
```diff
@@ -0,0 +1,60 @@
...
+				pendingID = 0;
+				while (pendingID < infos.Count && infos[pendingID] != null)
+					pendingID++;
+				if (pendingID == infos.Count)
+					infos.Add(info);
+				else
+					infos[pendingID] = info;
+			}
+			modSystem.wallInfoMap[PosID] = pendingID;
+		}
```
Mirsario, line 57: These 3 methods can be turned into 1 with the right parameters.

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/Default/UnloadedChestInfo.cs
Rartrin: Either do like Mirsario said and also add `readonly` to it.
`internal byte readonly chestStyle;`

Or just make it a readonly auto property and leave it UpperCamelCase (considering that it's non-private).
`internal byte ChestStyle { get; }`

patches/tModLoader/Terraria/ModLoader/Default/UnloadedInfo.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Terraria.ModLoader.IO;
+
+namespace Terraria.ModLoader.Default
+{
+	internal class UnloadedInfo
+	{
+		public string modName;
+		public string name;
+		internal ushort FallbackType;
```
Rartrin, line 9: Make these readonly properties and use UpperCamelCase.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
Rartrin: Alternatively, a private method to handle all the pendingID stuff. Though, if things could be generalized to use generics, that would be good too. Probably would need some changes to UnloadedTilesSystem.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedWallInfo.cs
Rartrin: Add a space between `name,` and `fallbackType`.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -265,53 +458,93 @@ internal struct TileTables
...
+							// Place UnloadedChest (required to preserve the inventory and re-namings
+							WorldGen.PlaceChestDirect(i, j + 1, UnloadedTilesSystem.UnloadedChest, 0, -1);
+						}
 					}
-					UnloadedTilesSystem unloadedTiles = ModContent.GetInstance<UnloadedTilesSystem>();
-					int pendingFrameID = unloadedTiles.pendingInfos.IndexOf(info);
-					if (pendingFrameID < 0) {
-						pendingFrameID = unloadedTiles.pendingInfos.Count;
-						unloadedTiles.pendingInfos.Add(info);
+					if (tile.type == PendingDresser) {
```
Rartrin, line 500: `else if` perhaps?

patches/tModLoader/Terraria/ModLoader/Default/UnloadedInfo.cs
Solxanich: Not sure about this one. 
Readonly wouldn't work with current subclasses for the different tile types constructing base values afaik (namely looking at chests while thinking this). I've swapped from public to internal, but that's not getting to what you're looking for I presume.

Per Mirsario's comment, these are attributes so camelCase not Upper.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
Solxanich: Agreed. All order-sensitive stuff is dealt with, so good catch.

patches/tModLoader/Terraria/Terraria.csproj.patch
Solxanich: Fixed. Removed lines; not sure how that got in there, but they're gone

patches/tModLoader/Terraria/ModLoader/Default/UnloadedWallInfo.cs
Solxanich: Resolved spacing issues in all info subclasses.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
Solxanich: Noted - it appears I switched along the way for that class.
Fixed the casing.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedDresser.cs
Solxanich: Fixed all instances of spacing error.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedChestInfo.cs
Solxanich: Fixed this line to be internal readonly byte chestStyle.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedInfo.cs
Rartrin: I was mainly talking about the public ones fields, the conventions for attributes is to use public properties rather than public fields.
As for `internal`, everywhere I've seen them used, they follow nearly the same conventions as `public` because they are practically the same except you have to whitelist which external assemblies can use them. In this case, these attributes are meant to be used outside of the class itself so a property would just make more sense then a field and be cleaner.

As for readonly, I understand what you mean about not being able to be readonly. I wasn't able to see much else that so I was just putting it out there as something to check.

Does make me wonder if there would be a better way to handle it from an object oriented approach to make it more protected. Setting a base field in an extending class outside of its constructor is generally a suboptimal approach.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
Solxanich: So, I refactored all the info subclasses, as since the original refactoring of that, a lot has changed.
I was able to reduce it down to one info class again which meant I could mostly clean this to one method. This should be 'better', although I can't say I feel is now perfect.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedInfo.cs
Solxanich: Thanks for the education there Rartrin; I didn't quite understand the difference between them all as I've only started to do OOP this year. Makes a lot more sense now; and I'm all for safe coding practices.

I made them readonly when I refactored the UnloadedInfo back down to one class, so that's been fixed. Realized I could just store a customData tagcompound in Info for chests and could reduce the classes from there.

Thanks for the review!

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
Solxanich: Alright, got a burst of inspiration and am happy now with the end version of this file.
Very simple class with one method, and short method too.

patches/tModLoader/Terraria/Terraria.csproj.patch
Solxanich: Re: fixed. File no longer a part of PR

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
Solxanich: @Mirsario - are you content with the changes made to this file to clean it up?
If so, I'd like to resolve the change and move towards merging

Solxanich:
Going to toss this here now, as a summary, since part 3 scope is nixed as already covered and no point in having an issue anymore:

1) Refactor across for UnloadedInfo instances and related positionMaps so that can reduce further the repeat code in UnloadedSystemsWorld and related
2) Refactor UnloadedItems to take advantage of the changes made in unloaded tile rework
3) Aim to resolve #302

Solxanich:
To dos for current branch based on insight gained from CB's early review comments, and subsequent revealing of opportunities:
1) fix indexOf for positionindexer per original unloadedtileinfo
2) switch position maps datatype to ordered dictionary to explicit order, rework FloorGetValue to use binary search instead of enumeration
3) realize that can calculate chest style during load instead of retrieving, thus eliminating legacy load flag and need for flags2; further can use chest style of placeholder to store chest style instead of custom data, thus trimming that down to simpler.
4) double check why pending tile variants and can't just grab variants flags when needed?
5) switch the code in pending container handling to change upcoming types to be considered handled instead of looking back
6) Create a flow chart of what TileIO+UnloadedModTilesSystem does so can figure out how to make it understandable

Solxanich:
Additional changes todo after discussion with CB:
1) Prevent hammering of unloadedTile to prevent frame data issues
2) refactor ModTile unloadedChest, unloaded dresser into ModTile
3) Use a calculated, overrideable field (unloadedType) in ModTile to determine variants as part of the og tile. Then can just copy-paste the field from og to placeholder. Use a quick constant byte to indicate what each means.
4) Eliminate chestInfos as its redundant after all changes made; there is nothing unique compared to standard tile
4) Swap all dicts to the appropriate combination of list, array, structure for performance; in particular, the tileTables in TileIO being a big red flag 
5) UnloadedPosIndexing should be static; its just methods, doesn't need identity
6) Eliminate the last artifact of pending flags from the old frame-based system. Store FullName to a field on UnloadedTile, and use that for the UnloadedInfo indexing check in PosIndexing, and rebuild positionMaps each load. Simplify TileTables as possible.
7) Rename fallbackType to possibly "vanillaFallbackOnModDeletion" or "vanillaReplacementType", and implement in ModTile the field to actually get it.

And anything else I see after doing all that. Move to draft.

Solxanich:
I've addressed the vast majority of feedback and made changes accordingly.
The following feedback I haven't addressed as written for following reasons:
1) UnloadedPosIndexing, if we were to support multiplayer run-time tile identification would imply that this should remain non-static to handle multiple different identification requests, assuming I understand static correctly
2) I was not able to fully eliminate the dictionairies in TileIO tileTables. I've brought it down from 9 dictionairies to 3. The main issue is that the tile sweep on load time creates a situation where the key's appear many to one, in addition to out of order. I believe dictionairies are appropriate here.
3) there is a few spots where refactoring could be done, but not at my current knowledge level of C#. I've flagged the three locations I've identified with comments. 
4) Added a method for getting tile placeholder type instead of a property on TileIO, based on passing string and subsequent lookup. Mod authors shouldn't need access to specify, as could lead to unpredictable behavior with current code design.

Solxanich:
Finished running testing; verified legacy load compatibility, and verified all methods work as expected.
RE: ready for review.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
```diff
@@ -5,113 +5,49 @@
...
-			infos.Clear();
-			pendingInfos.Clear();
+			tileInfos.Clear();
+			wallInfos.Clear();
 		}
 
 		public override TagCompound SaveWorldData() {
 			return new TagCompound {
-				["list"] = infos.Select(info => info?.Save() ?? new TagCompound()).ToList()
+				["tileInfos"] = tileInfos.Select(info => info?.Save() ?? new TagCompound()).ToList(),
```
Chicken-Bones, line 26: null infos are allowed?

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
```diff
@@ -5,113 +5,49 @@
...
-				bool frameImportant = infoTag.ContainsKey("frameX");
-				var info = frameImportant ?
-					new UnloadedTileInfo(modName, name, infoTag.GetShort("frameX"), infoTag.GetShort("frameY")) :
-					new UnloadedTileInfo(modName, name);
-
-				infos.Add(info);
-
-				int type = ModContent.TryFind(modName, name, out ModTile tile) ? tile.Type : 0;
+			// Process tiles
+			tileInfos.AddRange(TileIO.tileInfos);
```
Chicken-Bones, line 33: I think `TileIO` should have a more active interface with this class. Rather than merging two separate lists in this hook

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -12,106 +12,261 @@ namespace Terraria.ModLoader.IO
...
+				}).ToList(),
+
+				["wallMap"] = wallList, // Lists of all active wall types and restoration mapping
+				["wallInfoMap"] = wallInfoMap.Select(pair => new TagCompound {
+					["posID"] = pair.posID,
+					["infoID"] = pair.infoID
+				}).ToList(),
+
+				["data"] = ms.ToArray(), // Array of locational-specific data
+			};
```
Chicken-Bones, line 162: 100/100 good stuff

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -12,106 +12,261 @@ namespace Terraria.ModLoader.IO
...
+					tabData.modName = modName;
+					tabData.name = name;
+					
+					ushort fallbackType = 0;
+					if (wallTag.ContainsKey("fallbackType"))
+						fallbackType = wallTag.Get<ushort>("fallbackType");
+
+					tabData.fallbackID = fallbackType;
+
+					UpdateUnloaded pendingInfo = new UpdateUnloaded(wallInfos, WallsContext);
```
Chicken-Bones, line 239: this kind of temporary helper class is not a typical C# pattern. I'd recommend using static methods when possible.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
```diff
@@ -12,106 +12,261 @@ namespace Terraria.ModLoader.IO
...
 				string modName = wallTag.GetString("mod");
 				string name = wallTag.GetString("name");
-				tables.walls[type] = ModContent.TryFind(modName, name, out ModWall wall) ? wall.Type : (ushort)0;
+								
+				tabData.id = ModContent.TryFind(modName, name, out ModWall wall) ? wall.Type : (ushort)0;
+				
+				if (tabData.id == 0) {
+					tabData.id = ModContent.Find<ModWall>("ModLoader/UnloadedWall").Type;
+					tabData.modName = modName;
+					tabData.name = name;
```
Chicken-Bones, line 231: should `tabData` have an `unloadedInfo` field instead?

Chicken-Bones (changes requested):
Good improvements since last time. I'll admit, I ran out of brain before I finished this review, but...

What are your thoughts on not representing unloaded data in the save file? That avoids having to merge newly unloaded tiles/walls with previously unloaded ones, and avoids having to handle a partial restoration of previously unloaded tiles/walls from a different save format.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
direwolf420: Yes, the length of the list of infos never changes, "recovered" data is just nulled out. this keeps indexes consistent

Solxanich: Thanks for responding Dire - Expounding a bit, the only usage is having fixed indices being stored in the mapping. 
The nulled entries "should" then be reusable as posIndexing should use null entries instead of adding net new.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
```diff
@@ -0,0 +1,70 @@
...
+				}
+
+				if (posMap[split].posID <= this.posID) {
+					minimum = split;
+				}
+
+				if (posMap[split + 1].posID > this.posID) {
+					maximum = split;
+				}
+
```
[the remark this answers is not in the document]
Solxanich: Given we are looking for the interval from posMap[split] to posMap[split + 1] containing posID, it may be more effective to separate it out as it decreases the amount of iterations for elements closer to center. IIRC it should be net beneficial, as >80% of domain benefits.

An example being that if the posID is in the dead middle (say 50.1 / 100) - if I use if else, it would find the minimum is 50, but then take another few checks to determine the maximum is also 50.

Solxanich:
@Chicken-Bones - per our discussion, I've gone a whole new route and rewrote TileIO with just doing tiles for now. Can you take a look, before I start thinking about generics for doing walls using the same methods?

@direwolf420  - the PosIndexer has been generalized and should be closer to what you were looking for - let me know if other methods to add.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
Solxanich: Not applicable anymore.

patches/tModLoader/Terraria/ModLoader/IO/TileIO.cs
Solxanich: Also not applicable anymore.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
Solxanich: Class is gone, so not applicable.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTilesSystem.cs
Solxanich: The entries shouldn't be null anymore, as tracking indexes are all run time now.
I've kept the null check for now, but it *should* be safe to remove, assuming everything works.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTile.cs
```diff
@@ -1,37 +1,138 @@
...
+			Player player = Main.LocalPlayer; 
+			var modSystem = ModContent.GetInstance<UnloadedTilesSystem>();
+			var posIndex = new UnloadedPosIndexing(i, j);
+
+			//NOTE: Onwards only works in singleplayer, as the lists aren't synced afaik.
+			int infoID = posIndex.FloorGetValue(TileIO.tileInfoMap.ToArray());
+			var infos = modSystem.tileInfos;
+
+			if (infoID < infos.Count) {
+				var info = infos[infoID];
```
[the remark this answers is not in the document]
Solxanich: I've simplified this to be a fair few less lines. 
PosIndexer.GetKeyFromPos() returns the index for a given position, in the provided position map, which is more general and usable elsewise, but still simplified.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPosIndexing.cs
```diff
@@ -0,0 +1,70 @@
...
+		public void GetCoords(out int posX, out int posY) {
+			posY = posID % Main.maxTilesY;
+			posX = posID / Main.maxTilesY;
+		}
+
+		public void MapPosToInfo(List<UnloadedInfo> infos, List<TileIO.posMap> posMap, List<TileIO.posMap> prevPosMap = null, UnloadedInfo info = null)  {
+			ushort pendingID = 0;
+			if (info == null && prevPosMap.Count > 0) {
+				pendingID = (ushort) (FloorGetValue(prevPosMap.ToArray()) + infos.Count);
```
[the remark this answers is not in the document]
Solxanich: Fixed. Misunderstood the request while discussing. 
Uses a list while generating, and converts list to an array for run time.

Performance is slightly slower on saving, didn't notice any visual indication of performance drop during run time when using 
[ReplaceTilesCommand.zip](https://github.com/tModLoader/tModLoader/files/6031450/ReplaceTilesCommand.zip)

Should be good now.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTile.cs
```diff
@@ -1,37 +1,138 @@
...
-	{
-		public override string Name{get;}
+	[Autoload(false)] // need multiple named versions
+	public class UnloadedTile : ModTile {
+		public override string Name { get; }
+
+		internal bool isSolid;
+		internal bool isSemi;
+		internal bool isChest;
+		internal bool isDresser;
```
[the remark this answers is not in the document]
Solxanich: TODO self: probably subclass, can unique-ify the textures so they can be distinguished, such as chests and dressers

patches/tModLoader/Terraria/ModLoader/Default/UnloadedInfo.cs
```diff
@@ -0,0 +1,39 @@
...
+				return false;
+			}
+			if (modName != other.modName || name != other.name) {
+				return false;
+			}
+			return true;
+		}
+
+		public override int GetHashCode() {
+			int hash = name.GetHashCode() + modName.GetHashCode();
```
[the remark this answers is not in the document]
Solxanich: The concern here is collision, correct?
Prime multiply and add only holds for uniqueness on counting numbers iirc. Given GetHashCode can return negative, that won't be sufficient to eliminate collision. I can abs, but that still isn't perfect.  
XOR has same negative issue. It prevents overflow so I tend to prefer it, so I'll implement with absolute as that's better than adding, but I'm open to better ideas.

patches/tModLoader/Terraria/ModLoader/PosData.cs
```diff
@@ -0,0 +1,150 @@
...
+		/// </summary>
+		/// <param name="posMap"></param>
+		/// <param name="posID"></param>
+		/// <returns></returns>
+		public static int FindIndex<T>(this PosData<T>[] posMap, int pos) {
+			int minimum = -1, maximum = posMap.Length;
+			while (maximum - minimum > 1) {
+				int split = (minimum + maximum) / 2;
+
+				if (posMap[split].pos <= pos)
```
Solxanich, line 83: Reverse this, again per original review, its wrong for an interval search as the exit condition takes wasted cycles to reach

patches/tModLoader/Terraria/ModLoader/IO/ModEntry.cs
```diff
@@ -1,51 +1,41 @@
...
-			if (modName != other.modName || name != other.name) {
-				return false;
-			}
-			return true;
+		public ModEntry(ModBlockType block) {
+			type = block.Type;
+			modName = block.Mod.Name;
+			name = block.Name;
+			vanillaReplacementType = block.vanillaFallbackOnModDeletion;
+			throw new NotImplementedException();// unloadedType = ?
```
Solxanich, line 27: Unloaded type was implemented, there was a method in TileIO basic to calc it.
Why is this here?

patches/tModLoader/Terraria/ModLoader/Default/UnloadedTile.cs
```diff
@@ -2,40 +2,25 @@
...
-			if (infoID < infos.Count) {
-				var info = infos[infoID];
-
-				if (info != null) {
-					player.cursorItemIconEnabled = true;
-					player.cursorItemIconID = -1;
-					player.cursorItemIconText = $"{info.modName}: {info.name}";
-				}
+			var info = TileIO.tileEntries[TileIO.unloadedTileLookup.Lookup(i, j)];
+			if (info != null) {
```
Solxanich, line 17: Need to test this

patches/tModLoader/Terraria/ModLoader/IO/TileIO_Basic.cs
```diff
@@ -7,206 +7,81 @@
 namespace Terraria.ModLoader.IO
 {
 	internal static partial class TileIO {
-		static string tileEntries = "tileMap";
```
Solxanich: Why is this deleted with no replacement?

Solxanich (commented):
Some to dos.

patches/tModLoader/Terraria/ModLoader/IO/WallEntry.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using Terraria.ID;
+
+namespace Terraria.ModLoader.IO
+{
+	internal class WallEntry : ModEntry
+	{
+		public static Func<TagCompound, WallEntry> DESERIALIZER = tag => DeserializeTag(tag);
+
+		public static WallEntry DeserializeTag(TagCompound tag) {
```
Solxanich, line 10: This *probably* could be generalized and moved up to ModEntry, but I missed that when doing it.

patches/tModLoader/Terraria/ModLoader/PosData.cs
```diff
@@ -0,0 +1,166 @@
...
+		/// <summary>
+		/// Searches for the interval posMap[i].posID < provided posID < posMap[i + 1].posID.
+		/// </summary>
+		/// <param name="posMap"></param>
+		/// <param name="posID"></param>
+		/// <returns></returns>
+		public static int FindIndex<T>(this PosData<T>[] posMap, int pos) {
+			if (posMap.Length == 0) {
+				throw new ArgumentException($"Can't find the index in an empty posMap. Please verify map is non-empty before calling.");
+			}
```
Chicken-Bones, line 94: please remove this condition. I updated the iteration to return -1 when `pos` is < the first entry in the map (or the map is empty)

This works naturally with the builder, as there's essentially a phantom element at `-1` with the value of `default(T)` this means you don't ever have to check that the map is non-empty.

Conceptually, an empty map means that the entire world is filled with `default(T)`, which is a sensible thing to work with

Chicken-Bones (changes requested):

bvcbvc:
i understand none of this, but yes unloaded tiles deserve this treatment

patches/tModLoader/Terraria/ModLoader/PosData.cs
```diff
@@ -116,26 +130,44 @@ public static class PosData
...
-			var i = posMap.FindIndex(pos);
-			bool fail = i == -1;
-			data = fail ? default : posMap[i].value;
-			return !fail;
-		}
+		/// <summary>
+		/// General purpose lookup function. Always returns a value (even if that value is `default`).
+		/// See <see cref="PosData{T}.OrderedSparseLookupBuilder.OrderedSparseLookupBuilder(int, bool, bool)"/>for more info
+		/// </summary>
+		public static T Lookup<T>(this PosData<T>[] posMap, int x, int y) => posMap.Lookup(CoordsToPos(x, y));
```
Solxanich, line 141: Modders can always directly call posData.Find() if using default values as valid values and use that to check if data is valid.

Solxanich (commented):
Review commit 1.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedWall.cs
```diff
@@ -7,7 +7,7 @@ public class UnloadedWall : ModWall
 		public override string Texture => "ModLoader/UnloadedWall";
 
 		public override void SetDefaults() {
-			TileIO.IsUnloadedWall.Add(Type);
+			TileIO.Tiles.unloadedTypes.Add(Type);
```
Solxanich, line 10: Double check

Solxanich, line 10: Double check if should be Walls

patches/tModLoader/Terraria/ModLoader/IO/ModEntry.cs
```diff
@@ -9,26 +9,28 @@ public abstract class ModEntry : TagSerializable
 		public string unloadedType;
 		public ushort loadedType;
 
-		public bool IsUnloaded => loadedType != type;
-
-		public abstract string DefaultUnloadedType { get; }
-
-		internal virtual void SetData<B>(B block) where B : ModBlockType {
+		protected ModEntry(ModBlockType block) {
```
Solxanich, line 12: Double check how got constructors to work in generics?

patches/tModLoader/Terraria/ModLoader/IO/TileIO_Basic.cs
```diff
@@ -6,265 +7,243 @@
 namespace Terraria.ModLoader.IO
 {
 	internal static partial class TileIO {
-		// todo, resize arrays or content tags
-		public static List<ushort> IsUnloadedTile = new List<ushort>();
-		public static List<ushort> IsUnloadedWall = new List<ushort>();
+		public abstract class IOImpl<TBlock, TEntry> where TBlock : ModBlockType where TEntry : ModEntry
```
Solxanich, line 10: Privacy mismatch. Internal

patches/tModLoader/Terraria/ModLoader/IO/TileIO_Basic.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
Solxanich, line 1: Figure out where call:
TileIO.LoadNewWorld

Solxanich (commented):
Review commit 2; will need to look at TileIO basic in VS as its non-trivial to look through

patches/tModLoader/Terraria/ModLoader/IO/ModEntry.cs
Solxanich: Didn't just used inheritance.

patches/tModLoader/Terraria/ModLoader/IO/TileIO_Basic.cs
Solxanich: Was unneeded. LoadEntries assigns CreateEntries off the get go, and unloadedEntryLookup is assigned to by builder.Build in LoadData before it is used.

patches/tModLoader/Terraria/ModLoader/PosData.cs
```diff
@@ -23,16 +38,13 @@ public class OrderedSparseLookupBuilder
...
 
-				list.Add(new PosData<T>(pos, value));
-			}
-
-			public void ClusteredAdd(int x, int y, T value) => ClusteredAdd(PosData.CoordsToPos(x, y), value);
-
-			public void ClusteredAdd(int pos, T value) {
-				if (pos <= last.pos)
-					throw new ArgumentException($"Must build in ascending index order. Prev: {last.pos}, pos: {pos}");
+				if (compressEqualValues && !insertDefaultEntries && pos > last.pos+1) {
```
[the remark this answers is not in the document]
Solxanich: Yeah, no ! on insertDefaultEntries.

Issue #1176, opened 2020-11-01
GlobalTile.KillTile probably shouldn't be called during world gen


hamstar0:
### Description
Strange side effects (Guide not existing in the world) from my KillTile hook (spawns custom critters from cut grass) being called during world gen.

### Log file
None applicable.

### Steps to reproduce
See above.

### Expected behavior 
A living guide at the default spawn.

### Actual behavior
Guide doesn't exist for some reason.

### Reproduction frequency
Always?

JavidPack:
It is expected, and desired. 

If you want to prevent drops, I suggest checking if(!noItem && !effectOnly && !WorldGen.stopDrops), or just maybe use DropCritterChance to manually spawn instead, since it already checks those.

Chicken-Bones:
This should be documented somewhere. If there are particularly relevant flags for KillTile behaviour, then we could include them as parameters in 1.4 to make it clear they should be considered.

Mirsario:
This is a discussion issue, don't just close it right away. I can't come up with any reason to run KillTile in worldgen, and checking against WorldGen.gen is boilerplate that could slow down generation.

Mirsario:
Some pointed out in discord that sometimes multitiles need fixes applied in KillTile. So, yeah, let's just add a parameter into the context in 1.4.

Chicken-Bones:
@JavidPack is this still relevant? Is there an easy guideline here we can put in some docs? A recommendation to check `WorldGen.generatingWorld`?

JavidPack:
I think modders might just need to replicate the logic of WorldGen.KillTile, still. Conceptually TileLoader.KillTile both sets behavior flags and acts on them, so it's another poorly designed hook.

I think it'll need to be 2 or more hooks, but I'll need to make sense of the code more to be sure. 

`WorldGen.stopDrops` is also relevant, it is what stops meteors from spawning items/critters.

PR #1267, opened 2021-01-04
[1.4] Add worldGen to KillTile hooks

fakelag:
### What is the addition?
Add `worldGen` to KillTile hooks in reference to https://github.com/tModLoader/tModLoader/issues/1176

Based on original issue, world generation is relevant context in this hook, and I doubt it will add any performance overhead.

World gen state might be obtainable from some global interfaces or variables? It might still be relevant enough to be included as a param.

```C#
public class ExampleOre : ModTile
{
	public override void KillTile(int i, int j, ref bool fail, ref bool effectOnly, ref bool noItem, bool worldGen) {
		if (!worldGen) fail = true; // unbreakable, but doesn't get in the way of world generation passes
	}
}
```

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
@@ -127,7 +127,7 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
The `gen` parameter you're passing is static. You can use `WorldGen.gen` from anywhere in your mod

Issue #114, opened 2017-01-25
Multiplayer Framerate


Rashaln:
I attempted to set up a tModLoader server recently, but experienced severe frame skipping and a notable overall framerate decrease when connected to it. I have so far found no fix for it, and I can only seem to trace it to tModLoader in general. I am running a tModLoader server on a Ubuntu 16.04 Live Linux box on my home network, connecting from a Windows 10 client. No matter how far I turn down Terraria settings; tiny window, everything off(except frame skip) or turned down, I lose 10-20 FPS (With frequent stuttering from frame skip) when connecting to my server. I've done a clean install of Terraria, and tested multiple configurations; vanilla Terraria runs fine in single- and multi-player at high graphical settings, tModLoader runs fine in single-player with and without mods, at the same graphical settings, but tModLoader multiplayer, with or without mods, has significant performance problems. My server doesn't appear to be the direct cause, as it stays under 50% of either core while tModLoader is running.

Some system info:
Client:
Terraria 1.3.4.4, TML 0.9.1
Windows 10 Pro 64bit build 14393
AMD Athlon 64 X2 4600+ 2.4ghz
4GB RAM
Nvidia GeForce 9600GT

Server:
Terraria Linux server 1.3.4.4, TML Linux 0.9.1
Ubuntu 16.04 LTS 64-bit (Running off USB with persistence)
Intel Core 2 Duo E8500 3.16ghz
4GB RAM
Onboard graphics

Perhaps I'm just missing something, (I've had bad luck with troubleshooting lately) but hopefully someone can help. I will provide more detailed information if necessary.

Chicken-Bones:
What happens when you run the server locally on your windows machine?

Rashaln:
Ah I can't believe I forgot to test in that configuration. My computer chugged a bit, as expected, but the local game ran with far fewer hitches and NPC warping than it did when connecting to my server, and Host&play actually ran surprisingly smoothly.

Chicken-Bones:
And last thing to test is, Could you run a windows server on another machine? (see if it's a remote/local issue or a windows/linux issue)

Rashaln:
Connecting to the Windows server did provide a better experience than the Linux one, but there is still a disconcerting performance hit. The game didn't drop large numbers of frames, but I did still experience a 10-20 fps drop overall.

Rashaln:
Any word on this?

Pop000100:
try disabling Frameskip my brother had similar problem with it

[...]

PR #1272, opened 2021-01-08
[1.3,1.4] multiplayer CPU Usage fix

Solxanich:
### What is the bug?
Issue #114 - GPU rendering being bottlenecked by CPU frame submission.
CPU bottlenecking was being caused by continuous read operations (from innerClientLoop thread) occurring on a single-core app thus hogging the CPU from other tasks. 
 
### How did you fix the bug?
Re-added Thread.Sleep(1) to NetPlay.cs, and added a comment as to why it is there.

A full redesign of Vanilla's netcode could be done to avoid this dependency (basically eliminate NetStream objects and related async reads that induce this issue).
Defaulting to 2 or more cores for the application, or reducing the number of threads so that one core could handle all would also be options, but again are largely Vanilla redesigns.


patches/tModLoader/Terraria/Netplay.cs.patch
```diff
@@ -84,12 +84,15 @@
...
+@@ -512,8 +_,9 @@
  								Main.statusText = Connection.StatusText + ": " + (int)((float)Connection.StatusCount / (float)Connection.StatusMax * 100f) + "%";
  							}
  						}
 -
 -						Thread.Sleep(1);
++						/*instead of letting the thread that checks for new data continually run on the CPU 
++						// (thus taking away time from other processes) make the thread sleep for 1 millisecond between calls. */
++						Thread.Sleep(1); 
  					}
```
Mirsario, line 96: This is present in vanilla, so just delete this entire specific patch instead?

Mirsario (changes requested):

patches/tModLoader/Terraria/Netplay.cs.patch
Chicken-Bones: Is this fix confirmed? Like, it does make some sense but it's not on the main thread right?

Solxanich: For Mirsario - there are other patches in Netplay.cs.patch that we're added and are independent to this, so I'm hesitant to delete the whole patch file. That said, I agree cleanup the patch file as seeing remove and add looks really ugly, just not sure how without directly editing it.

For Chicken-Bones - steam (social) connections invoke a new thread in class method StartSocialClient(). Likewise TCP connections invoke a new thread in StartTcpClient(). 
And just to confirm the fix via recordings in case any doubt.
Before: (on 1.4 branch)
https://user-images.githubusercontent.com/59670736/104056926-7d1acd80-51ae-11eb-804a-e89d22a8d9ed.mp4

 After: (on this branch)
https://user-images.githubusercontent.com/59670736/104057341-39749380-51af-11eb-98a7-607388173a3b.mp4

Mirsario: I did not mention anything about files.

Solxanich: Okay, then I'm lost on your request. 
Thread.Sleep(1) was removed 4+ years ago. I'm not sure why it's still showing up in this diff.
Thread.Sleep(1) + comment was added yesterday to the branch. That is the change.

Again, semi-new to patching, so I'm not sure why the removal statement is showing up. Realistically, the final patch file should just be add comment. I'll try and clean that up so its like that, but I'm only aware of direct edit the patch file to make that so

Mirsario: I removed the patch myself. There's not much reason for the comment of what basically is just vanilla code.

On one of the videos you posted, you accidentally go into singleplayer.

Solxanich: Oh, I see what you mean now. Yeah, makes sense. Thanks for making change. Main reason for comment is so that it didn't get removed later again by someone else, but yeah, vanilla is vanilla so shouldn't be touching anyways.

Yeah, in the base case I go into singleplayer for first 20ish seconds to show base CPU, then mutliplayer for remainder to show previous CPU. (hence why its 20 sec longer)
The new case I just do multiplayer since no point in doubling down on singleplayer recording (although I did open it in SP to be sure its consistent, separately).
Should have timestamped original post, fair point.

Mirsario: I see, didn't notice that it's longer. I'll merge this now I guess.

PR #1275, opened 2021-01-09
[Changes Requested][1.3] Resolve multiplayer frame drops in 1.3 version, matching PR 1272 of 1.4

Solxanich:
Resolves longstanding issue of multiplayer frame drops.
Patch 1.3 per the following change: https://github.com/tModLoader/tModLoader/pull/1272/commits/31713772a7d5c2c8534199581cfc4dd994c985f5
As part of: https://github.com/tModLoader/tModLoader/pull/1272

*My opinion* is we should patch this on 1.3 as well, while 1.4 is still being developed.

### DIFF
--- a/patches/tModLoader/Terraria/Netplay.cs.patch
+++ b/patches/tModLoader/Terraria/Netplay.cs.patch
@@ -123,7 +123,6 @@
  							NetMessage.SendData(8, -1, -1, null, Main.player[Main.myPlayer].SpawnX, Main.player[Main.myPlayer].SpawnY);
  						}
  
-
 @@ -307,6 +_,7 @@
  					num = Connection.State;
  				}
--- a/patches/tModLoader/Terraria/Netplay.cs.patch
+++ b/patches/tModLoader/Terraria/Netplay.cs.patch
@@ -123,15 +123,7 @@
  							NetMessage.SendData(8, -1, -1, null, Main.player[Main.myPlayer].SpawnX, Main.player[Main.myPlayer].SpawnY);
  						}
  
-@@ -296,8 +_,6 @@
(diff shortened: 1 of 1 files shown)


Solxanich:
@JavidPack 
Just checking if you've had a chance to look at this backport fix from 1.4. 
No worries if not; not sure when a next update, if ever, would happen for 1.3 tML

Solxanich:
https://github.com/tModLoader/tModLoader/issues/114  - so apparently someone has already verified this backport in 1.3.
I'm gonna change title to be ready for merge.

Chicken-Bones:
1.4 fixed this in vanilla. I'd rather we backport the 1.4 fix. This will work fine, but the sleep will induce latency up to 16ms (windows timeslice size) in packet handling. 

1.4 calls the update methods directly from the main thread, rather than relying on a worker thread to post update actions to a queue.

Solxanich:
Hmm. Unfortunate; we removed the sleep in 1.3 and earlier in order to prevent idle cycles in main, at the cost of unnecessary multiplayer reads. 

Now that 1.4 is separate thread for async netcode, that earlier decision no longer applies so we could comfortably fix it.

I'm going to close given that a) the fix is based on 1.4 vanilla code; b) I don't have a 1.3 contributor environment setup; and c) this doesn't particularly seem to be a rush fix relative to 1.4 betas, with a workaround already posted in the og issue.

Solxanich:
I'm reopened this as I shouldnt say this is a non-priority by myself

Solxanich:
Hmm. On further thought, I think it would be appropriate to just remove the patch so we atleast match vanilla on that. 
Improving 1.3 performance by adding the second thread would be better, but I think just doing this is sufficient as atleast makes performance notably better than it is today, and brings it to be comparable to 1.3 vanilla

PR #1278, opened 2021-01-11
[1.4] Add ModifyLuck, PreModifyLuck, and ModifyTorchLuck.

kittenchilly:
These hooks allow you to modify everything regarding a players luck, with some new ExampleMod examples (ExampleLuckPlayer in Common)
All three hooks are in ModPlayer. ModifyLuck just simply modifies the luck value after vanilla makes their calculations, which is what will be used in most cases. PreModifyLuck is a bool that allows you to modify values before vanilla's luck calculations or stop vanilla luck calculations from happening in the first place. ModifyTorchLuck can modify... torch luck; it needs to be a separate hook since all torches accumulate into one torch luck variable, which is capped.


ExampleMod/Common/Players/ExampleLuckPlayer.cs
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		public override void ModifyTorchLuck(ref float positiveLuck, ref float negativeLuck) { //ModifyTorchLuck is more unique than the others as it specifically deals with luck given by torches.
+			//ModifyTorchLuck is separated from the other hooks since all torches nearby contribute to the same amount of overall torch luck, which has a cap.
+			if (ExampleTorchNearby) { //If an ExampleTorch is nearby (set in ExampleTorch's file)...
+				positiveLuck += 1f; //...the torch has a positive influence on overall torch luck.
+			}
+			else
+				negativeLuck += 0.1f; //...otherwise it has a small negative influence on overall torch luck.
+
```
Mirsario, line 36: Don't mix bracket styles/presence within connected blocks.

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -340,6 +340,30 @@ public abstract class ModPlayer : ModType
...
+		/// <param name="luck"></param>
+		public virtual void ModifyLuck(ref float luck) {
+		}
+
+		/// <summary>
+		/// Allows you to modify the variables that affect overall torch luck. Increase positiveLuck if you want it to affect luck positively, and increase (note: increase) negativeLuck if you want to affect luck negatively.
+		/// </summary>
+		/// <param name="positiveLuck"></param>
+		/// <param name="negativeLuck"></param
+		public virtual void ModifyTorchLuck(ref float positiveLuck, ref float negativeLuck) {
```
Mirsario, line 363:
```suggestion
		/// <param name="positiveLuck"></param>
		/// <param name="negativeLuck"></param>
		public virtual void ModifyTorchLuck(ref float positiveLuck, ref float negativeLuck) {
```

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
kittenchilly: oh oops lol

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -340,6 +340,30 @@ public abstract class ModPlayer : ModType
...
+		/// <param name="luck"></param>
+		public virtual void ModifyLuck(ref float luck) {
+		}
+
+		/// <summary>
+		/// Allows you to modify the variables that affect overall torch luck. Increase positiveLuck if you want it to affect luck positively, and increase (note: increase) negativeLuck if you want to affect luck negatively.
+		/// </summary>
+		/// <param name="positiveLuck"></param>
+		/// <param name="negativeLuck"></param>
+		public virtual void ModifyTorchLuck(ref float positiveLuck, ref float negativeLuck) {
```
Chicken-Bones, line 363: This is really hard to use correctly. Please add much more info on exactly how `positiveLuck` and `negativeLuck` work.

Final `torchLuck` can only ever be 0.5 or 1, which results in either 0.1 or 0.2 luck bonus.
Tell modders that the normal use case is to "add 1 to positive luck for a good luck torch, or 1 to negative luck for a bad luck torch" "overall positiveLuck > 1 has no effect" "add a smaller amount (eg 0.5) for smaller impact, or add 2 to negativeLuck to negate all torch luck completely."

Chicken-Bones (changes requested):
Good PR! Just needs some multiplayer and documentation improvements

patches/tModLoader/Terraria/MessageBuffer.cs.patch
```diff
@@ -361,6 +361,14 @@
  							tileEntity.ID = num48;
  							TileEntity.ByID[tileEntity.ID] = tileEntity;
  							TileEntity.ByPosition[tileEntity.Position] = tileEntity;
+@@ -3304,6 +_,7 @@
+ 						obj.luckPotion = luckPotion;
+ 						obj.HasGardenGnomeNearby = hasGardenGnomeNearby;
+ 						obj.RecalculateLuck();
++						PlayerHooks.ModifyLuck(obj, ref obj.luck); //Neccesary? Should this be in PlayerIO instead?
```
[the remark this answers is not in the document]
kittenchilly: I did that initially, but I didn't like how returning PreModifyLuck false also made it so ModifyLuck didn't do anything (since it cancelled out all of RecalculateLuck), so I put it outside the method.

Chicken-Bones:
Use a `goto skipVanillaLuck;` statement

ExampleMod/Common/Players/ExampleLuckPlayer.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Terraria;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	public class ExampleLuckPlayer : ModPlayer
+	{
+		public bool ExampleTorchNearby;
```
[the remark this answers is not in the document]
kittenchilly: This is ModPlayer, its all client-side. Luck is calculated client-side, so theres not really a point in syncing it with the server.

Solxanich (commented):
Ah ok. I see. I took a look at PR more in depth since Discord 
So there are two luck data pieces associated with the tile itself - positive, neg luck multipliers. 
Kitten, Id consider it so that:
1) Nearbyeffects in exampletorch adds the tile id to a dynamic list of modded torches in modplayer instead of a bool for each torch
2) for each tile id in nearby torch list, modifytorchluck accesses its declared positive and negative luck multipliers as set in their instance of modtile
3) see if can refactor away in that system to remove need for exampleluckplayer modifyTorchLuck

As an aside, you could also do a list for item ids (ideally a combined list) if we want to generalize beyond tiles under this same style, I would anticipate

Chicken-Bones (approved):

Mirsario (approved):
I would prefer to have this later be replaced with a more declarative version, involving a list (or yields) of structures with values that will get added up together, linked to descriptions of sources of associated additions or subtractions from luck. That way people would be able to enumerate these lists in a readonly form later on and selectively modify their values. This would let people easily mess with luck in gameplay mods, and show a list of active luck modifiers in helper/GUI mods. What's here just works for content mods, I guess.

PR #1280, opened 2021-01-13
[1.4] Fix NPCID.Sets.MPAllowedEnemies ignoring modded entries + ExampleMod adjustment

direwolf420:
### Description
Fixes a hardcoded NPCID.Count value to use NPCID.Sets.MPAllowedEnemies.Length instead to make it see modded entries.

Also adds new 1.4 "boss summon" item code by adjusting current "Plantera Item" in ExampleMod.

### DIFF
--- a/ExampleMod/Content/Items/Consumables/PlanteraItem.cs
+++ b/ExampleMod/Content/Items/Consumables/PlanteraItem.cs
@@ -6,37 +6,59 @@ using Terraria.ModLoader;
 
 namespace ExampleMod.Content.Items.Consumables
 {
+	//This is the Item used to summon a boss, in this case the vanilla Plantera boss.
 	public class PlanteraItem : ModItem
 	{
 		public override void SetStaticDefaults() {
 			DisplayName.SetDefault("Plantera");
 			Tooltip.SetDefault("The wrath of the jungle");
 			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 3;
-			ItemID.Sets.SortingPriorityBossSpawns[item.type] = 12; // This helps sort inventory know that this is a boss summoning item.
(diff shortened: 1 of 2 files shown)


ExampleMod/Content/Items/Consumables/PlanteraItem.cs
```diff
@@ -6,13 +6,23 @@
...
+			ItemID.Sets.SortingPriorityBossSpawns[Type] = 12; // This helps sort inventory know that this is a boss summoning Item.
+
+			//This is set to true for all NPCs that can be summoned via an Item (calling NPC.SpawnOnPlayer). If this is for a modded boss,
+			//write this in the bosses file instead
+			NPCID.Sets.MPAllowedEnemies[NPCID.Plantera] = true;
+		}
+
+		public override void Unload() {
+			//If the mod is disabled, the flag should be reset again to retain the original state of the game
+			NPCID.Sets.MPAllowedEnemies[NPCID.Plantera] = false;
```
Chicken-Bones, line 25: @Mirsario should tML be resetting this array? I guess we don't revert the effects of GlobalX on Sets arrays normally. 
Wait.. we re-run the static constructors which should reset them yes?

Chicken-Bones: @direwolf420 please test if this is required.

Chicken-Bones (changes requested):

ExampleMod/Content/Items/Consumables/PlanteraItem.cs
Mirsario: Yes, all sets get reset, unloading isn't needed.

direwolf420: Yes, just tested, printing false when ExampleMod is unloaded again, will fix

PR #1281, opened 2021-01-13
[1.4] Auto treasure bag droptable registering + NPC setDefaultsToType constructor

direwolf420:
### Description
With 1.4, the old methods of dropping items from NPCs are now obsolete (literally, the `NPC.NPCLootOld()` is not referenced anywhere), with it the way TML added boss bags to the drop logic. 

Added a comment on the obsolete method in case of compatibility, aswell as made it so after vanilla registers all boss loot, TML registers boss bags using the previous 1.3 logic, meaning modders don't have to manually register bags in 1.4 (They sort of had to in 1.3 by checking `Main.expertMode` and calling `NPC.DropBossBags()`).

As an aside, added a convenience constructor for `NPC` (the same way that a previous commit added a new `Item` constructor) that automatically calls SetDefaults with the given type on it.

### DIFF
--- a/patches/tModLoader/Terraria/NPC.cs.patch
+++ b/patches/tModLoader/Terraria/NPC.cs.patch
@@ -30,6 +30,19 @@
  		public bool midas;
  		public bool ichor;
  		public bool onFire;
+@@ -284,6 +_,12 @@
+ 		private static int maximumAmountOfTimesLadyBugRainCanStack = 10 * ladyBugRainTime;
+ 		public static int offSetDelayTime = 60;
+ 
++		public NPC() { }
++
++		public NPC(int setDefaultsToType, NPCSpawnParams spawnparams = default(NPCSpawnParams)) {
++			SetDefaults(setDefaultsToType, spawnparams);
++		}
++
+ 		public bool CanTalk {
+ 			get {
+ 				if (isLikeATownNPC && aiStyle == 7 && velocity.Y == 0f)
 @@ -653,7 +_,7 @@
  		public static void UpdateFoundActiveNPCs() {
  			for (int i = 0; i < 200; i++) {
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -562,10 +575,11 @@
  					Main.npc[i].active = false;
  					NetMessage.SendData(23, -1, -1, null, i);
  				}
-@@ -59769,6 +_,7 @@
+@@ -59769,6 +_,8 @@
  			if (type == 551)
  				itemType = 3860;
  
++			//Left in for compatibility. It is now in ItemDropDatabase.RegisterBosses
 +			NPCLoader.BossBag(this, ref itemType);
```
Chicken-Bones, line 583: Is this advisable? Is this hook actually usable for anything here?

direwolf420: The method actually doesn't contain any of the new 1.4 bosses, for the record. And yeah I just left it in cause I didn't know if I should remove it or not, don't know what tmls policies on keeping legacy tml methods inside legacy vanilla methods is.

direwolf420: Actually I think it can be removed, given it is auto-registered AND auto-dropped through the drop handler. Mods used to have to call npc.DropBossBags() in NPCLoot manually, if they would still do it in 1.4 in the OnKill hook, bags would drop twice.

direwolf420: Upon further checking vanilla code regarding bags, I noticed one boss not utilizing the logic this PR would use: Eater of Worlds. Its in practically every respect unique, there's exceptions for it everywhere, no surprise this is also the case here. If a mod were to implement a boss with the same logic (any segment could potentially drop the bag, but only one bag should be dropped), it's easy to remove the drop rule from the NPC and insert a custom boss bag rule (the same one EoW uses).

direwolf420:
Closed due to opening a can of worms for other boss related content (trophies, masks, relics, pets).
Adding a boss bag is still about the same amount of code you had to write before 1.4

PR #1286, opened 2021-01-14
[1.4] ModifyGlobalLoot

kittenchilly:
I forgot this in the initial bestiary PR, but there is now a separate hook in GlobalNPC that allows you to modify global loot rules (loot rules that apply to literally every npc, like souls and biome keys). Also includes a small example in ExampleNPCLoot for how to use it.


Mirsario (changes requested):
There should be a way to enumerate and remove such drops, same as with normal NPCLoot.

kittenchilly:
> There should be a way to enumerate and remove such drops, same as with normal NPCLoot.

Vanilla doesn't have that, so I'd have to patch vanilla a bunch to get it to work that same as the normal NPCLoot. I don't think its really necessary anyways since drops are going to be conditional anyways (besides removing vanilla global loot, I guess)

Mirsario:
> 
> 
> > There should be a way to enumerate and remove such drops, same as with normal NPCLoot.
> 
> Vanilla doesn't have that, so I'd have to patch vanilla a bunch to get it to work that same as the normal NPCLoot. I don't think its really necessary anyways since most drops using this are going to be conditional. (Besides removing vanilla global loot, I guess)

I did it now with 1 patch, which changes protection of a field.

Mirsario (approved):

PR #1288, opened 2021-01-15
[1.4] Boss bar support; ExampleMod: Minion Boss

direwolf420:
## What is the new feature?
Adds three ModTypes pertaining to "boss health bars".

Also adds a full boss example of the "boss body with minions" archetype to better showcase advanced boss bars, as well as more advanced NPC control, drops, bestiary, and summoning.

Read my experimental PR (#1261) here for an explanation on the vanilla system.

Before elaborating on how they all work together, I'll first explain each addition in detail, and what changes to vanilla/TML had to be done to make it work.

#### Implementation and Features
This implements `IBigProgressBar` (what gives it updating and drawing capabilities). It is used to create custom boss health bars for modded and vanilla NPCs.
Just like the vanilla counterparts, a boss bar does not know which NPCs it is tied to. NPCs now have a `BossBar` property that modders can freely assign vanilla and modded bars to. In contrast to vanilla, this is not "set in stone" after load, but dynamic.

ModBossBar tracks displayed life/shield percentages for the boss bar for better mod compatibility (see ModBossBarStyle).

If the modder doesn't override `ModifyInfo`, TML assumes a "single-segment" NPC and automatically fetches the life%, reducing boilerplate for modders.

Because it's a ModTexturedType, an optional(!) texture with the same name is chosen as the base texture for the boss bar.
As long as the modder keeps the dimensions the same (matching the vanilla texture), they don't have to write any manual draw code.

Another convenient overridable method is `GetIconTexture`, letting you choose any texture as the icon for the boss bar (defaults to the "?" icon otherwise).
You can achieve the same by using the `PreDraw` hook, but I chose to make `GetIconTexture` as it fits the purpose most modders want to use boss bars for.

This class is convenient for modders to use, because it gives full freedom over everything related to the boss bar - 
Without having to write any code inside the NPCs themselves! And as a reminder: any NPC with a boss head index will automatically display a common boss bar shared among all simple bosses, so for a modder, it is not even mandatory to make a ModBossBar, only if you want special behavior.

#### Examples
* `ExampleBossBar`:
    * Basic entity boss bar focusing on drawing
    * Uses a custom colored texture
    * Assigns a custom icon
    * Modifies the draw position based on life%
 
* `MinionBossBossBar`:
    * Basic custom boss bar with code for life, icon, and shield
    * Assigns one of two icons based on what the boss dictates
    * Displays a shield% based on the life of remaining minions

#### Implementation and Features
This is used to modify existing boss health bars (vanilla and modded), mainly their drawing using `PreDraw` and `PostDraw`.
Implementation specifics are listed further below.

#### Examples
* `ExampleGlobalBossBar`:
    * If the bars' associated NPC is the Eye of Cthulhu:
        * Makes the icon rainbow colored
        * Shows "GlobalBossBar Showcase" in the center of the bar

#### Implementation and Features
Very similar to the ModMenu PR (#1043) in purpose and implementation, it allows modders to let the player choose between styles for boss bar functionality and drawing.
This feature is specifically tailored towards mods that fully change/replace the vanilla boss bar implementation (YABHB, Calamity, UI overhaul mods).

Modders can decide whether the ModBossBarStyle should piggyback off of vanilla updating and/or drawing through overridable properties `PreventUpdate` and `PreventDraw`.

If `PreventDraw` returns false, no GlobalBossBar hooks will run by default (except if the modder manually decides to call these methods).

Because vanilla doesn't have this concept of boss bar styles (in contrast to player health/mana styles and minimap styles), vanilla behavior has been turned into a class itself that simply allows updating and drawing.
A selectable option has been added to the "Interface" menu in both the main menu and ingame menu, handled the same way as ModMenu switching.

This is the most "code-heavy" feature from the modders' perspective, as they have to reimplement/call things themselves (but for the mods it's tailored towards, they already have the code required for this).

#### Examples
* `ExampleBossBarStyle`:
    * Showcases very basic code for a custom boss bar style that only changes "single-segment" boss bars
    * Lets special bars and modded bars through without changes

### Interactions
Update and draw code of the vanilla boss bar "style" is wrapped within checks of the currently selected ModBossBarStyle, meaning it could bypass the other two types' hooks completely (as explained in **3.**).
The part of vanilla update code that decides which bar to display first checks for modded bars, then vanilla.

PreDraw code runs in this logic: First GlobalBossBar, then ModBossBar, a single `return false` in any GlobalBossBar can prevent ModBossBar code from being ran. The `PostDraw` hook runs regardless of `PreDraw` returning false, but compared to other TML implementations, includes a "skipped" parameter to identify what `PreDraw` returned. The order is reversed here, so GlobalBossBar has the final draw.
This lets a GlobalBossBar know if a ModBossBar skipped vanilla drawing (for whatever reason), and if it wants to respect this or not.

TML-added draw methods all use a newly added `BossBarDrawParams` struct that contains draw parameters for `DrawFancyBar_TML` (explained further below), otherwise the method signatures would get extremely long. It also makes adding new parameters easier in the future.

### Vanilla Changes
The draw code adjustments are the most interesting and probably controversial parts (implementation wise):
How are mods able to hook into the vanilla drawing? By having TML intercept the method(s) vanilla uses to draw them.

Vanilla uses two helper methods to draw bars. These helper methods are identical, apart from one thing: the one with the `shieldPercent` parameter uses slightly different icon aligning maths.
This override is also only used by the draw code of the lunar pillars. To reduce TML code, I have decided to redirect the non-`shieldPercent` override to this one, which THEN redirects to a custom method.
If the _vanilla_ draw logic runs, TML sets a value to a nullable info variable, which is then checked inside the method(s) to identify if the call is made within a boss bar drawing context.
These methods only redirect the calls if this variable has a value, otherwise they don't (to preserve calls of these helper methods outside of this context).

The downside to this redirection stuff is that vanilla draws using the `shieldPercent` override now have the icon offset depending on the icon's size (if it's intercepted by TML).
In the case of lunar pillars, it's only offset by 2 pixels downwards, not noticeable without a side-by-side comparison. Phew.

The upside is the custom method `DrawFancyBar_TML`; it's located in BossBarLoader and is free to use, its biggest benefit being that it comes with a lot more customization options (the vanilla methods are very limited) that the draw hooks from the new classes are based on.
It only works with a fixed-size texture though (as explained in **1.**).

#### Minor Changes
* Made `IBigProgressBar` public
* Made `BigProgressBarSystem` partial and added a `TryGetSpecialVanillaBossBar` method that complements `BossBarLoader.NpcToBossBar`
* Exposed the private `NeverValidProgressBar` through a getter property so it can be used by modders
* Because of the in-game menu boss bar style insertion, the "Interface" menu is slightly tighter packed to accommodate for the new setting

## ExampleMod
Adding to the examples tailored towards showcasing the new classes, this PR also adds a full-fledged boss with "fairly basic" AI and fully documented code (albeit just recolored Eye of Cthulhu sprites).
For general use, it adds a `DownedBossWorld`; its purpose in ExampleMod should be maintaining downed boss booleans.

The boss consists of the body class, and the minion class. With it comes:
 * Bestiary entries for both, and drops for the body
 * A summon item
 * A boss bag
 * The aforementioned boss bar
 * Body gore
 * A hostile projectile

I eagerly await feedback on both the TML additions and the ExampleMod ones.


direwolf420:
Screenshots:
[In-game menu option](https://media.discordapp.net/attachments/571666551179051030/797942072198823946/Screenshot_5.png)
[Main menu option](https://cdn.discordapp.com/attachments/571666551179051030/798152396928188426/Screenshot_1.png)

[ExampleBossBarStyle](https://cdn.discordapp.com/attachments/571666551179051030/798157089745797130/Screenshot_5.png)

Videos:
[Minion Boss (Stage 1)](https://cdn.discordapp.com/attachments/445276626352209920/798695925085634560/boss1_bar.webm)
[Minion Boss (Stage 2)](https://cdn.discordapp.com/attachments/445276626352209920/798695934526881842/boss2.webm)

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
```diff
@@ -0,0 +1,323 @@
...
+			bool modify = true;
+
+			if (NpcToBossBar(npc.type, out ModBossBar bossBar)) {
+				drawParams.barTexture = GetTexture(bossBar).Value;
+				modify = bossBar.PreDraw(spriteBatch, npc, ref drawParams);
+			}
+
+			foreach(GlobalBossBar globalBossBar in globalBossBars) {
+				modify &= globalBossBar.PreDraw(spriteBatch, npc, ref drawParams);
+			}
```
Flashkirby, line 209: Would be worth rearranging this so GlobalBossBar is checked first, and only call ModBossBar if the global predraw returned true.

See implementation of mixed mod/global as seen in https://github.com/tModLoader/tModLoader/blob/5e311a223dc23b0ded123c5a0ae13c880528f477/patches/tModLoader/Terraria.ModLoader/NPCLoader.cs#L698

Flashkirby (commented):
Looks good for the most part, just a few bits I would want to double check!

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
Mirsario: Yes, globals should be able to have full control of ModX, and be able to prevent its hooks from running in cases like this one.

patches/tModLoader/Terraria/ModLoader/ModBossBar.cs
```diff
@@ -0,0 +1,125 @@
...
+		/// </summary>
+		public IReadOnlyList<int> ValidNPCs { get; internal set; } = new List<int>();
+
+		internal int index;
+
+		private float lifePercent;
+		public float LifePercent => lifePercent;
+
+		private float shieldPercent;
+		public float ShieldPercent => shieldPercent;
```
Mirsario, line 27: Again, use autoproperties.

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+	/// <summary>
+	/// A class that is used to swap out the entire boss bar display system with your own implementation
+	/// </summary>
+	public abstract class ModBossBarStyle : ModType
+	{
+		/// <summary>
+		/// The string that gets saved in the config to represent this ModBossBarStyle. Defaults to "TheMod/TheModBossBarStyle".
+		/// </summary>
+		public string SavedName => $"{Mod?.Name ?? "Terraria"}/{Name}";
```
Mirsario, line 14: "Defaults to"? Was this supposed to be virtual? If yes - why? If no - you can remove this and change ModType.FullName to have null-coalescing on Mod like here without introducing any problems.

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -394,7 +394,7 @@ internal static void LoadConfiguration()
 			Main.Configuration.Get("ShowModMenuNotifications", ref notifyNewMainMenuThemes);
 			Main.Configuration.Get("LastSelectedModMenu", ref MenuLoader.LastSelectedModMenu);
 			Main.Configuration.Get("KnownMenuThemes", ref MenuLoader.KnownMenuSaveString);
-
+			Main.Configuration.Get("BossBarStyle", ref BossBarLoader.lastSelectedStyle);
```
Mirsario, line 397:
```suggestion
			Main.Configuration.Get("BossBarStyle", ref BossBarLoader.lastSelectedStyle);

```

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
```diff
@@ -0,0 +1,323 @@
...
+
+			NPC npc = Main.npc[index];
+
+			if (NpcToBossBar(npc.type, out ModBossBar value))
+				value.PostDraw(skipped, spriteBatch, npc, drawParams);
+
+			foreach (GlobalBossBar globalBossBar in globalBossBars) {
+				globalBossBar.PostDraw(skipped, spriteBatch, npc, drawParams);
+			}
+		}
```
Mirsario, line 227:
```suggestion
		public static void PostDraw(bool skipped, SpriteBatch spriteBatch, BigProgressBarInfo info, BossBarDrawParams drawParams) {
			int index = info.npcIndexToAimAt;
			if (index < 0 || index > Main.maxNPCs)
				return; //Invalid data, abort

			NPC npc = Main.npc[index];

			foreach (GlobalBossBar globalBossBar in globalBossBars) {
				globalBossBar.PostDraw(skipped, spriteBatch, npc, drawParams);
			}

			if (NpcToBossBar(npc.type, out ModBossBar value))
				value.PostDraw(skipped, spriteBatch, npc, drawParams);
		}
```

patches/tModLoader/Terraria/ModLoader/ModBossBar.cs
```diff
@@ -0,0 +1,125 @@
...
+		public virtual void SetStaticDefaults() {
+		}
+
+		//abstract because otherwise it would be useless
+		/// <summary>
+		/// Allows you to specify a list of NPC types this ModBossBar is associated with. Has to contain atleast one type.
+		/// <para>If you need to fetch this list again, use this.ValidNPCs instead.</para>
+		/// </summary>
+		/// <returns>List of NPC types this ModBossBar is associated with</returns>
+		protected abstract IEnumerable<int> GetValidNPCs();
```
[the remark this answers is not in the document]
direwolf420: it is InitializeValidNPCs

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
direwolf420: why should Global be first in PostDraw? It's not like this for NPC or Item PostDraw

ExampleMod/Common/Systems/DownedBossSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+			Remember that Bytes/BitsByte only have 8 entries. If you have more than 8 flags you want to sync, use multiple BitsByte:
+				This is wrong:
+			flags[8] = downed9thBoss; // an index of 8 is nonsense. 
+				This is correct:
+			flags[7] = downed8thBoss;
+			writer.Write(flags);
+			BitsByte flags2 = new BitsByte(); // create another BitsByte
+			flags2[0] = downed9thBoss; // start again from 0
+			// up to 7 more flags here
+			writer.Write(flags2); // write this byte
```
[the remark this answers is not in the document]
direwolf420: The comments are copied from old examplemod, and yea I literally thought about suggesting bitarray 15 minutes ago

patches/tModLoader/Terraria/ModLoader/ModBossBar.cs
direwolf420: cant use ref with properties

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
direwolf420: copied from ModMenu. I was using ModType.FullName at first and it errored for the non-autoloaded DefaultBossBarStyle where i thought about the same approach you suggested, but didn't act on it

ExampleMod/Common/Systems/DownedBossSystem.cs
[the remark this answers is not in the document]
direwolf420: I think this it not worth the trouble for most people as it requires a bunch of boilerplate for sending/receiving and then reassigning the flags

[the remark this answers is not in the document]
direwolf420: Added an example anyway

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
direwolf420: Regarding this, Here is the code from ModMenu, including the comment
https://github.com/tModLoader/tModLoader/blob/34fd0def138e1edf2ebf8aa06646dd76955d08d2/patches/tModLoader/Terraria/ModLoader/ModMenu.cs#L19-L22

https://github.com/tModLoader/tModLoader/pull/1034 refers to the ModType PR, that got merged 10 days after the ModMenu one. The obvious problem with using ModType.FullName instead of whatever SavedName is getting currently is that since it's not autoloaded by anything, no mod belongs to it, so it throws.

https://github.com/tModLoader/tModLoader/blob/34fd0def138e1edf2ebf8aa06646dd76955d08d2/patches/tModLoader/Terraria/ModLoader/ModType.cs#L19-L21

As Mirsario pointed out, it could be changed for ModType, would it affect anything? @Rartrin

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
direwolf420: Mirsario told to ping you @Chicken-Bones. He proposes Global `PostDraw` to be before Mod `PostDraw`, unlike other draw methods in TML.

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
Rartrin: First off, the using "Terraria/Name" is only for vanilla things. We reserve it everywhere else for that and some things would likely break otherwise. Plus, it's always good to be consistent.

Secondly, the real question is why is Mod not getting set. It doesn't matter how it is being loaded. Regardless of if it is added through autoloading or through manual calling, `Load` must be getting called. We can't allow for ModTypes to be exist without a mod and frankly, there is no point in extending from ModType otherwise. If FullName is being called and Mod is null, throw an exception because it means something is likely running in the wrong hook.

direwolf420: Currently, there is a `VanillaBossBarStyle` class with `[Autoload(false)]`, and it is initialized in the static constructor of `BossBarLoader`, so it can be added directly to the list of boss bar styles. No registering takes place, so it's not visible to `ModTypeLookup`, nor are basic `ModType` variables properly set.

This design is a straight copy from ModMenu, and I *think* it is done like this to avoid having this style assigned to ModLoaderMod, which gets loaded and unloaded, all while main menu things are happening and possibly checking its value. After all, CB merged it, so he must've had a reason to think this is the correct design.

The current problem obviously is that if some mod would want to check CurrentStyle, and its Mod field, it could possibly be null if the style is vanilla (This same problem exists in the ModMenu implementation aswell).

I could try changing it tomorrow so it belongs to ModLoaderMod, and see if any issues arise.

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
Mirsario: I looked at NPCLoader once more and am retracting this. Haven't seen that there's already hooks in which globals are called afterwards on purpose, though I think you told me that.

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
Mirsario: `ModType.Mod` will be null in many cases related to vanilla dummies that never get registered and/or unloaded with mod reloading, and i.e. aren't associated with the internal TML mod. Like with instances of the internal VanillaPlayerDrawLayer and ModMenus. I will commit null coalescing on the main branch in a minute.

ExampleMod/Content/Items/Consumables/MinionBossSummonItem.cs
```diff
@@ -0,0 +1,67 @@
...
+using Terraria.ID;
+using Terraria.GameContent.Creative;
+using Terraria.ModLoader;
+using ExampleMod.Content.NPCs.MinionBoss;
+
+namespace ExampleMod.Content.Items.Consumables
+{
+	//This is the item used to summon a boss, in this case the modded Minion Boss from Example Mod. For vanilla boss summons, see PlanteraItem.cs
+	public class MinionBossSummonItem : ModItem
+	{
```
Mirsario, line 12: May wanna remove that plantera summoning item we already had.

ExampleMod/Content/NPCs/MinionBoss/MinionBossBody.cs
```diff
@@ -0,0 +1,473 @@
...
+			else {
+				DoFirstStage(player);
+			}
+		}
+
+		private void SpawnMinions() {
+			if (!SpawnedMinions) {
+				SpawnedMinions = true;
+
+				if (Main.netMode != NetmodeID.MultiplayerClient) {
```
Mirsario, line 286: Short-circuit to reduce indentation more frequently.

ExampleMod/Content/NPCs/MinionBoss/MinionBossBody.cs
```diff
@@ -0,0 +1,473 @@
...
+using Terraria.GameContent.Bestiary;
+using Terraria.GameContent.ItemDropRules;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.NPCs.MinionBoss
+{
+	//The main part of the boss, usually refered to as "body"
+	[AutoloadBossHead] //This attribute looks for a texture called "ClassName_Head_Boss" and automatically registers it as the NPC boss head icon
+	public class MinionBossBody : ModNPC
```
Mirsario, line 20: Great documentation.

patches/tModLoader/Terraria/GameContent/UI/BigProgressBar/BigProgressBarSystem.TML.cs
```diff
@@ -0,0 +1,9 @@
+namespace Terraria.GameContent.UI.BigProgressBar
+{
+	public partial class BigProgressBarSystem
+	{
+		//If any mods need to fetch special vanilla bars
+		public bool TryGetSpecialVanillaBossBar(int netID, out IBigProgressBar bossBar) =>
+			_bossBarsByNpcNetId.TryGetValue(netID, out bossBar);
```
Mirsario, line 7: This is much more preferable
```suggestion
		//If any mods need to fetch special vanilla bars
		public bool TryGetSpecialVanillaBossBar(int netID, out IBigProgressBar bossBar)
			=> _bossBarsByNpcNetId.TryGetValue(netID, out bossBar);
```

Mirsario (commented):
Are NPCs able to assign bars for themselves outside of ModBossBar?

ExampleMod/Content/BossBars/MinionBossBossBar.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+	//Showcases a non-basic custom boss bar
+	public class MinionBossBossBar : ModBossBar
+	{
+		private int bossHeadIndex = -1;
+
+		protected override IEnumerable<int> InitializeValidNPCs() {
+			return new List<int> { ModContent.NPCType<MinionBossBody>() };
+			//We don't include the minion here as it's not an essential "segment" or body part of the boss itself (unlike Twins, Moonlord, Golem etc.)
+		}
```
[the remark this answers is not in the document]
direwolf420: Changed to array, yield return is not as beginner friendly, and doesn't make much sense as it's only called once and the result is cached in the `ReadOnlyList<int> ValidNPCs` property.

ExampleMod/Content/Items/Consumables/MinionBossSummonItem.cs
direwolf420: I added a few comments to `MinionBossSummonItem`, making the plantera summoning item obsolete. Can be removed when this is merged.

ExampleMod/Content/NPCs/MinionBoss/MinionBossBody.cs
direwolf420: got rid of two indentations

patches/tModLoader/Terraria/GameContent/UI/BigProgressBar/BigProgressBarSystem.TML.cs
direwolf420: adjusted and added docs aswell

direwolf420:
> Are NPCs able to assign bars for themselves outside of ModBossBar?
@Mirsario No, and I see no point in this. There is only one vanilla bar worth assigning, and its the `NeverValid` one that simply doesn't display anything, it can be reimplemented by the modder within less than 10 lines. The `CommonBigProgressBar` is "assigned" by default (if NPC has a boss head index, and no special vanilla or modded bars). Almost all the other special vanilla bars are tailored towards the specific vanilla NPCs and will not work with any other NPCs.

I think it's also cleaner not to have any code spread between two files.

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
Rartrin: Then we should solve the issue in an object oriented way rather than hard coding it. `FullName` is working as intended, the problem is that Mod is null.

patches/tModLoader/Terraria/ModLoader/Default/DefaultBossBarStyle.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Terraria.ModLoader.Default
+{
+	/// <summary>
+	/// This is the default boss bar style - the way vanilla draws boss bars.
+	/// </summary>
+	[Autoload(false)]
+	internal class DefaultBossBarStyle : ModBossBarStyle
```
Solxanich, line 7: This class feels unnecessary at a glance - what is its purpose?
If it needs to exist temporarily, why can't it use a constructor to set name, and be discarded when done?
Alternatively, if it is just showing how to get a quick vanilla style bar, wouldn't that fit EM better?

ExampleMod/Content/BossBarStyles/ExampleBossBarStyle.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace ExampleMod.Content.BossBars
+{
+	//Showcases very basic code for a custom boss bar style that is selectable in the menu in "Interface"
+	public class ExampleBossBarStyle : ModBossBarStyle
+	{
+		public override bool PreventDraw => true; //Prevents the vanilla bars from drawing
+
+		public override void Draw(SpriteBatch spriteBatch, IBigProgressBar currentBar, BigProgressBarInfo info) {
+			if (currentBar != null) {
```
Chicken-Bones, line 14: early return please

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
```diff
@@ -0,0 +1,326 @@
...
+
+		/// <summary>
+		/// Matches NPC types to their ModBossBar
+		/// </summary>
+		internal static void BuildReverseLookup() {
+			for (int i = 0; i < bossBars.Count; i++) {
+				ModBossBar bossBar = bossBars[i];
+				var validNPCs = bossBar.ValidNPCs;
+
+				//The same NPC can exist in multiple bars. In that case, only take the first found occurence
```
Chicken-Bones, line 159: this seems problematic? Is there any way for a player to select which boss bar they want for a given NPC? At the very least there should be a log message.

Chicken-Bones (approved):
Overall approve, I didn't dive too deep. Have a look at my comments.

patches/tModLoader/Terraria/ModLoader/Default/DefaultBossBarStyle.cs
direwolf420: This is internal to tml, and is required to be in line with how ModBossBarStyle is managed in the menus. (And it's a mirror of the ModMenu design). An example for a ModBossBarStyle exists in EM.

direwolf420:
Reworked how the NPC-to-Boss Bar relation works.

Added `public IBigProgressBar BossBar { get; set; }` to NPC.TML.cs: now modders can simply assign a bar (vanilla or modded) anywhere (usually `SetDefaults`). In the process, eliminated any "static" code that was handling picking boss bars (`ModBossBar.InitializeValidNPCs` for example).

(I am not sure if adding this property to that file is the proper way, should it be a field instead, maybe it should be in NPC.cs?)

patches/tModLoader/Terraria/ModLoader/BossBarLoader.cs
direwolf420: Resolved in https://github.com/tModLoader/tModLoader/pull/1288/commits/e66364ddb45fdd7340b7ffe07f9fafb438288963

patches/tModLoader/Terraria/ModLoader/Default/DefaultBossBarStyle.cs
Solxanich: Thanks for clarifying, more of just questioning how its used - 
Looking at what you pointed at, it's used as a 'backing' correct? 

Then to reiterate my question of constructors, keeping in mind this comment chain ( https://github.com/tModLoader/tModLoader/pull/1288#discussion_r561351794) why are we not using a constructor such as set `string mod ="VanillaSpoof"` in the default style so FullName and TryFind() work? 
Or, for that matter, actually directly change its full name so it works without using a constructor if its always the same "fake" mod?

patches/tModLoader/Terraria/ModLoader/ModBossBarStyle.cs
direwolf420: Due to tml commit https://github.com/tModLoader/tModLoader/commit/f79079253b05c0d734839fb212eb234fd480e508, can this now be resolved? @Rartrin

Chicken-Bones (approved):

Mirsario (approved):

PR #1292, opened 2021-01-19
[1.4] ModGeneration, new worldgen hooks

kittenchilly:
I guess you could consider this a continuation of my previous worldgen PR which I think was too much. So I closed it and completely remade some of what it had by scratch, plus more, especially with the advent of ModSystems.
1. New ModGeneration class, all the previous world gen hooks were moved here from ModSystems.
2. ~~Implement a native version of Large World Enabler. The system used may actually improve performance on normal worlds as well.~~ Split off into #1505 
3. Three new world generation hooks: SetWorldGenDefaults, ModifyMeteorTasks, and ModifyAltarTasks.
4. ExampleMod examples for the new class

If you're wondering about OOP based worldgen, I really don't think its possible without ripping out vanilla code. All the pieces just aren't there unlike how playerlayers were. GenPasses are created during worldgen, some gen steps dont even use genpasses, etc.


Mirsario (changes requested):
Great first steps, but I think you misunderstood some things about the suggested design in discord. In it, we won't have these hardcoded (i.e. repeated and not generalized) ModifyXTasks hooks. Instead, all these vanilla generations would be static instances of a `VanillaGeneration` class, which will derive from `ModGeneration` (I don't like the Mod prefix for this reason, `PlayerDrawLayer` doesn't have it), exactly like with `VanillaPlayerDrawLayer`, and static fields of its instances in `PlayerDrawLayers.cs`. `ModGeneration` should instead have a hook for just declaring its initial roster of GenPasses. We can then add a way to add or disable genpasses from the outside, or make a GlobalGeneration class for dealing with that. I don't know which is better. But there should definitely be topological sorting seen in player layers involved in dealing with genpasses.

kittenchilly:
This is so old I'm just gonna close it.

PR #1295, opened 2021-01-23
Fixed DamageClass throwing an exception before the game can even start

ThomasSpeedrunner:
Title says it all. Somehow, a null check was added to the DamageClass PR right before it got merged which caused the game to throw an error before even starting properly, preventing all forms of actually opening the game. This PR fixes this.

### DIFF
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -17,11 +17,13 @@ namespace Terraria
 
 		private DamageClass _damageClass = DamageClass.Generic;
 		/// <summary>
-		/// The damage type of this Item. Assign to DamageClass.Melee/Ranged/Magic/Summon/Throwing for vanilla classes, or ModContent.GetInstance<T>() for custom damage types.
+		/// The damage type of this Item.
+		/// Assign to DamageClass.Melee/Ranged/Magic/Summon/Throwing for vanilla classes, or ModContent.GetInstance<T>() for custom damage types.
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -17,11 +17,13 @@ public partial class Item : TagSerializable
...
 		/// <summary>
-		/// The damage type of this Item. Assign to DamageClass.Melee/Ranged/Magic/Summon/Throwing for vanilla classes, or ModContent.GetInstance<T>() for custom damage types.
+		/// The damage type of this Item.
+		/// Assign to DamageClass.Melee/Ranged/Magic/Summon/Throwing for vanilla classes, or ModContent.GetInstance<T>() for custom damage types.
+		/// Set to DamageClass.Generic to not have any one damage class, but still have the weapon be boosted by universal bonuses.
 		/// </summary>
 		public DamageClass DamageType {
 			get => _damageClass;
-			set => _damageClass = value ?? throw new ArgumentException("DamageType cannot be null");
+			set => _damageClass = value;
```
Mirsario, line 26: This isn't a fix, this is a rollback.

Mirsario (changes requested):

patches/tModLoader/Terraria/Item.TML.cs
Mirsario: The actual fix would be to not autoload the default classes with the default mod. Why are they done so..

Mirsario: image

Mirsario:
Fixed this.

PR #1299, opened 2021-01-25
[1.4] Accessory Slot Creation Rework

Solxanich:
### What is the new feature?
Credit to ModernMak for initially working on cleaning up the accessory code, which gave me the idea to actually make extra accessory slots easier for modders.
Well, currently adding accessory slots is not a clean affair. Lot's of cloning, netmessaging, drawing, and that just gets you the functional slot. 
It ends up being a lot of boilerplate code, doing the same stuff in multiple mods.
Ignoring that, not all mods cover all functionality that a slot has, vanilla does some fun stuff thats easy to miss, such as post load fixes.
The ability to retrieve your items when you unload the mod is lost.
Finally, we have the issue of UI cluttering from such slots. 

This PR aims to fix all of that by making a ModAccessorySlot ModType. An example of the player facing UI behavior is as follows (1 custom location slot, and two additional standard drawn slots): 

https://user-images.githubusercontent.com/59670736/133655350-f178d1a4-f9ff-478c-9899-71d6ff909e7c.mp4

### Example Usage:
See ExampleMod ExampleAccessorySlots for usage such as:
image

 ### Concerns:
The interdependency of AccessorySlotLoader.list.Count & ModAccessorySlotPlayer array.Length while dealing with multiple different player characters with varying unloaded slots may result in errors in Multiplayer. 
Non-trivial to test locally, so this particular scenario is untested and possible problem.
There are possible solutions, however, would prefer to avoid implementing either unless this is an actual problem and can see what in particular is the breaking point to select the appropriate fix path around.


Chicken-Bones (commented):
tl;dr: skip to the last paragraph.

Thanks for making this PR as a partially implemented proposal. This lets us evaluate the maintenance impacts of your suggested approach, and discuss actual code. Let me start by saying that I'm definitely in favour of tML providing an easy mechanism to add extra accessory slots. The picture you've provided is exactly what we should hope to achieve with this PR. Please see the individual code review comments below.

While you've noted that save/load and hooks for 'customisation' of slots are not yet implemented, we need to discuss these to settle on a design for this feature. The current hook design does allow mods to create their own slots, but they must handle rendering, storage, networking and interactions, replicating a lot of vanilla functionality. 

Normally, if a mod adds an accessory slot, and the mod is unloaded, the accessory in it is inaccessible until the mod is restored. With tML handling this feature, we can save accessories centrally, while retaining which mod owns their 'slot', and if the mod is disabled (or the slot removed) then we can use the vanilla 'no effect' rendering to allow players to remove their accessories from the missing slots and then have the slots disappear.

I also expect mods to want to restrict what can be placed in their slots, or violate certain normal rules (like allowing duplicate accessory use, or making a slot which applies effects twice).

With this in mind, the number of accessory slots cannot simply be a public variable on player. Mods must declare (possibly via a `ModEquipSlot` class) their slots, for tML to internally sort, render, store and network. 

The approach here, of storing mod accessory slots in the same arrays as vanilla, requires changing many constants, and hooking many IO and render locations, incurring significant maintenance cost. Additionally, mods are still going to want to create 'special' slots outside these arrays, and I fear that relying too heavily on vanilla code paths may leave them out on interoperability.

Please see my mod https://github.com/Chicken-Bones/UtilitySlots (which I updated just for this review), for an example of exactly what building a current equipment slot entails. UtilitySlots is a little bit special in that it adds its own 'page', but it should illustrate what's currently difficult and missing. In particular, note that an `On` hook is required for `Player.UpdateVisibleAccessories` and that there's currently no way interact with the fast equip feature (`ArmorSwap`) without an IL hook.

I think the best approach will actually be for tML to implement the extra slots using available hooks, and the builtin `ModLoader` mod. This will mean minimal patches, and that tML is using the same API to create the shared pool of 'plain extra accessory slots' as mods can use to make more specialised content.

Solxanich:
Wow, thank you for the expedited review! I wasn't expecting a review for a bit as I had labeled this draft and was still cleaning up things here and there, as well as checking for items missing in description. Glad to hear their is an useful outcome here, although design is up for discussion.

I guess to address in order, I'll start by adding the info on what I was thinking for part 2 that I clearly missed adding to original desc. Part 2 I was thinking would:
1) Swap the integer to a proper tagcompound list with data:
A) index of assigned slot -> so that a mod can override and custom draw of the slot, as well as modify slot using existing hooks
B) name of mod -> so can check against and not display when unloaded
C) a pointer to the mod's player class to obtain data such as ShouldDisplay for character and Custom Conditions on placement -> but I suspect I'd have to implement differently this to be cleaner than initial idea
2) Add a single, unified load method and single save method to handle the additional slots data, instead of all overall the place like vanilla is
3) check code to ensure functionality of two above is fully implemented.

For particular code, agree with all flagged stuff; missed how we handle vanilla disabling, so I'll clean that up if continuing on this design path.

For the comment, by paragraph:
1) Agree - save and load is important. My thoughts are tML should self manage saving and loading of additional slots in terms of how many, and what items were there. Custom conditions, custom draws should be handled by the mod imo
2) See part 2 notes (again, sorry, should've waited before creating PR so that I could spend time ensuring all notes are on it instead of sleeping on the draft)
3) Been basing initial thinking off of WingSlot. So yes, see part 2 notes, 1c. Hadn't thought about intention break normal rules, so good catch on that.
4) I'll need to think on this when it comes to custom slots. In terms of maintenance, agree lots of number swaps to variables, although not sure maintenance created here. The assumption I'm making is most of vanilla code is basically frozen, never going to change, at this point so I'm not sure maintenance in terms of updating with Vanilla is something I'd consider personally. 
5) I'll take a look at your mod. Thank you for updating it; good to have a relevant comparison
6) Yeah, alternatives definitely exist. I'll take a look at going the modloader route with its own management of slots instead; I imagine it will still accomplish general same idea, although with a bit longer blocks of custom tML code instead of many small vanilla patches.

Tl;dr I jumped the gun on posting PR without verifying all details presented.
Also think we should review our approach to vanilla patches given vanilla code is hard-frozen now (exempting steam workshop and related minor changes)

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2423,6 +2432,130 @@
...
+ 				int num39 = -1;
+-				for (int num40 = 0; num40 < 10; num40++) {
+-					if ((num40 == 8 && !flag4) || (num40 == 9 && !flag5))
++				int modRowCorrect = 0;
++				for (int num40 = 0; num40 < player[myPlayer].totalEquipSlots; num40++) { // Added support for modded accessory slots; does not support custom draw conditions natively here
++					if ((num40 == player[myPlayer].demonHeartSlot && !flag4) 
++						|| (num40 == player[myPlayer].masterModeSlot && !flag5))
+ 						continue;
++					if (num40 == player[myPlayer].vanillaEquipSlots) // begin a new column for modded accessory slots, with functionals being unique
++						modRowCorrect = num39 + 1;
```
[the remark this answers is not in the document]
Solxanich: I actually disagree with being in a separate column, despite implementing as such, as it leads to more unnecessary patches relative to the intended feature (namely modifying draws for defense, multiplayer team selections, etc). 
After seeing your UtilitySlots mod, I think it best to implement as a new page, and maybe columns subsequent from there. 
Further, we shouldn't just continually stack downwards in vanilla fashion, as that will inevitably lead to slots rendering offscreen, and thus vanilla's Purple Screen of Render Error. (For example, modSlots >= 11 will purple screen as I don't do more than one column...)

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -48,15 +48,42 @@
...
++		public int vanillaEquipSlots => vanillaEXAccSlots + vanillaBasicAccSlots + totalArmorSlots;
++		public int totalEquipSlots => moddedEXAccSlots + vanillaEquipSlots;
++		public int vanillaContainerSlots = 10;
++		public int vanillaHotbarSlots = 40;
++		public int vanillaCoinAmmoSlots = 8;
++		public int vanillaInventorySlots => vanillaHotbarSlots + vanillaCoinAmmoSlots + vanillaContainerSlots;
++		public int moddedInventorySlots = 0;
++		public int totalInventorySlots => moddedInventorySlots + vanillaInventorySlots;
++		//public bool extraAccessory; -> renamed to demonHeartUsed to avoid confusion
++		public bool demonHeartUsed;
```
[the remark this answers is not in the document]
Solxanich: Agreed - this goes back to the same point Can"InsertVanilla"AccessoryBeShown

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -1244,26 +1244,31 @@ private class HookList
 		/// Returns the wing item that the player is functionally using. If player.wingsLogic has been modified, so no equipped wing can be found to match what the player is using, this creates a new Item object to return.
 		/// </summary>
 		public static Item GetWing(Player player) {
```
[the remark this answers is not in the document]
Solxanich: I'll double check it does nothing otherwise in the player.cs, but yeah, when looking at it, it's very redundant and overly-complicated. 
I'll look to remove it independent of design path.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -48,15 +48,42 @@
...
++		public int vanillaEXAccSlots = 2;
++		public int vanillaBasicAccSlots = 5;
++		public int totalArmorSlots = 3;
++		public int vanillaEquipSlots => vanillaEXAccSlots + vanillaBasicAccSlots + totalArmorSlots;
++		public int totalEquipSlots => moddedEXAccSlots + vanillaEquipSlots;
++		public int vanillaContainerSlots = 10;
++		public int vanillaHotbarSlots = 40;
++		public int vanillaCoinAmmoSlots = 8;
++		public int vanillaInventorySlots => vanillaHotbarSlots + vanillaCoinAmmoSlots + vanillaContainerSlots;
++		public int moddedInventorySlots = 0;
```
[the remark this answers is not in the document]
Solxanich: I'll remove this. Agreed that it just makes things more complicated when too many unnecessary things included.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -48,15 +48,42 @@
...
++		public int masterModeSlot => totalArmorSlots + vanillaBasicAccSlots + 1;
  		private bool dontConsumeWand;
  		public int tankPet = -1;
+ 		public bool tankPetReset;
+@@ -683,7 +_,7 @@
+ 		public const int miscSlotHook = 4;
+ 		public const int SupportedMiscSlotCount = 5;
+ 		public Item[] armor = new Item[20];
+-		public Item[] dye = new Item[10];
++		public Item[] dye = new Item[20];
```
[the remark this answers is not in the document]
Solxanich: This is a Homer Simpson "Doh!".
Wonder when I changed that; I'll fix that to not change

Chicken-Bones:
> Tl;dr I jumped the gun on posting PR without verifying all details presented.
As mentioned before, this was a good state to create a draft in and get feedback.

> The assumption I'm making is most of vanilla code is basically frozen, never going to change, at this point so I'm not sure maintenance in terms of updating with Vanilla is something I'd consider personally.
> Also think we should review our approach to vanilla patches given vanilla code is hard-frozen now (exempting steam workshop and related minor changes)
This is not currently the stance of the tML dev team. Even if we could guarantee that the vanilla code was going to be hard frozen, improvements to ILSpy and Roslyn based reformatters are still on the table. Even if we were more open to sweeping changes, we prefer keeping the original vanilla code for comparison, and creating new code paths which better model the goals of tML. That and vanilla code is often a fragmented mess and difficult to reason about.

patches/tModLoader/Terraria/Main.cs.patch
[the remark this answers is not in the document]
Chicken-Bones: I'm biased against a separate page, but looking at the vanilla UI there's really no space there. Even with another page you'll only get 7 more slots though. I think the best idea is to do what you've done with the stacking up of new accessory slots in another column, and just moving the MP icons and armor around

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1119,8 +1160,57 @@
  			}
  
  			if (whoAmI == Main.myPlayer && luckPotion != oldLuckPotion) {
-@@ -8274,12 +_,19 @@
- 			}
+@@ -8252,34 +_,49 @@
+ 		}
+ 
+ 		public bool IsAValidEquipmentSlotForIteration(int slot) {
++			/*
```
Solxanich, line 1167: I'll need to fix this

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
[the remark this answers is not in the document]
Solxanich: I haven't removed this method as it is used in horizontal and vertical wing speed fetches in ItemLoader. For now, I don't touch ItemLoader.cs, so I don't think I'll tackle that here.

patches/tModLoader/Terraria/ModLoader/Default/DefaultPlayer.cs
```diff
@@ -0,0 +1,183 @@
+using System.Linq;
+using Terraria.ModLoader.IO;
+using System;
+using System.IO;
+
+namespace Terraria.ModLoader.Default
+{
+	public class DefaultPlayer : ModPlayer
```
Chicken-Bones, line 8: `ModAccessorySlotPlayer` please

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -80,12 +87,12 @@ public static class EquipLoader
 			//Sets
 			LoaderUtils.ResetStaticMembers(typeof(ArmorIDs), true);
 			WingStatsInitializer.Load();
-			
+
```
Chicken-Bones, line 90: accidental removal of spaces?

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -298,5 +305,487 @@ public static class EquipLoader
...
 		}
+
+		internal static ModAccessorySlot GetModAccessorySlot(int slot) {
+			ModContent.TryFind<ModAccessorySlot>(ModAccessorySlot.moddedAccSlots[slot], out ModAccessorySlot mAccSlot);
+			if (mAccSlot == null) {
+				mAccSlot = new UnloadedAccessorySlot(slot);
+			}
+
+			return mAccSlot;
+		}
```
Chicken-Bones, line 316: I think a different class would be better for these methods. The role of `EquipLoader` is just mapping item ids to equip types and ids. Mod accessory slot loading/hooks should go in a new class.

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+//NOTE: Does not fully support: ItemLoader.GetGamepadInstructions and related gamepad stuff.
+
+namespace Terraria.ModLoader
+{
+	/// <summary>
+	/// A ModAccessorySlot instance represents a net new accessory slot instance. You can store fields in the ModAccessorySlot class. 
+	/// </summary>
+	public abstract class ModAccessorySlot : ModType
+	{
+		internal virtual bool skipRegister => false;
```
Chicken-Bones, line 14: needs a better name? Something about not being part of the vanilla slot list, and being manually rendered/positioned by a mod?

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+		internal static int scrollbarSlotPosition = 0;
+
+		// List of registered identifiers for modded accessory slots. Used in DefaultPlayer.
+		internal static List<string> moddedAccSlots = new List<string>();
+
+		/// <summary>
+		/// Called when the ModAccessorySlot is registered. Registers to parent list.
+		/// Requests for a modded Slot, and stores assigned Slot. DefaultPlayer will auto-manage the items themselves.
+		/// </summary>
+		internal protected void Initialize() {
```
Chicken-Bones, line 30: `private`?
`internal` allows accesses from tML outside the `ModAccessorySlot` class
`protected` allows accesses from subclasses

Just merge this into `Register`

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+		/// <summary>
+		/// This function allows for custom textures and colours to be drawn for the accessory slot. Called for Dyes, Vanity, and Functionals.
+		/// Runs in place of ItemSlot.Draw() if this returns true.
+		/// Receives data:
+		/// <para><paramref name="inv"/> :: the array containing all accessory slots, yours is inv[slot] </para>
+		/// <para><paramref name="slot"/> :: which is the index for inventory that you were assigned </para>
+		/// <para><paramref name="position"/> :: is the position of where the ItemSlot will be drawn </para>
+		/// <para><paramref name="context"/> :: 12 => dye; 11 => vanity; 10 => functional </para>
+		/// </summary>
+		public virtual bool DrawModded(Item[] inv, int context, int slot, Vector2 position) {
```
Chicken-Bones, line 63: normally I'd say call this `CustomDraw` or something, but actually, just move the `ItemSlot.Draw` call from `DrawRedirect` in here, and then if mods override this, they'll be specifying custom behaviour. Just like inheritance was meant to be :)

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+			return false;
+		}
+
+		/// <summary>
+		/// Override to set conditions on when the slot is available, Example: the demonHeart is consumed and in Expert mode in Vanilla.
+		/// </summary>
+		public virtual bool CanUseSlot() {
```
Chicken-Bones, line 70: `IsSlotAvailable`? perhaps as a property? summary should mention that this controls _visibility_ and that if the condition is false but an item was previously inserted, the slot will disappear when the item is removed.

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+
+		internal int slot;
+		internal int index;
+
+		// Setting toggle for stack or scroll accessories/npcHousing
+		internal static bool scrollSlots = true;
+		internal static int scrollbarSlotPosition = 0;
+
+		// List of registered identifiers for modded accessory slots. Used in DefaultPlayer.
+		internal static List<string> moddedAccSlots = new List<string>();
```
Chicken-Bones, line 24: put this in ~`EquipLoader`~ `ModAccessorySlotLoader`

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,83 @@
...
+		public virtual bool CanUseSlot() {
+			return true;
+		}
+
+		/// <summary>
+		/// Override to set conditions on what can go in slot. Return false to prevent the item going in slot. Return true for dyes, if you want dyes. Example: only wings can go in slot.
+		/// Receives data:
+		/// <para><paramref name="checkItem"/> :: the item that is attempting to enter the slot </para>
+		/// </summary>
+		public virtual bool LimitWhatCanGoInSlot(Item checkItem) {
```
Chicken-Bones, line 79: `CanAcceptItem`?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1491,13 +1482,13 @@
...
  				if (armor[k].prefix == 73)
-@@ -9264,6 +_,22 @@
+@@ -9264,7 +_,25 @@
  
  				if (armor[k].prefix == 80)
  					meleeSpeed += 0.04f;
 +
 +				ItemLoader.UpdateEquip(armor[k], this);
-+			}
+ 			}
```
Chicken-Bones, line 1491: accidental indentation change? or fix?

patches/tModLoader/Terraria/UI/ItemSlot.cs.patch
```diff
@@ -36,6 +36,29 @@
  			if (Main.cursorOverride == 2) {
  				if (ChatManager.AddChatText(FontAssets.MouseText.Value, ItemTagHandler.GenerateTag(item), Vector2.One))
  					SoundEngine.PlaySound(12);
+@@ -425,11 +_,11 @@
+ 			Player player = Main.player[Main.myPlayer];
+ 			bool flag = Main.mouseLeftRelease && Main.mouseLeft;
+ 			if (flag) {
+-				if (OverrideLeftClick(inv, context, slot))
++				if (OverrideLeftClick(inv, Math.Abs(context), slot))
```
Chicken-Bones, line 44: needs a comment

Chicken-Bones (changes requested):
Looking good so far. I'll need to take another pass through the actual player and equip slot logic changes. 

Also: how does ModAccessorySlot work with a custom draw position/condition (eg for UtilitySlots or a mod that wants a special slot somewhere). Can we get an example?

patches/tModLoader/Terraria/Player.cs.patch
Solxanich: Fix, I believe. Instead of us adding the closing '}', we now use the existing vanilla '}'.
I'll take another look for any that aren't a fix.

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
Solxanich: Hmm. This is a swap from somebody hitting tab 3+ times to just a new line.
Accidental, probably, but I think is better to use just a new line and not new line + multiple tabs.

ExampleMod/Common/Players/ExampleAccessorySlot.cs
```diff
@@ -0,0 +1,35 @@
...
+	public class ExampleModAccessorySlot1 : ModAccessorySlot
+	{
+		// If the class is empty, everything will default to a basic vanilla slot.
+	}
+
+	public class ExampleModAccessorySlot2 : ModAccessorySlot
+	{
+		// If the class is empty, everything will default to a basic vanilla slot.
+	}
+
```
Solxanich, line 18: Add CustomDrawAccessorySlot : ModAccessorySlot with overriding XLoc, YLoc and DrawModded

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,69 @@
...
+			if (pendingID < 0) {
+				pendingID = AccessorySlotLoader.moddedAccSlots.Count;
+				AccessorySlotLoader.moddedAccSlots.Add(this.FullName);
+			}
+
+			this.slot = pendingID;
+		}
+
+		/// <summary>
+		/// This function allows for custom textures and colours to be drawn for the accessory slot. Called for Dyes, Vanity, and Functionals.
```
Solxanich, line 40: Double check the comments to ensure are accurate

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -0,0 +1,510 @@
...
+			Main.spriteBatch.Draw(value4, new Vector2(xLoc2, yLoc2), Microsoft.Xna.Framework.Color.White * 0.7f);
+
+			Microsoft.Xna.Framework.Rectangle rectangle = new Microsoft.Xna.Framework.Rectangle(xLoc2, yLoc2, value4.Width, value4.Height);
+			if (!(rectangle.Contains(new Microsoft.Xna.Framework.Point(Main.mouseX, Main.mouseY)) && !PlayerInput.IgnoreMouseInterface)) {
+				return;
+			}
+
+			Player player = Main.LocalPlayer;
+			player.mouseInterface = true;
+			player.mouseInterface = true;
```
Solxanich, line 95: Uhh, double take?

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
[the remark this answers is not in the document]
Solxanich: Resolved by storing "equippedWings" on the player at same time as setting winglogic.

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,86 @@
...
+		// Properties to control which parts of accessory slot draw
+		public virtual bool DrawFunctionalSlot => true;
+		public virtual bool DrawVanitySlot => true;
+		public virtual bool DrawDyeSlot => true;
+
+		protected sealed override void Register() {
+			type = Loader.Register(this);
+		}
+
+		private bool MySlotContainsAnItem() {
```
Chicken-Bones, line 42: Suggestion, invert this and call it `IsEmpty`
Can use the suggested properties above.
Should probably also check dye slot.

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,86 @@
...
+	/// <summary>
+	/// A ModAccessorySlot instance represents a net new accessory slot instance. You can store fields in the ModAccessorySlot class. 
+	/// </summary>
+	public abstract class ModAccessorySlot : ModType
+	{
+		internal int type;
+		internal int index;
+		public AccessorySlotLoader Loader => LoaderManager.Get<AccessorySlotLoader>();
+
+		// Properties to preset a location for the accessory slot
```
Chicken-Bones, line 19: There should probably be get+set properties for `Item`, `VanityItem`, `Dye` and `ShowVisuals`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1672,7 +1685,7 @@
  		}
  
  		private void ApplyEquipVanity(int itemSlot, Item currentItem) {
-+			ApplyEquipVanity(currentItem); //remove itemSlot parameter and make public so mods can call from their own accessory slots
++			ApplyEquipVanity(currentItem);
```
Chicken-Bones, line 1688: accidental comment revert?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1644,11 +1656,12 @@
...
  			}
  
-@@ -9334,6 +_,7 @@
+@@ -9334,6 +_,8 @@
  				maxTurrets++;
  			}
  
-+			/* wing loop is merged into ApplyEquipFunctional
++			/*
++			// wing loop is merged into ApplyEquipFunctional
```
Chicken-Bones, line 1664: accidental comment formatting change?

patches/tModLoader/Terraria/UI/ItemSlot.cs.patch
```diff
@@ -260,40 +412,145 @@
...
+ 			if (dyeSlotCount < 0)
+ 				dyeSlotCount = 9;
  
-@@ -2008,6 +_,9 @@
- 						num3 = k;
+@@ -1974,43 +_,70 @@
+ 				player.armor[num + 2] = item.Clone();
+ 			}
+ 			else if (item.accessory) {
++				//TML: Rewrote ArmorSwap for accessories under the PR #1299 so it was actually readable. No vanilla functionality lost in transition
```
Chicken-Bones, line 460: preferred approach to a whole method rewrite is to make another method and call it while commenting out the vanilla impl

patches/tModLoader/Terraria/ModLoader/Default/UnloadedAccessorySlot.cs
```diff
@@ -0,0 +1,25 @@
...
+		internal UnloadedAccessorySlot(int slot, string oldName) {
+			type = slot;
+			Name = oldName;
+		}
+
+		public override bool IsSlotValid() { 
+			return false;
+		}
+
+		// Only display unloaded slots if in singleplayer. Don't need people carrying items in to multiplayer they shouldn't have :(
```
Chicken-Bones, line 17: unsure about this, may cause confusion. The item can switch to 'unloaded item' independently of the slot. I think if a server removes a mod, or a player joins a server which doesn't have the mod adding the slot, they should still be able to get the item and choose how/if to use it with the other slots they have available here.

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -0,0 +1,206 @@
...
+
+				// Place loaded items in to the correct slot
+				exDyesAccessory[type] = dyes[i];
+				exHideAccessory[type] = visible[i];
+				exAccessorySlot[type] = items[i];
+				exAccessorySlot[type + order.Count] = items[i + order.Count];
+			}
+		}
+
+		// The following netcode is adapted from Chicken-Bone's UtilitySlots:
```
Chicken-Bones, line 89: *ChickenBones'

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -0,0 +1,536 @@
...
+			}
+
+			Main.inventoryBack = flag3 ? new Color(80, 80, 80, 80) : color;
+
+			int yLoc = 0, xLoc = 0;
+			bool customLoc = false;
+
+			if (modded) {
+				ModAccessorySlot mAccSlot = list[slot];
+				customLoc = mAccSlot.XLoc != 0 || mAccSlot.YLoc != 0;
```
Chicken-Bones, line 163: I don't suppose there's any actual desire for custom location (0, 0)? Might be better to return -1 for 'no custom location'? Not a big deal either way.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1608,15 +1614,21 @@
 +			for (int k = 0; k < 10; k++) {
 +				if (IsAValidEquipmentSlotForIteration(k))
 +					VanillaUpdateEquip(armor[k]);
- 			}
++			}
++
++			LoaderManager.Get<AccessorySlotLoader>().VanillaUpdateEquipsMirror(this);
```
Chicken-Bones, line 1619: I'd prefer if all equip related `AccessorySlotLoader` functionality was implemented via `ModAccessorySlotPlayer `. That way we can ensure feature parity with mod accessory slots, and the `ModPlayer` api in terms of having equipment effects/visuals applied from custom sources.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -935,6 +927,20 @@
  		}
  
  		public void UpdateDyes() {
+@@ -6382,11 +_,12 @@
+ 					UpdateItemDye(i < 10, hideVisibleAccessory[num], armor[i], dye[num]);
+ 				}
+ 			}
++			LoaderManager.Get<AccessorySlotLoader>().VanillaUpdateDyesMirror(this);
```
Chicken-Bones, line 934: See below comment about using a `ModPlayer`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3063,6 +3091,18 @@
...
  					flag = true;
  					break;
+@@ -25059,8 +_,10 @@
+ 			if (armor[12].legSlot >= 0)
+ 				legs = armor[12].legSlot;
+ 
+-			if (!dead)
++			if (!dead) {
+ 				UpdateVisibleAccessories();
++				LoaderManager.Get<AccessorySlotLoader>().VanillaUpdateVisibleAccessoriesMirror(this);
```
Chicken-Bones, line 3101: See above comment about using a `ModPlayer`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4117,6 +4167,16 @@
  			Vector2 pointPoisition = Main.MouseWorld;
  			pointPoisition += offsetFromCursor;
  			LimitPointToPlayerReachableArea(ref pointPoisition);
+@@ -34132,6 +_,9 @@
+ 				return;
+ 			}
+ 
++			if (LoaderManager.Get<AccessorySlotLoader>().VanillaPreferredGolfBallMirror(ref projType))
++				return;
```
Chicken-Bones, line 4175: See above comment about using a `ModPlayer`. This one probably needs a new hook

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4814,10 +4874,22 @@
...
+ 					newPlayer.wingsLogic = newPlayer.armor[i].wingSlot;
++					newPlayer.equippedWings = newPlayer.armor[i];
++				}
+ 
+ 				if (type == 158 || type == 396 || type == 1250 || type == 1251 || type == 1252)
+ 					newPlayer.noFallDmg = true;
  
  				newPlayer.lavaTime = newPlayer.lavaMax;
  			}
++			LoaderManager.Get<AccessorySlotLoader>().VanillaLastMinuteFixesMirror(newPlayer);
```
Chicken-Bones, line 4892: See above comment about using a `ModPlayer`

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -0,0 +1,536 @@
...
+				if (type == 906 || type == 4038)
+					newPlayer.lavaMax += 420;
+
+				if (newPlayer.wingsLogic == 0 && ModSlotPlayer.exAccessorySlot[i].wingSlot >= 0) {
+					newPlayer.wingsLogic = ModSlotPlayer.exAccessorySlot[i].wingSlot;
+					newPlayer.equippedWings = ModSlotPlayer.exAccessorySlot[i];
+				}
+				
+				if (type == 158 || type == 396 || type == 1250 || type == 1251 || type == 1252)
+					newPlayer.noFallDmg = true;
```
Chicken-Bones, line 522: Rather than duplicating the vanilla code here, it would be ideal if the approach used for the rest of the equipment effects was followed. By which I mean:

Split the vanilla `LastMinuteFixes` method into one that takes an `Item` parameter, and the vanilla accessory slot iteration just calls that one. That way mods can call it too.

Then have a `LastMinuteFixes` hook on `ModPlayer` which you use to apply your custom accessory slots.

Chicken-Bones (changes requested):
Much improved! My notes are mostly around naming, and hooks in `Player`

direwolf420:
```cs
bool flag12 = false;
for (int num23 = 3; num23 < 10; num23++) {
    if (armor[num23].stack > 0 && armor[num23].wingSlot > -1)
        flag12 = true;
}
```

And `!flag12` instead of `wingsLogic == 0`.

Should be present where the fall damage is applied/calculated in Player.cs. More concrete:
Application: `Hurt(PlayerDeathReason.ByOther(0), num25, 0);`, the if conditional leading up to it should be adjusted.
Missing Code: `if (stoned) {`, above it, is where the above vanilla code is located.

direwolf420:
I know this was 6 years ago, but the above mentioned code is what you removed in this commit @Chicken-Bones: [commit](https://github.com/tModLoader/tModLoader/commit/60fbe9f708acc8cc81a34b13eb4ddcd89a654b63), do you remember why it had to be done like that? Was the bug with mounts not present back then?

Chicken-Bones:
> I know this was 5+ years ago, but the above mentioned code is what you removed in this commit @Chicken-Bones: [commit](https://github.com/tModLoader/tModLoader/commit/60fbe9f708acc8cc81a34b13eb4ddcd89a654b63), do you remember why it had to be done like that? Was the bug with mounts not present back then?

A poor assumption back then. I did not test with mounts. My intention was to replace the iteration loop with something that could check if a wing effect had been applied this frame (from any source). We still need a proper fix for this. As I mentioned in my review to solxan, effects should come from _items_ not from accessory slots specifically. Mods should be able to feed items into the `UpdateX` methods to produce effects each frame, regardless of if the item actually exists, or is in an accessory slot. 

One solution may be an additional flag in `Player` for 'hasWings' which is set when applying equip effects for a wing even if mounted.

Solxanich:
> I know this was 5+ years ago, but the above mentioned code is what you removed in this commit @Chicken-Bones: [commit](https://github.com/tModLoader/tModLoader/commit/60fbe9f708acc8cc81a34b13eb4ddcd89a654b63), do you remember why it had to be done like that? Was the bug with mounts not present back then?

I've updated the logic to now check the "Item equippedWings" field that this PR adds. Since this field would be accurate independent of mounts, it is much safer to work off of, and can be set by whatever, wherever.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedAccessorySlot.cs
Solxanich: Gonna make a judgement call - if a player has an op accessory going in, odds are they have other op items elsewhere on the player.
This kind of check would thus be redundant.
Will default back to just if there is an item

direwolf420:
If a mod would want to iterate over all accessories, including modded ones implemented through the PRs system, what code would it use?

Solxanich:
> If a mod would want to iterate over all accessories, including modded ones implemented through the PRs system, what code would it use?

Look at ItemSlot.tml.cs -> Accessory Swap. Gives a fairly clean iteration example

direwolf420:
Is all the code in `AccessorySwap` method public so it can be used in other mods?
Also, the iteration cuts off at half the accessory length, what does that skip, vanity slots?

Chicken-Bones:
> If a mod would want to iterate over all accessories, including modded ones implemented through the PRs system, what code would it use?

I've long told people this is not the intended design, and actively avoided exposing the ability to do this, because modded slots normally have something 'special' about them. The vanilla (and tML) design is for mods to be able to trigger 'item effects' by calling the appropriate methods on `Player`, regardless of source, or whether the item actually exists.

direwolf420:
There is a mod which requires full control over the accessory slots so it can block items going in/discard them if they somehow get in them anyway. That would require either hooking the tml methods using manual detours, or iterating over accessory slots.

https://github.com/hamstar0/tml-lockedabilities-mod/blob/57de8f32ef19c1f72579f9edc4eb3310cfa2c3c1/LockedAbilities/MyPlayer_Test_Armor.cs#L21

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -0,0 +1,426 @@
...
+		/// </summary>
+		public bool ModSlotCheck(Item checkItem, int slot) => CanAcceptItem(slot, checkItem) &&
+			!ItemSlot.AccCheck(Player.armor.Concat(ModSlotPlayer(Player).exAccessorySlot).ToArray(), checkItem, slot + Player.armor.Length);
+
+		/// <summary>
+		/// After checking for empty slots in ItemSlot.AccessorySwap, this allows for changing what the target slot will be if the accessory isn't already equipped.
+		/// DOES NOT affect vanilla behaviour of swapping items like for like where existing in a slot
+		/// </summary>
+		public void ModifyDefaultSwapSlot(Item item, ref int accSlotToSwapTo) {
+			for (int num = 0; num < ModSlotPlayer(Player).SlotCount(); num++) {
```
Chicken-Bones, line 399: It might be nicer to run this iteration in reverse order? Otherwise, if I have 2 slots which are both 'priority' for an item, it will pick the last one. (Unless I do some weird check on `accSlotToSwapTo`)

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		public bool ShowVisuals {
+			get => ModSlotPlayer.exHideAccessory[Type];
+			set => ModSlotPlayer.exHideAccessory[Type] = value;
+		}
+
+		protected sealed override void Register() => Type = LoaderManager.Get<AccessorySlotLoader>().Register(this);
+
+		public bool IsEmpty() => FunctionalItem.IsAir && VanityItem.IsAir && DyeItem.IsAir;
```
Chicken-Bones, line 64: might be nice for this, and `IsEnabled` to be properties for consistency?

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,116 @@
...
+		public virtual bool IsEnabled() => true;
+
+		/// <summary>
+		/// Override to replace the vanilla effect behavior of the slot with your own.
+		/// By default calls:
+		/// Player.VanillaUpdateEquips(FunctionalItem), Player.ApplyEquipFunctional(FunctionalItem, ShowVisuals), Player.ApplyEquipVanity(VanityItem)
+		/// </summary>
+		public virtual void ApplyEquipEffects() {
+			Player.VanillaUpdateEquip(FunctionalItem);
+			Player.ApplyEquipFunctional(FunctionalItem, ShowVisuals);
```
Chicken-Bones, line 91: `ApplyEquipFunctional` takes a second parameter `hideVisual`, so it should be `!ShowVisuals`

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
```diff
@@ -0,0 +1,116 @@
...
+		/// </summary>
+		/// <returns></returns>
+		public virtual bool IsVisibleWhenNotEnabled() => !IsEmpty();
+
+		/// <summary>
+		/// After checking for empty slots in ItemSlot.AccessorySwap, this allows for changing what the default target slot (accSlotToSwapTo) will be.
+		/// DOES NOT affect vanilla behaviour of swapping items like for like where existing in a slot
+		/// Return true to set this slot as the default targetted slot.
+		/// </summary>
+		public virtual bool ModifyDefaultSwapSlot(Item item, int accSlotToSwapTo) => false;
```
Chicken-Bones, line 114: I personally need a hook for when an item is quick swapped into my slot.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2740,6 +2761,73 @@
...
++				int randomLocationX = screenWidth - 64 - 28;
++
++				if (num39 == num38 && !_achievementAdvisor.CanDrawAboveCoins) {
++					_achievementAdvisor.DrawOneAchievement(spriteBatch, new Vector2(randomLocationX - 10 - 47 - 47 - 14 - 14, chkMax + 8), large: false);
++					UILinkPointNavigator.SetPosition(1570, new Vector2(randomLocationX - 10 - 47 - 47 - 14 - 14, chkMax + 8) + new Vector2(20f) * inventoryScale);
++				}
++
++				DrawDefenseCounter(randomLocationX, chkMax);
++
++				LoaderManager.Get<AccessorySlotLoader>().DrawAccSlots(num20);
```
Chicken-Bones, line 2825: Personal request. It would be great if there was a way to get access to `num20` for my slot draw positions. Maybe store it in a global variable in `Main`?

Chicken-Bones (changes requested):
Things I found while porting UtilitySlots! I'll probably have one more review/testing pass when you've implemented this feedback.

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -0,0 +1,426 @@
...
+			return true;
+		}
+
+		/// <summary>
+		/// Is run in AccessorySlotLoader.Draw. 
+		/// Generates a significant amount of functionality for the slot, despite being named drawing because vanilla.
+		/// At the end, calls this.DrawModded() where you can override to have custom drawing code for visuals.
+		/// Also includes creating hidevisibilitybutton.
+		/// </summary>
+		internal void DrawFunctional(Item[] access, bool[] visbility, int context, int slot, bool flag3, int xLoc, int yLoc) {
```
Chicken-Bones, line 241: Should use a `ref bool visibility` and drop the `slot` parameter

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
Solxanich: Good catch! Property itself is mis-named, should be HideVisuals as true is invisible

Solxanich:
> Is all the code in `AccessorySwap` method public so it can be used in other mods?

Last I checked yes, should all be public.

> Also, the iteration cuts off at half the accessory length, what does that skip, vanity slots?
Yes, skips vanity slots, same pattern as armor.



> > If a mod would want to iterate over all accessories, including modded ones implemented through the PRs system, what code would it use?
> 
> I've long told people this is not the intended design, and actively avoided exposing the ability to do this, because modded slots normally have something 'special' about them. The vanilla (and tML) design is for mods to be able to trigger 'item effects' by calling the appropriate methods on `Player`, regardless of source, or whether the item actually exists.

I disagree, there are several scenarios where one might want to iterate over what players have equipped, such as in rpg-esque mods like what Dire pointed to. We should be encouraging using of a common data storage point, for benefit of mod compatibility in terms of what items the player has actively equipped/in use, independent of the item's actual effects. 
For example, let's say I want to do an accessory set bonus mod, similar to armor set bonuses. 
I would not care which slots have the exact items, just that said items are equipped as is someplace.

> There is a mod which requires full control over the accessory slots so it can block items going in/discard them if they somehow get in them anyway. That would require either hooking the tml methods using manual detours, or iterating over accessory slots.
> 
> https://github.com/hamstar0/tml-lockedabilities-mod/blob/57de8f32ef19c1f72579f9edc4eb3310cfa2c3c1/LockedAbilities/MyPlayer_Test_Armor.cs#L21

For what you're describing, ItemLoader.CanEquipAccessory() should be a sufficient hook.

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
Solxanich: An OnEnterSlot(context) hook for all movement vectors or just from quick swap?

Solxanich:
Generic review comment so I don't forget: 
So, to summarize, internalize capability to iterate, add an IncompatibleWith/IsEquivalentWith hook?

CanBeEquippedWith seems best?
on ItemLoader?

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -139,5 +140,61 @@ public partial class Player
 			bool four = modBiomeFlags.Cast<bool>().Contains(true);
 			return !(one || two || three || four);
 		}
+
+		/// <summary>
+		/// Invoked at the end of loading vanilla player data from files to fix stuff that isn't initialized coming out of load.
+		/// Corrects the player.lavaMax time, wingsLogic, and no fall dmg to be accurate for the provided items in accessory slots.
+		/// </summary>
+		internal static void LoadPlayerLastMinuteFixes(Item item, Player newPlayer) {
```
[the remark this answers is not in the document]
Solxanich: Not a vanilla bug per se - the method really only runs during player select screen, so it doesn't really have any ability to know the slot is or isn't available.
I left it as private primarily because this method is more decorative than practical - it doesn't actually change anything in terms of rendering on player select nor functionality once in world (beyond the first tick).

I'll change it to public with the additional notes, but don't think we need to think too much about this method beyond that it has been matched for purposes of vanilla consistency.

patches/tModLoader/Terraria/Main.cs.patch
Solxanich: You can (and should) use the existing xLoc and yLoc properties to specify where you want the slot to be drawn - from there, AccessorySlot.DrawModded() will provide the exact positions it recommends through its position argument.

That said, after saying that and doublechecking the code, I can see how having that info might be helpful in deciding where to draw said slot less blindly.

Added AccessorySlotLoader.DrawVerticalAlignment and corresponding summary

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
Solxanich: DrawSlot() still needs a slot parameter, and propagates all the way to ItemSlot related functionality from this method.

So I've rewritten the method to just be drawing visibility so that the outcome makes sense.

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -139,5 +140,61 @@ public partial class Player
...
+				if (cShieldFallback != -1)
+					cShield = cShieldFallback;
+			}
+
+			if (ItemIsVisuallyIncompatible(item))
+				return;
+
+			if (item.wingSlot > 0) {
+				if (invisible && (velocity.Y == 0f || mount.Active))
+					return;
```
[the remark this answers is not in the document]
Solxanich: || This prevents the vanity accessory from rendering, unlike vanilla. The item in a wing's vanity slot may not be a wing.
Addressed.

|| Also, I think we could(should?) move this wingSlot check/hack to just check our new equipped wings field, and set wings to that, if the player has y velocity, before calling Player.UpdateVisibleAccessories.
Sounds reasonable, no point to have that in iterations when already storing the item itself

|| Both this method, and Player.UpdateVisibleAccessory should be public so mods can feed items into them.
Addressed

|| Really, this method should not take the vItem parameter. wings is set in UpdateVisibleAccessory anyway. && The iteration order for accessories is also important. All functional accessories should be iterated before all vanity accessories.
Why does the iteration order for accessories matter for functional before vanity? I personally can't see exactly what may break down at a glance, but is there something known that you are wary of?

patches/tModLoader/Terraria/ModLoader/ModAccessorySlot.cs
Solxanich: A hook seems unnecessary - why not just track on the AccessorySlot itself when FunctionalItem.Type changes?

Is it specifically that are looking for the context as to how it got in to the slot?

Solxanich: A hook seems unnecessary - why not just track on the AccessorySlot itself when FunctionalItem.Type changes?

Is it specifically that are looking for the context as to how it got in to the slot?

patches/tModLoader/Terraria/Player.TML.cs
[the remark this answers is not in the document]
Solxanich: Nvm, see the issue. Will fix.

Chicken-Bones (approved):

PR #1302, opened 2021-01-27
Portuguese Translations

xKirtle:
There was no European Portuguese Translation for tModLoader so I went ahead and translated it. Let me know if there's something wrong with it.

Note: The English Localization file that I was translating from had a typo in a word and I fixed it as well.


Mirsario:
Terraria only supports the following languages:
> Deutsch (German)
> Italiano (Italian)
> Français (French)
>Español (Spanish)
> Pусский (Russian)
> 简体中文 (Simplified Chinese)
> Português brasileiro (Brazilian Portuguese)
>Polski (Polish)

Placing another .json file in its localizations folder won't do anything. There was a plan to add a ModLanguage class and let mods add new languages, but this hasn't been implemented yet. I don't think we'll have tML add new languages on its own.

xKirtle:
Oh okay. I recommend mentioning that in the [Contributing Localization](https://github.com/tModLoader/tModLoader/wiki/Contributing-Localization) guide then.

Chicken-Bones:
I'd like to leave this open as a reminder to add additional language support. We could include these translations if/when we add it.

PR #1303, opened 2021-01-27
Made "DefaultToXX" in Item.cs Publicization

Crapsky233:
There are a few "DefaultToXX" methods in Item.cs such as DefaultToBow, DefaultToRangedWeapon etc. But unfortunally, they are private. This PR makes them the public. These may help modders while setting defaults to their items.

### DIFF
--- a/patches/tModLoader/Terraria/Item.cs.patch
+++ b/patches/tModLoader/Terraria/Item.cs.patch
@@ -336,6 +336,193 @@
  		}
  
  		public bool FitsAmmoSlot() {
+@@ -44077,7 +_,7 @@
+ 			return false;
+ 		}
+ 
+-		private void DefaultToSeaShelll() {
++		public void DefaultToSeaShelll() {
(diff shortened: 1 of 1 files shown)


Mirsario:
How is CloneDefaults not sufficient?

Crapsky233:
> How is CloneDefaults not sufficient?

When using CloneDefaults, it is usually necessary to reset some of the  stats again to get a different item. The "DefaultTo" method gives different kinds of items a quicker way to set stats. One line of code can meet the requirements, and it can also guide Modders to set stats for an item, it makes their task easier.
Also, "DefaultTo" as an additional method, it can complement with CloneDefaults.

JavidPack:
This makes sense. The DefaultTo stuff is for quickly setting up the required fields for a specific item type.

Chicken-Bones (approved):
Good alternative to `CloneDefaults` if you don't want to override all the properties of a specific item. I see this as a positive.

Issue #1271, opened 2021-01-07
Delete mod button


goldern:
### Description
As it's a hassle to be forced to: 
1. Disable the desired mod
2. Enter in "More info" at every single mod
3. Press **Delete** at every single mod, which turns into a headache fast
(if you disable tons of mods and then reload and try to remember which mods have been disabled to delete instead of disabled to _not_ be deleted)
Which is easier if you just close the game, go to the folders and delete manually every desired mod, but that's outside tModloader 

**I got an interesting idea**: to implement a button that marks the mod as **to be deleted** and be _greyed out_ 
until you press **Reload Mods**, which then it will start mass deleting all the mods in the **to be deleted** state.

### What does this proposal attempt to solve or improve?
It will make easier to correct mistakes if many mods are added but aren't what the player is looking for, changing multiple mods or making little changes in the mod list more clear and facilitating new users tmodloader experience, also it adds a way inside the game to delete multiple mods in less time.

### Sketch of the concept
![dumbterraridea](https://user-images.githubusercontent.com/5741155/103833845-b7765480-5060-11eb-992b-62f8b7ef58a3.png)
imgur: [https://imgur.com/GBcpKvP](url)

### How it works:
1. Press the trash icon in _each mod_ you want to delete, it will make said mod get the **to be deleted** state and have a lower opacity, can be undone by clicking again in the trash icon, restoring the 100% opacity]
2. Once "Reload mods" is pressed, it will delete all mods with the **to be deleted** state, and proceed to reload the remaining not deleted mods.

_The sketch button is just a suggestion, as I'm not very creative with buttons or design, it could be moved to another place, made smaller, whatever fits._

Mirsario:
Better than marking for deletion, imo it's better to just have the button clasicaly redirect to a deletion confirmation dialogue, which will be skippable if you hold something like Shift. Like deleting messages in Discord.

goldern:
The issue with direct deletion is that it will always trigger mod reload, else,  if the player directly deletes the mod and tries to play while it's loaded, the game would crash as it's referring to the previous content that was deleted, which in turn would lose time by constant reloads instead of deleting everything and reloading in one go.

Mirsario:
It doesn't sound like an issue to only allow deleting disabled mods.

Chicken-Bones:
I approve of the suggestion as is
You're welcome to PR it
However please note you can delete mods from the mods folder much more easily when the game is closed

Mirsario:
[...]

PR #1311, opened 2021-01-29
[1.4] Quick Mod Delete

xKirtle:
### What is the new feature?
This feature allows the user to quickly delete disabled mods without having to go into the More Info menu.
Because it's currently hard to quickly delete mods inside tModLoader.
Yes, issue #1271 has a similar concept design.
(Pressing L/R Shift skips the dialog box)
https://user-images.githubusercontent.com/30240961/106326907-2d8f6500-6275-11eb-9a59-02f711b5ccf7.mp4

Notes: 
- The trash icon might need to be updated, I simply made use of an already existing asset.
- I couldn't quite manage to center the wrapped text in the dialog box, I'd love some feedback (and help) on it.

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
@@ -338,8 +338,6 @@
 		// Misc
 		"OK": "OK",
 		"Continue": "Continue",
-		"Yes": "Yes",
-		"No": "No",
 		"DeleteModConfirm": "Are you sure you want to delete this mod permanently?",
 		"PatreonSetTooltip": "Thank you for your support!",
 		"WorldGenError": "A problem was encountered during world generation",
--- a/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
@@ -338,6 +338,9 @@
 		// Misc
 		"OK": "OK",
 		"Continue": "Continue",
+		"Yes": "Yes",
+		"No": "No",
+		"DeleteModConfirm": "Are you sure you want to delete this mod permanently?",
 		"PatreonSetTooltip": "Thank you for your support!",
 		"WorldGenError": "A problem was encountered during world generation",
 		"WorldIODataException": "The game encountered a mod data problem for a world.",
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
@@ -395,14 +395,7 @@ namespace Terraria.ModLoader.UI
 		}
 
 		private void QuickModDelete(UIMouseEvent evt, UIElement listeningElement) {
-			bool shiftPressed = false;
-			List<Keys> pressedKeys = Main.keyState.GetPressedKeys().ToList();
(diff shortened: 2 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
```diff
@@ -371,5 +393,82 @@ internal class UIModItem : UIPanel
...
+
+		private void QuickModDelete(UIMouseEvent evt, UIElement listeningElement) {
+			bool shiftPressed = false;
+			List<Keys> pressedKeys = Main.keyState.GetPressedKeys().ToList();
+			for (int i = 0; i < pressedKeys.Count; i++) {
+				if (pressedKeys[i] == Keys.LeftShift || pressedKeys[i] == Keys.RightShift) {
+					shiftPressed = true;
+					break;
+				}
+			}
```
Chicken-Bones, line 405: Use `Main.keyState..PressingShift()`

patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
```diff
@@ -371,5 +393,82 @@ internal class UIModItem : UIPanel
...
+		}
+
+		private void CloseDialog(UIMouseEvent evt, UIElement listeningElement) {
+			SoundEngine.PlaySound(SoundID.MenuClose);
+			_blockInput?.Remove();
+			_deleteModDialog?.Remove();
+		}
+
+		private void DeleteMod(UIMouseEvent evt, UIElement listeningElement) {
+			File.Delete(_mod.modFile.path);
```
Chicken-Bones, line 469: You must call `_mod.Close()` before deleting the tmod file. To allow the mod to close any open file handles (like streaming music).

You must also reload immediately after exiting the mods menu.

Chicken-Bones (changes requested):
Overall I'm happy with the design, though you should use the bin icon from the worlds/players UI.

patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
```diff
@@ -338,6 +338,9 @@
 		// Misc
 		"OK": "OK",
 		"Continue": "Continue",
+		"Yes": "Yes",
```
Steviegt6, line 341: You should use `LegacyMenu.104` (Yes) and `LegacyMenu.105` (No). These are already localized.

Steviegt6 (changes requested):

patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
xKirtle: The trash icon only shows up on disabled mods, shouldn't all mods have already no open file handles? I can try to make it work for enabled mods too, although I don't think it makes much sense to uninstall enabled mods.

xKirtle: Regarding the inventory trash icon, that's the one I'm using, except without a background border. img

Not sure if there's a blank border that I can add below it to make it look like all the other buttons.

patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
xKirtle: Did not know of those, thanks.

Mirsario (approved):
Looks and works well.

PR #1312, opened 2021-01-30
[Ready4Review] [1.4] Ported PR #749 to 1.4; RE: Fix scrolling a UI scrollbar also adjusting hotbar position

Solxanich:
NOTE: This replaces TehCupcakes #749. All credit to change belongs to them, and I've merely ported all changes one2one to 1.4 target. See original PR for its mostly completed review status.

My additional comments observed while initial porting it: Not sure why PlayerInput.cs Initialize() was modified in original; it doesn't show any need to be.

From the original:
"Description of the Change

Main.InGameUI.Update() -> Technically this happens first, but because the player's input hasn't been read yet, it's actually one step behind for picking up scrolls.
PlayerInput.UpdateInput() -> Sets scroll wheel values.
Player.Update() -> Updates the hotbar/recipes position based on scroll value.
This creates a problem, since we don't want the hotbar to move if a scrollbar has been scrolled, but we haven't reached that point yet. It's also not possible (or at least quite difficult) to identify if a scrollbar will update at the point of Player.Update(). My change records the input scroll value when attempting to move the hotbar and delays processing by one cycle. This gives the UI elements and mods an opportunity to reset the value if they handle the scrolling and want to cancel out the hotbar offset.

Alternate designs
I originally relocated the hotbar handling so it would happen just after the UI update, and added a condition that it would only update if a scrollbar event had not been handled. However, this was a bit of a more significant change to the original code, so it was suggested that I try delaying the processing a frame instead of changing the order altogether.

I also tried identifying in the player update method whether the user was hovering on a UI element with a scrollbar. I did not get this working, and decided there must be a better method.

Why this should be merged into tModLoader

Benefits
Better user experience.

Possible drawbacks
Some mods may have relied on the old behavior. For instance, BossChecklist (and many of @JavidPack's other mods) would calculate how much the hotbar moved and reverse it. (See here.) This fix breaks the workaround, but obviously it's better to have a proper fix in tModLoader that prevents stutter.

Applicable Issues
N/A

Sample Usage
N/A"


Solxanich:
Currently just have to test the original changes still do what intended.

Steviegt6:
We don't usually port existing PRs. The last time we ported commits, we did it in a wave, and it only took 2-3 days (and there were a considerable amount of commits). I'd just wait for the existing PR to get merged.

Itorius:
This is one of the issues I solve in my [input rewrite](https://github.com/Itorius/tModLoader/tree/1.4_inputrewrite) branch.

Chicken-Bones:
I think this is definitely the correct approach and I'd like to see it merged before Itorius' input rewrite. I haven't done a code review yet. It may be better to delay the visibility of scroll processing to non UI until the next frame, with a generic cancel scroll, rather than explicitly readying a hotbar scroll and then the UI cancelling it specifically

Solxanich:
tl;dr: I suspect doing this PR is not valuable due to input_rewrite, and further see this PR as an improved workaround (in current state) and not a fix. The question is, is this improved workaround needed before input_rewrite?

Hmm; before I begin working on testing and improving this, I'd like to confirm some information first with @Itorius - namely regarding this change and the input_rewrite.
Itorius, would this quick diagram summarize why the input_rewrite design 'eliminates' the original raised issue? 
image

Put a different way, assuming I got the gist of it, the rewrite doesn't use this old solution, so it is important to question if making a change provides any benefit to tML if whatever fix is implemented is going to subsequently be removed alongside the problem source in the rewrite. I don't see value in rushing this fix unless the input_rewrite is still a long ways from completion (2-3 months?) AND we are planning to go to Beta well ahead of then.

Coming back to this PR itself, I would agree general idea of addressing this issue is important; I don't particularly like that this PR feels more like improving the workaround to the issue to be friendlier instead of actually fixing it. I'll take a look at this further if we are pushing for this, and can't wait for input_rewrite.

Solxanich:
Hmm. Looking through this and working through it, I suspect it would be better to simply just grab a flag similar to Main.PlayerInventory (except you know, when player is hovering over the mod's UI) and use that to differentiate instead of playing with scrolling.

I've changed the design to be smarter imo. 

Example Usage for CheatSheet UIView class:

`PlayerInput.LockVanillaMouseScroll("CheatSheet:UIView", this.MouseInside);`

Chicken-Bones:
Look, this is great, much better tbh. However, I'm uncomfortable with a design that relies on a mod calling unlock. The best interim solution is a `ModSystem` hook, (or a `UIElement` one) which is automatically cleared by tML.

We also have to consider draggable windows, where they might be dragged on top of each other. Or maybe we don't, I have no idea how click events get 'consumed' by the 'top level component' either. UI is not my forte

Solxanich:
Hmm. I'm not too familiar with UI design either, for the matter. But based on my dive, my notes show the following:
1) Some people create their own UI classes, so inheritance isn't a good path
2) All mod UI's I looked at had a method similar to "bool IgnoreMouseInputs()", which they used to determine if their UI should handle mouse inputs or ignore them 
3) In particular, the code for IgnoreMouseInputs() blends well with this, as the code looks like
`` bool IgnoreMouseInputs() {
bool myUIContainsMouse = newRectangle(this.position.X, this.position.Y, this.Width, this.Height).Contains(Main.MouseScreen.ToPoint());
_PlayerInput.LockVanillaMouseScroll("CheatSheet:UIView", myUIContainsMouse);_
return myUIContainsMouse;
}
``
4) This design basically is that so long as the mouse is within your UI's window, we will suppress the vanilla mouse scroll events. It does not prevent mouse scrolls on multiple overlapping UI. Given point 2, I suspect the mod author already designs with ignoring mouse events from outside their UI, so this stacks on well. See CheatSheet's UIView HandleMouseInputs() as a good example of where I'm coming at this from. 

Now in terms of the overlapped UI, my understanding is that whichever UI window handles it first has the option to zero the PlayerInput.scrollDelta and prevent other UIs from scrolling. Most don't afaik. You'd probably end up scrolling both, but I think analyzing that is a separate edge case, as I don't know how you'd effectively determine what should or shouldn't scroll, be it both, the top level, or something else.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIScrollbar.cs.patch
```diff
@@ -1,9 +1,24 @@
...
  			_isDragging = false;
- 		}
- 	}
++		}
++
++		public override void MouseOver(UIMouseEvent evt) {
++			base.MouseOver(evt);
++			PlayerInput.LockVanillaMouseScroll();
++		}
++	}
```
Mirsario, line 21: This entire file shouldn't be a patch, and the added code should instead be in UIScrollbar.TML.cs. Just add `partial` in here.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIList.cs.patch
```diff
@@ -24,15 +32,31 @@
...
+@@ -86,6 +_,11 @@
+ 		public override void Recalculate() {
+ 			base.Recalculate();
+ 			UpdateScrollbar();
++		}
++
++		public override void MouseOver(UIMouseEvent evt) {
++			base.MouseOver(evt);
++			PlayerInput.LockVanillaMouseScroll();
  		}
```
Mirsario, line 60: This, AddRange, and ViewPosition should go into a partial file.

patches/tModLoader/Terraria/GameInput/PlayerInput.cs.patch
```diff
@@ -179,3 +179,20 @@
...
++		}
++
++		internal static List<string> MouseInModdedUI = new List<string>();
++		/// <summary>
++		/// Locks the vanilla scrollbar for the upcoming cycle when called. Autoclears in Player.
++		/// Takes a string to denote that your UI has registered to lock vanilla scrolling.
++		/// </summary>
++		public static void LockVanillaMouseScroll(string myUI) {
++			if (!MouseInModdedUI.Contains(myUI))
++				MouseInModdedUI.Add(myUI);
```
Mirsario, line 195: Should also go into a partial.

Mirsario (changes requested):
Watch me as I repeat the same thing 3 times below.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIList.cs.patch
Mirsario: The following isn't your change, but can you check if there's any reason why _items is public and not internal?

Solxanich: Regards to the latter, it was changed back in 2016 to be accessible for the ModBrowser; I suspect there wasn't an observed vanilla method to access the individual elements of the list. 
There is a fetchable enumerator now, so I feel comfortable changing to internal.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIScrollbar.cs.patch
Solxanich: Updated to use partials x3

patches/tModLoader/Terraria/ModLoader/Config/UI/ListElement.cs
```diff
@@ -57,6 +58,11 @@ class NestedUIList : UIList
 			//OverflowHidden = false;
 		}
 
+		public override void MouseOver(UIMouseEvent evt) {
+			base.MouseOver(evt);
+			PlayerInput.LockVanillaMouseScroll("modloader/ListElement");
```
Chicken-Bones, line 63: Capitals for `ModLoader` and we're good

patches/tModLoader/Terraria/GameContent/UI/Elements/UIList.cs.patch
```diff
@@ -1,38 +1,23 @@
...
 @@ -29,9 +_,9 @@
  			public override Rectangle GetViewCullingArea() => base.Parent.GetDimensions().ToRectangle();
  		}
  
 -		protected List<UIElement> _items = new List<UIElement>();
-+		public List<UIElement> _items = new List<UIElement>();
++		internal List<UIElement> _items = new List<UIElement>();
  		protected UIScrollbar _scrollbar;
 -		private UIElement _innerList = new UIInnerList();
 +		internal UIElement _innerList = new UIInnerList();
```
Chicken-Bones, line 20: why access change?

Mirsario: https://github.com/tModLoader/tModLoader/pull/1312#discussion_r577216900

patches/tModLoader/Terraria/ModLoader/Config/UI/ListElement.cs
Mirsario:
```suggestion
			PlayerInput.LockVanillaMouseScroll("ModLoader/ListElement");
```

Mirsario (approved):

JavidPack:
Just a note: LockVanillaMouseScroll won't work in MouseOver, MouseOver only triggers once when an element is initially hovered over. I used DrawSelf and it worked
```
if(IsMouseHovering)
	PlayerInput.LockVanillaMouseScroll("RecipeBrowser/UIHorizontalGrid");
```

PR #1316, opened 2021-02-02
[1.4] Buff draw hooks

direwolf420:
### What is the new feature?
Adds Mod/GlobalBuff `PreDraw`/`PostDraw` and `LeftClick/RightClick`. Adds a `BuffDrawParams` class so the method signatures don't get too long, and it makes for easier changes/additions to the functionality without being API-breaking.
It only changes three lines of code in Main.cs (icon drawing, text drawing, and mouse cursor boundary check) and exposes a total of 18 lines of effective vanilla code. The main portion of the patch contains declaring local variables (which are all just stuffed into `spriteBatch.Draw`  in vanilla), so modders can change and act on them using the provided hooks.

**The LeftClick hook is pending testing with controllers**

1. Most things have Pre/PostDraw.
2. Being able to change how buffs are drawn in general. Can reach from simple texture/color/positioning tweaks to entirely different buff representations.
3. Make things happen on clicking, the buffs are an interface, so click hooks are a given.

I have not included `rotation` and `scale` in `BuffDrawParams` because for these to be useful I would have to adjust vanilla drawing to account for origin. In the current, unmodified state, the origin is the draw position (top left of the icon), which is usually not what modders want, they want the center.
A modder can still manually draw the buff using `PreDraw` and choose all parameters freely.

### ExampleMod updates
Added more comments for `ExampleDefenseBuff`.
Added `AnimatedBuff`.
Added `ExampleGlobalBuff`.
Made `PartyZombie` apply `AnimatedBuff` on hit.


ExampleMod/Common/GlobalBuffs/ExampleGlobalBuff.cs
```diff
@@ -0,0 +1,69 @@
...
+			//Make the campfire buff have a different color and shake slightly.
+			if (type == BuffID.Campfire) {
+				drawParams.drawColor = Main.DiscoColor * Main.buffAlpha[buffIndex];
+				Vector2 shake = new Vector2(Main.rand.Next(-2, 3), Main.rand.Next(-2, 3));
+				drawParams.position += shake;
+				drawParams.textPosition += shake;
+			}
+
+			//If the buff is one of the three specified:
+			if (System.Array.IndexOf(new int[] { BuffID.Regeneration, BuffID.Ironskin, BuffID.Swiftness }, type) > -1) {
```
Mirsario, line 33: Constantly creating new arrays like that is not a good practice.
```suggestion
			if (type == BuffID.Regeneration || type == BuffID.Ironskin || type == BuffID.Swiftness) {
```

ExampleMod/Content/Buffs/AnimatedBuff.cs
```diff
@@ -0,0 +1,102 @@
...
+
+		public override bool PreDraw(SpriteBatch spriteBatch, int buffIndex, ref BuffDrawParams drawParams) {
+			//We draw our special texture here, the animation takes place in AnimatedBuffSystem.
+
+			//Use our animation spritesheet.
+			Texture2D ourTexture = animatedTexture.Value;
+			//Choose the frame to display.
+			//We reference the frame count and frame from the ModSystem here.
+			Rectangle ourSourceRectangle = ourTexture.Frame(verticalFrames: AnimatedBuffSystem.FrameCount, frameY: AnimatedBuffSystem.Frame);
+
```
Mirsario, line 71: There indeed is zero use in that system, lol.
```suggestion
	//This buff has an extra animation spritesheet, and also showcases PreDraw specifically.
	//(We keep the autoloaded texture as one frame in case other mods need to access the buff sprite directly and aren't aware of it having special draw code).
	public class AnimatedBuff : ModBuff
	{
		public const int FrameCount = 4; //Amount of frames we have on our animation spritesheet.
		public const int AnimationSpeed = 60; //In ticks.
		
		private static Asset<Texture2D> animatedTexture;

		public override void SetDefaults() {
			DisplayName.SetDefault("Animated Buff");
			Description.SetDefault("Animates.");

			if (Main.netMode != NetmodeID.Server) {
				//Do NOT load textures on the server!
				animatedTexture = Mod.GetTexture("Content/Buffs/AnimatedBuff_Animation");
			}
		}

		public override void Unload() {
			animatedTexture = null;
		}

		public override bool PreDraw(SpriteBatch spriteBatch, int buffIndex, ref BuffDrawParams drawParams) {
			//We draw our special texture here with a specific animation.

			//Use our animation spritesheet.
			Texture2D ourTexture = animatedTexture.Value;
			//Choose the frame to display, based on constants and the game's tick counts.
			Rectangle ourSourceRectangle = ourTexture.Frame(verticalFrames: FrameCount, frameY: Main.GameUpdateCount / AnimationSpeed % FrameCount);

```

ExampleMod/Content/Buffs/ExampleDefenseBuff.cs
```diff
@@ -8,12 +11,22 @@ public class ExampleDefenseBuff : ModBuff
...
+
+		//Optional
+		public override bool PreDraw(SpriteBatch spriteBatch, int buffIndex, ref BuffDrawParams drawParams) {
+			//You can use this hook to make something special happen when the buff icon is drawn (such as reposition it, pick a different texture, etc.).
+			//Here we make the icon have a lime green tint.
+			drawParams.drawColor = Color.LimeGreen * Main.buffAlpha[buffIndex];
+			
+			//Return true to let the game draw the buff icon.
+			return true;
+		}
```
Mirsario, line 30: Adding changes like this, which are unrelated to the file's name, may very well be the reason why the old EM turned in a mess quite quick. I think it'll be fine if you just stack this effect with the one in AnimatedBuff.

patches/tModLoader/Terraria/DataStructures/BuffDrawParams.cs
```diff
@@ -0,0 +1,62 @@
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+
+namespace Terraria.DataStructures
+{
+	/// <summary>
+	/// Holds data required for buff drawing.
+	/// </summary>
+	public struct BuffDrawParams
```
Mirsario, line 9: Great, maybe this should be done more frequently.

patches/tModLoader/Terraria/ModLoader/GlobalBuff.cs
```diff
@@ -57,5 +58,28 @@ public abstract class GlobalBuff : ModType
...
+
+		/// <summary>
+		/// Allows you to draw things after the buff has been drawn. skipped is true if you or another mod has skipped drawing the buff (possibly hiding it or in favor of new visuals).
+		/// </summary>
+		/// <param name="skipped"><see langword="true"/> if you or another mod has skipped drawing the buff in PreDraw (possibly hiding it or in favor of new visuals)</param>
+		/// <param name="spriteBatch">The spriteBatch that is drawn on</param>
+		/// <param name="type">The buff type</param>
+		/// <param name="buffIndex">The index in Main.LocalPlayer.buffType and .buffTime of the buff</param>
+		/// <param name="drawParams">The draw parameters for the buff</param>
+		public virtual void PostDraw(bool skipped, SpriteBatch spriteBatch, int type, int buffIndex, BuffDrawParams drawParams) {
```
Mirsario, line 82: `skipped` would be nicer as the last parameter here imo. Also, are there any use-cases for it? I don't think people would ever return false in PreDraw without doing their own drawing, they should disable an entire interface layer if they don't want buffs to be visible. And everyone should call tML hooks if they do buff rendering absolutely from scratch.

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/GlobalBuff.cs
direwolf420: One use case is to let the modder know something else returned false. Currently in all PostDraw hooks you cannot know this. Which sucks if you want to draw something that relies on the vanilla drawing, or your own PreDraw.

Though I have a hard time coming up with a convincing use case for `skipped`.

ExampleMod/Content/Buffs/ExampleDefenseBuff.cs
direwolf420: Good point, I'll move it

ExampleMod/Content/Buffs/AnimatedBuff.cs
direwolf420: Overengineering at its finest :cowboy_hat_face:

direwolf420:
Removed the skipped parameter due to lack of definitive use cases

Chicken-Bones (approved):
Looks good to me.

direwolf420:
I have added two new hooks to buffs, aswell as made a vanilla method public with added documentation. All these things need reviews.

Currently, RightClick will return false if atleast one of any GlobalBuff or ModBuff returns false. They do not shortcirquit eachother. 

LeftClick is currently untested with controllers: in vanilla, the buff cancelling for controllers is instead remapped from right clicking to "left clicking in the inventory", and currently there is a possibility where both hooks will run at the same time because of it. Do we just ignore LeftClick for controllers instead?

The method `private static void TryRemovingBuff(int i, int b)` has been made public in order to disincentive copying vanilla code that is used to cancel a buff by right clicking (in case mods want to cancel their buff in the LeftClick hook manually).

direwolf420:
Since 1.4.2, the `TryRemovingBuff` is now public. I have added more documentations to other buff draw related hooks, tell me what you think and if they are too descriptive or not descriptive enough.

direwolf420:
Regarding gamepad: Mirsario suggested that there should be "gamepad keyhints" support when selecting a buff that supports left clicking (`CanLeftClick` hook), and it should be `B` and `A` buttons (where `B` is currently used to cancel it)

Mirsario:
B in that context is "interact" (the pc right click button), and A is non-rebindable.

direwolf420:
I will restart this PR in another one without the `LeftClick` functionality as I seriously can't be bothered to write that as I don't own any controllers or a steam deck to test with.

PR #1335, opened 2021-02-20
ExampleWhip

Vaema:
Heya! I made the ExampleWhip, BaseWhip which is a code helper for whip weapons, and ExampleWhip's projectile. I always wanted to help out tModLoader by contributing to it because of me being a Terraria and tModLoader player and fan. I made ExampleWhip and the other stuff because no one has made these yet. So there we go, another example weapon that's actually a whip that some people wanted.


NotLe0n:
yikes

Mirsario:
Broken PR, stolen code from WeaponOut - https://github.com/Flashkirby/WeaponOut/blob/master/Projectiles/Whips/BaseWhip.cs?ts=4

PR #1336, opened 2021-02-22
Update tModLoader.json

Igmc:
Added missing translations.
Changed translations (like "pack") to make them similar to the ones in the official 1.4 Spanish localization.


JavidPack:
Thanks

PR #1339, opened 2021-02-25
[1.4] Remove skipped parameter from BossBarLoader.PostDraw

direwolf420:
Follow-up on the PRs #1316 and #1288, the latter one adding a skipped parameter on the `1.4` branch, which this PR removes. Reason: Lack of definitive use cases


direwolf420:
@Chicken-Bones

PR #1340, opened 2021-02-25
ExampleWhip and GetWhipSettings hook

jjohnsnaill:
### What is the new feature?
A new hook called `GetWhipSettings` in `ModProjectile` and `GlobalProjectile`.
It adds a single line to `Projectile.cs` in `GetWhipSettings`, and exposes 7 lines of vanilla code.
Without this hook, changing a whip's amount of segments and range multiplier is not possible (not counting IL edits).
As far as I'm aware, no.
### ExampleMod
Included is an example of a modded whip. It has many comments explaining how it works, and has a charging ability to showcase the new hook.


jjohnsnaill:
Whoops, wrong branch. 🙃

PR #1341, opened 2021-02-25
WhipSettings & ExampleWhip

jjohnsnaill:
### What is the new feature?
Adds the ability to make modded whips, configured via `Projectile.WhipSettings` and `DefaultToWhip` for convenience.

Without this, changing a whip's amount of segments and range multiplier is not possible (not counting IL edits).

A hook for `Projectile.GetWhipSettings` would work, and be detourable, but we prefer declarative data based configuration where possible.

### Sample Usage
```cs
	// This method quickly sets the whip's properties.
	Projectile.DefaultToWhip();

	// use these to change from the vanilla defaults
	Projectile.WhipSettings.Segments = 20;
	Projectile.WhipSettings.RangeMultiplier = 1f;
}
```

### ExampleMod
Included is an example of a modded whip. It has many comments explaining how it works, and has a charging ability to showcase the new hook.


ExampleMod/Content/Buffs/ExampleWhipDebuff.cs
```diff
@@ -0,0 +1,32 @@
...
+	public class ExampleWhipDebuff : ModBuff
+	{
+		public override void Update(NPC npc, ref int buffIndex) {
+			npc.GetGlobalNPC<ExampleWhipDebuffNPC>().markedByExampleWhip = true;
+		}
+	}
+
+	public class ExampleWhipDebuffNPC : GlobalNPC
+	{
+		public override bool InstancePerEntity => true;
```
Solxanich, line 23: Add a comment explaining why setting these as true are needed for this (it isn't inherently obvious why both need to be set)

ExampleMod/Content/Buffs/ExampleWhipDebuff.cs
```diff
@@ -0,0 +1,32 @@
...
+
+		public bool markedByExampleWhip;
+
+		public override void ResetEffects(NPC npc) {
+			markedByExampleWhip = false;
+		}
+
+		// Currently, this is inconsistent with vanilla, increasing damage after it is randomised instead of before. It will be changed to a different hook in the future.
+		public override void ModifyHitByProjectile(NPC npc, Projectile projectile, ref int damage, ref float knockback, ref bool crit, ref int hitDirection) {
+			if (markedByExampleWhip && !projectile.npcProj && !projectile.trap && (projectile.minion || ProjectileID.Sets.MinionShot[projectile.type])) {
```
Solxanich, line 34: Add a comment explaining why we also need to check both projectile.trap and projectile.npcProj, and can't just skip straight to the last two: projectile.minion or minionshot

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
```diff
@@ -0,0 +1,168 @@
...
+			get => Projectile.ai[0];
+			set => Projectile.ai[0] = value;
+		}
+
+		private float ChargeTime {
+			get => Projectile.ai[1];
+			set => Projectile.ai[1] = value;
+		}
+
+		public override void AI() {
```
Solxanich: How much of this is boilderplate that should be added to a template method ``CreateWhipAI()`` instead of every mod doing it?
At a glance, this seems like a lot of boilerplate code

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
```diff
@@ -0,0 +1,168 @@
...
+			Player owner = Main.player[Projectile.owner];
+			Projectile.rotation = Projectile.velocity.ToRotation() + MathHelper.PiOver2; // Without PiOver2, the rotation would be off by 90 degrees counterclockwise.
+
+			Projectile.Center = Main.GetPlayerArmPosition(Projectile) + Projectile.velocity * Timer;
+			// Vanilla uses Vector2.Dot(Projectile.velocity, Vector2.UnitX) here. Dot Product returns the difference between two vectors.
```
Solxanich: The engineer in me feels I ought to enhance your understanding of this.
To start, difference between vectors is just subtraction of the vectors. No fancy operation.

Basically, there are three main 'special' vector operations:
1) Cross product - finds a vector that is perpendicular to both of the original vectors (useless in Terraria for most)
2) Projection - finds the component of the vector that matches the direction of the base vector (useful for comparing to vectors along the same axis of interest as it removes perpendicular components)
3) Dot Product returns the sum of multiplying each of the components - ie X1*X2 + Y1*Y2. If the dot product is zero, it means the two vectors are perpendicular. So although you're end conclusion is right in that you can simply check the x component is nonzero, the better summary here is that dot product is ``sum of multiplying each of the components - ie X1*X2 + Y1*Y2`` where Y2 is zero.

ExampleMod/Content/Items/Weapons/ExampleWhip.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+	public class ExampleWhip : ModItem
+	{
+		public override void SetStaticDefaults() {
+			ItemID.Sets.SummonerWeaponThatScalesWithAttackSpeed[Type] = true; // This makes the item be affected by the user's melee speed.
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 1;
+		}
+
+		public override void SetDefaults() {
+			// Item.DefaultToWhip(ModContent.ProjectileType<ExampleWhipProjectile>(), 20, 2, 4); // Call this method to quickly set the whip's properties.
```
Solxanich: I can't tell which of these values 'I' would or wouldn't need to set if I were to use this.

Maybe consider having two separate SetDefaults() methods, one of them commented out, that shows how to use this?
Further, it would be helpful to indicate which of the following values should be adjusted for 'my' own implementation, be it via grouping or comments.

ExampleMod/Content/Items/Weapons/ExampleWhip.cs
```diff
@@ -0,0 +1,44 @@
...
+			Item.useStyle = ItemUseStyleID.Swing;
+			Item.useTime = 30;
+			Item.useAnimation = 30;
+			Item.UseSound = SoundID.Item152;
+			Item.channel = true; // This is used for the charging functionality. Remove it if your whip shouldn't be chargeable.
+			Item.noMelee = true;
+			Item.noUseGraphic = true;
+		}
+
+		// Please see Content/ExampleRecipes.cs for a detailed explanation of recipe creation.
```
Solxanich, line 26: Pretty sure someone PR'd a change to ExampleRecipe, so be sure to verify this comment is still relevant

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		public override void OnHitNPC(NPC target, int damage, float knockback, bool crit) {
+			target.AddBuff(ModContent.BuffType<ExampleWhipDebuff>(), 240);
+			Main.player[Projectile.owner].MinionAttackTargetNPC = target.whoAmI;
+		}
+
+		// This method draws a line between all points of the whip, in case there's empty space between the sprites.
+		private void DrawLine(List<Vector2> list) {
+			Texture2D texture = TextureAssets.FishingLine.Value;
```
Solxanich, line 68: Same thing as AI() -> Is this boilerplate code that every mod will have? Should this instead be a template method that mods call with a handful of parameters instead?

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		public override bool PreDraw(ref Color lightColor) {
+			List<Vector2> list = new List<Vector2>();
+			Projectile.FillWhipControlPoints(Projectile, list);
+
+			DrawLine(list);
+
+			// If you don't want to use custom code, you can instead call one of vanilla's DrawWhip methods. However, keep in mind that you must adhere to how they draw if you do.
+			// Main.DrawWhip_WhipBland(Projectile, list);
```
Solxanich: Please make it more obvious that all the following code is for Custom Draw - A simple comment preceding the rest of the code that everything to follow is Custom Draw would go a long ways for people who skip this line.

Also, very minor, but recommended comment styling is:
// Comment
//code
So can visually see what is comments without reading.

Solxanich (commented):
Good Start,
A few places that could use some polishing.

Beyond that, I'm a tad concerned at the amount of Boilerplate looking code in ExampleWhipProjectile. It would be worthwhile to consider a utility method for modders to invoke instead of having the same code re-hashed in many mods.

Chicken-Bones:
This whip PR is preferable due to the `GetWhipSettings` hook. However, I think it should only be on `ModProjectile` (because it's very specific). There's always `On` hooks if a global wants to modify a specific whip.

Please address Solxan's feedback, and update your PR. If you could look at #2214 and compare and note any differences that would be great too.

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
Mirsario: In my opinion, it's best to have a normal 'ExampleX' that can be used as a template, and then 'ExampleAdavancedX' that reimplements some vanilla code instead of reusing it, if needed.

Falcon-05:
There are a couple of things that should be added to ExampleWhipProjectile and ExampleWhipDebuff for consistency with vanilla:
1. ExampleWhipProjectile should have piercing damage falloff
2. ExampleWhipDebuff should have BuffID.Sets.IsAnNPCWhipDebuff (this makes sure that enemies/bosses that are immune to all debuffs except whip debuffs are not immune to the debuff)

ExampleMod/Content/Projectiles/ExampleWhipProjectile.cs
jjohnsnaill: I described Dot Product in an easy-to-understand way for new modders, or people unfamiliar with it. However, I did add your point about 0 meaning the two vectors are perpendicular.

ExampleMod/Content/Projectiles/ExampleWhipProjectileAdvanced.cs
```diff
@@ -0,0 +1,183 @@
...
+				Projectile.FillWhipControlPoints(Projectile, points);
+				SoundEngine.PlaySound(SoundID.Item153, points[points.Count - 1]);
+			}
+		}
+
+		// This method handles a charging mechanic.
+		// If you remove this, also remove Item.channel = true from the item's SetDefaults.
+		private void Charge() {
+			// Like other whips, this whip updates twice per frame (Projectile.extraUpdates = 1), so 120 is equal to 1 second.
+			if (owner.channel && ChargeTime < 120) {
```
Gabo6480: Where does this 'owner' variable comes from?
It seems to be missing inside of this function.

jjohnsnaill: oh so that's why it didn't build

Issue #319, opened 2018-03-18
ItemLoader.UseItemFrame doesn't get called.


Liburia:
### Description

ItemLoader.UseItemFrame doesn't get called if item.useStyle is 1, 2, 3, 4 or 5.

### Expected behavior 

Called regardless of item.useStyle.

### Actual behavior

```
if (this.inventory[this.selectedItem].useStyle == 5)
{
    ...
}
else if (ItemLoader.UseItemFrame(this.inventory[this.selectedItem], this))
{
    return;
}
```

### Reproduction frequency

Every time.

### Versions

`tModLoader v0.10.1.3`

Jofairden:
It would be useful if you could provide example files that reproduce this issue. I can test if you provide them.

Jofairden:
Since no further comment is given, I assume the problem is solved. Feel free to reply if it isn't.

JavidPack:
This one should stay open actually, I looked into it and it does seem a bit odd that ItemLoader.UseItemFrame can't affect the original useStyles.

JavidPack:
I think I'll wait until I hear more support for calling this regardless. The design of the hook as it currently is seems to assume that only 1 method will take effect, since it short circuits if any global item or moditem returns true. 

If someone can think of a good reason to call a GlobalItem.UseItemFrame regardless, that would help us decide.

Chicken-Bones:
Needs a usecase

Jofairden:
No activity/further reports.

JavidPack:
Yeah, we need to fix this in 1.4

Mirsario:
Fixed in https://github.com/tModLoader/tModLoader/commit/f850a3b9432a10bfee50c965b7e9476f9186c60c

PR #1342, opened 2021-02-26
ItemLoader UseItemFrame Fix Issue#319

Solxanich:
### What is the bug?
#319 

### How did you fix the bug?
I moved ItemLoader.UseItemFrame, which already returns a bool, from the end of the if chain to the beginning so it runs first, and thus attempts are now modded -> global -> vanilla instead of vanilla -> modded -> global.
The obvious concern being that this increases the amount of calculations as the global hook list gets called more often, but that's a given in ensuring globals are checked per the issue's decision

Don't know. Didn't think much on this as it seems straightforward enough.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -2915,15 +2915,6 @@
  					if ((double)itemAnimation < (double)itemAnimationMax * 0.333)
  						bodyFrame.Y = bodyFrame.Height * 3;
  					else if ((double)itemAnimation < (double)itemAnimationMax * 0.666)
-@@ -25525,7 +_,7 @@
- 								bodyFrame.Y = bodyFrame.Height * 2;
- 						}
- 					}
--				}
-+				}	
- 			}
- 			else if (pulley) {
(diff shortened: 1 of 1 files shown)


Mirsario (approved):

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -2905,15 +2905,16 @@
  			if (mount.Active) {
  				legFrameCounter = 0.0;
  				legFrame.Y = legFrame.Height * 6;
-@@ -25526,6 +_,8 @@
- 						}
- 					}
- 				}
-+				else
```
Solxanich, line 2912: Review from discord:
The hook should come at the end so that if someone wants to do a final tweak of piggyback  vanilla use style, they can.
Should wait to update until after Mirsario use item changes are done to ensure PR design relevancy.

Mirsario:
I approved this PR, and then forgot about it and improved the hooks on my own on the main branch. My bad? 🤔 

https://github.com/tModLoader/tModLoader/commit/f850a3b9432a10bfee50c965b7e9476f9186c60c

PR #1343, opened 2021-02-27
DustID

SmugBlanco:
Current state:
DustID is barebones and barely utilized due to that fact. You would often have to refer to the popular forum page 'Dust and Sound Catalogue 2' just to get a semi accurate image of what the dust may be used for.

Proposed changes:
This, a more flushed out DustID class who's variables are named after my best guess of it's original implementation purpose or most popular usage of it.


Mirsario (changes requested):
I should've told you the following in discord right away:
- Additions should be in DustID.TML.cs partial file.
- Don't remove or rename vanilla entries.
- There's no need for line breaks, but you can have them in the partial file, outside patches.

SmugBlanco:
My bad, undid and the changes to vanilla as you said and made the changes into a partial class.

patches/tModLoader/Terraria.ID/DustID.TML.cs
```diff
@@ -0,0 +1,322 @@
+﻿namespace Terraria.ID
+{
+	public static partial class DustID
+	{
+		//Naming is based on the best approximation of earliest implimentation/most popular
+		public const short Grass_Small = 3;
```
direwolf420, line 6: > implimentation

Mirsario:
```suggestion
	{
		//Naming is based on the best approximation of the earliest or most popular implementation.
		public const short Grass_Small = 3;
```

Mirsario (approved):
I had a complaint about underscores, but democracy happened:

PR #1346, opened 2021-02-28
[1.4] Rewrite Mana and Life Regen Hooks

Solxanich:
### What is the new feature?
Implement #1105, alongside other symmetrical life regen hooks by redesigning and implementing a new Regen struct and just 2 simplified hooks.

Give modders a more intricate way of messing around with mana.

"In my opinion, we're better off upgrading fields than making hooks, as it would be a pain to use ModPlayer hooks for, say, effects of a buff (that already has its own class). We're trying to improve designs, after all. What do you think about just using DamageModifiers here and in life regen?"
- Couldn't find Modifiers class, but have used fields better?

See example mod ExampleRegenPlayer.

### ExampleMod updates
Updated to add ExampleRegenPlayer as an example.


patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -914,5 +895,49 @@ public abstract class ModPlayer : ModType
...
+			/// <summary>
+			/// Allows you to modify the power of the player's natural life/mana regeneration. This can be done by setting a multiplication value. For example, campfires multiply it by 1.1, while walking multiplies it by 0.5.
+			/// </summary>
+			public float multiplyRegen;
+			/// <summary>
+			/// Allows you to force no positive regeneration to occur. For example, the "On Fire!" debuff prevents positive life regenaration.
+			/// </summary>
+			public bool maxZeroRegen;
+
+			public static Regen Create() {
```
Chicken-Bones, line 918: I normally prefer a static readonly `Default` field.

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -914,5 +895,49 @@ public abstract class ModPlayer : ModType
...
+				return new Regen { deltaRegen = 0, deltaRate = 0, multiplyRegen = 1, maxZeroRegen = false };
+			}
+			public void ResetVals() {
+				deltaRate = 0;
+				deltaRegen = 0;
+				multiplyRegen = 1;
+				maxZeroRegen = false;
+			}
```
Chicken-Bones, line 926: This isn't needed. Better to just assign `Default`

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -914,5 +895,49 @@ public abstract class ModPlayer : ModType
...
+		}
```
Chicken-Bones, line 927: class should have a `Combine` method like `DamageModifier`

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -11,6 +11,9 @@ public partial class Player {
 
 		public HashSet<int> nearbyModTorch = new HashSet<int>();
 
+		internal ModPlayer.Regen modManaRegen = ModPlayer.Regen.Create();
+		internal ModPlayer.Regen modLifeRegen = ModPlayer.Regen.Create();
```
Chicken-Bones, line 15: Do these actually need to be fields? They're only used in the one method, and they're `internal` so they should be local variables instead.

ExampleMod/Common/Players/ExampleRegenPlayer.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Terraria;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	public class ExampleRegenPlayer : ModPlayer
+	{
+		public override ModPlayer.Regen ModifyLifeRegen() {
+			var modifiers = ModPlayer.Regen.Create(); // Creates the struct with all values configured to do no modifications.
+			modifiers.deltaRate = 3f; // Increases the rate at which player reaches maximum natural life regeneration for every multiple of 300 up to 3600. Positives for buffs, negatives for debuffs.
```
Chicken-Bones, line 10: comments about how the field works above the line. Comments about exactly what this line does, on the line.

I can't actually tell what this line does... it's not a multiple of 300 so.. nothing?

ExampleMod/Common/Players/ExampleRegenPlayer.cs
```diff
@@ -0,0 +1,25 @@
...
+			modifiers.deltaRegen = 6f; // Increases the life per second recovered by typically 6 / 2 per second. Positives for buffs, negatives for debuffs.
```
Chicken-Bones, line 11: why is everything /2 here? Can we make it more sensible?

ExampleMod/Common/Players/ExampleRegenPlayer.cs
```diff
@@ -0,0 +1,25 @@
...
+			modifiers.multiplyRegen = -0.2f; // Decreases effective regen to 20% potency and negates it. Typically positive and >1 for strong buffs, and negative and closer to zero for weak debuffs.
```
Chicken-Bones, line 12: negative multipliers seem like a fundamentally bad idea... since if two mods both apply them....

ExampleMod/Common/Players/ExampleRegenPlayer.cs
```diff
@@ -0,0 +1,25 @@
...
+			return modifiers; // Return your modifiers 
+		}
+
+		public override ModPlayer.Regen ModifyManaRegen() {
+			var modifiers = ModPlayer.Regen.Create(); // Creates the struct with all values configured to do no modifications.
+			modifiers.deltaRate = -3f; // Increases the delay before a player regenerates mana. Positives for buffs, negatives for debuffs.
+			modifiers.deltaRegen = -2f; // Decreases the mana per second recovered by 2 / 2 per second. Positives for buffs, negatives for debuffs.
+			modifiers.multiplyRegen = 0.8f; // Typically use positive values. Decreases effective regen to 80% potency.
+			modifiers.maxZeroRegen = true; // Declares that you want effective regen, after handling all debuffs and prior to multiplication, to be less than or equal to zero. Typically true for debuffs, false for buffs.
```
Chicken-Bones, line 21: should just name this `stopRegen` or something? if it was just life I'd call it `disableHealing`.

ExampleMod/Common/Players/ExampleRegenPlayer.cs
```diff
@@ -0,0 +1,25 @@
...
+			return modifiers; // Return your modifiers
```
Chicken-Bones, line 22: There should probably be a way to alter/limit vanilla regen that isn't additive. If two mods try and cancel vanilla regen, simultaneously, they'll probably end up causing damage to the player.

Chicken-Bones (changes requested):
Good direction

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
Solxanich: Fair. No sense remaking the wheel every time.

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
Solxanich: Per original post, couldn't find that class. Was it changed or removed in the damage class rework?
Regardless, combine method is self-explanatory - combine the structs. 
Only thing I miss is name consistency : (

patches/tModLoader/Terraria/Player.TML.cs
Solxanich: No, no they don't, I need to watch for that better instead of fielding things all the time to start. Fixed.

ExampleMod/Common/Players/ExampleRegenPlayer.cs
Solxanich: Fixed. That line did nothing - that's why I had asked in Discord for anyone with an opinion on what the rate should be doing relative to their expectations.
Now its setup meaningfully, I think.

ExampleMod/Common/Players/ExampleRegenPlayer.cs
Solxanich: I've improved the documentation after rewriting some things further. The /2 was what we originally had.
There is a fair few random scalars in vanilla, so I'm finding hard pressed not to comment that it has the scalars (although notably, the new documentation doesn't rely on actually knowing the scalars)

ExampleMod/Common/Players/ExampleRegenPlayer.cs
Solxanich: This is an interesting one. 
We allowed this before, so I see no reason not to allow it now.
If both apply a negative, then they cancel each other ( and would have done so previously anyway).
I've updated the example to not use negatives, but just warn in the comment about what negatives in the field would do, mentioning both mod stacking and inversion. Some mod may want a debuff that inverts damaging debuffs and healing natural regen (ie some new player race type thing?)

ExampleMod/Common/Players/ExampleRegenPlayer.cs
Solxanich: Hmm. But wouldn't that still happen today? IE this doesn't make that worse.
That said, I can see the concept is to disable vanilla regen so that you can be confident its not involved.
The issues comes in if you then expect to also be able to disable all other mods' regen modifications in addition to vanilla, which I'd rather not go down that rabbit hole. 
Let mods decide if they want to be compatible, or set multiplyRegen to zero if they want to use their own completely separate system, called from the hook, that overrides everything.

Solxanich:
there probably is an issue in that I don't round it for bounding and just check bounds... I'll have to fix that probs

Solxanich:
I'm going to pause and leave this open; I think mixing in life regen hooks made things more complicated.
I'm going to fresh start a PR with just mana regen hooks, and once we're happy, than I can return and look at if life regen needs a similar update.

Solxanich:
I'm closing this as I don't believe I'll be returning to this PR; I'd prefer to continue with the new one having compared and thought on it a bit.
Some good lessons came out of this, and I'll keep the branch around to re-use good parts of as a refine the other pure Mana hooks PR.

PR #1349, opened 2021-03-04
Fix #1345

### LINKED ISSUE #1345: MusicID numbers are off by 1
This may be related to the Music Update for 1.4, but currently all MusicIDs are 1 too high. MusicIDs should now start with 0 (Overworld Day) up to 40 (Old Ones Army).

Chik3r:
### How did you fix the bug?
decreased the ids by 1

no


Chik3r:
Seems that it was correct before

PR #1355, opened 2021-03-11
Fixes individual armor vanity pieces not being updated

ghost:
### What is the bug?

Individual armor vanity pieces were not receiving `UpdateVanity` calls, because of being skipped by an (unintentional) magic number.

### How did you fix the bug?

Bug has been fixed by using a predefined constant to start looping for vanity piece checks using `Player.armor`.

Possibly. There's currently a constant being used in the initialization of the `for` loop in question, as well as `armor.Length`.
These numbers were 'magic' before, and if so required can be reversed to being just numbers (with the exception of the `13` becoming a `10` for proper vanity looping).

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -1555,11 +1555,15 @@
  			for (int m = 3; m < 10; m++) {
  				if (armor[m].wingSlot > 0 && IsAValidEquipmentSlotForIteration(m)) {
  					if (!hideVisibleAccessory[m] || (velocity.Y != 0f && !mount.Active))
-@@ -9305,12 +_,15 @@
+@@ -9305,12 +_,19 @@
  					wingsLogic = armor[m].wingSlot;
  				}
  			}
 +			*/
++
++			for (int n = 10; n < 13; n++) {
++				ApplyEquipVanity(n, armor[n]);
++			}
  
  			for (int n = 13; n < 20; n++) {
  				if (IsAValidEquipmentSlotForIteration(n))
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -1562,7 +1562,7 @@
 +			*/
 +
 +			for (int n = 10; n < 13; n++) {
-+				ApplyEquipVanity(n, armor[n]);
++				ItemLoader.UpdateVanity(armor[n], this);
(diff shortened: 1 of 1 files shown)


Solxanich:
Q: Is this a vanilla bug?  
11- 13 is armor
13- 20 is accessories

If vanilla is skipping vanity updates for 11-13, we should let Re;Logic know and have them quick patch before 1.4.2

Eldrazi:
Vanilla does not have any single pieces of vanity armor that update on their own; therefore I'd say this is no vanilla bug.

It this is not an acceptable way of trying to implement this, I'll implement the suggested logic, no problem 👍

Eldrazi:
I'm thinking this would be more according to the aforementioned logic then: a loop right before vanilla accessory vanity updates, on indices 10, 11, and 12.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1555,11 +1555,15 @@
...
  					if (!hideVisibleAccessory[m] || (velocity.Y != 0f && !mount.Active))
-@@ -9305,12 +_,15 @@
+@@ -9305,12 +_,19 @@
  					wingsLogic = armor[m].wingSlot;
  				}
  			}
 +			*/
++
++			for (int n = 10; n < 13; n++) {
++				ApplyEquipVanity(n, armor[n]);
```
Solxanich, line 1565: RE: Verify no reason to call vanilla vanity accessory updates when looping through the proper armor slots.
If no need, than just call itemloader.updatevanity and reduce # of calls

Eldrazi: Fair, should make more sense now 👍

Solxanich: Per discord, add a one line comment above the added for loop describing why it's there.
It would be good to drop a line confirming the "hideVisibility" array goes unused for slots 10-12 for sanity's sake, although I'd swear its unused.

Can merge when comment is on there, ping Mirsario in discord when done.

Some context:
Discord discussed if needs to check hideVisibility, but given this is helm, chest, greaves, don't believe need to check for the hook as player doesn't set it. Don't need to check accessory compat either, accessories control their compat, just came up as confusion

Mirsario (changes requested):
From discord: It seems more correct to remove the added code and just move all UpdateVanity calls to UpdateVisibleAccessory. Test this afterwards.

PR #1358, opened 2021-03-13
Modifiable Menu Buttons Through ModMenu

Steviegt6:
### What is the new feature?
This PR introduces the ability to modify the buttons on the main menu through `ModMenu`. Menu buttons will only be modified by the currently selected ModMenu, removing any chance of conflicts between mods.

This should be part of tML since `ModMenu` already gives you the ability to incorporate vustom drawcode and other neat stuff. This PR simply builds on that by allowing people to modify the menu buttons easily as well, giving modders more freedom.

The only real alternative design is making it not part of `ModMenu`, but instead `Mod`/`ModSystem`. (See my old PR: #1222)

See `ExampleModMenu.cs`.

### ExampleMod updates
Added an example usage of this new feature to `ExampleModMenu`, which makes the first button always rainbow and adds a new button to the main menu with a modified size and color that adjusts the offset of other buttons when hovered and resets it when clicked.

Example in action: https://gyazo.com/90ae03812a0eba465f23651f9d28ecf0.mp4

### DIFF
--- a/ExampleMod/Content/ExampleModMenu.cs
+++ b/ExampleMod/Content/ExampleModMenu.cs
@@ -50,10 +50,9 @@ namespace ExampleMod.Content
 			buttons.First().color = Main.DiscoColor;
 
 			// Use a foreach loop to modify all buttons in the list
-			foreach (MenuButton button in buttons) {
+			foreach (MenuButton button in buttons)
 				// Forcefully set the yOffsetPos of every button to _globalButtonOffset, which is modified through our ExampleButton below
 				button.yOffsetPos = _globalButtonOffset;
-			}
 
 			// Make sure we're on the title screen menu
 			// If we aren't, then return and don't execute the code below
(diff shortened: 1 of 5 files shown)


Solxanich:
So, to attempt to summarize the items from Discord:
1) On hold for other items of higher value relative to 1.4 milestone (ModAchievements, ModTranslations), and until a consensus direction is available
2) Consider splitting the scope of PR into two different target sets:
A) Modifying layout aspects of menu buttons (size, location, colour, text style, etc.). Not confirmed if wanted
B) Example Mod example of using Draw hook to add a ThemeMenuButton to the theme? Not confirmed if wanted.
C) An ability to add ExtrasMenuButton to a 'Mod Extras' menu. The Mod Extras menu accessible from main menu, should be obvious, and open a UI listing all added extra menu buttons. Not confirmed if wanted
3) A possible PR can be done to move Mod Sources menu into Mods menu
4) A possible PR can be considered for bumping the minimum height aspect up to 720 from 600 as it kinda should be fair in this decade to do so

Steviegt6:
Closing until further notice. ^^

PR #1359, opened 2021-03-15
Player Spawn Biome Check Coordinates Fix

Solxanich:
### What is the bug?
Currently, the player.spawn calls for return from death and recall items occur after the player biome has already been updated. This leads to at least one frame where the player is in the wrong biome; consequently there are rare occurences of not-yet spawned enemies from the previous biome 'following the player' and spawning in the new biome. 
In 1.4.2, a new method called player.forceUpdateBiomes() will be introduced to fix this vanilla bug.
We will need to add PlayerHooks.UpdateBiomes and related hooks currently in BiomeUpdate() to this new method to guarantee we receive the fix.

### How did you fix the bug?
Can't Yet.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -2974,6 +2974,16 @@
  				dead = false;
  				immuneTime = 0;
  			}
+@@ -26441,6 +_,9 @@
+ 			}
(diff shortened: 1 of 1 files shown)


Solxanich:
Close as 1.4.2 issue will remind us of this when relevant.

Issue #1106, opened 2020-09-26
"CanStackInWorld" hook for ModItem


TotallyNotAScalie:
### Description
A hook for ModItem which returns a boolean, and having a parameter for the item the given item instance is attempting to stack with. The return will determine if the two item instances should combine into one (stacking) with each other. It would also enable modders to create item => item collision interactions for various applications like crafting systems or NPC spawning.

### What does this proposal attempt to solve or improve?
Modders may wish to have items which have different stacking behavior in inventories and in the world, as it is currently the only way to stop items from stacking in-world is to set their maximum stack size to 1 or to hackily change it depending on it's state. This would allow for features such as drops that should visually appear seperate on the ground but stack in the inventory like currencies, more freedom in designing visuals related to item spawning, and possible applications for interactions between items in-world other than stacking.

### Which (other) solutions should be considered?
Possibly a more general hook concerning stack-ability, for additional control over the stacking behavior of items in inventories.

Chicken-Bones:
Can you give us some specific examples of some items which could benefit from this? Rather than general possibilities?

TotallyNotAScalie:
-An item which, when dropped, floats in a circle around a point, without combining in stack with others of the same type
-A item storing special data, such as a charge level, that can only stack with items of identical charge levels
-Some sort of "crystal shard" item that would drop in a large quantity on destroying a crystal, with unique sprites on the ground that would not stack, but would stack in the inventory under one sprite 
-Dropping some sort of "fire core" and "ice core" items on the ground together to create an explosion

Chicken-Bones:
Excellent. Suggestion approved

JavidPack:
So, we've had troubling discussions on how to implement a previously suggested CanStack hook (more generally, such as in the inventory.). This hook will have to come after we figure that one out, and piggy back on the logic. There are a lot of issues with stacking stackable items, and a large amount of logic that will need to be updated as well.

Chicken-Bones:
I feel like an 'in world' stacking hook could be implemented separately though. It could only prevent items which could previously stack from stacking in the world, and allow for effects instead (if called on all disparate sets of items). The problem with general purpose can stack is that it's called in many different scenarios, but in this case it should be fine

Itorius:
I also had the idea of implementing a CanStack hook for modded items. However due to the insane amounts of places where that would have to be implemented I decided not working on it (at least just yet).

PR #1362, opened 2021-03-16
Added a hook to prevent items from stacking in the world

MrSlimeDude:
### What is the new feature?
This PR adds a hook for ModItem which returns a boolean which allows the modder to allow items to stack in the inventory but not in the world.
There is an issue for this PR: #1106 
Because it will allow mods to create unique in world item interactions based on certain criteria. (See issue)
See issue
```C#
using Terraria.ModLoader;

public class ExampleItem : ModItem
{

	...

	public override bool CanStackInWorld(Item item2) {
		return false; //Prevents the item from stacking in the world but still allows it to stack in the inventory 
	}

}

```
### ExampleMod updates

None

### DIFF
--- a/patches/tModLoader/Terraria/Item.cs.patch
+++ b/patches/tModLoader/Terraria/Item.cs.patch
@@ -676,6 +676,16 @@
  				base.velocity.Y += gravity;
  				if (base.velocity.Y > maxFallSpeed)
  					base.velocity.Y = maxFallSpeed;
+@@ -45614,6 +_,9 @@
+ 					num2 *= 2;
+ 
+ 				if (num < (float)num2) {
++					if (!ItemLoader.CanStackInWorld(this, item))
++						continue;
++
+ 					position = (position + item.position) / 2f;
+ 					velocity = (velocity + item.velocity) / 2f;
+ 					int num3 = item.stack;
 @@ -45935,6 +_,9 @@
  		}
  
--- a/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
@@ -481,10 +481,11 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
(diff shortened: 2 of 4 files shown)


Solxanich:
Good base to the PR. At a glance, my main comment is to rename item1 and item2 parameters to have meaningful names. Its indistinguishable in globalItem and confusing if you aren't familiar with how its setup internally.


Beyond the code, a couple of questions of consideration for the formal reviewer and author alike, can't tell from the issue;
1) Should the ability to force items to stack in world that don't stack in inventory also exist? And if so, how manage conflicts? I only forsee this being used to improve end-user performance,  so I don't think it's needed.
2) Should the stacking check include context sensitivity? IE should we allow specifying only prevent stacking if the player dropped the item AND not if it dropped from other sources?
3) As the outcome raised is merging two in world items in to something different,
A) should this PR include the additional hook/modifications to hooks to facilitate?
B) does tML need to consider how to handle in-world crafting recipe conflicts?

Chicken-Bones (approved):
The position of this hook means it will only be called for 2 items of the same `type`. Please add this to the documentation for the hooks. 

In response to @Solxanich 's questions
1. I don't see a use case for forced stacking at this stage
2. I don't think additional context is required. This hook covers the current use cases.
3. Merging 2 items for a 'recipe' or 'interaction' would require a different hook, and should be designed in a declarative manner similar to other recipes, where tML chooses the recipe to 'apply' and mods can see which recipes exist.

MrSlimeDude:
@Solxanich  I agree with your naming suggestion but was unable to think of any better names myself so I tried to explain the name in the method documentation. If you are able to think of one I would be happy to rename the parameters in a new commit. Thank you for the feedback.

Solxanich:
The documentation is good and clarifies well enough for me.
I'll drop a line in the discord to let someone know its ready for merge as it looks like all feedback is addressed. 
Thanks for the contribution! If you're interested in contributing more, I'd recommend dropping a line in our discord to have yourself added to the 'collaborators' chat as its a good place to get modular feedback on larger PRs

PR #1371, opened 2021-03-20
More improvements related to the flow of time

ThomasThePencil:
- The rate at which time flows and the rate at which tiles in the world update may both now be set to decimal values.
- The rate at which events in the world occur (e.g. fallen star/fairy spawns, weather changes) can now be set independently of `Main.dayRate`, although by default it will have the same value `Main.dayRate` would have after vanilla calculations.
- Updated `ModSystem.ModifyTimeRate` to better reflect the new and improved system of time distortion.

The main benefit of this change set is that it becomes possible to **slow the flow of time (and tile/world updates)** compared to the rate at which time normally passes in the game (1 in-game minute per 1 real-life second, or 1 min/sec). For example, setting `Main.dayRate` to 0.5 causes the game to move at 0.5 min/sec, causing the game's time to move half as quickly as normal. Furthermore, you are now able to decide if events should update faster or slower than they normally would with a certain `Main.dayRate` value --- this should grant much more flexibility in how time is handled.

As extra clarification, **this does not hinder or otherwise affect aspects of the game that would not normally be affected by changing the flow of time**. The only changes made were changes which were required to ensure things whose behavior DOES change with the flow of time still behaved properly following the conversion of `Main.dayRate` and related variables.

"Why not use floats?"
Doubles are much more precise (the relative inaccuracy of float values is why the Lunar Pillars are able to slowly sink into the ground as they hover), and they are already used by a few important time-related fields (most notably `Main.time`, the field responsible for keeping track of the time universally in the current world) in the base game. Using doubles for the rest of the fields, as such, seems like the most appropriate option. While you could *technically* argue that floats would take up slightly less patch size, it seems counterintuitive to consciously give up precision and consistency (especially on something as fundamental to the game as the flow of time) simply to lower the number of lines shown by the patch by a relatively-insignificant amount, as the change would hardly prevent any of the significant changes from needing to be made *anyway*.

"Isn't `Main.dayRate` cast to a float a few times?"
For rendering the sky, yes. These casts aren't really relevant to the PR, though, as the difference made by changing them at all would be so minimal it's practically unnoticeable, and since you can safely cast a double to a float, it seems unnecessary to change them when they already work fine after the change.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
@@ -1,0 +1,11 @@
+--- src/Terraria/Terraria/GameContent/Events/MysticLogFairiesEvent.cs
++++ src/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs
+@@ -8,7 +_,7 @@
+ 	public class MysticLogFairiesEvent
+ 	{
+ 		private bool _canSpawnFairies;
+-		private int _delayUntilNextAttempt;
++		private double _delayUntilNextAttempt;
+ 		private const int DELAY_BETWEEN_ATTEMPTS = 60;
+ 		private List<Point> _stumpCoords = new List<Point>();
+ 
--- a/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
@@ -34,7 +34,7 @@
  				if (_delayUntilNextAttempt == 0) {
  					_delayUntilNextAttempt = 60;
  					TrySpawningFairies();
-@@ -60,10 +_,16 @@
+@@ -60,10 +_,17 @@
  
  			int oneOverSpawnChance = GetOneOverSpawnChance();
  			bool flag = false;
@@ -49,6 +49,7 @@
 -					flag = true;
 +						flag = true;
 -					break;
++						_timePass -= Math.Floor(_timePass);
 +						break;
 +					}
 +					_timePass -= 1;
--- a/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
@@ -34,25 +34,20 @@
  				if (_delayUntilNextAttempt == 0) {
  					_delayUntilNextAttempt = 60;
  					TrySpawningFairies();
-@@ -60,10 +_,17 @@
+@@ -60,12 +_,15 @@
  
  			int oneOverSpawnChance = GetOneOverSpawnChance();
  			bool flag = false;
 +
 +			_timePass += Main.desiredWorldEventsUpdateRate;
-+			if (_timePass >= 1) {
-+				double _timePassReference = _timePass;
 -			for (int i = 0; i < Main.dayRate; i++) {
-+				for (int i = 0; i < (int)_timePassReference; i++) {
--				if (Main.rand.Next(oneOverSpawnChance) == 0) {
-+					if (Main.rand.Next(oneOverSpawnChance) == 0) {
--					flag = true;
-+						flag = true;
--					break;
-+						_timePass -= Math.Floor(_timePass);
-+						break;
-+					}
-+					_timePass -= 1;
++			for (int i = 1; i < (int)_timePass; i++) {
+ 				if (Main.rand.Next(oneOverSpawnChance) == 0) {
+ 					flag = true;
+ 					break;
  				}
  			}
++			_timePass %= 1.0;
  
+ 			if (!flag)
+ 				return;
--- a/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
@@ -1,6 +1,6 @@
 --- src/Terraria/Terraria/GameContent/Events/MysticLogFairiesEvent.cs
 +++ src/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs
-@@ -8,7 +_,7 @@
+@@ -8,14 +_,16 @@
  	public class MysticLogFairiesEvent
  	{
  		private bool _canSpawnFairies;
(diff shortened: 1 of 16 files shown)


Mirsario (changes requested):
Use floats, double is excessive and rare.

patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
```diff
@@ -0,0 +1,57 @@
...
+-			for (int i = 0; i < Main.dayRate; i++) {
++				for (int i = 0; i < _timePassReference; i++) {
+-				if (Main.rand.Next(oneOverSpawnChance) == 0) {
++					if (Main.rand.Next(oneOverSpawnChance) == 0) {
+-					flag = true;
++						flag = true;
+-					break;
++						break;
++					}
++					_timePass -= 1;
```
Solxanich, line 54: Does this need to reduced prior to break as well? Otherwise _timePass would have be >=1 on the next cycle as well.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1543,15 +1551,43 @@
...
+-			for (int l = 0; l < Main.dayRate; l++) {
 -				float num6 = Main.maxTilesX / 4200;
++			timePass += Main.desiredWorldEventsUpdateRate;
++			if (timePass < 1)
++				return;
++
++			double timePassReference = timePass;
++			for (int l = 0; l < timePassReference; l++) {
++				timePass -= 1.0;
 +				float num6 = Main.maxTilesX / 4200f; // Selfish fix for falling stars on extra small worlds
```
Solxanich, line 1576: This isn't your change, but I'm seeing that we are removing and than adding the same line. (1568 and 1576 on this).
Mind fixing?

Solxanich (commented):
Mostly a quick sanity check on the Github code.
High level comments:
1) Should probably bound dayRate between 86400 and 1E-38 to avoid edge cases like div by Zero
2) Will likely need a legacy flag as you are changing how data is saved and stored (obviously can't read a stored int32 as 64 bit double). I also don't see the IO/WorldFile having its Reader.Write changed so please double check that.

patches/tModLoader/Terraria/GameContent/Events/MysticLogFairiesEvent.cs.patch
Chicken-Bones: Suggested: remove `_timePassReference` and `if (_timePass >= 1)`, to reduce indenting. The `(int)` cast will floor, and the for loop won't execute if `_timePass < 1`.

After the loop, do `_timePass %= 1` to only keep the remainder.

patches/tModLoader/Terraria/GameContent/Events/Sandstorm.cs.patch
```diff
@@ -0,0 +1,39 @@
...
++							for (int i = 0; i < (int)timePassReference; i++) {
+-							if (Main.rand.Next(num) == 0)
++								if (Main.rand.Next(num) == 0)
+-								StartSandstorm();
++									StartSandstorm();
++								timePass -= 1.0;
++							}
+ 						}
+ 					}
+ 				}
```
Chicken-Bones, line 39: Same comment as MysticLogFairiesEvent.cs.patch

patches/tModLoader/Terraria/Graphics/Effects/SkyManager.cs.patch
```diff
@@ -1,15 +1,58 @@
...
++			for (int i = 0; i < (int)_timePassReference; i++) {
+ 				LinkedListNode<CustomSky> linkedListNode = _activeSkies.First;
+ 				while (linkedListNode != null) {
+ 					CustomSky value = linkedListNode.Value;
+@@ -34,6 +_,7 @@
+ 
+ 					linkedListNode = next;
+ 				}
++				_timePass -= 1.0;
+ 			}
```
Chicken-Bones, line 42: Same as MysticLogFairiesEvent.cs.patch

patches/tModLoader/Terraria/Graphics/Effects/SkyManager.cs.patch
```diff
@@ -1,15 +1,58 @@
...
+ 			return MathHelper.Clamp(num, 0f, 1f);
++		}
 +
 +		internal void DeactivateAll() {
 +			foreach (string key in _effects.Keys) {
 +				if (this[key].IsActive())
 +					this[key].Deactivate(new object[0]);
 +			}
-+		}
+ 		}
```
Chicken-Bones, line 56: diff derp?

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1020,6 +1056,15 @@
  			else if ((double)player[myPlayer].position.Y >= worldSurface * 16.0 + (double)(screenHeight / 2) && !WorldGen.oceanDepths((int)(screenPosition.X + (float)(screenWidth / 2)) / 16, (int)(screenPosition.Y + (float)(screenHeight / 2)) / 16)) {
  				if (player[myPlayer].ZoneHallow) {
  					newMusic = 11;
+@@ -9868,7 +_,7 @@
+ 			if (cloudBGActive > 0f)
+ 				cloudBGActive = 0f;
+ 
+-			if (cloudBGActive == 0f && rand.Next((int)((float)(num2 * 12 / ((dayRate == 0) ? 1 : dayRate)) / num3)) == 0) {
++			if (cloudBGActive == 0f && rand.Next((int)((float)(num2 * 12 / ((dayRate == 0) ? 1E-38 : dayRate)) / num3)) == 0) {
```
Chicken-Bones, line 1064: please explain

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1082,6 +1127,27 @@
...
++
++			timePass += dayRate;
++			if (timePass < 1)
++				goto skipWeatherUpdates;
++			double timePassReference = timePass;
+-			for (int i = 0; i < dayRate; i++) {
++			for (int i = 0; i < (int)timePassReference; i++) {
+ 				UpdateWeather(gameTime);
++				timePass -= 1.0;
+ 			}
```
Chicken-Bones, line 1145: Same as MysticLogFairiesEvent.cs.patch

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3182,6 +3248,19 @@
...
+@@ -47097,9 +_,9 @@
+ 			if (gameMenu || netMode == 2)
+ 				bgTopY = -200;
+ 
+-			int num3 = dayRate;
++			double num3 = dayRate;
+-			if (num3 < 1)
++			if (num3 < 0)
+-				num3 = 1;
++				num3 = 0;
```
Chicken-Bones, line 3260: comments required

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3433,6 +3525,19 @@
...
+@@ -48991,9 +_,9 @@
+ 			if (invasionX == (double)spawnTileX)
+ 				return;
+ 
+-			float num = dayRate;
++			double num = dayRate;
+-			if (num < 1f)
++			if (num < 0)
+-				num = 1f;
++				num = 0;
```
Chicken-Bones, line 3537: comments required

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3449,6 +3554,77 @@
...
++						else if (cloudBGActive >= 1f && rand.Next((int)(num2 * 4.25)) == 0)
+ 							StartRain();
+-						else if (ladyBugRainBoost > 0 && rand.Next(num2) == 0)
++						else if (ladyBugRainBoost > 0 && rand.Next((int)num2) == 0)
+ 							StartRain();
+ 					}
+ 
+ 					if (!raining && !NPC.BusyWithAnyInvasionOfSorts()) {
+-						int num3 = (int)(1728000.0 / (double)dayRate);
++						int num3 = (int)(1728000.0 / (double)((dayRate == 0) ? 1E-38 : dayRate));
```
Chicken-Bones, line 3599: this will just overflow... (maybe to int.MaxValue?) still not ideal

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1543,15 +1551,43 @@
...
  					continue;
+@@ -45702,8 +_,8 @@
+ 			}
+ 		}
+ 
+-		public static int GetWorldUpdateRate() {
++		public static double GetWorldUpdateRate() {
+-			int result = Math.Min(Main.desiredWorldTilesUpdateRate, 24);
++			double result = Math.Min(Main.desiredWorldTilesUpdateRate, 24);
+ 			if (CreativePowerManager.Instance.GetPower<CreativePowers.FreezeTime>().Enabled)
```
Chicken-Bones, line 1588: In all the places where you have 1E-38, consider investigating how freezeTime works, and using that instead.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Graphics/Effects/SkyManager.cs.patch
ThomasSpeedrunner: ...how did that even

okay

PR #1378, opened 2021-03-22
Added Example Minecart (Updated with commetns now)

FullOfXP:

Custom minecart was a thing I heard alot of ppl ranting about, there was no ExampleMinecart, nothing on the internet... and the only mod that iirc had it was Thorium, so I spent some time digging in terraria and I found all of the lines needed to make a custom minecart happen.

Example minecart is one of the most faqs in `#help-with-code`

I dont undetrstand what `mountData.jumpBoost` does so if someone wants to comment on that go ahead

image_2021-03-22_223040


ExampleMod/Buffs/ExampleMinecartBuff.cs
```diff
@@ -0,0 +1,22 @@
...
+		{
+			DisplayName.SetDefault("Example Minecart");
+			Description.SetDefault("Example your rails");
+			Main.buffNoTimeDisplay[Type] = true;
+			Main.buffNoSave[Type] = true;
+		}
+
+		public override void Update(Player player, ref int buffIndex)
+		{
+			player.mount.SetMount(ModContent.MountType<Mounts.ExampleMinecart>(), player);
```
Steviegt6, line 18: I feel like a least a single comment explaining the purpose of the buff would be helpful.

ExampleMod/Mounts/ExampleMinecart.cs
```diff
@@ -0,0 +1,64 @@
...
+			mountData.flyingFrameCount = 0;
+			mountData.flyingFrameDelay = 0;
+			mountData.flyingFrameStart = 0;
+			mountData.inAirFrameCount = 0;
+			mountData.inAirFrameDelay = 0;
+			mountData.inAirFrameStart = 0;
+			mountData.idleFrameCount = 1;
+			mountData.idleFrameDelay = 10;
+			mountData.idleFrameStart = 0;
+			mountData.idleFrameLoop = false;
```
RighteousRyan1, line 56: Not exactly sure, but maybe explain what a few of these do?

ExampleMod/Items/MinecartKeys.cs
```diff
@@ -0,0 +1,42 @@
...
+			item.width = 20;
+			item.height = 30;
+			item.useTime = 20;
+			item.useAnimation = 20;
+			item.useStyle = ItemUseStyleID.SwingThrow;
+			item.value = 30000;
+			item.rare = ItemRarityID.Green;
+			item.UseSound = SoundID.Item79;
+			item.noMelee = true;
+			MountID.Sets.Cart[ModContent.MountType<ExampleMinecart>()] = true; //this line makies it automatically equip into aminecart slot
```
Steviegt6, line 29: 1. Sets should be set in SetStaticDefaults for consistency's sake and because SetStaticDefaults is called first.
2. Why not do this in ExampleMinecart's code?

ExampleMod/Mounts/ExampleMinecart.cs
```diff
@@ -0,0 +1,64 @@
...
+			if (Main.netMode != NetmodeID.Server)
+			{
+				mountData.textureWidth = mountData.frontTexture.Width;
+				mountData.textureHeight = mountData.frontTexture.Height;
+			}
+		}
```
Steviegt6, line 62: Extremely noticeable lack of comments. Anyone without any prior mount experience won't understand this at all. I certainly don't know the purpose of some of this stuff, even with reasonably guessing from names.

Steviegt6 (changes requested):

ExampleMod/Mounts/ExampleMinecart.cs
FullOfXP: time to comment every single line

Issue #1383, opened 2021-03-25
[1.3 and 1.4] Modded torches lack "smart cursor" support by default


direwolf420:
### Description
Torches have a few special behaviors in vanilla, that tml mostly supports by default (no modder boilerplate required), except "smart cursor". Providing this support on tmls side by default will bring torches fully in line with vanilla behavior.

### What does this proposal attempt to solve or improve?
What is supported by tml:
* Dry torch, water torch and glowstick handling with "smart select" (selected item depends on the tile the player is mouseovering)
* Torch flames

What is not supported by tml:
* "Smart cursor" (the game auto-aims at a suitable tile position to do things in - in the torches case, place it at an appropriate location around the player with the least amount of light)

### Which (other) solutions should be considered?
The modder writing their own custom smart cursor code (status quo), no consistency with other mods or vanilla guaranteed.

### Bonus
1.3:
* The method responsible for this is `Player.SmartCursor_Torch`, which currently only works with the vanilla torch tile type (4).

1.4:
* There exists `ItemID.Sets.WaterTorches`, which is not compatible with modded torches as its always checked against the vanilla torch tile type (4) first.
* The method responsible for this is `SmartCursorHelper.Step_Torch` and works the same as the 1.3 equivalent.

Solxanich:
@direwolf420 - do you want to close this now that the PR was merged and create a new issue for the leftovers, 
OR do you want to leave open with this comment stating as a TODO on this?

"Modded torches should never be added to ItemID.Sets.WaterTorches or ItemID.Sets.Torches at the moment because once they are then some Terraria code runs that calls TorchID.TorchColor for out of range values. Work would have to be done to allow modded toches to be included in the torch sets and would require changing either how modded torches are implemented or how Terraria handles torch lighting. For some reason, ModItem.AutoLightSelect seems to be completely unused in the code other than the definition. I feel that implementing AutoLightSelect or allowing modded torches in ItemID torch sets might be best for a PR of its own where a global hook could potentially be implemented with ModItem.AutoLightSelect as well to be consistent with other hooks in ModItem or modded torches are overhauled to be more similar to Terraia's torches which could also reduce torch boilerplate code."

direwolf420:
I suppose I could make a second issue that adresses the 1.4 stuff

PR #1391, opened 2021-03-26
Allow modded torches to support smart cursor

MrSlimeDude:
### What is the new feature?
This allows modded torches to be placed by smart cursor without any extra work from the modder.
There is an issue for it: #1383 
There really is no reason that modders should have to do this themselves as it is an easy fix on the tModLoader side.
None other than forcing modders to do it themselves.
The only thing that modders have to do to use this new feature is put `TileID.Sets.Torch[Type] = true;` in their torch's SetDefaults method which they should already be doing.
### ExampleMod updates
None

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/SmartCursorHelper.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/SmartCursorHelper.cs.patch
@@ -10,3 +10,12 @@
  						break;
  					case 314:
  						if (providedInfo.player.gravDir == 1f)
+@@ -2149,7 +_,7 @@
+ 			int reachableEndY = providedInfo.reachableEndY;
+ 			_ = providedInfo.screenTargetX;
+ 			_ = providedInfo.screenTargetY;
+-			if (providedInfo.item.createTile != 4 || fX != -1 || fY != -1)
++			if (providedInfo.item.createTile < 0 || !TileID.Sets.Torch[providedInfo.item.createTile] || fX != -1 || fY != -1)
+ 				return;
+ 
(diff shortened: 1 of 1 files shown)


direwolf420:
Nice, have you looked into the functionality of `ItemID.Sets.WaterTorches` and if it's necessary to expand its scope to support modded entries aswell? Or is everything already covered by overriding `ModItem.AutoLightSelect` and setting `wetTorch` to true?

MrSlimeDude:
Modded torches should never be added to `ItemID.Sets.WaterTorches` or `ItemID.Sets.Torches` at the moment because once they are then some Terraria code runs that calls `TorchID.TorchColor` for out of range values. Work would have to be done to allow modded toches to be included in the torch sets and would require changing either how modded torches are implemented or how Terraria handles torch lighting. For some reason, `ModItem.AutoLightSelect` seems to be completely unused in the code other than the definition. I feel that implementing `AutoLightSelect` or allowing modded torches in `ItemID` torch sets might be best for a PR of its own where a global hook could potentially be implemented with `ModItem.AutoLightSelect` as well to be consistent with other hooks in `ModItem` or modded torches are overhauled to be more similar to Terraia's torches which could also reduce torch boilerplate code.

The consensus on Discord was that upper bounds checking was not required when accessing `TileID.Sets` as if it was out of range it would indicate a serious problem with either the mod or tModLoader.

direwolf420:
> For some reason, `ModItem.AutoLightSelect` seems to be completely unused in the code other than the definition.

That's not good, guess it went under during moving patches

PR #1392, opened 2021-03-27
[1.4] Modded ItemTime Fixes

DarkLight66:
### What is the bug?
Some parts setting `Player.itemTime` or otherwise using `Item.useTime` were not accounting for modded use time hooks. Notably, this affected pickaxes , the gravedigger's shovel, hammers when braking walls, and mechanical tools such as drills and chain saws (for different  reasons than the other normal tools).
There was also a case of `Player.itemAnimation` being assigned without accounting for modded use animation hooks, but i'm not sure if this was done on purpose or not due to it being directly inside the `ApplyItemAnimation` method.

### How did you fix the bug?
I searched all instances i could find in the `Player` class of `Item.useTime` being assigned or compared to and changed them so they account for modded hooks, by replacing them with a `ApplyItemTime` call or a `PlayerHooks.TotalUseTime` call depending on context.

Some sections could probably have used `itemTimeMax` instead of calling `PlayerHooks.TotalUseTime`.


Chicken-Bones:
### Partially addressed by:
- #2288 
- ad03f8ce06ea77a02753d380b6e6a9598843ce25
- #3795

### Not addressed:

In `PlaceThing_TryReplacingTiles`
```.patch
-						SetItemAnimation((int)((float)bestPickaxe.useTime * pickSpeed));
+						ApplyItemAnimation(bestPickaxe.useTime * pickSpeed, this, bestPickaxe);
```

In `ItemCheck_UseMiningTools_TryHittingWall`
```.patch
-			itemTime = sItem.useTime / 2;
+			ApplyItemTime(sItem, 0.5f);
```

Magic mirror and teleportation potion hardcoded `itemTime` checks. Note that the proposed implementation calling usetime hooks in the comparison risks breaking the items if the modifiers change during use. Better to compare to a fraction of `itemTimeMax`

PR #1396, opened 2021-03-28
Fixed a crash with ModMenu during mod building

OliHeamon:
### What is the bug?

This bug occurs when a there is a non-vanilla ModMenu active and it is unloaded. This can cause AssetDisposedException's because it attempts to draw the unloaded textures (caused by unloading occurring on a different thread).

### How did you fix the bug?

Setting the ModMenu to MenutML at the beginning of the unload process.

Something like https://github.com/Mirsario/Terraria-BetterRussian/blob/master/BetterRussianMod.cs, which uses a TaskCompletionSource to await the unloading of the menu assets.


Mirsario:
Can be improved further later.

PR #1398, opened 2021-03-28
1.4

Quethed:
Tried to make a set of corruptable blocks, found 2 important code hardcoded.

Adding this support would allow the author to extend the corruption, crimson, or hallow. This would be great for all those custom biomes in mods. Adding support for vanilla blocks is good too.

WorldGen.Convert(int, int, int, int) should have ModWorld support to add extra entries.
WorldGen.GERunner(int, int, float, float, bool) should also have ModWorld support to add extra entries, likely even the same method.
Any more occurrences would help too.

ModTile could be an alternate, but it could make code unnecessarily lengthy and might not be able to support vanilla blocks. GlobalTile is a another good alternate.

I would use this for my newest in development mod that overhauls the spreading biomes. I own the slime biome mod and it could use the ability to convert into a spreading biome too. Thorium and Calamity have an ocean biome. Tremor when it was in development had an arctic glacier biome. Spirit mod when in development had the reach biome and I believe a spirit biome too. Crystillium had a crystal biome. The list is extensive, my point is that this could be a small change for a huge impact.

Example mod should have conversion variants for ExampleBlock and use around this code:
//ModTile.SetDefaults()
TileID.Sets.Corrupt[Type]=true;
TileID.Sets.Crimson[Type]=true;
TileID.Sets.Hallow[Type]=true;
//ModWorld
public override void TileCountsAvailable(int[] tileCounts){
	Main.evilTiles+=tileCounts[mod.TileType("PurpleExampleBlock")];//corrupt variant
	Main.bloodTiles+=tileCounts[mod.TileType("RedExampleBlock")];//crimson variant
	Main.holyTiles+=tileCounts[mod.TileType("PinkExampleBlock")];//hallow variant
}
//The new method idea
//This should handle all vanilla conversion methods and projectiles.
public override void ModifyTileConversionTable(int type, ref int corrupt, ref int crimson, ref int hallow){
	//Starts with all numbers the type.
	//Ends with the conversion chart nesessary.
	//For multiple pure variants, add multiple entries.
	//Conversion toward pure could default to the first entry.
	if(type==mod.TileType("ExampleBlock")){
		corrupt=mod.TileType("PurpleExampleBlock");
		crimson=mod.TileType("RedExampleBlock");
		hallow=mod.TileType("PinkExampleBlock");
	}
}
//Same but for walls
public override void ModifyWallConversionTable(int type,ref int corrupt,ref int crimson,ref int hallow){}

The new method would require small addition documentation.

	-Quethed


JavidPack:
If you wish to make an issue/feature request, please do so at https://github.com/tModLoader/tModLoader/issues . It seems you have mistakenly make a pull request.

Issue #410, opened 2018-09-08
ExampleMod Request: Animated Gun (Vortex Beater)


JavidPack:
An example of an animated gun has been frequently requested. The example would need to document how ExampleGun and itself differ. (ModItem vs ModProjectile)

PR #1401, opened 2021-03-29
Added Example Animated Gun

andrewlord607:
Added an example of Animated Gun like Vortex Beater https://github.com/tModLoader/tModLoader/issues/410
I have given some explanations of how this works in the comments on the code.

### DIFF
--- a/ExampleMod/Content/Items/Weapons/ExampleAnimatedGun.cs
+++ b/ExampleMod/Content/Items/Weapons/ExampleAnimatedGun.cs
@@ -1,0 +1,64 @@
+﻿using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Items.Weapons
+{
+	//This example show how to made an Animated Gun(like Vortex Beater)
+	//The main idea is to create in this script an animated Projectile that will behave like a Gun(ExampleAnimatedGunProjectile)  
+	//Therefore, most of the properties of the weapon are not needed, but we can still use them in the projectile.
+	public class ExampleAnimatedGun : ModItem
+	{
+		public override string Texture => "Terraria/Images/Item_" + ItemID.VortexBeater;
+
+		public override void SetStaticDefaults() {
+			Tooltip.SetDefault("This is a modded animated gun.");
+		}
+
+		public override void SetDefaults() {
+			//Important properties for Animation
+			Item.channel = true; //Set wapon to channel mode. When the attack button is held, we create only one projectile(animated gun projectile).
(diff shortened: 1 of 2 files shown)


Mirsario (changes requested):
Seems like a horrible idea to use projectiles for this. What's stopping people from doing this without them? If there is anything like that, we can just resolve that.

andrewlord607:
> Seems like a horrible idea to use projectiles for this. What's stopping people from doing this without them? If there is anything like that, we can just resolve that.

Yes, I totally agree with you
This is how it is implemented in the vanilla terraria. Only projectiles have animation functionality with _Projectile.frame,Projectile.frameCounter_ and etc.
ModItem does not have such functionality.

I can do this with PostDrawInWorld and spriteBatch.Draw, but it takes a few extra methods to adjust the frame timings (like in Projectiles).
I can also try to do this, but for me it sounds not just like an Example, but like the **implementation of animation in the ModItem class**.
What do you think about this?

andrewlord607:
Okay
I tried diffrent drawing hooks for this and none of them cant draw sprites while the item is in the player's hands (Maybe I'm wrong?).
I think this is the **main reason** why you can't do without a projectile.

JavidPack:
Every single advanced weapon in the game uses held projectiles. Unless item code changed drastically recently, then I think it'll have to be a projectile. These examples are also examples people will go to for examples of deciphering vanilla code. Various reasons for using a projectile include no support for animation on items, and probably a bunch of MP considerations that help the behavior sync and display properly.

Mirsario:
There is nothing about projectiles' animations that gets synced in MP, and if they bring anything to MP - it's desync. It's the main reason why I dislike the idea of using them for this, even after the move of all firing code to ModProjectile (GlobalItem.Shoot enjoyers will not be happy). We can definitely just add the hooks needed to implement such stuff with just ModItems. I'm also interested in the ability to mess with item sourceRectangles through a GlobalItem, without player layer injections.

andrewlord607:
> There is nothing about projectiles' animations that gets synced in MP, and if they bring anything to MP - it's desync. It's the main reason why I dislike the idea of using them for this, even after the move of all firing code to ModProjectile (GlobalItem.Shoot enjoyers will not be happy). We can definitely just add the hooks needed to implement such stuff with just ModItems. I'm also interested in the ability to mess with item sourceRectangles through a GlobalItem, without player layer injections.
> 
> The 1.3 EM may be a bunch of adjusted vanilla code, but I'd really like 1.4 EM to be an actual example of a well programmed mod. I am still hoping for a single-entity worm npc one day.

Okay, I get you point. Hooks for alternating item's texture would be awesome(I think you should create **issue** for this). 
I close this PR until it is implemented

Issue #1403, opened 2021-03-30
[1.4] ExampleMod stuff to port from the Old directory


Mirsario:
### Description
This is a list of ExampleMod classes/files to port from the 'Old' directory of 1.4. Nearly anyone can help us out with this, as long as you know tiny bits of C# and are aware of what good code looks like.

**Code porting**
- During the porting process, the code should be heavily cleaned up, and potentially redesigned to match the new higher standards of 1.4 EM.
- Most of the time, content will need to be renamed to match what it provides an example for.
- Put a lot of comments explaining everything, and, if needed, raise the quality of comments already present there.
- Please improve readability, and don't be afraid to use line breaks to separate parts of the code. Like assignments, calls, and statements.

**Art**
There is zero need to improve quality of the art during the port process, but if you're an artist and dying inside while looking at something - you can just make it use the new "Examplium" palette already present on most ported items.

### To port & improve:
> - [ ] ExampleConfig
> - [ ] ExampleCustomCurrency
> - [ ] ExampleSurfaceBgStyle
> - [ ] ExampleUgBgStyle
> - [ ] Car -> ExampleCarMount
> - [ ] CarMount -> ExampleCarMountBuff
> - [ ] CarKey -> ExampleCarMountItem
> - [ ] ExampleJavelin & Projectile
> - [ ] PurityWisp -> ExampleHoveringMinion
> - [ ] Undead -> ExampleAdvancedDebuff
> - [ ] AddTimeCommand -> ExampleTimeCommand
> - [ ] CoinCommand -> ExampleCoinsCommand
> - [ ] ItemCommand
> - [ ] NPCCommand -> ExampleSummonCommand
> - [ ] AdvancedDust -> ExampleAdvancedDust
> - [ ] Bubble -> ExampleBubbleDust
> - [ ] EtherealFlame -> ExampleFlameDebuff
> - [ ] ExampleDrawTooltips -> ExampleAdvancedTooltipDrawingItem
> - [ ] ExampleFishingRod
> - [ ] ExampleBag
> - [ ] Face -> ExampleTooltipsItem
> - [ ] Fertilizer -> ExampleFeritilizerItem (Fix in multiplayer!)
> - [ ] ExclusiveAccessory -> ExampleExclusiveAccessory
> - [ ] ExampleBreastplate
> - [ ] ExampleCostume
> - [ ] ExampleHelmet
> - [ ] ExampleHood
> - [ ] ExampleLeggings
[...]

PR #1412, opened 2021-03-31
ExampleHoveringMinion

J164:
#1403 

Includes:

Renames PurityMinion to ExampleHoveringMinion
Adds additional comments to ExampleHoveringMinion


Mirsario:
Well, these are objective improvements, but unfortunately it seems that there was some mishap. The issue you referenced is 1.4 specific. This PR targets `master`, which is 1.3. Make a new branch out of '1.4', redo these improvements with its ExampleMod there (you'll notice that it actually has that 'Old' directory mentioned in the issue), then recreate the PR with that branch, targeting `1.4`.

PR #1417, opened 2021-04-02
[1.4] Update tModLoader.json

Igmc:
Added new string from the English lang file (l. 348)


direwolf420:
> [14:40:49] [1/INFO] [Terraria]: Steam Cloud Quota: 0.0 bytes available
You need to free up space on the cloud. Just to note: Terraria and tModLoader share the same cloud space, so clearing out Terrarias cloud saved files should also open it up for tModLoader.

PR #1424, opened 2021-04-04
A minion that always stays above your head

dpoopoop:
## What are the changes to ExampleMod?
it adds a rod that summons a minion that always stays above your head. 

it also teaches you about accessing variables from outside the ModPlayer.

the variable is made so you can summon only one at a time, and when you switch to another overhead minion in the same mod, it cannot be used, until the player manually clears the minion buff that activates the variable.


ExampleMod/Projectiles/Minions/ExampleOverheadMinion.cs
```diff
@@ -0,0 +1,84 @@
...
+                        shootToY *= distance * 3;
+                        int damage = projectile.damage;                  
+                        
+                        Projectile.NewProjectile(projectile.Center.X, projectile.Center.Y, shootToX, shootToY, mod.ProjectileType("ExampleOverheadMinionProjectile"), damage, 4f, Main.myPlayer, 0f, 0f); 
+                        Main.PlaySound(2, (int)projectile.position.X, (int)projectile.position.Y, 28); //28 is the sound
+                        projectile.ai[0] = 0f;
+                    }
+                }
+            }
+            projectile.ai[0] += 1f;
```
Chik3r, line 80: This code seems oddly similar to lines 57 to 85 from here: https://pastebin.com/1SNQ0sZd and the only difference seems to be that you removed the comments, and changed the distance check and the time between shots, but everything else is the same.
That pastebin is code from an `Al0n37` tutorial (https://www.youtube.com/watch?v=pqb0FKsvKY4)

Issue #1008, opened 2020-07-13
[Discussion] Shoot hook rework proposal/ideas


direwolf420:
### Description
This issue is related to `ModPlayer.Shoot` and `ModItem/GlobalItem.Shoot` and their behavior. Ill outline some of the issues of its current implementation and some proposed fixes.

### Application Scope
Currently, both hooks are called in
```cs
else if (PlayerHooks.Shoot(this, item, ref vector, ref num70, ref num71, ref shoot, ref Damage, ref KnockBack)
	&& ItemLoader.Shoot(item, this, ref vector, ref num70, ref num71, ref shoot, ref Damage, ref KnockBack)) {
```
which is at the very end of the chain relevant for all projectile spawning from item use. A big number of valid shooting items (I would say about 50%, other modders agree with this number) won't ever enter this part of the code, making the hook useless if you want to detect projectiles being shot from items in a guaranteed manner. (Vanilla code is still inconsistent as to where it spawns projectiles sometimes, so it won't be able to cover 100% of items anyway, but even 80% is better than 50%)

Possible fix would be to add another hook with no return value that simply executes when an item was used to shoot a projectile (Possible overlap with UseItem (which is also a bit inconsistent because of item.useStyle logic, but that's another topic)).

### General Use
The hooks return value (bool) indicates if the game should proceed spawning the projectile with the given parameters or not.
Modders use this hook to change the way projectiles are spawned, or just spawn new ones entirely. In combination with how the hooks are called (ModPlayer ones being split from the Item ones), this can cause shortcircuiting due to the `&&`.

Possible fix would be to split up the logic into `CanShoot` and `Shoot` hooks. As for the shortcircuiting, change it to `&` or combine it into one `Loader` method.

Mirsario:
> Possible fix would be to split up the logic into `CanShoot` and `Shoot` hooks.

Seems like the best solution to me.

Ozzatron:
I very highly support splitting into `Shoot` and `CanShoot` and disabling the short circuiting. Calamity has needed to perform a few ugly workarounds for some items ignoring Shoot hooks (Minishark comes to mind) or tolerate it only working 90% of the time.

Similarly, numerous headaches relating to Calamity's Rogue class can be traced back to the one highlighted line of code with the hooks. This is especially problematic when unofficial Calamity expansions are installed because they tend to have their own projectile behavior code that isn't reliably called.

It seems clear to me that both changes will provide significant concrete improvements.

[...]

PR #1432, opened 2021-04-07
[1.4] Shoot Rework

DarkLight66:
### What is the new feature?
Splits the old `Shoot` hook into `CanShoot`, `ModifyShootStats` and `Shoot`. This aims to solve the problems discussed in #1008.

* `CanShoot` would be used to cancel an item's shooting behavior easily.
* `ModifyShootStats` would be used to modify the stats of the projectile being shot, without dealing with manual projectile spawning directly.
* `Shoot` would be used to add additional projectiles to the shooting behavior and, in the case of the `ModItem` version of the hook, allow overriding the original projectile's spawning behavior.

I couldn't figure out a better way of handling how the `Shoot` hook should work than only allowing the `ModItem` version to change the original projectile spawning behavior, as i'm not sure which version of the hook should take priority in overriding that, and how to handle multiple versions of the hooks doing so. I'm partly relying on a future PR tackling projectile spawning thanks to the vanilla addition of projectile sources.

### ExampleMod updates
The files `ShortswordGlobalItem`, `ExampleGun`, `ExampleMinigun`, `ExampleModifiedProjectilesItem`, `ExampleShootingSword`, `ExampleShotgun`, `ExampleFishingRod` and `ExampleSimpleMinion` have been updated to the new hooks, although none of them use `CanShoot`.


patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -480,20 +480,39 @@ public abstract class ModPlayer : ModType
...
+		/// Allows you to add additional projectiles to an item's shooting mechanism.
+		/// </summary>
+		/// <param name="item"> The item being used. </param>
+		/// <param name="source"> The projectile source's information. </param>
+		/// <param name="position"> The center position of the projectile. </param>
+		/// <param name="velocity"> The velocity of the projectile. </param>
+		/// <param name="type"> The ID of the projectile. </param>
+		/// <param name="damage"> The damage of the projectile. </param>
+		/// <param name="knockback"> The knockback of the projectile. </param>
+		public virtual void Shoot(Item item, IProjectileSource source, Vector2 position, Vector2 velocity, int type, int damage, float knockback) {
```
Chicken-Bones, line 513: The type of `source` should be `ProjectileSource_Item_WithAmmo`

Chicken-Bones (changes requested):
Looks great. Only one suggestion

ExampleMod/Content/Projectiles/Minions/ExampleSimpleMinion.cs
```diff
@@ -70,13 +70,12 @@ public class ExampleSimpleMinionItem : ModItem
 			Item.shoot = ModContent.ProjectileType<ExampleSimpleMinion>(); // This item creates the minion projectile.
 		}
 
-		public override bool Shoot(Player player, ref Vector2 position, ref float speedX, ref float speedY, ref int type, ref int damage, ref float knockBack) {
+		public override void ModifyShootStats(Player player, ref Vector2 position, ref Vector2 velocity, ref int type, ref int damage, ref float knockBack) {
 			// This is needed so the buff that keeps your minion alive and allows you to despawn it properly applies
 			player.AddBuff(Item.buffType, 2);
```
direwolf420, line 76: Shouldn't `player.AddBuff(Item.buffType, 2);` be in the separate Shoot hook? Since it has nothing to do with modifying.

DarkLight66: Arguably you could say that it shouldn't `Shoot` either, since it has nothing to do with shooting. Vanilla doesn't even add minion buffs in the shooting code, and the closest hook to the place were vanilla applies its minion buffs would be `CanUseItem`, which works because the code that applies the buff doesn't actually care if the item can be used or not. Not sure if that would be proper hook usage though.

PR #1434, opened 2021-04-07
[1.4] ExampleHoveringMinion

J164:
#1403

Changes made:

Renamed, relocated, and added comments to:

- Purity Wisp -> ExampleHoveringMinion
- Purium Flame (dust) -> ExampleHoveringMinionDust
- Hover Shooter -> ExampleHoverShooter
- Minion -> ExampleMinion
- Purity Bolt -> ExampleMinionProjectile

Notes:

- I wasn't completely sure about what ExampleHoverShooter did so I did not add any comments to it. Some help on that would be appreciated.
- If there are better names for any of these files I will change them (I made up most of them)


ExampleMod/Content/Projectiles/Minions/ExampleHoveringMinion.cs
```diff
@@ -0,0 +1,88 @@
+using ExampleMod.Dusts;
+using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.ID;
+using static Terraria.ModLoader.ModContent;
```
Mirsario, line 5: It has been decided that 'using static' definitions shouldn't be in ExampleMod, to avoid confusion.

Mirsario (changes requested):
The branch does not compile (ignore that one checkmark, it's a CI bug), and it uses lowerCamelCase property names, even though they were changed to UpperCamelCase in January, and this PR is from April. Have you not set up a TML workspace to develop EM correctly?

PR #1437, opened 2021-04-09
[1.4] Big item use rework.

Mirsario:
## UseItemRework
This PR reworks UseItem to return a nullable boolean, and moves its main (there are multiple) call sites to the ApplyItemTime methods, fixing it not being called with a lot of vanilla items. This also lets GlobalItems properly control whether or not ItemTime should be applied.
```cs
/// Return true if the item actually did something, to force itemTime.
/// Return false to keep itemTime at 0.
/// Return null for vanilla behavior.
```

## UseAnimation hook
This PR adds a UseAnimation hook, which runs at the start of item animations.

## Multiplayer synchronization rework?
UseItem hasn't been working correctly *at ALL*, both in 1.3 and 1.4, due to remote players having very poor item use synchronization. In 1.3, UseItem would be called every frame on remote players, due to a lot of the shoot code never being called for them, and, therefore, itemTime not being ever reset from zero.
I've removed a bit of vanilla item use synchronization and made remote players run more code, based on their inputs, which are always synchronized. This also fixes a bunch of vanilla bugs, like sounds not being played correctly. Quite a radical change, but we're at the last major update of the game, and I think this is worth it, for modders' sake.

## Use Multiplier Hooks
image
The UseTimeMultiplier and MeleeSpeedMultiplier hooks were horribly confusing:
- UseTimeMultiplier wasn't even a multiplier, but a divisor.
- MeleeSpeedMultiplier could easily cause issues, like making the clockwork assault rifle shoot more or less than 3 projectiles it would shoot in vanilla.
- There wasn't a hook for just animation time.
- Why did MeleeSpeedMultiplier have 'Melee' in its name?

I've turned UseTimeMultiplier into an actual multiplier, and replaced MeleeSpeedMultiplier with two new hooks:
- UseAnimationMultiplier - Multiplies itemAnimation, similarly to UseTimeMultiplier, without a care for possible use count implications.
- UseSpeedMultiplier - Multiplies the item's overall usage speed, affecting even reuseDelay (unlike the other hooks.) BUT, it now additionally prevents issues with wrong use counts. This hook is completely safe to use for status effects, like an overcharge that doubles overall weapon speed. And even if you return 9999 here, the clockwork assault rifle will still shoot 3 shots.

## Player itemTime & itemAnimation desychronization fix
Vanilla has had a really horrible bug in it since the very beginning of the game. With it, every time the player would use an item that has autoReuse on it, the itemAnimation value would start lagging behind itemTime, the offset increasing by 1 every time the item is fired/used. In the video below, you can quite easily notice that the firing sound starts to heavily desync after a while.

https://user-images.githubusercontent.com/6795251/124508971-a3c26580-ddd9-11eb-9e81-32498e1a04d3.mp4

This issue comes from very random execution orders that lie inside the Player.ItemCheck method:
```diff
- 1. Reuse delay is applied
- 2. Item animation is applied if button is pressed
- 3. Item animation is reduced
- 4. Hold / Use styles are invoked
- 5. Item time is reduced
- 6. Item logic applies item time
```
It is resolved by changing the execution orders to the following:
```diff
+ 1. Item animation is reduced
+ 2. Item time is reduced
+ 3. Reuse delay is applied
+ 4. Item animation is applied if button is pressed
+ 5. Item logic applies item time
+ 6. Hold / Use styles are invoked
```
(Thanks @ThomasThePencil for help with this, been quite confusing)

## Other fixes
- Fixed weapons not being able to fire when their itemAnimation is at 1.
- The 2 in `Max(2, ...)` calls around total time methods is now 1.
- Total time methods are now in CombinedHooks class.


patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -296,11 +296,17 @@ public abstract class GlobalItem : GlobalType<Item>
...
 		/// </summary>
-		public virtual bool UseItem(Item item, Player player) {
-			return false;
-		}
+		public virtual bool? UseItem(Item item, Player player) => null;
+
+		/// <summary>
+		/// Allows you to make things happen when an item's use animation starts.
+		/// </summary>
+		public virtual void UseAnimation(Item item, Player player) { }
```
direwolf420, line 309: missing parameter documentation

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -728,23 +728,40 @@ GlobalItem Instantiate(GlobalItem g)
...
+
+			bool? modItemResult = item.ModItem?.UseItem(player);
+
+			return result ?? modItemResult;
+		}
+
+		private static HookList HookUseAnimation = AddHook<Action<Item, Player>>(g => g.UseAnimation);
+		
+		public static void UseAnimation(Item item, Player player) {
+			foreach (var g in HookUseAnimation.arr) {
```
direwolf420, line 760: artefact from mergedtesting

```suggestion
			foreach (var g in HookUseAnimation.Enumerate(item.globalItems)) {
```

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
Mirsario: There's kind of no point in documentation like 'item: The item.'.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3559,7 +3576,7 @@
  						ApplyItemTime(item);
  					}
 -					else if (itemTime == item.useTime / 2) {
-+					else if (itemTime == PlayerHooks.TotalUseTime(item.useTime, this, item) / 2) {
++					else if (itemTime == CombinedHooks.TotalUseTime(item.useTime, this, item) / 2) {
```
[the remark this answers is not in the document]
Mirsario: True, but I need to check if that field is updated properly first.

DarioDaF:
Tested `Player.ItemCheck_Inner` "spaghetti" reordering, this action will result in hitboxes being updated after item use because `Hold / Use styles are invoked` gets last in the chain resulting in bug: https://github.com/tModLoader/tModLoader/issues/1721 .
Furthermore removing the reordering (commenting the gotos) will result in collaterally fixing also https://github.com/tModLoader/tModLoader/issues/1770 and https://github.com/tModLoader/tModLoader/issues/1767 related to how feral claw updates stuff.
Empirically in `HandleItemHolding` the `ItemCheck_ApplyHoldStyle` is **never called** for melee weapons (not checked wit other weapons, but seems reasonable to think it's never called ever) because due to reordering `itemAnimation > 0` is always true maybe?

PR #1439, opened 2021-04-09
Implemented Modded Torch Smart Select Along With Other QOL Improvements  

ZakBerkelaar:
### What is the new feature?
This allows modded torches to be selected by smart select while also reducing the general torch boilerplate code that is required for modded torches.
This should resolve the rest of #1383 for 1.4.
Typically for modders to implement torches they have to override `HoldItem` and `PostUpdate` where they then have to do the standard check to ensure that the player isn't underwater before spawning particles and creating the light. The code almost always looks the same and is already handled by the game. This PR allows modders to just add the torch item to `ItemID.Sets.Torches` and possibly `ItemID.Sets.WaterTorches` and then override the new method `TorchVFX` where they can then return the color of light to create and the type of particles to spawn. This automatically handles torch smart select and removes the boilerplate code previously present.
A new `TileID` set could be created just for smart select or it could be done the way it was in 1.3 where you need to override `AutoLightSelect` and then set the bools `dryTorch`, `wetTorch` and `glowstick` respectively.
```c#
	ItemID.Sets.Torches[Type] = true; //Add the item to the torch set
	...
}

//Override the new method
public override void TorchVFX(out float R, out float G, out float B, out int DustType) {
	//Set the color of the light
	R = 1f;
	G = 1f;
	B = 1f;
	DustType = ModContent.DustType<Sparkle>(); //Set the type of particles the torch should emit
}
```
### ExampleMod updates

Updated to use the new torch system.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModItem.cs
@@ -871,12 +871,17 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Allows you to tell the game whether this item is a torch that cannot be placed in liquid, a torch that can be placed in liquid, or a glowstick. This information is used for when the player is holding down the auto-select hotkey.
+		/// Allows you to determine the color of light and dust that this torch emits when held or dropped on the ground
 		/// </summary>
-		/// <param name="dryTorch">if set to <c>true</c> [dry torch].</param>
-		/// <param name="wetTorch">if set to <c>true</c> [wet torch].</param>
-		/// <param name="glowstick">if set to <c>true</c> [glowstick].</param>
-		public virtual void AutoLightSelect(ref bool dryTorch, ref bool wetTorch, ref bool glowstick) {
+		/// <param name="R">The red component of the torch light (0-1)</param>
+		/// <param name="G">The green component of the torch light (0-1)</param>
+		/// <param name="B">The blue component of the torch light (0-1)</param>
+		/// <param name="dustId">The id of the dust that the torch emits</param>
+		public virtual void TorchVFX(out float R, out float G, out float B, out int dustId) {
+			R = 0;
+			G = 0;
+			B = 0;
+			dustId = DustID.Torch;
 		}
 
 		/// <summary>
--- a/patches/tModLoader/Terraria/ModLoader/ModItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModItem.cs
@@ -873,14 +873,14 @@ namespace Terraria.ModLoader
 		/// <summary>
 		/// Allows you to determine the color of light and dust that this torch emits when held or dropped on the ground
 		/// </summary>
-		/// <param name="R">The red component of the torch light (0-1)</param>
-		/// <param name="G">The green component of the torch light (0-1)</param>
-		/// <param name="B">The blue component of the torch light (0-1)</param>
+		/// <param name="r">The red component of the torch light (0-1)</param>
+		/// <param name="g">The green component of the torch light (0-1)</param>
+		/// <param name="b">The blue component of the torch light (0-1)</param>
 		/// <param name="dustId">The id of the dust that the torch emits</param>
-		public virtual void TorchVFX(out float R, out float G, out float B, out int dustId) {
-			R = 0;
-			G = 0;
-			B = 0;
+		public virtual void TorchVFX(out float r, out float g, out float b, out int dustId) {
+			r = 0;
+			g = 0;
+			b = 0;
 			dustId = DustID.Torch;
 		}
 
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/ModLoader/ModItem.cs
```diff
@@ -871,12 +871,17 @@ public abstract class ModItem : ModTexturedType
...
-		/// <param name="wetTorch">if set to <c>true</c> [wet torch].</param>
-		/// <param name="glowstick">if set to <c>true</c> [glowstick].</param>
-		public virtual void AutoLightSelect(ref bool dryTorch, ref bool wetTorch, ref bool glowstick) {
+		/// Allows you to determine the color of light and dust that this torch emits when held or dropped on the ground
+		/// </summary>
+		/// <param name="R">The red component of the torch light (0-1)</param>
+		/// <param name="G">The green component of the torch light (0-1)</param>
+		/// <param name="B">The blue component of the torch light (0-1)</param>
+		/// <param name="dustId">The id of the dust that the torch emits</param>
+		public virtual void TorchVFX(out float R, out float G, out float B, out int dustId) {
```
direwolf420, line 880: Parameters should be lowercase (r, g, b). Also, concider possible use-cases of a GlobalItem hook (and resulting ItemLoader method). This would change calls of the form `ModItem.TorchVFX` to `ItemLoader.TorchVFX`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3765,6 +3765,23 @@
...
  					if (sItem.type == 726)
  						Main.projectile[num178].magic = true;
+@@ -34203,8 +_,14 @@
+ 				else if (num2 >= 427)
+ 					num = num2 - 426;
+ 
++				int num3 = 0;
++				if (num >= 22) {
++					sItem.ModItem.TorchVFX(out R, out G, out B, out num3);
++				}
```
direwolf420, line 3775: What does num stand for here? And concider a safety null check via null coalesc
```suggestion
+				if (num >= 22) {
+					sItem.ModItem?.TorchVFX(out R, out G, out B, out num3);
+				}
```

patches/tModLoader/Terraria/ModLoader/ModItem.cs
MrSlimeDude: I don't think that this needs a `GlobalItem` hook as it doesn't really add anything that couldn't be done before without `HoldItem` and `PostUpdate` overrides in `ModItem` and those already have `GlobalItem` hooks.

patches/tModLoader/Terraria/Player.cs.patch
MrSlimeDude: Yea now that I am looking into it I don't really know why I did that so I have changed it to a safer null check on `ModItem` which should do the job just as well without the potential issues of how I was doing it.

patches/tModLoader/Terraria/ModLoader/ModItem.cs
```diff
@@ -871,12 +871,17 @@ public abstract class ModItem : ModTexturedType
...
-		/// <param name="wetTorch">if set to <c>true</c> [wet torch].</param>
-		/// <param name="glowstick">if set to <c>true</c> [glowstick].</param>
-		public virtual void AutoLightSelect(ref bool dryTorch, ref bool wetTorch, ref bool glowstick) {
+		/// Allows you to determine the color of light and dust that this torch emits when held or dropped on the ground
+		/// </summary>
+		/// <param name="r">The red component of the torch light (0-1)</param>
+		/// <param name="g">The green component of the torch light (0-1)</param>
+		/// <param name="b">The blue component of the torch light (0-1)</param>
+		/// <param name="dustId">The id of the dust that the torch emits</param>
+		public virtual void TorchVFX(out float r, out float g, out float b, out int dustId) {
```
Mirsario, line 880: ModifyTorchEffects would be a better name.
Also, make the 3 RGB parameters into a Vector3 lightColor

Mirsario (changes requested):
Good boilerplate reduction. I would prefer there to be a global version of the hook to control vanilla torch effects, but then there would have to be something similar for tiles, and the possibility of item and tile effects desyncing is not great. With TorchId and dumb luck rules being present, I'm thinking that perhaps there should be a way to define torch types and have items just link to torch types with IDs. But it could be weird to, say, make a GlobalTorchType, so I'm not fully sure about that. Needs discussing later on.

MrSlimeDude:
I agree that having a hook for vanilla torches would be nice and I would be happy to add it into this PR if desired (although I'm not sure where to put it as it would affect both items and tiles). The torch tiles and items should not be able to desync as they both call `TorchID.TorchColor` passing their torch ID to the method. Despite the inconveniences, I feel that it might be best if the potential vanilla torch hook was not combined with the modded torch hook as vanilla determines the torch id using several different methods including `tile.frameY / 22`, and `Item.placeStyle` which would make it difficult to determine if it was a modded torch or not. Making a GlobalTorchType should probably be discussed later as I agree that it could be weird just to have a type for torches specifically but torches are also a weird case considering that biome luck is a thing. A GlobalTorchType would also probably be best for another PR after ModBiome is merged.

PR #1441, opened 2021-04-10
AwesomePrefix -> ExamplePrefix

Valjean42:
This is my first contribution to tModLoader so I'd rather keep it small
Related to #1403 
Added `ExamplePrefix` to ExampleMod based on the older `AwesomePrefix`.
I don't think so.


ExampleMod/Content/Prefixes/ExamplePrefix.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            return false;
+        }
+
+        // Use this function to modify these stats for items which have this prefix: Damage Multiplier, Knockback Multiplier, Use Time Multiplier,
+        //			Scale Multiplier (Size), Shoot Speed Multiplier, Mana Multiplier (Mana cost), Crit Bonus.
+        public override void SetStats(ref float damageMult, ref float knockbackMult, ref float useTimeMult,
+            ref float scaleMult, ref float shootSpeedMult, ref float manaMult, ref int critBonus)
+        {
+            damageMult *= 1.05;
```
Chicken-Bones, line 54: This should include `_power` and probably be something like +20% for awesome, and +40% for really awesome.

ExampleMod/Content/Prefixes/ExamplePrefix.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class ExamplePrefix : ModPrefix
+    {
+        private readonly byte _power;
+
+        // See documentation for vanilla weights and more information.
+        // Note: a weight of 0f might still be rolled. See CanRoll to exclude prefixes.
+        // Note: if you use PrefixCategory.Custom, actually use ModItem.ChoosePrefix instead.
+        public override float RollChance(Item item)
+            => 5f;
```
Chicken-Bones, line 15: maybe use a ternary and have a lower roll chance for the higher power variant?
@JavidPack thoughts? Is this overcomplicating what should be a simple example?

ExampleMod/Content/Prefixes/ExamplePrefix.cs
```diff
@@ -0,0 +1,67 @@
...
+        public override bool Autoload(ref string name)
+        {
+            if (base.Autoload(ref name))
+            {
+                mod.AddPrefix("Awesome", new ExamplePrefix(1));
+                mod.AddPrefix("ReallyAwesome", new ExamplePrefix(2));
+            }
+
+            return false;
+        }
```
Chicken-Bones, line 47: This looks like 1.3 code. Have you tested this in 1.4?

Chicken-Bones (changes requested):
Thanks for working on this! Please double check that this builds and works properly on 1.4

ExampleMod/Content/Prefixes/ExamplePrefix.cs
Valjean42: Gotcha, that's done

ExampleMod/Content/Prefixes/ExamplePrefix.cs
direwolf420: adding prefixes like this requires ModSystem.Load or Mod.Load now and some additional code for the prefix constructor (Name property), here is an example:
Where it's called: https://github.com/direwolf420/ClickerClass/blob/1.4_experiment/ClickerClass.cs#L26
Prefix code and registering (see AddContent): https://github.com/direwolf420/ClickerClass/blob/1.4_experiment/Prefixes/ClickerPrefix.cs#L8

Valjean42: I saw, working on it at the moment (Apparently testing would require me to update Visual Studio, so this takes some time)

Valjean42:
As the commit says, it's pretty hard to make the class add 2 different prefixes, so I settled for one prefix in the class, in the name of keeping things simple enough for new modders

JavidPack:
Just ping one of us and mention this pr for access to that chat.

Prefixes are a great example for multi loading the class. I think we want to keep this as the example that loads multiple types from a single class. People in the chat can help you figure that out. I know a lot of things have changed so it would be most efficient to double check with people familiar with the new systems to do it right

ExampleMod/Content/Prefixes/ExamplePrefix.cs
JavidPack: We can keep it simple, but a comment showing a switch would probably be good, since a real mod would probably make 4 copies

Mirsario:
> 
> 
> Prefixes are a great example for multi loading the class. I think we want to keep this as the example that loads multiple types from a single class. People in the chat can help you figure that out. I know a lot of things have changed so it would be most efficient to double check with people familiar with the new systems to do it right

I'm trying very hard to not have offtopic examples in 1.4 EM. If there's a need for an example of multi-loading, it should be done in a separate file that states in its name that it provides an example of that. i.e. `ManualContentLoadingExample.cs`.

Here, it makes a ton more sense to just use two classes with inheritance and autoloading. It is not like there's going to be generics or for loops involved in common creation of prefixes.

Mirsario (approved):
Redid the example to use inheritance, removed the manual loading.

PR #1457, opened 2021-04-21
[1.4] Projectile drawing hooks rework

DarkLight66:
### What is the new feature?
* Make it so `ProjectileLoader.PreDrawExtras` is once again called by the vanilla projectile drawing code.
* Removed the `spriteBatch` parameter from all projectile drawing hooks, as vanilla now uses `Main.EntitySpriteDraw` to draw projectile sprites. Also changed the documentation to mention this.
* Changed vanilla patches in `Main.DrawProj` in such a way so that:

  - `ProjectileLoader.PreDrawExtras` is called before all drawing, and now can also prevent vanilla yoyo string drawing code from running.
  - `ProjectileLoader.PreDraw` now runs before all other vanilla drawing code and can prevent it from running, including the new code added in 1.4, which was previously placed before it and couldn't be stopped.
  - `ProjectileLoader.PostDraw` now properly runs after all vanilla drawing code.
* Changed `lightColor` parameter in `PreDraw` so it's passed by reference.
* Changed parameter names of `DrawBehind` projectile hooks to not be as verbose, and added `overPlayers` parameter that represents the cache of projectiles to draw in front of players added in 1.4 .
* Fixed `ModProjectile.GlowTexture` not working with projectiles that have multiple frames.

* Currently the `PreDrawExtras` hooks, which is normally used to override the drawing of things that usually go behind normal projectile drawing, like strings or chains, don't get called.
* To maintain consistency with vanilla, the `Main.EntitySpriteDraw` method should be used for all drawing of projectiles, as it takes care of drawing the sprites with the proper shader from dyes.
* Previously (as right now `PreDrawExtras` isn't even called by the game) yoyo strings couldn't be stopped from drawing.
* Currently there's a lot of projectiles that cannot be stopped from drawing by `PreDraw`, all of which were added in 1.4.
* Currently `PostDraw` doesn't get called in most cases where vanilla drawing code different from the default takes place.
* Passing `lightColor` to `PreDraw` by reference essentially allows modifying whether the projectile is affected by lighting, as well as the color of the light.
* Being able to draw projectiles over players easily is a feature sought after by modders.
* `ModProjectile.GlowTexture` not working with projectiles that have multiple frames severely reduces its usability.

Don't think so.

### ExampleMod updates
`ExampleBullet`, `ExampleBobber`, `ExampleAnimatedProjectile` and `ExampleHook` have been updated to use the new versions of the drawing hooks, and so they use `Main.EntitySpriteDraw`. Also fixed `ExampleHook` not drawing the chain correctly.


Chicken-Bones (approved):
Looks great! Clever patches.

PR #1463, opened 2021-04-21
[1.4] Reimplement TileCollideStyle projectile hooks

DarkLight66:
### What is the new feature?
* Made it so `ProjectileLoader.TileCollideStyle` is once again called by the vanilla `Projectile.HandleMovement` method, in the same section as in 1.3 .
* Added `hitboxOffsetMult` parameter, which allows adjusting the tile collision hitbox position.

* Currently, apart from the hook not being called at all, `TileCollideStyle` only allows modifying the size of the tile collision hitbox, but not to offset its position. Some vanilla projectiles do offset the hitbox position, so passing the variable used for this to the hooks allow mods to do the same.

The `hitboxOffsetMult` variable can be confusing to use. I tried to provide an explanation of its usage, but i believe it could better.

```cs
public override bool TileCollideStyle(ref int width, ref int height, ref bool fallThrough, ref Vector2? hitboxOffsetMult) {
	width = 12;
	height = 12;
	hitboxOffsetMult = new Vector2(1, 0.5f);
	return true;
}
```
The above code would resize the tile collision hitbox of the projectile to be 12 x 12. If `hitboxOffsetMult` is null, the new hitbox would have its center aligned with the projectile's center, which would be equivalent to `hitboxOffsetMult` being `new Vector2(0.5f, 0.5f)`, but the code above would make the new hitbox have its **bottom center** align with the projectile's center. Do note that the way the vanilla code is setup, either `width` or `height` have to be changed from -1 for `hitboxOffsetMult` to have any effect.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalProjectile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalProjectile.cs
@@ -94,14 +94,15 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
-		/// Allows you to determine how a projectile interacts with tiles. Width and height determine the projectile's hitbox for tile collision, and default to -1. Leave them as -1 to use the projectile's real size. Fallthrough determines whether the projectile can fall through platforms, etc., and defaults to true.
+		/// Allows you to determine how a projectile interacts with tiles. Return false if you completely override a projectile's tile collision behavior. Returns true by default.
 		/// </summary>
-		/// <param name="projectile"></param>
-		/// <param name="width"></param>
-		/// <param name="height"></param>
-		/// <param name="fallThrough"></param>
+		/// <param name="projectile"> The projectile. </param>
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
I think you should remove the requirement for `overrideWidth` or `overrideHeight` to be set, for `hitboxOffsetMult` to work. Looking at the vanilla line:
`vector2 = (vector.HasValue ? (base.Center - new Vector2(num, num2) * vector.Value) : new Vector2(base.position.X + (float)(base.width / 2) - (float)(num / 2), base.position.Y + (float)(base.height / 2) - (float)(num2 / 2)));`

Running that all the time shouldn't change the default behaviour.

I also think `hitboxOffsetMult` should be called `hitboxCenterFrac`

Remove the nullability, and maybe even move
```cs

	int num = (overrideWidth != -1) ? overrideWidth : base.width;
	int num2 = (overrideHeight != -1) ? overrideHeight : base.height;
```
before the hook, so you can pass `ref num` and `ref num2` in.

PR #1468, opened 2021-04-22
[1.4] SetStaticDefaults unification through ModType

direwolf420:
### Brief
Singleton classes have a `SetDefaults` method instead of `SetStaticDefaults` which is confusing and inconsistent. Most classes do not have this method at all, which is required to complement other `ModType`s and reduce code splitting to achieve the same thing you could only do before by using `ModSystem/Mod.PostSetupContent`.

### Description
The use of `SetStaticDefaults` and `SetDefaults` was only clearly split for non-singleton classes like `ModItem` or `ModProjectile`: in `SetDefaults` you would assign stats (`Item.damage = 10;`, `Projectile.timeLeft = 60;`), in `SetStaticDefaults` define various static properties (`Item.staff[Type] = true;`, `ProjectileID.Sets.Homing[Type] = true;` etc).
Singleton classes like `ModTile` or `ModMount` have only `SetDefaults`, which is used for both of these purposes at once (assign static arrays aswell as properties of itself). Many smaller, non-content related classes also do not have any equivalent hooks at all.

#### This brings up several problems:
**Problem Nr.1:** Singleton-`SetDefaults` are misnamed. This causes confusion for modders and misconception regarding what the hooks actually do for the non-singleton version. As an example, I have seen many modders misuse `ModItem.SetDefaults` and mix in code that should be in `SetStaticDefaults` instead. Another example would be thinking that if a `SetDefaults` hook exists, that means the type is a non-singleton, for example `ModBuff`, and then treat it like a non-singleton and add fields to the class (which don't work like they intended to - they act like static variables and cause issues in multiplayer).
**Solution:** For how they are called by tModLoader and what their purpose is, they should be renamed to `SetStaticDefaults`. `ModType`s affected by this:

```
GlobalTile
GlobalWall
InfoDisplay
ModTile (part of ModBlockType)
ModWall (part of ModBlockType)
ModBuff
ModDust
ModMount
ModPrefix
```

**Problem Nr.2:** 27 out of 36 `ModType`s are missing a `SetStaticDefaults` (and for tml, a properly sealed `SetupContent` method for a half of them). This means, for anything that should run after content has finished initially loading, you have to move to `ModSystem/Mod.PostSetupContent`, spreading your code for no reason.
**Solution:** Add `SetStaticDefaults` to `ModType` directly, and make all remaining `ModType`s call it through a sealed `SetupContent` override. `ModType`s affected by this are all remaining 27 not listed in **Solution Nr.1**.

### Additional items
* Added documentation to all `ModType` members so that their purpose is more clear (modders can use them too!)
* Changed `ModPrefix.AutoDefaults` to `AutoStaticDefaults` to fit with the pattern of `ModItem/Projectile/NPC.AutoStaticDefaults`
* Fixed SetupContent not being previously sealed for the following:

```
DamageClass
GlobalInfoDisplay
```

* Fixed Register not being previously sealed for the following:

```
InfoDisplay
PlayerDrawLayer
```

### Drawbacks of current implementation
* `ModSystem` gets special treatment by having its loading related hooks get called separately. This introduces overlap between `SetStaticDefaults` and `PostSetupContent`. The main difference is that the `ModSystem`-specific hook runs after ALL the mod's content has gone through its loading and setup phases, whereas each `ModType` is isolated in itself, any types loaded after it have not yet finished setting up. This PR simply seals the methods that should be sealed, so it will not cause confusion.
* ~~Loss of `ModType`-specific documentation, such as what common properties most modders want to change (`Main.tileSolid[]`, `Main.debuff[]`, `DisplayName.SetDefault` etc.).~~
Actually added back the summaries via logic-less overrides -- Mirsario.
`ModType`s affected by this:

```
ModTile (renamed) (part of ModBlockType)
ModWall (renamed) (part of ModBlockType)
ModBossBar
ModBuff (renamed)
ModDust (renamed)
ModItem
ModNPC
ModProjectile
ModMount (renamed)
```

### Alternatives
* Since only 9 `ModType`s have their method renamed, and another 27 do not have one, it should be concidered to rename it to `PostSetupContent`, as it fits the "load stage" better.
* During `Mod.SetupContent`, call `SetStaticDefaults` after `SetupContent`, remove those calls from the individual `SetupContent`s but keep them sealed. The problem with this is that some `ModType`s have code running before and after calling `SetStaticDefaults`, and this ordering would be broken.

### ExampleMod updates
ExampleMod and ModLoaderMod have been updated to use the new method names where required.


Chicken-Bones:
I think `SetupContent` should have a default implementation which calls `SetStaticDefaults`

I don't think there's much benefit to sealing it. We can keep the ones which have actual implementation sealed to make it clearer they're not meant to be overridden, but it feels like boilerplate. 

`SetStaticDefaults` should be `protected` (since it's only called by `SetupContent`)

Mirsario (approved):
Discussed CB's suggestions in discord, merging as is.

PR #1473, opened 2021-04-23
Brazilian Portuguese localization addition

sahlaysta:
Test for now

template=localization


JavidPack:
This is good, no syntax errors as far as I can see, but we might not be updating the 1.3/master branch much further. For future contributions it might be more suitable to target the 1.4 branch, as that is where all the current work is being done in preparation for updating tmodloader to 1.4

Issue #1466, opened 2021-04-22
[1.4] NET5 & 1.4.2.2 Post Merge - Pre Alphas - TODO & Issues


Solxanich:
- [x]  Update InstallVerifier
  - [x] Seek official confirmation on if can or can't include steam_api.dll in release zip and/or Github from Steam Rep 
    - [x] Possibly switch from using steam_appid.txt to a more automated way to determine if its Steam or GoG
    - [x]  ~~Possibly encrypt hashes so that only TML Team and View and Edit them.~~
- [x]  Review and confirm bat and shell scripts are up to spec for needs:
  - [x]  Icon is set correctly
    - [x] Preferred use is runtime, not scripts: https://wiki.libsdl.org/SDL_SetWindowIcon + GameWindow.Handle
  - [x] Re-verify Unix shell scripts
  - [ ] Possibly replace start-tML.bat with apphost 64bit exe + steamInstall.vdf
    - [x] Portable install script fails on Windows7 PowerShell
    - [x] .tMod extension no longer associated to tModLoader.exe, because exe doesn't exist.
      - [x] Need to associate Opens With to tModLoader.dll on Windows if not using apphost
- [ ]  UPnP in UwUPnP logs minimal troubling errors in server log
https://discord.com/channels/103110554649894912/445276626352209920/834736569931661342
- [ ] Performance issues: #1470 
- [x] Achievements are being granted again when weren't pre-merge, need to look in to
- [x] it still says wants framework45 targetting pack, I guess that got missed? 
- [x] Direwolf420's weird graphical issues: https://discord.com/channels/103110554649894912/445276626352209920/835184240122658866 / https://github.com/tModLoader/tModLoader/issues/1487
- [x] Solxan's Underground Background issue when running ~OpenGL: https://discord.com/channels/103110554649894912/445276626352209920/836285660197290035
- [x]  update https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-contributors
- [x] Update the migration guide link in mod sources UI to look nicer
- [x] update the tML 1.4 tag in discord
- [x] look into why music doesn't loop in vanilla build (IE Terraria.sln)

kittenchilly:
- [ ] Item Definition Elements are broken and crash the game when loaded (possibly other definition elements too?) (#1467)

cozykoko:
Why does it state AMD performance issues, when the user in #1470 doesnt even have AMD hardware in his pc?

Solxanich:
[...]

PR #1481, opened 2021-04-26
Misc Net5 Improvements Series 1

Solxanich:
Per #1466 
1) Replace resource hacker with runtime setting of Icon via cross-plat SDL (have to use SDL for GameWindow.Handle, not WinAPI, no wonder I couldn't get it working earlier)
2) Add argument "-console" to prevent console from being hidden
3) Change MigrationLink colour from hard to read Black to easy to read Red


patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -198,7 +198,7 @@ internal class UIModSources : UIState
 
 			var font = FontAssets.MouseText.Value;
 			Vector2 sizes = font.MeasureString(versionUpgradeMessage);
-			Color color = Color.Black;
+			Color color = Color.Red;
```
Mirsario, line 201: 🧐
```suggestion
			Color color = Color.IndianRed;
```

PR #1482, opened 2021-04-28
[1.4] Add IgnoredByGrowingSaplings

Antirhinnum:
### What is the new feature?
* A new TileID Set, named `IgnoredByGrowingSaplings`, that is used to allow tiles to be treated as empty when growing saplings - specifically, in the `WorldGen.EmptyTileCheck()` method.

* Currently, growing saplings into trees calls `WorldGen.EmptyTileCheck()` to ensure the tree can actually grow. In this method, in addition to inactive tiles, a hardcoded list of other tiles (like grass, thorns, etc.) are counted as "empty" when checking saplings. If any tile not in this hardcoded list is detected, then the sapling will not grow.
* An example of this being an issue is a modder adding in a new type of foliage, like a Hallowed version of the Corruption's thorns. In the current system, any sapling near these Hallowed thorns would be unable to grow, despite being able to grow near Corrupt thorns.

* Adding a new method to `ModTile` and/or `GlobalTile` that allows the user to determine whether a tile is counted as empty for any parameters of `WorldGen.EmptyTileCheck()`.

```cs
public override SetDefaults() {
    TileID.Sets.IgnoredByGrowingSaplings[Type] = true;
    ...
}
```

### DIFF
--- a/patches/tModLoader/Terraria/ID/TileID.TML.cs
+++ b/patches/tModLoader/Terraria/ID/TileID.TML.cs
@@ -36,6 +36,9 @@
 
 			/// <summary> Whether or not this tile counts as a lava source for crafting purposes. </summary>
 			public static bool[] CountsAsLavaSource = Factory.CreateBoolSet();
+
+			/// <summary> Whether or not saplings count this tile as empty when trying to grow. </summary>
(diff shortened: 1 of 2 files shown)


JavidPack:
Looks good to me. I looked at the method and it seems that the name of the new set matches what is happening in the method.

JavidPack (approved):

PR #1500, opened 2021-05-03
Cleanup concerns on ModBiome post-merge

Solxanich:
### What is the bug?
Syntax issues in ModBiome PR, that weren't caught prior to merge. These include: "the Loader<T> properties are lower case, the zoneForest property in player-tML.cs is lower case, ModAVFX should be ModAvfx, and the Loaders static class has undesirable public set capability, when should be private set".

Others may exist that have not been captured/flagged, so if curious, please consult against the commit here: https://github.com/tModLoader/tModLoader/commit/e893b3ea7c91b0772db88a32c580517bb989aa96

### How did you fix the bug?
Fixed per Style Guide and Discord direction.

### Additional Details regarding ModSceneEffect
ModSceneEffect now does the handling of choosing scene effects instead of separate hooks, so that tML can give proper attention to designated priorities. Notably, it has an IsSceneEffectActive return method, and Priority property associated to it. It should be derived directly when adding scene effects that were controlled by any of the following hooks, with the exemption of ModTypes that derive this class already such as ModBiome. Multiple small, derived classes may be required to accomplish the same functionality. This change does not affect existing ModNPC music implementations at time of writing. 

The following changes have been made with respect to ChooseStyle Hooks, with possibly multiple ModSceneEffect classes being required to fully replace the Hook:
1) ModSystem.ChooseWaterStyle Hook -> ModSceneEffect.WaterStyle Property
2) ModSystem.ChooseMusic Hook -> ModSceneEffect.Music Property
3) ModSurfaceBgStyle.ChooseStyle Hook -> ModSceneEffect.SurfaceBackgroundStyle Property
4) ModUgBgStyle.ChooseStyle Hook -> ModSceneEffect.UndergroundBackgroundStyle Property


Chicken-Bones:
Suggestion `Avfx[Priority]` -> `AmbienceEffect` or `AmbienceFX`
I assume "avfx" currently stands for "audio-visual-effects"

Chicken-Bones:
I'll probably further replace the `LoaderManager` backend with static backing fields and something more singleton based. Though I wonder how we could expand this to mods having their own loaders. 

Either way... I want this PR to focus on the public API/naming, and get that stabilised ASAP

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -102,9 +102,9 @@ public partial class Player {
 
 
 		/// <summary>
-		/// Container for current AVFX client properties such as: Backgrounds, music, and water styling
+		/// Container for current SceneEffect client properties such as: Backgrounds, music, and water styling
 		/// </summary>
-		public AVFXLoader.AVFXInstance currentAVFX = new AVFXLoader.AVFXInstance();
+		public SceneEffectLoader.SceneEffectInstance currentSceneEffect = new SceneEffectLoader.SceneEffectInstance();
```
Mirsario, line 107: API fields should be properties.

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -119,9 +119,9 @@ public partial class Player {
 		public bool InModBiome(ModBiome baseInstance)=> modBiomeFlags[baseInstance.ZeroIndexType];
 
 		/// <summary>
-		/// The zone field storing if the player is in the purity/forest biome. Updated in <see cref="UpdateBiomes"/>
+		/// The zone property storing if the player is in the purity/forest biome. Updated in <see cref="UpdateBiomes"/>
 		/// </summary>
-		public bool zonePurity { get; set; } = false;
+		public bool ZonePurity { get; set; } = false;
```
Mirsario, line 124: Why not ZoneForest? Also, unnecessary assignment.
I think that the biome PR should've eliminated all the Zone properties, not add new ones. There should  be wrapping over vanilla, same way as with DamageClass.

Mirsario (approved):
We'll solve these later.

PR #1506, opened 2021-05-07
Generalized all ModContent.XType and GetX methods

Mirsario:
## Description

Heavily cleans up and unhardcodes `ModContent` by generalizing all `int XType<T>()` methods into `int GetId<T>()`.

For now, all methods like `ModItem GetModItem(int type)` are removed with no local alternative. Loaders' methods still exist.
Later, these should be replaces with `int` overloads for Get and TryGet methods.


Chicken-Bones:
Going to close this, feel free to re-open if you ever want to work on it again. Note that the opportunity to done it without maintaining backwards compatibility with the current methods has passed. Overall I think the benefit provided by this would be limited. In my experience, `struct ID` types have a bit too much friction to be a suitable replacement for `int type` in Terraria.

PR #1518, opened 2021-05-11
[1.4] Shoot Rework Reloaded

DarkLight66:
### What is the new feature?
Allowing mods to override shooting behavior of vanilla items by returning false in `GlobalItem` and `ModPlayer` `Shoot` hooks (again!).

It was a feature in 1.3. In my shoot rework PR (#1432), i changed it so only `ModItem.Shoot` could override default shooting behavior, to stop the scenario were multiple mods that override shooting behavior spawned their own projectiles, each of which were supposed to override the default one. Now i believe this should be something the mods in question should deal with, not tModLoader.

Maybe there is a design out there that would solve the issue with multiple mods overriding shooting i mentioned above.

### ExampleMod updates
Updated `ShortswordGlobalItem.Shoot`, and made it return false to showcase the effect of overriding shooting, as the blades of the shotswords are projectiles in 1.4.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -598,12 +598,12 @@ namespace Terraria.ModLoader
 		/// <summary>
 		/// Calls each GlobalItem.Shoot hook then, if none of them returns false, calls the ModItem.Shoot hook and returns its value.
 		/// </summary>
-		public static bool Shoot(Item item, Player player, ProjectileSource_Item_WithAmmo source, Vector2 position, Vector2 velocity, int type, int damage, float knockback, bool vanillaShoot) {
+		public static bool Shoot(Item item, Player player, ProjectileSource_Item_WithAmmo source, Vector2 position, Vector2 velocity, int type, int damage, float knockback, bool defaultResult = true) {
 			foreach (var g in HookShoot.Enumerate(item.globalItems)) {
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -594,16 +594,18 @@ GlobalItem Instantiate(GlobalItem g)
...
+			bool vanillaShoot = item.ModItem?.Shoot(player, source, position, velocity, type, damage, knockback) ?? true;
+
 			foreach (var g in HookShoot.Enumerate(item.globalItems)) {
-				g.Shoot(item, player, source, position, velocity, type, damage, knockback);
+				vanillaShoot &= g.Shoot(item, player, source, position, velocity, type, damage, knockback);
 			}
 
-			return item.ModItem?.Shoot(player, source, position, velocity, type, damage, knockback) ?? true;
+			return vanillaShoot;
 		}
```
Mirsario, line 609: Returning false in GlobalItem should cancel the ModItem call.

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
Mirsario: Additionally, the same should somehow apply to modplayer. But returning false in ModPlayer shouldn't stop GI shoot from happening. Makes for some spaghetti.

Mirsario (approved):

PR #1520, opened 2021-05-11
[1.4] Allow Guns to shoot with no ammo

DarkLight66:
### What is the new feature?
Adds the `NeedsAmmo` hook to `ModItem` and `GlobalItem`, which allows any ammo consuming item to shoot when there's no valid ammo for the item left in the player's inventory, in which case the item will shoot as if the default ammo, the item with the type used as the ammo class's ID, is being used.
When shooting with no ammo, the code related to ammo consumption doesn't run, including the `ConsumeAmmo` and `OnConsumeAmmo` hooks.

Some modders may want to allow shooting without ammo when using certain equipment, while under certain buffs, when using certain weapons, etc.

Maybe also allowing to change which ammo type the item will use while having no ammo could be allowed, though it would require validation to make sure the ammo being selected also belongs to the ammo class in the item's `useAmmo` field, as the vanilla code doesn't expect certain items to use ammo outside their class.

### ExampleMod updates
Check the updated `ExampleMinigun` for an example of this hook's usage.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -4931,8 +4931,9 @@
 +
 +			bool shootWithNoAmmo = false;
 +			if (!canShoot && !ItemLoader.NeedsAmmo(sItem, this)) {
-+				canShoot = shootWithNoAmmo = true;
-+				item.SetDefaults(sItem.useAmmo);
++				item = ContentSamples.ItemsByType[sItem.useAmmo];
++				if (item.ammo == sItem.useAmmo)
++					canShoot = shootWithNoAmmo = true;
 +			}
 +
  			if (!canShoot)
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -2207,6 +2207,16 @@
  							SmartSelect_SelectItem(i);
  							return;
  						}
+@@ -12560,6 +_,9 @@
+ 								}
+ 							}
+ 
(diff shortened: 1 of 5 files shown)


fakelag:
I think changing the ammo type could also be cool as you pointed out. What kind of problems would there be with validation?

Maybe there could be a design where it would be a ref argument to the hook, so it could be optionally set by a modder

DarkLight66:
Actually, i had to put the `NeedsAmmo` hook in places were the code only checks if the player has ammo and nothing else, so setting which ammo to shoot should probably go in a separate hook in `Player.PickAmmo`, but the `PickAmmo` hook in there already allows changing the projectile set by the ammo, as well its damage, speed and knockback, and modders can know if the player is shooting with no ammo by checking if `NeedsAmmo` was called, so in the end i don't think there's much of a need for a hook to change the ammo used there.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4032,6 +4063,29 @@
...
+ 
+ 			return canUse;
+@@ -36484,6 +_,12 @@
+ 				}
+ 			}
+ 
++			bool shootWithNoAmmo = false;
++			if (!canShoot && !ItemLoader.NeedsAmmo(sItem, this)) {
++				canShoot = shootWithNoAmmo = true;
++				item.SetDefaults(sItem.useAmmo);
```
Mirsario, line 4083: SetDefaults calls can be super slow
```suggestion
+				item = ContentSamples.ItemByType[sItem.useAmmo];
```

Mirsario (changes requested):

Mirsario (approved):

Issue #1521, opened 2021-05-12
Ammo damage discrepancy between vanilla and tModLoader


DarkLight66:
### Description
The way tModLoader applies damage modifiers to ammo differs from vanilla, as it calculates the damage modifier to apply to the ammo by roughly estimating the damage bonus the weapon received and then applying it to the ammo. This has the problem that:
 * The result of applying the damage modifier to the weapon was rounded down when the damage code had to cast the resulting value to an integer. Say a ranged weapon has a damage of 5 and the player has a ranged damage bonus of 50%, the weapon will end up dealing 7 damage instead of 7.5 due to casting, which would translate to 40% damage bonus instead, and that is the bonus that will be applied to the ammo it uses.
 * If the weapon happens to have a different damage class from the ammo, or deals no damage to begin with but the ammo does, the ammo will end up not being affected by modifiers that should apply to it. In vanilla, some weapons that would be affected by this are the flare gun, which has no damage type, but the flares it uses as ammo deal ranged damage; and the coin gun, which deals no damage by itself, but uses ammo that deals considerable amounts.
 * In vanilla, certain damage modifiers that apply to ranged weapons don't actually apply to ranged ammo of the corresponding type, but they do in tModLoader, namely the `arrowDamage`, `bulletDamage` and `rocketDamage` ones.
 * Archery potions, which increase both the damage of weapons that use arrows as ammo, and the damage of arrow ammo items themselves, apply their damage modifier in `PickAmmo` in vanilla. In tModLoader, this is applied as a multiplier to `arrowDamage`, which has the issue i mentioned above.
   * [Edit] Additionally, the move to `arrowDamage` doesn't account for the molten quiver's 2 damage bonus to normal arrows.
   * [Edit2] Also, the multiplier is applied to the sum of weapon and ammo damage in vanilla, while in tModLoader it's only applied to each one individually. In a lot of cases, this will result in 1 less total damage for the projectile. This suggests the archery potion bonus is a very special case, and modders trying to imitate such a bonus should apply it somewhere more appropriate, like in the `PickAmmo` hook.
 * As an aside, there's no way for modders to apply custom damage modifiers to ammo specifically.
 
[...]

PR #1523, opened 2021-05-14
Fix Item Damage issues

DarkLight66:
### What is the bug?
All info can be found in issue #1521 

### How did you fix the bug?
* Changed the method signature of `GetWeaponDamage` to include a `forTooltip` parameter (defaulting to `false`) that tells the method whether it is being called from the drawing code of the item's tooltip, and moved the relevant damage calculations in the tooltip drawing code to `GetWeaponDamage`.
* Made it so ammo damage modifiers are applied through `GetWeaponDamage`, and removed the ranged ammo check so modded ammo can be modified too, as vanilla only adds ranged and unclassed ammo.
* Restored the vanilla behavior of the archery potion buff, as such an effect should probably be applied in the `PickAmmo` hooks instead.
* Changed how `arrowDamage`, `bulletDamage` and `rocketDamage` `StatModifier`'s are combined with `rangedDamage`, by casting them to float first then multiplying them, so it's more in line with how vanilla does it.
  * However, i didn't re-add the casting to int between steps of the vanilla `GetWeaponDamage` code, as the damage calculations of the vanilla tooltip drawing code multiplies the bonuses first before applying them to the damage and casting to int, so i'm keeping the more sensible of the two.
* Adding `5E-6f` to the damage calculations before casting to int, including the summon damage calculation in `GetWeaponDamage`, as that's the only one it doesn't apply to, even though the tooltip version of the code does add it. Incidentally vanilla actually doesn't really make use of the damage returned by `GetWeaponDamage` for minions and turrets, as it recalculates the damage each tick in the projectile's `Update` method.

### Are there alternatives to your fix? ###
I don't think so.

### Additional Notes ###
Actual in-game testing of the changes can be difficult, as vanilla damage randomness makes it hard to be sure whether the values match or not.


Chicken-Bones:
Implemented via #2196.
Sorry yours got lost in the review pile

Issue #1328, opened 2021-02-11
[1.4 alpha] Mod Browser is extremely slow, recalculates children too much


Solxanich:
### Description
When opening the Mod Browser in the dev/alpha version, the UI will lag and stutter significantly.

### Log file
![image](https://user-images.githubusercontent.com/59670736/107691305-88a75b80-6c68-11eb-82fc-2ad9a477b45c.png)
^UI List

### Steps to reproduce
n/a

### Expected behavior 
Framerate should be smooth like 1.3 Mod Browser

### Actual behavior
Framerate is smooth like a mountain range.

### Reproduction frequency
100% reoccurence

### Additional Information
We should probably disable the Mod Browser button in 1.4's main menu for now, and look into fixing it when it makes sense be it in SW environment or otherwise.

cozykoko:
Isnt this closable now?

Solxanich:
This has been confirmed closable, we think

PR #1529, opened 2021-05-18
AlphaPrep: Disable ModBrowser, cleanup InstallVerifier check, add betaWelcomeMessage

Solxanich:
### What is the change?
In preparation for moving from Indev to Alpha, this PR addresses a handful of small ToDos.

1) Disables Mod Browser button - Mod Browser as is, is 1.3. Steam Workshop PRs will restore this overtime. As well, its terribly slow due to #1328 
2) Cleanups InstallVerifier check to not require deleting steamapp_id.txt. This file is needed for GoG's steam workshop; also was unnecessary extra install step when better programming exists.
3) Adds a betaWelcomeMessage to clearly denote this is the beta version. **Only appears every run for release builds**

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Engine/InstallVerifier.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Engine/InstallVerifier.cs
@@ -93,20 +93,18 @@ namespace Terraria.ModLoader.Engine
 			}
 
 			// If its clearly a steam install/launch, use Steam API.
-			if (Directory.GetCurrentDirectory().Contains("steamapps") || Program.LaunchParameters.ContainsKey("-steam")) {
+			if (Directory.GetCurrentDirectory().Contains("steamapps") || Program.LaunchParameters.ContainsKey("-steam"))
 				return CheckSteam();
-			}
 
 			Logging.tML.Info("Checking if GoG or Steam...");
 
-			// If can't find a GoG folder, than it is a Steam launch
+			// If can't find a GoG folder, than assume it must be a Steam launch
 			if (!ObtainVanillaExePath(out var steamApiPath, out var exePath))
 				return CheckSteam();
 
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/ModLoader/Engine/InstallVerifier.cs
```diff
@@ -86,11 +86,12 @@ internal static class InstallVerifier
 				return false;
 			}
 
-			// Whether the steam_appid.txt file exists, indicating we'd have to check steam installation
-			if (File.Exists(SteamAppIDPath))
-				return CheckSteam();
+			// Check if we are within a steamapps subfolder OR are trying to force run as steam install; if not, return GoG.
+			if (!(Directory.GetCurrentDirectory().Contains("steamapps") || Program.LaunchParameters.ContainsKey("-steam"))) {
+				return CheckGoG();
```
Mirsario, line 91: What if someone has a dev variant way outside steam? Instead this should check for whether or not steam_api.dll/unix variant (or maybe CSteamworks.dll) is present in the _vanilla_ game's folder (at `../` or `../Terraria/`). I've mentioned this in Discord.

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/Engine/InstallVerifier.cs
Solxanich: Had included the launch parameter for it, but had aimed to avoid rewriting large parts of InstallVerifier to facilitate. 
I've since enhanced it to check each of the steam_api variants & attempt to rule out GoG if Terraria folder is not nearby; although my certainty on unix steam_api location is limited, in particular for MacOS. 
~~Although feels more convoluted for this since can't just call the api to give steam Terraria install location. ~~

Mirsario (approved):

Issue #1489, opened 2021-04-30
[1.4] Rename character button flickers


cozykoko:
### **Steps to reproduce**

Just hover over the character rename button and you can see the tooltip flickering.

### **Expected behavior** 

 Tooltip should not flicker.

Mirsario:
There is a lot of flickering bugs in vanilla that happen when you turn off frameskip with a more than 60hz monitor. Are we sure that this is TML's issue?

cozykoko:
I tested this in vanilla and it was not the case.

Steviegt6:
> There is a lot of flickering bugs in vanilla that happen when you turn off frameskip with a more than 60hz monitor. Are we sure that this is TML's issue?

Absolutely. It's an issue caused by a mistake in a patch I believe, making the text position use an incorrect variable that messes with its offset. It's overlayed on top of the rename button, causing flickering if you hover over the text, as it cancels the hovering of the button, which cancels the text and allows hovering again, rinse & repeat.

cozykoko:
Fixed.

PR #1537, opened 2021-05-22
[1.4] Fix rename character label flickering

Fatiz:
### What is the bug?
Bug description is found in #1489. 
The bug is caused by the UIText object being positioned over the `Rename` button causing the OnMouseOver event to constantly fire causing the flickering text.

### How did you fix the bug?
Fixed this by assigning the UITexts left position to use the `num` variable just like in `UIWorldListItem.cs`. This causes the UIText button to be positioned off the of the button preventing the flicker.

Initially, I was going to use the position of the `Rename` button to determine the placement of the `_ButtonLabel` however after cross comparing code to `UIWorldListItem.cs` it seems using the already existent `num` variable that is used is most likely the best practice.

### Note
This is my first pull request and I'm unsure if the patch is correct. I applied the patch using the tool and all seemed to be fine but feel free to let me know 🙂


JavidPack:
This is good, thanks.

Issue #1532, opened 2021-05-19
[1.4] Mod Config Saving/Revert Changes buttons don't properly disappear


direwolf420:
### Description
On 1.4, doing changes to a ModConfig ingame causes the Saving/Revert Changes buttons to show up (normal behavior), but clicking them does not cause any feedback: the buttons don't disappear, making the user think that nothing happened, yet it does save the config properly into the json file.

### Log file
No logs

### Steps to reproduce
Get a mod with a ModConfig and atleast one field in it. Open the game, open the mods config, change a value, press save.

### Expected behavior 
Buttons disappear.

### Actual behavior
Buttons don't disappear.

### Reproduction frequency
100%.

### Additional Information
Closing and then reopening the ModConfig afterwards makes the buttons disappear. EXCEPT if when you click "Revert to default", then save, open and close, the buttons don't go away, until you manually change something again.

JavidPack:
Should be fixed with the recent commits

PR #1539, opened 2021-05-23
[1.4] Update Mod Config buttons to refresh UI when pressed

Fatiz:
### What is the bug?
Issue #1532 | Issue is caused by the buttons not getting removed from the `uIElement` collection when buttons are pressed.

### How did you fix the bug?
Made changes for buttons related functions to call `onActivate` - This fixes the issue listed as well as updates the UI with the default / original values when `RestoreDefaults` and `RevertConfig` are called respectively.

Two alternatives I can think of are:
1. Instead of redrawing the UI, you instead just manually remove the elements from the UI when the buttons are pressed. i.e. `uIElement.RemoveChild(saveConfigButton)`. This seems like a more roundabout fix as opposed to refreshing the UI itself.
2. Create a `UpdateUI` function that will essentially perform the same actions as `onActivate` - I was not opposed to this however, I felt as though it was unnecessary code duplication when it seems `onActivate` does what is needed perfectly. If it is felt a `UpdateUI` or similar function is necessary, it can be done instead! 🙂

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Config/UI/UIModConfig.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Config/UI/UIModConfig.cs
@@ -283,6 +283,7 @@ namespace Terraria.ModLoader.Config.UI
 			//else
 			//{
 			DoMenuModeState();
+			OnActivate();
 			//}
 		}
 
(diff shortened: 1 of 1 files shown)


JavidPack:
I took a different approach for a fix that I think better fits the intended path in bbb70c4580488064ee5def7cfcd11fbf7293a7bc

PR #1086, opened 2020-09-13
.NET 5 and Win-64-FNA Port

Dual-Iron:
### What is the new feature?
.NET Core 3.1 compatibility. Sixty-four-bit FNA windows may be PRed into 1.4 soon after this is functional. FNA windows and .NET Core should ideally be ported in any order, I'm just PRing .NET Core first.
### Why should this be part of tModLoader?
Using the modern .NET Core 3.1 should allow tModLoader to persist into the future and would give modders access to new and greater language features, .NET services, and other useful tools.
### Are there alternative designs?
Not really, this is the logical step from .NET Framework.
### Sample usage for the new feature
Not applicable [yet]
### ExampleMod updates
Will be needed, todo.

PR #1544, opened 2021-05-27
[1.4] Fix .NET Framework 4.0 compilation errors

void8192:
### What is the bug?
Terraria doesn't compile with the .NET Framework (`Terraria_LegacyNetFramework.csproj`) after #1086.

### How did you fix the bug?
- Added conditional directives to separate .NET Core and .NET Framework code.
- Added legacy-supported MP3Sharp and FNA libraries.

I don't think so.


Chicken-Bones:
Thanks for this. There's some cleanup to be done here, but we're not planning on supporting net40/net45 any more. The transition period this was intended for didn't last very long. We'll keep it around, but it'll only get used if we need to debug a difference between Terraria and tModLoader with XNA or FNA or something.

PR #1549, opened 2021-06-01
Hook for sign text changed.

JakobHartman:
### What is the new feature?
When a sign is finished being edited, I want to notify the tile that an update has been made.

I want to create a mod that allows a user to create trackers using signs, notifying the tile that the sign has been updated, I can cleanly update text on the mini map, to what the sign says.

I have spent many days and haven't found one.

```
public override void SignDidChange(int i, int j)
{
  string text = Main.sign[Sign.ReadSign(i, j, true)].text;

  ModTranslation name = CreateMapEntryName();
  name.SetDefault(text);
  AddMapEntry(new Color(19, 255, 0), name);

  entryCount++;
}
```

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2555,6 +2555,15 @@
  				if (npc[player[myPlayer].talkNPC].type == 20) {
  					SoundEngine.PlaySound(12);
  					npcChatText = Lang.GetDryadWorldStatusDialog();
+@@ -29565,6 +_,8 @@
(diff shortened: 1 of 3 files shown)


htd-tstepien:
Sorry if some offtopic, but I'm browsing PRs around here, and I'm seeing quite a lot \`ifs\` with arguments that really drag on a lot of space.

Such a question from a total amateur of JS/NodeJS rather than C - wouldn't the code be more readable if you put the \`ifs\` arguments into a described variable? It would be clear what this \`if\` is checking.

PR #1552, opened 2021-06-03
Added Exampleboomerang

crownarch:
Added exampleboomerang


JavidPack:
Looks like this fell through the cracks. We are fine with an Example Boomerang in ExampleMod, but this example needs to have the AI code written out rather than using aistyle 3 for it to be useful as a learning example. 

If anyone or you wants to make such an example for current tModLoader, that would be great, I'll be sure to merge it quickly.

PR #1559, opened 2021-06-06
RegisterHotKey 'Keys' parameter.

RighteousRyan1:
## The Change
Changed `defaultKey` to be a `Microsoft.Xna.Framework.Input.Keys` parameter instead of being a string parameter. Maybe a need for a deprecated method?

## The Why
This needs a change because some people don't know all of the serializable names for keys.


Mirsario:
Addressed in the above commit. I've just added an overload to RegisterHotkey, which you can find in HotkeyLoader.

PR #1562, opened 2021-06-08
[1.4] New "ActsLikeTownNPC" & "SpawnsWithCustomName" sets for NPCs

MutantWafflez:
### What is the new feature?

In 1.4, a new boolean property was created called `isLikeATownNPC` for NPCs that when true, gives an NPC the functionality of a town NPC in terms of AI, animation, and attacks, while not requiring other things such as housing or a minimap head. This PR adds two new hooks, `ActsLikeTownNPC` and `SpawnsWithCustomName` in `ModNPC` that takes advantage of this new boolean and allows for easier naming of non-town NPCs.

This should be a part of tModLoader due to the fact that before, to get the functionality of the Bone Merchant from vanilla, you had to do some sort of method detouring or IL editing to achieve the proper effect. With these new hooks, modders can simply make the hooks return true and the NPC will act very similarly to the Bone Merchant if everything else is implemented correctly.

Likely not. Not that I can immediately think of, at least.

See below:

### ExampleMod updates

Added a new NPC called Example Bone Merchant. Functions very similarly to `ExamplePerson`, and showcases the new hooks and the other things required to make an NPC that acts similarly to vanilla's Bone Merchant.


ExampleMod/Content/NPCs/ExampleBoneMerchant.cs
```diff
@@ -0,0 +1,155 @@
...
+			NPCID.Sets.AttackTime[Type] = 90; // The amount of time it takes for the NPC's attack animation to be over once it starts.
+			NPCID.Sets.AttackAverageChance[Type] = 30;
+			NPCID.Sets.HatOffsetY[Type] = 4; // For when a party is active, the party hat spawns at a Y offset.
+
+			// Influences how the NPC looks in the Bestiary
+			NPCID.Sets.NPCBestiaryDrawModifiers drawModifiers = new NPCID.Sets.NPCBestiaryDrawModifiers(0) {
+				Velocity = 1f, // Draws the NPC in the bestiary as if its walking +1 tiles in the x direction
+				Direction = 1 // -1 is left and 1 is right. NPCs are drawn facing the left by default but ExamplePerson will be drawn facing the right
+				// Rotation = MathHelper.ToRadians(180) // You can also change the rotation of an NPC. Rotation is measured in radians
+				// If you want to see an example of manually modifying these when the NPC is drawn, see PreDraw
```
direwolf420, line 36: The comment regarding PreDraw does not apply to this file (copied from ExamplePerson)

ExampleMod/Content/NPCs/ExampleBoneMerchant.cs
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		//This hook is the most important part of this NPC. Since it is true, it tells the game that we want this NPC to act like a town NPC without being one.
+		//What that means is: the NPC will have the AI of a town NPC, will attack like a town NPC, and have a shop (or any other additional functionality if you wish) like a town NPC.
+		//However, the NPC will not have their head displayed on the map, will de-spawn when no players are nearby or the world is closed, will spawn like any other NPC, and have no happiness button when chatting.
+		public override bool ActsLikeATownNPC() => true;
+
+		//Make sure to allow your NPC to chat, since being "like a town NPC" doesn't automatically allow for chatting.
+		public override bool CanChat() => true;
+
```
direwolf420, line 64: ExampleMod convention is always using braces even on single line methods

```suggestion
		//However, the NPC will not have their head displayed on the map, will de-spawn when no players are nearby or the world is closed, will spawn like any other NPC, and have no happiness button when chatting.
		public override bool ActsLikeATownNPC() {
			return true;
		}
		
		//Make sure to allow your NPC to chat, since being "like a town NPC" doesn't automatically allow for chatting.
		public override bool CanChat() {
			return true;
		}

```

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -715,6 +724,15 @@
  			if (Main.npc[newNPC].type == 1 && Main.player[num7].RollLuck(180) == 0)
  				Main.npc[newNPC].SetDefaults(-4);
  
+@@ -62697,7 +_,7 @@
+ 		}
+ 
+ 		private static void GiveTownUniqueDataToNPCsThatNeedIt(int Type, int nextNPC) {
+-			if (TypeToDefaultHeadIndex(Type) != -1 || Type == 453) {
++			if (TypeToDefaultHeadIndex(Type) != -1 || Type == 453 || NPCLoader.ActsLikeATownNPC(Type)) {
+ 				Main.npc[nextNPC].GivenName = getNewNPCName(Type);
```
direwolf420, line 733: Why does this add a check here? Terraria would use `isLikeATownNPC` instead of `Type == 453` otherwise

MutantWafflez: I added the check here cause `isLikeATownNPC` is non-static and the method here is static, with the `Type` parameter coming from several other methods including `NewNPC`, thus requiring a `NPCLoader` call. `Type == 453 || NPCLoader.ActsLikeATownNPC(Type)` is equivalent to `isLikeATownNPC` as a result.

direwolf420: `Main.npc[nextNPC].isLikeATownNPC` could work, but does it represent the same logic? Could require a different hook since vanilla would otherwise check that, but it didn't for some reason

MutantWafflez: Apologies due to my lack of understanding, but is there something wrong with specifically the `NPCLoader` call being used here? I might be missing something obvious, and again sorry for my ignorance in this case.

direwolf420:
You can extend the functionality of `SpawnsWithCustomName` to return `bool?` with 3 modes:
null: Default behavior
true: Force to roll a custom name
false: Prevent a custom name

ExampleMod/Content/NPCs/ExampleBoneMerchant.cs
```diff
@@ -0,0 +1,163 @@
...
+			NPC.DeathSound = SoundID.NPCDeath1;
+			NPC.knockBackResist = 0.5f;
+
+			AnimationType = NPCID.Guide;
+		}
+
+		//This hook is the most important part of this NPC. Since it is true, it tells the game that we want this NPC to act like a town NPC without being one.
+		//What that means is: the NPC will have the AI of a town NPC, will attack like a town NPC, and have a shop (or any other additional functionality if you wish) like a town NPC.
+		//However, the NPC will not have their head displayed on the map, will de-spawn when no players are nearby or the world is closed, will spawn like any other NPC, and have no happiness button when chatting.
+		public override bool ActsLikeATownNPC() {
```
Solxanich: Please convert both this and SpawnsWithCustomName to be properties instead of methods.
There doesn't look like any need for fully fleshed out virtual methods here.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -114,6 +114,20 @@ public abstract class ModNPC : ModTexturedType
...
+		/// </summary>
+		/// <returns></returns>
+		public virtual bool? SpawnsWithCustomName => null;
+
+		/// <summary>
+		/// Whether or not this NPC will act like a town NPC in terms of AI, animations, and attacks; but not in other regards, like the bone merchant in vanilla. Returns false by default.
+		/// </summary>
+		/// <returns> </returns>
+		public virtual bool ActsLikeATownNPC => false;
+
```
Mirsario: I don't see much reason for these 2 to not just be NPCID.Sets.* entries. Especially since you deemed GlobalNPC entries unnecessary.

Mirsario (changes requested):

Mirsario (approved):

PR #1567, opened 2021-06-08
Asset Loading Rework! Proper async loading. Graphics access safety.

Chicken-Bones:
Completely rewrote asset loading to use a common call path for sync and async loading. 
Texture deserialization (reading) and asset creation are split only if the current thread is not the main thread. 
Requesting the same asset asynchronously and then via ImmediateLoad will not cause a deadlock.

Added FNA as a submodule, pointing to a fork where an exception is thrown if graphics functions are ever called outside the main thread.

Removed ModAssetRepository
Simplified all content sources to use a common base class. Implementation is dead simple.
Removed the ability to have 2 assets with the same name but different extensions. .cs files don't count as assets because there's no reader.

Removed ImageSharp dependency. FNA3D Image calls work for us.

Note that `ModContent.GetAsset` defaults to `AssetRequestMode.AsyncLoad` ~~where `AssetRepository.Request` defaults to `ImmediateLoad`~~. Async load is perfect for mod assets, because we make sure all async loads are completed via `Mod.TransferAllAssets` at the end of loading. 

Note that async does not yet mean _on demand_ and that the only way for an asset to be reloaded currently is switching resource packs. (Reloading mods discards all the old assets and makes a whole new `AssetRepository`)

TODO:
- Moved to #1612


Solxanich:
Regarding ResourcePackContentValidator, Yorai:
"It was meant to validate whole resource packs in one go, and let you know of rejections immediately - as right now the resource pack system only rejects on-demand and you don't know whether you have an issue until you are actually experiencing it.
that said this feature never made it to fruition, and most of what you see in the compiled code is just a dummy state.
that is unused, too.
so you can feel free to remove/ignore it. "
Appx. Jun 21 8 am MST

PR #1587, opened 2021-06-23
Renamed ZonePurity player variables to ZoneForest for consistency wit…

Seraph-ED:
Renamed ZonePurity player variables to ZoneForest for consistency with vanilla's use of the term for the default biome


Solxanich:
A couple of quick non-binding comments/things to check.

If my memory serves, ZonePurity does not include consideration to the vertical zones - surface, underground and caverns in particular. 

As such, I would instead anticipate something like three bools for each of ZoneForest, ZoneNormalUnderground, and ZoneNormalCaverns that each derive ZonePurity and corresponding vertical layer if we are aiming for more accuracy?

Mirsario:
With solxan's comment, I'm leaning towards just keeping this 'ZonePurity'.

Issue #1560, opened 2021-06-07
[1.4] .NET Libraries are not found


BlackTeal:
### Description
Whenever I try to use a .NET Library (in my case ImGui.NET which is a .NET Standard 2.0 Library), my Mod just doesn't find it (it fails to Load the Library), the Library is referenced in build.txt with the dllReferences thing (dllReferences = ImGui.NET) and it is present in the lib Directory inside the Directory of my Mod

### Log file
[client.log](https://github.com/tModLoader/tModLoader/files/6608966/client.log)

### Steps to reproduce
1. Create your 1.4 Mod Project
2. Add any .NET Library (not sure if it happens with other Libraries than .NET Standard 2.0 Libraries, I tried with the ImGui.NET Library, which is a .NET Standard 2.0 Library) compatible with .NET 5 as a Project Reference, or download one from nuget
3. In the build.txt file, add the name of the DLL of the Library to the dllReferences
4. Add the .NET Library to the lib Directory inside the Directory of your Mod
5. Use it in any way so your Mod tries to Load the Library

### Expected behavior
The mod loads the library, without any exceptions or crashes

### Actual Behavior
The mod throws System.Reflection.ReflectionTypeLoadException exception when it tries to load the .NET Library

### Reproduction frequency
100% as far as I have tested, the Mod never is able to find the .NET Library DLL, not even if I put it in the tModLoader Directory or in the Mod DLL Directory

Chicken-Bones:
Please send a zip of your mod from the mod sources directory.

BlackTeal:
[BlackTeal.zip](https://github.com/tModLoader/tModLoader/files/6613564/BlackTeal.zip)

Solxanich:
@BlackTeal - 
A problem I note with the ImGui.net lib you're using in that mod is that its missing several other parts to it.
IE, all the cimgui.dll and its variations for each of the platforms.
Further, I see its supposed to be a NuGet package, but with what you have, it doesn't match the contents of the package either.

Although a fix may be done to eliminate the error resulting from not finding ImGui.Net, I won't anticipate that it will start working fully for you until said problem is addressed on your end.

BlackTeal:
Yes, I'm aware that it's missing the native libraries, but as the managed library wasn't being loaded at all, I reported that as a bug, I don't know why it doesn't match the package (because I have took it somewhere in the .nuget folder from my user directory) but thanks for telling me those stuff though.

Chicken-Bones:
[...]

PR #1609, opened 2021-07-08
Fix resolving assemblies from tmod files

Solxanich:
### What is the bug?
#1560 tmodloader was unable to resolve embedded assemblies from .tmod files.

### How did you fix the bug?
Added a custom resolver to resolve assembly probing.

Functionally, no.


Mirsario (approved):

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -83,6 +84,23 @@ private class ModLoadContext : AssemblyLoadContext
 				return asm;
 			}
 
+			public void LoadCustomResolver() => AppDomain.CurrentDomain.AssemblyResolve += ModCustomResolver;
+			public void UnloadCustomResolver() => AppDomain.CurrentDomain.AssemblyResolve -= ModCustomResolver;
+
+			// In Net5, in order for the embedded dll references to resolve, I implemented a custom resolver. 
+			// This fixes Issue #1560, and allows mods to include and load 3rd party dlls. Utilizes the existing... 
+			// 'assemblies' dictionairy to retrieve the byte array to load in to the domain.
```
Mirsario, line 92:
```suggestion
			// 'assemblies' dictionary to retrieve the byte array to load in to the domain. 
```

Mirsario (changes requested):
Actually, no, this looks weird. I don't think AppDomain.CurrentDomain has a need to be used in the context of ALCs. Is ModLoadContext.Load override not sufficient?
CB had some other concerns.

PR #1636, opened 2021-07-27
[1.4] Add new set check in NewNPC to spawn in reverse

direwolf420:
### What is the new feature?
Adds an `NPCID.Sets` set to expand a niche use case of forcing an NPC to spawn at the end of the `Main.npc` array instead of looking for an empty slot from the beginning. This is used for Queen Bee and Golem in vanilla to display their minions/body parts infront of the boss, due to how the drawing order is set up.

Modders had to previously either detour `NPC.NewNPC` to force this behavior manually, or draw the "secondary" NPCs within the "primary" NPC to customize the draw order.

The set name and its description could be different/improved.

Vanilla has something similar with `Item.NewItem`s `reverseLookup` parameter, but that's used for different purposes.

`public override void SetStaticDefaults() => NPCID.Sets.SpawnFromLastEmptySlot[NPC.type] = true;`

### ExampleMod updates
Technically, `MinionBossBody` could get this, but the minions it spawns are a good distance away from the main body, which means no overlap, so draw order is not an issue there.


JavidPack:
We already have the spawnnpc hook for this, right? There was never a need to detour anything to get this effect

direwolf420:
SpawnNPC does not fire when summoning through SpawnOnPlayer or the multiplayer-received end of it.

PR #1640, opened 2021-07-28
New exceptions on invalid drop rule

Ishigh1:
### What is the new feature?
To help people to debug the drop rates, with this PR, all the vanilla drop rules will throw an exception if the minimum amount of something is below the maximum.
This design prevents people from changing the DropOneByOne.parameters during the game, but I believe that if someone is techy enough to want it, they will have a better design by creating a custom class anyway
I also chose to prevent the bonus per chunk per player from reducing the spread to avoid some unexpected bugs with an high number of players


Chicken-Bones (approved):
Looks good to me!

PR #1642, opened 2021-07-29
Update WorldGen.cs.patch for Mac Recursion limitng

Solxanich:


Solxanich:
Unnecessary after stack minimum fix

PR #1644, opened 2021-07-29
Random Assortment of Patches from MacOS testing

Solxanich:
export COMPlus_DefaultStackSize=4000000 so that the default secondary thread MacOS stack limit of 512kb becomes 4MB and add logging for it being received (untested if it fixes Mac stack size issues)

update Workshop Publisher to error on missing icon/content folder prior to upload instead of during

add logging worldgen stage to log file for easier debugging

update SteamGameServer.Init to actually use its bool return and path better.

update the logging around launch parameters so that if none are used, it doesn't show up in the log 

update the Mac library look change to also include creating the symlink so as to avoid the SegFault issues seen on Linux re-occuring.

### DIFF
--- a/solutions/ReleaseExtras/RuntimeFiles/InstallNetFramework.sh
+++ b/solutions/ReleaseExtras/RuntimeFiles/InstallNetFramework.sh
@@ -22,6 +22,9 @@ else
   ln -sf "$library_dir/libSDL2-2.0.so.0" "$library_dir/libSDL2.so"
 fi
 
+# Ensure sufficient stack size on MacOS secondary threads, doesn't hurt for Linux
+export COMPlus_DefaultStackSize=4000000
+
 #Parse version from runtimeconfig, jq would be a better solution here, but its not installed by default on all distros.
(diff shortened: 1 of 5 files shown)


solutions/ReleaseExtras/RuntimeFiles/InstallNetFramework.sh
```diff
@@ -15,13 +15,18 @@ echo "This may take a few moments."
 # Additionally, something in dotnet is requesting 'libSDL2.so' (instead of 'libSDL2-2.0.so.0' that is specified in dependencies)
 # without actually invoking managed NativeLibrary resolving events!
 if [[ "$(uname)" == Darwin ]]; then
-  export DYLD_LIBRARY_PATH="$script_dir/Libraries/Native/OSX"
+  library_dir="$script_dir/Libraries/Native/OSX"
+  export DYLD_LIBRARY_PATH="$$library_dir"
```
Mirsario, line 19:
```suggestion
  export DYLD_LIBRARY_PATH="$library_dir"
```

PR #1656, opened 2021-08-01
Getters for GameModeData including journey powers

Ishigh1:
### What is the new feature?
It includes ways to get the different values affected by journey and for the worthy.

These getter are a bit long and can be quite common, so this allows modders to get the right value easily and others that may want to change it can affect these methods to affect everyone instead of doing for each compatible mod.

This PR is open to more upgrades, I can think of including custom flags, make the setters protected and the getters virtual to allow modder to make their own GameModes or add a List of modifiers which could include the for the worthy one and affect the result of the getter.

The getters would mainly be needed when changing the damage/health value of a projectile after the SetDefaults, either with a phase system or when changing the values of a manually spawned vanilla entity.


patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		public float GetEnemyDamageModifier() {
+			float enemyDamageModifier = GetPureEnemyDamageModifier();
+			if (Main.getGoodWorld) {
+				if (IsJourneyMode) {
+					enemyDamageModifier *= 2;
+				}
+				else {
+					enemyDamageModifier += 1;
```
Chicken-Bones, line 26: Vanilla code for this is just ...
I think this should be `enemyDamageModifier = enemyDamageModifier + 1) * (1 + 1/3f);`

because `ScaleStats_ApplyGameMode` has
```
			if (Main.getGoodWorld)
				damage += damage / 3;
```
at the end

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+	public partial class GameModeData
+	{
+		public float GetPureEnemyDamageModifier() {
+			float enemyDamageModifier = EnemyDamageMultiplier;
+
+			if(IsJourneyMode) {
+				CreativePowers.DifficultySliderPower difficultySliderPower =
+					CreativePowerManager.Instance.GetPower<CreativePowers.DifficultySliderPower>();
+				enemyDamageModifier *= difficultySliderPower.StrengthMultiplierToGiveNPCs;
```
Chicken-Bones, line 13: `if (power.GetIsUnlocked())` ?

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,72 @@
...
+			}
+			
+			return enemyDamageModifier;
+		}
+
+		public float GetEnemyDamageModifier() {
+			float enemyDamageModifier = GetPureEnemyDamageModifier();
+			if (Main.getGoodWorld) {
+				if (IsJourneyMode) {
+					enemyDamageModifier *= 2;
```
Chicken-Bones, line 23: Where does this come from in vanilla?

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,72 @@
...
+			if(IsJourneyMode) {
+				CreativePowers.DifficultySliderPower difficultySliderPower =
+					CreativePowerManager.Instance.GetPower<CreativePowers.DifficultySliderPower>();
+				enemyDamageModifier *= difficultySliderPower.StrengthMultiplierToGiveNPCs;
+			}
+			
+			return enemyDamageModifier;
+		}
+
+		public float GetEnemyDamageModifier() {
```
Chicken-Bones, line 19: I think these should probably be properties, and they should have the original name with `Effective` in front of them.

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,72 @@
...
+					enemyDamageModifier *= 2;
+				}
+				else {
+					enemyDamageModifier += 1;
+				}
+			}
+
+			return enemyDamageModifier;
+		}
+		public float GetPureEnemyMaxLifeMultiplier() {
```
Chicken-Bones, line 32: similarly, the `Pure` property getters should probably be called `EnemyMaxLifeMultiplerWithPowersApplied`

Chicken-Bones (changes requested):
Needs more comments indicating the vanilla code that this mirrors. 

Given that Vanilla has all kinds of things like `GetAttackDamage_ForProjectiles` as well as scaling based on number of players. This does seem helpful in reducing code duplication, but I'm not sure it captures the whole story, and without use cases it's hard to maintain. Some ExampleMod code would help.

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
Ishigh1: This method simply always return true

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
```diff
@@ -0,0 +1,117 @@
...
+		{
+			get
+			{
+				float enemyDamageModifier = EnemyDamageMultiplier;
+
+				if (IsJourneyMode)
+				{
+					CreativePowers.DifficultySliderPower difficultySliderPower =
+						CreativePowerManager.Instance.GetPower<CreativePowers.DifficultySliderPower>();
+					enemyDamageModifier *= difficultySliderPower.StrengthMultiplierToGiveNPCs;
```
Mirsario: The entire file violates the code style, including with that trailing assignment.

Mirsario (changes requested):

patches/tModLoader/Terraria/DataStructures/GameModeData.TML.cs
Ishigh1: How does it violates the code style ? I don't see it

Chicken-Bones:
I finally had the opportunity to cleanup this mess in vanilla for 1.4.5. It was a long road. The things you provided here were helpful, but the vanilla picture was such a mess that I wasn't able to review it at the time. My apologies for this taking so long

PR #1665, opened 2021-08-02
Save & load for NPCs

Ishigh1:
### What is the new feature?
Adds the ability to save custom data on NPCs (via `Mod/GlobalNPC`), even if the type of NPC normally wouldn't be saved because it's not a `townNPC` or `NPCID.Sets.SavesAndLoads[npc.type]` is false

An NPC which needs to save won't despawn naturally.

override `Mod/GlobalNPC.NeedSaving()` to make non-townNPCs save.

Take a look at the examples! 
Can be used for anything where an NPC might want to 'persist'. Eg, enemies that stole money or items, tamed animals, fancier merchants.

Probably none as simple as this. 

From `ExampleZombieThief`
```cs
public override bool NeedSaving() {
	return StolenItems >= 10; // Only save if the NPC has more than 10 stolen items, to avoid keeping the NPC in memory if it only has few
}

public override void SaveData(TagCompound tag) {
	if (StolenItems > 0) {
		// Note that at this point it may have less than 10 stolen items, if another mod or part of our decides to save the NPC
		tag["StolenItems"] = StolenItems;
	}
}

public override void LoadData(TagCompound tag) {
	StolenItems = tag.GetInt("StolenItems");
}
```

### ExampleMod updates
`ExamplePerson` Remembers how many times it has been talked to.
`ExampleTravelingMerchant` An ExampleTravellingMerchant. Uses Save/Load to remember the randomized shop for the day.
`ExampleZombieThief` A zombie which steals ExampleItems and stays around if it's stolen enough of them.
`ExampleGlobalNPC` Makes townNPCs double prices if they've been hit by a player (and saves this data until the NPC dies)


Mirsario:
@Ishigh1 Can now be updated according to #1814 's changes and new standards.

patches/tModLoader/Terraria/IO/WorldFile.cs.patch
```diff
@@ -188,7 +188,7 @@
  			for (int j = 0; j < Main.npc.Length; j++) {
  				NPC nPC2 = Main.npc[j];
 -				if (nPC2.active && NPCID.Sets.SavesAndLoads[nPC2.type]) {
-+				if (nPC2.active && NPCID.Sets.SavesAndLoads[nPC2.type] && !NPCLoader.IsModNPC(nPC2)) {
++				if (nPC2.active && NPCLoader.SavesAndLoads(nPC2) && !NPCLoader.IsModNPC(nPC2)) {
```
Chicken-Bones: I'm wary about changing vanilla saving logic. Is there a reason for a mod to enable saving for a vanilla NPC? Would this break vanilla compat?

Chicken-Bones: I suppose you need this to save globals on vanilla NPCs...
I need an example of such in ExampleMod before this can be merged

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1218,5 +1218,22 @@ internal static void DrawBehind(NPC npc, int index)
 				}
 			}
 		}
+
+		private delegate bool DelegateNeedSaving(NPC npc);
+		private static HookList HookNeedSaving = AddHook<DelegateNeedSaving>(g => g.NeedSaving);
+
+		public static bool SavesAndLoads(NPC npc) {
+			if (npc.townNPC && npc.type != NPCID.TravellingMerchant)
+				return false;
```
Chicken-Bones: You sure about this? Pretty sure modded townNPCs should be able to save and load

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1218,5 +1218,22 @@ internal static void DrawBehind(NPC npc, int index)
...
+			if (NPCID.Sets.SavesAndLoads[npc.type] || (npc.ModNPC?.NeedSaving() == true))
```
Chicken-Bones, line 1256: need to consider the usage of this set, what's the intended use for it? Should modders use it at all? Should it be fully deprecated and calls replaced with `NPCLoader.SavesAndLoads`?

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -125,62 +126,157 @@ internal static class WorldIO
...
+							["data"] = data
+						});
+
+						data = new TagCompound();
 					}
 				}
+
+				TagCompound tag;
+
+				if (NPCLoader.IsModNPC(npc)) {
```
Chicken-Bones, line 161: should probably check `modNPC.NeedsSaving` here, or you'll confuse people about why `SaveData` gets called even when they said `NeedsSaving` is false, just because there's a global or some other reason the NPC is being saved

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
Ishigh1: Sorry if the name is a bit misleading, SavesAndLoad is only for saving outside of town NPCs, they have their own saving not related to this one
So they are always saved and the call to the hook is unneeded

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
Ishigh1: I didn't want to change that but yes it seems fair to remove the access to this array. It's only useful because vanilla uses it honestly, so it's as you want

patches/tModLoader/Terraria/IO/WorldFile.cs.patch
Ishigh1: Well this change doesn't really change the behavior of tmodloader as you could already change the previous array.
It would make sense though to only save vanilla NPCs that were intended to be saved, just wanted to follow the logic in place.
If you want a change to it if the array becomes unavailable to modders, feel free to say it
Also for ExampleMod, what kind of example do you want ?

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
Ishigh1: The whole logic behind the NPC save logic is that NPCs can be saved even if the modNPC saving is set to false, and town NPC are a special case where you always save (mainly because it's the previous behavior and everyone wants their town NPC to be saved)

patches/tModLoader/Terraria/IO/WorldFile.cs.patch
Chicken-Bones: I think this just needs reverting. This controls what goes in the vanilla save file, so it should use the vanilla `SavesAndLoads` check, `nPC2.active && NPCID.Sets.SavesAndLoads[nPC2.type] && !NPCLoader.IsModNPC(nPC2)`

PR #1680, opened 2021-08-02
Re-implement Music Loading in 1.4

Steviegt6:
### What is the new feature?
This pull request adds back in music loading, allowing for modders to use their own custom music. This picks up on the original `1.4_musicfix` branch.

There could be, but this seemed like the most reasonable approach.

```cs
    public class TestMenu : ModMenu
    {
        public override int Music => MusicLoader.GetMusicSlot("Rejuvena/Sounds/Music/Sky_Tower");
    }
```
Example of setting the music of a `ModMenu` to the value returned by `MusicLoader.GetMusicSlot`.

### ExampleMod updates
Added example music boxes and fixed music in biomes.


Steviegt6:
Everything is functioning now, as far as I'm aware.

patches/tModLoader/Terraria/Audio/LegacyAudioSystem.cs.patch
```diff
@@ -47,12 +47,21 @@
...
++
  		public void LoadCue(int cueIndex, string cueName) {
  			CueAudioTrack cueAudioTrack = new CueAudioTrack(SoundBank, cueName);
+ 			TrackNamesByIndex[cueIndex] = cueName;
+@@ -244,7 +_,7 @@
+ 		}
+ 
+ 		public void UpdateCommonTrackTowardStopping(int i, float totalVolume, ref float tempFade, bool isMainTrackAudible) {
+-			if (!WaveBank.IsPrepared)
++			if (!WaveBank.IsPrepared || AudioTracks[i] is null) // should never really be null, save for some in-between loading if music is forced to play before mod loading is complete
```
Steviegt6, line 64: In my testing, it appeared that it could be temporarily null if you load a mod and it brings you to a ModMenu with custom music and attempts to play that music before mod loading is finished. It still plays fine once loading is complete.

Mirsario (approved):
Cleaned up and stuff. The way the file extensions are checked is concerning, but I think we can merge for now.

Issue #1677, opened 2021-08-02
[1.4] Extracting dlls out of mods doesn't work


saschahi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Extracting dlls out of mods doesn't work

### Log File

Found nothing about the extraction in there anyway.

https://pastebin.com/CReThPr8

### Steps to reproduce

1. Extract a mod
2. look for extracted dlls in "ModLoader\Beta\references\mods" or "ModLoader\references\mods"
3. Don't find them

### Expected Behavior

It should put extracted dlls in
"ModLoader\Beta\references\mods"
(was confirmed this is expected behaviour on discord)

### Actual Behavior

It doesn't put extracted dlls in
"ModLoader\Beta\references\mods"
and the entire "references\mods" path doesn't exist

### Reproduction frequency

tested it with one of my mods I updated from the old tModLoader version and 2 random mods I downloaded from the workshop

### Additional Information

The folders "references\mods" also weren't created.

I also tried to manually create this path but it still doesn't work.

Chik3r:
Have you tried going to the extracted mod's folder and seeing if the dll is there? If it isn't there, it might be that the developers of that mod aren't allowing people to extract the dlls. (Done by setting `hideCode` to `true` in `build.txt`)
If the dll is being extracted, but not copied, someone would have to fix that.

(Extracted mod folder is `Terraria\ModLoader\Beta\Mod Reader\ModName` I think)

saschahi:
> 
> 
> Have you tried going to the extracted mod's folder and seeing if the dll is there? If it isn't there, it might be that the developers of that mod aren't allowing people to extract the dlls. (Done by setting `hideCode` to `true` in `build.txt`)
> If the dll is being extracted, but not copied, someone would have to fix that.
> 
> (Extracted mod folder is `Terraria\ModLoader\Beta\Mod Reader\ModName` I think)

yep. i tried that. and no i don't have hidecode on (like I said, one of the mods I tried it on was my own mod)

and yep, I also looked there. there IS a dll that generated, but it's not the correct one.

JavidPack:
K, They will now end up in Mod Sources/Beta/Mod Libraries

PR #1681, opened 2021-08-02
Fix extracted dll not being copied to references

Chik3r:
### What is the bug?
#1677 
Extracted `.dll`s aren't copied to `references` because it doesn't expect dll files to be named `ModName.dll`

### How did you fix the bug?
Made the extractor know dll files can be named `ModName.dll`

Maybe

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
@@ -93,7 +93,7 @@ namespace Terraria.ModLoader.UI
 							src.CopyTo(dst);
 					}
 
-					// Copy the dll to ModLoader\Mod Libraries for easy collaboration.
+					// Copy the dll/xml to ModLoader/Mod Sources/Mod Libraries for easy collaboration.
 					if (name == $"{mod.Name}.dll") {
 						string modReferencesPath = Path.Combine(ModCompile.ModSourcePath, "Mod Libraries");
 						Directory.CreateDirectory(modReferencesPath);
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
@@ -94,7 +94,7 @@ namespace Terraria.ModLoader.UI
 					}
 
 					// Copy the dll to ModLoader\references\mods for easy collaboration.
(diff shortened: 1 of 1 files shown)


saschahi:
Hey, I'm the guy that created the issue.

just a heads up that in the pre-1.4 era the dlls would be extractes as modname_version.dll

for example: ProjectT_v0.5.5.dll

I'm not sure if that changed (I haven't dug into the 1.4 tModLoader source yet) but I just thought I should mention that since your fix doesn't seem to mention a version.

JavidPack:
This is probably fine, but we might be able to remove all the other options, and I'm not sure what our plans are for the references folder, since we aren't using it for the previous reasons anymore. I'll come back to this after we talk.

Chicken-Bones:
Lets get this done. Suggested path should be `Mod Sources/Mod Libraries`  or `mod_libraries` or something

patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
```diff
@@ -94,7 +94,7 @@ internal class UIExtractMod : UIProgress
 					}
 
 					// Copy the dll to ModLoader\references\mods for easy collaboration.
-					if (name == "All.dll" || PlatformUtilities.IsXNA ? name == "Windows.dll" || name == $"{mod.Name}.XNA.dll" : name == "Mono.dll" || name == $"{mod.Name}.FNA.dll") {
+					if (name == "All.dll" || PlatformUtilities.IsXNA ? name == "Windows.dll" || name == $"{mod.Name}.XNA.dll" : name == "Mono.dll" || name == $"{mod.Name}.FNA.dll" || name == $"{mod.Name}.dll") {
```
Steviegt6, line 97: Can't we get rid of the `.FNA.dll` check now, too, since they're always FNA?

steviegt6 (changes requested):

patches/tModLoader/Terraria/ModLoader/UI/UIExtractMod.cs
Steviegt6: Oh, yeah, we can also get rid of the legacy `.dll` checks (`Windows.dll` and `Mono.dll`), as well as `.XNA.dll` files. I *believe* they've all been streamlined to `ModName.dll`.

Chik3r:
> Suggested path should be `Mod Sources/Mod Libraries`

Made it copy the dll to that folder now

> Oh, yeah, we can also get rid of the legacy `.dll` checks

Removed them, now only copies `ModName.dll`

PR #1685, opened 2021-08-04
Fixed issue where Mod Browser sorting by recent did not work.

### LINKED ISSUE #1682: [1.4] Mod Browser's "Sort by recently updated" filter not working

- [x] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

The "Sort by recently updated" filter in the Mod Browser doesn't properly sort mods by their update times. It appears to sort correctly within groups, but then those groups are out of order.
Mod Browser Sorting

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/6922692/client.log)

### Steps to reproduce

1. Open the Mod Browser
2. Select the "Sort by recently updated" filter option
3. See the results

### Expected Behavior

Mods should be sorted by when they were last updated when using this option, with the most recently updated at the top, and the least recently updated at the bottom.

### Actual Behavior

Mods are all over the place. There seems to be a general tendency to have older mods at the bottom, but it can vary quite a bit going down the list.

### Reproduction frequency

100%

### Additional Information

_No response_

OliHeamon:
### What is the bug?
Fix #1682.

### How did you fix the bug?
Changed the way the timestamps are compared. I didn't look much into it, but it seems like the timestamps returned by the Steam Workshop are in a human-readable format, whereas the Mod Browser returned a Unix timestamp, This means the comparison (which was by string ordinal originally) broke when the format changed.

This fix just parses the timestamp into a `DateTime` object so that they can be compared properly.

No.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModDownloadItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModDownloadItem.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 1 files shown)


JavidPack:
Has an issue with assuming the original text is in the en-US culture, I'll push a fix independently.

Issue #1695, opened 2021-08-04
[1.4] Hammer change from walls mode to blocks mode.


htd-tstepien:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Maybe not everyone knows this, but in Vanilla Terraria when you start hitting a wall with the hammer and move the mouse on the screen, nothing will be destroyed except the walls. Only when you stop holding the mouse button and "attack" a block then the block will be modified (and you'll be not able to destroy walls).

### Steps to reproduce

1. Put a wall by the door
2. Move the hammer with the mouse during the destruction (without clicking) outside the wall (e.g. under the door)
3. The door will be destroyed because the hammer suddenly gains the ability to edit the block, besides destroying the walls

### Expected Behavior

A hammer shouldn't be able to destroy walls and edit blocks at the same time.

### Actual Behavior

A hammer is be able to destroy walls and edit blocks at the same time.

### Reproduction frequency

always

Mirsario:
Should now be fixed via #1698

PR #1698, opened 2021-08-04
Hammer fix

Ishigh1:
### What is the bug?
#1695 

### How did you fix the bug?
I found out that releaseUseItem was the culprit, it was set after the hammer use due to useItemFix, so I fixed it, verifying that the order of the other things using it was preserved according to vanilla

I don't think so, unless someone manages to understand how releaseUseItem is intended to work, I will stick to how vanilla works

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3547,9 +3547,9 @@
 +			// 2. Item animation is applied if button is pressed
 +			// 3. Item animation is reduced
 +			// 4. Hold / Use styles are invoked
-+			//    - ReleaseItem set at the end of this
-+			// 5. Item time is reduced
-+			// 6. Item logic applies item time
++			// 5. 'releaseUseItem' is set
++			// 6. Item time is reduced
++			// 7. Item logic applies item time
 +			//
 +			// TML:
 +			// 1. Item animation is reduced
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3547,6 +3547,7 @@
 +			// 2. Item animation is applied if button is pressed
 +			// 3. Item animation is reduced
 +			// 4. Hold / Use styles are invoked
++			//    - ReleaseItem set at the end of this
 +			// 5. Item time is reduced
 +			// 6. Item logic applies item time
```
Mirsario, line 3552:
```suggestion
+			// 4. Hold / Use styles are invoked
+			// 5. 'releaseUseItem' is set
+			// 6. Item time is reduced
+			// 7. Item logic applies item time
```

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -3556,6 +3557,7 @@
 +			// 3. Reuse delay is applied
 +			// 4. Item animation is applied if button is pressed
 +			// 5. Item logic applies item time
++			//    - ReleaseItem set at the beginning of this
 +			// 6. Hold / Use styles are invoked
```
Mirsario, line 3561:
```suggestion
+			// 5. 'releaseUseItem' is set
+			// 6. Item logic applies item time
+			// 7. Hold / Use styles are invoked
```

PR #1701, opened 2021-08-05
Made mod references work

Pbone3:
### What is the bug?
Loading a mod with references would always fail, because the assembly loading for referenced mods did not work in 1.4

### How did you fix the bug?
I made it so assemblies of referenced mods are loaded handled similarly to dllReferences, and added a resolve event that tries to load an assembly from loaded dllReferences or modReferences.

Yes. Chicken Bone's original vision with the new Assembly Load Context loading was to load mods and theirs dependencies into a single assembly load context, instead of relying on resolving assembly references. This is probably the preferable long term feature, though this one may still be helpful for developers trying to port their mod who simply need a working modReference system.


Chicken-Bones:
Fixed in https://github.com/tModLoader/tModLoader/commit/a18ceaf714aa3fec3091227b1840ce221d41a8b4

PR #1703, opened 2021-08-05
[1.4 Support bugs] Many Batch and Workshop Cleanup

Solxanich:
1) Batch files will now step people through manual portables installation when it fails to install.
2) Removed dependency on Vanilla finder.HasItem to determine if CreateItem() vs. UpdateItem(). We won't create new items if the item has already a publish ID assigned (reduces dupes)
3) added a few lines of logging around create and update
4) removed ending CreateItem result if WorkshopAgreement not done so that UpdateItem runs.
5) Reworked dependency on writing to manifest out of UpdateItem to be checked before doing anything.
6) Removed defunct code, refactored ForceCallbacks method to simplify for later.
7) Passed the HasUpdate bool to try and force downloading the Workshop Item again, even if Steam doesn't think there is an update.

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.cs.patch
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.cs.patch
@@ -125,7 +125,7 @@
  					ERemoteStoragePublishedFileVisibility visibility = GetVisibility(publicity);
  					_entryData = new SteamWorkshopItem {
  						Title = itemTitle,
-@@ -171,19 +_,31 @@
+@@ -171,19 +_,32 @@
  						ContentFolderPath = contentFolderPath,
  						Tags = tags,
  						PreviewImagePath = previewImagePath,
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.cs.patch
```diff
@@ -160,7 +162,43 @@
...
+ 						UpdateItem();
+ 					}
+@@ -233,17 +_,19 @@
+ 				private void UpdateItem() {
+ 					string headerText = GetHeaderText();
+ 					if (!TryWritingManifestToFolder(_entryData.ContentFolderPath, headerText)) {
+-						_endAction(this);
+-						return;
++						// If this fails to write, any future attempts to publish the item will always create a new item. We don't want that, so give clear message that something is wrong. - Solxan
++						Logging.tML.Fatal("Failed to write critical publishing information to the item content folder. Ending user interactions with the Workshop immediately to preserve integrity.");
```
Solxanich, line 197: Force ending the game instance is an excessive solution to #1706  which prevents tossing an exception.
The main comment remains the goal, but reviews say don't like current. Is there a good alternative?

Possibly consider switching to write to a static field the item publish info (publish ID, content folder source), and have any future attempts be blocked by the existence of real data in the field and display an appropriate error result prompt to inform why publishing ingame is blocked. Two main concerns with this: is convoluted and still doesn't stop someone from just restarting game to try again.

Chicken-Bones: People will definitely restart to try again. Is there nothing that prevents publishing a workshop item with the same name as one that already exists? Particularly with the same author? We could require they delete the old one? (Haven't followed most of the discussion around workshop so please excuse the question if it's dumb due to lacking context)

PR #1708, opened 2021-08-07
ZonePurity variable rename v2

Seraph-ED:
Alternative solution for the ZonePurity variable naming inconsistency based on Solxan's input on the previous PR, use THIS one instead of the former:

- instead of ZoneForest the the default biome flag is now named ZonePure and does not account for layers (the pure vs purity distinction was made because Pure implies a more general descriptor whereas Purity implies a specific zone, as ZonePure is now split up into different biome specific flags i went with the former, might be overthinking this a bit)

- ZoneForest is true if both ZonePure is true and the player is in the overworld layer, ZoneNormalUnderground and ZoneNormalCaverns are the same for the underground and cavern layers respectively

### DIFF
--- a/patches/tModLoader/Terraria/Player.TML.cs
+++ b/patches/tModLoader/Terraria/Player.TML.cs
@@ -118,6 +118,7 @@ namespace Terraria
 		public bool InModBiome(ModBiome baseInstance) => modBiomeFlags[baseInstance.ZeroIndexType];
 
 		/// <summary>
+<<<<<<< Updated upstream
 		/// The zone property storing if the player is in the purity/forest biome. Updated in <see cref="UpdateBiomes"/>
 		/// </summary>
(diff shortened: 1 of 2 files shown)


cozykoko:
why dont you just close the other one?

Seraph-ED:
i thought only the people with merge perms could do that, im a bit new to this sorry

Chicken-Bones:
Your files have some change markers left in them. Please have a look at the Files Changed panel here in github, and cleanup the formatting.

Also since these flags are 'derived' from other flags I suggest getting rid of the 

`InZoneX()` methods and replacing the properties with get-only accessors like

`public bool ZoneNormalCavern => ZonePure && ZoneRockLayerHeight;`

Mirsario:
The above PR implemented this one.

PR #1709, opened 2021-08-07
ILoadable.IsLoadingEnabled, ModProperties cleanup

pbone64:
### What is the new feature?
`ILoadable.IsLoadingEnabled(Mod)` controls if methods like `Mod.AddContent` should actually add this content.
The `ModProperties` cleanup was at the suggestion of Chicken Bones, and makes the properties easier to find

`ILoadable.IsLoadingEnabled(Mod)` replaces the functionality of 1.3's `bool Autoload(ref string)` method. It allowed modders to implement features such as items that can be disabled in configs.
`ModProperties` was rather convoluted to access before and you'd basically only know about it if you dove into tML code or someone told you about it. Moving it to properties in `Mod` makes its values much easier to find through VS autocomplete.

Yes. A wealth of alternative designs were discussed in the tModLoader discord's `#collaborators` channel. The current design is the one that me, Chicken Bones, and jopojelly came to agree on.

See `ExampleWings.cs`.

### ExampleMod updates
`ExampleModConfig` was added, explaining config attributes and including an option to disable `ExampleWings`. `ExampleWings` was given a `IsLoadingEnabled` implementation where it disables or enables it based on the `ExampleModConfig` option. `ExampleWings` was chosen as it was the item in 1.3 that showcased how to disable items through configs.


patches/tModLoader/Terraria/ModLoader/ILoadable.cs
```diff
@@ -11,6 +11,12 @@ public interface ILoadable
 		/// <param name="mod">The mod instance associated with this type.</param>
 		public abstract void Load(Mod mod);
 
+		/// <summary>
+		/// Whether or not this type should be loaded when it's told to. Returning false disables <see cref="Mod.AddContent"/> from actually loading this type
+		/// </summary>
+		/// <param name="mod">The mod instance trying to add this content</param>
+		public abstract bool IsLoadingEnabled(Mod mod);
```
Chicken-Bones, line 18: default interface implementation for this?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ILoadable.cs
Pbone3: Ah, forgot to change that back when I was messing with it, trying to get it to show up on ModTypes (I'm not too familiar with interfaces implementating each other). Fixed now

PR #1714, opened 2021-08-08
Add keybind to 1.4 ExampleMod

sdegueldre:
This commit adds the random buff keybind that was present in 1.3 ExampleMod to 1.4 ExampleMod

I've taken the approach to create a separate player class for this as has been done for luck and inventory, let me know if this is not desirable.

I'm not sure if I should create a separate class for registering the keybind that would also have an unload method that would set the static field to null, kind of like what is currently done for the recipes.

As a side note, I'm getting the warning that ExampleMod did not fully unload whenever I reload the mods, but I was getting this warning even before my commit so I'm not sure if I should be concerned about that.

This is my first PR on this project so please be kind if I've forgotten anything or missed the mark on some of these changes, and advise as to any changes that are needed. Thank you for your time!


JavidPack:
Thanks a bunch!

sdegueldre:
Thanks for the merge!

Issue #1728, opened 2021-08-12
Multiple Language Support in Workshop


Solxanich:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

Yes

### Description

Building on #1365 , would be good to be able to supply multiple translations to Steam for Workshop items.

The ISteamUGC interface allows for setting and getting translated descriptions from a workshop item, and having multiple thereof.
It would be trivial to add from a Workshop perspective, but from a mod building and packaging perspective some work would be needed.

### What does this proposal attempt to solve or improve?

TModLoader, now that it is on Workshop, is seeing a massive increase in Chinese player-base visibility as the Great Firewall moves out of the picture and Chinese modders can get their mods up.

Being able to have both English and the appropriate dialect of Chinese (I presume Mandarin) available would greatly improve player QoL. 

Likewise, there is a sizeable Russian community that stands to benefit as well.

Steviegt6:
I've touched on this myself before in `#collaborators`. Simply appending the language code to `description` should suffice functionally, or have `description.txt` default to `en-US` for legacy's sake and append other language codes (i.e. `desctipn-zh-Hans.txt`) for localized descriptions. As for mod names, there's been some different opinions on their practicality, but it'd be as easy to implement on our side as just adding more optional `build.txt` fields (such as `modName-zh-Hans` or something).
We'd wanna also modify existing tML code to display translations.

Solxanich:
May also suggest the possibility of separating out the language config files from the .tmod file and into a Localization folder delivered alongside the .tmod (since Steam Workshop is a folder of items).
Than it would be trivial to move the existing localizable build.txt fields in to those config files, and fetch them as needed (although it is still likely implementable even if we don't separate out from the .tmod files). 
May also be other benefits from such changes, such as furthering goal of eliminating build.txt in favor of using csproj directly.

Issue #1727, opened 2021-08-11
[1.4] Visit Workshop Button in the Mod Info Section


Rijam:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

No

### Description

Add a button to visit the mod's Steam Workshop page in the mod info section in addition to the normal "visit the mod's homepage" button.

Here is a mock-up image that I made.
![tMLFeatureRequest](https://user-images.githubusercontent.com/11262234/129108722-ee3f7b6b-db59-4e8f-aa9a-704346fa7fc9.png)

### What does this proposal attempt to solve or improve?

It would allow users to easily visit any mod's Steam Workshop page while in game without having to manually search for the mod on the Workshop themselves. This would be beneficial if the user wanted to see the comments on the mod's Workshop page, see how many subscribers it has, or look at the version history.

JavidPack:
Would people want this to open in their browser? or in the steam overlay browser?

Rijam:
> Would people want this to open in their browser? or in the steam overlay browser?

Personally I would prefer the Steam Overlay, but for GOG users it would have to be the browser. If it is easier just to make it the browser for everyone, then that would work too.

Ishigh1:
Steam browser isn't exactly the best, I'd prefer the favorite browser, but it could be something asked when opening workshop for the first time, with an option to remember

Steviegt6:
> Would people want this to open in their browser? or in the steam overlay browser?

Configurable? Default to overlay.

Rijam:
Thanks Solxan and everyone else!

Issue #1724, opened 2021-08-11
[1.4 Alpha] Tell Players to Check Steam Logs After Workshop Failures


Steviegt6:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

No

### Description

When Steam relays a failure, such as the generic `k_EResultFail` as demonstrated in #1716, the error screen should tell the user that receiving support requires checking their Steam-specific workshop logs (`C:\Program Files (x86)\Steam\logs\workshop_log.txt` on windows).

### What does this proposal attempt to solve or improve?

This proposal attempts to make debugging issues with the workshop easier. It's hard to communicate through GitHub's issue tracker when the issue-maker does not make a follow-up post or doesn't quite understand how to use GitHub. Telling them that providing the log file is required will reduce the amount of context-less issues and makes debugging easier for everyone.

### Which (other) solutions should be considered?

N/A as far as I'm aware, but we could also directly log everything to our logs. Not entirely sure how Steam logs their workshop stuff, though, so that may not be viable.

Issue #1719, opened 2021-08-09
[1.4] Updating Enabled Mods Refuses to Replace Files


Steviegt6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

When updating mods in-game through the Mod Browser, if a mod you're attempting to update is enabled and the file isn't released, an update will still be attempted and will fail, meaning that the mod still requires updating but does not give the user any feedback regarding the issue.

### Steps to reproduce

1. Enable a mod that needs updating.
2. Update all mods or any single mod.
3. Observe the attempted update result in no modified files.

### Expected Behavior

Attempting to update an enabled mod should either:
1. Actually replace the file.
2. Cancel with some form of feedback.
3. Schedule the update for during a mod reload or something of that sort.
4. Not appear as updatable when the mod is enabled.

### Actual Behavior

There's zero feedback and it appears as though the mod cannot be updated.

### Reproduction frequency

100% reproduction frequency. This happens with all mods that have updates.

PR #1732, opened 2021-08-12
Collection of Workshop Improvements

Solxanich:
ToDos:
- [x]  #1727
- [x]  Workshop to Mod Browser support for: #1728
- [x]  #1724 
- [x] Fix GoG calling SteamAPI.Init(), thus eliminating dedicated servers preventing people from running the client
- [x] Rework UIModDownloadItem into two parts: a UIModDownloadItem with graphics, and a ModDownloadItem with all the metadata so as to resolve leftover null check ToDos from #1726 
- [x] Actually implementing Playtime Tracking
- [x] Investigate PublishId available from the local mod instead of relying exclusively on the Mod Browser to load
- [x] Investigate threading the fetching of mods data on the browser based on the page level queries, so can scale better speed wise.
- [x] Pre-validate description, title length 
- [x] Fix Mod updating for unloading/reloading #1719 
- [x] Fix UIPublishMod visual errors
- [x] #1716  Fix Mod Browser query breaking when exact multiple of 50 items on the Workshop.
- [x] #1746 Fix localization missing in Workshop implementation.


Mirsario (changes requested):
Useless 8/16 partial review. More later.

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -279,77 +328,174 @@ internal class ModManager
...
-				// Call the appropriate variant
-				if (SteamUser)
-					SteamUGC.StopPlaytimeTracking(modsById, (uint)modsById.Length);
-				else
-					SteamGameServerUGC.StopPlaytimeTracking(modsById, (uint)modsById.Length);
+				if (!new AQueryInstance().GetPageCountFast())
+					return false;
+
+				//TODO: Threaded query of multiple pages at a time appears un-supported within Steam.
+				// Concerned for future speed of querying the Steam Workshop (ie at 1000+ items).
```
jackbondpreston, line 404: Have you tried using the version which uses a cursor[1][2] instead of page number? It _may_ be faster[3], and I think the current method might be limited to 1000 pages total[3]. Should also cut out some pagination logic.

[1] https://store.steampowered.com/news/group/4145017/old_view/1751242076600910671
[2] https://github.com/rlabrecque/Steamworks.NET/blob/9416761cab7fa6b226803c95a0cc7ab8dd03b43c/CodeGen/steam/isteamugc.h#L211
[3] https://github.com/rlabrecque/Steamworks.NET/blob/9416761cab7fa6b226803c95a0cc7ab8dd03b43c/CodeGen/steam/isteamugc.h#L386

jackbondpreston: If this turns out to be parallelisable, I imagine that would be a lot faster. But if not this is probably worth pursuing.

Solxanich: Hmm, I don't see that in the standard documentation: https://partner.steamgames.com/doc/api/ISteamUGC#CreateQueryAllUGCRequest
That said, when the release notes say its deprecated, and with the upgrade Mirsario did to Steamworks.Net 15, I guess that is the new proper method.
Thanks for pointing this out, much easier to deal with now than later.

Solxanich: Made the switch, improved the speed somewhat, but if the method is deprecated and subject to removal, its best we don't keep using the old method. 

Current performance is roughly 0.4 seconds per 50 items (or 2 seconds per 250 items). Not bad, but at 2000 items plus, that's 16 seconds of waiting. Still a concern, and parallellization still isn't possible.

patches/tModLoader/Terraria/ModLoader/UI/UIModInfo.cs
```diff
@@ -183,6 +214,15 @@ internal class UIModInfo : UIState
 			Utils.OpenToURL(_url);
 		}
 
+		private void VisitModSteamPage(UIMouseEvent evt, UIElement listeningElement) {
+			SoundEngine.PlaySound(10);
+			if (Social.Steam.WorkshopHelper.ModManager.SteamUser && Steamworks.SteamUtils.IsOverlayEnabled() && (Main.graphics.IsFullScreen || Main.graphics.PreferredBackBufferWidth >= 1920 || Main.graphics.PreferredBackBufferHeight >= 1080))
+				Steamworks.SteamFriends.ActivateGameOverlayToWebPage("http://steamcommunity.com/sharedfiles/filedetails/?id=" + _publishedFileId, Steamworks.EActivateGameOverlayToWebPageMode.k_EActivateGameOverlayToWebPageMode_Modal);
+			else
+				Utils.OpenToURL("http://steamcommunity.com/sharedfiles/filedetails/?id=" + _publishedFileId);
```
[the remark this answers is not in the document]
Solxanich: Hmm, debate this one. Don't follow the logic as the overlay isn't always preferable; at low resolutions, the web browser built into Steam Overlay is pretty pathetic. Doesn't scale well since the overlay browser lacks a zoom feature, as an example:
image

I should have included a comment explaining that in the code.

And all the checks for verifying the user is logged in occurs elsewhere; no reason why viewing info about a mod should require yet another check, so not following the "require logged in to Steam logic", SteamAPI.Init() is responsible for that in the first place.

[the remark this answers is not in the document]
Mirsario: The lowest resolution we should care about is 1280x720. If it looks at least barely usable in that - overlay should be preferred.

Mirsario (approved):

patches/tModLoader/Terraria/ModLoader/UI/UIModInfo.cs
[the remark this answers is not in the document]
Mirsario: More than usable:
image

PR #1734, opened 2021-08-13
[1.4] [breaking] (On)ConsumeAmmo fixes and refactor

direwolf420:
### What is the bug?
1. GlobalItem.ConsumeAmmo was not called for `item` globals
2. GlobalItem.OnConsumeAmmo was calling `ammo` globals but used `item`

### How did you fix the bug?
Use proper variables (clarify `item` usage to specify `weapon` and `ammo`).
Refector the hooks:
* `ConsumeAmmo` -> `CanConsumeAmmo` + `CanBeConsumedAsAmmo`
* `OnConsumeAmmo` -> `OnConsumeAmmo` + `OnConsumedAsAmmo`
* Add them to `CombinedHooks`

Add the corresponding "other item" to the ModItem hooks?


Mirsario:
The feedback in discord was that ConsumeAmmo hooks should be split into ConsumeAmmo and ConsumeAsAmmo, to avoid confusion.
Or ConsumeAmmoItem for the first name, for even less?

direwolf420:
well, one is for items to control whether their use will consume ammo (possibly based on the ammo it shoots), the latter is for the ammo itself determining that (possibly based on the item that shoots it)

ConsumeAmmoItem sounds clunky, the pendant would be ConsumeAmmoAmmo?

patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
```diff
@@ -36,6 +36,15 @@ public static class CombinedHooks
 			PlayerLoader.OnMissingMana(player, item, neededMana);
 		}
 
+		public static bool CanConsumeAmmo(Player player, Item weapon, Item ammo) {
+			return PlayerLoader.CanConsumeAmmo(player, weapon, ammo) & ItemLoader.CanConsumeAmmo(weapon, ammo, player);
+		}
```
Mirsario, line 41: CanX hooks are to use short-circuiting. 
```suggestion
		public static bool CanConsumeAmmo(Player player, Item weapon, Item ammo) {
			return PlayerLoader.CanConsumeAmmo(player, weapon, ammo) && ItemLoader.CanConsumeAmmo(weapon, ammo, player);
		}
```

direwolf420: What about CanUseItem, it uses a single `&` too

Mirsario: The fact that official TML examples misused a hook is utterly bad. All CanX hooks should short-circuit on all levels, unless their return values are nullable. For DualUseWeapon, there should be something else.

PR #1735, opened 2021-08-13
[1.4] Ported RecipeHelper

ExterminatorX99:
Ported RecipeHelper


ExampleMod/Content/ExampleRecipes.cs
```diff
@@ -21,6 +21,10 @@ public static class ExampleRecipes
 			Examples_AddRecipes(mod);
 
 			Examples_ModifyRecipes(mod);
+
+			// See RecipeHelper.cs
+			RecipeHelper.AddExampleRecipes(mod);
+			RecipeHelper.ExampleRecipeEditing(mod);
```
Steviegt6, line 27: Would probably make more sense to make use of an `ILoadable`'s `Load()` or `ModSystem`'s `OnModLoad()` to do this.

ExampleMod/Content/RecipeHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+				.Where(recipe => items.All(item => recipe.requiredItem.Any(reqItem => reqItem.type == item.type && reqItem.stack >= item.stack)));
+
+			// loop every recipe found by the finder
+			foreach (Recipe r in recipes) {
+				r.requiredItem.RemoveAll(req => req.type == ItemID.Chain); // for the currently looped recipe, delete the Chain ingredient.
+			}
+
+			// The following is a more precise example, finding an exact recipe and deleting it if possible.
+			List<int> groups = new() {
+				RecipeGroupID.IronBar // add a new recipe group, in this case the vanilla one for iron or lead bars.
```
Steviegt6, line 82: May be worthwhile to explain how to retrieve modded group IDs?

steviegt6 (changes requested):
Looks pretty good to me, though I feel like it could use some more examples that focus specifically on modded items and groups. I understand that this is a port of a previously-existing class from the original ExampleMod, but there's no reason to avoid improvement.

ExampleMod/Content/RecipeHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content
+{
+	// In this class we separate recipe related code from our main class
+	public static class RecipeHelper
+	{
+		// Here we've made a helper method we can use to shorten our code.
+		// This is because many of our recipes follow the same terminology: one ingredient, one result, one possible required tile
+		private static void MakeSimpleRecipe(Mod mod, string modIngredient, short resultType, int ingredientStack = 1, int resultStack = 1, string reqTile = null)
```
Steviegt6, line 14: Ah, by the way: I believe we want to generally move away from using string to get instances of loaded types. We could use this method as an example of generics as well.

steviegt6 (approved):
Looks good now, only thing I can say is that it may be cleaner to make an overload for tiles instead of just inputting a generic hen it isn't needed.

Mirsario:
> I'm not sure on what this file teaches people. It looks like a leftover from when locals were a must for creating recipes, and so the process was way more annoying.

Steviegt6:
> > I'm not sure on what this file teaches people. It looks like a leftover from when locals were a must for creating recipes, and so the process was way more annoying.

That appears to be the case, but I feel like giving a couple examples of how to deal with recipe creation through slightly more unusual ways could prove beneficial to some people, even if it's an extreme minority. I suppose a merge isn't required but I do believe some could find this useful, even if they just use it to get a basic concept of how to deal with recipes.
Some methods in `RecipeHelper` are definitely less helpful than others, even as examples, though.

PR #1738, opened 2021-08-13
Recipe Ordering

Ishigh1:
### What is the new feature?
I added 3 new methods to allow ordering recipes.

Usually, new recipes are placed at the end of the recipe list, meaning the crafting menu shows the recipe of different mods separately, so all the recipes separated by mods, instead of being grouped by utility or materials. 

I chose to leave MoveAt() public as Main.recipe is public anyway, but it could be made internal or private as it utility hould be marginal (an could only be used to put a recipe as first or last.

### Secondary changes
To make recipe search more reliable, recipes are only deleted after the recipe setup phase.
Also to allow recipes to be moves without introducing a new variable, Recipe.Register() returns the recipe.

Any iron recipe can be put after the vanilla ones instead of at the end of the recipe list.

### ExampleMod updates
I only updated the wings to place it before the wings as the wings are all together in the recipe list, couldn't figure out a nice ordering for the others.


patches/tModLoader/Terraria/ModLoader/RecipeLoader.cs
```diff
@@ -57,6 +57,28 @@ public static class RecipeLoader
 			}
 		}
 
+		/// <summary>
+		/// Deletes the recipes flagged for deletion
+		/// </summary>
+		internal static void DeleteRecipes() {
```
Solxanich: I'm not a fan of the idea of recipes being deleted.
This easily can run away in to some weird cross mod interactions with null/missing objects.

Let's focus on disabling recipes instead by adding an always false condition to them. This poses less risk to cause unintended cross-mod interactions

Issue #1672, opened 2021-08-02
[1.4] Researching Modded Items and disabling the Mod makes players not save correctly


Mayonnaisinator:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

There are two branches to this issue depending on steps taken:

1. If a modded item is researched in Journey mode, and then the mod is disabled, the player file doesn't save correctly and reverts back to before the item was researched.
2. Alternatively, if you research, close the world, re-open the world with the player, and _then_ leave and disable the mod, the player becomes unplayable. Clicking on the player on the menu does nothing, and the player appearance reverts to default, and it says "Classic", even if the player is a Journey mode character. Re-enabling the mod does then restore the ability to play as that player.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/6916000/client.log)

### Steps to reproduce

1. Research a modded item in Journey mode, and make changes to your inventory so that it is noticeable
2. Exit the world and disable the mod
3. Re-open a world using that player, and the player will have reverted back to before the item was researched, so the inventory will be back how it was before

Alternatively:
1. Research a modded item in Journey mode, and make changes to your inventory so that it is noticeable
2. Exit the world, and then re-open a world using that player
3. Exit again, and now disable the mod
4. Try selecting the player on the menu, but you won't be able to as clicking it does nothing. The player's apperance will have lost their armour/vanity, and it will say that the player is a classic player, despite being a Journey player
5. Re-enable the mod and the player can be selected again

### Expected Behavior

If a modded item is researched and the mod is disabled, the modded item should disappear from the research menu. Upon re-enabling the mod, that item should re-appear in the research menu. If an item is only partially researched, its progress should also be preserved. Obviously, the player's inventory should save correctly throughout all of this.

### Actual Behavior

Player's inventory/state isn't saving when modded items are researched. Note that this actually also happens if a modded item is only partially researched (e.g. 10/100), so it doesn't even have to be fully researched and appear in the menu for the issue to happen.

### Reproduction frequency

I can get this to happen with 100% frequency following the steps above. I used ExampleMod to test this extensively, but I did have this happen with Magic Storage yesterday.

Erisfiregamer1:
Hmm, try importing the character to Terrasavr and removing all research. Does that fix it?

Mayonnaisinator:
> 
> 
[...]

Issue #1659, opened 2021-08-01
[1.4] Journey Mode Duplication Unload Preservation Isue


Solxanich:
### Description
whenever you duplicate something from a mod then unload the mod, the duplicated items within the menu gets replaced with items, not just random items but items from a mod of priority. So, one the example mods items gets replaced with the luminite storage unit if you have magic storage installed, its been consistent.
- From TheControllerWolf in discord.

Expected problem: those IDs are changed at runtime, so if something matched the id of another mod at the time while it was loaded, would assume that causes that issue

PR #1740, opened 2021-08-14
[1.4] Fix player research saving

direwolf420:
### What is the bug?
#1659, #1672
Research data was using `persistentId` to save modded content, which is incompatible with how modded data is usually saved.

### How did you fix the bug?
To preserve research state for unloaded mods/items, it now saves using mod/name pairs to handle safe retrieval.

This brings it in line with other saving (such as unloaded `ModPlayer` data, or unloaded kill counts).

I also included a check to not carry over the old data pre-PR that was on the player regarding `persistentId` to not keep it as baggage in `unloadedResearch`. Please tell me if I should remove it again.

```cs
if (tag.ContainsKey("persistentID"))
    continue;
```

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
@@ -101,14 +101,14 @@ namespace Terraria.ModLoader.IO
 
 		public static List<TagCompound> SaveResearch(Player player) {
 			var list = new List<TagCompound>();
-			Dictionary<string, int> dictionary = new Dictionary<string, int>(player.creativeTracker.ItemSacrifices._sacrificeCountByItemPersistentId);
-			foreach (KeyValuePair<string, int> item in dictionary) {
-				ContentSamples.ItemNetIdsByPersistentIds.TryGetValue(item.Key, out int netID);
-				ContentSamples.ItemsByType.TryGetValue(netID, out Item realItem);
-				if (ItemLoader.NeedsModSaving(realItem)) {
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
```diff
@@ -118,12 +118,23 @@ internal static class PlayerIO
...
 		public static void LoadResearch(Player player, IList<TagCompound> list) {
 			foreach (var tag in list) {
-				ContentSamples.ItemNetIdsByPersistentIds.TryGetValue(tag.GetString("persistentID"), out int netID);
-				ContentSamples.ItemsByType.TryGetValue(netID, out Item realItem);
-				if (ItemLoader.NeedsModSaving(realItem)) {
-					player.creativeTracker.ItemSacrifices._sacrificeCountByItemPersistentId[tag.GetString("persistentID")] = tag.GetInt("sacrificeCount");
-					if (ContentSamples.ItemNetIdsByPersistentIds.TryGetValue(tag.GetString("persistentID"), out int value2))
-						player.creativeTracker.ItemSacrifices.SacrificesCountByItemIdCache[value2] = tag.GetInt("sacrificeCount");
+				if (tag.ContainsKey("persistentID"))
+					continue; //Discard tags from previous insufficient implementation pre-alpha so they are not carried over to unloadedResearch
```
Mirsario, line 122: Same thing without referencing the past.
```suggestion
				if (!tag.ContainsKey("mod") || !tag.ContainsKey("name"))
					continue; // Discard tags from previous insufficient implementation pre-alpha so they are not carried over to unloadedResearch
```

Mirsario (approved):

PR #1741, opened 2021-08-14
[1.4, breaking] Lots of DamageClass-related fixes and improvements

ThomasThePencil:
This PR aims to once again overhaul DamageClass by adding a lot of new functionality, alongside fixing some still-free-roaming errors that weren't caught with the original overhaul and adding a lot of long-missing functionality related to controlling the stats which `DamageClass` covers.

### Additions
- Renamed `DamageClass.NoScaling` to `DamageClass.Default`.
- All items and projectiles now use `DamageClass.Default` by default instead of `DamageClass.Generic`. **This should fix a LOT of vanilla item/projectile stat scaling problems!**
- All vanilla summon projectiles now use `DamageClass.Summon` as their `DamageType`. This shouldn't cause any stat scaling discrepancies with vanilla, although extra testing never hurts.
- Added a new hook in `DamageClass` called `AllowStandardCrits` --- this allows you to enable or disable standard crit calcs for all items/projectiles of a given damage type. This hook is used for both `DamageClass.Default` and `DamageClass.Summon`.
- `Projectile` now has a `<T>()` overload for `CountsAsClass`, just as `Item` does.
- Added a hook in `DamageClass` allowing easy visibility control for various common statistical tooltip lines. This will be made obsolete once that tooltip overhaul happens, but until then, this'll work just fine.
- Implemented the mythical second half of the crit fix (make projectiles set their own critical strike chance on creation and keep that chance synced, then check against it when striking enemies).
- Added new fields in `DamageClass`: attack speed and armor penetration.
- Armor penetration may now also be freely modified for `Item`s and `Projectile`s.
- Updated ExampleMod with a bunch of new examples for DamageClass features new and old. This doubles as my way of testing the waters for a new approach to ExampleMod which, if it goes over well, can easily be applied to a lot of the rest of ExampleMod.

#### Vanilla
- The Sharpening Station's armor penetration boost can no longer be applied to non-melee attacks (this was easiest to pull off with summons) by holding a melee weapon.
- The long-standing and equally infamous 'crit swap' exploit is no longer possible.
#### tML
- Minion shots and the like are now actually able to hit things again, if they weren't before.
- Whips and various other summon attacks are no longer able to critically strike in situations where they shouldn't be able to.


direwolf420:
(I somehow can't link code in this PR, maybe cause it's a draft idk)
The part where you comment out some cases in the `switch (type) {` switch should include a comment for why it was commented out and where to find the corresponding new code.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
```diff
@@ -1,65 +1,68 @@
...
 
 namespace ExampleMod.Content.DamageClasses
 {
 	public class ExampleDamageClass : DamageClass
 	{
 		public override void SetStaticDefaults() {
-			// Make weapons with this damage type have a tooltip of 'X example damage'.
+			// This makes weapons with this damage type have a tooltip of 'X example damage'.
+			// TO-DO: Move to .hjson file and actually teach people how they work.
 			ClassName.SetDefault("example damage");
```
Solxanich: This is an easy TODO, and should be done before merging as such.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
```diff
@@ -1,65 +1,68 @@
...
 		}
 
 		protected override float GetBenefitFrom(DamageClass damageClass) {
-			// Make this damage class not benefit from any otherclass stat bonuses by default, but still benefit from universal/all-class bonuses.
+			// This method lets you make your damage class benefit from other classes' stat bonuses by default, as well as universal stat bonuses.
```
Solxanich, line 17: This is a lot of good documentation - let's try to make sure a summary of it ends up on the Mod Migration wiki page, as no doubt there will be further questions on this to come.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
```diff
@@ -1,65 +1,68 @@
...
+			// Default is, you guessed it, the default damage class. It doesn't scale off of any class-specific stat bonuses or universal stat bonuses.
+			// There are a number of items and projectiles that use this, such as thrown waters and the Bone Glove's bones.
+			// Generic, on the other hand, scales off of all universal stat bonuses and nothing else; it's the base damage class upon which all others that aren't Default are built.
 			if (damageClass == DamageClass.Generic)
 				return 1f;
 
-			// Now, I know you're just dyin' to get into the fun side of things, so let's have ourselves some demonstrations.
-			// Feel free to uncomment one of the below! Experiment, play with the variables a bit. See what works best for you!
+			return 0f;
+			// To explain how the return value works, it behaves like a percentage, with 0f being 0%, 1f being 100%, and so on.
```
Solxanich: Shouldn't this comment block be above the 'return 0' statement, per the style guidance?

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
```diff
@@ -1,65 +1,68 @@
...
+		}
 
-			// PROMPT: You want your damage class to benefit at a much higher rate from a vanilla class' stat boosts.
-			// The below makes your class benefit at 500% effectiveness from all magic stat bonuses.
-			/*
+		public override bool CountsAs(DamageClass damageClass) {
+			// This method allows you to make your damage class benefit from otherclass effects (e.g. Spectre bolts, Magma Stone) based on what returns true.
+			// Note that unlike GetBenefitFrom, you do not need to account for universal bonuses in this method.
+			// For this example, we'll make our class count as melee and magic for the purpose of class-specific effects.
+			if (damageClass == DamageClass.Melee)
```
Solxanich, line 40: I get the feeling that someone may raise that this may be a place where being declarative might be better.

I might suggest considering if you can write it as ``Is 'damageClass' in {DamageClass.Melee, DamageClass.Magic}`` in all cases, or if this has a reason to be a method.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
```diff
@@ -1,65 +1,68 @@
...
+			// This method lets you set default statistical modifiers for your example damage class.
+			// Here, we'll make our example damage class have more critical strike chance and armor penetration than normal.
+			player.GetCritChance<ExampleDamageClass>() += 4;
+			player.GetArmorPen<ExampleDamageClass>() += 10;
+			// These sorts of modifiers also exist for damage (GetDamage), knockback (GetKnockback), and attack speed (GetAttackSpeed).
+			// You'll see these used all around in referencce to vanilla classes and our example class here.
+		}
 
-			return 0;
+		public override bool AllowStandardCrits() {
```
Solxanich: This should be a property; I can't see any reason to have a method.

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -28,6 +28,15 @@ public partial class Item : TagSerializable, IEntityWithGlobals<GlobalItem>
 			set => _damageClass = value ?? throw new ArgumentException("DamageType cannot be null");
 		}
 
+		private int _armorPenetration = 0;
+		/// <summary>
+		/// The number of defense points that this item can ignore on its own. Cannot be set to negative values. Defaults to 0.
+		/// </summary>
+		public int ArmorPenetration {
+			get => _armorPenetration;
+			set => _armorPenetration = Math.Max(value, 0);
```
Solxanich: I'm concerned this may cause confusion when some mod inevitably sets it to a negative value and successfully builds and runs it without testing this in particular.
Consider tossing an exception instead so the modder is obligated to fix the stupid on their end.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1712,28 +1713,34 @@
...
  						toolTipLine[numLines] = num3 + Lang.tip[5].Value;
-@@ -15152,6 +_,14 @@
+@@ -15152,8 +_,16 @@
  						toolTipLine[numLines] = num5 + Lang.tip[5].Value;
  						numLines++;
  					}
 +					*/
-+					// TODO: make this not inherently exclude summons. minions should be allowed to crit in the right circumstances, too
-+					if (!item.summon) {
++
```
Solxanich: unnecessary extra line?

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1712,28 +1713,34 @@
...
-@@ -15152,6 +_,14 @@
+@@ -15152,8 +_,16 @@
  						toolTipLine[numLines] = num5 + Lang.tip[5].Value;
  						numLines++;
  					}
 +					*/
-+					// TODO: make this not inherently exclude summons. minions should be allowed to crit in the right circumstances, too
-+					if (!item.summon) {
++					
++					if (item.DamageType.AllowStandardCrits() && item.DamageType.ShowStatTooltipLine(player[myPlayer], "CritChance")) {
```
Solxanich: I don't recall seeing any example mod comment or documentation explaining that you need to have both aspects set:
AllowCrits is true, and is showing that it has Crit Chance. 
Ensure that this behaviour is covered someplace.

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -5,8 +5,8 @@ namespace Terraria.ModLoader
 {
 	public abstract class DamageClass : ModType
 	{
+		public static DamageClass Default { get; private set; } = new DefaultDamageClass();
```
Solxanich, line 8: For support sanity's sake, please add "[Breaking Changes]" to the title so people have a higher chance of clueing in.

patches/tModLoader/Terraria/ModLoader/DamageClassData.cs
```diff
@@ -3,13 +3,17 @@
 	internal struct DamageClassData
```
Solxanich: Any reason for internal instead of public?
Is there a reason not to allow mods to access members of this struct/instantiate this struct? I imagine several mods would want to be able to read the active player stat modifiers

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -56,6 +56,10 @@ public class SummonDamageClass : VanillaDamageClass
 		protected override string LangKey => "LegacyTooltip.53";
 
 		protected override float GetBenefitFrom(DamageClass damageClass) => damageClass == Generic ? 1f : 0f;
+
+		public override bool AllowStandardCrits() => false;
```
Solxanich: This line looks odd at a glance.
Mind adding a comment explaining in to the code?
I would have expected this to appear in SummonDamageClass instead.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -2548,10 +2575,34 @@
...
+ 			num32 *= ItemID.Sets.BonusMeleeSpeedMultiplier[inventory[selectedItem].type];
+ 			meleeSpeed = 1f + num32;
++			*/
++			meleeSpeed = new StatModifier(1 + ((meleeSpeed.Additive - 1) * ItemID.Sets.BonusMeleeSpeedMultiplier[inventory[selectedItem].type]), meleeSpeed.Multiplicative);
+ 			if ((double)pickSpeed < 0.3)
+ 				pickSpeed = 0.3f;
+ 
++			// TO-DO: potentially add a setting in Mod that can make this not a thing
+-			if (meleeSpeed > 3f)
++//			if (meleeSpeed > 3f)
```
Solxanich: The typical style recommended for commenting out code is as follows:
``\\code``
On the other hand, for comments, just for comparison
``\\ Comment``

This Style is to help tell at a glance whether a comment is code or guidance

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -2548,10 +2575,34 @@
...
++				meleeSpeed = new StatModifier(3f, meleeSpeed.Multiplicative);
+ 
+ 			if (tileSpeed > 3f)
+ 				tileSpeed = 3f;
+@@ -18146,7 +_,7 @@
+ 				}
  			}
  
- 			meleeSpeed = 1f / meleeSpeed;
+-			meleeSpeed = 1f / meleeSpeed;
```
Solxanich, line 2608: Accidental removal instead of comment out?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4066,7 +4121,10 @@
  			int projToShoot = sItem.shoot;
  			float speed = sItem.shootSpeed;
  			int damage = sItem.damage;
-@@ -32542,7 +_,9 @@
+ 			if (sItem.melee && projToShoot != 699 && projToShoot != 707 && (uint)(projToShoot - 877) > 2u)
+-				speed /= meleeSpeed;
++				speed *= meleeSpeed;
```
Solxanich: A commentary here please? No context here for why it was switched.

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -14,18 +15,55 @@ public partial class Projectile : IEntityWithGlobals<GlobalProjectile>
...
 		}
 
+		private int _armorPenetration = 0;
+		/// <summary>
+		/// The number of defense points that this projectile can ignore on its own. Cannot be set to negative values. Defaults to 0.
+		/// On spawn, if this projectile was fired frrom a weapon, this value has the total armor penetration of the weapon that made the projectile added to itself.
+		/// </summary>
+		public int ArmorPenetration {
+			get => _armorPenetration;
+			set => _armorPenetration = Math.Max(value, 0);
```
Solxanich: Same comment as in the other spot for ArmorPenetration.

Likewise for CritChance

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -14,18 +15,55 @@ public partial class Projectile : IEntityWithGlobals<GlobalProjectile>
...
+		/// The critical strike chance modifier of this projectile. Cannot be set to negative values. Defaults to 0.
+		/// On spawn, if this projectile was fired frrom a weapon, this value has the total critical strike chance of the weapon that made the projectile added to itself.
+		/// </summary>
+		public int CritChance {
+			get => _crit;
+			set => _crit = Math.Max(value, 0);
+		}
+
+		private static void HandlePlayerStatModifiers(IProjectileSource spawnSource, Projectile projectile) {
+			if (spawnSource is null || !(spawnSource is ProjectileSource_Item || spawnSource is ProjectileSource_Item_WithAmmo))
```
Solxanich: Minor performance recommendation:
Swap from ``!( a || b)`` to ``!a && !b`` so that short circuiting kicks in more often.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -555,6 +675,24 @@
  				localAI[0] = 1f;
  			}
  		}
+@@ -26715,7 +_,7 @@
+ 					break;
+ 			}
+ 
+-			float meleeSpeed = player.meleeSpeed;
++			float meleeSpeed = 1f / player.meleeSpeed;
```
Solxanich: Do we need to worry about player.meleeSpeed ever being zero?
Similar to another review comment, why the change?

Solxanich (changes requested):
Overall a fairly well developed PR.
Just some minor syntax/cleanup items that I noticed.

Beyond that, I'd like to a change regarding how player.melee speed is handled in projectile.cs.patch - there's a possibility left open for a div0 error after the changes, so lets be smart and code defensively on that.

I haven't play tested this at all, so I'm taking you at your word on it, outside of gleaming that it looks reasonable in the diff.

patches/tModLoader/Terraria/ModLoader/DamageClassData.cs
ThomasThePencil: personally I never knew why it was internal, fine with makin' it public if desired

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
ThomasThePencil: done

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
ThomasThePencil: it is in SummonDamageClass though

patches/tModLoader/Terraria/ModLoader/DamageClassData.cs
steviegt6: Make everything public, this is a modding API.

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
Solxanich: Oh, my bad, missed the line skip bar on the diff.
Looks good, apologies for confusion.

(Please mark as resolved, I for some reason can't despite being the reviewer.)

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
Solxanich: (Resolved)

patches/tModLoader/Terraria/ModLoader/DamageClassData.cs
Solxanich: (resolved)

patches/tModLoader/Terraria/Player.cs.patch
Solxanich: (Resolved)

patches/tModLoader/Terraria/Player.cs.patch
Solxanich: (Resolved

patches/tModLoader/Terraria/Projectile.cs.patch
Solxanich: (Resolved)

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
ThomasThePencil: personally, I can't see a real reason as to why this would be better, as there's no way to modify existin' `DamageClass`es' return values on these methods (`GlobalDamageClass` doesn't exist --- and, if you ask me, it probably doesn't need to, anyway) and, the way I see it, the way I have it set up already is both intuitive and versatile. if it ain't broke, don't fix it

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
ThomasThePencil: technically addressed this on Discord already, but tl;dr this is laid out the way it is so that the modder who looks at it is gradually introduced to the information necessary to understand what's goin' on, rather than havin' everything thrown at them at once and bein' expected to just throw it into their own mod and experiment --- it's become increasingly clear that that won't work when some people aren't exactly bright enough to do it without first bein' told why copypastin' exactly what's there into their mod's codebase didn't work, so we're gonna try somethin' a little different given that you kinda need to know what's what with this feature to work with it to any extent past the simplest stuff

Solxanich: Sounds reasonable to me.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
Mirsario: Declarative designs for both `GetBenefitFrom` and `CountsAs` here would mean performance though. As well as less code.

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -121,6 +121,19 @@ public partial class Player
 		/// </summary>
 		public ref StatModifier GetKnockback<T>() where T : DamageClass => ref GetKnockback(ModContent.GetInstance<T>());
 
+		/// <summary>
+		/// Gets the armor penetration modifier for this damage type on this player.
+		/// This returns a reference, and as such, you can freely modify this method's return value with operators.
+		/// </summary>
+		public ref int GetArmorPen<T>() where T : DamageClass => ref GetArmorPen(ModContent.GetInstance<T>());
```
Mirsario: Rename to `GetArmorPenetration`, no need to shorten anything like a C developer.

ExampleMod/Content/Items/Accessories/ExampleStatBonusAccessory.cs
```diff
@@ -38,6 +40,15 @@ public class ExampleStatBonusAccessory : ModItem
...
 			// In this case, we're adding 100% knockback additively, but only for our custom example DamageClass (as such, only our example class weapons will receive this bonus).
 			player.GetKnockback<ExampleDamageClass>() += 1f;
+
+			// GetArmorPen is functionally identical to GetCritChance, but for the armor penetration stat instead.
+			// In this case, we'll add 5 armor penetration to magic weapons.
+			player.GetArmorPen(DamageClass.Magic) += 5;
+
+			// GetAttackSpeed works similarly to GetCritChance and GetArmorPen, but uses a float instead of an int.
+			// In this case, we'll make ranged weapons 15% faster to use.
+			// NOTE: DO NOT SET TO ZERO OR NEGATIVE VALUES. DOING SO WILL CAUSE PROBLEMS.
```
Mirsario: There should be checks & exceptions for that.

Mirsario (changes requested):
Pretty good. We'll need to test whether the `meleeSpeed` shenanigans have affected the item use rework's item time & animation related results.

ExampleMod/Content/DamageClasses/ExampleDamageClass.cs
ThomasThePencil: there is no conceivable way that this has enough of a performance impact to even remotely justify doin' it (even `GetBenefitFrom` isn't checked often enough per tick to present opportunity for noticeable improvement) and the amount of code necessary is already basically none. sorry, but I'm not convinced

Solxanich: Alright, so, I guess to rephrase this a different way, I'd like to see the CountsAs vritual method replaced by a CountsAs public field consisting of ``public List<DamageClass> CountsAs``. 
The corresponding call to CountsAs can than be replaced with ``this.CountsAs.Contains(damageClass)``

The advantages of this approach range from higher modifiability in terms of just adding or removing from the list (thus avoiding the need for a global), and performance gains in removing the performance overhead associated with virtual methods.

This is a fairly trivial change from a work involved perspective to push back on, is there a particular technical reason you're against changing it?

PR #1747, opened 2021-08-16
[1.4] Generic HasBuff

ExterminatorX99:
### What is the new feature?
Generic overload for `Player.HasBuff` and `NPC.HasBuff`

Reduces bloat caused by `HasBuff(ModContent.BuffType<ModBuff>())` calls

Can't think of any

`player.HasBuff<ExamplePetBuff>()`

### ExampleMod updates

Used the new overload in `ExampleLightPetProjectile`, `ExamplePetProjectile` and `ExampleSimpleMinion`


Steviegt6:
This is real nice, though you could just make an extension method. No reason for it not to be added unless someone wants to be a stickler for patches.

JavidPack:
I'm not sure this is a good idea, we want to teach people that the XType is used exactly the same as the XID class, and I feel that making separate approaches like this prevents new modders from understanding that concept. Plus, if we do this for all methods that take an ID, it's going to get ugly. Not doing it for all would be just as confusing.

Steviegt6:
> I'm not sure this is a good idea, we want to teach people that the XType is used exactly the same as the XID class, and I feel that making separate approaches like this prevents new modders from understanding that concept. Plus, if we do this for all methods that take an ID, it's going to get ugly. Not doing it for all would be just as confusing.

Extension methods it is, then.

Mirsario (approved):

direwolf420:
I'd keep the method but just not use it in EM, to appease jopos point
Similar thing exists with the Recipe.AddIngredient method

patches/tModLoader/Terraria/NPC.TML.cs
```diff
@@ -20,6 +20,10 @@ public partial class NPC : IEntityWithGlobals<GlobalNPC>
 		/// </summary>
 		public IBigProgressBar BossBar { get; set; }
 
+		public bool HasBuff(int type) => FindBuffIndex(type) != -1;
+
+		public bool HasBuff<T>() where T : ModBuff => HasBuff(ModContent.BuffType<T>());
+
```
Mirsario, line 26:
```suggestion
		/// <summary> Returns whether or not this NPC currently has a (de)buff of the provided type. </summary>
		public bool HasBuff(int type) => FindBuffIndex(type) != -1;

		/// <inheritdoc cref="HasBuff(int)" />
		public bool HasBuff<T>() where T : ModBuff
			=> HasBuff(ModContent.BuffType<T>());

```

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -50,6 +50,10 @@ public partial class Player
 		}
 		*/
 
+		public bool HasBuff(int type) => FindBuffIndex(type) != -1;
+
+		public bool HasBuff<T>() where T : ModBuff => HasBuff(ModContent.BuffType<T>());
+
```
Mirsario, line 56:
```suggestion
		/// <summary> Returns whether or not this Player currently has a (de)buff of the provided type. </summary>
		public bool HasBuff(int type) => FindBuffIndex(type) != -1;

		/// <inheritdoc cref="HasBuff(int)" />
		public bool HasBuff<T>() where T : ModBuff
			=> HasBuff(ModContent.BuffType<T>());

```

Mirsario (approved):

PR #1752, opened 2021-08-17
german localization fix

Indikativ:
fixed typos and removed spaces
because almost every word in german language is **one** word. for example: "Modliste" not "Mod Liste" (mod list)


stefnotch:
Native Austrian speaker here, I personally think "Mod Liste" is fine German. That's mainly due to me seeing "Mod" as a very English word and I'll treat it as such.

> Deine Mod

This one makes me wonder: What's the correct article for "Mod"? [The Duden](https://www.duden.de/rechtschreibung/Mod) may have an entry for it, but arguably it's not a very useful one, since that's a very different sort of "Mod".

Indikativ:
i think you are right

also: Mod -> Modification -> Modifikation (german) -> die Modifikation -> die Mod

Issue #1753, opened 2021-08-17
[1.4] Cutting in text boxes (Ctrl+X) sometimes crashes the game.


BlueStaggo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Sometimes, when you press Ctrl+X on a text box, tModLoader crashes. I keep accidentally crashing myself because I usually press Ctrl+X to clear the duplication menu search box.

### Log File

```
[11:18:52] [Main Thread/INFO] [tML]: Starting tModLoader client 1.4.2.3+0.12-1.4-Alpha-8305af16ddd7b2e2a959aa05831ab63bd3880daa-5249331498694505270
[11:18:52] [Main Thread/INFO] [tML]: Log date: 17/08/2021
[11:18:52] [Main Thread/INFO] [tML]: Running on Windows X64 NetCore 5.0.8
[11:18:52] [Main Thread/INFO] [tML]: Executable: C:\Program Files (x86)\Steam\steamapps\common\tModLoader\tModLoader.dll
[11:18:52] [Main Thread/INFO] [tML]: Working Directory: C:\Program Files (x86)\Steam\steamapps\common\tModLoader
[11:18:52] [Main Thread/INFO] [tML]: Developer mode enabled
[11:18:55] [Main Thread/DEBUG] [tML]: Native Resolve: FNA, Version=21.8.0.0, Culture=neutral, PublicKeyToken=null -> FNA3D
[11:18:55] [Main Thread/DEBUG] [tML]: 	attempting load C:\Program Files (x86)\Steam\steamapps\common\tModLoader\Libraries\Native\Windows\FNA3D.dll
[11:18:55] [Main Thread/DEBUG] [tML]: 	success
[11:18:56] [Main Thread/INFO] [tML]: Checking Steam installation...
[11:18:56] [Main Thread/DEBUG] [tML]: Native Resolve: Steamworks.NET, Version=15.0.1.0, Culture=neutral, PublicKeyToken=null -> steam_api64
[11:18:56] [Main Thread/DEBUG] [tML]: 	attempting load C:\Program Files (x86)\Steam\steamapps\common\tModLoader\Libraries\Native\Windows\steam_api64.dll
[11:18:56] [Main Thread/DEBUG] [tML]: 	success
[11:18:56] [Main Thread/INFO] [tML]: Steam installation OK.
[...]

PR #1757, opened 2021-08-17
[1.4] WIP: Do not free set data memory because ownership is reserved to the OLE …

DarioDaF:
### What is the bug?

Bug is related to crash on clipboard cut on windows as per issue: https://github.com/tModLoader/tModLoader/issues/1753

### How did you fix the bug?

Looking at documentation and other stack overflow it seems that:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclipboarddata
> If SetClipboardData succeeds, **the system owns the object** identified by the hMem parameter. The application may not write to or **free** the data once ownership has been transferred to the system, *[...]*

Removing the free from the code resulted in no more errors, but I didn't test in depth if this causes some memory leaks, a heads up from someone more familiar with win api can probably confirm or deny this more easily.

If the problem was the one supposed this is the easyest fix

### Anomalies found

*ReLogic/Windows/Clipboard.cs* presents:
```csharp
#if WINDOWS
    System.Windows.Forms.Clipboard.SetText(text);
#else
    NativeClipboard.SetText(text);
#endif
```
where I expected the WINDOWS path to be taken (where the managed set text will definitely never present the issue). I'm not sure if it is a solution setup problem or if it is intended to stay away from `WinForms` or other dependencies it might have, if it is a solution problem it is important to address that as fast as possible (could have various breaks here and there)


DarioDaF:
After a deeper research seems it has roots in `Terraria` solution rewriting, because `NativeClipboard` is fully inside that patch

PR #1758, opened 2021-08-18
parentVal translation fix

Ishigh1:
### What is the bug?
hjson translation with $parentVal doesn't work

### How did you fix the bug?
I added the replacement from ".$parentVal" to "" on the translation creation

The implementation could maybe be a bit cleaner but that's the simplest


jackbondpreston:
Oops, thanks @Ishigh1

Issue #1773, opened 2021-08-22
Use Item Fix Post-Merge TODOs & Issues.


Solxanich:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

No

### Description

The following are bugs reported that have been demonstrated to be due to the Use Item Fix PR #1437 and may require a fix:

- [x] #1721 
- [x] #1770 
- [x] #1767 
- [x] #1863 
- [x] #2478 
- [x] #3775 

The following are Post-Merge TODOs left by the PRs:
- [x] #2059 

### What does this proposal attempt to solve or improve?

Address & fix bugs found in deployed Alpha (or later) version of tModLoader 1.4, that have been investigated and shown to originate from the Use Item Fix PR.

PR #1774, opened 2021-08-23
[1.4] Simple itemAnimation mismatch patch and itemTime now passing trough zero

DarioDaF:
Demonstration patch of zero itemAnimation needed in AI_019_Spears for Kill

### What is the bug?
https://github.com/tModLoader/tModLoader/issues/1773

### How did you fix the bug?
Forcing item to act at the end of the frame (but cutting the method so that AI processing boundries are in the middle).

Anything else is probably better, a discussion should be done on the need to have consistent animation and time variables (if the issue was the animation an use time desync there are other ways around it keeping them desynced by a constant 1 frame instead of have them overlap).

This is probably the hackiest fix possible, just keeping vanilla order of times etc without the "spagetti" rework solves the issue too and why it was reworked that way (with actual mod examples or bugs that vanilla would make hard to solve).


DarioDaF:
tML current behaviour with `useStyle` of no autoswing sword: 
image

DarioDaF:
With this patch: 
image

DarioDaF:
Feral Claw swing with PR: 
image

As seen itemAnimation never gets down to 0, making the swing take 1 frame less.
While this patch and no `goto` (aka vanilla order) pass trough `itemAnimation == 0`

DarioDaF:
Lance behaviour, no 0 means no cleanup of projectiles: LanceTML

Guns somehow never get to 0 no matter what: GunTML

So if the idea is to keep al "resetting" on 1s (having last animation being 1 frame longer as an effect) all the places outside `ItemCheck` in vanilla code need to be fixed by subtracting one? so taking this as a base PR and removing the frame skip part (so it's like tML but with the reuse bugs, those reuse will be fixed by the subtract 1 on vanilla)

DarioDaF:
None of the above present desync (due to the fact that max animation is supposed to always be 1 less than use time in vanilla when they matter).

DarioDaF:
Final push of what I will be testing locally that will potentially become obsolete when Thomas finds an alternative solution (mine moves all `itemTime` to pass trough 0, autoReuse or not (without changing the actual frame counted useTime ofc).

Chicken-Bones:
Superceded by #2351

PR #1797, opened 2021-08-30
[1.4] World size logging, ExampleMod fixes, UnloadedTile mouseover, docs

direwolf420:
### What is the new feature?
First of all, sorry for the bloated PR.

1. Adds world size logging to the server on world load, and to the client entering world. In singleplayer, only the latter will show, so no duplicate info in that case
2. Removes the `MouseOverFar` from `UnloadedTile`, to make it like in 1.3
3. ExampleMod:
    * Makes every item journeymode-researchable
    * Fixes comments regarding `ItemDropRule.Common` (outdated sig and param names)
    * Fixes comments and usages of `ItemID.Sets.AnimatesAsSoul`
    * Add more usages of `Projectile.Opacity` instead of alpha
    * Add bonus comment for easy framing of a texture
4. Documentation for `Main.RegisterItemAnimation` and `ItemID.Sets.AnimatesAsSoul`

1. World size: Debugging, information for modders getting reports about features not working correctly which might be tied to world size.
2. MouseOverFar: Distracting and unnecessary in cases where you are surrounded with unloaded tiles. And most of the time nothing stops you to get close to a tile to inspect it. (Out of all tiles in vanilla, only things like graves or signs display text when mouseovering far away. I would not put unloaded tiles into that category.)
3. ExampleMod: Self-explanatory
4. Docs: The two methods are usually misunderstood in what they do, especially since the latter ID set was changed in it's purpose for 1.3, but kept its "old" name

### DIFF
--- a/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
+++ b/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
@@ -67,8 +67,8 @@ namespace ExampleMod.Content.Items.Consumables
 		public int exampleLifeFruits;
 
 		public override void ResetEffects() {
-			//Increasing health in the ResetEffects hook in particular is important so it shows up properly in the player select menu
-			//and so that life regeneration properly scales with the bonus health
+			// Increasing health in the ResetEffects hook in particular is important so it shows up properly in the player select menu
+			// and so that life regeneration properly scales with the bonus health
 			Player.statLifeMax2 += exampleLifeFruits * ExampleLifeFruit.LifePerFruit;
 		}
 
--- a/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
+++ b/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
@@ -20,6 +20,7 @@ namespace ExampleMod.Content.Items.Consumables
 
 		public override void SetStaticDefaults() {
 			Tooltip.SetDefault($"Permanently increases maximum life by {LifePerFruit}\nUp to {MaxExampleLifeFruits} can be used");
+
 			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 10;
 		}
 
@@ -48,6 +49,7 @@ namespace ExampleMod.Content.Items.Consumables
 			// This handles the 2 achievements related to using any life increasing item or getting to exactly 500 hp and 200 mp.
 			// Ignored since our item is only useable after this achievement is reached
 			// AchievementsHelper.HandleSpecialEvent(player, 2);
+			// TODO re-add this when ModAchievement is merged?
 			return true;
 		}
 
@@ -65,6 +67,8 @@ namespace ExampleMod.Content.Items.Consumables
 		public int exampleLifeFruits;
 
 		public override void ResetEffects() {
+			//Increasing health in the ResetEffects hook in particular is important so it shows up properly in the player select menu
+			//and so that life regeneration properly scales with the bonus health
 			Player.statLifeMax2 += exampleLifeFruits * ExampleLifeFruit.LifePerFruit;
 		}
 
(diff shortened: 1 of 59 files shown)


ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
```diff
@@ -65,6 +67,8 @@ public class ExampleLifeFruitPlayer : ModPlayer
 		public int exampleLifeFruits;
 
 		public override void ResetEffects() {
+			//Increasing health in the ResetEffects hook in particular is important so it shows up properly in the player select menu
+			//and so that life regeneration properly scales with the bonus health
```
Mirsario, line 71:
```suggestion
			// Increasing health in the ResetEffects hook in particular is important so it shows up properly in the player select menu
			// and so that life regeneration properly scales with the bonus health
```

ExampleMod/Content/Items/ExampleSoul.cs
```diff
@@ -16,8 +16,8 @@ public class ExampleSoul : ModItem
 			// Registers a vertical animation with 4 frames and each one will last 5 ticks (1/12 second)
 			// Reminder, (4, 6) is an example of an item that draws a new frame every 6 ticks
 			Main.RegisterItemAnimation(Item.type, new DrawAnimationVertical(5, 4));
-			
-			ItemID.Sets.AnimatesAsSoul[Item.type] = true; // Makes the item have a 4-frame animation while in world (not held.)
+			ItemID.Sets.AnimatesAsSoul[Item.type] = true; // Makes the item have an animation while in world (not held.). Use in combination with RegisterItemAnimation
+
```
Mirsario, line 20: This set is legacy as hell. The old comment is true - the animation is hardcoded to 4 frames.

ExampleMod/Content/Items/ExampleTooltipsItem.cs
```diff
@@ -13,8 +14,12 @@ public class ExampleTooltipsItem : ModItem
 			// See here for help on using Tags: http://terraria.gamepedia.com/Chat#Tags
 			Tooltip.SetDefault("How are you feeling today?"
 				+ $"\n[c/FF0000:Colors ][c/00FF00:are ][c/0000FF:fun ]and so are items: [i:{Item.type}][i:{ModContent.ItemType<ExampleMountItem>()}][i/s123:{ItemID.Ectoplasm}]");
+			
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 1;
 
 			Main.RegisterItemAnimation(Item.type, new DrawAnimationVertical(30, 4));
+			ItemID.Sets.AnimatesAsSoul[Item.type] = true; // Makes the item have an animation while in world (not held.). Use in combination with RegisterItemAnimation
+
```
Mirsario, line 22: You have 4 frames here too, which is why you thought that this set isn't hardcoded.

ExampleMod/Content/Projectiles/ExampleAdvancedAnimatedProjectile.cs
```diff
@@ -108,15 +109,19 @@ public class ExampleAdvancedAnimatedProjectile : ModProjectile
...
+			// If texture without animation frameHeight is always texture.Height and startY is always 0
 			int frameHeight = texture.Height / Main.projFrames[Projectile.type];
 			int startY = frameHeight * Projectile.frame;
 			
 			// Get this frame on texture
 			Rectangle sourceRectangle = new Rectangle(0, startY, texture.Width, frameHeight);
+
+			//Alternatively, you can skip defining frameHeight and startY and use this:
+			//Rectangle sourceRectangle = texture.Frame(1, Main.projFrames[Projectile.type], frameY: Projectile.frame);
+
```
Mirsario, line 121:
```suggestion

			// Alternatively, you can skip defining frameHeight and startY and use this:
			// Rectangle sourceRectangle = texture.Frame(1, Main.projFrames[Projectile.type], frameY: Projectile.frame);

```

Mirsario (changes requested):

Mirsario (approved):

PR #1803, opened 2021-09-01
[1.4] Add ObsidianLockBox to OpenVanillaBag methods

direwolf420:
### What is the bug?
The opening of the "ObsidianLockBox" item, which is new in 1.4, was not concidered by any of these hooks that are usually used in that context: `ConsumeItem`, `PreOpenVanillaBag`, and `OpenVanillaBag`.
I also adjusted the comments/docs to the updated code (all but 1 method was renamed to uppercase, and also a property not related to the PR was referenced with the wrong casing in a comment)

### How did you fix the bug?
Add the method calls in the proper contexts

You can make an argument for renaming the "lockBox" parameter to "goldenLockBox" as it would avoid ambiguity. (But the ItemID is still just called `LockBox`)


Steviegt6:
While we're at it, would it be worthwhile to switch this over to an enum?

PR #1807, opened 2021-09-01
Add Error Handling 'no callback received' Failure Mode

Solxanich:
### What is the bug?
There is currently a gap where if no callback is received (which shouldn't happen as either Game Server API or SteamAPI will have initialized if its reaching such a code point, and thus callbacks must be delivered), but nevertheless, doesn't hurt to make sure the failure mode is caught instead of an infinite loop.

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -3,6 +3,7 @@ using Steamworks;
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -486,14 +491,18 @@ internal class AQueryInstance
 
 					_queryHook.Set(call);
 
+					int counter = 0;
 					do {
-						// Do Pretty Stuff if want here
+						if (++counter > 2000) // 10 seconds maximum alotted time before assumed no connection
```
Mirsario, line 496:
```suggestion
						if (++counter > 2000) // 10 seconds maximum allotted time before assumed no connection
```

Issue #1809, opened 2021-09-02
[1.4] Fatal Unloading due to presence of IL.Method


Solxanich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Bug summary: 
Fatal Unloading due to presence of IL.Method hooks on Release Builds of tModLoader.

This is almost certainly a bug within TerrariaHooks / MonoMod. 

### Log File

This also holds true for any Mod using "IL.Method += Method" to modify hooks. Can be easily tested with this skimmed version of the ExampleMod code.
[ToBeDeleted.zip](https://github.com/tModLoader/tModLoader/files/7095332/ToBeDeleted.zip)

### Steps to reproduce

 - On Release build of tML, reproduction steps are "IF ExampleMod's enabled status does not change over the first two reload events, it will crash on the second reload while unloading monomod hooks" which is fairly easy to do if you build+reload, than just reload it.
Debug Build is uneffected.

### Expected Behavior

Unloads.

### Actual Behavior

Crashes.

### Reproduction frequency

100% for specifically the reproduction steps provided.

### Additional Information

Finally, the following sequence is noteworthy: "Start w/ no mods -> Enable GenericUnaffectedMod and reload -> Enable AffectedMod and reload -> Disable GenericUnaffectedMod and reload -> Infinite reload events unlocked"

Solxanich:
Fixed in #1857

PR #1810, opened 2021-09-02
Add Error Logging To Wrap Fatal Unloading of Hooks on Release tML

Solxanich:
Why should this be merged?
- This moves away from a fatal unloading due to random looking error (https://github.com/tModLoader/tModLoader/issues/1809) to instead log an error on mod loading on release versions of tML. This is more practical in terms of isolating the issue to a known location and making clear what is the breaking code included in a mod.

Why isn't this a bug fix?
- No fix could be found for the bug. Its very nature appears too transient and random for my sanity


Solxanich:
PullRequest outdated,

PR #1816, opened 2021-09-04
WIP: Refresh config ui after save/revert config

zzp198:
WIP: Refresh ui after save/revert config.

At present, the button still exists after clicking the button, no good interactive feedback.

Now will be refresh feedback after clicking the save or revert button.


zzp198:
It has been mentioned in the previous PR

Issue #1721, opened 2021-08-10
[1.4] Melee lingering hitbox damage


direwolf420:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Melee hitbox is "active" for one tick on item use on the old location, causing damage to one NPC in it when you use a melee item again.
(Item location not updating if items are not in use is vanilla behavior and normally has no effect on the game)

[Demonstration video shown using reused code on item glowmasks that draws at all times](https://cdn.discordapp.com/attachments/711551818194485259/874659270757613689/lingeringMeleeHitbox.webm)

### Steps to reproduce

Can be tested on a dummy, hit it once with a melee item, walk back, swing the weapon, and even though it is out of range, the dummy receives damage. repeated swings behave as expected (no damage)

### Reproduction frequency

50% reproduction value on each try.

### Additional Information

I tested the latest build and a build the commit prior to [this one](https://github.com/tModLoader/tModLoader/commit/2174a830efad027c35812e55c0e55023a18cf2ac), same thing.

Rosethorns:
Repeated swings DO NOT behave as expected - what has happened is that it has now adopted the last location of the previous swing so if you swing twice in the same spot then there is a hitbox at that position AS WELL as the proper hitbox.

DarioDaF:
lingering hitbox seems to not overlap with previous 100%, it seems to be very near to player hitbox, if I hit overlapping the dummy with the player I always get the hit when away

direwolf420:
> 
> 
> Repeated swings DO NOT behave as expected - what has happened is that it has now adopted the last location of the previous swing so if you swing twice in the same spot then there is a hitbox at that position AS WELL as the proper hitbox.

I cannot reproduce this. I tried both autoswinging while moving in/out of the dummy, and also just clicking once, move away, then autoswing. Can you elaborate on how you got this to happen?

Chicken-Bones:
Fixed in #2351

PR #1821, opened 2021-09-05
Minimal fix for item holding hitbox order

DarioDaF:
### What is the bug?

https://github.com/tModLoader/tModLoader/issues/1721

### How did you fix the bug?

Just reordered hitbox specifiying code between after use code and before hit code in `ItemCheck`

Technically https://github.com/tModLoader/tModLoader/pull/1774, but this will change some `itemTime` and `itemAnimation` behaviour (from a mod dev prespective) so in Discord was decided to search for other possible solutions to see the options.

So this is a small patch to only a part of the problem that is reasonable and doesn't change as much as the other pull request.


Chicken-Bones:
Superceded by #2351

PR #1822, opened 2021-09-05
1.4 NeedSaving removal

Ishigh1:
### What is the removed feature?
On GlobalItem, there's a method, NeedSaving and following a discussion on Discord, it was said that it should be removed

Essentially, you can easily achieve the same thing by making saving null do the same, it removes the useless NeedSaving => true as they can simply do this check at the start of Save() and return null if it doesn't need saving.

### ExampleMod updates
Surprisingly, there's no GlobalItem Save/Load on ExampleMod, so nothing to update

Note : I didn't test this code with item saving yet


Mirsario:
Merged with #1814

Issue #1671, opened 2021-08-02
[1.4] Materials Tab in Journey Mode menu not being filled automatically


Mayonnaisinator:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

In the Journey Mode item duplication menu, the "Materials" tab is not being automatically populated like the other categories. This is likely because no items have the "Materials" tooltip added while they are in the Journey menu (and Cheat Sheet's menu, for that matter). This tooltip is only added once the item is out of the menu, and in the inventory, which is inconsistent with vanilla, where the tooltip is already present on all items in the research menu.
![Empty Materials](https://user-images.githubusercontent.com/81387852/127836047-557a1dd1-da47-4107-8b9a-8c99204e8b33.png)
![Vanilla Materials](https://user-images.githubusercontent.com/81387852/127836053-2ce21b39-c504-46d9-bf14-17f42ab5ed0a.png)
![tML Materials](https://user-images.githubusercontent.com/81387852/127836058-8b810ced-33a8-4eb8-9e35-a45e1916ad7e.png)

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/6915766/client.log)

### Steps to reproduce

1. Research any item that is a material.
2. See it populate the other relevant categories as expected.
3. The Materials category is still empty.

### Expected Behavior

Any item that is a material (and therefore has the "Material" tooltip) should appear in the "Materials" tab once researched, as in vanilla.

### Actual Behavior

No items are appearing in the Materials tab.

### Reproduction frequency

100% reproduction, using both GamerMod's /research command, and manually researching items.

MunchyG:
i have noticed this too, honestly not that worried about the tab, not working, more annoying that while browsing items you can't see if they a material.

Mirsario:
Should be fixed with the PR above merged.

PR #1829, opened 2021-09-10
[1.4] Fix empty Materials tab in Journey Mode

direwolf420:
### What is the bug?
The Materials tab in the Journey Mode Item menu is empty.

### How did you fix the bug?
Add `ContentSamples.FixItemsAfterRecipesAreAdded` after the tml call to `Recipe.SetupRecipes` to mirror the behavior found in `Main.Initialize`.

No


direwolf420:
Fixes #1671

PR #1841, opened 2021-09-14
Make InstallNetFramework.bat standalone again

Solxanich:
Also add a -sysdotnet argument so that could feasibly launch via the scripts in to local system dotnet again


Solxanich:
Gonna do this in a separate PR with other minor stuff

PR #1844, opened 2021-09-15
[1.4] Make 'ProcessGroupsForText' and 'AcceptedByItemGroups' public

ExterminatorX99:
### What is the bug?
Recipe methods 'ProcessGroupsForText' and 'AcceptedByItemGroups' were (mistakenly?) made internal

### How did you fix the bug?
Made them public again


Steviegt6:
Pretty sure these were intentionally made internal; there may be some other way to achieve the same effect? Either way, I see no point in them being internal. It doesn't stop people from executing the code, it just makes it slower to do so.
+1

PR #1847, opened 2021-09-16
Fix mono launch script on non-standard LD_LIBRARY_PATH

oatmealine:
### What is the bug?

On non-standard `LD_LIBRARY_PATH` environment values (such as NixOS) the provided tModLoader mono launch script will not locate essential libraries (PulseAudio, xinput, etc.) and will be unusable.

### How did you fix the bug?

Instead of forcefully setting `LD_LIBRARY_PATH` to `lib:lib64` (which would load ./lib/, /lib/, ./lib64/ and /lib64/) it's instead added to the path with `$LD_LIBRARY_PATH:lib:lib64`

Not that I know of. This is how environment values that contain paths are usually added to in Unix shell scripts, and the current version only works by accident (I wasn't even aware you could load /lib with just `lib`).

This may also need similar fixes on MacOS, but I do not have a Mac PC so I cannot personally test any changes.

### DIFF
--- a/solutions/ReleaseExtras/tModLoader-mono
+++ b/solutions/ReleaseExtras/tModLoader-mono
@@ -18,7 +18,7 @@ if [ "$UNAME" == "Darwin" ]; then
 		export DYLD_INSERT_LIBRARIES="$STEAM_DYLD_INSERT_LIBRARIES"
 	fi
 else
-	export LD_LIBRARY_PATH=lib:lib64
+	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:lib:lib64"
 fi
 
 # move all kickstart mono libraries to a sys/ folder to remove conflicts


Mirsario:
Yeah, unix support is pretty barebones in 1.3.
It should be checked whether the linker prefers system libraries in this & previous contexts. If the answer is the same to both of them, then this is safe to merge. It's preferable for it to be a no though. Although I also see that MacOS uses the same kind of export (so it's a no to your last statement), but mac package manager users are relatively rare.

oatmealine:
I'm pretty sure the original would first check for `./lib` and `./lib64` and then move on to `/lib` and `/lib64`, but I'll have to test which order it uses whenever multiple paths are supplied to `LD_LIBRARY_PATH`.

oatmealine:
`LD_CONFIG_PATH="lib:lib64"; ldconfig -p | grep libpng`,
`LD_CONFIG_PATH="$LD_CONFIG_PATH:lib:lib64"; ldconfig -p | grep libpng` and `LD_CONFIG_PATH="lib:lib64:$LD_CONFIG_PATH"; ldconfig -p | grep libpng`
all output the same on my system because from what I can tell from reading up on ldconfig online, it prefers `/usr/lib` anyways as that's hardcoded directly in the source code. So it'd still prefer system libraries before this patch on standard `LD_CONFIG_PATH` systems.
I've yet to test this on NixOS and similar, however, so the results may differ there. All I know is setting `LD_CONFIG_PATH` to just `lib:lib64` will make it pretty badly fail on those systems, and changing the launch script to the proposed version fixes it.

Mirsario:
Merged, but I'm not sure if there will be any more 1.3 updates.

PR #1850, opened 2021-09-18
is there going to be linux support? because i like switched to linux because its generally faster in terms of both gameplay and dashboard

z5vix:
is there going to be full linux support


JavidPack:
It works fine on Linux. Also, you should not use a pull request for this. If you have a support question, your best bet is to come to discord and chat with us there.

PR #1860, opened 2021-09-21
[1.4] Town NPC StepUp fix + ExampleTable

direwolf420:
### What is the bug?
(This is also a problem in 1.3 btw)
Modded work benches, anvils, and similar furniture are not being treated the same as vanilla counterparts in regards to town NPC collision. The NPC usually walks "by" the furniture, but modded furniture is not concidered for that and it will perform "step up", effectively walking "on" the furniture.

### How did you fix the bug?
The hardcoded check for such furniture is replaced via a new `TileID.Sets.IgnoredByStepUp` set used specifically for that purpose.

The name and description of the set could be made better. The naming is similar to the existing `IgnoredByGrowingSaplings`. I kept it as precise as possible without referencing any town NPC in its name or description code (`NPCAIStyleID.Passive` for example), since the set is only checked in a non-related method (`Collision.StepUp`), which just gets a special parameter passed in WHEN the AI is being executed. Modders can also call that method themselves without the town NPC.

### ExampleMod
Adjusted `ExampleWorkbench` to use this new set, and fix it's name to be in line with vanilla.
Added `ExampleTable` (so that this special set is also used, cause it might get ignored when someone else would implement ExampleTable later).


direwolf420:
Should I add a disclaimer that only "furniture with a walkable surface usable for NPC housing" should be put in the `IgnoredByStepUp`  set? Stuff like anvil, work bench, table, but not stuff like furnace, magic ball, loom

Steviegt6:
> Should I add a disclaimer that only "furniture with a walkable surface usable for NPC housing" should be put in the `IgnoredByStepUp` set, and would it be as comments in EM or as docs in tml? Stuff like anvil, work bench, table, but not stuff like furnace, magic ball, loom

Yeah, seems important to make note of.

PR #1876, opened 2021-09-29
Allow non public constructor on ILoadable implementations

ExterminatorX99:
### What is the bug?
ILoadable implementations with a non public parameterless constructor are ignored

### How did you fix the bug?
Allowed non public constructors

No


Chicken-Bones:
Do you not want this anymore?

ExterminatorX99:
Allowing it can't hurt, but I don't need it anymore

PR #1878, opened 2021-10-01
[1.4] Forgotten breaking change in tagCompound

Ishigh1:
The Set method got an overload in 1.3, but the old method wasn't removed to prevent breaking mods, but in 1.4 we can do this change so the sooner this method is removed, the less mods it will break (if it breaks any)


Ishigh1:
PR merged with #1877

PR #1879, opened 2021-10-01
[1.4] IReadOnlyTagCompound - interface

ExterminatorX99:
### What is the new feature?
IReadOnlyTagCompound interface

### Why should this be part of tModLoader?
Allows explicitly defining that a TagCompound should only be read from.

### Are there alternative designs?
readonly ref struct => https://github.com/tModLoader/tModLoader/pull/1884

### Sample usage for the new feature
LoadData hooks

### ExampleMod updates
Fixes

PR #1884, opened 2021-10-02
[1.4] ReadOnlyTagCompound - readonly ref struct

ExterminatorX99:
### What is the new feature?
ReadOnlyTagCompound readonly ref struct

Allows explicitly defining that a TagCompound should only be read from.

interface => https://github.com/tModLoader/tModLoader/pull/1879

LoadData hooks

### ExampleMod updates
Fixes

### DIFF
--- a/ExampleMod/Common/Systems/DownedBossSystem.cs
+++ b/ExampleMod/Common/Systems/DownedBossSystem.cs
@@ -39,7 +39,7 @@ namespace ExampleMod.Common.Systems
 			// }
 		}
 
-		public override void LoadWorldData(TagCompound tag) {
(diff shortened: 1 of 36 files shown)


ExterminatorX99:
Alternative: https://github.com/tModLoader/tModLoader/pull/1879

Chicken-Bones:
I think this is much better than #1879 and I'm not opposed to it. But you should probably make methods which can return `TagCompound` return readonly as well.

Personally I don't think any of this is necessary. If people get their load save hooks backwards it's pretty easy to spot and figure out.

ExterminatorX99:
I've gotta agree. Doesn't seem necessary

PR #1897, opened 2021-10-08
[1.4] Fixed FNA discrepancy between Terraria and tModLoader

DarioDaF:
Terraria solution was using FNA.dll bundled in patch and not the latest one used also by tModLoader solution, since Terraria is where FNA port lives it makes sense to have it updated too (also `Game` constructor was missing due to difference in version)


Chicken-Bones (changes requested):
Okay I think I got it

PR #1901, opened 2021-10-09
[1.4] Fixed Terraria solution

DarioDaF:
As the title tells, moved all "new FNA" related changes to TModLoader solution instead of Terraria to keep stock FNA in the base build


DarioDaF:
Decided to keep our FNA in Terraria and asserting loading on main thread instead

PR #1904, opened 2021-10-11
[1.4] [DRAFT] XML comments cleanup and addressing warnings in general (in a very lax way) [fixed]

DarioDaF:
Doesn't really fall into a bugfix or new feature category, just some cleanup following reasonable VS warnings avoiding big changes (i.e. no addressing deprecated uses of functions like in Lang).

This is the fixed tree version of #1898

Other changes in progress:
- [ ] Fixing silent override or hiding of methods
- [ ] Removing or fixing non matching `<param />` elements (adding missing ones seems like a thing to do later, muffling the warnings with a dumb no-op description for argument feels not productive nor useful)

### DIFF
--- a/patches/tModLoader/Terraria/ID/BuffID.cs.patch
+++ b/patches/tModLoader/Terraria/ID/BuffID.cs.patch
@@ -1,14 +1,11 @@
 --- src/Terraria/Terraria/ID/BuffID.cs
 +++ src/tModLoader/Terraria/ID/BuffID.cs
-@@ -1,5 +_,8 @@
+@@ -1,4 +_,5 @@
  using ReLogic.Reflection;
- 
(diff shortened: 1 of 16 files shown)


steviegt6 (approved):
Looks good, thanks.
We should probably consider redoing a lot of documentation since a lot of it needs prettying up, but that's for a different PR, I suppose.

JavidPack:
Seems like a lot of these have been fixed throughout recent changes. I'm going to close this since I think most are already done and the merge conflicts would be very difficult to manage. If you notice any of these still in the most recent version of the source, let me know on discord and I can fix them directly, or you can make another PR. Sorry about this, sometimes PRs don't get reviewed quickly enough and they get outdated.

PR #1922, opened 2021-10-26
[1.4] Nuke the World

ExterminatorX99:
Nuke the useless World class


Chicken-Bones:
The patches here are pretty light on and wouldn't be hard to maintain, but I don't think there's much use case here.

JavidPack:
I also agree that we probably don't need to do this at this time.

Mirsario:
Yorai has now removed the class.

PR #1923, opened 2021-10-26
[1.4] ExampleMod Recipes rework

ExterminatorX99:
Remove the static ExampleRecipes class
Add the Recipes class that inherits from GlobalItem
Fix an example
Use RecipeGroupID instead of string group names. String version was left in for completion

No


Chicken-Bones (approved):
Having to make a `GlobalItem` and use `AppliesToEntity` is lame.

But this looks like the way to do it at the moment, and it's an improvement over what was previously there.

Mirsario (approved):

PR #1926, opened 2021-10-29
[1.4] `Player.QuickSpawnItem` return item index

ExterminatorX99:
### What is the new feature?
`Player.QuickSpawnItem` now returns the index of the spawned item
Added Direct variants to `Player.QuickSpawnItem`

Allows modifying the data of the spawned item

Add more arguments to `Player.QuickSpawnItem`

```cs
Item item = player.QuickSpawnItemDirect(ItemID.Zenith); // or `Main.item[Player.QuickSpawnItem(ItemID.Zenith)]`
item.prefix = PrefixID.Legendary;
```

### ExampleMod updates
None

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -739,7 +739,7 @@
  			if (Main.rand.Next(15) == 0 && Main.hardMode) {
  				int number = Item.NewItem((int)position.X, (int)position.Y, width, height, 602);
  				if (Main.netMode == 1)
-@@ -3942,15 +_,34 @@
+@@ -3942,15 +_,33 @@
  						break;
  					}
  			}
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -739,40 +739,21 @@
...
-+ 		}
  
++		/*
  		public void QuickSpawnItem(int item, int stack = 1) {
  			int number = Item.NewItem((int)position.X, (int)position.Y, width, height, item, stack, noBroadcast: false, -1);
-@@ -3951,6 +_,21 @@
+ 			if (Main.netMode == 1)
+ 				NetMessage.SendData(21, -1, -1, null, number, 1f);
  		}
++		*/
```
Chicken-Bones, line 756: Why comment this rather than editing it directly?

Chicken-Bones:
Looks almost perfect

PR #1927, opened 2021-10-30
[1.4] Patch and general cleanup

ExterminatorX99:
Move some methods into partials
Fix XML comment in `Item.TML.cs`
Instead of deleting a bunch of lines in `Main`, comment them
Remove unused field and methods from `Netplay`
Slightly cleanup `Recipe` patch

Update `TerrariaHooks.dll`


Mirsario (approved):
Good, but it may be needed to reimplement #340's patches later.

ExterminatorX99:
Looks like it was fixed in 1.4 vanilla by introducing `Main._worldPreparationState`

PR #1929, opened 2021-10-31
[1.4] Fix mannequin accessory IOOB exception

ExterminatorX99:
### What is the bug?
When hovering an accessory over a mannequin's accessory slot, an IOOB exception occurs

### How did you fix the bug?
Ensured that loops only ever index to the last element of the array

Can't see any simple ones

### DIFF
--- a/patches/tModLoader/Terraria/UI/ItemSlot.TML.cs
+++ b/patches/tModLoader/Terraria/UI/ItemSlot.TML.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using Terraria.Audio;
 using Terraria.ModLoader;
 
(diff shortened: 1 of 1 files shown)


Mirsario:
@Solxanich has raised multiple problems about this PR. The fact that accessories don't handle mannequin cases is actually a design issue that will need more than just this.

Solxanich:
Thanks Mirsario for reminder.

To summarize my main point from Discord, this is treating a symptom of the problem, not fixing the actual problem: tml accessory hooks/modslot are currently fundamentally designed incompatibily with Mannequins.

The first question to be asked here is whether or not mannequins should have mod slots.
If yes, a fair bit of work beyond just this is needed. This likely won't work from a UI perspective and could get quite crowded.
If no, the answer of which I'm partial to, than we need to isolate mannequin from tml so that any accessory can go in to the given 2 slots. This is arguably cleaner, as it requires less changes and doesn't involve using additional UI space.

Progressing with the No path, I than anticipate the following minimum high level work needed for this:
1) remove AccCheck from being called for Display Doll
2) remove ItemLoader.CanEquipAccessory hooks from being invoked on Display Doll

PR #1934, opened 2021-11-04
your discord doesn't work

idkman2021:
your discord doesn't work
and neither does the google docs file
add the issues tab


idkman2021:
dude oh my god I hate github

PR #1944, opened 2021-11-09
Example CI building

Solxanich:
This PR adds the ability for mods to be published using github workflows, among other CI options.
This feature is required for collaboration on Mods.

This includes supplying both a sample batch file, and sample workflow.yaml file. 

The usage of SteamCMD requires special attention, and would recommend staged User Acceptance Testing.

Ideally, once this is implemented, this allows us to move towards taking advantage of github actions to dispatch calls to subscribed modders, and subsequently either have their mods build immediately for the new version & publish, or just let them know in general.


Solxanich:
Closing in favour of code needed in #2133

Issue #1943, opened 2021-11-09
Deleting a mod from the workshop twice causes a crash


Truhard:
### Is there an existing issue for this?

- I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

If you download a mod from the browser, then go back to the browser, reselect the mod, click delete on the downloaded mod from the browser, then select it again and click delete again, the game crashes. 

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/7500948/client.log)

### Steps to reproduce

1. Open workshop
2. Install any mod
3. Delete it
4. Reselect it
5. Delete it again
6. Profit :(

### Additional Information

AMD RX 5700 graphics card

Solxanich:
@Truhard - this issue should be resolved.
Please close if it is the case

PR #1945, opened 2021-11-09
Add clearing the Installed localmod to delete routine

Solxanich:
Bug fix for #1943 

This isn't a perfect fix, although it eliminates the crash scenario, it still leaves a visual error on the mod browser where the mod is both installed and not installed on the main UI, but correctly shown in the ModInfo UI.

Reloading the browser obviously fixes this now graphical error.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
@@ -240,6 +240,12 @@ namespace Terraria.ModLoader.UI.ModBrowser
 			return UpdateNeeded = true;
 		}
 
+		internal void ModifyUIModDownloadItemInstalled(string modName, Core.LocalMod installed) {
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
```diff
@@ -240,6 +240,12 @@ internal partial class UIModBrowser : UIState, IHaveBackButtonCommand
 			return UpdateNeeded = true;
 		}
 
+		internal void ModifyUIModDownloadItemInstalled(string modName, Core.LocalMod installed) {
+			var modbrowsermod = _items.IndexOf(_items.FirstOrDefault(x => x.ModDownload.ModName.Equals(modName, StringComparison.OrdinalIgnoreCase)));
+			if (modbrowsermod >= 0)
```
Chicken-Bones, line 245:
```suggestion
			var moddownload = _items.Select(x => x.ModDownload).FirstOrDefault(x => x.ModName.Equals(modName, StringComparison.OrdinalIgnoreCase));
			if (moddownload != null)
				moddownload.Installed = installed;
```

PR #1951, opened 2021-11-10
Fix Conflict between Settings button and Defense Icon

Solxanich:
image

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2741,15 +2741,15 @@
  				}
  
  				inventoryBack = color;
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2974,6 +2974,15 @@
  			if (mouseItem.stack <= 0)
  				mouseItem.type = 0;
  
+@@ -33523,7 +_,7 @@
+ 					num4 += num5;
+ 				}
+ 
+-				if (amountOfExtraAccessorySlotsToShow >= 1 && (screenHeight < num3 || (screenHeight < num4 && mapStyle == 1))) {
```
Solxanich, line 2982: Add a comment on why this was added

PR #1960, opened 2021-11-13
Swap to Release Build

Solxanich:
At some point, we will have to make the judgement call to swap over to using the Release Build instead of the Debug build.

This should be done during a period where there is ample support presence, and minimal other ongoing changes in case something goes wrong.

This PR reminds us to do this in the coming months.


JavidPack:
Handled directly

PR #1961, opened 2021-11-14
[1.4] Fix constantly requesting csproj update

ExterminatorX99:
### What is the bug?
The Mod Sources UI is constantly requesting the csproj to be updated

### How did you fix the bug?
Fix the upgrade check

No


Mirsario:
> Fixed a recently introduced typo

One does not just go out of their way to add a typo to an area that wasn't rewritten from scratch. It should remain to be a backslash.

> That's the standard for VS & MSBuild. They're smart enough to make it work on unix (Confirmed.)
But `/`, on the other hand, breaks on some VS versions.

Issue #1940, opened 2021-11-07
[1.4]hidden by taskbar[windows11]


moe41:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows11

### Platform

Steam

### Description

when display mode:1366×768 100%(from windows settings>system>display),bottom of window is hidden by taskbar.
(resolution:1366×720)i can't push maximize button.
![無題](https://user-images.githubusercontent.com/66813644/140630917-965a4f0a-6016-49d0-b472-b31bf504dec7.png)
comparison:from vanilla terraria,resolution:1366×696(the same display mode and miximized window)
![無題2](https://user-images.githubusercontent.com/66813644/140631022-c5c16f6a-f662-4f1e-b096-a33fe603e750.png)

moe41:
when open inventory, "settings" is hidden by taskbar.
![無題](https://user-images.githubusercontent.com/66813644/140648475-2c493707-a957-440e-9b37-82301c545efe.png)

moe41:
![無題](https://user-images.githubusercontent.com/66813644/142625985-e8cba1e8-be88-48f5-b254-8d56c1efbea9.png)
latest version, social media buttons are hidden by taskbar.

inikishev:
temporarily you can edit game config to make UI scale smaller

moe41:
in config.json(from "C:\Users\(username)\Documents\My Games\Terraria\ModLoader\Beta\"),i tried to edit
from "DisplayHeight": 720, to "DisplayHeight": 696
but returned "DisplayHeight": 720

inikishev:
> in config.json(from "C:\Users(username)\Documents\My Games\Terraria\ModLoader\Beta"),i tried to edit from "DisplayHeight": 720, to "DisplayHeight": 696 but returned "DisplayHeight": 720

edit the UI scale setting instead, idk what the name is

moe41:
UI scale? where?
manual window sizing?

inikishev:
> UI scale? where? manual window sizing?

ni its called ui scale. Search for uiscale or something like that

moe41:
![無題](https://user-images.githubusercontent.com/66813644/154858960-0774ed35-f0bf-4df4-aed3-183d7dbc7497.png)

moe41:
[...]

PR #1972, opened 2021-11-18
Fix window sizing

### LINKED ISSUE #1889: [1.4] Windows 'fullscreen' resolution not compatible with the "Go Windowed" button

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Currently, when one goes to "fullscreen" via the button in settings, this leads to the window being resized as windowed borderless-  ie it matches screen size.

However, the reverse direction of "Go Windowed" doesn't resize the window resolution back down, so one ends up with a window with no controls, despite having 'gone windowed'.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/7280547/client.log)

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

Solxanich:
### What is the bug?
Fix #1889  & a support bug found where using fullscreen resolution leads to display settings getting reset on tmodloader updates.

### How did you fix the bug?
Wrapped display changes related to fullscreen with a boolean resolution correction to allow for windowed borders.
Added a set fullscreen display settings to occur before tmodloader writes settings (thus fixing the new version load order problem)

### DIFF
--- a/patches/tModLoader/Terraria/Main.TML.cs
+++ b/patches/tModLoader/Terraria/Main.TML.cs
@@ -37,8 +37,10 @@ namespace Terraria
 		public static Color MouseTextColorReal => new Color(mouseTextColor / 255f, mouseTextColor / 255f, mouseTextColor / 255f, mouseTextColor / 255f);
 		public static bool PlayerLoaded => CurrentFrameFlags.ActivePlayersCount > 0;
 
+		// Used to prevent situations where when going from borderless Fullscreen display to windowed display the width and height don't get resized to be able to access key window functions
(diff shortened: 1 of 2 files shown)


moe41:
but #1940 isn't fixed yet?

Solxanich:
Hi Moe,
Unfortunately, this PR didn't have the intention to address that issue.

Support for windows 11 has not been addressed, and is pending someone who has it taking time to fix it. I myself do not so I can't target a fix to it

PR #1977, opened 2021-11-19
[1.4] Beard accessory support, ExampleBeard

direwolf420:
### What is the new feature?
Makes beards an EquipType to support alongside 1.4.3 added beards.

### ExampleMod updates
Added ExampleBeard which uses the new code, including comments for making a pre-1.4.3 beard [(Ginger Beard)](https://terraria.fandom.com/wiki/Ginger_Beard).


patches/tModLoader/Terraria/ModLoader/EquipType.cs
```diff
@@ -18,6 +18,7 @@ public enum EquipType
 		Shield,
 		Neck,
 		Face,
+		Beard,
 		Balloon
```
Mirsario: So that people don't need to rebuild their balloon enthusiastic mods
```suggestion
		Balloon,
 		Beard,
```

Mirsario (approved):

PR #1978, opened 2021-11-20
Better portable storage support

pbone64:
### What is the new feature?
Added better portable storage support, by introducing a tracker (similar to the void lens and piggy banks) for Safes and Defenders Forges, adding a set that lets you count a projectile as a portable storage projectile, and added the `BankID` class to easily reference special bank ids.

1.4.3 greatly improved the portable storage system, but it still wasn't at the point where mods could easily add their own without IL editing. This saves mods from multiple annoying IL edits.

Yes, probably. The way sounds are handled could be changed to deterministic, however I can't think of a good way to do that right now

Here is an example of a portable safe projectile:
```
public class Awa : ModProjectile
{
    {
        ProjectileID.Sets.CountAsBank[Projectile.type] = BankID.Safe;
    }

    public bool init = false;

    public override void AI()
    {
        if (!init)
        {
            Player localPlayer = Main.player[Projectile.owner];

            localPlayer.chest = BankID.Safe;
            for (int i = 0; i < 40; i++)
            {
                ItemSlot.SetGlow(i, -1f, chest: true);
            }

            localPlayer.safeProjTracker.Set(Projectile);
            localPlayer.chestX = (int)Projectile.Center.X;
            localPlayer.chestY = (int)Projectile.Center.Y;
            localPlayer.SetTalkNPC(-1);
            Main.SetNPCShopIndex(0);
            Main.playerInventory = true;
            SoundEngine.PlaySound(SoundID.MenuOpen);
            Recipe.FindRecipes();

            init = true;
        }
    }

    public override void PlayBankCloseSound()
    {
        SoundEngine.PlaySound(SoundID.MenuClose);
    }
}
```
### ExampleMod updates
I did not update ExampleMod.

Side note: For some reason, an old commit was included in this. The old commit doesn't actually change anything, and I'm not sure why GitHub included it in this PR.


direwolf420:
Add documentation to ClearPortableBankProjectileTrackers, then it's mergeable imo!

PR #1982, opened 2021-11-21
[1.4] Allow tModLoader to count for Terraria playtime

### LINKED ISSUE #1540: Achievements and Play-time
### Description
Please add the ability to merge tmodloader and terraria like before, So that we can earn Achievements and Play-time (towards terraria). This was possible in pervious versions... With the steam release that all changed for no real reason. I asked on the discord and it seems they aren't going to bring that back. Don't really understand why the ability to manual install is removed, The "Manual install guide" is really just a fix for the gog version and family sharing players, You still of to go thru steam.

A QoL really. You get your Mods and you get your Achievements and Play-time.

A launcher just like before that runs tml thru the terraria.exe file. (I think that's how it worked)

### LINKED ISSUE #1979: 1.4 Alpha check Terraria version
### Version

Latest public-1.4-alpha

### OS

Linux

### Platform

Steam

### Description

Currently the 1.4 alpha (and maybe the stable version) does not check for the Terraria version, and instead claims that game assets are missing. Attached is my crashlog for today, containing the crashes for Glow 310 missing, and after a few tries I made a fake Glow 310 file and it complained that Extra 244 was missing, I kept creating files and it kept discovering new missing ones until I updated to the latest vanilla terraria on steam which contained the missing game assets.
[client-crashlog.txt](https://github.com/tModLoader/tModLoader/files/7574337/client-crashlog.txt)

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/7574344/client.log)
Here is the client log after I created around 15 fake buff files and then realized I was probably on the wrong version of the game.

### Steps to reproduce

Have old version of terraria. 
Install tmodloader.
Get in-descriptive error message.

### Expected Behavior

It should warn the user about the wrong version.

### Actual Behavior

It only notices missing asset files.

### Reproduction frequency

Any time when versions don't match.

### Additional Information

My temporary solution for users on reddit: [https://old.reddit.com/r/Terraria/comments/qxnlqw/tmodloader_14_issues/hlddgxo/](url)

Solxanich:
### What is the new feature?
I had originally planned to prototype this as a mod; however, it proved infeasible to do so due to launch timing limitations.
Specifically, the Terraria shim needs to be launched before tModLoader, and can't be launched from tModLoader itself.

At any rate, this new feature makes Playtime count towards Terraria. It does so by using a Terraria shim program to fake running Terraria.

Things left to do on this PR:

- [ ] Make a mod for granting achievements to Vanilla
- [x] Add the communication between the shim and tModLoader
- [x] Closes #1979 

Resolves #1540  with respect to playtime


Mirsario:
Discussed in Discord; Closed the aforementioned issue with the following comment:
> Adding or acknowledging support for this is anti-productive, as using a separate `appid` would mean not being able to play via steam networking with TML users who are on default configurations. And, now, there's an even bigger issue in 1.4 - making this work with steam workshop is an ever so giant can of worms.

The seemingly experimental solution for play-time tracking in this PR also seems like way too many shenanigans for something that most people will not want.

PR #1988, opened 2021-11-24
Begin Migration from 105600 Cloud

### LINKED ISSUE #1993: Corrupt character selection
### Version

Latest Official

### OS

Windows

### Platform

Steam

### Description

Character selection screen is not acting as expected. tModLoader keeps displaying unnamed, corrupt characters which are apparently saved to the cloud. If I turn off cloud storage for these characters and re-enter the character selection there are always at least two new cloud stored, unnamed characters. I have tested this on both macOS and Windows 10.

I managed to make a working, local character on macOS but managed to delete it by trying to delete one of the corrupt characters mentioned above. On Window 10, when trying to hit play on a newly created character it simply freezes and I have to force quit.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/7597980/client.log)

### Steps to reproduce

1. Start tModLoader
2. Select "Single Player"
3. There are two unnamed characters, with the default looks, 100 health and 0 mana, with cloud saving on
4. Set cloud saving off, go back and select single player again. There are now four characters, the previous two, and always two new with cloud saving on.
5. Sometimes one of the characters display a long string of special characters and symbols.
6. Behaviour when trying to delete some of these characters is weird, and could delete other characters.

### Expected Behavior

There should be no corrupted characters and I should be able to create a new character and a new world and store them both online on steam and play normally.

### Actual Behavior

The character selection screen is corrupt.

### Reproduction frequency

Each time I run tModLoader

### Additional Information

I have tried on macOS, loading mods manually as the mod browser does not work on macOS, and got to play successfully. However I deleted my character trying to delete a corrupted character that was created automatically by tModLoader.

I have also tried in Windows 10 via boot camp on an intel MacBook pro, on a fresh steam install, and managed to reproduce the same issues regarding the corrupted characters, but not being able to play at all with a local character.

Solxanich:
This PR moves us away from 105600 cloud file usage.
A similar PR will be needed to modify 1.3.

Resolves #1993, #1996 by removing cloud files off of Steam.

The core outcomes of this PR are:
1) Migrate all files off of 105600 cloud and into the SaveFile directory
2) Prevent players from moving files on to the cloud during the transition period.

The following is pertinent information to announce to Steam Users:
1) That we are in the process of changing Steam Cloud 'providers' and will be disabling adding new files to the Cloud in the short term of the next two weeks
2) That players should either run tModLoader to migrate their files off of the old Steam Cloud provider and in to their Save File directory, or download the files manually via https://store.steampowered.com/account/remotestorageapp/?appid=105600
3) This change applies to both 1.3 release and 1.4 Alpha, and will be completed by Dec. 15th, so we ask that all players take the time to migrate during this window. After this window, files will only be available to download manually via the aforementioned link.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/Elements/UICharacterListItem.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/Elements/UICharacterListItem.cs.patch
@@ -144,14 +144,15 @@
  		private void DeleteMouseOut(UIMouseEvent evt, UIElement listeningElement) {
  			_deleteButtonLabel.SetText("");
  		}
-@@ -178,8 +_,12 @@
+@@ -178,8 +_,13 @@
  		private void CloudButtonClick(UIMouseEvent evt, UIElement listeningElement) {
  			if (_data.IsCloudSave)
  				_data.MoveToLocal();
 -			else
 +			else {
 +				ModLoader.Engine.Steam.RecalculateAvailableSteamCloudStorage(); //Only recalculate when about to put the file to cloud
-+				if (!ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient(_fileSize))
++				// Disable Putting on 105600 Cloud Temporarily Nov 23 2021 - Solxan
++				if (true || !ModLoader.Engine.Steam.CheckSteamCloudStorageSufficient(_fileSize))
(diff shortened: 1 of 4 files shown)


Solxanich:
Superseded by:
https://steamcommunity.com/groups/SteamClientBeta/announcements/detail/3114798907236840432

Issue #1693, opened 2021-08-04
[1.4] Music 'stutters' at certain points


Mayonnaisinator:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

The music sometimes 'stutters' and will repeat a note. Happens at about 1 second in the video below.

https://user-images.githubusercontent.com/81387852/128204548-827ccffe-4444-49d5-a1de-d1857c27023c.mp4

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/6932240/client.log)

### Steps to reproduce

1. Wait for music to stutter
2. Repeat step 1

### Expected Behavior

Music shouldn't stutter

### Actual Behavior

Music sometimes stutters

### Reproduction frequency

It doesn't seem to happen too often, maybe only once per loop of the song.

NicolasPL64:
+1

Alaric323:
Can confirm, almost seems to happen in the same spots every time.

Soulfurious:
I've also wondered whether or not this was a phenomenon unique to TML or just the Overhaul version I was using.

WildArticunoAppeared:
Yep. Same here. And I only have cheating/debugging mods.

WildArticunoAppeared:
I'm gonna listen to all tracks and then list which ones stutter and maybe even at which points.

Mirsario:
Should be fixed now.

fihn14:
this is still an issue for me

Chicken-Bones:
As a workaround, please try replacing `Libraries\Native\Windows\FAudio.dll` in your tModLoader steam dir with https://github.com/tModLoader/tModLoader/blob/927e175d6492eb428e8514249416f1c3c106782f/patches/Terraria/Terraria/Libraries/Native/Windows/FAudio.dll?raw=true

Solxanich:
Possibly Related: #2323 , #2466

Xaymar:
> As a workaround, please try replacing Libraries\Native\Windows\FAudio.dll in your tModLoader steam dir ...

This appears to have fixed it, but I'm not entirely sure. There is still a noticable stutter when the music is supposed to loop, but it is much less noticable compared to before.

ChrisBuilds:
Still experiencing this bug with 1.4.3.6

Music repeats a note every 20-30 seconds.

PR #2000, opened 2021-11-28
use native Win32 platform version of FAudio instead of ffmpeg

ubnt-intrepid:
### What is the bug?

On Windows, BGM is intermittently interrupted at short intervals (refs: #1693, [forum thread]).

[forum thread]: https://forums.terraria.org/index.php?threads/im-new-to-troubleshooting-terraria-mod-issues.98816/

As mentioned in the thread above, the following log messages were dumped on interrupting the BGM:
```
[wmav2 @ 0000025d2446f840] overflow (1865 > 1864) in spectral RLE, ignoring
```

These log messages were due to ffmpeg, and a similar problem was reported on their issue tracker:
* https://trac.ffmpeg.org/ticket/691

I also experimented by replacing ffmpeg DLLs in `Libraries/Natives` with the latest 4.4.1 build, but it didn't solve the slutter problem.

### How did you fix the bug?

This patch forces FAudio backend to use the native Win32 audio platform instead of ffmpeg on Windows. 

* Upgrade `FAudio.dll` to the upstream version with `PLATFORM_WIN32=ON`
* Remove unnecessary DLLs from ffmpeg

<details>
<summary>Build information</summary>

Environment:
  * Windows11, MinGW-w64 on MSYS2
  * `MINGW64_NT-10.0-22000 DESKTOP-XXXXXX 3.2.0-340.x86_64 2021-08-02 16:30 UTC x86_64 Msys`

Instructions:
```shell-session
$ echo "$(pwd), $(git rev-parse --short HEAD)"
/d/Sources/FAudio, de0c1f8

$ mkdir _build && cd $_
$ cmake -DPLATFORM_WIN32=ON ../
$ ninja
$ ls
... FAudio.dll ...
```

</details>

* Build FAudio during setup rather than including pre-build DLLs.


Chicken-Bones:
Interesting. When did FAudio add the platform win32 option? Does it work on all versions of windows?

ubnt-intrepid:
> When did FAudio add the platform win32 option?

Since [21.07](https://github.com/FNA-XNA/FAudio/releases/tag/21.07). A concern is that they mention this feature as "experimental", but it seems to work fine at least in my environment. 

> Does it work on all versions of windows?

The native Win32 platform backend [uses Windows Core Audio API](https://docs.microsoft.com/en-us/windows/win32/coreaudio/core-audio-apis-in-windows-vista), that introduced in Vista and improved in Windows7, and it should be work on most of upstream Windows versions. I'm sorry but I don't currently own the older versions of Windows and it has not be tested yet.

Solxanich:
Quick question here;
Is the bug confirmed only observed on Windows?
Just want to identify if a similar change needs to be considered for Unix platforms.
Beyond that, has this been tested with mods that add Music, such as ExampleMod's example biome? Want to make sure we aren't breaking anything mods wise

Chicken-Bones:
The mention of wmav2 codec means it should only affect tracks from the vanilla wavebank

Mirsario (approved):
Seems to be working well, thanks. GStreamer being removed is really great news -- It was such a pain to deal with that I legit rolled back FAudio to using ffmpeg directly.

PR #2005, opened 2021-11-28
[1.4] (2) Fixed FNA discrepancy between Terraria and tModLoader

DarioDaF:
Replaces PR https://github.com/tModLoader/tModLoader/pull/1897 after the recent changes.
Main difference since last fix is the omission of all the logic to neatly organize in subfolders and string manipulation madness done in tML for `.dll`s, figured it wouldn't be a big loss and makes more sense to touch only tML project files if something must change there in the future


patches/tModLoader/Terraria/MonoLaunch.cs.patch
```diff
@@ -1,53 +1,31 @@
...
  		// FNA is requested by both Terraria and ReLogic.dll
- 		var loaded = new Dictionary<string, Assembly>();
- 
+@@ -24,6 +_,7 @@
  		AppDomain.CurrentDomain.AssemblyResolve += delegate (object sender, ResolveEventArgs sargs) {
  			string resourceName = new AssemblyName(sargs.Name).Name + ".dll";
  			string text = Array.Find(typeof(Program).Assembly.GetManifestResourceNames(), (string element) => element.EndsWith(resourceName));
 +			
  			if (text == null)
  				return null;
```
Chicken-Bones: Whitespace diffs

patches/Terraria/Terraria/Terraria.csproj.patch
```diff
@@ -74,13 +68,13 @@
...
-+    <None Update="Libraries/Native/**" CopyToOutputDirectory="PreserveNewest" />
++    <Content Include="Libraries/Native/**" CopyToOutputDirectory="PreserveNewest" />
 +  </ItemGroup>
 +  <Target Name="CopyToSteamDir" AfterTargets="Build">
 +    <ItemGroup>
 +      <BinFiles Include="$(TargetDir)**" />
 +    </ItemGroup>
 +    <Message Text="Copying $(AssemblyName) to Steam Dir..." Importance="high" />
-+    <Copy SourceFiles="@(BinFiles)" DestinationFolder="$(TerrariaSteamPath)/%(RecursiveDir)" />
++    <Copy SourceFiles="@(BinFiles)" DestinationFolder="$(tModLoaderSteamPath)/%(RecursiveDir)" SkipUnchangedFiles="True" />
```
Chicken-Bones: Intentional change here? We previously built the Terraria sln into the Terraria steam dir. I don't think Terraria.csproj has the required asset path redirects for the side-by-side content sharing etc.

patches/Terraria/Terraria/Terraria.csproj.patch
```diff
@@ -74,13 +68,13 @@
...
 +    <None Remove="Libraries/XNA/**" />
 +    <None Remove="Libraries/net40/**" />
-+    <None Update="Libraries/Native/**" CopyToOutputDirectory="PreserveNewest" />
++    <Content Include="Libraries/Native/**" CopyToOutputDirectory="PreserveNewest" />
 +  </ItemGroup>
 +  <Target Name="CopyToSteamDir" AfterTargets="Build">
 +    <ItemGroup>
 +      <BinFiles Include="$(TargetDir)**" />
 +    </ItemGroup>
 +    <Message Text="Copying $(AssemblyName) to Steam Dir..." Importance="high" />
```
Chicken-Bones: Maybe while you're at it, you can put the destination folder in this message

Chicken-Bones (commented):
Looks mostly good, I'll have to check against my local install that none of the tML files have changed.

patches/Terraria/Terraria/Terraria.csproj.patch
DarioDaF: Fair point I didn't wanna copy the full block there, I should probably just point and do all the stuff tML does, because now a build that is not `Debug` will be called `Terraria.exe` overwriting the game, also a bunch of modified dlls might be used by vanilla terraria due to name clashing (mainly ReLogic).

DarioDaF:
Ok, moved output directory for Terraria back to the actual Steam folder where `Content` is, moved the code to clean the output to the Terraria solution so it's not a bunch of DLLs potentially conflicting with Vanilla.
Just for cleanliness I added a property with `_` in the solution that is the actual output folder, so both tML and Terraria can use it in the convoluted Targets without diffs

Issue #1718, opened 2021-08-09
[1.4 Windows] StackOverflow when breaking 128+ gravity affected blocks in a vertical stack


Solxanich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

When breaking 128+ falling block stack at the bottom, it triggers stack overflow.

On Windows, the COMPlus_DefaultStackSize enviro variable doesn't appear to work to give a quick solution to increase the stack size. The default secondary thread stack size on Windows is 1.5MB (https://github.com/dotnet/runtime/issues/33622#issuecomment-599462300)

This is only really triggerable in the desert, as no other location will typically have that large a vertical count of blocks.

### Log File

Replication world:
[The_Clear_Tilt_of_Argon.zip](https://github.com/tModLoader/tModLoader/files/6955797/The_Clear_Tilt_of_Argon.zip)

Log:
[client.log](https://github.com/tModLoader/tModLoader/files/6955798/client.log)

Debugger output:
[New Text Document.txt](https://github.com/tModLoader/tModLoader/files/6955801/New.Text.Document.txt)

### Steps to reproduce

1. Load the replication world
2. Walk left to sand pillar
3. Break the bottom sand block in any column. 
4. Crash

Breaking the second block from the bottom (ie the 127th) is ok.

### Reproduction frequency

100%

Solxanich:
Fixed in #1857

Solxanich:
Re-opening; tested on a release build instead of debug build.
Still bugged on debug.

Solxanich:
Patched in #2054

PR #2008, opened 2021-11-29
[1.4] Adding default stack runtime config 4MB

DarioDaF:
Hopefully fixing https://github.com/tModLoader/tModLoader/issues/1718.
If this works on all plaforms we can consider removing from the scripts the enviroment variable setting for Unix and Mac, but having it twice doesn't hurt until confirmed working.


Solxanich:
Please complete the following additional changes related to deploying this change:
1) Update Program.cs to have the following code block: 
``LaunchMain(isServer);
		}

		//IMPORTANT: We really want arg parsing, logging, and monomod initialized before static Main()
		// Moving all calls to Main to LaunchGame_() mitigates this risk
		//IMPORTANT: Main thread stack size does NOT respect DefaultStackSize runtime config. 
		// Running LaunchGame_ via a new thread addresses this to provide the buffer vanilla recursion needs
		// NOTE: Static Main is currently still on the Main thread with this design. Parameters weren't working.
		private static void LaunchMain(bool isServer) {
			Main.dedServ = isServer;
			new Thread(LaunchGame_).Start();
		}

		public static void LaunchGame_() {``
		
2) Remove the stack debugging code in Program.tml.cs
3) Remove the Unix environment variable export in InstallNetFramework.sh
4) Cleanup any thread naming to be appropriate

And that should wrap this up. I'd recommend testing in the example world provided in the issue ref'd.

If you know how to do Thread.Start with arguments, feel free to simplify 1)'s code

Chicken-Bones:
From Discord:
Revert the thread name change, leave the new thread as "Main Thread"
Call the original launch thread "Entry Thread"
Maybe set the entry thread as a background thread?

DarioDaF:
Set also the thread as background (after spawning the other one).
Tested with the sand pillar and no crash as expected

PR #2017, opened 2021-12-03
Use Correct Reference Assemblies in CSharpCompiler

Solxanich:
The existing code for getting framework references is only valid for net framework.
We have to change what we do for .netcore (and thus .net6)

The issue is that the old code utilizes the .... I'm tired. Just read this https://github.com/jaredpar/basic-reference-assemblies

This swaps to use the net6 package from the library that one person made to fix the bug.
Reference assemblies aren't default provided by the SDK anywho so modders would have headache, so might as well go for the lib at that point and just eat the 5 MB heft.


Mirsario (approved):
I'd delete the old code rather than comment it out.

Chicken-Bones:
Learn something new every day. Looks great!

PR #2027, opened 2021-12-06
[1.4] Remove trailing whitespace

ExterminatorX99:
Trailing whitespace is evil.
I'm good, so I killed it.


patches/Terraria/Terraria/Main.cs.patch
```diff
@@ -481,22 +481,22 @@
...
-@@ -49869,6 +_,8 @@
- 			else if (Terraria.Graphics.Effects.Filters.Scene["MoonLordShake"].IsActive()) {
+@@ -49870,6 +_,8 @@
  				Terraria.Graphics.Effects.Filters.Scene.Deactivate("MoonLordShake");
  			}
-+			
-+			SkipMoonLordEffects:
  
++			SkipMoonLordEffects:
++
```
steviegt6, line 496: Your modification to this patch here removed the newline above this label but added a newline below it, unsure of if that was intentional.
It also caused the differ to add an extra line of patch context.

steviegt6 (changes requested):
A small concern.

patches/Terraria/Terraria/Main.cs.patch
ExterminatorX99: I don't see a problem.
Both patches result in
```cs
// Empty line
SkipMoonLordEffects:
// Empty line
```

ExterminatorX99: And the context line was just changed. There is one less at the top, but one extra at the bottom

PR #2044, opened 2021-12-15
[1.4] Simplify ExampleCommandSign to ExampleSign

xNarnia:
This commit reduces ExampleCommandSign to ExampleSign. This removes the CommandExecutor from the Sign example to capture the functionality of it better. 
I'll be making a new ExampleCommandTile later that shows off the Command executor functionality.

The example has a lot of documentation built-in with comments. There is no additional documentation needed.


Solxanich:
Thank you for the PR.

I notice that the files modified are in the Old folder. Does this PR include porting it in to 1.4? 
If so, I would suggest making a copy of the file and putting the copy in to the proper ExampleMod locations, not the Old folder.

Than when so ever the ExampleCommandTile gets done, the original file in the Old folder can be re-utilized to create that.

xNarnia:
Heyo! This is a 1.3 example that has not been tested on 1.4.
I wanted to simplify the example from ExampleCommandSign to ExampleSign to make it easier to search for and find at a glance (but did it very late).

Solxanich:
Thanks for the clarification. 
The 1.3 branch is where the 1.3 examples live, but isn't going to see any more updates.

Given the current state of this Github, it would be most appropriate to port the changes to the 1.4 examplemod as that is the current default branch and where changes may actually get merged. Would it be possible for you to update the example to 1.4 proper? I'd be happy to review and merge soon after

xNarnia:
> 

Will do!

PR #2047, opened 2021-12-18
Data-oriented tiles

Mirsario:
## Description
This PR changes the way the game stores tiles to be more data-oriented, turning `Tile` into what is essentially an on-stack key to any data associated with its position. In usage, the type remains about the same as if it was still a class with data.

In short, the design provided here achieves the following:
- Data is no longer stored by references, cutting memory usage by 12 (24 on x64) bytes per tile (242MB/484MB) in a large world). But we'll likely expand some bit packs into full values to improve speed.
- Mods can add their own tile data.
- Large world generation performance is actually improved by ~30% at average?
- Data is now stored on type-basis, allowing usage of direct memory operations to, say, set all of the world's tiles to 0, or copy tons of it in one swoop around IO.

The implementation affects minimal amount of patches, best we can do really!

## TODOs
- [x] ~~The implementation of tile data definition is very thoughtless and should be redone.~~
  * Done. Tile data organised to:
    * Maximize performance
    * Make space for 64 liquid ids (used spare bits from rearranging the vanilla header fields)
    * Logically group the bitfields to make implementing schematics easier.
- [x] ~~What's the purpose behind multiple simultaneous Tilemaps? Who can create one? When?~~
  * Denied. Only 1 tilemap. A reference schematic implementation would be useful in tML, but that can be another PR. Schematics can use plain arrays of tiledata, but IO needs much thought.
- [x] Do we want to use chunks for mod data? They wouldn't be hard to do. 
  * Postpone for another PR, easy to do with a different generic accessor on Tile.


Chicken-Bones:
Confirmed, a regular large world with all tiles loaded in tML uses 806.5MB (564.5MB in vanilla 32bit). This PR uses 363.1MB

PR #2053, opened 2021-12-21
[1.4] Workshop menu tweaks

kittenchilly:
Just tweaks the workshop menu to put more focus on the mod menu and remove the TBD button that nothing will likely ever come of.

image

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/States/UIWorkshopHub.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/UIWorkshopHub.TML.cs
@@ -39,7 +39,7 @@ namespace Terraria.GameContent.UI.States
 		}
 
 		private UIElement MakeButton_OpenModBrowserMenu() {
-			UIElement uIElement = MakeFancyButton("Images/UI/Workshop/HubWorlds", "tModLoader.MenuDownloadMods");
+			UIElement uIElement = MakeFancyButton("Images/UI/Workshop/HubPublishWorlds", "tModLoader.MenuDownloadMods");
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/GameContent/UI/States/UIWorkshopHub.cs.patch
```diff
@@ -55,33 +55,28 @@
...
+@@ -98,6 +_,25 @@
  			uIElement3.HAlign = 1f;
  			uIElement3.VAlign = 1f;
  			uIElement2.Append(uIElement3);
 +			*/
 +
 +			// Added by TML [
 +			var modsMenu = MakeButton_OpenModsMenu();
-+			modsMenu.HAlign = 0f;
++			modsMenu.Width.Set(-10f, 1f);
```
JavidPack, line 66: Try this with `modsMenu.Width.Set(0f, 1f);`, it might fix the weird spaces on left and right

patches/tModLoader/Terraria/GameContent/UI/States/UIWorkshopHub.TML.cs
```diff
@@ -50,16 +50,5 @@ partial class UIWorkshopHub
 			Interface.modBrowser.PreviousUIState = this;
 			Main.MenuUI.SetState(Interface.modBrowser);
 		}
-
-		private UIElement MakeButton_TBD() {
```
Solxanich, line 54: Please comment out this code instead of removing it - we may reuse this later if we find an agreed on use case for the fourth button.

Issue #1745, opened 2021-08-15
[1.4] Multiplayer characters both "mine" when only one clicks 


MunchyG:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

so basically if two (or more idk) people in multiplayer are near each other and one [player 1] of them attempts to mine (or any action, placing included) if the other player's [player 2] cursor is in the range of "player 1" then it will double peform the action, so say "player 1" is trying to mine with a pickaxe, if "player 2"'s cursor is in range it will mine that block too, even if "player 2" doesn't have a pickaxe at all. 

### Log File

[client2.log](https://github.com/tModLoader/tModLoader/files/6988175/client2.log)
[server.log](https://github.com/tModLoader/tModLoader/files/6988177/server.log)
[client.log](https://github.com/tModLoader/tModLoader/files/6988179/client.log)

### Steps to reproduce

have 2 or more players.
have the players close and all looking at different blocks
have one player attempt to mine a block

### Expected Behavior

think it is pretty clear

"player 1" shoudn't also "mine" the blocks under "player 2"'s cursor 

### Actual Behavior

"player 1" also "mines" the blocks under "player 2"'s cursor, as well as the one under his cursor

### Reproduction frequency

so far 100%. every time we have left clicked and the cursors overlapped in "build/mine" range it has double performed the action

### Additional Information

this was originally noticed in journey mode (cos long reach) with mods, and i just thought it was a mod causing the issue but decided to test it without any mods loaded and the same issue arose. This may just be a journey mode thing though cos all my testing was also in journey mode. (have attached both of my clients and the server log hosting while i was testing)

DBotThePony:
[Video demonstrating the issue](https://icf.dbotthepony.ru/Videos2/2021-10-02_14-32-54.mp4)

also on video you can hear soundtrack is distorted

MunchyG:
any updates on how patching this bug is going

Solxanich:
@MunchyG  the issue should now be suitably patched.

If you can confirm, then please close this.
If we don't hear anything, we will assume it is resolved and this can be closed

MunchyG:
Should get around to testing it later today, I will let you know how it goes.

MunchyG:
From some very quick testing I would say it is good (however it is local host testing, but given the code changes I don't think it should make a difference)

PR #2059, opened 2021-12-28
Remedy for duplicate item use on remote clients. Fixes #1817, #1745

### LINKED ISSUE #1817: [1.4] Block placement/removal issue in multiplayer

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

When trying to place/remove blocks in multiplayer on the Tmod 1.4 alpha, if a player (A) is standing close enough to another player's (B) cursor to be able to place/remove a block at B's cursor location, they will place/remove the block at both A's and B's cursor location, instead of just A's

### Log File

_No response_

### Steps to reproduce

requires 2 players (A and B)
1. B places their cursor over the tile where they wish to place/remove a block
2. A moves to a location close enough to place/remove a block at B's cursor location
3. A moves their cursor to a location other than B's cursor location where they wish to place/remove a block
4. A attempts to place/remove a block
5. the block will be placed/removed at both A and B's cursor locations instead of just at A's cursor location

### Expected Behavior

The block should only be placed at A's cursor location

### Actual Behavior

The block is placed at both A and B's cursor locations

### Reproduction frequency

I was able to reproduce the issue fairly easily, it is easier to reproduce if A has a larger block placement/removal range

### Additional Information

_No response_

Chicken-Bones:
## This is a REMEDY and not the correct long term fix

Introduced by #1437

In Vanilla, when a remote player uses an item, none of the logic for whether the item use is possible, or using the item runs.
tModLoader attempts to make `ApplyItemTime` called on remote players. It does so by disabling the vanilla remote player check, allowing the code to run the 'can use item' test, applying item time, and then returns before the item is actually used.

Unfortunately, this was only correctly implemented (and presumambly tested) on `ItemCheck_TurretAltFeatureUse`, `ItemCheck_MinionAltFeatureUse` and `ItemCheck_Shoot`. Meaning that all other item uses ran on remote clients as well, often using substituted incorrect information from the remote client (such as cursor position).

This currently affects wiring tools, lawn mower, instruments, buckets, mining tools, rod of discord, life/mana crystal/fruit, demon heart, torch god's favor, event items, boss spawners, combat book, pet licenses, golf ball state, all placement functions, critter releasing and the bone glove.

### How did you fix the bug?
I inserted the remote player check to skip the above listed functions (and applying item time for them).

Apply the splitting of can use item, apply item time and use item correctly into all vanilla code (many places, hard).
Also note that in many places, item usability depends on non synchronised state, such as cursor position, and therefore can't be determined correctly on remote players. 

A potential solution would be actually syncing the `ApplyItemTime` call itself, in addition to the current approach of syncing the item use attempt.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3676,7 +3676,7 @@
 -				if (whoAmI == Main.myPlayer) {
 +
 +				// TML attempts to make ApplyItemTime calls run on remote players, so this check is removed. #ItemTimeOnAllClients
-+				if (whoAmI == Main.myPlayer || true) {
++				if (true) {// previously: if (whoAmI == Main.myPlayer)
  					bool flag4 = true;
  					int type2 = item.type;
  					if ((type2 == 65 || type2 == 676 || type2 == 723 || type2 == 724 || type2 == 757 || type2 == 674 || type2 == 675 || type2 == 989 || type2 == 1226 || type2 == 1227) && itemAnimation != itemAnimationMax - 1)
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3659,7 +3659,7 @@
  			if (itemTime > 0) {
  				itemTime--;
  				if (ItemTimeIsZero && whoAmI == Main.myPlayer) {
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
It may also be necessary to note in the docs that `UseItem` runs on all clients, as this is changed behaviour to 1.3

PR #2062, opened 2021-12-28
Use aiStyle -1 by default for modded NPCs

Chicken-Bones:
### What is the bug?
`NPC.aiStyle` 0 is undocumented in the id list, and has functionality making it an unreasonable default for modded NPCs.
Many modders use aiStyle -1 to get no vanilla ai, but it's not immediately clear that negative values are permitted.

### How did you fix the bug?
Default modded npcs to aiStyle -1

Add a new aiStyle for modded NPCs called 'DoNothing'
Add a new aiStyle for the bound vanilla NPCs


JavidPack:
Seems fine to me. I'm just not sure if modders want this to be the default

PR #2069, opened 2022-01-02
Nuget support for mods

Chik3r:
### What is the new feature?
Support for nuget packages/dependencies in TML mods.
It works by making TML compile mods by loading and compiling the `.csproj` file and then copying the required `.dll`s to `lib/`, instead of creating a new `CSharpCompilation` by manually loading the source files and references in `dllReferences`/`weakReferences`/`modReferences`

Because nuget

Maybe

Just include a nuget package in your `.csproj`, for example: `<PackageReference Include="RestSharp" Version="106.15.0" />`, and you are ready to go.
[Here](https://git.ppeb.me/Chik3r/TestNugetMod) is an example of a mod using nuget and mod references


patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
```diff
@@ -439,29 +475,38 @@ private string DllRefPath(BuildingMod mod, string dllName)
...
 
-			var parseOptions = new CSharpParseOptions(LanguageVersion.Preview, preprocessorSymbols: preprocessorSymbols);
+			if (project.CompilationOptions is not null) {
+				var newOptions = project.CompilationOptions
+					.WithAssemblyIdentityComparer(DesktopAssemblyIdentityComparer.Default)
+					.WithOptimizationLevel(preprocessorSymbols.Contains("DEBUG") ? OptimizationLevel.Debug : OptimizationLevel.Release);
 
-			var emitOptions = new EmitOptions(debugInformationFormat: DebugInformationFormat.PortablePdb);
+				if (project.CompilationOptions is CSharpCompilationOptions csharpComp)
+					newOptions = csharpComp.WithAllowUnsafe(allowUnsafe);
```
ExterminatorX99, line 491: Is overwriting the above assigned `newOptions` intentional?

Chik3r: Yes, it's intentional. The `newOptions` is of type `CompilationOptions`, which doesn't have an `allowUnsafe` field, so in case `newOptions` is of type `CSharpCompilationOptions`, we cast it and set the `allowUnsafe` field to whatever was passed to TML. I could make it so it always assumes it is going to compile a C# project, but it would throw an error if it tried to compile a VB.NET or F# project, and I'd prefer to leave it like this just in case someone wanted to write a mod using another .NET language.

ExterminatorX99: `CompilationOptions` is immutable, which means calling `CompilationOptions.WithX` doesn't modify the `CompilationOptions`, but instead returns a brand new instance with that option applied.
Meaning `newOptions = csharpComp.WithAllowUnsafe(allowUnsafe);` overwrites/deletes the options you set for `newOptions`.

What the current code does is:
1. Create `newOptions` with `AssemblyIdentityComparer` and `OptimizationLevel`
2. If `project.CompilationOptions` is `CSharpCompilationOptions`, completely ignore the above actions and only set `AllowUnsafe`

ExterminatorX99: The fix for this would be to replace
`if (project.CompilationOptions is CSharpCompilationOptions csharpComp)`
with
`if (newOptions is CSharpCompilationOptions csharpComp)`

Chik3r: ohhhh, I didn't see that, thanks for letting me know

Chik3r: Should be fixed now

ExterminatorX99: Cool. Close this thread please

Chicken-Bones:
I like the feature, and I'll test it on Linux and Mac. This will change all mod building to require VS/msbuild, which I'm okay with, but we need to make sure that's clearly communicated in setup and error messages for new modders.

If we're going this far, then why not just making ingame building call `msbuild mod.csproj` and rely on the AfterBuild target to do the packaging? I believe this was one of our long-term todos anyway.

Chicken-Bones:
I tested on Linux, and the msbuild resolving doesn't work, unsurprisingly. Running `dotnet build` works, however some of the reference assemblies end up in the lib folder and the mod fails to load.
```
[19:57:33] [.NET ThreadPool Worker/ERROR] [tML]: An error occurred while loading UtilitySlots
The mod(s) have been automatically disabled.
System.BadImageFormatException: Could not load file or assembly 'Microsoft.CSharp, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL'. Reference assemblies should not be loaded for execution.  They can only be loaded in the Reflection-only loader context. (0x80131058)
File name: 'Microsoft.CSharp, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL' ---> System.BadImageFormatException: Cannot load a reference assembly for execution.
   at System.Runtime.Loader.AssemblyLoadContext.InternalLoad(ReadOnlySpan`1 arrAssembly, ReadOnlySpan`1 arrSymbols)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromStream(Stream assembly, Stream assemblySymbols)
   at Terraria.ModLoader.Core.AssemblyManager.ModLoadContext.LoadAssembly(Byte[] code, Byte[] pdb) in tModLoader\Terraria\ModLoader\Core\AssemblyManager.cs:line 61
   at Terraria.ModLoader.Core.AssemblyManager.ModLoadContext.LoadAssemblies() in tModLoader\Terraria\ModLoader\Core\AssemblyManager.cs:line 44
   at Terraria.ModLoader.Core.AssemblyManager.InstantiateMods(List`1 modsToLoad, CancellationToken token) in tModLoader\Terraria\ModLoader\Core\AssemblyManager.cs:line 195
   at Terraria.ModLoader.Core.ModOrganizer.LoadMods(CancellationToken token) in tModLoader\Terraria\ModLoader\Core\ModOrganizer.cs:line 137
   at Terraria.ModLoader.ModLoader.Load(CancellationToken token) in tModLoader\Terraria\ModLoader\ModLoader.cs:line 126
   ```

Chik3r:
MSBuild resolving doesn't work on linux because the .NET runtime that is installed by tML doesn't seem to find the installed SDKs. (First `dotnet --info` is from global .NET, second `dotnet --info` is from tML's dotnet)
image
I've also modified MSBuildLocator to see what `dotnet` it is using (since it uses `dotnet --info` to find `MSBuild`), and it seems to be using the tML installed `dotnet`
image
This only seems to happen on Linux (and probably also on MacOS), since on Windows both the global `dotnet` and the tML installed `dotnet` can find the SDKs ( https://paste.mod.gg/crtxnemruegc/0 ).



There's probably a way to fix the `dotnet` issue, but I don't know how, so I'll try to make mod building work using the `AfterBuild` target

Chik3r:
This is now being developed in #2472.
(git didn't want to let me merge `1.4`, so I decided to make a new branch and just start a new PR from scratch. ~~why cant git just work~~)

PR #2075, opened 2022-01-05
[1.4] Remove bashism from dotnet version detection logic

RunningDroid:
### What is the bug?
If `/bin/sh` is provided by [dash](http://gondor.apana.org.au/~herbert/dash/) instead of [bash](https://www.gnu.org/software/bash/) the path to dotnet will be `dotnet/6.0.0\r/dotnet` (if the filesystem uses the bag-of-bytes approach to filenames (like Ext4); if it doesn't, then dotnet-install.sh probably fails to create the directory and bails.)

### How did you fix the bug?
I replaced the bash expression in the version detection logic with a call to [tr](https://en.wikipedia.org/wiki/Tr_(Unix)).

Replacing the `#!/bin/sh` shebang with `#!/bin/bash`. Bash is a dependency for Steam on Linux, but I don't know if it is installed by default on all Linux distributions.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/InstallNetFramework.sh
+++ b/patches/tModLoader/Terraria/release_extras/InstallNetFramework.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #Author: covers1624
 # Provided for use in tModLoader deployment. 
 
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
My preference is for the alternative fix. I'm happy with `bash` being a dependency for tML on Linux

RunningDroid:
Ok, would you prefer I open a new PR or re-use this one?

Chicken-Bones:
This one is fine :)

RunningDroid:
Ok, I've updated the PR

PR #2076, opened 2022-01-06
Caching sandstone chests

Ishigh1:
Shinies in one of the slowest passes to generate a large world, as it scans the whole world to find the top-most and the bottom-most sandstone chest for EACH sandstone chest generated.
Without the cache, I took 41 seconds to generate a large world, it went down to a single second with my PR active. (The total time of the generation without it took 216 seconds so it saved about 18.5% of the time)


Mirsario (approved):

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -526,6 +527,34 @@
...
+ 					int num10 = -1;
+ 					for (int num11 = (int)Main.worldSurface - 100; num11 < Main.maxTilesY - 200; num11++) {
+@@ -22647,7 +_,6 @@
+ 							if (Main.tile[num12, num11].wall == 216 || Main.tile[num12, num11].wall == 187) {
+ 								if (num9 == -1)
+ 									num9 = num11;
+-
+ 								num10 = num11;
+ 								break;
+ 							}
```
ExterminatorX99: Since this code is commented out, this silly patch can be removed

Ishigh1: Done

kittenchilly:
This looks great. I hope this opens the door to other worldgen optimizations in the future from vanilla.

PR #2080, opened 2022-01-08
UnifiedRandom for PrefixLoader.Roll

SadOnion:
### What is the bug?
When using Item.Prefix() with argument -1, -2 or -3 in a loop and modded prefix is chosen by PrefixLoader.Roll() it will be same prefix no matter how many times Item.Prefix() is called as long as Environment.TickCount returns different value (which is when 1 milisecond passes realtime). It is happening because PrefixLoader.Roll creates new WeightedRandom every time it is called and WeightedRandom default constructor creates new UnifiedRandom with seed = Evnvironment.TickCount
### How did you fix the bug?
Added additional argument to PrefixLoader.Roll and passed unfiedRandom object from Item.Prefix. 
I don't think so


Chicken-Bones:
Great PR, easy fix. Just be careful about including formatting fixes in your PR. Those lines at the top of `GetDrawHitbox` should be fixed in the `Terraria` patches, not the `tModLoader` ones

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -587,6 +588,17 @@
...
+ 		}
+ 
+ 		public static Rectangle GetDrawHitbox(int type, Player user) {
+-			if(Main.dedServ)
++			if (Main.dedServ)
+-			return Rectangle.Empty;
++				return Rectangle.Empty;
+ 
+ 			Main.instance.LoadItem(type);
+ 			if (ItemID.Sets.IsFood[type])
```
Chicken-Bones:
```suggestion
```

Chicken-Bones (commented):
Accidental formatting change

Issue #1699, opened 2021-08-05
[1.4] Blurriness


heh372:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

you can see here in the screenshot attaced below of how blurry everything is, and i have no clue why and how is this happening.

![Screenshot (594)](https://user-images.githubusercontent.com/66488498/128276866-8964525b-5f8b-41cf-b2ed-52debe1429a1.png)

### Log File

it did not result in a crash, theres no need of log files.

### Steps to reproduce

the issue is persistent and it's going to be like this for the time being.

### Expected Behavior

uhhhh i guess i expected it to be not as blurry since my 1.3 tml is perfectly fine

### Actual Behavior

b-blurriness, 

### Reproduction frequency

it's always there

### Additional Information

okay so this is where can i go all out without answering some irrelevant inquiry questions.

heres a list of things i did (none of them worked though):

- disabled geforce experience's in-game overlay, nothing happened
- made some changes to config.json, borderless windowed and resolution to be specific, nothing happened
- steam launch options, proven useless
- changed my display scale (well theres another person who reported the exact same thing in tml's official discord server, and by doing this they managed to solve it), did not work for me
- re-installed the entire thing, nothing happened!!!

and yeah thats all there is,

Proxiehunter:
I think this is the same issue I'm having although I was distracted from the blurriness by the fact that the UI appears to be shrunken. It is much smaller and very difficult to see despite using the same graphics settings I'm using in vanilla Terraria.

Chicken-Bones:
Please post logs regardless of if there's a crash. Also post a picture of vanilla Terraria with the same UI settings for comparison.

heh372:
> Please post logs regardless of if there's a crash. Also post a picture of vanilla Terraria with the same UI settings for comparison.

yes

JavidPack:
This issue has no recent activity or reproducible steps, so it is being closed. We assume this issue has solved itself in some manner. Feel free to open a new issue if the issue persists and fill out the issue template completely.

PR #2085, opened 2022-01-10
Design for High DPI auto support

### LINKED ISSUE #2078: Add support for high dpi displays
### Description

Add support for high dpi displays.
Use https://wiki.libsdl.org/SDL_GetDisplayDPI & FNA high dpi variable: https://github.com/FNA-XNA/FNA/wiki/7:-FNA-Environment-Variables

Reports of blurriness on M1 and Windows

Solxanich:
### What is the bug?
Closes #2078 .

Might improve #1699 , maybe even close.

### How did you fix the bug?
Used SDL2 to obtain primary monitor DPI values, and added checks against the threshold value of 96.
If higher, we enable the environment variable for FNA.

I am unable to test the fix, as my (only) monitor is exactly 96 DPI, so there is no difference either way.

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -1,6 +1,3 @@
-using MonoMod.RuntimeDetour;
-using MonoMod.Cil;
-using Mono.Cecil.Cil;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


stefnotch:
If I recall correctly, it's somewhat possible to test this on Windows by going to the "Display settings" and scrolling down to "Scale and layout"
image

Solxanich:
Tested. 
Two issues:
1) Resizing - if someone had played before, they will need to have resolution fixed when playing with windows zoom - low priority, tolerable
2) UI does not properly update scaling logic with DPI so need to fix that before merge

Thanks for the tips stefnotch!

PR #2100, opened 2022-01-19
[1.4] EM Mod.Call

mikhailmikhalchuk:
Added a `Mod.Call()` example, with the ability to retrieve and set values, and comments on how each step works. Also added miscellaneous comments for more clarity in other files.

No.


ExampleMod/Content/NPCs/ExampleBoneMerchant.cs
```diff
@@ -83,6 +83,7 @@ public class ExampleBoneMerchant : ModNPC
 		}
 
 		public override void HitEffect(int hitDirection, double damage) {
+			// Causes dust to spawn when the NPC takes damage.
 			int num = NPC.life > 0 ? 1 : 5;
```
RighteousRyan1, line 87: Maybe explain how ternary operators work?

mikhailmikhalchuk: I was considering including a comment like this, and others, but decided against it as I thought it best to keep the comments solely regarding the API itself rather than the language. If this is common practice however, it is open to change.

ExampleMod/ExampleMod.cs
```diff
@@ -26,6 +28,55 @@ public class ExampleMod : Mod
...
+			// This allows mod developers to access Example Mod's data without having to set it a reference.
+			// Mod calls are not exposed by default, so it will be up to you to publish appropriate calls for your mod, and what values they return.
+
+			// Make sure the call doesn't include anything that could potentially cause exceptions.
+			if (args is null) {
+				throw new Exception("Arguments cannot be null!");
+			}
+			if (args.Length <= 0) {
+				throw new Exception("Arguments cannot be empty!");
+			}
```
Mirsario:
```suggestion
			if (args is null) {
				throw new ArgumentNullException(nameof(args), "Arguments cannot be null!");
			}

			if (args.Length == 0) {
				throw new ArgumentException("Arguments cannot be empty!");
			}
```

ExampleMod/ExampleMod.cs
```diff
@@ -26,6 +28,55 @@ public class ExampleMod : Mod
...
+
+			// Make sure that the argument is a string using pattern matching.
+			// Since we only need one parameter, we'll take only the first item in the array.
+			if (args[0] is string content) {
```
Mirsario:
```suggestion
			// This check makes sure that the argument is a string using pattern matching.
			// Since we only need one parameter, we'll take only the first item in the array.
			if (args[0] is string content) {
```

Mirsario (approved):
Wondering if the call method should be put into a partial.

PR #1866, opened 2021-09-23
[1.4] [DRAFT] Unification and improvements in bootstrap scripts

DarioDaF:
### What is the bug?

#1859

### How did you fix the bug?

Part of it by bundling a BusyBox-like for windows and run all under bash script
Yet to write is to update Powershell by bundling a msi for it (or downloading it if necessary), but looking into regenerating env vars after install and deciding if it is ok installing a "system" update silently is ok...

### Are there alternatives to your fix?

Any compiled thing that needs to be our artifact is annoying, so if there is any decent no dep scripting language with a very small interpreter and flexible enough that could be an option

PR #2106, opened 2022-01-21
Use bash to run scripts that source scripts that contain bashisms

RunningDroid:
### What is the bug?
If tModLoader is launched via start-tModLoader.sh then InstallNetFramework.sh is run by `/bin/sh` instead of `/bin/bash`, and if `/bin/sh` is provided by [dash](http://gondor.apana.org.au/~herbert/dash/) instead of [bash](https://www.gnu.org/software/bash/) then this line fails to remove the carriage return:
https://github.com/tModLoader/tModLoader/blob/18ed9d7f4cd4efe00f3e1a48f71ba911efbbb2a4/patches/tModLoader/Terraria/release_extras/InstallNetFramework.sh#L37
### How did you fix the bug?
I fixed the bug by replacing the `#!/bin/sh` shebang with `#!/bin/bash`.
Converting the Bash scripts to Posix shell.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/start-tModLoader.sh
+++ b/patches/tModLoader/Terraria/release_extras/start-tModLoader.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 cd "$(dirname "$0")"
 script_dir="$(pwd -P)"


Solxanich:
Hi there,

Thanks for the PR. I notice that this may be duplicated in #1866 which is for a massive overhaul of these scripts so that they are cleaned up. 


Does that PR contain this bug fix, or is this different to it? I'm not too familiar with Unix to know one way or the other.

I've attached a zip of those files [LaunchScriptsTest.zip](https://github.com/tModLoader/tModLoader/files/7918736/LaunchScriptsTest.zip)
If you want to try those instead directly, feel free, and let us know as we do want to merge that PR soon once tested

RunningDroid:
PR #1866 does fix this bug, I'd just opened this PR because I'd forgotten to correct the shebang for start-tModLoader.sh in #2075 and #1866 was a draft at the time. Since #1866 is no longer a draft I'll close this PR.

Issue #2057, opened 2021-12-24
[1.4] Expert and Master Mode Accessory Slots Cause Some UI Slots to be Transparent in Non-Expert/Master Worlds


Rijam:
### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Here is a small visual bug with the UI.
If you have an accessory equipped in the 6th or 7th slot and then join a world that doesn't allow those slots, the UI elements of the Coin, Ammo, Chest, and Crafting slots become transparent (or, you know, more transparent than they should be). The slots still work like normal, so it is purely a visual thing.
Unequipping the accessories in the extra slots fixes the transparency.

![vanilla](https://user-images.githubusercontent.com/11262234/147314511-27eba048-c350-488f-89da-673dea1c11bd.png)
![tModLoader1](https://user-images.githubusercontent.com/11262234/147314517-aec98f9e-7511-430b-874e-e0284765d9f6.png)
![tModLoader2](https://user-images.githubusercontent.com/11262234/147314520-92aec874-c8b8-4732-a8ce-7bf2b864da5a.png)

The screen shots were taken with no mods loaded.

Edit: It also happens in Journey Mode if you change the difficulty setting.

### Log File

N/A

### Steps to reproduce

1. Equip an accessory in the 6th and/or 7th while in an Expert or Master Mode world.
2. Join a Classic world.
3. Coin/Ammo/Chest/Crafting UI slots are transparent

### Expected Behavior

The UI slots stay normal.

### Actual Behavior

The UI slots become very transparent.

### Reproduction frequency

Always

Solxanich:
So, I know roughly what the issue is, but since may be awhile until I could do a fix;

The transparency is intended for the extra accessory slots so that one may retrieve the accessories while in a different world, as a non contributing slot.

The problem is the way things get drawn in Main, all the values reuse the same variables, so the transparency change needs to be reset - so with the recent changes to how accessory slots get drawn, the reset for transparency was probably lost. 
The rough area for it is where DrawDefenseIcon() was called, so someone just needs to cleanup whatever stupid I did there

PR #2121, opened 2022-01-28
Fixed transparent inventory slots caused by expert and master mode accessory slots in non-expert/master worlds

hannest03:
### What is the bug?

#2057

### How did you fix the bug?

The color of the slot background was changed to be transparent, but this change was never reverted.
After the transparent slot was rendered I changed the color back to the default solid color.


Solxanich:
Fixes #2057

Issue #1932, opened 2021-11-03
Multiplying Buckets


Mionikoi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Other (please specify elsewhere)

### OS

Windows

### Platform

Steam

### Description

Version 1.4.2.1 tmodLoder v0.12.1.4 Alpha 3fb60956, built 10/30/4:25pm

condition tested
started with 3 empty buckets

Action: hold down mouse on a body of water 
Result: buckets of water will multiply to 99 in inventory with 1 bucket in hot bar.
retested results after closing, opening, saving and exiting, and creating new world. Results are consistent with water.

### Steps to reproduce

1. 3 empty buckets in inventory
2. click on a body of water
3. hold until a stack of 99 is achieved + 1 in hotbar.

### Expected Behavior

1. Fill 3 empty buckets with water
2. Empty 3 buckets of water.
3. 3 buckets are always in inventory if space allows.

### Actual Behavior

So long as there is more than a single bucket the buckets of water will multiply until 99 are in inventory and 1 bucket is in hotbar. After the 99 stack is achieved the multiplication stops until a stack of 3 are in hot bar again.

### Reproduction frequency

Every time so long as 2 or more empty buckets are in the hotbar. one bucket will empty and fill normally. 2 or more buckets were tested and always multiply buckets of water. 

### Additional Information

No attempts with honey or Lava have been tested yet. World seed is 3.3.1.1323050566, Master, large, corruption. Created September 5, 2021

Ishigh1:
There is no info on the enabled mods, usually this is caused by Max Stack Plus

Chicken-Bones:
You must give us a log file. Please come to the discord if you don't know how to do that.

ExterminatorX99:
Tested with only Cheat Sheet enabled

https://user-images.githubusercontent.com/16169502/141526099-a88541f4-5e81-4f7c-b852-671033e782f1.mp4

https://user-images.githubusercontent.com/16169502/141526396-89f8e677-eef0-4eff-aedf-ed7cabb4e225.mp4

https://user-images.githubusercontent.com/16169502/141528843-dc12c9e1-622a-4030-b047-1041c868d4c9.mp4

esotericist:
for what it's worth, i experienced this last week, but i'm no longer able to duplicate the problem since the 1.4.3.2 patch integration. it's possible this has been resolved?

edit: i stand corrected. it started happening again later in a session. weird.

PR #2122, opened 2022-01-29
Fix bucket duplication bugs

### LINKED ISSUE #2077: [1.4 Alpha] Water Bucket Duplication
NO, but it is similar to issue #1932 

### Version
v1.4.3.2 tmodloader v0.12 1.4 Alpha a878801a

### OS
Windows

### Platform
Steam

### Condition
Started with two buckets, one full of water, one empty
This was tested without any mods enabled*

### Steps to reproduce
- Required Material: 1 water bucket, 1 empty bucket
- Place Water and HOLD

### Expected Behavior
Emptying the water bucket should place only one source of water and move to stack of empty buckets

### Actual Behavior:
For every empty bucket in an inventory slot, aside from the singular water bucket, water may be placed continuously until each bucket is turned into a stack of empty buckets.

### Reproduction frequency
Consistent

### Additional Information
No attempts with honey have been tested yet. Lava was attempted but did not work.

einarmo:
### What is the bug?

This fixes #2077 and #1932. The bug in general is that there is no check for the stack being empty after decrementing stack size while using buckets, so if you held the mouse down you would continue to "use" the empty stack (decrementing `item.stack` to negative numbers).

### How did you fix the bug?

I added a small check at the end of `ItemCheck_UseBuckets` which simply clears the item if the stack is empty. This is only reached after a bucket has been used somehow, so it should be largely harmless. To stop the item animation from playing I also set `Main.blockMouse`.

To me this smells like a symptom of some deeper issue, but I do not know enough about the codebase to be sure what, or how to solve it. If a similar duplication glitch happens with other stackable but not `consumable` items, it is likely to be a similar issue caused by the same underlying lack of a check somewhere.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3969,6 +3969,18 @@
  						if (tile.halfBrick()) {
  							WorldGen.PoundTile(x, y);
  							if (Main.netMode == 1)
+@@ -32713,6 +_,11 @@
+ 						NetMessage.sendWater(tileTargetX, tileTargetY);
+ 				}
+ 			}
++
++			if (sItem.stack <= 0) {
++				sItem.TurnToAir();
++				Main.blockMouse = true;
++			}
+ 		}
+ 
+ 		private void ItemCheck_PlayInstruments(Item sItem) {
 @@ -32922,7 +_,8 @@
  				}
  


einarmo:
I'm realizing that this isn't the way to really fix this. I've spent some time trying to find the underlying cause, here is what I've managed to piece together:

 1. Last bucket in stack is used, `itemAnimation > 0`, `itemTime == 0`. These are both conditions for the bucket being used at all.
 2. Since `item.IsAir` (since stack is now 0), `itemTime `is not reset, and remains at zero.
 3. Next, item is _not_ cleared, because `itemAnimation > 0`.
 4. Next iteration, `itemAnimation `is decremented by 1 then if it hits 0 it is reset, in `ApplyItemAnimation`, called from `ItemCheck_StartActualUse`.
 5. Because `itemTime` was not reset, it is still zero, and since `itemAnimation` is always positive due to 4), we will continue using the empty stack of buckets.

To me this highlights at least one potential issue:

The three critical events related to itemAnimation have been reordered. In vanilla: `Reset if zero`, `Decrement`, `Check equal to zero for deleting the item`. In TML: `Decrement`, `Reset if zero`, `Check equal to zero for deleting`. So _unless something else kills them, items that decrement in stack size are used forever_.

This explains half the issue, that you can keep using the empty bucket _forever_. This can easily be fixed in a (safe?) way by adding `item.stack > 0` to the condition doing animation reset. In vanilla this condition is implicit, because if `itemAnimation` was 0 on the last iteration, then `item` would be nothing, and `item.useStyle` would be 0.

Another change is that `ApplyItemTime` no longer resets `useTime` if item is air. Which means that the bucket can be used every frame, not just infinitely.

For some reason making it so that buckets in particular can reset `useTime` even if they are air solves the problem. But I really don't know why, it _should_ just make the problem rare (since buckets useTime is 10, and and animation length is 14, so there are 4 frames to duplicate bucekts), but it seems something else somewhere resets the item.

While I was trying to figure this out, I found a random piece of code in `Update `that apparently sets all items with stack size 0 or type 0 to air every 10 frames (something related to gems or something), which by pure coincidence prevents the problem from occurring in this case. (though not if useTime is made lower using the new system). Not sure if this is really how this is not also a bug in vanilla.

Sorry for the long ramble, I'm probably not entirely correct, but I think I have found at least one real problem with the item use change.

einarmo:
I'll probably give this another shot later, this needs more thinking. The more I look at it the more inane vanilla behavior seems.

For example the bees knees bug: The bug is _completely expected and correct_, given the code. If `itemAnimation` is x and `itemTime` is x - 1, then the objectively correct behavior is that you should be able to use it twice in a cycle. Like if `itemAnimation` is 15 and `itemTime` is 4, then you should be able to use it three times in an animation cycle. If the use item rework changed this (I don't know how), then that technically fixed a weird vanilla bug. While not great, I think the solution is to just lower `itemAnimation` for bees knees.

The bucket bug just happens because `itemTime` < `itemAnimation` _and_ it decrements stack size (i.e. can invalidate a "global" use condition). The way I see it there are two major issues here:

First: I can't see any reason why vanilla item use logic would run if the stack size is 0. In the majority of cases, this is impossible, and if it is possible it is obviously a bug.

Second: The animation = 0 bug I mentioned in the previous comment. This one is due to the item rework.

The ApplyUseTime thing is frankly a detour. Incrementing `itemTime` for air makes no sense and is perfectly fine to ignore.

Frankly, I think both of these are vanilla bugs that didn't occur in vanilla due to pure coincidence.

einarmo:
Closing this, I'll make a new PR with my suggestion.

PR #2127, opened 2022-01-31
[1.4] CanStack & prefix weapon stacking

direwolf420:
### Description
This PR attempts to tackle several things:
**1.** Prevent stackable weapons that are prefixed from being stacked together in most relevant contexts (does not include accessories, ammo, coins etc)
**1.5**. Allow stacking items that a modder has flagged as reforgeable to be reforged
**2.** Introduce a generic `CanStack` hook in all contexts where item stack manipulation between two items take place. Currently only `CanStackInWorld` exists
**3.** **[[Mod breaking, but not often used]]** Clarify usage of several `Item` methods:
```cs
  IsTheSameAs -> internal. Porting: item.type == compareItem.type
  IsNotTheSameAs -> renamed to IsNotSameTypePrefixAndStack
```

An example for `CanStack` was added in the form of a bag that can only be opened by players who didn't craft it.

## 1. Prefix weapon stacking
Mostly integrated into **2.**
  
## 1.5. Opt-in stacking item reforging
Implementation: New property: `Item.AllowReforgeForStackableItem` for use in `SetDefaults` to give the item the capability of being reforgeable even when stack is not 1.
For vanilla items, this will not automatically make them reforgeable, as the list of items that can be reforged is hardcoded, and tml only applies general rules to modded items. The modder would have to use `GlobalItem.ChoosePrefix` and manually specify the list of prefixes.

Gameplay implications that were decided upon:
  * Stackable items with that flag set cannot roll a prefix on craft
  * Stackable items with that flag set cannot be crafted if this item is already prefixed and held in the mouse
  * Stackable items with that flag set will scale their reforge price based on the stack size
  * Bananarang and Light Disc (the only stackable non-ranged/thrower weapons) are **not** able to be reforged by default (even if the flag is true). Modders have to manually specify the list of accepted prefixes using ChoosePrefix)

## 2. CanStack
Implementation:
  * Checks for prefix equality to cover **1.**.
  * The first item parameter _should_ always be the item that the stack is increased for. Not mentioned in the docs as such for this reason, as it's not guaranteed. (There might be untested cases for the rather nested code related to `Player.ItemSpace` where I couldn't properly make sense of all its usages and which item stack it changes, but in most cases its implementation should work as expected.)

Situations where `CanStack` is applicable:
  * Any item pickup from the world (be it through player overlap, grabbing items from item/weapon frames, mannequins etc, covers void vault too)
  * in-world overlapping/stacking
  * in-inventory clicking with an item onto a different item
  * in-chest stacking (and quick stack to nearby chests)
  * in-chest/inventory sorting
  * shop purchase while a prefixed item is in the mouse

In the case of `CanStackInWorld`, `CanStack` is also called and is prioritized.

Situations where it was omitted:
  * _Automatic_ coin stacking (usually related to their dedicated slots)
  * Ammo inventory sorting (related to their dedicated slots)

Gameplay implications that were decided upon:
  *  Changing the coin stacking behavior will only partially work, the things the game makes automatically (i.e. shop interactions, coin pickup) rely on their intrinsic "conversion rate" of 100:1 between tiers. An option moving forward is to blacklist `CanStack` entry for coins only.
  
## 3. `Item` methods
`IsTheSameAs` has been made internal, as it might be confusing in a modding context for comparing two items, whereas the method JUST checks the type, nothing else.
`IsNotTheSameAs` got renamed because it might not be clear for a modder when to use it, so it's new name is now `IsNotSameTypePrefixAndStack`. (Implementation wise, it's been internalized, and a replacement method added to `Item.TML.cs`)


Chicken-Bones (approved):

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -19,6 +19,17 @@ public partial class Item : TagSerializable, IEntityWithGlobals<GlobalItem>
 
 		public RefReadOnlyArray<Instanced<GlobalItem>> Globals => new(globalItems);
 
+		/// <summary>
+		/// Set to true in SetDefaults to allow this item to receive a prefix on reforge even if maxStack is not 1.
+		/// <br>This prevents it from receiving a prefix on craft.</br>
+		/// </summary>
+		public bool allowReforgeForStackableItem = false;
```
Mirsario: Change to an autoproperty, public fields are not to appear in classes in APIs.
```cs
		public bool AllowReforgeForStackableItem { get; set; }
```

patches/tModLoader/Terraria/ModLoader/ModItem.cs
```diff
@@ -671,11 +671,21 @@ public abstract class ModItem : ModTexturedType
 		public virtual void OpenBossBag(Player player) {
 		}
 
+		/// <summary>
+		/// Allows you to decide if this item is allowed to stack with another of its type.
+		/// <br>This is only called when attempting to stack with an item of the same type.</br>
+		/// <br>This is usually not called for coins and ammo in the inventory/UI.</br>
+		/// <br>This covers all scenarios, if you just need to change in-world stacking behavior, use <see cref="CanStackInWorld"/></br>
+		/// </summary>
```
Mirsario: This is how breaks are supposed to be used.. Can you apply to this to the rest?
```suggestion
		/// Allows you to decide if this item is allowed to stack with another of its type.
		/// <br/> This is only called when attempting to stack with an item of the same type.
		/// <br/> This is usually not called for coins and ammo in the inventory/UI.
		/// <br/> This covers all scenarios, if you just need to change in-world stacking behavior, use <see cref="CanStackInWorld"/>
		/// </summary>
```

Mirsario (changes requested):
Code looks good, just nitpicks:

Mirsario (approved):

PR #2132, opened 2022-02-04
Create BirthdayParty.cs.patch

Lefthandednot:
### What is the bug?
Calling BirthdayParty.WorldClear() wouldn't deactivate the party custom sky (i.e. balloons in background).

### How did you fix the bug?
Removing the change to the _wasCelebrating variable ensures the appropriate path is followed on the next call of UpdateTime.

A call to the SkyManager could be introduced directly in the WorldClear method. This would require additional logic to ensure it not be called when the custom sky is not active or the net mode inappropriate.  This logic already exists in BirthdayParty.UpdateTime() so duplication seems unnecessary.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Events/BirthdayParty.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Events/BirthdayParty.cs.patch
@@ -1,0 +1,10 @@
+--- src/Terraria/Terraria/GameContent/Events/BirthdayParty.cs
++++ src/tModLoader/Terraria/GameContent/Events/BirthdayParty.cs
+@@ -138,7 +_,6 @@
+ 			GenuineParty = false;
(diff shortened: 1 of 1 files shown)


direwolf420:
Where is BirthdayParty.WorldClear normally called in vanilla, and why would a modder want to call it?

Lefthandednot:
> Where is BirthdayParty.WorldClear normally called in vanilla, and why would a modder want to call it?

The method is usually called during the process of loading or exiting a world during vanilla, when the party sky wouldn't normally be active anyway, however a modder my wish to call the method to stop all ongoing parties, manual and natural, with a single function call.

direwolf420:
Can you test if this properly works when you have an ongoing party in world 1, leave, then go into world 2 without a party?
And also, the same scenario but party state reversed?
Does the sky behave the exact same way as vanilla?

Issue #2048, opened 2021-12-18
[1.4] Rain doesnt fall in modded biomes


Hallam9K:
### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

The texture for rain is invisible while in modded biomes, I assume its because of how different colored rain works in 1.4 (rain sprite below)
![RainOriginal](https://user-images.githubusercontent.com/32880148/146655398-8a4023fe-73a2-4f47-b49a-d32317a426b1.png)

### Steps to reproduce

1. Create a modded biome
2. Go in it while its raining

Solxanich:
Fixed in #2195

PR #2160, opened 2022-02-20
Add system to let modded biomes have rain

einarmo:
### What is the new feature?

An additional two methods to ModWaterStyle which lets mods define their own rain texture, or use a variant of the vanilla textures.

This is noted as an issue in #2048.

Modded surface biomes should be able to have their own rain.

Almost certainly. I've been back and forth between a few different approaches. Consider this a suggestion. It impacts three different places in vanilla code:

 - The two methods in Rain.cs, which are more or less the same method duplicated twice.
 - DrawRain in Main.cs

You _might_ be able to get away with not modifying Rain.cs, but the fact that waterStyle is not explicitly included in Rain instances mean that this complicates things a fair bit, and also limits modders flexibility.

See example mod changes for a basic example.

Since the texture is returned from a method, with some cleverness you could make animated rain with this.

### ExampleMod updates

ExampleWaterStyle has an example of basic usage, making rain in the example biome green.

### DIFF
--- a/ExampleMod/Common/Players/ExampleKeybindPlayer.cs
+++ b/ExampleMod/Common/Players/ExampleKeybindPlayer.cs
@@ -2,13 +2,15 @@ using Terraria;
 using Terraria.ID;
 using Terraria.GameInput;
 using Terraria.ModLoader;
+using ExampleMod.Common.Systems;
 
 namespace ExampleMod.Common.Players
 {
+	// See Common/Systems/KeybindSystem for keybind registration.
 	public class ExampleKeybindPlayer : ModPlayer
 	{
 		public override void ProcessTriggers(TriggersSet triggersSet) {
-			if (ExampleMod.RandomBuffKeybind.JustPressed) {
+			if (KeybindSystem.RandomBuffKeybind.JustPressed) {
 				int buff = Main.rand.Next(BuffID.Count);
 				Player.AddBuff(buff, 600);
 				Main.NewText($"ExampleMod's ModKeybind was just pressed. The {Lang.GetBuffName(buff)} buff was given to the player.");
--- a/ExampleMod/Common/Systems/KeybindSystem.cs
+++ b/ExampleMod/Common/Systems/KeybindSystem.cs
@@ -1,0 +1,22 @@
+﻿using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Systems
+{
+	// Acts as a container for keybinds registered by this mod.
+	// See Common/Players/ExampleKeybindPlayer for usage.
+	public class KeybindSystem : ModSystem
+	{
+		public static ModKeybind RandomBuffKeybind { get; private set; }
+
+		public override void Load() {
+			// Registers a new keybind
+			RandomBuffKeybind = KeybindLoader.RegisterKeybind(Mod, "Random Buff", "P");
+		}
+
+		// Please see ExampleMod.cs' Unload() method for a detailed explanation of the unloading process.
+		public override void Unload() {
+			// Not required if your AssemblyLoadContext is unloading properly, but nulling out static fields can help you figure out what's keeping it loaded.
+			RandomBuffKeybind = null;
+		}
+	}
+}
--- a/ExampleMod/ExampleMod.cs
+++ b/ExampleMod/ExampleMod.cs
@@ -8,13 +8,9 @@ namespace ExampleMod
 	{
 		public const string AssetPath = $"{nameof(ExampleMod)}/Assets/";
 
-		public static ModKeybind RandomBuffKeybind;
 		public static int ExampleCustomCurrencyId;
 
 		public override void Load() {
-			// Registers a new keybind
-			RandomBuffKeybind = KeybindLoader.RegisterKeybind(this, "Random Buff", "P");
-
 			// Registers a new custom currency
 			ExampleCustomCurrencyId = CustomCurrencyManager.RegisterCurrency(new Content.Currencies.ExampleCustomCurrency(ModContent.ItemType<Content.Items.ExampleItem>(), 999L, "Mods.ExampleMod.Currencies.ExampleCustomCurrency"));
 		}
@@ -27,9 +23,6 @@ namespace ExampleMod
 
 			// NOTE: When writing unload code - be sure use 'defensive programming'. Or, in other words, you should always assume that everything in the mod you're unloading might've not even been initialized yet.
 			// NOTE: There is rarely a need to null-out values of static fields, since TML aims to completely dispose mod assemblies in-between mod reloads.
-
-			// Not required if your AssemblyLoadContext is unloading properly, but nulling out static fields can help you figure out what's keeping it loaded.
-			RandomBuffKeybind = null;
 		}
 	}
 }


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2516,6 +2516,21 @@
...
+@@ -28382,7 +_,13 @@
+ 				Rain rain = Main.rain[i];
+ 				if (rain.active) {
+ 					value.X = rain.type * 4;
++
++					Texture2D tex = value2;
++					if (rain.waterStyle >= maxLiquidTypes) {
++						tex = LoaderManager.Get<WaterStylesLoader>().Get(waterStyle).GetRainTexture();
++					}
++
```
direwolf420: Is there a way that declaring a new texture can be avoided, and instead `value2` reused? Just so that the patch below can be removed

patches/tModLoader/Terraria/ModLoader/ModWaterStyle.cs
```diff
@@ -60,6 +60,23 @@ public abstract class ModWaterStyle : ModTexturedType
 		public virtual Color BiomeHairColor() {
 			return new Color(28, 216, 94);
 		}
+
+		/// <summary>
+		/// Returns the texture to be used when drawing rain of this water type.
+		/// Default uses the vanilla rain texture.
+		/// </summary>
```
direwolf420: Line breaks in xml summaries should be prepended with `<br/>`: `<br/>Default uses the vanilla rain texture.`

ExampleMod/Content/Biomes/ExampleWaterStyle.cs
```diff
@@ -21,5 +23,13 @@ public class ExampleWaterStyle : ModWaterStyle
...
 		public override Color BiomeHairColor()
 			=> Color.White;
+
+		public override byte GetRainType() {
+			return (byte)Main.rand.Next(3);
+		}
+
+		public override Texture2D GetRainTexture() {
+			return Request<Texture2D>("ExperimentMod/Content/Biomes/ExampleRain").Value;
+		}
```
direwolf420: Should be "ExampleMod/..."

direwolf420 (commented):
Quick review of the code itself, not the design.

ExampleMod/Content/Biomes/ExampleWaterStyle.cs
direwolf420: Also get rid of the `using static Terraria.ModLoader.ModContent;` and fix the errors that come from it by qualifying the methods used from ModContent. static usings are discouraged for ExampleMod

patches/tModLoader/Terraria/Main.cs.patch
einarmo: Just use `else { value2 = TextureAssets.Rain.Value }`, yeah, probably better.

ExampleMod/Content/Biomes/ExampleWaterStyle.cs
direwolf420: Cool so it looks like that entire file wasnt yet updated to ExampleMod standards. The `=>` notation for method bodies is not used in it. Revert it to regular block bodies (like the one you had for `GetRainType` originally).

patches/tModLoader/Terraria/Main.cs.patch
direwolf420: isn't `value2` already assigned to that? No need for the else then

einarmo: This happens for each loop iteration so it needs to be reverted.

einarmo: I.e. each particle of rain on the screen, which may be a mix of vanilla and modded.

direwolf420: Oh, alright then

ExampleMod/Content/Biomes/ExampleWaterStyle.cs
```diff
@@ -1,25 +1,42 @@
...
-			=> Color.White;
+		public override Color BiomeHairColor() {
+			return Color.White;
+		}
+
+		public override byte GetRainType() {
+			return (byte)Main.rand.Next(3);
+		}
+
+
```
direwolf420: redundant newline here

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2516,6 +2516,21 @@
  		}
  
  		public void DrawItems() {
+@@ -28382,6 +_,14 @@
+ 				Rain rain = Main.rain[i];
+ 				if (rain.active) {
+ 					value.X = rain.type * 4;
++
++					if (rain.waterStyle >= maxLiquidTypes) {
++						value2 = LoaderManager.Get<WaterStylesLoader>().Get(waterStyle).GetRainTexture();
```
Chicken-Bones: Should this say `Get(rain.waterStyle)`?

einarmo: Yeah, haven't had time to test after the various corrections. I'll make sure to do so...

Chicken-Bones:
I haven't tested it myself or checked in VS, but it looks right and probably won't break anything. Let me know when you've tested ut and are happy it works

einarmo:
> I haven't tested it myself or checked in VS, but it looks right and probably won't break anything. Let me know when you've tested ut and are happy it works

Alright. Here's what I did to test it, for the record:
 - Loaded up a mod containing ExampleSurfaceBiome and other necessary code.
 - Created the biome.
 - Saw that the rain in the biome was green.
 - Tried crossing from the example biome to other biomes to verify that other biomes rain is still good, and that it transitions correctly, also when going from a custom biome to a biome that isn't forest.

It works as it should, i.e. you can have rain of both custom and vanilla colors on the screen at the same time.

patches/tModLoader/Terraria/ModLoader/ModWaterStyle.cs
```diff
@@ -60,6 +60,23 @@ public abstract class ModWaterStyle : ModTexturedType
...
 			return new Color(28, 216, 94);
 		}
+
+		/// <summary>
+		/// Returns the texture to be used when drawing rain of this water type.
+		/// <br/>Default uses the vanilla rain texture.
+		/// </summary>
+		public virtual Texture2D GetRainTexture() {
+			return TextureAssets.Rain.Value;
+		}
```
Mirsario: This should return Asset<Texture2D> (and the rendering code might need to halt if the .Value of the asset is null.)

Mirsario (changes requested):

patches/tModLoader/Terraria/ModLoader/ModWaterStyle.cs
einarmo: That makes sense. I'll make it skip the raindrop if the asset value is null. Presumably we still want to render vanilla rain.

direwolf420: Should it though? What if you programmatically generate a Texture2D? Cheatsheet does this for some of its stuff afaik

einarmo: It's a quick revert to go back to Texture2D, it being more dynamic is an argument in favor. In defense of using Asset, the rest of the textures referenced in this file are all assets in some static array.

Chicken-Bones: I think this is good, dynamic assets are a more generally required todo.

PR #2169, opened 2022-02-27
GetDrawHitbox Accounting for Item Animations

Smug-B:
### What is the bug?
Mouse text displaying item names did not account for modded animated items
See: https://cdn.discordapp.com/attachments/941876757022277673/947559840811745290/unknown.png
For an example of what I meant

### How did you fix the bug?
Patched GetDrawHitbox to include a check for registered item animations

No.


JavidPack:
Addressed in [8bc6384](https://github.com/tModLoader/tModLoader/commit/8bc63849088d87ad9bd66c9ed66c3e5bfa5d8dbb)

PR #2173, opened 2022-03-03
Improved extensibility of config UI by making the relevant classes mostly public

Tyfyter:
### What is the new feature?
Improved extensibility of config UI by making the relevant classes mostly public

Currently making a config UI element type with functionality similar but not identical to a pre-existing type requires entirely reimplementing the other type with whatever changes are desired

I am not aware of any better way to accomplish this than reducing access limitations

Custom collection UI elements for custom collection types


JavidPack:
Sorry I didn't get back on this, I was hesitant to make them public since I want to rewrite them without breaking mods. Since 1.3 is legacy now, I'll close this.

PR #2190, opened 2022-03-06
Localized folders

Ishigh1:
Currently only the name of the hjson files are taken in account, so if you want to divide a big mod into multiple files, you must put them in different folders, then name them all the same.
With this PR, the culture can be indicated with one more way : either you do as normal and name the hjson with the name of the culture, either you name a parent folder this way and give any arbitrary name to the hjson file


patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Linq;
+
+namespace Terraria.Localization;
+
+public partial class GameCulture
+{
+	public static GameCulture FromPath(string path) {
+		path = Path.ChangeExtension(path, null);
+
+		string[] split = path.Split("/");
```
ghost, line 11: Probably should be using `Path.DirectorySeparatorChar`

ghost (changes requested):

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
Ishigh1: (It doesn't work here, path contains "/" while Path.DirectorySeparatorChar is "\\" (at least on Windows)

ghost (approved):

PR #2195, opened 2022-03-12
Add system to let modded biomes have rain

einarmo:
Weird stuff happened when I tried to update my PR and it closed on its own? Either way, this is the same code, updated to latest tmodloader.

### What is the new feature?

An additional two methods to ModWaterStyle which lets mods define their own rain texture, or use a variant of the vanilla textures.

This is noted as an issue in #2048.

Modded surface biomes should be able to have their own rain.

Almost certainly. I've been back and forth between a few different approaches. Consider this a suggestion. It impacts three different places in vanilla code:

 - The two methods in Rain.cs, which are more or less the same method duplicated twice.
 - DrawRain in Main.cs

You _might_ be able to get away with not modifying Rain.cs, but the fact that waterStyle is not explicitly included in Rain instances mean that this complicates things a fair bit, and also limits modders flexibility.

See example mod changes for a basic example.

Since the texture is returned from a method, with some cleverness you could make animated rain with this.

### ExampleMod updates

ExampleWaterStyle has an example of basic usage, making rain in the example biome green.


patches/tModLoader/Terraria/ModLoader/ModWaterStyle.cs
```diff
@@ -60,6 +61,23 @@ public abstract class ModWaterStyle : ModTexturedType
...
+		/// <summary>
+		/// Returns the texture to be used when drawing rain of this water type.
+		/// <br/>Default uses the vanilla rain texture.
+		/// </summary>
+		public virtual Asset<Texture2D> GetRainTexture() {
+			return TextureAssets.Rain;
+		}
+
+		/// <summary>
+		/// Return the type of rain used. Equal to the offset in the rain texture divided by four.
```
Solxanich: The use of Type here is potentially confusing. 

I would prefer we use Variant or Variation as its a varying the drawn portion of an singular texture of an existing type, not a full blown Type (ie ModType for how we use the term) in itself.

Likewise for every other usage of Type in this PR,  use Variant instead

ExampleMod/Content/Biomes/ExampleWaterStyle.cs
```diff
@@ -1,25 +1,42 @@
...
 			b = 1f;
 		}
 
-		public override Color BiomeHairColor()
-			=> Color.White;
+		public override Color BiomeHairColor() {
+			return Color.White;
+		}
+
+		public override byte GetRainType() {
```
Solxanich: Suggest Type -> Variant

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2525,6 +2525,26 @@
...
  		public void DrawItems() {
+@@ -28356,6 +_,19 @@
+ 				Rain rain = Main.rain[i];
+ 				if (rain.active) {
+ 					value.X = rain.type * 4;
++
++					if (rain.waterStyle >= maxLiquidTypes) {
++						value2 = LoaderManager.Get<WaterStylesLoader>().Get(rain.waterStyle).GetRainTexture().Value;
++						if (value2 == null) {
++							if (flag)
```
Solxanich, line 2536: Suggest reducing nesting by using short circuit 
(Value2 == null && flag)

Solxanich (commented):
A couple of word choice cleanups
Previous PR: #2160

patches/tModLoader/Terraria/Main.cs.patch
einarmo: That would be functionally different, there's a `continue` in there as well which does not require `flag`.

patches/tModLoader/Terraria/ModLoader/ModWaterStyle.cs
einarmo: It's `type` in terraria code, hence the wordchoice, but I agree that `variant` is better.

Issue #1688, opened 2021-08-04
[1.4] Minion shots deal critical hits


Falcon-05:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Minion shots (projectiles that are shot by minions) are dealing critical hits even though minions should not have critical strike chances, excluding whip tags.

### Steps to reproduce

1. Summon a minion that shoots projectiles
2. The minion will attack enemies and occasionally crit

### Expected Behavior

The minions don't crit.

### Actual Behavior

The minions are dealing critical hits.

### Reproduction frequency

All of the time, though the crit chance is quite low (likely 4%)

### Additional Information

Here are two examples of minion shots dealing critical hits:
https://gyazo.com/9c1b1dbf174d7f29a76c379332c7897d
https://gyazo.com/943c4ace6e5c5e96e27e0d0047ac7338

Falcon-05:
After some more testing, it seems that whips (and projectiles spawned by whips, such as cool whip's snowflake projectile) are also affected by this issue.

Blookiwi:
I tried reproducing this oddity with the Tempest Staff, but it doesn't seem to happen on my side, even with my regular hyrbid set...
Even with Xeno and Hornet Staves :/
However, I can confirm that whips can indeed have some critical strikes.

Mirsario:
A temporary fix has been merged a few days ago - ede936c . But I'll keep this issue open until it's replaced/finalized.

Falcon-05:
I haven't tested it, but you should probably make sure stuff like the chlorophyte leaf crystal can't crit either

Falcon-05:
Yep, leaf crystal is dealing critical hits

Chicken-Bones:
Fixed in #2196

PR #2196, opened 2022-03-12
[1.4, HEAVY breakage inbound] the great DamageClass overhaul: part 2

ThomasThePencil:
# IN WHICH
the new `DamageClass` feature gets a supremo makeover again, because the first time wasn't nearly enough

...oh, and this PR's a remake/remaster of #1741, if you're curious. had to remodel it with the DST crossover update series in mind, because merge conflicts are cringe

## why does this exist?
that's a good question. the last PR that I did to overhaul `DamageClass`...wasn't great. it left a sour taste in my mouth, it left a lotta stuff unfinished or unimplemented, and it's even allowed a bug related to summon class crits to run rampant for months on end. so...this PR's meant to at least somewhat finish what I started with that first PR, and generally add a lot of `DamageClass` features that'll give people plenty of new toys to play with as we finally transition into the period of people actually portin' their mods to the new and (hopefully) improved API

### additions and changes
- renamed `DamageClass.NoScaling` to `DamageClass.Default`, because that's what it is (see next point)
- all items and projectiles now use `DamageClass.Default` by default instead of `DamageClass.Generic`. **this fixes (or at least, should fix) a LOT of vanilla item/projectile stat scalin' problems**
- all vanilla summon projectiles now use `DamageClass.Summon` as their `DamageType`
- added a new hook in `DamageClass` called `UseStandardCritCalcs` --- this allows you to enable or disable standard crit calcs for all items/projectiles of a given damage type. this hook's used for both `DamageClass.Default` and `DamageClass.Summon`.
- renamed the `GetBenefitFrom` and `CountsAs` hooks (in `DamageClass`) to `GetModifierInheritance` and `GetEffectInheritance`, respectively. this'll probably get rid of a lot of confusion for tML devs and modders alike
- on the subject of stat inheritances, they're now handled through a new struct, `StatInheritanceData`, which allows for more specific modifications to how your damage classes benefit from certain others. take a gander at ExampleMod if you're unsure
- `Projectile` now has a `<T>()` overload for `CountsAsClass`, just as `Item` does
- added a hook in `DamageClass` allowin' easy visibility control for various common statistical tooltip lines. this'll be made obsolete once that tooltip overhaul happens, but until then, this'll work just fine
- added flat and base stat bonuses to `StatModifier` and simplified a number of hooks accordingly. you can utilize these by appendin' `.Flat` or `.Base`, respectively, to things that return a `StatModifier`
- implemented the mythical second half of the crit fix (make projectiles set their own critical strike chance on creation and keep that chance stored, then check against it when strikin' enemies). this comes with the ability to give specific projectiles additional crit chance, if you want
- added new fields in `DamageClass`: attack speed and armor penetration
- armor penetration may now also be freely modified for `Item`s and `Projectile`s
- attack speed modifiers are now unified across classes as a `float`; It affects `useTime` `useAnimation` and `reuseDelay` uniformly, making it a plain DPS multiplier.
  - this is a deviation from vanilla melee speed, which only affected `useAnimation`, resultin' in a buff to some weapons.
  - `meleeSpeed` is longer set to its reciprocal, and is no longer capped at 3.0x speed
- added a `forTooltip` parameter to `GetWeaponDamage` which allows the additional multiplier that someone will potentially eventually use for tooltips to be applied automagically, savin' people a bit of time (addition derived from #1523)
- restored the addition of 5E-06f which existed in the original calcs. not sure how this changes things, but hey, better safe than sorry (addition transplanted from #1523)
- added `GetTotal[stat goes here]` methods for all stats currently supported by `DamageClass`, allowin' people to easily fetch the overall multiplier for any given stat and `DamageClass` combo
- updated ExampleMod with a bunch of new examples for `DamageClass` features new and old
- updated the documentation for basically everything this PR touches

#### vanilla
- the Sharpening Station's armor penetration boost can't be applied to non-melee attacks by holdin' a melee weapon (this was easiest to pull off with summons) anymore
- the long-standin' and equally infamous 'crit swap' exploit is officially and irrevocably dead, and I killed it
- fixed the Forbidden armor set bonus' ancient...sand...tornado...attack...thing. more specifically, it should no longer benefit twice from `DamageClass.Generic`'s stat boosts. I'm not sure whether this should go under bug fixes or vanilla differences, but it seems more like a bug, so /le shrug
#### tML
- minion shots and the like are now actually able to hit things again, if they weren't before
- finally fixed whips bein' able to crit. note to self: look into the whip dupe bug later...
- fixed minion and sentry damage not scalin' with all the right bonuses
- vanilla's weapon-type-specific damage bonuses (read: those from Archery Potions, the Magic Quiver and its tinkers, and Shroomite armor) should now be applied correctly again (fix derived from #1523)
#### issues here on Git
assumin' I didn't royally fuck up somewhere, this should solve everything mentioned by issues #1521 and #1688, as well as provide an alternate solution to #780. and lastly...

with this PR, I have now settled #1173 to a degree where I'm comfortable with it...at least, for the moment

## vanilla differences
since this PR's a lot more rooted in fixin' vanilla's problems than the last one, there are inevitably some changes to how items work. with that in mind, here are those changes briefly described, and their balancin' implications:
- melee speed is now an actual attack speed modifier like the rest, which means that it affects `useTime`, a trait not inherent to vanilla. please take this time to be more attentive to your projectile swords and other such weaponry, as you may find some of them are a teensy bit (read: a lot) more effective with melee speed boosts than they used to be
- as a result of the changes to attack speed, Obsidian armor's whip attack speed multiplier is now additive rather than multiplicative

to better explain the implications of the first thing, here's a spreadsheet: https://docs.google.com/spreadsheets/d/1Fcy_IyHCq_Ly44KSJZHziYS_9hD5pNCwsj7eCGwGlO0/edit?usp=sharing
it has weapon DPS difference data for a number of common weapon types that would actually change in some capacity from the melee speed changes. most melee weapon types, funnily enough, don't actually benefit in terms of attack speed on this front, so the examples on this sheet should suffice to explain what you do and don't need to be concerned over

## portin' notes (what you need to know)
for the most part, this stuff's probably self-explanatory or explained well enough by ExampleMod, BUT, for convenience's sake, here's all the stuff you need to know as a modder wantin' to get your portin' done so you can make use of the new improvements. **this PR will almost certainly shatter your mod, and you'll need to know this stuff to get it back on track, so pay attention**

### `DamageClass` stuff
- `protected float GetBenefitFrom(DamageClass damageClass)` -> `public StatInheritanceData GetModifierInheritance(DamageClass damageClass)`
  - see ExampleDamageClass in ExampleMod for more information on how to use this
- `public bool CountsAs(DamageClass damageClass)` -> `public bool GetEffectInheritance(DamageClass damageClass)`
- `DamageClass.NoScaling` -> `DamageClass.Default`

### `Player` stuff
- `minionKB` -> `GetKnockback(DamageClass.Summon).Base`
- `armorPenetration` -> `GetArmorPenetration(DamageClass.Generic)`
- `meleeSpeed` -> `GetAttackSpeed(DamageClass.Melee)`
  - no longer set to reciprocal right before `ModPlayer.PostUpdateMiscEffects`
- `whipUseTimeMultiplier` ->`GetAttackSpeed(DamageClass.SummonMeleeSpeed)`
- `arrowDamage`, `bulletDamage`, and `rocketDamage` are all `StatModifier`s now instead of `float`s
- `GetCritChance(DamageClass damageClass)` and its generic overload now give a `float` instead of an `int`, but their usage remains virtually unchanged; only use those decimal places if you *really* know what you're doin'
- `ModifyWeaponCrit` ... `ref int crit` -> `ref float crit`

### other things to note
- all `ModifyWeapon[XYZ stat here]` hooks in `ModPlayer`, `GlobalItem`, and `ModItem` have lost their `flat` parameters
- whips now use `DamageClass.SummonMeleeSpeed` instead of `DamageClass.Summon`


patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -98,20 +98,22 @@ public Item QuickSpawnItemDirect(IEntitySource source, int type, int stack = 1)
...
 
-		internal void ResetDamageClassData() {
+		internal void ResetDamageClassData()
+		{
 			damageData = new DamageClassData[DamageClassLoader.DamageClassCount];
 
-			for (int i = 0; i < damageData.Length; i++) {
-				damageData[i] = new DamageClassData(StatModifier.One, 0, StatModifier.One);
+			for (int i = 0; i < damageData.Length; i++)
+			{
```
Mirsario: Autoformat most of your additions with the editorconfig working.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4462,16 +4611,16 @@
...
-+			float flat = 0f;
-+
 +			//TODO: Add flat damage to StatModifier?
++			/*
 +			flat += meleeAddDamage * sItem.DamageType.benefitsCache[DamageClass.Melee.Type];
 +			flat += rangedAddDamage * sItem.DamageType.benefitsCache[DamageClass.Ranged.Type];
 +			flat += magicAddDamage * sItem.DamageType.benefitsCache[DamageClass.Magic.Type];
 +			flat += minionAddDamage * sItem.DamageType.benefitsCache[DamageClass.Summon.Type];
++			*/
 +
```
Mirsario: These lines weren't vanilla and are to be removed.

Mirsario (changes requested):
Discussed in Discord:
I found multiple troubles related to the addition of item speeds into the picture:
- `meleeSpeed` was something that only affected `itemAnimation`. What the added methods deal with isn't too clear.
- Vanilla summon weapons inherit `meleeSpeed` without inhering anything else melee related. This means that basic benefit scales aren't satisfactory for reimplementing that.
- Etc.

My current recommendation is to split away speed-related code from this PR, as everything else isn't even barely as complicated of a topic.

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -539,7 +536,6 @@ public abstract class GlobalItem : GlobalType<Item>
 		/// </summary>
 		/// <param name="armorSlot">head armor (0), body armor (1) or leg armor (2).</param>
 		/// <param name="type">The equipment texture ID of the item that the player is wearing.</param>
-		/// <param name="male">True if the player is male.</param>
 		/// <param name="equipSlot">The altered equipment texture ID for the legs (armorSlot 1 and 2) or head (armorSlot 0)</param>
```
direwolf420: not related to PR

patches/tModLoader/Terraria/ModLoader/ModItem.cs
```diff
@@ -148,7 +148,6 @@ public abstract class ModItem : ModTexturedType
 		/// Use AllowPrefix to prevent rolling of a certain prefix.
 		/// </summary>
 		/// <param name="pre">The prefix being applied to the item, or the roll mode. -1 is when the item is naturally generated in a chest, crafted, purchased from an NPC, looted from a grab bag (excluding presents), or dropped by a slain enemy (if it's spawned with prefixGiven: -1). -2 is when the item is rolled in the tinkerer. -3 determines if the item can be placed in the tinkerer slot.</param>
-		/// <param name="rand">The random number generator class to be used in random choices</param>
 		/// <returns></returns>
```
direwolf420: not related to PR

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -102,16 +102,16 @@ public Item QuickSpawnItemDirect(IEntitySource source, int type, int stack = 1)
...
 		}
 
 
 		/// <summary>
-		/// Gets the crit modifier for this damage type on this player.
+		/// Gets the crit chance modifier for this damage type on this player.
 		/// This returns a reference, and as such, you can freely modify this method's return value with operators.
-		/// </summary>
+		/// </summary> 
 		public ref int GetCritChance<T>() where T : DamageClass => ref GetCritChance(ModContent.GetInstance<T>());
```
direwolf420: whitespace

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4397,13 +4573,12 @@
...
-+			StatModifier kbModifier = StatModifier.One;
-+			var currentModifiers = damageData;
-+			var scalings = sItem.DamageType.benefitsCache;
-+			for (int i = 0; i < currentModifiers.Length; i++) {
-+				kbModifier = kbModifier.CombineWith(currentModifiers[i].knockback.Scale(scalings[i]));
++			StatModifier kbModifier = StatModifier.Default;
++			for (int i = 0; i < damageData.Length; i++) {
++				StatInheritanceData inheritanceData = sItem.DamageType.DoubleCheckStatInheritanceData(DamageClassLoader.DamageClasses[i], this, sItem);
++				kbModifier = kbModifier.CombineWith(damageData[i].knockback.Scale(sItem.DamageType.CheckClassStatInheritance(sItem.DamageType, this, sItem).knockbackInheritance));
 +			}
```
direwolf420: `inheritanceData` isn't used here, but the next statement uses `sItem.DamageType` instead of `DamageClassLoader.DamageClasses[i]`, what's the intended one?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4438,18 +4611,15 @@
...
-+			StatModifier damage = StatModifier.One;
-+			var currentModifiers = damageData;
-+			float[] scalings = sItem.DamageType.benefitsCache;
-+
-+			for (int i = 0; i < currentModifiers.Length; i++) {
-+				damage = damage.CombineWith(currentModifiers[i].damage.Scale(scalings[i]));
++			StatModifier damage = StatModifier.Default;
++			for (int i = 0; i < damageData.Length; i++) {
++				damage = damage.CombineWith(damageData[i].damage.Scale(DamageClassLoader.statInheritanceCache[sItem.DamageType.Type, i].damageInheritance));
 +			}
```
direwolf420: This looks distinctly different compared to previous similar code, what's up with that?

direwolf420 (commented):
General consistency/readability requests

patches/tModLoader/Terraria/Player.cs.patch
ThomasThePencil: the intended one is `DamageClassLoader.DamageClasses[i]`. must've been an oversight on my end

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -17,15 +17,75 @@ public partial class Projectile : IEntityWithGlobals<GlobalProjectile>
...
+				projectile.CritChance += (actualSpawnSource.Entity as Player).GetWeaponCrit(actualSpawnSource.Item);
+				projectile.ArmorPenetration += (actualSpawnSource.Entity as Player).GetWeaponArmorPenetration(actualSpawnSource.Item);
+			}
+			else {
+				EntitySource_ItemUse_WithAmmo actualSpawnSource = spawnSource as EntitySource_ItemUse_WithAmmo;
+				if (actualSpawnSource.Entity is not Player)
+					return;
+				projectile.CritChance += (actualSpawnSource.Entity as Player).GetWeaponCrit(actualSpawnSource.Item);
+				projectile.ArmorPenetration += (actualSpawnSource.Entity as Player).GetWeaponArmorPenetration(actualSpawnSource.Item);
+			}
```
Chicken-Bones: Second half of method is redundant, because `EntitySource_ItemUse_WithAmmo : EntitySource_ItemUse`

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -17,15 +17,75 @@ public partial class Projectile : IEntityWithGlobals<GlobalProjectile>
...
+		/// The number of defense points that this projectile can ignore on its own. Cannot be set to negative values. Defaults to 0.
+		/// On spawn, if this projectile was fired from a weapon, this value has the total armor penetration of the weapon that made the projectile added to itself.
+		/// </summary>
+		public int ArmorPenetration {
+			get => _armorPenetration;
+			set {
+				if (value < 0)
+					throw new Exception("A projectile's armor penetration value cannot be set below 0.");
+				else
+					_armorPenetration = value;
```
Chicken-Bones:
```suggestion
				if (value < 0) throw new ArgumentException($"{nameof(Projectile)}.{nameof(ArmorPenetration)} must be >= 0");
					_armorPenetration = value;
```

and similarly below.

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -25,36 +23,72 @@ public abstract class DamageClass : ModType
...
+
+		/// <summary> 
 		/// This lets you define the classes that this DamageClass will count as (other than itself) for the purpose of armor and accessory effects, such as Spectre armor's bolts on magic attacks, or Magma Stone's Hellfire debuff on melee attacks.
 		/// Returns false in all cases by default, which does not let any other classes' effects trigger on this DamageClass.
+		/// For a more in-depth explanation and demonstration, refer to ExampleMod/Content/DamageClasses/ExampleDamageClass.
+		/// DO NOT USE THIS TO CHECK IF AN ITEM OR PROJECTILE'S DAMAGE TYPE MATCHES THE INPUT DAMAGE CLASS. For that purpose, use (Item/Projectile).CountsAsClass(DamageClass damageClass) instead.
 		/// </summary>
 		/// <param name="damageClass">The DamageClass which you want this DamageClass to gain effects from.</param>
-		public virtual bool CountsAs(DamageClass damageClass) => false;
+		public virtual bool CheckClassEffectInheritance(DamageClass damageClass) => false;
```
Chicken-Bones: Propose `GetEffectInheritance`

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -25,36 +23,72 @@ public abstract class DamageClass : ModType
...
 		/// <param name="damageClass">The DamageClass which you want this DamageClass to benefit from statistically.</param>
-		protected virtual float GetBenefitFrom(DamageClass damageClass) => 0;
-
-		internal void RebuildBenefitCache() {
-			benefitsCache = DamageClassLoader.DamageClasses.Select(GetBenefitFrom).ToArray();
-			benefitsCache[Type] = 1f;
-		}
-
-		public float GetCachedBenefitFrom(DamageClass damageClass) => benefitsCache[damageClass.Type];
+		public virtual StatInheritanceData CheckBaseClassStatInheritance(DamageClass damageClass) => new StatInheritanceData(0f, 0f, 0f, 0f, 0f, 0f, 0f);
```
Chicken-Bones: Propose `GetModifierInheritance`

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -25,36 +23,72 @@ public abstract class DamageClass : ModType
...
+		/// This lets you define the classes that this DamageClass will benefit from (other than itself) for the purposes of stat bonuses, such as damage and crit chance.
+		/// This returns either null (no dynamic effect) or a struct called StatInheritanceData.
+		/// This is used to allow extensive specifications for what your damage class can and can't benefit from in terms of other classes.
+		/// For a more in-depth explanation and demonstration, refer to ExampleMod/Content/DamageClasses/ExampleDamageClass.
+		/// THIS METHOD SHOULD ONLY BE USED FOR STAT INHERITANCES GRANTED BY ARMOR, ACCESSORIES, OR OTHER EFFECTS. For base stat inheritances, use CheckBaseClassStatInheritance instead.
+		/// </summary>
+		/// <param name="damageClass">The DamageClass which you want this DamageClass to benefit from statistically.</param>
+		/// <param name="player">The player being referenced for dynamic calculations.</param>
+		/// <param name="item">The item being referenced for dynamic calculations.</param>
+		public virtual StatInheritanceData? CheckDynamicClassStatInheritance(DamageClass damageClass, Player player, Item item) => null;
```
Chicken-Bones: Propose `GetModifierInheritance`. I question whether there are valid use cases for this method, can we leave this out?

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -25,36 +23,72 @@ public abstract class DamageClass : ModType
...
-		/// <summary> This is the name that will show up when an item tooltip displays 'X [ClassName]'. This should include the 'damage' part. </summary>
+		/// <summary>
+		/// This is the name that will show up when an item tooltip displays 'X [ClassName]'.
+		/// This should include the 'damage' part.
+		/// </summary>
 		public string DisplayName => DisplayNameInternal;
 
 		internal protected virtual string DisplayNameInternal => ClassName.GetTranslation(Language.ActiveCulture);
 
+		public StatInheritanceData DoubleCheckStatInheritanceData(DamageClass damageClass, Player player, Item item) {
```
Chicken-Bones: If we decide to keep the 'dynamic' `GetModifierInheritance` version, then that can just call the 'base' one as a default implementation and this method wouldn't be required.

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -25,36 +23,72 @@ public abstract class DamageClass : ModType
...
+		/// <summary> 
+		/// This lets you define default stat modifiers for all items of this class (e.g. base crit chance).
+		/// </summary>
+		public virtual void SetDefaultStats(Player player) { }
+
+		/// <summary>
+		/// This lets you decide whether or not your damage class can use standard crit chance calculations.
+		/// Setting this to false will also hide the critical strike chance line in the tooltip of any item that uses this DamageClass.
+		/// </summary>
+		public virtual bool AllowStandardCrits => true;
```
Chicken-Bones: Maybe just `CanCrit`? Normally I'd suggest just not inheriting any crit chance, but this does have the benefit of hiding the tooltip line...

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -9,59 +10,70 @@ public abstract class VanillaDamageClass : DamageClass
...
 
 	public class SummonDamageClass : VanillaDamageClass
 	{
 		protected override string LangKey => "LegacyTooltip.53";
 
-		protected override float GetBenefitFrom(DamageClass damageClass) => damageClass == Generic ? 1f : 0f;
+		public override StatInheritanceData CheckBaseClassStatInheritance(DamageClass damageClass) => damageClass == Generic ? new StatInheritanceData() : new StatInheritanceData(0f);
+
+		public override StatInheritanceData? CheckDynamicClassStatInheritance(DamageClass damageClass, Player player, Item item) {
+			if (ItemID.Sets.SummonerWeaponThatScalesWithAttackSpeed[item.type] && damageClass == Melee)
```
Chicken-Bones: I would really really prefer if there was another class, called `SummonMeleeDamageClass` which just had different base inheritance, and then this set could be removed/hidden, and everything would be much more declarative.

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -9,59 +10,70 @@ public abstract class VanillaDamageClass : DamageClass
 
 		protected abstract string LangKey { get; }
 
-		public override bool CountsAs(DamageClass damageClass) => false;
+		public override bool CheckClassEffectInheritance(DamageClass damageClass) => false;
```
Chicken-Bones: `false` is already the default here.

patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
```diff
@@ -3,15 +3,47 @@
 
 namespace Terraria.ModLoader
 {
+	public struct StatInheritanceData
```
Chicken-Bones: Move this out from `DamageClassLoader.cs` and just name it `StatInheritance`

patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
```diff
@@ -3,15 +3,47 @@
...
+	{
+		public float damageInheritance;
+		public float useTimeInheritance;
+		public float useAnimationInheritance;
+		public float useSpeedInheritance;
+		public float critChanceInheritance;
+		public float knockbackInheritance;
+		public float armorPenInheritance;
+
+		public StatInheritanceData() {
```
Chicken-Bones: `default` constructors for structs are ill advised. Follow the `StatModifier` pattern. 
Consider `StatInheritanceData.None` and `StatInheritanceData.Full`

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -9,59 +10,70 @@ public abstract class VanillaDamageClass : DamageClass
...
+			player.GetCritChance(this) = 4;
+		}
 	}
 
 	public class MeleeDamageClass : VanillaDamageClass
 	{
 		protected override string LangKey => "LegacyTooltip.2";
 
-		protected override float GetBenefitFrom(DamageClass damageClass) => damageClass == Generic ? 1f : 0f;
+		public override StatInheritanceData CheckBaseClassStatInheritance(DamageClass damageClass) => damageClass == Generic ? new StatInheritanceData() : new StatInheritanceData(0f);
```
Chicken-Bones: I think this is a sensible default is for all custom `DamageClass`. With the proposed `static` fields, consider `damageClass == Generic ? StatInheritance.Full : StatInheritance.None`

patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
```diff
@@ -32,23 +64,34 @@ public static class DamageClassLoader
...
+			RebuildEffectInheritanceCache();
 		}
 
 		internal static void Unload() {
 			DamageClasses.RemoveRange(DefaultClassCount, DamageClasses.Count - DefaultClassCount);
 		}
 
-		private static void RebuildCountsAsCache() {
-			countsAs = new bool[DamageClassCount, DamageClassCount];
+		private static void RebuildStatInheritanceCache() {
```
Chicken-Bones: This is... not actually used. It may be a good optimisation, but lets remove it for now

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
ThomasThePencil: even if we can't think of an immediate use case, somebody'll find one, and when they do, they'll be thankin' me for doin' this. you know how this community is with new possibilities

it's got the potential to make wrestlin' vanilla stuff into doin' what we want easier, and it's bound to open up a shitload of new routes for people to take with their mod design. we're keepin' it

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
ThomasThePencil: not inheritin' any crit chance won't solve the fact that the class itself can be given crit chance by accessories and other things. modifyin' stat inheritances doesn't work on the class you're selectin' the stat inheritance modifiers of

as for the name, `CanCrit` is noticeably more vague. `AllowStandardCrits` is straightforward and immediately tells the modder what's up. I think I'll be keepin' it as is

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
ThomasThePencil: fair enough

patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
ThomasThePencil: fair enough. not entirely sure why I kept it in the first place

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
ThomasThePencil: not entirely sure I understand. the static version and dynamic version are two different methods, and static is used in cases where a Player or Item isn't available

patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
ThomasThePencil: can separate from `DamageClassLoader.cs`, but I fail to see a substantial reason the name should be changed to something easier to misinterpret

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
ThomasThePencil: can do. new system makes it way easier to set up something better for whips. should be nice enough on patches, too

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
ThomasThePencil: went with `GetStaticModifierInheritance` due to keepin' dynamic stat inheritances, but yeah

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
Chicken-Bones: We really need compelling use cases for everything we add. Features without use cases are difficult to maintain, can be a source of confusion for modders, and make meta-systems (like a mod which displays damage class relationships) harder to make.

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
Chicken-Bones: Ah, I was saying they can have the same name, because they have different parameters, but as above I'm in favour of removing it. All tML hooks need sensible examples.

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
Chicken-Bones: I'd be happy for it to go either way in summon's case. If someone wants to add summon crit, why not. On the other hand, being very clear that 'this damage class can't crit' is also good and useful.

`CanCrit` should be the name, or `AllowCrits` remove the `Standard` because that implies that non-standard crits may be allowed (what would a non-standard crit be) (yes, people will ask this)

ThomasThePencil: "what would a non-standard crit be"
Lightning Aura sentries can crit if you wear full Monk (16.67%) or Shinobi Infiltrator (25%) sets, and minions can crit if the target is tagged by either Morning Star (5%) or Kaleidoscope (10%). these use fixed chances and are calculated entirely separately from normal crit calcs. I can add this answer somewhere in the ExampleMod documentation and/or an eventual Completely Accurate Guide to damage classes if need be, but I think it's seriously worth notin'. `CanCrit` implies that if set to false, an item or projectile from the DamageClass cannot crit **at all**, even if specifically given non-standard crit calculations to consider, which simply isn't the case; summon weapons simply can't crit *from standard crit calcs*, hence the name of the property

with all this said, I think I know a better name for the property: `Use(Normal/Standard)CritCalcs`, which is even more clear about the actual function of the property. whether it uses Normal or Standard is up for debate, but I'm leanin' towards Standard

Chicken-Bones: Well, the more you know. Sounds fine to me! But maybe then `ShowCritTooltip` (or the tooltip hook you already have) should be separate from whether standard crits are enabled, since you might have your own crit logic and still want to display the tooltip?

Chicken-Bones: `UseStandardCritCalcs` is a good name, with a comment about the above

ThomasThePencil:
as I FINALLY finish up with this God-forsaken PR, here's some stuff that's currently out of scope, as this PR is already _multiple normal PRs in size_ as is, but which should probably be investigated at some point. may add to this list later if things come up, but don't take any of these as things that will be added on this PR; please just review what I have so I can be done

- dynamic effect inheritance
this has applications with e.g. an accessory that allows melee weapons which shoot projectiles to have those projectiles proc magic effects in exchange for a mana cost when used
most applications, however, would require the addition of the next item on the list

- GlobalDamageClass implementation. how do you make this make sense? seriously, it's dumb
either way, has applications primarily for removin' the limitations on summon crits (I can figure a lot of people will want to do this)
in tandem with dynamic effect inheritance, it could also be used for generally creatin' interestin' effects that interface with vanilla damage classes
however, allowin' people to mess with existin' damage classes is...a very slippery slope, and if we're gonna do that, it's GOTTA be done right, so for now it's been omitted

- potential refactor of inheritances on the whole to use a single, unified method
primary concerns are oversimplfiication, potential troubles with providin' learnin' resources, and the inevitable question of "how well can you keep the average new modder from not blowin' up their game"
however, a single, unified method could also provide an easier way to get all the learnin' related to it into one spot

- potential refactor of `ModifyWeaponXYZ` methods to be a single `ModifyWeaponStats` method, utilizin' a new `DamageClassData` class which becomes extensible with modded stats and vanilla's non-universal stats
primary concerns are "holy shit mana cost is dumb (and so is minin' speed)" and how practical such extensibility becomes for the average modder to take advantage of; we've already made a simple damage increase too verbose as it is

- addition of new structs for crit chance, attack speed, and armor pen to properly limit how people interact with them
could probably be avoided through sufficiently effective learnin' resources, but I know how scared most of this community is of actually havin' those, so

why did I throw away over a year of my life on a feature people will almost certainly take for granted, you ask?
the answer is "because I knew it could be better, and because I knew people would eventually find dissatisfaction with its features as it was originally implemented". I think that's a good enough response to that question to warrant something of this scale

as a last note, I hate whips. they have actively worked against about half the stuff in this overhaul. they are dead to me. I'm never workin' with them again before I make them melee weapons for DD. my final message. thank you, and good fuckin' night

Issue #1918, opened 2021-10-23
Mounts are inside blue box (or grey or gold)


Tsspidermine:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Mac

### Platform

Steam

### Description

When I get on a mount such as the Drill Containment Unit, the texture for the mount is inside of a blue box
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/51246489/138568622-ce4be546-ee70-4773-9b70-8cbd2cdf0d2d.png">

### Steps to reproduce

1. Get on mount by either using the item, or equipping it and pressing R

### Expected Behavior

There should be no blue box around the mount's texture

### Actual Behavior

There is a blue box enshrouding the mount

### Reproduction frequency

Every time I use my mount, I get the box.

Jawell:
Duplicates:
#2158 
#2191

Mirsario:
Issue seems to be common but exclusive to macs, aforementioned duplicates have extra info.
Seems like both of the provided logs use OpenGL rendering, need to check whether Vulkan/MoltenVK changes anything.
For anyone who wants to try, you need to start the game in a shell using:
`FNA3D_FORCE_DRIVER=Vulkan ./start-tModLoader.sh`

Solxanich:
For those on Steam, you can also set the steam launch parameter to have:
/gldevice:Vulkan.

Steam Launch Parameter location shown below:
https://media.discordapp.net/attachments/871289059396448337/937463311405056000/unknown.png

Jawell:
@Solxanich thank you, it's works for me

Can we implement something like auto choose rendering without launch parameters? Or this issue relevant only for tmodloader alpha version? Because, I don't see any problems in native terraria 1.4

Solxanich:
An indirect fix was attempted, however not all Mac devices supported Vulkan.
The fix was reverted, and thus re-opened

Goseale:
I have the same issue and /gldevice:Vulkan didnt work. I'm running PopOS 22.01 LTS with KDE Plasma

Shadowblitz16:
I'm on linux (KDE neon 5.24) and my boxes are grey.
![image](https://user-images.githubusercontent.com/2453707/169920269-bbb00ff7-15a5-4230-8f4e-a30db3c5d680.png)

Shadowblitz16:
Trying vulkan gives some strange visual bugs...
[...]

PR #2208, opened 2022-03-18
[1.4] Force Vulkan on MacOS

ExterminatorX99:
### What is the bug?
#1918 

### How did you fix the bug?
Force Vulkan rendering on MacOS

No clue. Ask Solxan


JavidPack:
This fix is leading to some mac users not being able to launch the game when they previously could. I had them delete this added line and it worked again for them.

[client (2).log](https://github.com/tModLoader/tModLoader/files/8307760/client.2.log)
 

Graphics card of this log: Radeon Pro Vega 20

Solxanich:
Hmm. Possibly revert then?
No launch is worse than graphics error. 

Just spitballing; if neither OpenGL nor Vulkan work, and Metal did, what are the options?

Chicken-Bones:
can we fallback to OpenGL? I wonder what they actual blue box issue is, it shouldn't be happening :|

PR #2209, opened 2022-03-19
[1.4] Make Held Items Use Registered Animations

absoluteAquarian:
### What is the new feature?
Item use animations notoriously do not support animations.  These changes 1) allow the item's registered animation affect its use sprite and 2) affect the item's hitbox to be the source rectangle of the animation's frame.

Modder convenience, mostly.  Currently, in order to have an item have an animation while in use, you have to use a held projectile.

None that I can think of.

n/a

### ExampleMod updates
n/a


absoluteAquarian:
Not sure why so many commits appear, but the only changes in this PR are from the most recent commit.

PR #2212, opened 2022-03-20
Fix ScriptCaller.sh never checking for system dotnet

ppebb:
### What is the bug?
ScriptCaller.sh never actually checks for a system install of dotnet.

### How did you fix the bug?
Check dotnet exists on the system on the path, then compare the version string with the one in runtimeconfig.json

Someone could probably write it cleaner


Chicken-Bones:
This is not a bug. tML should always run with the portable .NET so that we have a consistent environment across all users. The only  exception is when running tML from Visual Studio.

Future .NET versions are not guaranteed to be backwards compatible, particularly because MonoMod relies on JIT internals.

PR #2214, opened 2022-03-20
Added example Whip files

Japaneixxx:
Added example whip files because at the moment there is none

I dont think it needs additional documentation


Chicken-Bones:
Currently considering #1341 over this because of the hook. If there's anything you'd like to add to that example to combine the two, I'd welcome your feedback

PR #2220, opened 2022-03-22
Example Flail

dpoopoop:
It adds an example flail with the new 1.4 flail AI, with 2 options for flail head rotation!

Maybe change some of the variable's names that don't really explain the variable fully


ilija04:
Is this done? I really want to add flails in my mod, so i attempted it using the ExampleAdvancedFlail's code copied directly from this PR but i forgot that hooks are either missing or private. Looking forward to it!

PR #2228, opened 2022-03-26
[1.4] NPC Buff Arrays Rework + Netcode Optimizations

absoluteAquarian:
### What is the new feature?
* `NPC.buffType` and `NPC.buffTime` were changed from `int[]` to `System.Collections.Generic.List<int>`
* netcode for `MessageID.SyncNPCBuffs` was modified to only send as much information as how many buffs as the NPC has
* `NPC.DelBuff(int buffIndex)` was changed to `NPC.DelBuff(ref int buffIndex, bool quiet = false)`
  * The change to `buffIndex` was necessary for code that loops through `NPC.buffType`/`NPC.buffTime` to not break when removing buffs via `NPC.DelBuff()`
  * The addition of `quiet` is for certain netcode optimizations (`NPC.DelBuff()` was being called in a loop responsible for clearing buffs that have ended, which would cause netcode spam)
* `NPC.maxBuffs` was made obsolete.  Modders should use `NPC.BuffCount` on an `NPC` instance instead.

Previously, NPCs had a hard limit of 5 buffs/debuffs active at any given moment.  This PR removes that limit entirely.

None that I can think of, but any better alternatives are welcome.

n/a

### ExampleMod updates
n/a

### DIFF
--- a/patches/tModLoader/Terraria/NPC.cs.patch
+++ b/patches/tModLoader/Terraria/NPC.cs.patch
@@ -136,7 +136,7 @@
  			waterMovementSpeed = (lavaMovementSpeed = 0.5f);
  			honeyMovementSpeed = 0.25f;
  			netOffset *= 0f;
-@@ -1949,15 +_,24 @@
+@@ -1949,15 +_,30 @@
  				oldPos[j].Y = 0f;
  			}
  
@@ -147,8 +147,14 @@
  			}
 +			*/
 +
-+			buffTime = new();
-+			buffType = new();
++			if (ModNet.AllowVanillaClients) {
++				//Vanilla behavior:  a constant size of 5 buff slots
++				buffTime = new(new int[5]);
++				buffType = new(new int[5]);
++			} else {
++				buffTime = new();
++				buffType = new();
++			}
 +
 +			if (buffImmune.Length != BuffLoader.BuffCount)
 +				Array.Resize(ref buffImmune, BuffLoader.BuffCount);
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -130,13 +136,24 @@
...
  			netOffset *= 0f;
-@@ -1954,10 +_,14 @@
- 				buffType[k] = 0;
+@@ -1949,15 +_,24 @@
+ 				oldPos[j].Y = 0f;
  			}
  
++			/*
+-			for (int k = 0; k < 5; k++) {
++			for (int k = 0; k < BuffCount; k++) {
```
ExterminatorX99: Unnecessary patch

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -844,6 +923,39 @@
...
+-			for (int i = 0; i < 4; i++) {
+-				if (buffTime[i] == 0 || buffType[i] == 0) {
+-					for (int j = i + 1; j < 5; j++) {
+-						buffTime[j - 1] = buffTime[j];
+-						buffType[j - 1] = buffType[j];
+-						buffTime[j] = 0;
+-						buffType[j] = 0;
+-					}
+-				}
+-			}
```
ExterminatorX99: Comment out instead?

patches/tModLoader/Terraria/NPC.cs.patch
absoluteAquarian: right, forgot about that one

patches/tModLoader/Terraria/NPC.cs.patch
absoluteAquarian: good point

Issue #1491, opened 2021-04-30
{Far Future] ModAchievements & tML Achievements


Solxanich:
### Description
Add a new system for having in-game achievements in tModLoader, that are independent to steam.
An example reference for ModAchievements could be the modded library [Achievement Lib](https://mirror.sgkoi.dev/Mods/Details/AchievementLib)

It may also be fruitful to consider tML specific achievements (such as playing X hours, using more than 4GB of memory, wearing a full patreon/dev set, playing ExampleMod bc example achievement) 

### What does this proposal attempt to solve or improve?
Reduction of 3rd party libraries that accomplish features that should be in tML.

### Which (other) solutions should be considered?

TBD, this is FAR future

Patcheresu:
Some suggestions:
Playing TML for a certain number of hours
Creating a world that has a mod alter worldgen
Downloading 'any' mod
Making 'any' mod
Assigning a key to 'any' mod altered hotkey in the settings menu
Beating a Pre-Hardmode non-Vanilla boss in TML
Beating a Hardmode non-Vanilla boss in TML
Making furniture, finding weapons, crafting items that aren't from Vanilla
Beating Moon Lord in TML

Steviegt6:
> Some suggestions:
> Playing TML for a certain number of hours
> Creating a world that has a mod alter worldgen
> Downloading 'any' mod
> Making 'any' mod
> Assigning a key to 'any' mod altered hotkey in the settings menu
> Beating a Pre-Hardmode non-Vanilla boss in TML
> Beating a Hardmode non-Vanilla boss in TML
> Making furniture, finding weapons, crafting items that aren't from Vanilla
> Beating Moon Lord in TML

Pre-made tML achievements are kind of dumb. tML is a modding API and mod loader, not a content mod in itself (save for some Patreon sets, an old April Fool's item back from when Blushie was the main dev, and a few items to help with mod compatibility (unloaded items, Starter Bag).

Steviegt6:
> ### Description
> Add a new system for having in-game achievements in tModLoader, that are independent to steam.
> An example reference for ModAchievements could be the modded library [Achievement Lib](https://mirror.sgkoi.dev/Mods/Details/AchievementLib)
> 
> It may also be fruitful to consider tML specific achievements (such as playing X hours, using more than 4GB of memory, wearing a full patreon/dev set, playing ExampleMod bc example achievement, setting a hotkey, downloading any mod)
> 
[...]

PR #2254, opened 2022-04-05
[1.4] [Feature] Modded Achievements

steviegt6:
### What is the new feature?
Enables mods to add modded achievements.

Achievements are a well-known vanilla feature and many modders and players want mods to be able to add achievements.

To be discussed.

TODO

### ExampleMod updates
TODO


patches/tModLoader/Terraria/ID/AchievementID.cs
```diff
@@ -0,0 +1,114 @@
+﻿namespace Terraria.ID
+{
+	/// <summary>
+	///		Constant IDs for vanilla achievements, provided by tModLoader.
+	/// </summary>
+	public static class AchievementID
```
xIceFox, line 6: maybe do this as an enum? Then you can pick the Achievement ID with AchievementID.Timber for example

Steviegt6: No.

Chicken-Bones:
Linking to #1491

Issue #2250, opened 2022-04-04
[1.4] IndexOutOfRangeException at the end of Credits roll after defeating Moon Lord


blahblahbal:
### Version

Other (please specify elsewhere)

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

I defeated the Moon Lord in a fresh world and watched the credits, and at the end of them, I got an IndexOutOfRangeException. I'm the latest 1.4-stable on Steam (v2022.3.35.3).

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/8406144/client.log)

### Steps to reproduce

1. Spawn and defeat the Moon Lord.
2. Watch the credits until the end
3. Note the error in the chat

### Expected Behavior

No error

### Actual Behavior

The error shows up in chat

### Reproduction frequency

I haven't tested it more than once, since it's a rather long process

### Additional Information

Nothing

Solxanich:
Fixed in #2256

PR #2256, opened 2022-04-05
[1.4] Fix credit roll crashing at the end when trying to draw player

direwolf420:
### What is the bug?
#2250:
Due to `Segments.PlayerSegment._player` being initialized way before any mods are loaded, calling methods on this player (which vanilla does) will crash since no `ModPlayer`s are set up for it.

### How did you fix the bug?
Reinitialize the entire `CreditsRollSky` object in `ModContent.ResizeArrays` (running on load and unload) which contains the `Segments.PlayerSegment` constructor, reinitializing the player aswell, and making sure `Player.modPlayers` is populated.

Also removed outdated/redundant comments from 1.3 (where `Player` ctor had an optional parameter) to reduce patch count.

Making `_player` static, essentially a "shared" rendering dummy. Problem is the fact that vanilla uses this object to bind it to cutscene actions, reinitializing this object afterwards will disconnect it.


steviegt6 (approved):

steviegt6:
LGTM

PR #2264, opened 2022-04-09
Add IsSlime, IsBunny, IsSquirrel, IsButterfly and IsBird + example.

Cyrillya:
### What is the new feature?
Add `IsSlime`, `IsBunny`, `IsSquirrel`, `IsButterfly` and `IsBird` to `NPCID.Sets`.
It is quite similar wiith `NPCID.Sets.IsDragonfly` in vanilla.

For example, if you wanted to make all slimes 2x bigger, you would have to use a long "if" to determine if it was one of the many slimes in vanilla. And it doesn't support another mod that adds slimes at all.
With these new static `NPCID.Sets`, modders can more easily accomplish these functions. It also performs well in cross-mod content if the modder sets the values correctly.

### ExampleMod updates
In ExampleCustomAISlimeNPC.cs: `NPCID.Sets.IsSlime[Type] = true`
PartyZombie also has a fix to make it count as a zombie. In PartyZombie.cs: `NPCID.Sets.Zombies[Type] = true`

### DIFF
--- a/ExampleMod/Content/NPCs/ExampleCustomAISlimeNPC.cs
+++ b/ExampleMod/Content/NPCs/ExampleCustomAISlimeNPC.cs
@@ -44,6 +44,8 @@ namespace ExampleMod.Content.NPCs
 			// DisplayName.SetDefault("Flutter Slime"); // Automatic from localization files
 			Main.npcFrameCount[NPC.type] = 6; // make sure to set this for your modnpcs.
 
+			NPCID.Sets.IsSlime[Type] = true; // This NPC will be considered a slime.
+			
 			// Specify the debuffs it is immune to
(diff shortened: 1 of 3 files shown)


Chicken-Bones:
hmmm, sure, You may want to consider all the critters in `NPCID.Sets.CountsAsCritter`, perhaps should also modify the checks in `MechSpawn`?

Ouroel:
At this point we should really just use a BitsByte array so we can compress it 8fold

direwolf420:
Imo making sets like this opens a can of worms. Why not add sets for all other "archetypes" of enemies?
Also this might be covered/obsolete once this is implemented: <https://github.com/tModLoader/tModLoader/pull/1083>

PR #2277, opened 2022-04-11
Re-implemented internal names for AddEquipTexture() and ported the equip methods from Mod to EquipLoader

IbanPlay:
### What is the new feature?
`Mod.AddEquipTexture()` got turned into `EquipLoader.AddEquipTexture()`, and it can now be used to register equipment textures through an internal name instead of being locked into using the name of their associated item as an internal name. This also allows mods to register equip textures without needing an item reference, and most importantly, allows them to register multiple equip textures of the same type for the same item.

This was a feature in 1.3 tMod, and its removal ends up forcing modders to rely on custom player draw sets to achieve effects that previously were much simpler to implement.

None that I can think of, since this is really just a port of a 1.3 tMod feature that got left out during the 1.4 port due to whoever was working on it not knowing the use cases for it (?).

The main draw of letting mods use internal names for `EquipTexture`s is that multiple ones of the same type can be registered for the same item, which lets mods easily implement alt textures for armor sets and accessories. 

### ExampleMod updates
Added an extra part to `ExampleCostume` and `ExampleCostumePlayer` which shows how mods can implement multiple equip textures of the same type on the same item.


patches/tModLoader/Terraria/ModLoader/Default/Developer/Jofairden/Jofairden_Head.cs
```diff
@@ -13,9 +13,9 @@ internal class Jofairden_Head : JofairdenArmorItem
 		}
 
 		public override bool IsVanitySet(int head, int body, int legs)
-			=> head == Mod.GetEquipSlot(nameof(Jofairden_Head), EquipType.Head)
-			&& body == Mod.GetEquipSlot(nameof(Jofairden_Body), EquipType.Body)
-			&& legs == Mod.GetEquipSlot(nameof(Jofairden_Legs), EquipType.Legs);
+			=> head == EquipLoader..GetEquipSlot(Mod, nameof(Jofairden_Head), EquipType.Head)
+			&& body == EquipLoader..GetEquipSlot(Mod, nameof(Jofairden_Body), EquipType.Body)
+			&& legs == EquipLoader..GetEquipSlot(Mod, nameof(Jofairden_Legs), EquipType.Legs);
```
steviegt6: `EquipLoader..` -> `EquipLoader.`

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,131 @@ internal static int GetPlayerEquip(Player player, EquipType type)
 				_ => 0,
 			};
 
+		/// <summary>
+		/// Adds an equipment texture of the specified type, internal name, and associated item to your mod.<br/>
+		/// You can then get the ID for your texture by calling EquipLoader.GetEquipTexture, and using the EquipTexture's Slot property.<br/>
+		/// Returns the ID / slot that is assigned to the equipment texture.
+		/// </summary>
+		///  <param name="mod">The mod the equipment texture is from.</param>
```
steviegt6:
```suggestion
		/// <param name="mod">The mod the equipment texture is from.</param>
```

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,131 @@ internal static int GetPlayerEquip(Player player, EquipType type)
...
+			return AddEquipTexture(mod, new EquipTexture(), item, type, texture);
+		}
+
+		/// <summary>
+		/// Adds an equipment texture of the specified type and internal name to your mod<br/>
+		/// (Use this if you don't want to associate an item with the texture)<br/>
+		/// You can then get the ID for your texture by calling EquipLoader.GetEquipTexture, and using the EquipTexture's Slot property.<br/>
+		/// Returns the ID / slot that is assigned to the equipment texture.
+		/// </summary>
+		///  <param name="mod">The mod the equipment texture is from.</param>
```
steviegt6:
```suggestion
		/// <param name="mod">The mod the equipment texture is from.</param>
```

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,131 @@ internal static int GetPlayerEquip(Player player, EquipType type)
...
+
+		/// <summary>
+		/// Adds an equipment texture of the specified type, internal name, and associated item to your mod.<br/>
+		/// This method is different from the other AddEquipTexture in that it takes in both an associated item and internal name at once.<br/>
+		/// If both are provided, the EquipTexture's name will be set to the internal name, alongside the keys for the equipTexture dictionnary.<br/>
+		/// NOTE : If multiple EquipTextures of the same type are registered for the same item, the first one to be added will be the one automatically displayed on the player and mannequins<br/>
+		/// All other parameters are the same as the other AddEquipTexture.
+		/// </summary>
+		/// <param name="equipTexture">The equip texture.</param>
+		/// /// <param name="mod">The mod the equipment texture is from.</param>
```
steviegt6:
```suggestion
		/// <param name="mod">The mod the equipment texture is from.</param>
```

steviegt6 (changes requested):
Formatting issues.

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,131 @@ internal static int GetPlayerEquip(Player player, EquipType type)
...
+		/// Adds an equipment texture of the specified type, internal name, and associated item to your mod.<br/>
+		/// You can then get the ID for your texture by calling EquipLoader.GetEquipTexture, and using the EquipTexture's Slot property.<br/>
+		/// Returns the ID / slot that is assigned to the equipment texture.
+		/// </summary>
+		/// <param name="mod">The mod the equipment texture is from.</param>
+		/// <param name="item">The item.</param>
+		/// <param name="type">The type.</param>
+		/// <param name="texture">The texture.</param>
+		/// <returns></returns>
+		public static int AddEquipTexture(Mod mod, ModItem item, EquipType type, string texture) {
```
Mirsario: This and the following methods (except the last one) could shorten their duplicate XML lines by using
```cs
/// <inheritdoc cref="AddEquipTexture(Mod, EquipTexture, ModItem, string, EquipType, string)"/>
```

and then only declaring parameters that are missing there.

❗  Also, to be discussed: Whether the Mod parameter is needed, while it could be taken from ModItem.

Mirsario (changes requested):
See stevie's comment, please make sure your branches are buildable and tested in-game.

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
IbanPlay: Mod parameter is required since part of the goal of this PR is to let the method be useable without the need of a ModItem reference, and making the Mod parameter be only needed in some of the overloads would be quite confusing. Did shorten the XML

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,106 @@ internal static int GetPlayerEquip(Player player, EquipType type)
...
+			if (!mod.loading)
+				throw new Exception("AddEquipTexture can only be called from Mod.Load or Mod.Autoload");
+
+			if (name == null && item == null)
+				throw new Exception("AddEquipTexture requires either an item or a name be provided");
+
+			ModContent.Request<Texture2D>(texture); //ensure texture exists
+
+			equipTexture.Texture = texture;
+			equipTexture.Mod = mod;
```
Solxanich: Echo Mirsario, the inclusion of Mod class here is not the cleanest. 
Suggest switching item.name for item.fullName so it aligns with other Loader practices

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -241,6 +241,106 @@ internal static int GetPlayerEquip(Player player, EquipType type)
...
+		/// All other parameters are the same as the other AddEquipTexture.
+		/// </summary>
+		/// <param name="mod">The mod the equipment texture is from.</param>
+		/// <param name="equipTexture">The equip texture.</param>
+		/// <param name="item">The item.</param>
+		/// <param name="name">The internal name.</param>
+		/// <param name="type">The type.</param>
+		/// <param name="texture">The texture.</param>
+		/// <returns>the ID / slot that is assigned to the equipment texture.</returns>
+		public static int AddEquipTexture(Mod mod, EquipTexture equipTexture, ModItem item, string name, EquipType type, string texture) {
```
Solxanich: Suggest mark Item, Name, and EquipTexture as optional and eliminate the above four helper methods.
Unnecessarily complicated, we don't need all these jumps.
Probably can just add a if(EquipTexture==null) then EquipTexture= new() if cant set the optional to default to new()

Solxanich (approved):
This is Merge-Ready assuming builds fine in CI.
Can merge when have Porting Notes.
I can't write up porting notes on my mobile, so might be a day or two unless added to PR/ping in discord

Issue #2266, opened 2022-04-10
Differentiate StarterBag and CanStack Bag Sprites


Setnour6:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

The tModLoader Starter Bag and the Example Mod CanStack bag appear too similar, and may confuse some players with what is what (you'll never know). I get that this feature isn't looked at, but I want to make that one tiny QoL change that might be better in the future (or however you perceive it).

In this proposal, I am only changing the CanStack item. All I did was try to make it appear slightly brighter so it can be slightly different.
If you want me to, I can just slap a big S on the sprite and maybe add a C next to it if you want me to.

_If you are reading this, please do not ask me to join the discord server, as I do not have discord and am not willing to make a discord account_

### What does this proposal attempt to solve or improve?

This improves overall viewability for what appears to be the same item, actually can be differentiated just by the different color tones and hues.
Say that for some reason, tooltips don't seem to appear (mod problem), and testing is conducted on the two items. The item that is the CanStack item will be a slightly different color, and can be better differentiated, so confusing one item for the other won't happen as often. Again, I can slap a big S on it, which stands for "Stack".

### Which (other) solutions should be considered?

Either I can solely change the color and tone of the sprite, add a big S on it (with a C next to it), or both.

Setnour6:
If y'all want something better, I can help work on that closed issue, ExampleCrossModItem, because I thought about doing it.

PR #2279, opened 2022-04-11
ExampleCanStackItem Resprite

Setnour6:
### What is the new feature?

This is a resprite of the ExampleCanStackItem.png item.

The tModLoader Starter Bag sprite and the ExampleCanStackItem both have the exact same sprite, and I want them to have a slightly different appearance as to not confuse them for one another.
This proposal improves overall viewability for what appears to be the same item, actually can be differentiated just by the different color tones and hues (and also text if you want me to add it).
Say that for some reason, tooltips don't seem to appear (mod problem), and testing is conducted on the two items. The item that is the CanStack item will be a slightly different color, and can be better differentiated, so confusing one item for the other won't happen as often. Again, I can slap a big S on it, which stands for "Stack".

There isn't any other already made alternative designs, but I can make them on request, such as a giant S in the middle of the sprite, or "CS" or "CanStack" written on it.

It's a sprite...

### ExampleMod updates

The CanStack sprite is part of Example Mod, so Example Mod will be updated.

### Other Notes
See tModLoader Issue/Request #2266 (Differentiate StarterBag and CanStack Bag Sprites)
This is Setnour6's first contribution and pull request for tModLoader 1.4. Feedback is appreciated.


JavidPack:
thanks.

Setnour6:
> thanks.

You're welcome!

PR #2284, opened 2022-04-13
Accept dots in hjson

Ishigh1:
Currently the hjson translation file doesn't handle dots in keys, because the jToken.Path tries to be clever by transforming "x.y" to "['x.y']". This PR solves this issue by recreating the Path method.
I did an handling of the arrays so that random message can be handled well, but didn't implement for the constructors because I didn't know what they were
This PR can also be considered as a way to have prefixes in hjson files as you can easily write it as the first line

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
@@ -223,10 +223,24 @@ namespace Terraria.ModLoader
 				// Parse JSON
 				var jsonObject = JObject.Parse(jsonString);
 				// Flatten JSON into dot seperated key and value
-				var flattened = jsonObject
-					.SelectTokens("$..*")
-					.Where(t => !t.HasValues)
-					.ToDictionary(t => t.Path, t => t.ToString());
+				var flattened = new Dictionary<string, string>();
+				foreach (JToken t in jsonObject.SelectTokens("$..*")) {
+					if (!t.HasValues) {
+						// Custom implementation of Path to allow "x.y" keys
+						string path = "";
+						JToken current = t;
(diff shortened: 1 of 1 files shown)


ghost:
maybe a little nonstandard in terms of HJSON and a little out of spirit of it perhaps, but tbh I don't see the harm in this and it can save a bit of space on top level prefixes as you mention.

unless anyone has any cases where this would break something (I hope not, please don't rely on the existing parsing of .s either)

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -223,10 +223,24 @@ internal static ModTranslation GetOrCreateTranslation(Mod mod, string key, bool
...
+								JProperty property => property.Name + (path == "" ? "" : "." + path),
+								JArray array => array.IndexOf(current) + (path == "" ? "" : "." + path),
+								_ => path
+							};
+							current = parent;
+						}
+
+						flattened.Add(path, t.ToString());
+					}
+				}
```
Mirsario: Line breaks, short-circuiting.
```suggestion
				var flattened = new Dictionary<string, string>();

				foreach (JToken t in jsonObject.SelectTokens("$..*")) {
					if (t.HasValues) {
						continue;
					}

					// Custom implementation of Path to allow "x.y" keys
					string path = "";
					JToken current = t;

					for (JToken parent = t.Parent; parent != null; parent = parent.Parent) {
						path = parent switch {
							JProperty property => property.Name + (path == string.Empty ? string.Empty : "." + path),
							JArray array => array.IndexOf(current) + (path == string.Empty ? string.Empty : "." + path),
							_ => path
						};
						current = parent;
					}

					flattened.Add(path, t.ToString());
				}
```

Mirsario (approved):

PR #2292, opened 2022-04-14
ExampleMod: ExamplePie.cs + ExamplePie.png

Lumelore:
Adds an example pie item. Code is commented and item uses the example palette. The example pie item shows how to make a food/drink item that can be placed on a plate. In addition it shows how to make it give multiple buffs when consumed.

No, they should not.


JavidPack:
Good job.

PR #2293, opened 2022-04-15
[1.4] Sprite Swap! Aka Swap ExampleCanStackItem and Starter Bag Sprites

Setnour6:
So I thought about the CanStack and Starter Bag sprites for a moment after I updated the CanStack item.
I had the feeling that maybe the resprite just feels off from how the Example Mod sprites usually are. They're grey and meant to be gray. Having the CanStack item be the only non-gray item seems off, now that I look at it.
This PR swaps both the CanStack and Starter Bag sprites, so now the Starter Bag has a lighter color, and the CanStack item now returns to it's original look, with a slight increase in light tones, so it's slightly more white. Of course, it's just a small thing and won't really be noticable.

### What is the new feature?

This is another resprite of the CanStack item, but also a resprite of the Starter Bag.

Seeing that the CanStack item is part of the example mod, and that it's one of the only non-grey "it can be grey and it pretty much was grey" items there, I decided to make a change to it again.
This change will restore originality to the sprites, and make it so the Starter Bag is brighter and slightly more appealing to look at (because it holds items and is part of tModLoader itself).

I can add an S or a C or CS text on the CanStack item sprite.

### ExampleMod updates

ExampleCanStackItem is getting a resprite again. (Sorry Javid)


Setnour6:
Whoops, wrong commit location...

Setnour6:
Well now I'm forced to close it and start a new PR

PR #2294, opened 2022-04-15
Fixed a vanilla bug with OGG audio track looping.

JamzOJamz:
### What is the bug?

Terraria 1.4 does not correctly implement the parsing/reading of .OGG file meta tags. In fact, the current code does not find them at all. Using music with loop points is impossible as long as this bug remains unfixed, which is undesirable.

### How did you fix the bug?

I changed one line in `OGGAudioTrack` to ensure meta tags are parsed correctly.

I've heard there's an instance of obsolete code inside of the `OGGAudioTrack` class, and changing it gives you the values directly without requiring any modification to the `TryGettingVariable` method. This fix, however, is the only one I found that managed to fix the problem.

### DIFF
--- a/patches/tModLoader/Terraria/Audio/OGGAudioTrack.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/OGGAudioTrack.cs.patch
@@ -1,11 +1,31 @@
 --- src/TerrariaNetCore/Terraria/Audio/OGGAudioTrack.cs
 +++ src/tModLoader/Terraria/Audio/OGGAudioTrack.cs
-@@ -57,7 +_,7 @@
+@@ -1,6 +_,7 @@
+ using Microsoft.Xna.Framework.Audio;
+ using NVorbis;
+ using System.IO;
++using System;
+ 
+ namespace Terraria.Audio
+ {
+@@ -57,8 +_,20 @@
  		}
  
  		private void TryGettingVariable(string vorbisComment, string variableWeLookFor, ref int variableValueHolder) {
--			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment, out int result))
-+			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment.Remove(0, variableWeLookFor.Length + 1), out int result))
++			// Vanilla: Code is faulty, tries to parse the whole comment as one integer.
++			/*
+ 			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment, out int result))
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Audio/OGGAudioTrack.cs.patch
```diff
@@ -0,0 +1,11 @@
+--- src/TerrariaNetCore/Terraria/Audio/OGGAudioTrack.cs
++++ src/tModLoader/Terraria/Audio/OGGAudioTrack.cs
+@@ -57,7 +_,7 @@
+ 		}
+ 
+ 		private void TryGettingVariable(string vorbisComment, string variableWeLookFor, ref int variableValueHolder) {
+-			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment, out int result))
++			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment.Remove(0, variableWeLookFor.Length + 1), out int result))
```
Mirsario, line 8:
```suggestion
-			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment, out int result))
+			if (vorbisComment.StartsWith(variableWeLookFor) && int.TryParse(vorbisComment.Substring(variableWeLookFor.Length), out int result))
```

Mirsario: Looks like this can technically throw ArgumentOutOfRangeException. What's the separator on these comments? Use IndexOf for it, continue if it's missing, use the index in a string.Substring call.

Mirsario (changes requested):

patches/tModLoader/Terraria/Audio/OGGAudioTrack.cs.patch
Mirsario: \*

JamzOJamz:
I’m sorry, I am new to Git pull requests. What do I have to do, or did you fix the solution yourself already?

Mirsario (approved):
Improved & tested everything now, thanks. Apparently VorbisReader.Comments is legacy, there was a warning suggesting .Tags.All instead, which is an easier to use dictionary.

As you're new to git:
In the future, if you'll be making any new PRs (even if not for TML), I suggest you to create separate branches for them, like `1.4_ogg_loop_points` as an example for this one. Makes things a bit easier both for you and everyone else.
Since you didn't do that, you may now need to hard reset your `1.4` branch to TML's. It's best to not push anything of your own there, and use it only as a template for new branches.

Issue #2239, opened 2022-04-02
Holding living leaf or wood wand makes wood rapidly drain from your inventory


AwhRyan:
### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When holding the leaf wand or living wood wand, wood rapidly drains from your inventory at about 50 per second without placing anything

### Log File

https://user-images.githubusercontent.com/102840279/161361150-45fba1aa-5af2-45b9-b31b-5b3e15cb6a00.mp4

### Steps to reproduce

Add wood into your inventory
Hold the living wood or leaf wand

### Expected Behavior

It shouldnt use wood until you click while using it

### Actual Behavior

Taking wood rapidly and seemingly deleting it.

### Reproduction frequency

Every time i use these wands.

Solxanich:
Looks like a duplicate of #1863  
Will close if is the case.

Issue #1863, opened 2021-09-22
[1.4] Block-Placing Wands Rapidly Deplete Blocks


DerpTaco99921:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

Holding a leaf wand after joining a world rapidly drains wood from your inventory.

### Steps to reproduce

1. Obtain wood and a leaf wand.
2. Save and exit.
3. Rejoin the world.
4. Hold the leaf wand.

### Expected Behavior

The wood in your inventory should only deplete by one when you use the leaf wand to place a block.

### Actual Behavior

The wood gets drained very quickly by just holding the wand.

### Reproduction frequency

100% of the time.

### Additional Information

This issue affects every type of block-placing wand (Bone Wand, Hive Wand, Leaf Wand, Living Wood Wand, Living Mahogany Wand, and Rich Mahogany Leaf Wand) with their respective block (Bone, Hive, Wood, Rich Mahogany Wood) that they use. 
When you obtain one of the wands, they function normally until you leave the world and rejoin with them in your inventory. The bug stops happening the moment you place a block with one of the wands, and then every wand starts working normally again.
I tested this no mods and the bug still occurs.

PR #2300, opened 2022-04-18
[1.4] Fix tile placing wands draining inventory

einarmo:
### What is the bug?

This PR fixes the bug in #2239, #1863, tile placing wands rapidly drains your inventory if you have not yet used anything when entering a world and select the wand.

### How did you fix the bug?

I added the condition `itemTimeMax != 0` to the check for removing items from wands.

This feels a little like a patch, but I don't really think it is. Every other comparison `itemTime == itemTimeMax` also checks `itemTimeMax != 0`. This particular line was changed in the item use rework to refer to `itemTimeMax`, so the logic here is not vanilla original.

Maybe? I don't know entirely why this was changed in the first place. The change seems reasonable, but it is clear to me that if you compare `itemTime` with `itemTimeMax` you also need to check `itemTimeMax != 0`.

As an aside, `dontConsumeWand` does nothing at all in vanilla. It is checked a few places, and repeatedly set to false, but the only place it is set to true is behind `if (0 == 0)`, no TML changes involved.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -4102,7 +4102,7 @@
  
  				if (i == Main.myPlayer) {
 -					if (!dontConsumeWand && itemTime == (int)((float)item.useTime * tileSpeed) && item.tileWand > 0) {
-+					if (item.tileWand > 0 && !dontConsumeWand && itemTime == itemTimeMax) {
++					if (itemTimeMax != 0 && item.tileWand > 0 && !dontConsumeWand && itemTime == itemTimeMax) {
  						int tileWand = item.tileWand;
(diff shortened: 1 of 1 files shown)


Ishigh1:
We already talked with Mirs about it, unsure if it's planned for another PR but if this PR is merged, it should at least also include a fix to the 2 other occurrences of the same issue, with this same check

einarmo:
I can't find any other references to `itemTimeMax` without this check or something equivalent. I didn't find another PR addressing this issue, but if there is one feel free to close this.

Ishigh1:
It was just a discussion on Discord, and it seems the two others were indeed fixed

Solxanich:
Fixes #2239 
Fixes #1863 

@ThomasThePencil  verified the change

PR #2307, opened 2022-04-20
[1.4] xID.TML IDs to names

Antirhinnum:
### What is the new feature?
Changes most of the constants in the xID(.TML).Sets classes to use their appropriate IDs.
One set (`ArmorIDs(.TML).Legs.Sets.HidesBottomSkin`) is not changed because the two values in the set have no names in ArmorIDs.Legs.
Some `TileID` sets explicitly prefix some constants with "`TileID.`" because a set shares a name with the constant and not prefixing causes the set to be used (Ex: `TileID.Sets.Stone` and `TileID.Stone`).

So the xID.TML.cs files are slightly more readable.

Do the same thing for the normal xID file Sets. Didn't do in this PR since it would patch the entirety of every `Sets` class, which I didn't think would be appreciated.
If it's wanted, I can also run through the non-TML ID files and apply the same changes I've already done to the TML ones.

N/A

### ExampleMod updates
N/A

### DIFF
--- a/patches/tModLoader/Terraria/ID/ArmorIDs.TML.cs
+++ b/patches/tModLoader/Terraria/ID/ArmorIDs.TML.cs
@@ -13,18 +13,18 @@ namespace Terraria.ID
 			partial class Sets
 			{
 				// Created based on 'fullHair' definition in 'Player.GetHairSettings'.
-				public static bool[] DrawFullHair = Factory.CreateBoolSet(10, 12, 28, 42, 62, 97, 106, 113, 116, 119, 133, 138, 139, 163, 178, 181, 191, 198, 217, 218, 220, 222, 224, 225, 228, 229, 230, 232, 235, 238, 242, 243, 244, 245, 272);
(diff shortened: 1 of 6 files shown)


JavidPack:
So, we probably won't be merging this. The maintenance cost is too high, and it actually adds a step for anyone researching a vanilla item in the source code. We've discussed in the past writing a special patch step to do this automatically through syntax rewriter algorithms, but we haven't been able to get it working 100% yet. Anything less than 100% will just add work to anyone working on anything.  (What I mean by that is they'll have to search for both `PlatinumCrown` and `51` throughout the whole solution to follow the adaption guide.)

Antirhinnum:
Understood. Thank you for the response.

JavidPack:
We discuss potential PRs frequently in the #collaborators channel on our discord. It might be useful to pop in and get some pre-feedback before spending a lot of effort on a future PR, if there is something else you are interested in contributing later. 

I'm sorry about this, it was well done and probably took a good deal of time to make all these changes.

PR #2311, opened 2022-04-21
[1.4] town NPC profile system but it's done right this time

ThomasThePencil:
## IN WHICH
I may or may not have redone this feature already. I hope that's cool
technically a bug fix PR but I kinda changed the system semi-significantly again in the process so

## why does this exist...again?
because the system I was originally in a rush to get onto preview sucked and had a very prominent crash (and general weird behavior) with modified vanilla profiles

## changelog
- changed `SetTownNPCProfile` into two different hooks
  - for `ModNPC`s, it's now the `TownNPCProfile` hook, allowin' for easy addition of a profile or set of conditional profiles to your modded NPC of choice
  - for `GlobalNPC`, it's now the `ModifyTownNPCProfile` hook, and rather than takin' the original dictionary as a parameter, it now takes the profile itself, meanin' the original dictionary no longer gets changed
- added a method to the unload process which forcibly resets all NPC textures modified by profiles, to catch any not-unloaded profile shenanigans and fix Claire's Crash (detailed below)

## bug fixes (all tML)
- the main point of this (hence the branch name) was to fix an issue I've affectionately dubbed "Claire's Crash" (in reference to the Princess revamp I've been testin' this against), in which loadin' into a world after replacin' a vanilla NPC's profile with one that calls different textures, exitin' the world, and then reloadin' mods would result in a crash
- this generally also fixes a lot of other funkiness with town NPC profiles, although I've made sure to include some extra stuff in ExampleMod juuuust in case people don't quite get the memo

## portin' notes
- `public override void ModNPC.SetTownNPCProfile(Dictionary<int, ITownNPCProfile> database)` => `public override ITownNPCProfile ModNPC.TownNPCProfile()`
- `public override void GlobalNPC.ModifyTownNPCProfile(NPC npc, Dictionary<int, ITownNPCProfile> database)` => `public override void GlobalNPC.ModifyTownNPCProfile(NPC npc, ref ITownNPCProfile profile)`


ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -345,14 +343,16 @@ public class ExamplePersonProfile : ITownNPCProfile
 		public int RollVariation() => 0;
 		public string GetNameForVariant(NPC npc) => npc.getNewNPCName();
 
+		public Asset<Texture2D> GetTextureNPCShouldUse() => ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson", AssetRequestMode.ImmediateLoad);
+
```
direwolf420:
```suggestion
		public Asset<Texture2D> GetTextureNPCShouldUse() {
			return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson", AssetRequestMode.ImmediateLoad);
		}

```

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -158,10 +158,9 @@ public abstract class ModNPC : ModTexturedType
 
 		/// <summary>
 		/// Allows you to set the town NPC profile that this NPC uses.
+		/// By default, returns null, meaning that the NPC doesn't use one.
 		/// </summary>
-		/// <param name="database">The list of town NPC profiles that currently exist.</param>
-		public virtual void SetTownNPCProfile(Dictionary<int, ITownNPCProfile> database) {
-		}
+		public virtual ITownNPCProfile TownNPCProfile => null;
```
direwolf420: Is there a reason this is a property and not a method? Clashes with `GetTextureNPCShouldUse` which is a method

direwolf420 (commented):
minor

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
ThomasThePencil: originally, I'd figured it'd be more convenient for it to be a property, but now that I look at it, havin' it be a method will probably be more convenient in the event somebody wants to do something like use different profiles for an NPC dependin' on progression. also just helps keep things consistent

ExampleMod/Content/NPCs/ExamplePerson.cs
ThomasThePencil: funny story, I had actually meant to remove this, since there's no longer a parameterless overload of `GetTextureNPCShouldUse` in `ITownNPCProfile`, but I must've forgotten to. thanks for pointin' this out

ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -347,12 +347,12 @@ public class ExamplePersonProfile : ITownNPCProfile
...
 			if (npc.IsABestiaryIconDummy && !npc.ForcePartyHatOn)
-				return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson");
+				return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson", AssetRequestMode.ImmediateLoad);
 
 			if (npc.altTexture == 1)
-				return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson_Party");
+				return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson_Party", AssetRequestMode.ImmediateLoad);
 
-			return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson");
+			return ModContent.Request<Texture2D>("ExampleMod/Content/NPCs/ExamplePerson", AssetRequestMode.ImmediateLoad);
```
Mirsario: ImmediateLoad shouldn't be anything required or expected. Use `asset.Wait?.Invoke()` calls instead.

Mirsario (changes requested):

Mirsario (approved):

PR #2312, opened 2022-04-22
Workshop tags finally, but Part 1 of implementation 

Solxanich:
This PR initialize Workshop Tags that properly work.

The templated tags can be revised & expanded upon as needed.

This is not a complete implementation, and should be viewed as a basic starting point to be expanded upon later.


patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -34,23 +45,13 @@ public partial class WorkshopSocialModule
 				return false;
 			}
 
-			if (BuildInfo.IsDev) {
+			if (false && BuildInfo.IsDev) {
```
Solxanich:
```suggestion
			if (BuildInfo.IsDev) {
```

Issue #2096, opened 2022-01-15
Add a `Clone` method to Terraria.Recipe


Soaryn:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Provide a new method in Terraria.Recipe named `Clone()` to provide a new instance of the same Recipe with all of the data copied to the new recipe. 

**Example:**
New item is introduced.
We want to loop over every recipe (added by mods or vanilla) and add a new recipe when a specific material is used.
This new recipe would be modified as such:
  - The specific material would be removed from the ingredients.
  - The new item we introduced would be added as an ingredient with a different stack size.
  - A new call back would be added to modify how the new item would be consumed

Key feature to note, is that the original recipe is still intact.
Something that needs to be addressed if this is added, is the ability to also remove callbacks, as `.Register` adds the Alchemy Callback which would break if registering a pure clone.

### What does this proposal attempt to solve or improve?

The intent is to allow new recipes to be added based on exact copies of other recipes. Say that you add a new item that has a special call back without modifying the original recipe. This item would work for similar recipes given a specific material.

Currently in order to create a copy of a new recipe, you can do so manually to some extent, but lack some critical information such as  `ConsumeItemHooks` or `OnCraftHooks` which would affect recipes such as Alchemy or other modded call backs.

### Which (other) solutions should be considered?

Opening up the call backs used in Recipe to not be internal. This, however, could open up new problems with mod compatibility, nor does it prevent other fields from being missed.

Solxanich:
Closed per #2314 and confirmation from Dire

PR #2314, opened 2022-04-22
[1.4] [Feature] Clone method for Terraria.Recipe

rjmarzec:
### What is the new feature?
Addressing feature request #2096, this feature adds cloning using the new ```Clone()``` method in ```Terraria.Recipe``` as an alternative to creating recipes using ```Create()```. The clone copies over the same properties of the original recipe (such as ```createItem```, ```requiredItem```, ```requiredTile```, and others, but also notably including ```ConsumeItemHooks``` and ```OnCraftHooks``` which can be otherwise hard for modders to access) while leaving the original recipe fully intact. 

As noted in #2096, recipes can currently be copied manually, but manually copied recipes may not share the same ```ConsumeItemHooks``` or ```OnCraftHooks``` as their originals, which can cause issues for unique modded call backs. Additionally, when creating large numbers of similar copies of complex recipes, it is much cleaner to simply make clones and a few small adjustments to each instead of needing to rebuild every property manually when creating a new recipe.

As proposed in #2096, one alternative is opening up call backs used in ```Terraria.Recipe``` to not be internal, but this could lead to other problems with mod compatibility on top of still not preventing other fields from being missed.

### Sample usage for the new feature & ExampleMod updates
A sample usage can be found from lines 93 to 121 of ```ExampleMod/Content/ExampleRecipes.cs```. In the sample, a new recipe is made by copying another recipe and adding/removing ingredients/tiles/conditions while still leaving the original intact. The end result is both recipes existing in game simultaneously without interference.
 
Original Recipe:
original recipe

Copied Recipe (with modified ingredients and conditions):
copied recipe

### Notes
This is my first non-trivial contribution to TML, so it's likely I broke some common practice/standard in the process. I recommend looking over my changes closely to make sure nothing looks too funky. Any feedback is super helpful and appreciated!

### DIFF
--- a/ExampleMod/Content/ExampleRecipes.cs
+++ b/ExampleMod/Content/ExampleRecipes.cs
@@ -90,6 +90,35 @@ namespace ExampleMod.Content
 
 				// When you're done, call this to register the recipe. Note that there's a semicolon at the end of the chain.
 				.Register();
+
+			///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+			// The following recipe showcases and explains cloning recipes and how they can modified to differ from the oringinal recipes they came from. //
+			///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+			// If you are making multiple recipes that are very similar with only a few differences or want multiple recipes to make use of the same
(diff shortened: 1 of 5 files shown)


JavidPack:
Good job, I made some small changes for various reasons, feel free to ask about any of them if you are curious why they were needed.

PR #2318, opened 2022-04-24
Swap ExampleCanStackItem and Starter Bag Sprites

Setnour6:
I made a huge mistake with the previous PR I made, so I closed it and made this one, again.
See PR (#2293) for more information.

### What is the new feature?

See PR (#2293)

See PR (#2293)

See PR (#2293)

### ExampleMod updates

ExampleCanStackItem is getting a resprite again. (Sorry Javid)
See PR (#2293)


Chicken-Bones:
I think this may be a misunderstanding. While I think both sprites could be improved the "Start Bag" is not part of ExampleMod. It is a tML feature, which is always present. When your combination of mods provides starting equipment which fills your entire inventory, the overflow is packed into a 'start bag' by tML.

For this reason, it should not follow the example mod colour scheme, and instead should look more 'part of vanilla'

Setnour6:
> I think this may be a misunderstanding. While I think both sprites could be improved the "Start Bag" is not part of ExampleMod. It is a tML feature, which is always present. When your combination of mods provides starting equipment which fills your entire inventory, the overflow is packed into a 'start bag' by tML.
> 
> For this reason, it should not follow the example mod colour scheme, and instead should look more 'part of vanilla'

Yes, but in a previous PR, I changed the sprite for the canstack item only, which was part of Example Mod.
Example Mod canstack item will go back to having a grey bag, and the starter bag will have the new white bag sprite.

PR #2321, opened 2022-04-26
NPC-capture-related hook expansion

ThomasThePencil:
# IN WHICH
I make a system for something you don't really think about unless you're actually designin' something that builds off of the idea

## why does this exist?
well, I'll be entirely honest with you: the main reason I looked into this was because I wanted to make a new type of bug net allowin' the capture of pixies if it specifically is used, as a corollary to the ideas introduced by 1.4's new Underworld critters. I then wondered if people that *don't* have a freely-accessible full-scale `ItemCheck` detour would want to do this sort of thing with their own modded content, and so I made actual hooks for this stuff

## changelog
- added the `CatchingTool` and `LavaproofCatchingTool` sets to `ItemID` --- the first allows you to define any given item as bein' a tool to catch NPCs with, while the second allows that item to automagically be able to catch the Underworld's lava critters
- added the `CanCatchNPC` hook to `ModPlayer`, `ModItem`, and `GlobalItem`, and the companion hook `CanBeCaughtBy` to `GlobalNPC` and `ModNPC`. these hooks allow the conditions under which a catch can happen to be controlled much more effectively, and from multiple places
- made several changes to the `OnCatchNPC` hook, previously exclusive to `ModNPC` and `GlobalNPC`:
  - modified the call location and intention of the hook itself: it is now designed to handle on-catch effects, rather than deal with the resultin' item (don't worry! this use case is still covered; keep readin' to learn more)
  - renamed the NPC variants of the hook to `OnCaughtBy` for consistency with `CanBeCaughtBy`
  - extended it to `ModPlayer`, `ModItem`, and `GlobalItem` for additional versatility/accessibility
  - made its various iterations give much more useful context
- added the `NPC.CheckCatchNPC` method as a way for modders to easily give non-standard NPC-catchin' tools (e.g. those that use projectiles instead of items) the ability to check for NPC catches
- added a new, unique entity source for catchin' NPCs, called `EntitySource_CatchEntity`. through the existence of `OnSpawn`, this completely fills the role that `OnCatchNPC` previously filled, and then some extra possibilities on top of that
- upgraded the `catchItem` NPC field and `makeNPC` Item field from `short`s to `int`s for convenience and consistency --- `NPC.DefaultToCapturedCritter` has been modified accordingly

## portin' notes
- `NPC.CatchNPC` is now internal. to check for NPC catchin' manually (e.g. for a projectile that catches NPCs), use `NPC.CheckCatchNPC` instead
- `GlobalNPC.OnCatchNPC(NPC npc, Player player, Item item)` => `GlobalNPC.OnCaughtBy(NPC npc, Player player, Item item, bool failed)`
  - `item` is now the item used to capture the NPC, rather than the item spawned on a successful catch
  - to modify the spawned item, check if the source `is EntitySource_CatchEntity` in `OnSpawn` instead
- `ModNPC.OnCatchNPC(Player player, Item item)` => `ModNPC.OnCaughtBy(Player player, Item item, bool failed)`
  - `item` is now the item used to capture the NPC, rather than the item spawned on a successful catch
  - to modify the spawned item, check if the source `is EntitySource_CatchEntity` in `OnSpawn` instead
- `npc.catchItem` and `item.makeNPC` are now `int`s instead of `short`s, and `item.DefaultToCapturedCritter` now takes an `int` instead of a `short`. adjust your critters accordingly


patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
```diff
@@ -154,5 +154,58 @@ public static class CombinedHooks
...
+				return canCaptureWithOverall.Value;
+			else
+				return item.type == 1991 || item.type == 3183 || item.type == 4821;
+		}
+
+		public static bool? CanCapture(Player player, NPC npc, Item item) {
+			bool? canCaptureOverall = null;
+			bool? canCaptureOnPlayer = PlayerLoader.CanCapture(player, npc, item);
+			bool? canCaptureOnItem = ItemLoader.CanCapture(item, npc, player);
+			bool? canCaptureOnNPC = NPCLoader.CanBeCapturedBy(npc, item, player);
```
ExterminatorX99: Should these be called after the early return checks?

patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
```diff
@@ -154,5 +154,58 @@ public static class CombinedHooks
 			}
 			return ret;
 		}
+
+		public static bool CanCaptureWith(Player player, Item item) {
+			bool? canCaptureWithOverall = null;
+			bool? canCaptureWithPlayer = PlayerLoader.CanCaptureWith(player, item);
+			bool? canCaptureWithItem = ItemLoader.CanCaptureWith(item, player);
```
ExterminatorX99: Should these be called after the early return checks?

patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
ThomasThePencil: probably, if purely for performance reasons. will do this

patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
ThomasThePencil: probably, if purely for performance reasons. will do this

direwolf420:
EDIT: approach 2 was chosen.

Naming consistency thing:
Current methods:
NPC.**Catch**NPC (vanilla)
Mod/GlobalNPC.On**Catch**NPC (tml)
Mod/GlobalItem/ModPlayer.Can**Capture**(With) (PR)

Notice the "catch" vs "capture" discrepancy
I propose two solutions to this:
1. Change the On**Catch**NPC to On**Capture**NPC so that all tml hooks are consistent (requires porting notes)
2. Change Can**Capture**(With) to Can**Catch**(With) so that all tml hooks + vanilla are consistent

Vanilla is inconsistent with it too: "capture" and "catch" are both used in the game and wiki (see Bug Net wiki page, it has both words in it). 
Food for thought.

direwolf420:
Also, I have a mod that implements a projectile (tied to an item, as it acts as a spear) that calls NPC.CatchNPC, with this change I would have to call the PR hooks myself, as CombinedHooks.CanCapture wraps around the call of NPC.CatchNPC.
I just noticed with 1.4 I would *also* have to copy the code with IsLavaBait in it into my mod aswell. Is there anything that can be done on tmls side to make it easier for me to make my own non-item CatchNPC basically? Maybe a helper method in NPC.TML.cs that wraps the stuff inside the for loop (after the continue) into a method that takes in `Player` and `flag` ("lava proof tool"), and  the item used to catch it with (in my case the held item while the projectile is out), helping resolve your todo. Then modders can call that instead of NPC.CatchNPC, and you can make NPC.CatchNPC internal

PR #2324, opened 2022-04-28
[1.4] Potion Sickness duration hook

ExterminatorX99:
### What is the new feature?
`ApplyPotionDelay` hook in `ModItem`, `GlobalItem` and `ModPlayer` which allows modifying the duration of the potion sickness debuff

Allows creation of items with functionality similar to that of Terraria's `Philosopher's Stone`

None that I can think of

```cs
// ModItem
public override void ApplyPotionDelay(Player player, bool quickHeal, ref int potionDelay)
{
	if (!quickHeal)
	{
		potionDelay = 300; // 5 seconds
	}
}

// ModPlayer
public override void ApplyPotionDelay(Item item, bool quickHeal, ref int potionDelay)
{
	if (someBool)
	{
		potionDelay /= 2; // Half as long
	}
}
```

### ExampleMod updates
`ExampleHealingPotion` uses the new hook


ExterminatorX99:
Better idea proposed in collaborators chat

Issue #1770, opened 2021-08-22
[1.4] Feral Claws autoswing locks player facing direction


DarioDaF:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

*Feral Claw* only autoswing keeps the player and weapon facing the direction of the first use

### Log File

Not applicable

### Steps to reproduce

Get a *Lead Broadsword* and equip *Feral Claws*.
Start to autoswing and try to move around

### Expected Behavior

Player switch face direction to the direction of motion **on use frame** (not at any moment)

### Actual Behavior

Player and weapon don't change facing direction even when moving

### Reproduction frequency

100%

### Additional Information

Player facing behaviour:
- Vanilla
  - *Terra Blade*: Facing the mouse cursor
  - *Lead Broadsword* with *Feral Claws*: Player motion direction
  - *Lead Broadsword*: Player motion direction

- tModLoader
  - *Terra Blade*: Facing the mouse cursor
  - *Lead Broadsword* with *Feral Claws*: Player motion direction **on first swing (no changing when moving around or moving cursor)**
  - *Lead Broadsword*: Player motion direction

JavidPack:
I can confirm this still happens.

Issue #1767, opened 2021-08-21
Spear's hitoboxes and particle effects remain when autoswinging


TrumanHanson:
When ever I use a spear with autoswing, even with all mods disabled, it leaves behind particle effects and hitboxes from All previous swings since I started swinging the spear, this is blatantly obvious when using a spear with lot's of particles such as the dark lance, as shown in this picture [https://drive.google.com/file/d/1mnMdmMXUU3Ggd5cYDxUJrkUCIfX4cMSY/view?usp=sharing](url)

Chicken-Bones:
Fixed in #2351

Issue #1654, opened 2021-08-01
[1.4 Alpha] Whips being drawn/rendered multiple times on auto-swing


RighteousRyan1:
### Description
Looks like whips are drawn multiple times if you hold down attack and auto-reuse them.

### Steps to reproduce
Just grab anything that makes you autoswing a whip, then hold down attack for multiple swings.

### Expected behavior 
It should not be getting drawn twice.

### Actual behavior
Nothing gameplay-like is affected, but it's a strange visual bug.

### Reproduction frequency
100%

Blookiwi:
It's not just a visual bug.
The duplicated whips actually deal damage, and it can get a bit broken.

Mirsario:
If this is related to 1.4_useitemfix, it should be real damaging projectiles, yes.

Chicken-Bones:
Fixed in #2351

PR #2329, opened 2022-04-30
[1.4] various post-item use rework fixes

ThomasThePencil:
# IN WHICH
I fix a number of issues related to the great item use rework of...how long ago was it, again?

either way, this is meant as a long-needed follow-up PR to #1437 

## it's all tML bug fixes, ladies and gents
- whips no longer start to duplicate themselves after autoswingin' for a short amount of time
(solves #1654)
- spears no longer build up more hitbox as they're used when in possession of autoswing
(solves #1767)
- melee weapons no longer have their hitbox linger weirdly near an enemy for a single frame
(solves #1721)

## and now, I'll address the Beholster in the room: Feral Claws interactions
(and conditional auto-reuse overall)
this solves #1770 --- by provin' that it's not an issue in the first place, but rather a fix to an issue that was already there

assume I have an Iron Broadsword and a Lead Broadsword. the Iron one has `autoReuse` set to true, while the Lead one has `autoReuse` set to false. 

### vanilla behavior
**Iron, no Feral:** locked into initial swing direction
**Iron w/ Feral:** locked into initial swing direction
**Lead, no Feral:** swing direction decided per-swing
**Lead w/ Feral:** swing direction decided per-swing despite havin' autoswing; inconsistent with pre-established broadsword autoswing mechanics

### stock tML behavior, post-item use rework
**Iron, no Feral:** locked into initial swing direction
**Iron w/ Feral:** locked into initial swing direction
**Lead, no Feral:** swing direction decided per-swing
**Lead w/ Feral:** locked into initial swing direction

notice that in both cases, the Iron Broadsword locks you into your initial swing direction regardless of whether or not you have Feral Claws equipped. this is how broadswords without `useTurn` set to true have always behaved with autoswing, both in vanilla and in tML --- it's only with 1.4's honestly wonky implementation of conditional autoswing bonuses that a third option became plausible in the first place. the item use rework essentially silently fixed this issue, only for some to mistakenly view it as a fully intended feature and cry out for its return. the case demonstrated with vanilla Lead w/ Feral has never been how these sorts of cases are supposed to be handled, which means tML has actually made this more consistent and thus increased intuitivity for modders and players alike --- and, as such, the issue here isn't actually an issue at all. on top of that, returnin' the behavior described here as an actual feature would require enough work, and would fly in the face of vanilla's established rules enough, to where it's genuinely better if a mod takes on that task. I rest my case

## what about alternate fixes for the stuff that's actually fixed?
the spear and whip fixes are tentative; **ideally, they'd be done with a more notable fundamental change to how they're handled on reuse**, but that's not what I came here to do and other people are likely to have better ideas on how to do that which I can't think of off of the top of my head. I simply aim to fix these issues well enough that we stop gettin' bug reports about things that can be tentatively fixed now and properly fixed later

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -2382,15 +2382,6 @@
  			}
  			else {
  				if (jumpBoost) {
-@@ -14215,7 +_,7 @@
- 						direction = -1;
- 				}
- 				else if ((itemAnimation == 0 || inventory[selectedItem].useTurn) && mount.AllowDirectionChange) {
--					direction = -1;
-+					direction = -1 ;
- 				}
- 
- 				if (velocity.Y == 0f || wingsLogic > 0 || mount.CanFly()) {
 @@ -14330,7 +_,7 @@
  				if (flag4)
  					num5 = 30;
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Superceded by #2351

PR #2355, opened 2022-05-10
JIT mods on load and allow extending from types in weak-refs

Chicken-Bones:
### What is the new feature?

JIT all methods from mods at load time, with attributes and a custom handler to specify what not to JIT. 

Adds two key attributes to allow interacting with weak-refs.
`[JitWhenModsEnabled(...)]` excludes a class/method/property from the load time JIT. Use this on any member which directly references a type from a mod that may not be present.

`[ExtendsFromMod(...)]` excludes a class from JIT and also from autoloading, making it now possible to have content which directly inherits from another mod, or only loads when that mod is present. The method `AssemblyManager.GetLoadableTypes(Assembly)` has been made public as a replacement for `Assembly.GetTypes()` if you want to do your own autoloading.

If a mod fails to JIT some methods, an error message is shown during load (example below).
```
An error occurred while loading ExampleModDepTest
The mod will need to be updated to match the current tModLoader version, or may be incompatible with the version of some of your other mods. Click the 'Open Web Help' button to learn more.
The mod(s) have been automatically disabled.

Terraria.ModLoader.Exceptions.JITException: 
In ExampleModDepTest.ExampleModCalls.method, Could not load type 'ExampleMod.Common.Configs.ExampleModConfig' from assembly 'ExampleMod, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
In ExampleModDepTest.ExampleModCalls..ctor, Could not load type 'ExampleMod.Common.Configs.ExampleModConfig' from assembly 'ExampleMod, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
In ExampleModDepTest.SeparateClass..ctor, Could not load type 'ExampleMod.Content.Items.ExampleItem' from assembly 'ExampleMod, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
In ExampleModDepTest.ExampleModCalls.get_property, Could not load type 'ExampleMod.Content.Items.ExampleItem' from assembly 'ExampleMod, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
In ExampleModDepTest.SeparateClass.method, Could not load type 'ExampleMod.Common.Configs.ExampleModConfig' from assembly 'ExampleMod, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

The PreJIT finds missing type/method/field errors which occur when a mod has been compiled against an old/different version of tML or a weak reference, before they cause a hard to detect/diagnose error at runtime which may be swallowed by a try-catch.

The `ExtendsFromMod` implementation currently uses `MetadataLoadContext`. This is slow because all assemblies in the hierarchy of every type in the mod must be loaded separately into the `MetadataLoadContext` for each mod, with no delegation and no way of resolving `BaseType` or `Interfaces` as strings instead of as full types. 

The actual performance impact has not been profiled, but imagine it might add a couple of seconds of load time if you have 20+mods on a slow PC?

A mod which has ExampleMod as a weak-ref:

```cs
using ExampleMod.Common.Configs;
using ExampleMod.Content.Items;

namespace ExampleModDepTest
{
    public class ExampleModCalls
    {
        [JITWhenModsEnabled("ExampleMod")]
        public static int property => ModContent.ItemType<ExampleItem>();


        [JITWhenModsEnabled("ExampleMod")]
        public ExampleModCalls()
        {
            ModContent.GetInstance<ExampleModConfig>().ExampleWingsToggle = true;
        }

        [JITWhenModsEnabled("ExampleMod")]
        public static void method()
        {
            ModContent.GetInstance<ExampleModConfig>().ExampleWingsToggle = true;
        }
    }

    [JITWhenModsEnabled("ExampleMod")]
    public class SeparateClass
    {
        public ExampleItem item = ModContent.GetInstance<ExampleItem>();

        public static ExampleModConfig method() => ModContent.GetInstance<ExampleModConfig>();
    }

    [ExtendsFromMod("ExampleMod")]
    public class ExampleItemSubclass : ExampleItem
    {

    }
}
```

## If you need more control
You can create a custom subclass of `MemberJitAttribute` or...
Set `Mod.PreJITFilter` to a custom override. Note that the base implementation of `ShouldJIT` checks for `MemberJitAttribute`
(and yes, `Type` extends `MemberInfo`)

```cs
public class PreJITFilter
{
    public virtual bool ShouldJIT(MemberInfo member) => ...
}
```


Chicken-Bones:
Note it is still not possible to extend from mod-types in another weak-referenced mod.

Setnour6:
Would the following in public class ExampleModCalls work instead?
```cs using ExampleMod.Common.Configs;
using ExampleMod.Content.Items;

namespace ExampleModDepTest
{
    public class ExampleModCalls
    {
        [JITWhenModsEnabled("ExampleMod")]
        public static int property => ModContent.ItemType<ExampleItem>();

        public ExampleModCalls()
        {
            ModContent.GetInstance<ExampleModConfig>().ExampleWingsToggle = true;
        }

        public static void method()
        {
            ModContent.GetInstance<ExampleModConfig>().ExampleWingsToggle = true;
        }
    }

    [JITWhenModsEnabled("ExampleMod")]
    public class SeparateClass
    {
        public ExampleItem item = ModContent.GetInstance<ExampleItem>();

        public static ExampleModConfig method() => ModContent.GetInstance<ExampleModConfig>();
    }

    /*
    [Autoload(false)]
    [JITWhenModsEnabled("ExampleMod")]
    public class ExampleItemSubclass : ExampleItem
    {

    }*/
}
```
Or does each entry have to be seperated by  [JITWhenModsEnabled("ExampleMod")]?

Chicken-Bones:
@Setnour6 The `ExampleModCalls` constructor and `ExampleModCalls.method` would fail at load time when "ExampleMod" is not present in your code there. As you can see in `SeparateClass` putting the attribute on the class exludes all its members from jitting, otherwise you must mark each method or property which references a type from an assembly that may not be present individually.

Setnour6:
> @Setnour6 The `ExampleModCalls` constructor and `ExampleModCalls.method` would fail at load time when "ExampleMod" is not present in your code there. As you can see in `SeparateClass` putting the attribute on the class exludes all its members from jitting, otherwise you must mark each method or property which references a type from an assembly that may not be present individually.

Ohhh okay. I see it now. Thank you!

PR #2356, opened 2022-05-10
The Grand French Localization Update

Setnour6:
Updates French Translation by a huge amount, to where every line here is french. I need people that know french to review this, because my basic French does not cut it. So far, Ishigh1 reviewed it.

### What did you just do?
I updated the French Translations and updated every single entry.
### Why should this be part of tModLoader?
See #2353.
### Are there alternative designs?
_no response_
### Sample usage for the new feature
_no response_
### ExampleMod updates

_no response_

PR #2362, opened 2022-05-10
The Actual Grand French Localization Update

Setnour6:
Fixed an issue where the French translations didn't apply. I failed to notice it for so long, even though it was right in front of my eyes the whole time.
I am embarrassed at myself, but it's okay, because it's fixed now.

See PR #2356 for information regarding the localization update.


Setnour6:
@Solxanich So I tested the new preview branch, and the French translations didn't apply. This PR fixes it.

Issue #2313, opened 2022-04-22
Hardmode Players no longer end up in recycle bin


JavidPack:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Previously, people could recover a hardmode player by checking the recycle bin on windows, but now we delete them directly.

We should probably change the preprocessor directive here for a runtime check, FileUtilities.cs. Whoever fixes this needs to verify the issue and the fix, as well as check for other similar situations. 
```
#if WINDOWS
				FileOperationAPIWrapper.MoveToRecycleBin(path);
#else
				File.Delete(path);
#endif
```

PR #2367, opened 2022-05-11
[1.4] Switch from preprocessor to runtime os checking for file deletion

miarodri:
### What is the bug?

#2313 : Hardmode Players no longer end up in recycle bin

### How did you fix the bug?

Switch from using preprocessor `#if WINDOWS` to runtime OS checks. Doing so also revealed a `System.AccessViolationException` when attempting to delete the files. According to https://stackoverflow.com/a/17648641, the struct should be aligned but not packed. Removing `Pack = 1` allowed deletion to work properly.

Not many. Specifically since we want to delete to the recycle bin without showing a dialog, we can't rely on the .NET function `FileSystem.DeleteFile` as it _will_ show error dialogs. We need to rely on the existing `FileOperationAPIWrapper.cs` for this behavior. `SHFILEOperation` is deprecated as of Windows Vista, but there's no need to expand this PR by updating it.


JavidPack:
Looks great, thanks!

PR #2373, opened 2022-05-12
[1.4, quickfix] Stop projectiles from sending an extra byte if extra AI is empty

Ouroel:
### What is the bug?

ProjectileLoader.SendExtraAI currently sends an extra unread byte which usually marks the number of elements in the array that is unread when bb22[2] is false. This removes the sent byte.

### How did you fix the bug?

Made NetMessage.SendData check for bb22[2] before sending the byte.

No.


direwolf420:
Doublechecked it, looks correct :heavy_check_mark:

Mirsario:
Thanks. Ignore my dumb edits, had to have a follow up commit - bec1398.

PR #2423, opened 2022-05-24
The Not-As-Grand Sound Playback Cleanup Follow-Up

ThomasThePencil:
# IN WHICH
I shamelessly knock off the name of the PR this is a follow-up to (#1084) and fix a number of sound issues that've come up in the wake of the sound overhaul

### changes
- made `ModBlockType.HitSound` default to `SoundID.Dig` rather than `null`

### bug fixes
- fixed Ivy, Rain Song, Stellar Tune, and Drum Set playin' various non-instrument sounds instead of the sounds they should be playin'
- fixed Bell, (Magical) Harp, and The Axe not properly obeyin' vanilla music pitch rules (no pitch variance, lower volume)
  - note to self: consult others about makin' this property public, and `Main.musicPitch`'s related netcode more universally applicable, at some point in the future
- fixed non-muted `PlaceTile` calls throwin' an error if occurrin' durin' worldgen
- fixed the `PitchVariance` property for `SoundStyle`s throwin' an exception on the wrong thing bein' < 0 when tryin' to set

### DIFF
--- a/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/Audio/SoundPlayer.cs
 +++ src/tModLoader/Terraria/Audio/SoundPlayer.cs
-@@ -1,24 +_,67 @@
+@@ -1,24 +_,66 @@
  using Microsoft.Xna.Framework;
  using ReLogic.Utilities;
  using System.Collections.Generic;
@@ -63,7 +63,6 @@
 +			
 +			// Handle 'UsesMusicPitch' (currently used solely to deal with vanilla's older instrument items).
 +			if (style.UsesMusicPitch) {
-+				styleCopy.Volume *= 0.75f;
 +				styleCopy.Pitch += Main.musicPitch;
 +			}
  
--- a/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
@@ -61,7 +61,7 @@
 +
 +			SoundStyle styleCopy = style;
 +			
-+			// Handle 'UsesMusicPitch'.
++			// Handle 'UsesMusicPitch' (currently used solely to deal with vanilla's older instrument items).
 +			if (style.UsesMusicPitch) {
 +				styleCopy.Volume *= 0.75f;
 +				styleCopy.Pitch += Main.musicPitch;
--- a/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/Audio/SoundPlayer.cs
 +++ src/tModLoader/Terraria/Audio/SoundPlayer.cs
-@@ -1,24 +_,66 @@
+@@ -1,24 +_,67 @@
  using Microsoft.Xna.Framework;
  using ReLogic.Utilities;
  using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
```diff
@@ -11,8 +12,11 @@ public abstract class ModBlockType : ModTexturedType
 		/// <summary> The internal ID of this type of tile/wall. </summary>
 		public ushort Type { get; internal set; }
 
-		/// <summary> The default style of sound made when this tile/wall is hit. Defaults to null. </summary>
-		public SoundStyle? HitSound { get; set; } = null;
+		/// <summary>
+		/// The default style of sound made when this tile/wall is hit.<br></br>
```
Mirsario: Are you not a certified _HTML programmer_?
```suggestion
		/// The default style of sound made when this tile/wall is hit.<br/>
```

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4508,6 +4508,68 @@
...
++				SoundEngine.PlaySound(SoundID.GuitarF, base.Center);
+ 		}
+ 
+ 		public void PlayDrums(float range) {
+ 			int num = 10;
+ 			float num2 = 1f / (float)num;
++			// tML: all these get changed to SoundIDs post-sound-overhaul to avoid misidentified drum sounds
+ 			if (range > num2 * 9f)
+-				SoundEngine.PlaySound(59, base.Center);
++				SoundEngine.PlaySound(SoundID.DrumCymbal2, base.Center);
```
Mirsario: This is all pointless, SoundID.TML.cs 's GetLegacyStyle does redirection already.

patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
```diff
@@ -59,10 +59,11 @@
 +				}
 +			}
 +
-+			// Handle 'UsesMusicPitch'.. This property is a weird solution, and is currently internal.
 +			SoundStyle styleCopy = style;
 +			
++			// Handle 'UsesMusicPitch' (currently used solely to deal with vanilla's older instrument items).
 +			if (style.UsesMusicPitch) {
++				styleCopy.Volume *= 0.75f;
```
Mirsario: Volume multiplication is out of place, should be done in styles instead.

Mirsario (changes requested):

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -225,6 +225,15 @@
  
  			for (int n = 0; n < Main.countsAsHostForGameplay.Length; n++) {
  				Main.countsAsHostForGameplay[n] = false;
+@@ -4085,7 +_,7 @@
+ 
+ 			Main.tile[x, num7].slope(0);
+ 			Main.tile[x, num7].halfBrick(halfBrick: false);
+-			PlaceTile(x, num10, 135, mute: false, forced: true, -1, 7);
++			PlaceTile(x, num10, 135, mute: true, forced: true, -1, 7); // tML: muted post-sound-overhaul to avoid a crash
+ 			for (int m = num8; m <= num10; m++) {
```
Mirsario: What crash? It should be fixed instead. What are the reproduction steps?

ThomasThePencil: literally just generate a world normally. no mods, no anything. as it is on current preview, attemptin' to generate a world while any non-muted `PlaceTile` calls are present throws an error due to the attempt to play the sound in a scenario where one can't be played

patches/tModLoader/Terraria/Player.cs.patch
ThomasThePencil: true as that may be, the fact that the newer instrument items such as Stellar Tune and Drum Set play the entirely wrong sounds currently as a result of said "redirection" suggests that either it isn't done well enough, or something's messed up with paths

edit: it was pathin' on the things themselves. fixed now

patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
ThomasThePencil: I mean, this is a replication of the vanilla rule for the instrument sounds that use this property, but w/e

patches/tModLoader/Terraria/WorldGen.cs.patch
ThomasThePencil: fixed in a better fashion by makin' `PlaceTile` calls never play a sound mid-worldgen

Mirsario (approved):

PR #2440, opened 2022-05-27
Add tModPorter (as not a submodule)

Chicken-Bones:
tModPorter is by no means finished, but the architecture is ready, and I've got it integrated with tML. 

I'd like to merge it now so that development can be on the main repository, and improvements can roll out to users straight away.

Current launch is via bat file, drag n drop or if you just click it'll prompt for path.
image

Future TODO is to integrate directly into mod sources menu.

Only thing I need help with is removing these files from the output (and steam) dir, as the exe is non-functional.
image

The `Tile.cs.patch` changes don't need to be in this PR, but it's just making some methods which were internalised from 1.3 -> 1.4 public again, because there was no need for them to be internal.


Solxanich (approved):
Approved per Discord: https://discord.com/channels/103110554649894912/937224651959062528/980152027801858078

Issue #2442, opened 2022-05-27
Sloping of non-solid tiles


Ved-s:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Add Global/ModTile.CanBeSloped which will override Main.tileSolid check in Player.ItemCheck_UseMiningTools_TryPoundingTile

PR #2444, opened 2022-05-28
Add CanBeSloped for non solid tiles

Ved-s:
### What is the new feature?

Feature allows non-solid tiles to be sloped, PR for issue #2442

Modders might use this to add more tile interactions

A full hook would be more flexible, but is probably unnecessary given the ability to return false from `Slope(i, j)`

```cs
TileID.Sets.CanBeSloped[Type] = true
```

### ExampleMod updates
`ExampleSlopedTile` (tile that can be sloped but is not solid)


direwolf420:
Docs: "Allows you to override vanilla tile slope behaviour for any non-solid tiles" is not quite accurate with the current code, when you return false in a GlobalTile hook, the subsequent type checks might still return true and continue the code. 
I would suggest splitting it up so that if the hook returns false, the type checks are skipped

Also, maybe concider an `OnSloped` hook aswell (that allows you to actually **override behavior**, CanX hooks should only be for allowing/preventing).

Ved-s:
`OnSloped` already exists, no? As `Slope()`

Chicken-Bones:
Any new hook needs a good example in ExampleMod

Ved-s:
I'll make an example tomorrow then

direwolf420:
Can you update GlobalTile docs aswell, or atleast change the hook placement/logic to prevent the type checks in case false is returned?

direwolf420:
Given how you set up the new hook, this doesn't make much sense as both true and false are handled equally within the TileLoader call, which is contrary to how most CanX hooks are set up. This opens it up to inconsistencies with how multiple GlobalTiles handle it, which might also depend on mod load order or when which hook returns something (Global 1 sets it to true, Global 2 sets it to false, vs Global 1 sets it to false, Global 2 sets it to true). The very first approach was closer to my suggested state than this one. Also naming wise I would revert it to the first approach too, Modify implies you can also directly change the tile there (which is false, Slope is the correct place).

direwolf420 (approved):

PR #2457, opened 2022-05-30
[1.4] More tModPorter tests

direwolf420:
### Description
Added a few common test cases encountered in porting.
This PR does NOT add the rewriters!

Also let me know if the way I added comments is acceptable/correct, this is my first tModPorter related PR.

I also noticed inconsistencies regarding line endings (CRLF/LF/Mixed), but it does not affect me.

### DIFF
--- a/tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
+++ b/tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
@@ -1,0 +1,17 @@
+using Terraria.ModLoader;
+using ReLogic.Graphics;
+using Terraria.GameContent; //FontAssets
+
+public class FontsTest : Mod
+{
+	void Method() {
+		DynamicSpriteFont font = null;
+
+		font = FontAssets.DeathText.Value;
+		font = FontAssets.ItemStack.Value;
+		font = FontAssets.MouseText.Value;
+
+		int index = 0;
+		font = FontAssets.CombatText[index].Value;
+	}
+}
--- a/tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
+++ b/tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
@@ -1,6 +1,6 @@
 using Terraria.ModLoader;
 using ReLogic.Graphics;
-using Terraria.GameContent; //FontAssets
+using Terraria.GameContent;
 
 public class FontsTest : Mod
 {
--- a/tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
+++ b/tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
@@ -4,4 +4,10 @@ using Terraria.ModLoader;
 public class GlobalItemTest : GlobalItem
 {
 	public override bool PreReforge(Item item) { return false; /* comment */ }
+
+	public override void ModifyWeaponKnockback(Item item, Player player, ref StatModifier knockback) { /* Empty */ }
+
+	public override void ModifyWeaponCrit(Item item, Player player, ref float crit) { /* Empty */ }
+
+	public override void ModifyWeaponDamage(Item item, Player player, ref StatModifier damage) { /* Empty */ }
 }
--- a/tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
+++ b/tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
@@ -1,10 +1,13 @@
 using Terraria;
 using Terraria.ModLoader;
 using Terraria.ModLoader.IO;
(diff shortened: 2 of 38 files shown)


tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
```diff
@@ -0,0 +1,17 @@
+using Terraria.ModLoader;
```
Chicken-Bones, line 1: we'll probably leave the unnecessary `using Terraria;`, as it's easy for modders to cleanup later, and figuring out whether it's now redundant is a bit of a pain.

tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
```diff
@@ -0,0 +1,17 @@
...
+using ReLogic.Graphics;
+using Terraria.GameContent; //FontAssets
```
Chicken-Bones: Do you think adding a comment about why usings were added is a good idea? Or should we just add them silently. What if they end up required by different rewriters for different reasons?

tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
```diff
@@ -1,7 +1,19 @@
 using Terraria;
 using Terraria.ModLoader;
+using Terraria.ModLoader.IO;
+using Microsoft.Xna.Framework;
```
Chicken-Bones, line 4: Additional framework import? Probably just a leftover.

tModPorter/tModPorter.Tests/TestData/ModProjectileTest.Expected.cs
```diff
@@ -1,9 +1,17 @@
 using System;
 using Terraria.ModLoader;
+using Microsoft.Xna.Framework; // Vector2, Color
+// Blank
```
Chicken-Bones: Yeah as above, we probably won't remove redundant usings. Could be considered helpful in the future? But it's a standard feature of VS

tModPorter/tModPorter.Tests/TestData/SoundTest.Expected.cs
```diff
@@ -0,0 +1,42 @@
...
+		SoundEngine.PlaySound(SoundID.Zombie10, player.position);
+
+		SoundEngine.PlaySound(SoundID.Item10, player.position);
+		SoundEngine.PlaySound(SoundID.NPCHit10, player.position);
+		SoundEngine.PlaySound(SoundID.NPCDeath10, player.position);
+		SoundEngine.PlaySound(SoundID.Zombie10, player.position);
+
+		//TODO modded sound retrieval&playback
+		//TODO non-dedicated IDs (Compile_Error req?)
+		//TODO pitch + volume
```
Chicken-Bones: These comments are great! It's good to be reminded why test cases exist. Just make sure the comments are in the same place in the input and expected `.cs` files.

tModPorter/tModPorter.Tests/TestData/SimpleRenamedVanillaMembersTest.Expected.cs
```diff
@@ -0,0 +1,44 @@
+using Terraria;
+using Terraria.Chat; // ChatHelper
+using Terraria.ID;
+using Microsoft.Xna.Framework;
+
+public class SimpleRenamedVanillaMembersTest
+{
+	void Method() {
+		var dresserX = Main.interactedDresserTopLeftX;
```
Chicken-Bones, line 9: You could try adding some of these to `Config.cs` yourself! (if you're interested). It doesn't yet have the support for renames across types, but for all the simple stuff that's just renamed within the same type, it should be one line to add!

Chicken-Bones (approved):
Great stuff, mostly some comments about imports. I'm happy to clean these up as I get to them, but I thought I'd review so you know what to expect for next time.

tModPorter/tModPorter.Tests/TestData/GlobalItemTest.Expected.cs
direwolf420: Rectangle for UseStyle

tModPorter/tModPorter.Tests/TestData/FontsTest.Expected.cs
direwolf420: I'd rather not leave them unless the tester/rewriter breaks because of it

Chicken-Bones: The test framework will complain that the file produced by tModPorter (which won't remove the using) is different from the Expected.cs

Chicken-Bones: Anyway, is stuff I can tweak in 20s, so no worries

PR #2470, opened 2022-06-01
[1.4] Move `Recipe.CreateRecipe` and `Recipe.CloneRecipe` to `Recipe`

ExterminatorX99:
### What is the new feature?
Clear up `Mod` by moving relevant methods to `Recipe`

Clarity of recipe creating and good design

Keep as is

ExampleMod

### ExampleMod updates

Fix compile issues


patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -426,8 +441,8 @@ public static class ConsumptionRules
 			return this;
 		}
 
-		internal static Recipe Create(Mod mod, int result, int amount) {
-			var recipe = new Recipe(mod);
+		public static Recipe Create(int result, int amount = 1) {
+			var recipe = new Recipe(ModContent.LoadingMod);
```
Jofairden: Are we 100% certain we can safely do this? If so, Im not really against it.
Gonna tag CB to make sure of it.

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -426,12 +438,13 @@ public static class ConsumptionRules
...
 		}
 
-		internal static Recipe Create(Mod mod, int result, int amount) {
-			var recipe = new Recipe(mod);
-
+		public static Recipe Create(int result, int amount = 1) {
 			if (!RecipeLoader.setupRecipes)
 				throw new RecipeException("A Recipe can only be created inside recipe related methods");
 
+			ArgumentNullException.ThrowIfNull(RecipeLoader.CurrentMod);
```
Jofairden, line 445: I wonder if these checks should be going into the Recipe ctor instead, wouldn't that be cleaner? ( I don't remember if there is a usecase for passing null for the mod )

ExterminatorX99: Vanilla recipes use `null` for mod

Issue #953, opened 2020-06-02
Remove Build All


GoodPro712:
### Description

Remove the build all button in mod sources as it serves no real use.
I personally have never used the button in my year of modding.

### What does this proposal attempt to solve or improve?

This will improve tModlLoader by allowing space for a more useful button to put in its place.

### Which (other) solutions should be considered?

Keep it.

Chicken-Bones:
Cast your votes people. I vote for keep.

JavidPack:
I would never use it, and I hear a lot of people in the past lamenting accidentally clicking it.

direwolf420:
I curse that and the "Build + Reload All" button each time I click it (which is rare), but I just Alt + F4 when that happens. Wouldn't be a problem if I only had a handful of mods there, but its something like 40.
People who maintain multiple mods (and possibly libraries), already use VS + csproj to handle building the libraries and mods for testing. For publishing, you would build each mod separately anyway to doublecheck that it works as intended, and then publish.
For tml version migration: You would have to go through each mod separately again, to fix/update the code necessary. 

Just my two cents

Jofairden:
> I curse that and the "Build + Reload All" button each time I click it (which is rare)

Build + Reload all is very useful in my opinion, I use it all the time.

GoodPro712:
I also meant this for Build + Reload All. Anything that builds all is pretty useless in my opinion.

Jofairden:
> I also meant this for Build + Reload All. Anything that builds all is pretty useless in my opinion.

You always have to reload all mods when a mod is built, you cannot reload just one mod, it doesn't work like that.

GoodPro712:
That's not what I said. I was meaning both these buttons were useless
![image](https://user-images.githubusercontent.com/45357714/83867279-85201400-a76c-11ea-8a87-248ae2b895e4.png)

Jofairden:
Ah I see. Sorry I misunderstood

Chicken-Bones:
Looks like the votes are in! Lets remove it.

Steviegt6:
I wouldn't mind if it was removed, but TBH it doesn't really have to be removed if you just added a cancel button. I'm sure someone out there uses the button actively and we have nothing to replace it with as of now. Just my two cents though, IDRC either way, though.

direwolf420:
Cancel button might work in a way where it will prevent the next mods' build, as iirc interrupting the build process of one mod isn't possible

[...]

PR #2472, opened 2022-06-01
Mod building using `dotnet build`

Chik3r:
## What is the new feature?
This PR changes mod building to use `dotnet build` instead of integrating building into the tModLoader application itself. It does this by running a post-build task after the source code is built that packages the mod into a `.tmod` file. This PR allows mods to use NuGet packages and other assembly references directly via the `.csproj` without any of the weird file handling that was required before.

Changes:
- DLL references are now added using `<Reference>`s in a mod's `.csproj`.
	- They must point to a `.dll` file inside the mod source folder (`ModName/file.dll` or `ModName/path/to/file.dll`).
	- Replaces `dllReferences` from `build.txt`.
	- No longer limited to being inside the `lib/` folder, now they can be anywhere inside the mod's folder.
	- Being inside the mod's folder is required because the `BuildTools` get all the references and then filter the ones inside a mod's source folder to copy them to the `.tmod` file.
- `<PackageReference>`s are now supported
- `<ProjectReference>`s are now supported
	- They will be automatically copied to the `.tmod` file unless `Private` is set to `true`.
	- Transitive project references will also be copied unless they are disabled with `PrivateAssets` or `DisableTransitiveProjectReferences`
- `<ModReference>`s should be used instead of `modReferences` in `build.txt`
- `hideCode` has been removed and now `.dll` files are always extracted. 
- Localization files (`.hjson`) are now always extracted, even if `HideResources` is set to `true`.
- `noCompile` has been removed as it is no longer relevant and was rarely used if ever before, and was practically undocumented. The main use it previously had was making mods using other .NET languages, but that should now be possible without the need for `noCompile`.
Currently, if a mod developer wanted to use a NuGet package in their mod, they'd have to fetch the package's `.dll` file from NuGet, copy it to the `lib/` folder and add it to `dllReferences` inside `build.txt`.  If they needed to update the package, they would have to find the new `.dll` and copy it again. Since a custom MSBuild task is being run now after compiling the mod, it can automatically get all the required NuGet packages (added with normal `<PackageReference>`s) and copy them over to the `.tmod` file (while also internally adding them to `dllReferences`)
It was also not possible to use `<ProjectReference>`s without adding a task to copy the `.dll` file to the `lib/` folder before building. This PR solves that and allows `<ProjectReference>`s to be used for dependencies.

## How does this affect modders?
If you do not use any mod or dll references, this should have no effect for you. You should migrate your `build.txt` to the `.csproj` file, but can continue using it without issues. Below are more details on the changes that may affect you:
- If you deleted your `.csproj` file, you will need to create a new one. You can do this by clicking the `Upgrade .csproj file` in the mod sources menu.
- Mod properties should now be placed in the `.csproj` file.
	- As a result `build.txt` has been deprecated. You can still use it and everything will work the same as before, but migrating to the new system is recommended.
	- Most properties have the same name from `build.txt` but using PascalCase (ex: `displayName` to `<DisplayName>`). The only exception is `side` which has been renamed to `ModSide`.
	 - It is possible to migrate from `build.txt` to `.csproj` properties by using the in-game button in mod sources.
		Pasted image 20230917231436
	- Here is an example of the result:
		image
	- They should be placed inside a `<PropertyGroup>` as shown below:
```xml
<PropertyGroup>
  <Version>1.0.0</Version>
  <Author>Developer name</Author>
  <DisplayName>Mod name</DisplayName>
  ...
</PropertyGroup>
```
- `dllReferences` have been replaced with `<PackageReference>`s, `<ProjectReference>`s, and `<Reference>`s
	- `<PackageReference>`s are for NuGet packages.
	- `<ProjectReference>`s are for references to other .NET projects.
	- `<Reference>`s are for references to `.dll` files.
	- More details on how to use them and their syntax below.
- `modReferences` and `weakReferences` have been replaced with `<ModReference>`s
	- Add `Weak` as a metadata value to specify if the reference is a weak ref.
	- If you don't use any code from the referenced mod, you'll be able to compile just like before.
	- If you use code from the referenced mod, you will need to add a way to resolve the references by adding a path to the mod's `.dll` file with `HintPath` or to the mod's `.csproj` with `ProjectPath`.
		- You can obtain the mod's `.dll` by  clicking on it from the `Mods` menu and then extracting the mod, you can follow the steps here: [Expert Cross Mod Content · tModLoader/tModLoader Wiki (github.com)](https://github.com/tModLoader/tModLoader/wiki/Expert-Cross-Mod-Content#visual-studio-steps).
	- Previously you were able to compile from in-game without doing the above, but would fail to compile from Visual Studio. Now you will need to fix the references before being able to compile from in-game or VS.
	- More details on how to use them and their syntax below.

### Methods to build a mod
You can build a mod by doing any of the following options:
1. Clicking the `Build` button in-game
	- Works the same as before for most cases.
	- If you have references to symbols from other mods, you'll need to add a way to resolve them in your `.csproj` file. This wasn't needed before as TML automatically added them for you when building in-game.
	- Always built as `Release`
	- It calls `dotnet build` in the background
2. Building from Visual Studio, Rider, or other IDEs
3. Building from the command line with the `dotnet build` command
	- Mod properties can be passed as parameters when building (`dotnet build -p:PropertyName=Value`)

### How do I reference another mod now?
1. If you need to use any types (a class, method, variable, etc.) from the mod you are referencing, you'll need to extract the mod from the Mods menu in order to get it's `.dll` file. The `.dll` file can then be found in the `ModAssemblies` folder inside the `ModSources` folder.
2. Open your mod's `.csproj` and find the line that says `<PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />`
3. Insert a new line below and add `<ModReference Include="InternalModName"></ModReference>`, where `InternalModName` is the internal name of the mod you are referencing.
4. If you are creating a weak reference, add `Weak="true"` after `Include="InternalModName"`.
5. If you did step 1, you'll have to add `<HintPath>path/to/mod.dll</HintPath>` inside the `ModReference` element. `path/to/mod.dll` should match the relative path from your mod's folder to the referenced `.dll`  file.
6. Alternatively, you can use a `<ProjectPath>` if you have access to the referenced mod's source code. It should point to a `.csproj` file. Example: `<ProjectPath>../ReferencedMod/ReferencedMod.csproj</ProjectPath>`

**Examples:**
A weak reference to `ReferencedMod`
```xml
<ItemGroup>
  <PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />
  <ModReference Include="ReferencedMod" Weak="true"></ModReference>
</ItemGroup>
```

A strong reference to `StrongReferencedMod` with it's `.dll` file in the `ModAssemblies` folder.
```xml
<ItemGroup>
  <PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />
  <ModReference Include="StrongReferencedMod">
    <HintPath>../ModAssemblies/StrongReferencedMod.dll</HintPath>
  </ModReference>
</ItemGroup>
```

A strong reference to `MyReferencedMod` and its `.csproj` file
```xml
<ItemGroup>
  <PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />
  <ModReference Include="MyReferencedMod">
    <ProjectPath>../MyReferencedMod/MyReferencedMod.csproj</ProjectPath>
  </ModReference>
</ItemGroup>
```

## Available metadata for references
As mentioned before, references can be one of four types, each with metadata that can be changed.
1. `<Reference Include="path/to/file.dll" />`: A standard reference to a `.dll` file. The `.dll` file has to be inside the mod's folder. (This is because `@(Reference)` returns all references and there doesn't seem to be a way to filter only the references explicitly added by the `.csproj`). The following metadata can be set:
	- `Include`: This is required and corresponds to the path to the `.dll` file.
	- `Private`: This is optional and is set to `false` by default.  If set to `true`, the `.dll` won't be copied into the `.tmod` file or added to `dllReferences`
 2. `<ProjectReference Include="path/to/project.csproj" />`: A normal reference to another `.csproj`. The following metadata can be set:
	- `Private`: This is optional and is set to `false` by default.  If set to `true`, the project's output `.dll` won't be copied into the `.tmod` file or added to `dllReferences`
 3. `<PackageReference Include="Package" Version="1.0" />`: A normal NuGet reference. By default, they will be copied to the `lib/` folder inside the `.tmod` file.
	- `Private`:  This is optional and set to `false` by default. If set to `true`, the `.dll` of the NuGet package won't be copied to the `.tmod` file
4. `<ModReference Include="TestReferenceMod" Weak="false">`: A reference to another mod. If you use any type from another mod, you must add a way to resolve the references using `<HintPath>` or `<ProjectPath>`.
	- `Include`: This is required and should match the internal name of the mod you are referencing (e.g., `CalamityMod` or `ExampleMod`)
	- `Weak`: This is optional and is `false` by default. Setting this to `true` makes your reference a weak reference.
	- `HintPath`: This is optional and is empty by default. It should be set to a path to a `.dll` file of the mod. It will be ignored if `ProjectPath` has a value.
	- `ProjectPath`: This is optional and empty by default. It is an alternative to `HintPath` and will load a `.csproj` (as a `ProjectReference`) instead of a `.dll`

**Examples:**
An example of referencing multiple things:
```xml
<ItemGroup>
  <PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />
  <PackageReference Include="RestSharp" Version="106.15.0" />
  <PackageReference Include="SkiaSharp.Views" Version="2.88.0" />
  <!-- Private makes it so that the reference is not added to the output tmod file. -->
  <!-- That means that the .dll file is not copied over, and the reference is not added to the internal Info file -->
  <Reference Include="TerrariaNoRangePylon.dll" Private="true" />
</ItemGroup>
<ItemGroup>
  <ModReference Include="TestReferenceMod" Weak="false">
    <!-- HintPath lets you use a mod's .dll as a reference (the .dll is not placed inside the .tmod file) -->
    <HintPath>..\TestReferenceMod\bin\Debug\net6.0\TestReferenceMod.dll</HintPath>
    <!-- ProjectPath does the same as HintPath, but all the mod's types are loaded using a .csproj instead of a .dll file -->
    <!-- As ProjectPath is set here, HintPath will be ignored -->
    <ProjectPath>..\TestReferenceMod\TestReferenceMod.csproj</ProjectPath>
  </ModReference>
</ItemGroup>
```


Chicken-Bones:
You should pull the tML version from the tML dll itself. It's important that all the build tools match the actual version of tML used

Jofairden:
Wow, nice PR. I absolutely love it.
Could you say in a 1 liner how this will work for people not using a .csproj?
Like when we're talking about building the mod, it's great that you can just run TML alongside VS and build the mod using the dotnet command. However, you said:
> Mods are still able to be compiled with a build.txt file (which has to be deleted in order for the build info to be loaded from the .csproj) but the build will throw a warning.

Does this mean if you are using this method the in-game building won't function, and vice-versa? I would, optimally, only approve this if just both ways of building work alongside eachother. There are actually plenty of modders using just np++ (not very experienced) and building in-game, not debugging etc.

BuildTools/ModFile/BuildProperties.cs
```diff
@@ -0,0 +1,258 @@
...
+	private bool _hideResources = false;
+	public bool IncludeSource = false;
+
+	private string _eacPath = "";
+
+	// This .tmod was built against a beta release, preventing publishing.
+	internal bool Beta = false;
+	private string _homepage = "";
+	private string _description = "";
+	private ModSide _side;
```
Jofairden: Could you put public variables first, then all the private ones below it? Thanks (sort by visibility)

BuildTools/ModFile/BuildProperties.cs
```diff
@@ -0,0 +1,258 @@
...
+
+			// Make the first letter lowercase
+			propertyName = char.ToLowerInvariant(propertyName[0]) + propertyName.Substring(1);
+
+			ProcessProperty(properties, propertyName, propertyValue);
+		}
+
+		var refs = properties.RefNames(true).ToList();
+		if (refs.Count != refs.Distinct().Count())
+			throw new Exception("Duplicate mod/weak reference");
```
Jofairden: Maybe this should warrant a custom exception? How is this exception relayed right now when you use the dotnet command?

BuildTools/ModFile/BuildProperties.cs
```diff
@@ -0,0 +1,258 @@
...
+		//add (mod|weak)References that are not in sortBefore to sortAfter
+		properties._sortAfter = properties.RefNames(true).Where(dep => !properties._sortBefore.Contains(dep))
+			.Concat(properties._sortAfter).Distinct().ToArray();
+
+		return properties;
+	}
+
+	private static void ProcessProperty(BuildProperties properties, string property, string value) {
+		if (string.IsNullOrEmpty(value))
+			return;
```
Jofairden: Please just add it to the switch-case 
```
switch(property ?? String.Empty)
{
    case "":
        break;
```

BuildTools/ModFile/ContentConverters.cs
```diff
@@ -0,0 +1,49 @@
...
+					return true;
+				}
+				src.Position = 0;
+				return false;
+			default:
+				return false;
+		}
+	}
+
+	public static void ToRaw(Stream src, Stream dst) {
```
Jofairden: Is there a reason for not using this?
https://github.com/tModLoader/tModLoader/blob/dd44e70738e29e5ded0cb979355b671b41f56efe/patches/tModLoader/Terraria/ModLoader/IO/ImageIO.cs#L11

BuildTools/ModFile/FileEntry.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace tModLoader.BuildTools.ModFile;
+
+public class FileEntry
```
Jofairden, line 3: Same
https://github.com/tModLoader/tModLoader/blob/dd44e70738e29e5ded0cb979355b671b41f56efe/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs#L18

BuildTools/Tasks/PackageModFile.cs
```diff
@@ -0,0 +1,205 @@
...
+		List<ITaskItem> dllReferences = ReferencePaths.Where(x => x.GetMetadata("FullPath").StartsWith(ProjectDirectory)).ToList();
+		Log.LogMessage(MessageImportance.Low, $"Found {dllReferences.Count} dll references.");
+
+		foreach (ITaskItem taskItem in nugetReferences) {
+			string nugetName = "lib/" + taskItem.GetMetadata("NuGetPackageId") + ".dll";
+			string nugetFile = taskItem.GetMetadata("HintPath");
+
+			Log.LogMessage(MessageImportance.Low, $"Adding nuget {nugetName} with path {nugetFile}");
+			if (string.Equals(taskItem.GetMetadata("Private"), "true", StringComparison.OrdinalIgnoreCase))
+				continue;
```
Jofairden: How about movin the Log after the conditional and adding a log about skipping the private nuget file

BuildTools/Tasks/PackageModFile.cs
```diff
@@ -0,0 +1,205 @@
...
+			modProperties.AddDllReference(taskItem.GetMetadata("NuGetPackageId"));
+		}
+
+		foreach (ITaskItem dllReference in dllReferences) {
+			string dllPath = dllReference.GetMetadata("FullPath");
+			string dllName = Path.GetFileNameWithoutExtension(dllPath);
+
+			Log.LogMessage(MessageImportance.Low, $"Adding dll reference with path {dllPath}");
+			if (string.Equals(dllReference.GetMetadata("Private"), "true", StringComparison.OrdinalIgnoreCase))
+				continue;
```
Jofairden: Same as before

BuildTools/Tasks/PackageModFile.cs
```diff
@@ -0,0 +1,205 @@
...
+			string? weakRef = modReference.GetMetadata("Weak");
+
+			Log.LogMessage(MessageImportance.Low, $"Adding mod reference with mod name {modName} [Weak: {weakRef}]");
+			modProperties.AddModReference(modName, string.Equals(weakRef, "true", StringComparison.OrdinalIgnoreCase));
+		}
+	}
+
+	private List<ITaskItem> GetNugetReferences() {
+		Dictionary<string, ITaskItem> nugetLookup = PackageReferences.ToDictionary(x => x.ItemSpec);
+		if (nugetLookup.ContainsKey("tModLoader.CodeAssist")) nugetLookup.Remove("tModLoader.CodeAssist");
```
Jofairden: Maybe make it more generic, i.e. 'ignoredNugetPackages' or something we can easily configure. In case we change them or make more.

BuildTools/Tasks/PackageModFile.cs
```diff
@@ -0,0 +1,205 @@
...
+
+		List<ITaskItem> nugetReferences = new List<ITaskItem>();
+		foreach (ITaskItem referencePath in ReferencePaths) {
+			string? hintPath = referencePath.GetMetadata("HintPath");
+			string? nugetPackageId = referencePath.GetMetadata("NuGetPackageId");
+			string? nugetPackageVersion = referencePath.GetMetadata("NuGetPackageVersion");
+
+			if (string.IsNullOrEmpty(nugetPackageId)) continue;
+			if (!nugetLookup.ContainsKey(nugetPackageId)) continue;
```
Jofairden: Joining with || 
```
if (string.IsNullOrEmpty(nugetPackageId)
   || !nugetLookup.ContainsKey(nugetPackageId)) continue;
   ```
   
   Or is there a specific reason for separating them?

File: BuildTools/Tasks/PackageModFile.cs [hunk 10]
```diff
@@ -0,0 +1,205 @@
...
+			Log.LogWarning($"Expected {nugetLookup.Count} nuget references but found {nugetReferences.Count}.");
+		return nugetReferences;
+	}
+
+	private List<ITaskItem> GetModReferences() {
+		List<ITaskItem> modReferences = new List<ITaskItem>();
+		foreach (ITaskItem modReference in ModReferences) {
+			string? modPath = modReference.GetMetadata("HintPath");
+			if (string.IsNullOrEmpty(modPath))
+				modPath = modReference.GetMetadata("ProjectPath");
```
Jofairden: You can make this shorter
```
string? modPath = modReference.GetMetadata("HintPath") ?? modReference.GetMetadata("ProjectPath");
```

Jofairden: Im assuming it's just a null check, since you've made string nullable

BuildTools/Tasks/PackageModFile.cs
```diff
@@ -0,0 +1,205 @@
...
+		Log.LogMessage(MessageImportance.Normal, $"Found {modReferences.Count} mod references.");
+		return modReferences;
+	}
+
+	private BuildProperties GetModProperties() {
+		string buildInfoFile = Path.Combine(ProjectDirectory, "build.txt");
+		if (File.Exists(buildInfoFile)) {
+			Log.LogWarning("Using deprecated build.txt file");
+			return BuildProperties.ReadBuildInfo(buildInfoFile);
+		}
```
Jofairden: I wonder if this is needed. (How obsolete / deprecated is build.txt if it is needed for building in-game?)
Why not just read the properties as done and just ignore build.txt (not using for the dotnet command)?

Jofairden: We can even make this backwards compatible 
i.e. modders dropping the build.txt we can just check in-game for these .csproj properties like you're doing in this PR, and use that instead of build.txt 
For those not using .csproj we keep the simple key = value support in build.txt

ExampleMod/build.txt
```diff
@@ -1,8 +0,0 @@
-author = TML Team
-displayName = Example Mod
-homepage = https://tmodloader.net/
-hideCode = false
-hideResources = false
-includeSource = true
-buildIgnore = *.csproj, *.user, obj\*, bin\*, .vs\*, Old\*
-includePDB = true
```
Jofairden: If a modder now gets ExampleMod and tries to build in-game, is it going to fail?

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -105,14 +105,14 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
 				Top = { Pixels = -65 }
 			};
 			buttonBA.WithFadedMouseOver();
-			buttonBA.OnClick += BuildMods;
+			//buttonBA.OnClick += BuildMods;
```
Jofairden: So the build all functionality is completely gone?

Jofairden (commented):
So, yeah, again I really like this PR. Looks good
However my single question does remain: what happens to those not using .csproj ? I see the build logic was moved to this buildtool, and I think that's fine as long as we also support building without visual studio. I also don't see code that upgrades modder's  current build.txt to this new format. Ideally this update silently fixes it for modders internally so they dont have to be worried about breaking anything, while we communicate the change. Also, building all mods at once is gone? Where did it go ?

BuildTools/Tasks/PackageModFile.cs
Jofairden: Just saw this PR Is in favor of totally dropping the build.txt
If so, just resolve this conversation.

ExampleMod/build.txt
Jofairden: Same counts here

BuildTools/ModFile/BuildProperties.cs
Chik3r: If you have duplicate mod references, it will not throw an error right now as MSBuild either merges the references, or only uses one of the `ModReference` (im not sure which of these options it does, havent tested it)

BuildTools/ModFile/ContentConverters.cs
Chik3r: iirc, if I used the `ImageIO` methods in the `tModLoader` solution, I'd have to load all the libraries that tModLoader uses, which would require copying them next to the `BuildTools` dll file (for MSBuild to load them). My solution to this was to copy the method and make `BuildTools` use `ImageSharp` (to load images), making it necessary to only copy a single dll instead of many more.

BuildTools/ModFile/FileEntry.cs
Chik3r: Same reason, adding tModLoader would require copying its libraries

BuildTools/Tasks/PackageModFile.cs
Chik3r: I don't remember/know why I made it nullable (GetMetadata doesnt return null from a quick test, it returns `""`), but I need to check if it's empty

ExampleMod/build.txt
Chik3r: It shouldn't fail

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
Chik3r: It's been hidden from the UI for a while, this just removes more parts of it

Chik3r:
> Could you say in a 1 liner how this will work for people not using a .csproj?

It will not be possible to build mods without having a `.csproj` present. But the only way to not use a `.csproj`  right now is by setting `noCompile = true` (as far as I know), something that few people ever do and its only for special cases. If you do need to use `noCompile`, you could change the `OutputPath` that the task uses to point to a precompiled dll

>  I also don't see code that upgrades modder's current build.txt to this new format.

A button to move build.txt properties to the csproj has been added now

> Also, building all mods at once is gone? Where did it go ?

It's been gone for a while but I dont know exactly when it was removed. The code to use it still existed, but the button was just hidden. This pr now erases some of that code, but it shouldn't affect anything for a modder/user

> Does this mean if you are using this method the in-game building won't function, and vice-versa?

In-game building will still work, but now it runs a command (`dotnet build`) instead of compiling the dlls, packing the tmod file and doing other stuff. The same command can be used from anywhere to build your mods

steviegt6:
> [...] something that few people ever do and its only for special cases.

It's also worth noting that moving to using `dotnet build` (and MSBuild in general) eliminates practically all use cases `nocompile` previously had: namely F#/VB.NET mods (and anything else that compiles to CLI-compatible binaries).

BuildTools/ModFile/ContentConverters.cs
steviegt6: I believe ImageSharp might have some incompatibilities on Linux, even when manually providing binaries or installing them through the system package manage (at least on NixOS, which is somewhat unique). Might be a cause for concern, I can have a friend test this.

Destructor-Ben:
Are project references automatically copied to the lib folder too?

Chik3r:
> Are project references automatically copied to the lib folder too?

They are copied if the `Private` metadata isn't set to `true`


> Could you say in a 1 liner how this will work for people not using a .csproj?

They won't be able to build their mod. Unless someone manually deleted the `.csproj`, every mod should have a `.csproj` so they should still be able to compile their mod. When building in-game it should all work the same for a modder and the only changes they might need to do is port the build properties from `build.txt` to the `.csproj` file, but this is not obligatory and `build.txt` still works.


> Also, building all mods at once is gone? Where did it go ?

It has been gone for some time from the UI, see #953. All I've done is remove parts that were left over.


> I also don't see code that upgrades modder's current build.txt to this new format

There is now a button to migrate build properties from `build.txt` to the `.csproj` file, check the updated PR description.


> Does this mean if you are using this method the in-game building won't function, and vice-versa? I would, optimally, only approve this if just both ways of building work alongside eachother. There are actually plenty of modders using just np++ (not very experienced) and building in-game, not debugging etc.

You can still build in-game and through the CLI/Rider/VS/etc, this just unifies how they all work. Any modder using NP++, notepad, or any other tool can still build in-game or by using the `dotnet build` command

PR #2476, opened 2022-06-02
[Draft] Add ChooseBestTorch Hook (resolves #2475)

### LINKED ISSUE #2475: [1.4] [Feature Proposal] Torch God Manipulation Hook

Yes

Mod capability as a Modder

### Description

Add a new hook that allows modders to control whether a Torch God replacement torch should be placed and what tile ID and style should be used.

Allows modders to provide custom biome torches, among other things.

Requires discussion.

steviegt6:
### What is the new feature?
Resolves #2475, adds a new hook that lets modders manipulate the torch placed when Torch God's Favor is active (as well as other tiles, technically).

A proposal to move this functionality to `ModSceneEffect` was raised, but I am not concerned since vanilla doesn't have a concept of scene effects and attempting to use a `ModSceneEffect` for arbitrary custom logic is not desirable here. Additionally, this logic is traditionally located only in `Player`, and I wish to keep it that way. Torch variations shouldn't be restricted to `ModSceneEffect`s since you can absolutely use this functionality for other reasons, such as custom torch variations for vanilla biomes, accessories and whatnot that may disable or modify every placed torch, among other possibilities.

This hook allows modders to manipulate more about tiles that just when Torch God's Favor is active, due to the nature of the hook and its ability to toggle vanilla behavior and override the need for Torch God's Favor. There is no easy or convenient way around this.

See #2475.

Not really, but there are a some ways to rewrite the code. needs discussion.

See ExampleChooseBestTorch.cs.

### ExampleMod updates
See ExampleChooseBestTorch.cs.

### DIFF
--- a/ExampleMod/Common/Players/ExampleChooseBestTorch.cs
+++ b/ExampleMod/Common/Players/ExampleChooseBestTorch.cs
@@ -1,0 +1,35 @@
+﻿using ExampleMod.Content.Biomes;
+using ExampleMod.Content.Tiles;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	public class ExampleChooseBestTorch : ModPlayer
+	{
+		// ChooseBestTorch gives you control over the type and style of a torch if it's manipulated by Torch God's Favor.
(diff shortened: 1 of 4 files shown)


steviegt6:
Can't mark this as a draft after making it, I think, but I have several things to sort out with this PR at a later date.

PR #2477, opened 2022-06-02
Various Bestiary fixes and automation

JavidPack:
Fixes ModBiome.SetStaticDefaults being unable to properly assign DisplayName

Restores ExampleSurfaceBiome map background display

Fixes PartyZombie to show ExampleSurfaceBiome background

Tweaks icon_small to actually fit in bestiary filter menu

ModBiome.BestiaryIcon and BackgroundPath now autoload a default texture path

Fixes SceneEffect.SpecialVisuals not being called on all active scene effects

Bonus:
GetOrCreateTranslation now remembers translations it generates.

Breaking:
ModBiome.BestiaryIcon and BackgroundPath use path that includes modname to fit the convention
ModPlayer.SetMapBackgroundImage -> ModSceneEffect.MapBackground


ExampleMod/Common/Players/ExampleBiomePlayer.cs
```diff
@@ -0,0 +1,16 @@
+﻿using ExampleMod.Content.Biomes;
+using Microsoft.Xna.Framework.Graphics;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	public class ExampleBiomePlayer : ModPlayer
+	{
+		public override Texture2D GetMapBackgroundImage() {
```
Solxanich: This should be fixed at the ModSceneEffect level to automatically happen.
This should not be a part of ModPlayer.

PR #2480, opened 2022-06-02
Update tModLoader.json

Cyrillya:
Replace the names of some biomes with the translations from the Official Terraria Chinese Wiki
Add tags


JavidPack:
Looks like the translations this was doing have already been merged separately. Thanks for doing this.

PR #2481, opened 2022-06-02
1.4 main menu buttons

JavidPack:
Adds tModLoader specific menu buttons to allow users easier access to support and learning resources.


Jofairden:
Nice , I like it. However, Steam is very strict about having Patreon links in menus. Have we checked this ?

PR #2482, opened 2022-06-02
1.4 better asset load errors

JavidPack:
Better asset load errors, asset load errors while loading the mod now result in in-game error, rather than separate window.

Errors also now show a colored edit distance, to facilitate locating typos.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
@@ -7,6 +7,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using Terraria.Localization;
 using Terraria.ModLoader.Core;
+using Terraria.ModLoader.Exceptions;
 using Terraria.ModLoader.UI;
 
 namespace Terraria.ModLoader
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -323,6 +323,13 @@ public static class ModContent
 				mod.PostSetupContent();
 				SystemLoader.PostSetupContent(mod);
 				mod.TransferAllAssets();
+				if(mod.AssetExceptions.Count > 0) {
+					if (mod.AssetExceptions.Count == 1)
+						throw mod.AssetExceptions[0];
+
+					if (mod.AssetExceptions.Count > 0)
+						throw new MultipleException(mod.AssetExceptions);
+				}
```
Chicken-Bones: This code should go in `TransferAllAssets`

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -81,12 +81,40 @@ internal void PrepareAssets()
...
 			};
 		}
 
 		internal void TransferAllAssets() {
 			Assets.TransferAllAssets();
 		}
+
+		internal List<Exception> AssetExceptions = new List<Exception>();
+		internal void OnceFailedLoadingAnAsset(string assetPath, Exception e) {
+			if (!Main.gameMenu) {
```
Chicken-Bones: should just have a bool in the class, which is set in `TransferAllAssets` above like `initialTransferComplete = true` and we add to `AssetExceptions` until that bool is true.

PR #2496, opened 2022-06-04
Prevent SDL2 loading on servers.

Mirsario:
This is a bugfix. Currently, mods (or even the base game it would seem) can very easily trigger FNAPlatform's static constructor on servers, which causes SDL2 and graphics APIs to be loaded into memory.
image

This results in not only servers needlessly requiring natives to be around, but also breaks headless servers that may not have a desktop window manager and/or graphics & audio drivers installed.

Implemented at `TerrariaNetCore` stage by setting `FNA_PLATFORM_BACKEND` to `NONE` when `dedServ` is true. **Functionality in FNA is added in our fork of it.** The envvar's name was in a FNA comment -- seems to be no longer used. The value is new.


Chicken-Bones:
Looks good to me. Happy to merge to preview. For stable... need to make sure we don't otherwise update FNA... messy.

PR #2500, opened 2022-06-05
[1.4] ReadOnly `TmodFile.FileEntry,cachedBytes`

ExterminatorX99:
### What is the bug?
`TmodFile.FileEntry,cachedBytes` is intended to be readonly, but unfortunately there was no ReadOnlySpan on .NET 4.5

We're now on .NET 6

### How did you fix the bug?
Switch to `ReadOnlyMemory` and `ReadOnlySpan` where applicable

Keep as is


Chicken-Bones:
It's fine... but a bit ugly. The caching system isn't really used anymore atm. Was your goal here just to remove a TODO, or to make `Mod/ModContent.GetFileBytes` safer?

ExterminatorX99:
The goal was to remove the TODO

PR #2504, opened 2022-06-05
[1.4] Update Polish (pl-PL) localization

NothyariS0:
### Description

Added missing text to Polish (pl-PL) localization


JavidPack:
Thanks a bunch.

PR #2505, opened 2022-06-05
[1.4] Add fullyDestroyed parameter to (Mod/Global)BlockType.KillSound()

### LINKED ISSUE #2433: Mod/GlobalTile.KillSound needs fail parameter

Yes

Mod capability as a Modder

### Description

The `fail` parameter in `KillTile_PlaySounds` could be passed to the api methods. It would indicate if the tile is being hit or actually breaking. Most tiles in the game make the same sound if you hit them without breaking or hit them and break. Crystal Block is an example of a tile that makes the normal mine sound on early hits, but changes to the crystal break sound once the tile is fully mined. This is not possible with the current hook.

_No response_

_No response_

Antirhinnum:
### What is the new feature?
Added ~~`fail`~~ `fullyDestroyed` parameter to `ModBlockType.KillSound()`, `GlobalBlockType.KillSound()`, `TileLoader.KillSound()`, and `WallLoader.KillSound()`, and updates tModPorter to fix these method signatures. Resolves #2433.

Requested in #2433, allows more modder control. From the issue:
> Crystal Block is an example of a tile that makes the normal mine sound on early hits, but changes to the crystal break sound once the tile is fully mined. This is not possible with the current hook.

Not that I can think of.

See `ExampleAnimatedTile`.

### ExampleMod updates
Updates `ExampleAnimatedTile` to showcase the feature.

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleAnimatedTile.cs
+++ b/ExampleMod/Content/Tiles/ExampleAnimatedTile.cs
@@ -2,6 +2,7 @@
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Terraria;
+using Terraria.Audio;
 using Terraria.DataStructures;
 using Terraria.GameContent.Creative;
 using Terraria.ID;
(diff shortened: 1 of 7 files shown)


JavidPack:
Looks great. We'll get to this soon.

Chicken-Bones:
Thanks for adding the tModPorter lines!

My only comment here is that you need to give the parameter a better name. I'd consider `fullyDestroyed` or `justDamaged` as two potentially good names (with opposite meanings).

JavidPack:
`fail` is used in other hooks we have and matches what the source code terminology, I think changing it just makes it harder to follow the logic, actually. Thoughts @Chicken-Bones ?

Chicken-Bones:
Your call. `fail` is fine, but the documentation needs to be clearer. Eg `true if the tile is only partially damaged, false if the tile is destroyed`

Issue #1854, opened 2021-09-19
[1.4] Add a hook to builderAccStatus


zzp198-old:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

Yes

### Description

![`U`RWKKW_CGRR2 IJX(X91](https://user-images.githubusercontent.com/43930552/133914489-ffbc4d66-41e6-4563-96e7-6e12756617bb.png)

This hook allows players to customize icons and click events, etc

### What does this proposal attempt to solve or improve?

A fast local player configuration switch.
Yes, but it's not necessary. It's just a proposal.

### Which (other) solutions should be considered?

Add a ModBuilder class, which rewrites index, draw, texture, position, etc

JavidPack:
Sounds reasonable. If someone is attempting this, something similar to the implementation of the info display system would be a good guide to emulate. https://github.com/tModLoader/tModLoader/pull/1370

PR #2511, opened 2022-06-07
[1.4] Added support for Modded Builder Toggles

xKirtle:
### What is the new feature?
Modded Builder Toggles implementation
To provide a standard way for modders to add their own builder toggles to behave in a similar fashion to vanilla's ones. Requested in #1854 
Included in the Example Mod examples added.
[Functionality showcase imgur link](https://i.imgur.com/gUO2Nww.mp4)
### ExampleMod updates

Included in the video. One builder toggle that is active when the user isn't holding any items and another that is "dimmed" when clicked and its state becomes "Off"


FoxAmes:
Would love to see this merged, does it need anything other than a rebase to 1.4.4 to be considered?

JavidPack:
I'm not positive how much it needs besides being updated to 1.4.4. We were busy at the time and this one fell through the cracks. I don't imagine it would need much more, I remember it being functional

xKirtle:
Mirsario suggested at the time that another route could be taken instead of ModX and follow the tooltips approach, but I don't think anything was ever settled on. I can revisit this PR and update it to 1.4.4 but would love to know your opinions on which approach should be used.

JavidPack:
Sorry, I forgot to reply it seems. I don't see a way for this to not have a ModX class. The design looks great to me. I want to prioritize this, if you are still around, updating the PR to 1.4.4 shouldn't be too difficult. I noticed builderAccStatus changed slightly, so we'll need to look into that. 

I also noticed looking at the patches that it doesn't seem like saving and loading was implemented. Note how vanilla code needs to be fixed to save the original number of slots, and additional code is needed in PlayerIO to save the modded ones. InfoDisplayLoader is once again a good example of that.

JavidPack:
Port of this PR has been merged. Thanks for putting this together.

PR #2515, opened 2022-06-07
[1.4] Example Crystal

BasicallyIAmFox:
Example Crystal, those act similarly to Crystal Shards (including spreading and item form)

Code-wise? Probably yes.
Anything else? Most likely no.


ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,119 @@
...
+        public override ushort GetMapOption(int i, int j) => (ushort)(Main.tile[i, j].TileFrameX / 18);
+
+        public override void KillTile(int i, int j, ref bool fail, ref bool effectOnly, ref bool noItem)
+        {
+            int toDrop = 0;
+            switch (Main.tile[i, j].TileFrameX / 18)
+            {
+                case 0:
+                    toDrop = Mod.Find<ModItem>("ExampleCrystalItem").Type;
+                    DustType = Mod.Find<ModDust>("Sparkle").Type;
```
Chicken-Bones: Use `ModContent.ItemType<ExampleCrystalItem>()` and  `ModContent.DustType<Sparkle>()`

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,119 @@
...
+    internal class ExampleCrystal : ModTile
+    {
+        public override void SetStaticDefaults()
+        {
+            TileObjectData.newTile.Width = 1;
+            TileObjectData.newTile.Height = 1;
+            TileObjectData.newTile.CoordinateWidth = 16;
+            TileObjectData.newTile.CoordinateHeights = new int[] { 16 };
+            TileObjectData.newTile.CoordinatePadding = 2;
+            TileObjectData.newTile.StyleHorizontal = true;
```
Chicken-Bones: comments on some of these lines would be nice

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,119 @@
...
+            Main.tileShine[Type] = 500;
+            Main.tileSpelunker[Type] = true;
+        }
+
+        public override ushort GetMapOption(int i, int j) => (ushort)(Main.tile[i, j].TileFrameX / 18);
+
+        public override void KillTile(int i, int j, ref bool fail, ref bool effectOnly, ref bool noItem)
+        {
+            int toDrop = 0;
+            switch (Main.tile[i, j].TileFrameX / 18)
```
Chicken-Bones: may as well just make this an if statement. You probably want to return if `fail` as well

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,119 @@
...
+    internal class ExampleCrystalGrow : GlobalTile
+    {
+        public override void RandomUpdate(int i, int j, int type)
+        {
+            if (Main.hardMode && type == TileID.Stone && j > Main.rockLayer && WorldGen.genRand.NextBool(110))
+                // this code line above means, that your crystal will grow only if: it's Hardmode, on Stone, in rock layer, and with same chance as vanilla Crystal Shards.
+            {
+                int num = WorldGen.genRand.Next(4);
+                int num2 = 0;
+                int num3 = 0;
```
Chicken-Bones: please give these variables sensible names

Chicken-Bones (commented):
Just needs some reformatting, and cleanup of the vanilla code.

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: alr

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: wdym here? like,
`fail = true`?
or like check?

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: alr, will try

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox:
```
TileObjectData.newTile.CoordinateWidth = 16; // Width of tile on spritesheet
            TileObjectData.newTile.CoordinateHeights = new int[] { 16 }; // Sets an offset
            TileObjectData.newTile.CoordinatePadding = 2; // Sets an "empty space" that should be between each 16x16 sprite in spritesheet
```
would that be alright?

JavidPack:
I'm looking at the vanilla TileID.Crystals code and this seems completely different. Did you adapt this from the code or just try things until it acted close enough to the vanilla tile? I'm seeing things like custom drawing in the code. Also, KillTile shouldn't be used to drop items, KillTile is used to determine if a tile will be killed, not necessarily if it is being killed.

BasicallyIAmFox:
yeah, i tried until it acted close enough to the vanilla tile
and vanilla crystals ofc would use custom drawing lol, it has gelatine crystal thing in it that summons qs
and it has additional 4 frames those used for something

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,118 @@
...
+                    additionalX = 1;
+                }
+                else if (coordinateChance == 0)
+                {
+                    additionalY = -1;
+                }
+                else
+                {
+                    additionalY = 1;
+                }
```
Jofairden: Merge these ifs
```
                if (coordinateChance == 0)
                {
                    additionalX = -1;
                    additionalY = -1;
                }
                else if (coordinateChance == 1)
                {
                    additionalX = 1;
                    additionalY = 1;
                }
```

Alternatively even
```
additionalX = !coordinateChance ? -1 : 1;
additionalY = additionalX;
```
But I guess that's too advanced for example mod

ExampleMod/Content/Tiles/ExampleCrystal.cs
Jofairden: Can't this just be done in the static defaults like example block?
https://github.com/tModLoader/tModLoader/blob/1.4/ExampleMod/Content/Tiles/ExampleBlock.cs#L15

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: trueee
alr, will commit another change now

BasicallyIAmFox: and wait, isn't `coordinateChance` an integer?
so you can't use `!coordinateChance` at this point

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: yeah, it can be

ExampleMod/Content/Tiles/ExampleCrystal.cs
Jofairden: Yes it's an integer, but booleans are integers too, really. False = 0, True = 1
So change it to 
```
additionalX = coordinateChance != 0 ? 1 : -1;
additionalY = additionalX;
```
and it would work (otherwise you would have to convert to boolean)

BasicallyIAmFox: trueeee, but `coordinateChance` is `WorldGen.genRand.Next(4)`, and `additional(X/Y)` default values are 0, so there should be a check for if `coordinateCheck` is not 2, like:
```
if (coordinateCheck != 2)
{
// code
}
```
alternatively,
```
additionalX = coordinateChance != 2 ? (coordinateChance != 0 ? 1 : -1) : 0;
additionalY = additionalX;
```

Jofairden: Your original if statement just checks 0 or 1, which is like a boolean true or false. I didn't look at the specifics of the function

BasicallyIAmFox: yeah, whatever

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,116 @@
...
+                if (coordinateChance == 0)
+                {
+                    additionalX = -1;
+                    additionalY = -1;
+                }
+                else if (coordinateChance == 1)
+                {
+                    additionalX = 1;
+                    additionalY = 1;
+                }
```
Jofairden: I've added some better comments and made the variables clearer in the tile placement section.
However, I am still bothered by this section. We're generating a random choice out of 4 options, but we only use 2 of them and in a weird way. It makes more sense to me if there are 4 options of using the coordinateChance, because the crystal could grown on any of the 4 sides a tile has (top, right, bottom, left). So the appropriate thing to do here would be do end up with 4 combinations:
```
x = 1, y = 0 (right)
x = -1, y = 0 (left)
x = 0, y = 1 (bottom)
x = 0, y = -1 (top)
```
I will leave this update to you as a tiny challenge :)

ExampleMod/Content/Tiles/ExampleCrystal.cs
```diff
@@ -0,0 +1,116 @@
...
+                    }
+					
+					// If there are less than 2 crystals, we can proceed to spawn one
+                    if (amountOfCrystals < 2)
+                    {
+                        int style = (short)WorldGen.genRand.Next(1);
+                        if (WorldGen.PlaceTile(i + additionalX, j + additionalY, TileID.Crystals, true, style: style)) // Places crystal shards first, because hardcoded!
+                        {
+                            Main.tile[i + additionalX, j + additionalY].TileType = ModContent.TileType<ExampleCrystal>(); // ... if succesfully placed, then replace with our crystal
+                            Main.Map.Update(i + additionalX, j + additionalY, Main.Map[i + additionalX, j + additionalY].Light); // also don't forget to update it on map, so it won't appear as crystal shards
```
Jofairden: I wonder what exactly is hardcoded here? It would help if you add the explanation of it in a comment. So why do we have to place a normal crystal first?

Jofairden (changes requested):

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: i took spreading thing from vanilla, so idk why its like that 🤷 
imma look in source again
EDIT: looked again, no difference
EDIT 2: looked in tml decompiler source, it's basically same code...
EDIT 3: i tested ingame and it seems like it worked just fine, so idk how it works at this point xd
EDIT 4: imma look what i can do with 4 combinations you gave me here

ExampleMod/Content/Tiles/ExampleCrystal.cs
BasicallyIAmFox: the placement (frame Y thing)

Jofairden: I dont get it? What is hardcoded so that we have to place a normal crystal first and replace it, instead of just placing our crystal ?

BasicallyIAmFox: изображение_2022-06-09_070044882
(im bad at explaining things)

its only thing i found in <= 10 secs, might be more code in there

Jofairden: I don't really get it. You will definitely have to fully research the vanilla crystal and port that over with 100% certainty to Example Mod

ExampleMod/Content/Tiles/ExampleCrystal.cs
Jofairden: Done.

PR #2520, opened 2022-06-08
Some Sound Bug Fixes

absoluteAquarian:
### What is the bug?
There were two bugs that this PR fixes.

The first bug:  
The Eye of Cthulhu was not playing the pitched roar (ID 36, Style -1) properly due to SoundID.ForceRoar not accounting for any styles whatsoever.
* Beforehand, any uses of ID 36 by vanilla would always use Style 1, even if a Style other than 1 was specified.

The second bug:  
Exiting a world would play a Lavafall sound perpetually until a world was re-entered.
* This was due to a previous fix to play the Waterfall (ID 34) and Lavafall (ID 35) sounds in `WorldGen.SaveAndQuitCallback()` with a Style of 0 in order to stop the sounds
* However, like the previous bug, the special "style = volume" logic was not handled whatsoever.  Using a Style of 0 did not play the Waterfall/Lavafall sounds at no volume, nor did it stop the existing sound (aka the intended behaviour).

### How did you fix the bug?
I modified the code that handles playing the `SoundID.Waterfall` and `SoundID.Lavafall` sounds in `Main.Ambience()`.  
I also edited the relevant `SoundStyle` initializations and retrieval from `SoundID.TML.cs` as part of of the bug fixes.

None that I can think of.

### DIFF
--- a/patches/tModLoader/Terraria/Audio/ActiveSound.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/ActiveSound.cs.patch
@@ -124,6 +124,19 @@
  					num = 1f - num2 / ((float)Main.screenWidth * 1.5f);
  				}
  
+@@ -95,6 +_,12 @@
+ 						break;
+ 					case SoundType.Ambient:
+ 						num *= Main.ambientVolume;
++
++						// Added by TML to mimic the behavior of the LegacySoundPlayer code.
++						if (Main.gameInactive) {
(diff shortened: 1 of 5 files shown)


Mirsario:
Also sneaked in a fix for Ambient sounds not being muted when game's window is out of focus.
(Only really relevant in multiplayer.)

PR #2523, opened 2022-06-08
Fix DamageClass.MeleeNoSpeed benefiting from all classes

ThomasThePencil:
### What is the bug?
`DamageClass.MeleeNoSpeed` benefits from all other classes, not just Generic & Melee

### How did you fix the bug?
checked for `damageClass == Generic || damageClass == Melee` before returnin' the appropriate `StatInheritanceData`, and returned `StatInheritanceData.None` otherwise

no


Jofairden:
Assigning @Chicken-Bones to check it

PR #2529, opened 2022-06-08
[1.4] Call OnWorldLoad after clearWorld (if it precedes GenerateWorld) to initialize ModSystems fully and restore 1.3 behavior

direwolf420:
### Description
This restores 1.3 behavior for `ModSystem.OnWorldLoad` (`ModWorld.Initialize` called within `clearWorld`, before `GenerateWorld`) in a way that doesn't move it into `clearWorld` (as that is also called in an "unload"-y context, which is handled by `ModSystem.OnWorldUnload` separately)

This is important so that world data is set up to possibly get populated during worldgen (i.e. saving special coordinates in a list)

"Use PreWorldGen for such purposes" just invites the modder for oversights while they develop their mod as they need to remember adding such code in both `OnWorldLoad` and `PreWorldGen` (i.e. the list needs to be populated through IO/Net, so it HAS to be initialized in `OnWorldLoad` aswell)

Please discuss and bring up counterpoints here, previous discussions about this were fragmented on the discord

### DIFF
--- a/patches/tModLoader/Terraria/IO/WorldFile.cs.patch
+++ b/patches/tModLoader/Terraria/IO/WorldFile.cs.patch
@@ -44,14 +44,14 @@
  			if (!FileUtilities.Exists(Main.worldPathName, flag) && Main.autoGen) {
  				if (!flag) {
  					for (int num = Main.worldPathName.Length - 1; num >= 0; num--) {
-@@ -321,6 +_,7 @@
- 				}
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
As long as there is perfect parity between `OnWorldLoad` and `OnWorldUnload` I don't care (invocations should alternate)

direwolf420:
closed in favor of #2532

PR #2532, opened 2022-06-09
[1.4] OnWorldLoad/PreWorldGen doc updates

direwolf420:
### Description
This PR updates the docs of `ModSystem.OnWorldLoad` and `ModSystem.PreWorldGen` to reference each other, as now you might need to duplicate some of your code if you were relying on 1.3s `ModWorld.Initialize` to run before worldgen. Also updated tModPorter to notify modders about this "silent" change.


JavidPack:
wait, why are we telling people to do things twice? why don't we just have a hook for initializing world data structures?

Issue #2348, opened 2022-05-05
[1.4] Missing ExampleMod content


ilija04:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

This is kinda a follow up to #1403
Here are currently missing example features:

1. ExamplePylon (Implemented in #2564)
2. ExampleKite
3. ExampleJoustingLance (Implemented in #2618)
4. ExampleWhip (Implemented in #1341)
5. ExamplePaperAirplane (Implemented in #2379)
6. ExampleSentry
7. ExampleBoomerang
8. ExampleFlask (Ones that give weapon imbues)
9. ExampleMowedBlock (Mowing a block with the Lawn Mower for grass blocks, for golfing or having a nice flat grass around your house)
10. ExampleGemTree
11. ExampleVanityTree
12. ExampleRope
14. ExamplePaint (Is it possible to add a custom paint color?)
15. ExampleGem (To go with ExampleGemTree)
16. ExampleRopeCoil (To go with ExampleRope)
17. ExampleCoating (Works like Echo and Illuminant Coatings in 1.4.4)

There's probably more, i can't think of anything else

### What does this proposal attempt to solve or improve?

It makes easier for modders to add new content listed above

### Which (other) solutions should be considered?

None

Rijam:
See #1341 and #2214 for Example Whips.

ilija04:
I know, @Rijam, but they haven't been reviewed for months! That's why I included it in the list

ilija04:
Oh, and I will occasionally edit this to add more things that are missing in EM that I think of.

Jenosis:
might I suggest ExampleVanityTree/ExampleGemTree for this list?

ilija04:
Added them both to the list.

ilija04:
Been a while updating this, but I added in ExampleCoating to the list

PR #2564, opened 2022-06-16
Modded Pylons: ModPylon & GlobalPylon

MutantWafflez:
### What is the new feature?

A grandiose injection into vanilla's Pylon system with two new, important classes: `ModPylon` and `GlobalPylon`. `ModPylon` is an extension of `ModTile` and allows for full flexibility of a `ModTile` alongside being able to make said tile a pylon in any capacity one sees fit. `GlobalPylon` is its own type that allows for modification of BOTH vanilla and modded pylons, allowing modders to modify whatever aspect that they see fit.

Adds a couple of new `TileID.Sets` that streamlines the process, since it was impossible to change some things due to some Vanilla hardcode (PreventsSandfall, CountsAsPylon). 

Right now, there is no way to add your own modded pylons without using IL editing or detouring, which when multiple pylon mods are added, can definitely lead to conflict. It is also basically impossible (and very tedious) to modify vanilla pylons, which this PR also helps alleviate. Combining everything under the umbrella of tML will allow for cohesive compatibility so nobody is stepping on any toes.

I could of definitely been a bit more strict on what exactly you can define as a `ModPylon`, but overall I think giving full freedom of making anything that can be a ModTile be a ModPylon is very powerful, granted the modder knows what they're doing.

Check out ExampleMod's new `ExamplePylonTile`: https://github.com/MutantWafflez/tModLoader/blob/1.4_ModPylons/ExampleMod/Content/Tiles/ExamplePylonTile.cs

### ExampleMod updates
Implementation of ExamplePylon also is one step closer to closing #2348 .


ilija04:
I'm surprised you didn't mention #2348.

patches/tModLoader/Terraria/GameContent/NetModules/NetTeleportPylonModule.cs.patch
```diff
@@ -0,0 +1,83 @@
...
++using Terraria.ModLoader;
+ using Terraria.Net;
+ 
+ namespace Terraria.GameContent.NetModules
+@@ -14,20 +_,28 @@
+ 		}
+ 
+ 		public static NetPacket SerializePylonWasAddedOrRemoved(TeleportPylonInfo info, SubPacketType packetType) {
+-			NetPacket result = NetModule.CreatePacket<NetTeleportPylonModule>(6);
++			NetPacket result = NetModule.CreatePacket<NetTeleportPylonModule>(ModNet.AllowVanillaClients ? 6 : 32);
```
Chicken-Bones: You should be able to avoid changes to this class, as well as `TeleportPylonInfo` and `TeleportPylonType` if you assign modded pylons ids. (And restrict to max 255 pylon types total)

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
  		private int _cooldownForUpdatingPylonsList;
  		private const int CooldownTimePerPylonsListUpdate = int.MaxValue;
 -		private SceneMetrics _sceneMetrics = new SceneMetrics();
 +		internal SceneMetrics _sceneMetrics = new SceneMetrics();
  
- 		public List<TeleportPylonInfo> Pylons => _pylons;
+-		public List<TeleportPylonInfo> Pylons => _pylons;
++		public List<TeleportPylonInfo> AllPylons => _pylons;
++		public List<TeleportPylonInfo> VanillaPylons => _vanillaPylons;
++		public List<TeleportPylonInfo> ModdedPylons => _moddedPylons;
```
Chicken-Bones: Suggest exposing these as `IReadOnlyList`

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
++using Terraria.ModLoader;
+ using Terraria.Net;
+ 
+ namespace Terraria.GameContent
+@@ -14,11 +_,15 @@
+ 	{
+ 		private List<TeleportPylonInfo> _pylons = new List<TeleportPylonInfo>();
  		private List<TeleportPylonInfo> _pylonsOld = new List<TeleportPylonInfo>();
++		private List<TeleportPylonInfo> _vanillaPylons = new List<TeleportPylonInfo>();
++		private List<TeleportPylonInfo> _moddedPylons = new List<TeleportPylonInfo>();
```
Chicken-Bones: Doesn't look like there's a need for vanilla/modded split here

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 		public void Update() {
+ 			if (Main.netMode != 1) {
+@@ -32,7 +_,12 @@
+ 			}
+ 		}
+ 
+-		public bool HasPylonOfType(TeleportPylonType pylonType) => _pylons.Any((TeleportPylonInfo x) => x.TypeOfPylon == pylonType);
++		public bool HasPylonOfType(TeleportPylonType pylonType, int tileType = TileID.TeleportationPylon) {
++			if (pylonType == TeleportPylonType.Modded && tileType >= TileID.Count) {
++				return _pylons.Any(x => !string.IsNullOrEmpty(x.ModName) && ModContent.Find<ModPylon>(x.ModName, x.ModPylonName).Type == tileType);
```
Chicken-Bones: I'm seeing a lot of `ModContent.Find<ModPylon>(p.ModName, p.ModPylonName)`
Suggest having a `ModPylon` field on `TeleportPylonType` instead.

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 		}
+ 
+ 		public void HandleTeleportRequest(TeleportPylonInfo info, int playerIndex) {
+@@ -93,7 +_,14 @@
+ 			}
+ 
+ 			if (flag) {
+-				flag &= !NPC.AnyDanger(quickBossNPCCheck: false, ignorePillars: true);
++				bool forceSucceed = false;
++				if (PylonLoader.ValidTeleportCheck_PreAnyDanger(info) is { } value) {
```
Chicken-Bones: use `is bool value` 
Also, documentation on `GlobalPylon.ValidTeleportCheck_PreAnyDanger` says
"This method is called before both vanilla and modded pylons check their Danger requirements, and returning false will prevent those checks from taking place,"

But the code here does not prevent the call to `ModPylon.ValidTeleportCheck_AnyDanger`

Suggest:
```cs
flag &= !NPC.AnyDanger(quickBossNPCCheck: false, ignorePillars: true);
flag = PylonLoader.ValidTeleportCheck_PreAnyDanger(info) is bool value ? value :
    info.modPylon?.ValidTeleportCheck_AnyDanger(info) ?? flag;
```

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 				if (!flag)
+ 					key = "Net.CannotTeleportToPylonBecauseNotMeetingBiomeRequirements";
+ 			}
++			if (info.TypeOfPylon == TeleportPylonType.Modded && ModContent.TryFind(info.ModName, info.ModPylonName, out ModPylon destinationPylon))
++				destinationPylon.ValidTeleportCheck_DestinationPostCheck(info, ref flag, ref key);
+ 
+-			if (flag) {
+-				bool flag2 = false;
++			TeleportPylonInfo nearbyInfo = new TeleportPylonInfo();
++			bool flag2 = false;
```
Chicken-Bones: Just keep a placeholder brace for indentation and patch maintenance.
Also do `if (flag || true) // tML: check for destination pylon even if no source pylon`

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 						if (DoesPylonAcceptTeleportation(info2, player)) {
+ 							flag2 = true;
++							nearbyInfo = info2;
+ 							break;
+ 						}
+ 					}
+ 				}
+ 
+ 				if (!flag2) {
+-					flag = false;
```
Chicken-Bones, line 84: comment out with a comment please

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 							break;
+ 					}
+ 				}
+-			}
++			if (info.TypeOfPylon == TeleportPylonType.Modded && ModContent.TryFind(nearbyInfo.ModName, nearbyInfo.ModPylonName, out ModPylon nearbyPylon))
++				nearbyPylon.ValidTeleportCheck_NearbyPostCheck(nearbyInfo, ref flag, ref key);
++			PylonLoader.PostValidTeleportCheck(info, nearbyInfo, ref flag, ref flag2, ref key);
+ 
+-			if (flag) {
++			if (flag && flag2) {
```
Chicken-Bones, line 98: in comparison, no comment required here. Clear from above proposed comments, and the fact this change is additive what it does

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 					switch (num) {
+ 						default:
+ 							key = "Net.CannotTeleportToPylonBecausePlayerIsNotNearAPylon";
+@@ -166,9 +_,11 @@
+ 							break;
+ 					}
+ 				}
+-			}
++			if (info.TypeOfPylon == TeleportPylonType.Modded && ModContent.TryFind(nearbyInfo.ModName, nearbyInfo.ModPylonName, out ModPylon nearbyPylon))
++				nearbyPylon.ValidTeleportCheck_NearbyPostCheck(nearbyInfo, ref flag, ref key);
```
Chicken-Bones: `ref flag2` here?

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -146,9 +146,19 @@
  		}
  
  		public void SetupShop(int type) {
-@@ -2611,6 +_,7 @@
+@@ -2610,7 +_,17 @@
+ 				if (Main.player[Main.myPlayer].ZoneGlowshroom && num < 39)
  					array[num++].SetDefaults(4921);
  			}
++			foreach (ModPylon pylon in ModContent.GetContent<ModPylon>()) {
```
Chicken-Bones: pretty sure this doesn't actually work. `ModContent` only works on the actual type of an instance, not on its parent types. 

Should just have `PylonLoader.ModPylons`

patches/tModLoader/Terraria/GameContent/TeleportPylonsSystem.cs.patch
```diff
@@ -1,11 +1,209 @@
...
+ 			_pylons.Clear();
++			_vanillaPylons.Clear();
++			_moddedPylons.Clear();
+ 			foreach (TileEntity value in TileEntity.ByPosition.Values) {
+-				TETeleportationPylon tETeleportationPylon = value as TETeleportationPylon;
+-				if (tETeleportationPylon != null && tETeleportationPylon.TryGetPylonType(out TeleportPylonType pylonType)) {
+-					TeleportPylonInfo teleportPylonInfo = default(TeleportPylonInfo);
+-					teleportPylonInfo.PositionInTiles = tETeleportationPylon.Position;
++				TeleportPylonType pylonType = TeleportPylonType.Modded;
++				if (value.GetType().IsDefined(typeof(IsPylonEntityAttribute), true) || (value is TETeleportationPylon tETeleportationPylon && tETeleportationPylon.TryGetPylonType(out pylonType)))  {
```
Chicken-Bones: `IPylonTileEntity` marker interface rather than attribute please. Add it to vanilla's TE too. There's also really no reason for `TryGetPylonTypeFromTileCoords` in `TETeleportationPylon` not be `static`. And you could get the pylon type from the modded tile there.

patches/tModLoader/Terraria/IO/WorldFile.cs.patch
```diff
@@ -235,3 +235,12 @@
  					TileEntity.Write(writer, item.Value);
  				}
  			}
+@@ -2262,7 +_,7 @@
+ 			foreach (KeyValuePair<Point16, TileEntity> item in TileEntity.ByPosition) {
+ 				if (!WorldGen.InWorld(item.Value.Position.X, item.Value.Position.Y, 1))
+ 					list.Add(item.Value.Position);
+-				else if (!TileEntity.manager.CheckValidTile(item.Value.type, item.Value.Position.X, item.Value.Position.Y))
++				else if (!TileEntity.manager.CheckValidTile(item.Value.type, item.Value.Position.X, item.Value.Position.Y) && item.Value is not TETeleportationPylon)
```
Chicken-Bones: What does this do?

patches/tModLoader/Terraria/ModLoader/ModdedPylonsMapLayer.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.Xna.Framework;
+using System.Collections.Generic;
+using Terraria.GameContent;
+using Terraria.Map;
+
+namespace Terraria.ModLoader
+{
+	internal class ModdedPylonsMapLayer : IMapLayer
```
Chicken-Bones: If you make this `ModMapLayer` then you don't have to edit `Main.MapLayers`

Additionally... can we not combine this into vanilla `TeleportPylonsMapLayer`? I think `ModPylon.DrawMapIcon` should have an easier to use implementation where they just return the texture/color, and it's otherwise fully vanilla drawing.

patches/tModLoader/Terraria/ModLoader/IO/TileIO_TEs.cs
```diff
@@ -65,7 +66,19 @@ internal static partial class TileIO
 					}
 				}
 				//Otherwise, if the TE is vanilla, try to find its existing instance for the current coordinate.
-				else if (!TileEntity.ByPosition.TryGetValue(point, out tileEntity)) {
+				else if (TileEntity.ByPosition.TryGetValue(point, out tileEntity)) {
```
Chicken-Bones: hm... I don't really like this but I don't have a great solution right now. Perhaps we should make all modded pylons use a common, tML provided, modded TE?

Yes, I think you have to do this, because there's also 'unloaded tiles' to consider. With modded tiles and tile entities, pylons should survive a world load without the mod enabled.

patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -491,7 +492,7 @@ public static class ModContent
 			Config.ConfigManager.Unload();
 			CustomCurrencyManager.Initialize();
 			EffectsTracker.RemoveModEffects();
-			Main.MapIcons = new MapIconOverlay().AddLayer(new SpawnMapLayer()).AddLayer(new TeleportPylonsMapLayer()).AddLayer(Main.Pings);
+			Main.MapIcons = new MapIconOverlay().AddLayer(new SpawnMapLayer()).AddLayer(new TeleportPylonsMapLayer()).AddLayer(new ModdedPylonsMapLayer()).AddLayer(Main.Pings);
```
Chicken-Bones: See other comment about the map layer, I'd prefer not to have to keep this in sync with `Main.MapIcons`

patches/tModLoader/Terraria/NetMessage.cs.patch
```diff
@@ -372,6 +377,15 @@
  							short num11 = (short)Sign.ReadSign(j, i);
  							if (num11 != -1)
  								array2[num2++] = num11;
+@@ -1738,7 +_,7 @@
+ 								array3[num3++] = (short)num18;
+ 						}
+ 
+-						if (tile2.type == 597 && tile2.frameX % 54 == 0 && tile2.frameY % 72 == 0) {
++						if (TileID.Sets.CountsAsPylon.Contains(tile2.type) && tile2.frameX % 54 == 0 && tile2.frameY % 72 == 0) {
```
Chicken-Bones: If you use a modded pylon TE for all pylons then some of these `CountsAsPylon.Contains` patches should disappear

Chicken-Bones (commented):
Great work on this. Lots of feedback to look through. I haven't scrutinized the examples, ModPylon, GlobalPylon or PylonLoader classes yet, but this should be plenty for another review

cozykoko:
merging this should really be a priority imo

patches/tModLoader/Terraria/ModLoader/GlobalPylon.cs
```diff
@@ -0,0 +1,118 @@
...
+		/// Called right before both modded and vanilla pylons determine whether or not they can be placed. Returning false will prevent vanilla checks from taking place,
+		/// forcefully preventing the pylon from being placed. Returning null will enact vanilla behavior, and returning true will make the check forcefully succeed,
+		/// regardless of vanilla behavior. Returns null by default.
+		/// </summary>
+		/// <param name="x"> The current X position of where the tile in question is asking it can be placed. </param>
+		/// <param name="y"> The current Y position of where the tile in question is asking it can be placed. </param>
+		/// <param name="tileType"> The actual type of tile currently asking if it can be placed. </param>
+		/// <param name="pylonType"> The type of pylon belonging to this tile. </param>
+		/// <returns></returns>
+		public virtual bool? PreCanPlacePylon(int x, int y, int tileType, byte pylonType) {
```
Chicken-Bones: `byte` should be `TeleportPylonInfo`, as in the other hooks.

ExampleMod/Common/GlobalPylons/ExampleGlobalPylon.cs
```diff
@@ -0,0 +1,74 @@
...
+			}
+
+			//Since we aren't actually preventing the drawing of any map icons, we can just return the default value, which in this case is null (AKA vanilla behavior)
+			return base.PreDrawMapIcon(ref context, ref mouseOverText, ref pylonInfo, ref isNearPylon, ref drawColor, ref deselectedScale, ref selectedScale);
+		}
+
+		public override bool? PreCanPlacePylon(int x, int y, int tileType, byte pylonType) {
+			//What if we want to override the functionality for pylon placement?
+			//For example, let's always allow the players to place universal pylons, even if they already exist in the world:
+			if (pylonType == (byte)TeleportPylonType.Victory) {
```
Chicken-Bones: comment on how to do this check for a mod pylon?

I think you need to add `ModContent.PylonType` and it should probably return `TeleportPylonType` since we're using the enum everywhere.

patches/tModLoader/Terraria/ModLoader/ModPylon.cs
```diff
@@ -0,0 +1,178 @@
...
+	/// <br></br>
+	/// 9) Any <seealso cref="GlobalPylon"/> instances run <seealso cref="GlobalPylon.PostValidTeleportCheck"/>.
+	/// </remarks>
+	public abstract class ModPylon : ModTile
+	{
+
+		/// <summary>
+		/// What type of Pylon this ModPylon represents.
+		/// </summary>
+		public byte PylonType {
```
Chicken-Bones: similar to above comments, we either have to use `TeleportPylonType` everywhere or no-where. Fortunately, `Enum.ToString` does work with values outside the defined list, they just print as numbers. Though the first modded one might end up printing as 'Count' which could be confusing (unless we start modded above Count)

ExampleMod/Common/GlobalPylons/ExampleGlobalPylon.cs
```diff
@@ -0,0 +1,74 @@
...
+
+			return base.PreCanPlacePylon(x, y, tileType, pylonType);
+		}
+
+		public override bool? ValidTeleportCheck_PreBiomeRequirements(TeleportPylonInfo pylonInfo, SceneMetrics sceneData) {
+			//What if we want to do something based on the type of pylon in particular? Well all we have to do is check the pylon's type!
+			//Let's swap around the biome requirements for the Jungle Pylon and Snow Pylon, for example:
+			TeleportPylonType pylonType = pylonInfo.TypeOfPylon;
+
+			if (pylonType == TeleportPylonType.Jungle) {
```
Chicken-Bones: A better example would probably be to allow jungle pylon in the snow (as well as jungle). With this example, the false return will cancel out any other mod's attempt to make the jungle pylon more permissible

patches/tModLoader/Terraria/ModLoader/Default/TEModdedPylon.cs
```diff
@@ -0,0 +1,106 @@
...
+				ByPosition.Remove(new Point16(x, y));
+			}
+
+			Main.PylonSystem.RequestImmediateUpdate();
+		}
+
+		public override string ToString() => Position.X + "x  " + Position.Y + "y";
+
+		public override bool IsTileValidForEntity(int x, int y) {
+			return Main.tile[x, y].active() && TileID.Sets.CountsAsPylon.Contains(Main.tile[x, y].type) && Main.tile[x, y].frameY == 0 && Main.tile[x, y].frameX % 54 == 0;
```
Chicken-Bones: Should `% 54` be hardcoded here? Also do we need to check `CountsAsPylon`? Can we do `GetModPylonFromCoords` instead, that would give us the tile to query framing info on if we decided to put it there.

patches/tModLoader/Terraria/ModLoader/Default/TEModdedPylon.cs
```diff
@@ -0,0 +1,106 @@
...
+			moddedPylon.ID = AssignNewID();
+			moddedPylon.type = (byte)Type;
+
+			lock (EntityCreationLock) {
+				ByID[moddedPylon.ID] = moddedPylon;
+				ByPosition[moddedPylon.Position] = moddedPylon;
+			}
+
+			return moddedPylon.ID;
+		}
```
Chicken-Bones, line 45: This looks duplicated from `ModTileEntity.Place`, but without the benefit of cloning the `Mod` into it... If you just use that one then you shouldn't need to set `Mod` above

ExampleMod/Content/TileEntities/AdvancedPylonTileEntity.cs
```diff
@@ -0,0 +1,83 @@
...
+				Main.PylonSystem.RequestImmediateUpdate(); //This line is very important, don't forget it, or these entities won't register with vanilla's pylon system!
+			}
+
+			return placedEntityID;
+		}
+
+		public override void OnNetPlace() {
+			//This hook is only ever called on the server; its purpose is to give more freedom in terms of syncing FROM the server to clients, which we take advantage of
+			//by making sure to sync whenever this hook is called:
+			NetMessage.SendData(MessageID.TileEntitySharing, number: ID, number2: Position.X, number3: Position.Y);
```
Chicken-Bones, line 26: `Main.PylonSystem.RequestImmediateUpdate` doesn't get called here, like it does for `ModPylonTileEntity` and `TETeleportationPylon`

patches/tModLoader/Terraria/ModLoader/Default/TEModdedPylon.cs
```diff
@@ -0,0 +1,106 @@
...
+{
+	/// <summary>
+	/// This is a TML provided Tile Entity that acts extremely similar to vanilla's pylon TEs. If you plan
+	/// to make a pylon tile that closely resembles (or is identical to) vanilla's pylon TEs, you should use
+	/// this.
+	/// </summary>
+	public class TEModdedPylon : ModTileEntity, IPylonTileEntity
+	{
+
+		public override void NetPlaceEntityAttempt(int x, int y) {
```
Chicken-Bones, line 14: it would be nice if this could use more of the base impl. Also neither this nor `AdvancedPylonTileEntity` check `Main.PylonSystem.HasPylonOfType` like `TETeleportationPylon` does

Chicken-Bones (changes requested):
Looking much better. Didn't quite get through a full review, but the main thing I want to see is more code shared between `AdvancedPylonTileEntity`, `TEModdedPylon`, `ModTileEntity` and `TETeleportationPylon`

Ideally, `AdvancedPylonTileEntity` should extend from `TEModdedPylon`

ExampleMod/Content/TileEntities/AdvancedPylonTileEntity.cs
```diff
@@ -0,0 +1,72 @@
...
+			TileDimensions = ExamplePylonTileAdvanced.TileDimensions;
+		}
+
+		public override bool IsTileValidForEntity(int x, int y) {
+			Tile tile = Framing.GetTileSafely(x, y);
+
+			//This check is only if your tile has ONE style. If you plan on adding more styles, make sure to use modular arithmetic to ensure things are valid.
+			//For example, if your pylon tile has two styles and is 2 tiles wide, you should make sure tile.TileFrameX % 36 == 0. (assuming 18 pixels per tile)
+			return tile.HasTile && tile.TileFrameX == 0 && tile.TileFrameY == 0;
+		}
```
Chicken-Bones: No need for this override in this example.

patches/tModLoader/Terraria/ModLoader/Default/TEModdedPylon.cs
```diff
@@ -0,0 +1,114 @@
...
+
+		/// <summary>
+		/// This is the actual dimensions of the tile that this entity is attributed to. This exists so that multiplayer
+		/// syncing properly works, as sending the tile square requires a defined area to actually send to the clients.
+		/// If you are simply making a pylon that is the same size as a vanilla pylon (3x4 tiles), you will not need to change this.
+		/// </summary>
+		public Point16 TileDimensions {
+			get;
+			protected set;
+		} = new Point16(3, 4);
```
Chicken-Bones: Can you not get these from `TileObjectData.GetTileData(Tile)`?

ExampleMod/Content/Tiles/ExamplePylonTile.cs
```diff
@@ -0,0 +1,237 @@
...
+			//Basically, we only want the crystal to be drawn once, based off the top left corner, which is what this check is.
+			//The top left corner of a Pylon will have its FrameX divisible by 54 (since it's 3 tiles wide, and each tile on the tile sheet takes up 18x18 pixels),
+			//and its FrameY will be 0, since it's at the top of the tile sheet.
+			if (drawData.tileFrameX % 54 == 0 && drawData.tileFrameY == 0) {
+				//This method call basically says "Run SpecialDraw once at this position"
+				Main.instance.TilesRenderer.AddSpecialLegacyPoint(i, j);
+			}
+		}
+
+		public override void SpecialDraw(int i, int j, SpriteBatch spriteBatch) {
```
Chicken-Bones, line 113: this kind of stuff should go in advanced example. If this is the _default_ implementation, it should be in `ModPylon` (also applies to `DrawEffects, HasSmartInteract, RightClick` above)

ExampleMod/Content/Tiles/ExamplePylonTile.cs
```diff
@@ -0,0 +1,237 @@
...
+
+			//Finally, draw the highlight texture, if applicable.
+			int colorPotency = (crystalColor.R + crystalColor.G + crystalColor.B) / 3;
+			if (colorPotency > 10) {
+				Color selectionGlowColor = Colors.GetSelectionGlowColor(selectionType == 2, colorPotency);
+				Main.spriteBatch.Draw(crystalTexture.Value, drawPos - Main.screenPosition, highlightFrame, selectionGlowColor, 0f, origin, 1f, SpriteEffects.None, 0f);
+			}
+		}
+
+		public override void DrawMapIcon(ref MapOverlayDrawContext context, ref string mouseOverText, TeleportPylonInfo pylonInfo, bool isNearPylon, Color drawColor, float deselectedScale, float selectedScale) {
```
Chicken-Bones, line 118: This can stay, I guess. Still an annoying amount of code to duplicate in every pylon, especially if vanilla were to change something about pylons later. Be nicer if there was a default impl you could pass the icons to, maybe 2 functions?
`DefaultDraw` which returns `bool isMouseOver` and `TeleportFromMap`

ExampleMod/Content/Tiles/ExamplePylonTileAdvanced.cs
```diff
@@ -0,0 +1,243 @@
...
+				destinationPylonValid = false;
+				errorKey = "Mods.ExampleMod.MessageInfo.NearbyUnstablePylonIsOff";
+			}
+		}
+
+		public override void DrawEffects(int i, int j, SpriteBatch spriteBatch, ref TileDrawInfo drawData) {
+			//This time, we'll ONLY draw the crystal if the pylon is active.
+			if (drawData.tileFrameX % 36 == 0 && drawData.tileFrameY == 0 && TileEntity.ByPosition[new Point16(i, j)] is AdvancedPylonTileEntity { isActive: true }) {
+				Main.instance.TilesRenderer.AddSpecialLegacyPoint(i, j);
+			}
```
Chicken-Bones, line 135:
```suggestion
			if (TileEntity.ByPosition[new Point16(i, j)] is AdvancedPylonTileEntity { isActive: true }) {
				base.DrawEffects(i, j, spriteBatch, ref drawData);
			}
```

ExampleMod/Content/Tiles/ExamplePylonTileAdvanced.cs
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		public override void SpecialDraw(int i, int j, SpriteBatch spriteBatch) {
```
Chicken-Bones, line 138: Suggest moving the detailed example from `ExamplePylonTile` to here, with a message at the top "this is just a copy of the default implementation, allowing for adjustment if you want something more complex"

ExampleMod/Content/Tiles/ExamplePylonTileAdvanced.cs
```diff
@@ -0,0 +1,243 @@
...
+
+		public override void DrawMapIcon(ref MapOverlayDrawContext context, ref string mouseOverText, TeleportPylonInfo pylonInfo, bool isNearPylon, Color drawColor, float deselectedScale, float selectedScale) {
+			if (TileEntity.ByPosition[pylonInfo.PositionInTiles] is not AdvancedPylonTileEntity entity) {
+				//If for some reason we don't find the tile entity, we won't draw anything.
+				return;
+			}
+
+			//Depending on the whether or not the pylon is active, the color of the icon will change;
+			//otherwise, it acts as normal.
+			drawColor = entity.isActive ? Color.Green : Color.Red;
```
Chicken-Bones: This example works great with my proposed helper functions

ExampleMod/Content/Tiles/ExamplePylonTileAdvanced.cs
```diff
@@ -0,0 +1,243 @@
...
+			Main.cancelWormHole = true;
+			mouseOverText = Language.GetTextValue("Mods.ExampleMod.ItemName.ExamplePylonItemAdvanced");
+
+			if (Main.mouseLeft && Main.mouseLeftRelease) {
+				Main.mouseLeftRelease = false;
+				Main.mapFullscreen = false;
+				PlayerInput.LockGamepadButtons("MouseLeft");
+				Main.PylonSystem.RequestTeleportation(pylonInfo, Main.LocalPlayer);
+				SoundEngine.PlaySound(SoundID.Item6);
+			}
```
Chicken-Bones: I wonder how this works in multiplayer if the advanced pylon is off-screen in an area you haven't yet visited.

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -146,9 +146,19 @@
...
  			}
++			foreach (ModPylon pylon in PylonLoader.modPylons) {
++				int? pylonReturn = pylon.IsPylonForSale(type, Main.LocalPlayer, flag);
++				if (pylonReturn.HasValue && num < 39) {
++					array[num++].SetDefaults(pylonReturn.Value);
++				}
++				else if (pylonReturn.HasValue && num >= 39) {
++					Logging.tML.Warn($"Ran out of NPC shop space for {pylon.Mod.Name}'s pylon item {pylon.Name}");
++				}
++			}
```
Chicken-Bones: It would be good to move this to a tML ModNPC subclass in the `SetupShop` hook. That way it will work with mods that expand the shop

Chicken-Bones: Also replace the hardcoded `39` with `shop.item.Length`

patches/tModLoader/Terraria/ModLoader/Default/TEModdedPylon.cs
Chicken-Bones: Do we need to use `% 54` here? Can we use the tile object data to get a modulus? Or just not worry

Chicken-Bones (changes requested):
Should be the final review pass!

Chicken-Bones (approved):

Issue #2543, opened 2022-06-11
[1.4] Deerclops' ice spikes attack sounds are too low


Polaritz:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Deerclops ice spikes attack sound is hardly audible

### Steps to reproduce

1. spawn deerclops
2. wait until it uses ice spike attack

### Expected Behavior

Ice spike attack sounds play normally

### Actual Behavior

Ice spike attack sounds are way too low compared to vanilla

### Reproduction frequency

always

Mirsario:
Cannot reproduce in modern TML, the volume is exactly the same as in vanilla. Perhaps someone fixed this and forgot to close the issue.

Issue #2498, opened 2022-06-04
Eye of Cthulhu incorrect sound played when dashing


Nobisyu12:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

In Expert Mode and above when the Eye of Cthulhu does it's special faster dash it plays the normal roar sound effect. In vanilla, it plays a shorter and more high pitched version of the normal roar. Probably a simple mistake when making the new sound overhaul if I had to guess.

### Steps to reproduce

1. Spawn Eye of Cthulhu in Expert Mode or above difficulties
2. Get it to second form
3. Wait until it does the special dash
4. It plays the wrong sound

### Expected Behavior

It should play the correct sound

### Actual Behavior

It plays the wrong sound

### Reproduction frequency

100% of the time

Mirsario:
This has been fixed on `Jun 15, 2022` in [`#2520`](https://github.com/tModLoader/tModLoader/pull/2520), but I forgot to close this issue.

PR #2566, opened 2022-06-17
More sound fixes and new SoundStyle options

### LINKED ISSUE #2891: Minecart skid sounds are low-pitched
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

The minecart skid sound that plays when the player tries to move to the other direction or slow down while riding a minecart plays in a lower pitch than what it should sound like.
Here's the list of mods I'm using:
image
None of them seem to change the way minecarts are modified.

### Log File

_No response_

### Steps to reproduce

1. Create a new world and character
2. Go down to a railroad underground
3. Right click on the rails or press r near the railroad to enter your minecart
4. Move anywhere and then try to move to the opposite direction while still on the minecart
5. Minecart skid sounds emit like usual, but they are pitched down.

### Expected Behavior

Minecart skid sounds play as normal with no pitch differences, as expected in Vanilla Terraria.

### Actual Behavior

Minecart skid sounds play with a lower pitch.

### Reproduction frequency

100% of the time

### Additional Information

I don't know if this is intentional. If it is, then you may disregard this issue.

### LINKED ISSUE #3091: Autopause cuts off sounds
### Version

Latest 1.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When autopause is enabled any sound that is active wnen game pauses gets cut off

### Log File

_No response_

### Steps to reproduce

Enable autopause
Open inventory
Sound is playing only on the same tick as inventory opens (looks like Main.CanPauseGame runs before inventory key press check) then cuts off

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

All of the time

### Additional Information

_No response_

Chicken-Bones:
# What is the bug?
A fair few legacy sound style parameters were being ignored, and many other inconsistencies with vanilla.

# Fixes/Changes

## Many vanilla SoundStyle fixes
There are many fixes for incorrect sound behaviors to help them match Terraria behavior. The most crucial fixes are fixes to sounds incorrectly limiting or interrupting each other. Several vanilla sounds were incorrectly seen as the same as other sounds, limiting their ability to play at the same time. Some were limited to only playing 1 instance when vanilla behavior was to play as many instance as needed. High profile examples include zombie moans, coin sounds, bird sounds, meowmere sounds, flamethrower, and star cannon/falling star sounds. There were also some incorrect pitch values found and fixes as well. There were also some sounds that were supposed to be Ambient but weren't.

## Pause game behavior and SoundStyle.PauseBehavior
Before this PR, ALL sounds in tModLoader would be stopped when the game was paused. This is a bug (Fixes #3091). This was most noticeable with the `MenuOpen` sound when pausing the game. This did not match Terraria. This has been fixed, now sounds will continue playing as they do in Terraria. 

If you want the old behavior, set `SoundStyle.PauseBehavior` to `PauseBehavior.StopWhenGamePaused`. `PauseBehavior.PauseWithGame` is another option added in this PR which will resume the sound instance when resuming the game. `ActiveSoundShowcaseProjectile` has been updated to use these new options since it uses long sounds.

## SoundStyle.LimitsArePerVariant
A big change of this PR is the addition of `SoundStyle.LimitsArePerVariant` to indicate if the variants of a `SoundStyle` should be considered as separate sounds rather than the same sound for the purposes of `SoundLimitBehavior` and `MaxInstances`. For example, `Clown` does not use this, only 1 random sound of the 3 variants can be playing at a time, but the vanilla logic for `Toucan` says that each of the random sound variants can be playing at once, as if they were separate sounds.

To implement `LimitsArePerVariant`, the random choice decision had to be moved earlier in the logic to precede the `SoundStyle.IsTheSameAs` check.

## SoundStyle.IsTheSameAs
`IsTheSameAs` has been overhauled. Previously 2 sounds with different `Identifier`s but the same `SoundPath` would both evaluate to true ("the same"). This logic is incorrect, an `Identifier` should indicate that 2 `SoundStyle`s are separate even if their `SoundPath` matches. This bug would mistakenly prevent sounds using the `SoundPath` of "{Prefix}Zombie_" (but different `Identifier`s) from playing at the same time. Also, the `SelectedVariant` (randomly selected variant) is also taken into account during `IsTheSameAs` if `LimitsArePerVariant` is also true. 2 `SoundStyle` using `LimitsArePerVariant` with different `SelectedVariant` shouldn't be seen as "the same", allowing them both to coexist. The new `IsVariantOf` method is the same as `IsTheSameAs` but ignores the selected variants.

## SoundStyle.PitchRange getter
This was wrong and has been fixed.

## SoundStyle.GetRandomSound renamed to GetSoundEffect
The new name is more correct.

## SoundStyle.RerollAttempts 
`RerollAttempts` has been added, it allows a `SoundStyle` to try again if the randomly selected variant is already playing in an effort to find a variant that isn't actively playing.

# Porting Notes

- In fixing various bugs, this PR does change some existing behaviors some modders might have unwittingly relied on. We recommend that modders test the sounds in their mod that relate to the following:
- Sounds now continue playing when the game is paused by default. If this is not desired, consider setting setting `SoundStyle.PauseBehavior` to `PauseBehavior.StopWhenGamePaused` (or `PauseBehavior.PauseWithGame`).
- Various changes to `IsTheSameAs` now means that some `SoundStyle` that previously were seen as the same are now seen as different and vice versa. If you have multiple `SoundStyle` using the same `SoundPath` but want them to behave independently for sound limiting purposes, make sure to set `Identifier`. 
- Consider setting `SoundStyle.LimitsArePerVariant` if making use of the variants feature to allow each variant to have an independent instance limit. 

# TODOs
- [x] A lot of sounds need ~~`SoundLimitBehavior.Unlimited`~~ `MaxInstances` of 0.
- A lot of sounds need a 'slot per variant' Currently for e.g., all the `Zombie_` sounds share the same 'slot' and will replace/cancel each-other.
    - [x] Addressed via identifier spam for now.
- [x] The weights for the owl variants aren't quite right.
- [x] The owl variants have an unaddressed mixed behavior where the 'special' owl sounds don't stop other owl sounds from playing. I'd be fine with them all just being IgnoreNew
- [x] Fixes #2891 
- ~~#2498~~
- ~~#2543~~


direwolf420:
Another sound worth unlimiting is the sound fallen stars make when they drop down. If multiple are falling (very common with 1.4 increasing their spawn rate massively) the sounds get cut off which sounds really bad

Reportedly, the sounds  star cannon, super star shooter and star wrath are making are also affected by this.

Mirsario (approved):

patches/tModLoader/Terraria/ID/SoundID.TML.cs
```diff
@@ -64,31 +66,48 @@ partial struct SoundStyleDefaults
...
+	public static readonly SoundStyle Bird17 = Bird with { Variants = [17] };
+	public static readonly SoundStyle Bird18 = Bird with { Variants = [18] };
+	public static readonly SoundStyle Bird19 = Bird with { Variants = [19] };
 	public static readonly SoundStyle Critter = new($"{Prefix}Zombie_15", SoundType.Ambient) { Volume = 0.2f, PitchRange = (-0.1f, 0.3f), SoundLimitBehavior = IgnoreNew };
 	public static readonly SoundStyle Waterfall = new($"{Prefix}Liquid_0", SoundType.Ambient) { Volume = 0.2f, SoundLimitBehavior = IgnoreNew };
 	public static readonly SoundStyle Lavafall = new($"{Prefix}Liquid_1", SoundType.Ambient) { Volume = 0.65f, SoundLimitBehavior = IgnoreNew };
-	public static readonly SoundStyle ForceRoar = new($"{Prefix}Roar_0") { Identifier = "Terraria/Roar" };
-	public static readonly SoundStyle ForceRoarPitched = new($"{Prefix}Roar_0") { Pitch = 0.6f, Identifier = "Terraria/Roar" };
+	public static readonly SoundStyle ForceRoar = new($"{Prefix}Roar_0");
+	public static readonly SoundStyle ForceRoarPitched = new($"{Prefix}Roar_0") { Pitch = 0.6f };
```
Chicken-Bones: Removing these identifiers means you also need to remove the identifier from `Roar`

patches/tModLoader/Terraria/ID/SoundID.TML.cs
```diff
@@ -64,31 +66,48 @@ partial struct SoundStyleDefaults
...
-	// Explanation: There is a 1 in 300 chance for an owl to play a 'funni' easter egg sound variant.
-	public static readonly SoundStyle Owl = new($"{Prefix}Zombie_", stackalloc (int, float)[] { (110, 300f), (111, 300f), (112, 1f), (113, 1f), (114, 1f), }) {
-		PitchVariance = 0.2f
+	// There is a 1 in 300 chance for an owl to play one of 3 easter egg sound variants, one of which is rolled for before the other two.
+	public static readonly SoundStyle Owl = new($"{Prefix}Zombie_", stackalloc (int, float)[] {
+		(110, (299f / 300f) * (1f / 2f)),
+		(111, (299f / 300f) * (1f / 2f)),
+		(112, (1.0f / 300f) * (2f / 3f) * (1f / 2f)),
+		(113, (1.0f / 300f) * (2f / 3f) * (1f / 2f)),
+		(114, (1.0f / 300f) * (1f / 3f)),
```
Chicken-Bones: These can all be `1/300f * 1/3f`

patches/tModLoader/Terraria/ID/SoundID.TML.cs
```diff
@@ -64,31 +66,48 @@ partial struct SoundStyleDefaults
...
+	}) {
+		PitchVariance = 0.2f,
+		SoundLimitBehavior = IgnoreNew,
```
Chicken-Bones, line 113: Needs an `Identifier`

patches/tModLoader/Terraria/ID/SoundID.TML.cs
```diff
@@ -108,7 +127,7 @@ partial struct SoundStyleDefaults
 	public static readonly SoundStyle DrumFloorTom = new($"{Prefix}Item_148") { Volume = 0.7f, Identifier = "Terraria/Drums" };
 	public static readonly SoundStyle Research = new($"{Prefix}Research_", 1, 3);
 	public static readonly SoundStyle ResearchComplete = new($"{Prefix}Research_0");
-	public static readonly SoundStyle QueenSlime = new($"{Prefix}Zombie_", 115, 3) { Volume = 0.5f, SoundLimitBehavior = IgnoreNew };
+	public static readonly SoundStyle QueenSlime = new($"{Prefix}Zombie_", 115, 3) { Volume = 0.5f, SoundLimitBehavior = IgnoreNew, Identifier = "Terraria/QueenSlime" };
 	public static readonly SoundStyle Clown = new($"{Prefix}Zombie_", 121, 3) { Volume = 0.45f, PitchVariance = 0.15f, SoundLimitBehavior = IgnoreNew };
```
Chicken-Bones: Needs an `Identifier`

patches/tModLoader/Terraria/ID/SoundID.TML.cs
```diff
@@ -680,18 +705,17 @@ LegacySoundIDs.NPCHit or
...
 		LegacySoundIDs.Duck => Duck,
 		LegacySoundIDs.Frog => Frog,
-		LegacySoundIDs.Bird => Bird,
 		LegacySoundIDs.Critter => Critter,
 		LegacySoundIDs.Waterfall => Waterfall,
 		LegacySoundIDs.Lavafall => Lavafall,
 		LegacySoundIDs.ForceRoar => style switch { -1 => ForceRoarPitched, _ => ForceRoar },
-		LegacySoundIDs.Meowmere => Meowmere,
+		LegacySoundIDs.Meowmere => Meowmere with { Volume = style * 0.05f },
 		LegacySoundIDs.CoinPickup => CoinPickup,
```
Chicken-Bones: These two need to handle the `Main.starGame` case

Chicken-Bones (commented):
Added some change requests. `ShimmerWeak1` and `ShimmerWeak2` also need to share an identifier.

patches/tModLoader/Terraria/ID/SoundID.TML.cs
Mirsario: This bit was done by you. It seems to be more correct to be adding `MaxInstances = 0` while keeping the identifier, will do that.

Chicken-Bones: `ForceRoar` and `ForceRoarPitched` are `ReplaceOldest`. `Roar` is `IgnoreNew`. All use the same slot

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -75,6 +75,12 @@ public int RerollAttempts {
 	/// <summary> If true, this sound won't play if the game's window isn't selected. </summary>
 	public bool PlayOnlyIfFocused { get; set; } = false;
 
+	/// <summary>
+	/// If true, this sound will pause when the game is paused or unfocused.
+	/// <br/><br/> Currently these sounds will not resume when the game un-pauses, this might be a bug.
+	/// </summary>
+	public bool PauseWhenGamePausedOrUnFocused { get; set; } = false;
```
Chicken-Bones: Perhaps just `PauseWithGame`

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -58,6 +59,14 @@ public record struct SoundStyle
 	/// <summary> Determines what the action taken when the max amount of sound instances is reached. </summary>
 	public SoundLimitBehavior SoundLimitBehavior { get; set; } = SoundLimitBehavior.ReplaceOldest;
 
+	/// <summary>
+	/// How many times to attempt to find a variant that is not currently playing before applying the SoundLimitBehavior. Only has effect if LimitsArePerVariant is true. Defaults to 1.
+	/// </summary>
+	public int RerollAttempts {
```
Chicken-Bones, line 82: Since it's called _Re_roll attempts, it should default to 0, and thunder rerolls 6 times

patches/tModLoader/Terraria/ID/SoundID.cs.patch
```diff
@@ -119,7 +119,7 @@
 +		/*
  		return new LegacySoundStyle(42, count, variations, type);
 +		*/
-+		return new LegacySoundStyle($"{PrefixCustom}{name}{(variations > 1 ? "_" : null)}", 0, variations, type);
++		return new LegacySoundStyle($"{PrefixCustom}{name}{(variations > 1 ? "_" : null)}", 0, variations, type) with { MaxInstances = 0 };
```
Chicken-Bones: No need to use `with` here, you're constructing :)

Chicken-Bones (commented):
lgtm

PR #2579, opened 2022-06-19
Remove British spelling of endeavor

tristan-f-r:
This aims to remove the British spelling of endeavour to endeavor -- other words such as `license` are spelled with the american dialect, so switching this word will make it more consistent.


Chicken-Bones:
Sure I guess

PR #2583, opened 2022-06-20
Damage class name

majster2nn:
Added name for damage class so its not displayed in game while used by weapons as path to damage class.


Chicken-Bones:
Can we do this via the localisation file instead?

majster2nn:
nevermind didnt see that there

PR #2591, opened 2022-06-21
Example Whip Quick Fixes

Rijam:
### What is the bug?
Both Example Whip Projectiles were creating extra projectiles when swung continuously (most noticeable with Feral Claws/autoswing).
They were also missing the multihit penalty that all vanilla whips have, so I added that.

### How did you fix the bug?
By removing the two lines that set the `owner.itemAnimation` and `owner.itemTime`.

I don't think so. I don't see those two lines present in the vanilla source code, so I don't believe they are needed.


Chicken-Bones:
Fixed in 9ae5f96

Chicken-Bones:
ExampleWhip was fine, only ExampleWhipAdvanced was broken, you need to hold the animation during the charge mechanic.

PR #2594, opened 2022-06-21
Run SetupContent on all content from all mods before PostSetupContent

Chicken-Bones:
### What is the new feature?
Change `PostSetupContent` to run after `SetupContent` has been called for all `ModType` from all mods, not just this mod.

To make it clearer what's expected at this phase, and to make `PostSetupContent` even more useful, `ContentSamples.Initialize` has been moved to between `SetupContent` and `PostSetupContent`

`ModSystem.SetupContent` has also been unsealed, allowing you to fully utilise this load phase in your `ModSystem`s

With this, the 3 key phases are:
`Mod.AddContent` / `ModType.SetupContent` / `ModSystem.PostSetupContent + Mod.TransferAllAssets`
Followed by recipes (`AddRecipes` / `PostAddRecipes` / `PostSetupRecipes`)

New localisations for the 3 phases are:
`Adding Content` / `Configuring Content` / `Finalizing Content`

It's what people expect...
image

Makes `PostSetupContent` more useful for cross mod content. As an example, it should be safe to create item instances and lookup tables. A lot of mods end up using `AddRecipes` for stuff that should/could really go in `PostSetupContent`.

This of course, means that if mod B depends on mod A, and B had some code in a `SetStaticDefaults` method which depended on code from mod A's `PostSetupContent` it would break. But the code could just be moved to `PostSetupContent` in `B`.

Consider re-ordering more or less of the vanilla content setup

Should there be a separate message for the 2 phases? Currently every mod looks like it goes through 'Loading: ModName' twice.

### ExampleMod updates
None required

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
@@ -375,7 +375,7 @@
 		"Packaging": "Packaging: {0}",
 		// There should be ModName after ':'
 		"MSFinding": "Finding Mods...",
-		"MSInstantiating": "Instantiating Mods...",
+		"MSInstantiating": "Constructing Mods...",
 		"MSSandboxing": "Sandboxing: {0}",
 		"MSLoading": "Adding Content: {0}",
 		"MSResizing": "Resizing...",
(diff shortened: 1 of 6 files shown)


Chicken-Bones:
This will be a July preview feature unless it gets rejected

Chicken-Bones:
> Why is it now "constructing" instead of "instantiating"? What's the difference between the two?

No difference, it's just a slightly more user friendly localisation.

PR #2597, opened 2022-06-22
Port ExampleRobe to 1.4

jaydevelopsstuff:
Example Robe has been ported from the 1.3 ExampleMod.

Reference issue: #1403.

No


direwolf420:
I would add a comment above `EquipLoader.AddEquipTexture` that says that by passing `this` (which is the `ModItem` instance) the equip texture can be referenced using the `Name` property later as long as the `EquipType` has not already been used to register it already (for example via `AutoloadEquip`). And optionally the same comment but reversed above the `EquipLoader.GetEquipSlot` line.
You should probably also link to the "advanced" example somewhere: https://github.com/tModLoader/tModLoader/blob/1.4/ExampleMod/Content/Items/Armor/ExampleCostume.cs

Also concider this approach (caching the equip texture slot to not use `GetEquipSlot`):
```cs
private int legsSlot = -1;
legsSlot = AddEquipTexture(...);
equipSlot = legsSlot;
```

jaydevelopsstuff:
@direwolf420 I'll definitely add those comments, and in terms of caching the equip texture slot, is there any real benefit in this scenario since I'm only really accessing the slot once? Even if I don't end up using the caching method I'll definitely mention it in a comment.

I'll get to adding stuff tomorrow, thanks for the suggestions.

jaydevelopsstuff:
Yeah looking at it more, I definitely should add more documentation.

ExampleMod/Content/Items/Armor/Vanity/ExampleRobe.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Terraria;
+using Terraria.ModLoader;
+using Terraria.ID;
+
+namespace ExampleMod.Content.Items.Armor.Vanity
+{
+	// See also: ExampleCostume
+	[AutoloadEquip(EquipType.Body)]
+	internal class ExampleRobe : ModItem
```
jaydevelopsstuff: In the 1.3 ExampleRobe the class was `internal`, I'm not sure why this is necessary, but I included it in the port just in case.

direwolf420 (approved):

PR #2602, opened 2022-06-23
Recipe conditions player parameter

ExterminatorX99:
### What is the new feature?
Add `Player` parameter to the `Recipe.Condition` availability predicate

Allows mods to provide dummy players to be used instead of the local player

Keep current

```s
Player player = (Player) Main.LocalPlayer.Clone();
player.adjTile = myAdjTiles;

bool available = RecipeLoader.RecipeAvailable(Main.Recipe[123], player);
```

### ExampleMod updates
Use the player parameter instead of `Main.LocalPlayer`

### DIFF
--- a/ExampleMod/Content/ExampleRecipes.cs
+++ b/ExampleMod/Content/ExampleRecipes.cs
@@ -86,7 +86,7 @@ namespace ExampleMod.Content
 				// Adds a custom condition, that the player must be at <1/2 health for the recipe to work.
 				// The first argument is a NetworkText instance, i.e. localized text. The key used here is defined in 'Localization/*.hjson' files.
 				// The second argument uses a lambda expression to create a delegate, you can learn more about both in Google.
-				.AddCondition(NetworkText.FromKey("RecipeConditions.LowHealth"), r => Main.LocalPlayer.statLife < Main.LocalPlayer.statLifeMax / 2)
+				.AddCondition(NetworkText.FromKey("RecipeConditions.LowHealth"), (r, p) => p.statLife < p.statLifeMax / 2)
 
 				// When you're done, call this to register the recipe. Note that there's a semicolon at the end of the chain.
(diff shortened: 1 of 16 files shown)


Chicken-Bones:
Hm, I'm not sure about this. The usefulness seems limited. The fact that conditions for most things (biomes and adjacency) happen to be stored on the player is a bit of a coincidence. Vanilla has `Condition.TimeDay` and `Condition.TimeNight` for example.

I understand the sample usage, but not the motivation. Can you give me more details about why you would want to supply a dummy player? Surely you'd have to already know what the condition does, to know what variables in the dummy to set.

ExterminatorX99:
My main use case is for the Magic Storage crafting interface.
We want to have the recipe querying on a separate thread because it takes a while, even longer with Recursive Craft enabled.
The main reason we can't do it is because to satisfy the conditions we modify the local player data, but if the querying takes longer than a single frame the data gets reset to the normal values that don't reflect the conditions in the crafting interface.

ExterminatorX99:
> The fact that conditions for most things (biomes and adjacency) happen to be stored on the player is a bit of a coincidence. Vanilla has Condition.TimeDay and Condition.TimeNight for example.

As far as I can tell, conditions either rely on player data, or world/system data

Chicken-Bones:
I think checking recipe availability on another thread is too risky. Even if we were to supply a `Player` parameter to conditions, there's no guarantee people would use it. Similarly, there's no guarantee that the variables people might reference on their mod-systems are thread-safe either.

I also just noticed that there's `GlobalRecipe` which should be removed imo.
If `GlobalRecipe.RecipeAvailable` is removed, then you could check all recipes which don't have any `Conditions` on another thread, and just run the conditional recipes on the main thread.

You can also spread the effect of calculating recipes across multiple frames, with a stopwatch or recipe/frame limit.

PR #2606, opened 2022-06-24
[1.4] TileLoader.HasSolidTopCollision()

absoluteAquarian:
### What is the new feature?
Added a new hook to `TileLoader`:  `HasSolidTopCollision()`

Currently, "solid top" tiles can only be walked on for the uppermost subtiles in the spritesheet.  
This is a problem for mods like Magic Storage, which uses one spritesheet for all Storage Units (except the Creative Storage Unit), which results in the player being able to walk on only the basic Storage Unit and none of the others.

A possible alternative could be retrieving the TileObjectData for the tile and relying on CoordinateHeights.

n/a

### ExampleMod updates
Updated `Worm.cs` due to it previously reading from `Main.tileSolidTop[]`


absoluteAquarian:
Closed due to consensus not wanting a hook for this fix.

PR #2607, opened 2022-06-24
Disable quickedit in console

Chicken-Bones:
### What is the bug?
https://www.statmodel.com/download/QuickEdit.pdf

When you interact with the console in Windows 10, the console enters a "Select" mode, and the process blocks at the next attempt to write to stdout.

This is a problem for client users, who might accidentally click on the console or interact with it during setup. `Program.LaunchGame_` even has a call to `Platform.Get<IWindowService>().SetQuickEditEnabled(enabled: false);` because vanilla has worked around this for servers.

### How did you fix the bug?
Query and log the windows version. Then, if it's 10+, run a quick .NET 4.0 application (which should be present on all such systems) to call the kernerl32 methods required to disable quickedit.

I don't need to bounce all the windows scripts through yet another `busybox-sh.bat`, but it seemed like the nicest place to get the windows version in one place. Unfortunately `ver` is a cmd utility, and not a process that bash can access from inside busybox.


Solxanich (approved):
Approved, Only concern point is the for loop in busybox-sh.bat might break on Win7.
Merging to 1.4 but holding off on preview.

Issue #1612, opened 2021-07-12
Asset Loading Post-Merge TODOs


Mirsario:
Moved from #1567

> Note that async does not yet mean on demand and that the only way for an asset to be reloaded currently is switching resource packs. (Reloading mods discards all the old assets and makes a whole new AssetRepository)

**TODO:**
- [x] Change `AssetRepository.Request` to `internal` with `[InternalsVisibleTo(Terraria)]` and expose a request method for mods with the loading mode defaulting to `AsyncLoad` for consistency with `ModContent.Request`
- [x] FNA versioning in deps.json (csproj comment)
- [x] Remove obsolete methods (GetTexture/Sound, HasTexture/Sound from Mod/ModContent)
- [x] Figure out what `ResourcePackContentValidator` was ever meant to do.
  ```
  Yoraiz0r:
  It was meant to validate whole resource packs in one go, and let you know of rejections immediately
  - as right now the resource pack system only rejects on-demand and you don't know whether you have an issue until you are actually experiencing it.
  That said, this feature never made it to fruition, and most of what you see in the compiled code is just a dummy state.. that is unused, too.
  So you can feel free to remove/ignore it.
  ```
- [ ] Check all ExampleMod calls to `ModContent.Request` to make sure the loading mode is appropriate
- [ ] Let mods add to the asset reader pool.
- [ ] Allow resource packs to affect mod assets.
  - [ ] Wrap the content source list with delegating sources that prepend the mod name to the path.
  - [ ] Mod content validators; Auto-build a texture size chart.
- [ ] Allow mods to add resource packs for overriding vanilla assets.
  - [ ] A default one, if a mod contains a `Terraria/` directory.
  - [ ] Additional custom ones via code, can be enabled and disabled.
- [ ] Make better use of async asset loading.
  - [ ] Use the content validator to produce placeholder textures with appropriate size.
- [ ] Log rejections (and give content sources a decent `ToString()`.)
- [ ] Test performance impact (on load time) of opening new file streams for every TmodFile asset request.
- [ ] Don't constantly request glow textures.
- [ ] Test if mod building works in headless environments (don't know why it wouldn't.)
- [ ] Make Asset\<Texture2D\>.Value return dummy transparent textures until asynchronous loading finishes.

RighteousRyan1:
Resource packs that add/replace sounds cause issues on texture pack load. Causes the UI to bug out and in the end- results in a softlock.

PR #2617, opened 2022-06-25
Fix language packs don't work issue

### LINKED ISSUE #2615: Language pack don't work
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Language pack don't work at all

### Log File

_No response_

### Steps to reproduce

1. Enable a resource pack for modifying text
2. Get into the game and check it. The text has not changed.

### Expected Behavior

The text should be changed

### Actual Behavior

The text doen't change

### Reproduction frequency

_No response_

### Additional Information

_No response_

Cyrillya:
### What is the bug?
Fixes #2615

### How did you fix the bug?
In `LanguageManager.UseSources`, I found that the `assetNameStart` used `ToLower()`
```
string name = ActiveCulture.Name;
string assetNameStart = ("Localization" + Path.DirectorySeparatorChar + name).ToLower();
LoadLanguage(ActiveCulture, processCopyCommands: false);
foreach (IContentSource item in sourcesFromLowestToHighest) {
	foreach (string item2 in item.GetAllAssetsStartingWith(assetNameStart)) {
		string extension = item.GetExtension(item2);
```
But it doesn't use `ToLower()` in the method it calls `IContentSource.GetAllAssetsStartingWith()`
```
IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart, bool lower = false) =>
    EnumerateAssets().Where(s => s.StartsWith(assetNameStart));
```
So the `assetNameStart` never finds any asset name starts with `Localization\[LanguageName]`
The incoming parameter will only be `localization\[LanguageName]`, which is not uppercase

I replaced the method and call with these
```
string name = ActiveCulture.Name;
string assetNameStart = ("Localization" + Path.DirectorySeparatorChar + name).ToLower();
LoadLanguage(ActiveCulture, processCopyCommands: false);
foreach (IContentSource item in sourcesFromLowestToHighest) {
	// Added a "true" parameter so that it can search by lowercase paths
	foreach (string item2 in item.GetAllAssetsStartingWith(assetNameStart, true)) {
		string extension = item.GetExtension(item2);
```
```
IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart, bool lower = false) {
	if (lower)
		return EnumerateAssets().Where(s => s.ToLower().StartsWith(assetNameStart));
	return EnumerateAssets().Where(s => s.StartsWith(assetNameStart));
}
```
Now it will find the correct file

However, it didn't end here, since I found that `item.GetExtension(item2)` can actually get nothing but an empty string
I looked into the vanilla code of `GetExtension()`, in `FileSystemContentSource.GetExtension()`
```
public string GetExtension(string assetName) {
	if (!_nameToAbsolutePath.TryGetValue(assetName, out string value))
		throw AssetLoadException.FromMissingAsset(assetName);

	return Path.GetExtension(value) ?? "";
}
```
I found it used `Path.GetExtension(value) ?? ""`, so I replaced  `item.GetExtension(item2)` with it:
```
string extension = Path.GetExtension(item2) ?? "";
```
And the language packs just started to work.
This could be a temporary solution.

### DIFF
--- a/patches/tModLoader/ReLogic/Content/Sources/FileSystemContentSource.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/Sources/FileSystemContentSource.cs.patch
@@ -1,6 +1,10 @@
 --- src/TerrariaNetCore/ReLogic/Content/Sources/FileSystemContentSource.cs
 +++ src/tModLoader/ReLogic/Content/Sources/FileSystemContentSource.cs
-@@ -4,13 +_,17 @@
+@@ -1,16 +_,21 @@
+ using System;
+ using System.Collections.Generic;
+ using System.IO;
++using System.Linq;
  
  namespace ReLogic.Content.Sources;
  
@@ -19,13 +23,18 @@
  
  	public int FileCount => _nameToAbsolutePath.Count;
  
-@@ -21,8 +_,12 @@
+@@ -21,8 +_,17 @@
  			_basePath += Path.DirectorySeparatorChar;
  
  		BuildNameToAbsolutePathDictionary();
 +
 +		// Added by TML.
 +		SetAssetNames(_nameToAbsolutePath.Keys);
++
++		// #2617 - Allow using paths with extensions in GetExtension calls, because we happen to return such paths in EnumerateAssets().
++		foreach (var pair in assetExtensions.ToArray()) {
++			assetExtensions.TryAdd(pair.Key + pair.Value, pair.Value);
++		}
  	}
  
 +	/*
--- a/patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
@@ -7,7 +7,7 @@
  
  namespace ReLogic.Content.Sources;
  
-@@ -7,17 +_,48 @@
+@@ -7,17 +_,53 @@
  {
  	IContentValidator ContentValidator { get; set; }
  
@@ -54,6 +54,11 @@
 +		!Rejections.IsRejected(assetName) && GetExtension(assetName) != null;
 +
 +	// Added by TML.
-+	IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart) =>
-+		EnumerateAssets().Where(s => s.StartsWith(assetNameStart));
++	IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart, bool lower = false)
++	{
++		if (lower)
++			return EnumerateAssets().Where(s => s.ToLower().StartsWith(assetNameStart));
++
++		return EnumerateAssets().Where(s => s.StartsWith(assetNameStart));
++	}
  }
--- a/patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
@@ -1,13 +1,15 @@
 --- src/TerrariaNetCore/ReLogic/Content/Sources/IContentSource.cs
 +++ src/tModLoader/ReLogic/Content/Sources/IContentSource.cs
-@@ -1,5 +_,6 @@
+@@ -1,5 +_,8 @@
++using System;
  using System.Collections.Generic;
++using System.Diagnostics.CodeAnalysis;
  using System.IO;
 +using System.Linq;
  
  namespace ReLogic.Content.Sources;
  
-@@ -7,17 +_,53 @@
+@@ -7,17 +_,57 @@
  {
  	IContentValidator ContentValidator { get; set; }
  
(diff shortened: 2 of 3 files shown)


JavidPack:
There are fundamental issues with how the system has been changed by tmod. This probably solves a few of them. I went through similar issues when working on this same issue on https://github.com/tModLoader/tModLoader/compare/1.4...1.4_FixTextResoucePacks branch. I think we'll need feedback from the people who moved things around, see what the intention is for these methods and such, since they seem to have been changed for various purposes.

Shiranui-Izayoi:
I know this pull request is a bit old. But where did you find the files? Are they on the steam download of TModLoader or do I need to do something else?

I have run into this as well and wanted to fix it. But I just can't find "patches" folder or "LanguageManager.UseSources".

JavidPack:
> I know this pull request is a bit old. But where did you find the files? Are they on the steam download of TModLoader or do I need to do something else?
> 
> I have run into this as well and wanted to fix it. But I just can't find "patches" folder or "LanguageManager.UseSources".

No, the tmodloader download does not contain the tmodloader source code. To access the source code, you must follow the steps in https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-contributors. It is an involved process. If you aren't familiar with the source code already, fixing this will be very difficult. Please upvote the issue with 👍 if you'd like this issue prioritized.

Mirsario:
Rebased after more than a year, and changed the solution a bit:
- `ToLower()` is replaced with `StringComparison` parameters for performance.
- Instead of using `Path.GetExtension` and thus breaking non-FS content sources, `FileSystemContentSource` now populates its extension lookup to allow paths with extensions in `GetExtension`, since that's what `EnumerateAssets` yields.

My apologies for this taking so long to be reviewed & merged.

[Hungarian replacing English](https://steamcommunity.com/sharedfiles/filedetails/?id=2441620742):
image

patches/tModLoader/Terraria/Localization/LanguageManager.cs.patch
```diff
@@ -113,8 +113,13 @@
  		LoadLanguage(ActiveCulture, processCopyCommands: false);
 +		*/
  		foreach (IContentSource item in sourcesFromLowestToHighest) {
- 			foreach (string item2 in item.GetAllAssetsStartingWith(assetNameStart)) {
- 				string extension = item.GetExtension(item2);
+-			foreach (string item2 in item.GetAllAssetsStartingWith(assetNameStart)) {
+-				string extension = item.GetExtension(item2);
++			foreach (string item2 in item.GetAllAssetsStartingWith(assetNameStart, ignoreCase: true)) {
++				string extension = item.GetExtension(item2) ?? "";
```
Chicken-Bones, line 119: no need for the null coalesce patch here

patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
```diff
@@ -54,6 +56,15 @@
 +		!Rejections.IsRejected(assetName) && GetExtension(assetName) != null;
 +
 +	// Added by TML.
-+	IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart) =>
-+		EnumerateAssets().Where(s => s.StartsWith(assetNameStart));
++	[Obsolete, SuppressMessage("CodeQuality", "IDE0051:Remove unused private members")]
++	private IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart)
++		=> GetAllAssetsStartingWith(assetNameStart, ignoreCase: false);
```
Chicken-Bones, line 61: Why make this private member? Are mods reflecting it?

patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
```diff
@@ -54,6 +56,15 @@
...
++
++	// Added by TML.
++	IEnumerable<string> GetAllAssetsStartingWith(string assetNameStart, bool ignoreCase = false)
++	{
++		var comparison = ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
++
++		return EnumerateAssets().Where(s => s.StartsWith(assetNameStart, comparison));
```
Chicken-Bones, line 68: Is this really the correct solution? The entire asset system in vanilla is case insensitive. Should we just be using `.ToLowerInvariant()` everywhere instead?

patches/tModLoader/ReLogic/Content/Sources/FileSystemContentSource.cs.patch
```diff
@@ -19,13 +23,18 @@
...
  
  		BuildNameToAbsolutePathDictionary();
 +
 +		// Added by TML.
 +		SetAssetNames(_nameToAbsolutePath.Keys);
++
++		// #2617 - Allow using paths with extensions in GetExtension calls, because we happen to return such paths in EnumerateAssets().
++		foreach (var pair in assetExtensions.ToArray()) {
++			assetExtensions.TryAdd(pair.Key + pair.Value, pair.Value);
++		}
```
Chicken-Bones, line 37: Shouldn't this be in `SetAssetNames`? There are other content sources than just `FileSystemContentSource`

patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
Mirsario: To avoid runtime breakage from introducing a new parameter. I don't recall the protection modifier mattering in that case.

patches/tModLoader/ReLogic/Content/Sources/FileSystemContentSource.cs.patch
Mirsario: FSCS is the only content source that returns paths with extensions in EnumerateAssets. I went for the most minimal solution with the least chance of breaking anything further, as this fix is intended to be cherrypicked to stable. Future asset system reworks from #1612 can make more permanent improvements.

patches/tModLoader/ReLogic/Content/Sources/IContentSource.cs.patch
steviegt6: Without IgnoresAccessChecksTo, references to private members of another assembly (even if linked to public members at compile time) will be thrown; one would need to annotate their mod assembly with `[assembly: IgnoresAccessChecksTo("tModLoader")]` (or module) to avoid this. Late, but relevant.

PR #2618, opened 2022-06-25
Example Jousting Lance

Rijam:
### What are the new features?
(Credit to Chicken Bones for these 3 changes:)

`ProjectileID.Sets.NoMeleeSpeedVelocityScaling[]`
This set stops all melee speed increases and decreases from affecting the shoot velocity of the melee projectile. Even setting DamageClass.MeleeNoSpeed still affects the shoot velocity. This set is now used by `MonkStaffT2` (Ghastly Glaive), `MonkStaffT3` (Sky Dragon's Fury, but not the alt fire), Jousting Lance, Hallowed Jousting Lance, and Shadow Jousting Lance.

`Item.InterruptChannelOnHurt`
When the player gets hurt and is using a weapon with channel the weapon will be "put away".

`Item.StopAnimationOnHurt`
Similar to `Item.InterruptChannelOnHurt`, but the weapon will be "put away" instantly instead of letting the animation play out.

A custom held projectile melee weapon who bases their rotation and/or position on the shoot velocity will have the wrong position if the player has different than normal melee attack speed. The new projectile set is an easy way to for the modder to ensure that the shoot velocity is always the same.

If a modder wants to make a channel weapon "holster" when the player gets hit, the new fields are an easy way for them to do that.

For `ProjectileID.Sets.NoMeleeSpeedVelocityScaling[]`:
Manually checking the player's attack speed with `GetTotalAttackSpeed` and using `ModifiyShootStats()` to correct it. This would need to be done for every weapon who would want to do this.

For `Item.InterruptChannelOnHurt` and `Item.StopAnimationOnHurt`:
A player class could be created with a `Hurt()` override. Check that the player is holding the correct item, then check every projectile in the world if it is active, the owner is the player, and the projectile type is the one that the item shoots. Then set `Player.channel = false` (and additionally set `Player.itemAnimation` and `Player.itemAnimationMax` to 0).

As you can see, the new set and fields are much more convenient to use.

Example Jousting Lance (see below).

### Example Mod updates

Added an Example Jousting Lance. This includes an item and projectile. The Example Jousting Lance acts exactly like the vanilla Jousting Lances.

Here is a little image I made demonstrating the rotation of the Jousting Lance. 0 is the top left and 180° or π radians is the bottom right. It for some reason goes from 0 to 315° or 7π/4 radians and then goes to -45° or -π/4 radians to 0. I don't know if that's how all projectiles work, but it confused me enough to make this image lol.

JoustingLanceRotation

I say in the code comments that `ProjectileID.Sets.DismountsPlayersOnHit[]` will only dismount the player if they are hit with another jousting lance in PVP, but that is not necessarily true. A mod could add that set to their projectile and it'll dismount the player. The reason I said it the way I said it in the code comments is that is how the vanilla behavior is (only jousting lances are a part of the set).


ilija04:
About the `lanceHitboxBounds` part, you can just patch the vanilla code by changing the vanilla's `_lanceHitboxBounds` from `private` to `public`.

Rijam:
Changing the vanilla's `_lanceHitboxBounds` from `private` to `public` is not necessary because some people may want to change those values if they make their Jousting Lance bigger or smaller than the vanilla ones. But, your comment did remind me that the `lanceHitboxBounds` rectangle didn't need to be its own private variable (it is now just in `Colliding()`).

Rijam:
If/when #2773 merges, that will solve the issue with the prefixes here.

Rijam:
I have updated the code with `MeleePrefix()` and I think the PR is ready for a review.

For anyone looking to implement a jousting lance in Stable (2022.7) and wants to have the "correct" prefixes, you can use this hacky solution: https://github.com/Rijam/tModLoader/blob/5e9553dff3a91908481e20d92efb6e22c722b2d3/ExampleMod/Content/Items/Weapons/ExampleJoustingLance.cs#L64

ExampleMod/Content/Items/Weapons/ExampleJoustingLance.cs
```diff
@@ -0,0 +1,96 @@
...
+
+			double damageTaken = Main.CalculateDamagePlayersTake((int)damage, Player.statDefense);
+
+			if (!Player.immune && damageTaken >= 1.0 && Player.inventory[Player.selectedItem].type == itemType) {
+				for (int j = 0; j < Main.maxProjectiles; j++) {
+					if (Main.projectile[j].active && Main.projectile[j].owner == Player.whoAmI) {
+						Main.projectile[j].active = false;
+					}
+				}
+			}
```
Chicken-Bones: This will kill any other projectiles, including summons. You should be able to just set `Player.channel = false`. If you want the animation to stop immediately (and not play the remaining frames) then you can set `itemAnimation = 0` as well.

ExampleMod/Content/Items/Weapons/ExampleJoustingLance.cs
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// This will cause the Jousting Lance to become inactive if the player is hit with it out. Make sure to change the itemType to your item.
+	public class ExampleJoustingLancePlayer : ModPlayer
+	{
+		public override void Hurt(bool pvp, bool quiet, double damage, int hitDirection, bool crit, int cooldownCounter) {
+
+			int itemType = ModContent.ItemType<ExampleJoustingLance>(); // Change this to your item
+
+			double damageTaken = Main.CalculateDamagePlayersTake((int)damage, Player.statDefense);
```
Chicken-Bones: You don't need to recalculate damage taken here, or check immune. `Hurt` isn't called if damage taken is < 1 or the player is immune.

ExampleMod/Content/Items/Weapons/ExampleJoustingLance.cs
Chicken-Bones: You could, of course, keep this and do it vanilla style, but you need to check the projectile id.

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			// AIType = ProjectileID.JoustingLance; 
+		}
+
+		// This is the behavior of the Jousting Lances.
+		public override void AI() {
+			Player owner = Main.player[Projectile.owner]; // Get the owner of the projectile.
+			Vector2 center = owner.RotatedRelativePoint(owner.MountedCenter); // Get the center of the owner.
+			Projectile.direction = owner.direction; // Set the direction of the projectile to the same direction as the owner.
+			owner.heldProj = Projectile.whoAmI; // Set the owner's held projectile to this projectile.
+			owner.itemTime = owner.itemAnimation; // Set the time left on this item to the animation time.
```
Chicken-Bones: no need for this line. Potentially confusing. This is done by `SetDummyItemTime`

ExampleMod/Content/Items/Weapons/ExampleJoustingLance.cs
Chicken-Bones: Or, use `heldProj`

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			// Act like the normal Jousting Lance. Use this if you set the aiStyle to 19.
+			// AIType = ProjectileID.JoustingLance; 
+		}
+
+		// This is the behavior of the Jousting Lances.
+		public override void AI() {
+			Player owner = Main.player[Projectile.owner]; // Get the owner of the projectile.
+			Vector2 center = owner.RotatedRelativePoint(owner.MountedCenter); // Get the center of the owner.
+			Projectile.direction = owner.direction; // Set the direction of the projectile to the same direction as the owner.
+			owner.heldProj = Projectile.whoAmI; // Set the owner's held projectile to this projectile.
```
Chicken-Bones: Don't write comments that say what the code says. Write comments that say what the code _doesn't_ say. 

Comments should give context, and explain why. I can see that 'the `owner`'s `heldProj` is set to the `Projectile` `whoAmI` by reading the code.

Instead say:
Set `heldProj` so the projectile will be destroyed when the player drops/swaps the held item, or is interrupted.
(Say why the line exists)

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+				Projectile.Kill();
+			}
+
+			// Set the rotation of the projectile.
+			// For reference, 0 is the top left, 180 degrees or pi radians is the bottom right.
+			Projectile.rotation = (float)Math.Atan2(Projectile.velocity.Y, Projectile.velocity.X) + (float)Math.PI / 2f + (float)Math.PI / 4f;
+
+			// The Hallowed and Shadow Jousting Lance spawn dusts when the player is moving above a certain speed.
+			float minimumPlayerVelocity = 6f;
+			float minimumSpeed = 0.8f;
```
Chicken-Bones: no need for a `minimumSpeed` variable. Inline it into the condition below.

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			if (owner.ItemAnimationEndingOrEnded) {
+				Projectile.Kill();
+			}
+
+			// Set the rotation of the projectile.
+			// For reference, 0 is the top left, 180 degrees or pi radians is the bottom right.
+			Projectile.rotation = (float)Math.Atan2(Projectile.velocity.Y, Projectile.velocity.X) + (float)Math.PI / 2f + (float)Math.PI / 4f;
+
+			// The Hallowed and Shadow Jousting Lance spawn dusts when the player is moving above a certain speed.
+			float minimumPlayerVelocity = 6f;
```
Chicken-Bones: A better name would be `minimumDustVelocity`

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			float minimumSpeed = 0.8f;
+
+			// This Vector2.Dot is the dot product between the projectile's velocity and the player's velocity normalized to be between -1 and 1.
+			// What this means in this context is that the speed value will be closer to positive 1 if the player is moving in the same direction as the direction the lance was shot.
+			// Example: if the lance is shot up and to the right, the value here will be closer to 1 if the player is also moving up and to the right.
+			float speed = Vector2.Dot(Projectile.velocity.SafeNormalize(Vector2.UnitX * owner.direction), owner.velocity.SafeNormalize(Vector2.UnitX * owner.direction));
```
Chicken-Bones: A better name would be `movementInLanceDirection` or similar.

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			// Fade the projectile in when it first spawns
+			Projectile.alpha -= 40;
+			if (Projectile.alpha < 0) {
+				Projectile.alpha = 0;
+			}
+
+			float pushOutMulti = itemAnimationMax - itemAnimationMaxDiv - minimumItemAnimation; // This value will get larger the further out the projectile is from the player until it reaches it's maximum distance.
+			float pullInMulti = itemAnimationMaxDiv - smallerItemAnimation; // This value will get larger when the projectile is being pulled back in.
+			float scaleFactor = spawnDistance + flyOut * pushOutMulti - flyBack * pullInMulti;
+			Projectile.position += Projectile.velocity * scaleFactor; // Move the projectile based on the velocity and scaleFactor
```
Chicken-Bones: The names and math here is quite confusing. Especially because `Item.shootSpeed` throws a `3.5f` multiplier in. Also you have a mixture of alpha code, channel time code and tip positioning code all together.

I suggest changing `Item.shootSpeed` to `1f` and using code like below:
```cs
			int itemAnimationMax = owner.itemAnimationMax;
			// Remember, frames count down from itemAnimationMax to 0
			// Frame at which the lance is fully extended. Hold at this frame before retracting.
			// Scale factor (0.34f) means the last 34% of the animation will be used for retracting.
			int holdOutFrame = (int)(itemAnimationMax * 0.34f);
			if (owner.channel && owner.itemAnimation < holdOutFrame) {
				owner.SetDummyItemTime(holdOutFrame); // This makes it so the projectile never dies while we are holding it (except when we take damage, see ExampleJoustingLancePlayer).
			}

			// If the Jousting Lance is no longer being used, kill the projectile.
			if (owner.ItemAnimationEndingOrEnded) {
				Projectile.Kill();
				return;
			}

			int itemAnimation = owner.itemAnimation;
			// extension and retraction factors (0-1). As the animation plays out, extension goes from 0-1 and stays at 1 while holding, then retraction goes from 0-1.
			float extension = 1 - Math.Max(itemAnimation - holdOutFrame, 0) / (float)(itemAnimationMax - holdOutFrame);
			float retraction = 1 - Math.Min(itemAnimation, holdOutFrame) / (float)holdOutFrame;

			// distances are in pixels
			float extendDist = 22; // How far to fly out during extension
			float retractDist = extendDist / 2; // How far to fly back during retraction
			float tipDist = 98 + extension * extendDist - retraction * retractDist;

			Vector2 center = owner.RotatedRelativePoint(owner.MountedCenter); // Get the center of the owner.
			Projectile.Center = center; // Set the center of the projectile to the center of the owner. Projectile.Center is now actually the tip of the Jousting Lance.
			Projectile.position += Projectile.velocity * tipDist; // The projectile velocity contains the orientation of the lance, multiply it by the tipDist to position the tip

			// Set the rotation of the projectile.
			// For reference, 0 is the top left, 180 degrees or pi radians is the bottom right.
			Projectile.rotation = (float)Math.Atan2(Projectile.velocity.Y, Projectile.velocity.X) + (float)Math.PI * 3 / 4f;

			// Fade the projectile in when it first spawns
			Projectile.alpha -= 40;
			if (Projectile.alpha < 0) {
				Projectile.alpha = 0;
			}
```

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+			}
+
+			// If the Jousting Lance is no longer being used, kill the projectile.
+			if (owner.ItemAnimationEndingOrEnded) {
+				Projectile.Kill();
+			}
+
+			// Set the rotation of the projectile.
+			// For reference, 0 is the top left, 180 degrees or pi radians is the bottom right.
+			Projectile.rotation = (float)Math.Atan2(Projectile.velocity.Y, Projectile.velocity.X) + (float)Math.PI / 2f + (float)Math.PI / 4f;
```
Chicken-Bones: use `* 3/4f`

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
```diff
@@ -0,0 +1,210 @@
...
+		// This is the custom collision that Jousting Lances uses. 
+		public override bool? Colliding(Rectangle projHitbox, Rectangle targetHitbox) {
+			float rotationFactor = Projectile.rotation + (float)Math.PI / 4f; // The rotation of the Jousting Lance.
+			float scaleFactor = 95f; // How far back the hit-line will be from the tip of the Jousting Lance. You will need to modify this if you have a longer or shorter Jousting Lance. Vanilla uses 95f
+			float widthMultiplier = 23f; // How thick the hit-line is. Increase or decrease this value if your Jousting Lance is thicker or thinner. Vanilla uses 23f
+			float collisionPoint = 0f; // collisionPoint is needed for CheckAABBvLineCollision(), but it isn't used for our collision here. Keep it at 0f.
+
+			// This Rectangle is the width and height of the Jousting Lance's hitbox which is used for the first step of collision.
+			// You will need to modify the last two numbers if you have a bigger or smaller Jousting Lance. Vanilla uses (0, 0, 300, 300).
+			Rectangle lanceHitboxBounds = new Rectangle(0, 0, 300, 300);
```
Chicken-Bones, line 159: This all looks great. My only suggestion is to add
```cs
			Vector2 hitLineEnd = Projectile.Center + rotationFactor.ToRotationVector2() * scaleFactor;

			// Debug dusts, to easily see where the hit line starts and end
			// Dust.NewDustPerfect(Projectile.Center, DustID.Pixie, Velocity: Vector2.Zero);
			// Dust.NewDustPerfect(hitLineEnd, DustID.Pixie, Velocity: Vector2.Zero);
```

Chicken-Bones (changes requested):
Thanks for your contribution! Sorry it's taken me so long to review it. This looks high quality, with great attention to detail.

ExampleMod/Content/Projectiles/ExampleJoustingLanceProjectile.cs
Rijam: Yeah you are right. I just copied and pasted the decompiled source code and explained what it was doing. It didn't even cross my mind to refactor the code to make it more understandable.

Rijam:
For anyone looking to implement a Jousting Lance on Stable (2022.8) which doesn't have these new features, you can achieve them through this way:

Instead of adding `ProjectileID.Sets.NoMeleeSpeedVelocityScaling[]` to your projectile, add this to your item:
```cs
public override void ModifyShootStats(Player player, ref Vector2 position, ref Vector2 velocity, ref int type, ref int damage, ref float knockback)
{
	// If the player has increased melee speed, it will effect the shootSpeed of the Jousting Lance which will cause the projectile to spawn further away than it is supposed to.
	// This ensures that the velocity of the projectile is always the shootSpeed.
	float inverseMeleeSpeed = 1f / player.GetTotalAttackSpeed(DamageClass.Melee);
	velocity *= inverseMeleeSpeed;
}
```

And instead of adding `Item.StopAnimationOnHurt`, create a player class instead:
```cs
// This will cause the Jousting Lance to become inactive if the player is hit with it out. Make sure to change the itemType to your item.
public class ExampleJoustingLancePlayer : ModPlayer
{
	public override void Hurt(bool pvp, bool quiet, double damage, int hitDirection, bool crit, int cooldownCounter)
	{
		int itemType = ModContent.ItemType<ExampleJoustingLance>(); // Change this to your item

		if (Player.inventory[Player.selectedItem].type == itemType)
		{
			Player.channel = false;
			Player.itemAnimation = 0;
			Player.itemAnimationMax = 0;
		}
	}
}
```

I recommend removing these two things once you move to 2022.9 and use the new features instead.

PR #2620, opened 2022-06-25
Add PlayerLoader.HoverSlot hook

Cyrillya:
### What is the new feature?
Add `PlayerLoader.HoverSlot` method, similar to `PlayerLoader.ShiftClickSlot`

If modders want to modify the shift-click slot event (e.g. to put an item into their own ItemSlot UI), it's better if they can change `Main.cursorOverride` to tell the player what will happen

When a custom UI enabled, if the player hovers the cursor over an item in the inventory holding shift, the cursor icon will change to the icon for placing the item into a chest.

### ExampleMod updates
`ExampleShiftClickSlotPlayer` as a simple showcase of `ModPlayer.HoverSlot` and `ModPlayer.ShiftClickSlot`
If the player shift-clicks a gel, it changes its color and rarity.
dotnet_Xyc1PxvKhr


ExampleMod/Common/Players/ExampleShiftClickSlotPlayer.cs
```diff
@@ -0,0 +1,39 @@
...
+			if (context == ItemSlot.Context.InventoryItem && inventory[slot].type == ItemID.Gel) {
+				inventory[slot].color = Main.DiscoColor; // Change the color of the item into a "random" color
+				inventory[slot].rare = Main.rand.Next(ItemRarityID.Count); // Random rarity
+				SoundEngine.PlaySound(SoundID.Item4); // Play mana crystal using sound
+
+				// Block vanilla code so the item will not be picked up when it is clicked.
+				return true;
+			}
+			return base.ShiftClickSlot(inventory, context, slot);
+		}
```
Solxanich, line 25: Please add something to the tooltip to indicate this behaviour will happen, so that it isn't misinterpreted as a bug while examplemod is enabled.

Solxanich (changes requested):
Generally looks good. 
The Example needs a little attention to prevent confusion & forewarn someone who doesn't know this Example is present.

ExampleMod/Common/Players/ExampleShiftClickSlotPlayer.cs
Cyrillya: Done! Added GelGlobalItem.cs

ExampleMod/Common/GlobalItems/GelGlobalItem.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+using Terraria;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.GlobalItems
+{
+	// This is another part of the ExampleShiftClickSlotPlayer.cs that adds a tooltip line to the gel
+	public class GelGlobalItem : GlobalItem
+	{
+		public override void ModifyTooltips(Item item, List<TooltipLine> tooltips) {
```
Chicken-Bones, line 13: You're missing an override of `AppliesToEntity` so this just added your tooltip to every item in the game

ExampleMod/Common/Players/ExampleShiftClickSlotPlayer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		// Here we override the cursor style
+		public override bool HoverSlot(Item[] inventory, int context, int slot) {
+			// Apply our changes if this item is in inventory and is gel
+			if (context == ItemSlot.Context.InventoryItem && inventory[slot].type == ItemID.Gel) {
+				// If player is holding shift, use FavoriteStar texture, otherwise use CameraLight texture.
+				Main.cursorOverride = ItemSlot.ShiftInUse ? CursorOverrideID.FavoriteStar : CursorOverrideID.CameraLight;
```
Chicken-Bones, line 32: Why do you change this and return true, when the player isn't holding shift? Shouldn't you let it run through to the base call?

ExampleMod/Common/GlobalItems/GelGlobalItem.cs
Cyrillya: Forgot to do so. Now it's fixed

Cyrillya:
So when will this be merged?

Chicken-Bones (changes requested):

ExampleMod/Common/Players/ExampleShiftClickSlotPlayer.cs
Cyrillya: True. It's a mistake. Corrected now.

Chicken-Bones:
Example adjusted slightly ca6743c559cfa7675e9f0e01d5e3c98a570f0606

PR #2623, opened 2022-06-25
Add ModPlayer.ModifyFishingAttempt hook

direwolf420:
### What is the new feature?
`ModPlayer.ModifyFishingAttempt` hook for decent control over what catches will appear.

The current hook (`CatchFish`) allows you to add/override catches (which is perfectly suitable for most cases). This hook gives the modder control over the conditions leading up to deciding a catch (**Important: NOT if and when the catch will happen in general!**). This is important for further supporting mod compatibility, as long as other mods also utilize the fields provided in the hooks to pick what is cought (instead of just relying on randomness or arbitrary conditions).

It was quite difficult to figure out where to place the hook; Putting it before any other returns gives the modder full control over all parameters, such as the amount of liquid tiles required to fish (allowing fishing with just 1 liquid tile), and other environmental information. However, doing so will let vanilla overwrite assignments to the "rarities" and `crate`, so you wouldn't be able to modify those (unless I do multiple vanilla patches by changing the method sig of `FishingCheck_RollDropLevels` from `out`s to `ref`s.). Due to several returns being present, putting the hook somewhere before the last return would cause some data to not be initialized aswell. Placing it just before vanilla decides catches is the best compromise I have found.
Additionally, since `FishingAttempt` is passed as `ref`, and it also contains the cought item/NPC, I decided (for compatibility reasons) to zero those fields out after the hooks are called, as setting those is the purpose of `CatchFish` instead (that hook does NOT pass `FishingAttempt` as `ref`).

* "Enhanced Crate Potion" that quadruples crate chance
* "Fishing License" accessory that increases the chance to get higher quality fish
* Catching lava/honey fish in water

### ExampleMod updates
Added `ExampleCratePotion` and `ExampleCrateBuff` to showcase how to manipulate `FishingAttempt` fields to influence cought items.
Minor addition: Give `ExampleBuffPotion` a `DrinkParticleColors` entry (wasn't showcased yet, required for parity with vanilla potions and the new potion added here)


Chicken-Bones:
Ideally, we'd have a drop table system for fishing, but that's a later problem

PR #2633, opened 2022-06-26
Mod API proxying

Shockah:
Before anything else, I'll just say, that:
***~~I did not test this functionality with any tModLoader mods. I only checked if it builds.~~ I'm not exactly willing to come back for real to the Terraria modding scene, so I left it at that. The library that is leveraged here is already in use in a modding API for another game, Stardew Valley, and the related (super minimal) code was simply ported to tML.***

### What is the new feature?
This PR implements "mod API proxies". It leverages a library I created (called Pintail) to create proxies for arbitrary objects.

[Pintail's repository](https://github.com/Nanoray-pl/Pintail)

The library allows providing strongly-typed and/or duck-typed API (depending on the usage), without losing the benefit of not requiring the providing mod. The providing mod does not have to be referenced at all, not even as a weak dependency.

You provide an object you want to proxy, and a proxy interface type (it specifically has to be an interface). Pintail will generate a new type, that implements that interface, by delegating all of the method calls to the proxied object. The methods' bodies are built by emitting IL instructions.

The library is currently being used in the stable version of SMAPI, the modding API for Stardew Valley.
[SMAPI's website](https://smapi.io/)
[SMAPI's repository](https://github.com/Pathoschild/SMAPI)

Implementation in SMAPI:
[ModRegistryHelper.cs](https://github.com/Pathoschild/SMAPI/blob/a8a4d314dff1f80970090e9d364f36e787df73d6/src/SMAPI/Framework/ModHelpers/ModRegistryHelper.cs#L59-L101)
[InterfaceProxyFactory.cs](https://github.com/Pathoschild/SMAPI/blob/a8a4d314dff1f80970090e9d364f36e787df73d6/src/SMAPI/Framework/Reflection/InterfaceProxyFactory.cs)

Before Pintail was a thing, SMAPI used the same approach, but in a super simplified form (the APIs could only use types known by both parties) - I'm not sure for how long, but I've seen mods that are at least 2 years old, providing an API. This is the generally agreed upon standard in that community - it's either that or strong DLL references.

Nowadays, Pintail has been in use in stable builds for about 2 months I think, with barely any problems (one of those being a super edge case which only 1 mod ran into - fixed by now; the second of those was a minor problem which could be worked around on the mod side - also fixed already).

Actual usage of this feature looks like this:
* A providing mod returns some kind of object in its `Mod.GetAPI` method. It doesn't really matter what that object is - it can even be `this` (this happens a lot in the Stardew modding scene for the smaller mods). The important part is that only `public` methods can be proxied and accessed by other mods.
* A consuming mod that wants to access this API creates an interface, with methods matching the providing mod's API methods.
* The consuming mod calls `ModLoader.GetAPI<LatterMod.TheNewInterfaceWeJustMade>("<providing mod's name>")`
* The method returns null if the mod is not loaded or the proxy couldn't be created; otherwise it returns an `LatterMod.TheNewInterfaceWeJustMade` instance.
* The consuming mod can call any of the methods it defined. All the types that match get passed directly. Any other types get proxied.

Types and features that work with proxying (the ones I could at least remember about, I've probably forgotten about some):
* Any types both of the sides know (no proxying happening here)
* Types in "out" positions (result type, out parameters)
* Types in "in" positions (normal parameters)
* Types in "in/out" positions (ref parameters)
* Specialized enum handling - mapping enums by their `int` values (currently only enums with the `int` underlying type are supported, but this is the default, so not really a problem)
* Specialized array handling - they act kind of like an "in/out" position, but with special validation (they're still a pain in the ass to use though, I suggest using `IList<>` instead)
* Specialized "reconstructable" type handling - basically any type that has a Deconstruct method with a matching constructor - this makes Dictionary `KeyValuePair`s work with this system, which in turn also allows proxying `IDictionary<>`

Currently Pintail does not have any documentation, other than XML docs (which should be 100% complete for all public types).
There are a lot of unit tests, you can find them [here](https://github.com/Nanoray-pl/Pintail/tree/master/PintailTests).

`Mod.Call` is a way of providing similar APIs right now, but it's super painful to use - both on the providing and consuming side. The providing side has to check what kind of call it received and cast the received arguments to proper types. This is error-prone and can cause exceptions to be thrown if anything doesn't match (types or number of arguments), or requires a lot of validation code. Additionally, the IDE cannot help you in any way, as it straight up doesn't know what's the correct type. The consuming side ALSO has to cast the result to the proper type, so same issues here.

I don't know, you tell me

While I don't have any tML-specific examples, I can post several Stardew mod examples, that are already released and working without any issues.

[Project Fluent - main mod class](https://github.com/Shockah/Stardew-Valley-Mods/blob/2415c38f2d238444b473308096d3a83107c6459f/ProjectFluent/ProjectFluent.cs#L81)
[Project Fluent - the type whose instance we proxy in this case](https://github.com/Shockah/Stardew-Valley-Mods/blob/master/ProjectFluent/FluentApi.cs)
[Project Fluent - the interface on the API provider's side](https://github.com/Shockah/Stardew-Valley-Mods/blob/master/ProjectFluent/PublicAPI/IFluentApi.cs) (+ some other types in the same directory)
[Predictable Retaining Soil - the interface blueprint for which to create the proxy](https://github.com/Shockah/Stardew-Valley-Mods/blob/712f586a42e451216dd4e1c66f3725a67fd72be0/PredictableRetainingSoil/APIs/IFluentApi.cs)
[Predictable Retaining Soil - usage on the API consumer's side](https://github.com/Shockah/Stardew-Valley-Mods/blob/712f586a42e451216dd4e1c66f3725a67fd72be0/PredictableRetainingSoil/PredictableRetainingSoil.cs#L75-L76)
Note that the interface on the API provider's side is completely optional - it is only provided for the ease of use by potential consumers (just copy over the interface to your own mod's code and you can use it).

### ExampleMod updates
~~I did not update the ExampleMod.~~
The next comment has relevant links.

### DIFF
--- a/ExampleMod/Common/Systems/ModIntegrationsSystem.cs
+++ b/ExampleMod/Common/Systems/ModIntegrationsSystem.cs
@@ -2,6 +2,7 @@
 using Microsoft.Xna.Framework.Graphics;
 using System;
 using System.Collections.Generic;
+using Terraria;
 using Terraria.ModLoader;
 
 namespace ExampleMod.Common.Systems
@@ -24,7 +25,8 @@ namespace ExampleMod.Common.Systems
 
 			// Boss Checklist shows comprehensive information about bosses in its own UI. We can customize it:
 			// https://forums.terraria.org/index.php?threads/.50668/
-			DoBossChecklistIntegration();
+			//DoBossChecklistIntegration();
+			DoBossChecklistIntegrationViaAPI();
 
 			// We can integrate with other mods here by following the same pattern. Some modders may prefer a ModSystem for each mod they integrate with, or some other design.
 		}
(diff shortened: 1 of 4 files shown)


Shockah:
I've added a super crude example to Example Mod (Boss Checklist integration), although it definitely would need to be improved to be Example-Mod-worthy before being merged in - it's just to demonstrate and test the feature.

[Example Mod part](https://github.com/tModLoader/tModLoader/pull/2633/commits/6db3bc46dcbcbf20d45fbdba70845094ddb41587#diff-ed805f43266574e2cd5b16477219c383fb47ed1c3df29e3ffb101aff400a4c17)
[Boss Checklist part](https://github.com/Shockah/BossChecklist-fork/commit/5d61a129fc9e3d44b3ccc54b058aa4cf9f391eef)

PR #2635, opened 2022-06-27
Object Oriented Packet System (ModCustomPacket)

doombubbles:
### What is the new feature?

A system for Object Oriented packets utilizing the built-in tModLoader ModType system: `ModCustomPacket`

All Properties (with both a getter and a setter) within the class will be able to be automatically serialized to and from existing ModPackets / BinaryReaders.

The current ways of Sending and Handling ModPackets is all through a "god-class level hardcode" (as the TODO comment in ExampleMod.Networking.cs put it) in the Mod file, and usually leads to modders all creating their own Enums / IDs to handle packets. 
There's also some not-always-intuitive patterns with dealing with client vs server code execution that can be real gotchas to newer modders. 
This system seeks to improve those situations.

The current way to handle packets is obviously still functional, and modders would be able to 100% keep using it if they wanted.

As far as alternative OOP packet systems, I went through a couple iterations and this was the best one, mainly because it's got:

- Minimal reflection overhead; the `Delegate.CreateDelegate` calls create cached delegates that have [great performance](https://www.c-sharpcorner.com/article/boosting-up-the-reflection-performance-in-c-sharp/) getting and setting stuff
- Extensible custom serialization through creating new `ModPropertySerializer` types
- Compatibility with existing Enum/number types systems because the IDs start at -1 and count downward, leaving 0 and up open for easy Enum values

Since this started as a feature in my own separate mod, there may be additional/better ways to interact with the Terraria/tModLoader source code itself that I haven't taken advantage of.

A mod wanting to opt-in to custom packets would make their mod's `Mod.HandlePacket(BinaryReader, int)` method look something like:
```csharp
public override void HandlePacket(BinaryReader reader, int whoAmI) => ModCustomPacket.Handle(this, reader, whoAmI);
```
or
```csharp
public override void HandlePacket(BinaryReader reader, int whoAmI) {
	if (!ModCustomPacket.Handle(this, reader, whoAmI, out int packetType)) {
		// Can still interact with non-ModCustomPackets sent in the old way here using the packetType
	}
}
```

They'd then create a class that extends from `ModCustomPacket`, e.g.
```csharp
public class ExampleSyncPlayerPacket : ModCustomPacket
{
    /// <summary>
    /// This complex type is automatically correctly serialized (by reference)! <see cref="ModPlayerPropertySerializer"/>
    /// </summary>
    public ExampleLifeFruitPlayer ExampleLifeFruitPlayer { get; init; }
    
    // "init" just means that your code can only set this when creating the object and not afterward
    public int ExampleLifeFruits { get; init; }

    public override void HandlePacket() {
        ExampleLifeFruitPlayer.exampleLifeFruits = ExampleLifeFruits;
    }
}
```

Then, wherever they wanted to send the packet they could do so via:

```csharp
...
new ExampleSyncPlayerPacket {
    ExampleLifeFruitPlayer = this, ExampleLifeFruits = exampleLifeFruits
}.Send(toWho, fromWho);
...
```

Or, instead of sending the packet that traditional way, they could simplify things and do this:
```csharp
...
// HandleForAll will ensure the effects of the packet will be applied to all clients and the server, whether playing singleplayer or multiplayer
new ExampleTeleportToStatuePacket {
    NPC = this
}.HandleForAll();
...
```

### ExampleMod updates

`ExampleMod.Networking.cs`: Commented out the old stuff to show the new `ModCustomPacket.Handle` way

`ExamplePerson.cs`: Turning the existing packet interactions into a new `ExampleTeleportToStatuePacket`

`ExampleLifeFruit.cs`: Turning the existing packet interactions into a new `ExampleSyncPlayerPacket`

### Extra

I originally just made this for my own personal use, but it seemed like something that enough people could utilize that I decided to clean it up and PR it here. 

I also admit that I didn't do much research into any existing in-progress implementations of something like this, so if I've reinvented the wheel here or gone in a way different direction that y'all planned then RIP, so it goes lol. Still useful to myself in the interim at least :)

If this is could actually be useful to you guys though, you can also ping me in the tModLoader discord @doombubbles#1701 with questions/concerns. Also let me know if I should've gone with different naming conventions / namespaces for stuff.

P.S. I've tested this more extensively with my own mods than just the Examples in ExampleMod so far

### DIFF
--- a/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
+++ b/ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
@@ -73,11 +73,10 @@ namespace ExampleMod.Content.Items.Consumables
 		}
 
 		public override void SyncPlayer(int toWho, int fromWho, bool newPlayer) {
-			ModPacket packet = Mod.GetPacket();
-			packet.Write((byte)ExampleMod.MessageType.ExamplePlayerSyncPlayer);
-			packet.Write((byte)Player.whoAmI);
-			packet.Write(exampleLifeFruits);
-			packet.Send(toWho, fromWho);
+			new ExampleSyncPlayerPacket {
+				ExampleLifeFruitPlayer = this,
+				ExampleLifeFruits = exampleLifeFruits
+			}.Send(toWho, fromWho);
 		}
 
 		// NOTE: The tag instance provided here is always empty by default.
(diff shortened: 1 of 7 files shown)


PoctorDepper:
Is this still an ongoing development?

PR #2638, opened 2022-06-27
Remove GlobalRecipe

Chicken-Bones:
### What is the new feature?
`GlobalRecipe` is removed. 

- All functionality can be achieved elsewhere
- Less code to maintain, less ways of doing the same thing
- Can tell when a recipe has `Conditions` or `ConsumeItemHooks`
  - Makes advanced recipe features easier, because the availability of 'simple' recipes can be cached. (Helps with #2602)

No

Edit the recipes you want to edit in `ModSystem.PostAddRecipes` to add callbacks/custom conditions. 

Consider `Main.hidePlayerCraftingMenu` (previously added by tML) if you want to disable crafting.

### ExampleMod updates
None, it didn't have any `GlobalRecipe` examples.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalRecipe.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalRecipe.cs
@@ -1,39 +1,10 @@
-﻿namespace Terraria.ModLoader
+﻿using System;
+
+namespace Terraria.ModLoader
 {
-	/// <summary>
-	/// This class provides hooks that control all recipes in the game.
-	/// </summary>
-	public abstract class GlobalRecipe : ModType
+	// class left in for tModPorter
+	[Obsolete("Use Recipe.AddCondition/AddConsumeItemCallback/AddOnCraftCallback or GlobalItem.OnCreate", true)]
+	public class GlobalRecipe
 	{
-		protected sealed override void Register() {
-			ModTypeLookup<GlobalRecipe>.Register(this);
-			RecipeLoader.Add(this);
-		}
-
-		public sealed override void SetupContent() => SetStaticDefaults();
-
-		/// <summary>
-		/// Whether or not the conditions are met for the given recipe to be available for the player to use. This hook can be used for conditions unrelated to items or tiles (for example, biome or time).
-		/// </summary>
-		public virtual bool RecipeAvailable(Recipe recipe) {
-			return true;
-		}
-
-		/// <summary>
-		/// Allows you to make anything happen when the player uses the given recipe. The item parameter is the item the player has just crafted.
-		/// </summary>
-		/// <param name="item">The item created.</param>
-		/// <param name="recipe">The recipe used to create the item.</param>
-		public virtual void OnCraft(Item item, Recipe recipe) {
-		}
-
-		/// <summary>
-		/// Allows to edit the amount of item the player uses in a recipe.
-		/// </summary>
-		/// <param name="recipe">The recipe used for the craft.</param>
-		/// <param name="type">Type of the ingredient.</param>
-		/// <param name="amount">Modifiable amount of the item consumed.</param>
-		public virtual void ConsumeItem(Recipe recipe, int type, ref int amount) {
-		}
 	}
 }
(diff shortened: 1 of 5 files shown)


JavidPack:
We'll need feedback on how modders are using this. I imagine some game mode mods would be using this for global effects, like preventing crafting during certain situations. Adding a Conditions to every recipe in the game to support this seems like it is just muddling things. Similar to how the bestiary doesn't show global loot rules, this provided the same functionality for recipes.

If a mod did reimplement their global recipe checks as conditions, every recipe in the game would have the condition text show up on the guide, and that text wouldn't be too useful depending on the functionality the modder was trying to implement.

absoluteAquarian:
Magic Storage was previously using a GlobalRecipe to "hide" the crafting menu by making all recipes unavailable.
We have since moved to simply using `Main.hidePlayerCraftingMenu` to do that, since it's less of a hack and looks cleaner.

Chicken-Bones:
Intent to merge in 24hrs, unless feedback emerges

de17wpAcqqoWhXwUH:
We do use GlobalRecipe in tsorcRevamp, but exclusively the OnCraft hook. We use it to remove items from a custom item slot that we involve in crafting, which stores the players "souls". Its removal would mean we have to use a method swap instead. That wouldn't be the end of the world, but the ability to do things whenever crafting happens may be useful enough to be worth preserving via moving it somewhere else.

Chicken-Bones:
> We do use GlobalRecipe in tsorcRevamp, but exclusively the OnCraft hook. We use it to remove items from a custom item slot that we involve in crafting, which stores the players "souls". Its removal would mean we have to use a method swap instead. That wouldn't be the end of the world, but the ability to do things whenever crafting happens may be useful enough to be worth preserving via moving it somewhere else.

You can use `GlobalItem.OnCreate(ItemCreationContext context)` and check `if (context is RecipeCreationContext)`
tModPorter directs you there. https://github.com/tModLoader/tModLoader/pull/2638/files#diff-9e864215b616ea1a74827ad0413697a40d6bb2bf3bf06b31309b5fe81f566766R11

direwolf420:
One use of such a hook was to "spoof" crafting conditions like this:
<https://github.com/Fargowilta/Fargowiltas/blob/287b85f1d832ef851c4bae558b7cc5fa81cdff86/FargoGlobalRecipe.cs>
How would that be realized now?

Chicken-Bones:
> One use of such a hook was to "spoof" crafting conditions like this: https://github.com/Fargowilta/Fargowiltas/blob/master/FargoGlobalRecipe.cs#L12 How would that be realized now?

I would recommend an `On` hook in `FindRecipes` or just straight up `RecipeAvailable`. The main point of this PR is to make it clear which recipes are always available, and say, if a mod understands what certain conditions mean, they can make the recipes available in their own system (eg magic storage) by ignoring some of them, etc

That specific use case appears to be asking for ModCondition but uhh... fear. 
Another option would be to use reflection to replace the `InGraveyardBiome` condition or its delegate with a wrapped version

absoluteAquarian:
> > One use of such a hook was to "spoof" crafting conditions like this: https://github.com/Fargowilta/Fargowiltas/blob/master/FargoGlobalRecipe.cs#L12 How would that be realized now?
> 
> I would recommend an `On` hook in `FindRecipes` or just straight up `RecipeAvailable`. The main point of this PR is to make it clear which recipes are always available, and say, if a mod understands what certain conditions mean, they can make the recipes available in their own system (eg magic storage) by ignoring some of them, etc
> 
> That specific use case appears to be asking for ModCondition but uhh... fear. Another option would be to use reflection to replace the `InGraveyardBiome` condition or its delegate with a wrapped version

Magic Storage tracks its own bool for the Ecto Mist recipes and just changes `Main.LocalPlayer.ZoneGraveyard` before checking the Recipe objects via RecipeLoader, but that's likely not the norm.  Most mod devs would have to work around the existing Recipe system.

Chicken-Bones:
@absoluteAquarian for magic storage, you can always check recipe conditions yourself, since `RecipeAvailable` is just a loop over the conditions. If a condition is `InGraveyardBiome` and the player has an 'elemental assembler' then you could skip those conditions.

PR #1476, opened 2021-04-24
ExampleYoyo port

Falcon-05:
### What are the changes to ExampleMod?
Ports ExampleYoyo and its projectile to 1.4.

PR #2642, opened 2022-06-28
Example Yoyo Port to 1.4

Rijam:

Part of #1403, Ported Example Yoyo to 1.4. I made new sprites for it, added more comments, and cleaned up the existing comments.

No.
But, let me know if I should create a ExampleYoyoAdvanced which would use adapted AI instead of just relying on aiStyle 99.


ilija04:
There's also #1476.

Rijam:
Oh you're right. I might turn this into ExampleYoyoAdvanced then.

JavidPack:
I went ahead and merged some of your comments with the other PR and added you as an author of that one. If you'd like to make a more advanced yoyo example with a custom AI later, that is fine, but it isn't a high priority. Closing this one.

PR #2645, opened 2022-06-28
Modernize PlayerLoader, support custom ModPlayer hooks.

Mirsario:
PR does the following:
- Makes `PlayerLoader` use `Terraria.ModLoader.Core.HookList<T>`, instead of its old-old nested class.
- Adds `PlayerLoader.AddModHook` method, in line with `ItemLoader`, `NPCLoader`, and `ProjectileLoader` classes.
- Since `ModPlayer` is something in-between `ModType<,>` and `GlobalType<,>`, and those classes appear as a proof of past mistakes, I had to add `IIndexed` interface with `ushort Index { get; }` in it, and put that on `ModPlayer` and `GlobalType<,>` to allow the former to be used with some `HookList<T>` methods.
- Most of the changes is the replacement of `GlobalType.index` field with the `Index` auto-property. Ignore.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/FilteredEnumerator.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/FilteredEnumerator.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Linq;
+using System.Runtime.CompilerServices;
+
+namespace Terraria.ModLoader.Core
+{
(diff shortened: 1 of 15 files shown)


Chicken-Bones:
Looks good, I'll review and merge this soon. I've got some other hook cleanup to do as well this month.

PR #2651, opened 2022-06-29
OnConsumeItem() callback

andro951:
### What is the new feature?
Add an OnConsumeItem() callback to RecipeLoader.  This would occur slightly after the ConsumeItem() callback that exists.  

RecipeLoader.ConsumeItem(Recipe recipe, int type, ref int amount) only gives access to the item's type and amount.  The main thigs this cuts us off from is item fields/properties that have been changed and more importantly, GlobalItem fields/properties.
RecipeLoader.OnConsumeItem(Recipe recipe, Item item, ref int amountRemaining) would give more options discussed in sample usage.

ConsumeItem could be moved instead of creating a different callback, but I think it would be more beneficial to have both.  
ConsumeItem allows you to affect the total amount of items needed.
OnConsumeItem would allow you to affect the remaining amount needed based on the specific item being looked at.

Note: These can be accomplished with IL edits (I have), but the IL edits get missed my other mods that affect crafting such as Magic Storage and Better Crafting, causing incompatibilities.

1: Transfer values from an ingredient's GlobalItem to the result item's GlobalItem.  (This would require tracking the values by the modder, but that would be relatively easy.)
     Example: True Night's edge with 500 experience + True Excalibur with 300 experience -> Terra Blade with 800 experience.

2: Allow a specific instance of an item to be non-consumable from crafting.  Lets say you have an item in your mod called an infinity stone (cuz I'm creative) that can be installed on onto an item to make it so you can use it infinitely in crafting.  They would set the amountRemaining to 0 and return false to prevent the vanilla consume item code from running.

3: Allow other things to happen when consuming a specific item based on its GlobalItem fields or modified stats.  An example from my mod:  enchantment items are stored in an Item[] (max of 5) in the GlobalItem.  When crafting items with multiple ingredients, they could each have up to 5 enchantments.  I quickspawn the extra ones that don't fit.

### Summary
Many of these are examples from my mod, but in general, this would give modders more options to interact with crafting in unique ways.

### DIFF
--- a/ExampleMod/Common/GlobalItems/ExampleExperienceItem.cs
+++ b/ExampleMod/Common/GlobalItems/ExampleExperienceItem.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Terraria;
+using Terraria.ModLoader;
+using Terraria.ID;
+using Microsoft.Xna.Framework;
+using Terraria.ModLoader.IO;
+
+namespace ExampleMod.Common.GlobalItems
+{
+	public class ExampleExperienceItem : GlobalItem {
+		public int experience;
+		public int level;
+		public static int experiencePerLevel = 100;
+
+		public override bool InstancePerEntity => true;
+
+		public override bool AppliesToEntity(Item entity, bool lateInstantiation) {
+			//Apply this GlobalItem to all swords
+			return entity.DamageType == DamageClass.Melee && entity.noMelee == false && entity.useStyle == ItemUseStyleID.Swing;
+		}
+
+		public override GlobalItem Clone(Item item, Item itemClone) {
+			//Overriding clone is required for GlobalItems that set InstancePerEntity to true.
+			//It is expremely important to include complex fields and properties such as arrays, lists and dictionaries in the Clone method.
+			//If they are not included, they will act like static fields or properties.
+			ExampleExperienceItem clone = (ExampleExperienceItem)base.Clone(item, itemClone);
+			clone.experience = experience;
+			return clone;
+		}
+
+		public override void LoadData(Item item, TagCompound tag) {
+			experience = tag.Get<int>("experience");//Load experience tag
+			item.value += experience * 5;
+			UpdateLevel();
+		}
+		public override void SaveData(Item item, TagCompound tag) {
+			tag["experience"] = experience;//Save experience tag
+		}
+
+		public override void OnHitNPC(Item item, Player player, NPC target, int damage, float knockBack, bool crit) {
+			//The sword gains experience when damaging an npc.
+			int xp = damage;
+			experience += xp;
+			item.value += xp * 5;
+			UpdateLevel();
+		}
+
+		public void UpdateLevel() {
+			level = experience / experiencePerLevel;
+		}
+
+		public override void ModifyTooltips(Item item, List<TooltipLine> tooltips) {
+			if (experience > 0) {
+				UpdateLevel();
+				tooltips.Add(new TooltipLine(Mod, "level", $"Level: {level}") { OverrideColor = Color.LightGreen });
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -153,6 +153,68 @@
...
+-						else {
++							else {
+-							num -= item.stack;
++								num -= item.stack;
+-							array[k] = new Item();
++								array[k] = new Item();
++							}
+ 						}
+ 					}
+ 				}
```
Chicken-Bones, line 179: Use a goto to avoid changing indentation.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -153,6 +153,68 @@
...
+-							NetMessage.SendData(32, -1, -1, null, Main.player[Main.myPlayer].chest, l);
+-
+-						num = 0;
+-					}
+-					else {
+-						num -= item.stack;
+-						array[l] = new Item();
+-						if (Main.netMode == 1 && Main.player[Main.myPlayer].chest >= 0)
+-							NetMessage.SendData(32, -1, -1, null, Main.player[Main.myPlayer].chest, l);
+-					}
```
Chicken-Bones, line 197: comment out code rather than removing large amounts of it.

ExampleMod/Common/GlobalItems/ExampleExperienceItem.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Terraria;
+using Terraria.ModLoader;
+using Terraria.ID;
+using Microsoft.Xna.Framework;
+using Terraria.ModLoader.IO;
+
+namespace ExampleMod.Common.GlobalItems
+{
+	public class ExampleExperienceItem : GlobalItem {
```
Chicken-Bones, line 14: `ExampleMeleeExperienceGlobal`?

ExampleMod/Common/GlobalItems/ExampleExperienceItem.cs
```diff
@@ -0,0 +1,77 @@
...
+		public static int experiencePerLevel = 100;
+
+		public override bool InstancePerEntity => true;
+
+		public override bool AppliesToEntity(Item entity, bool lateInstantiation) {
+			//Apply this GlobalItem to all swords
+			return entity.DamageType == DamageClass.Melee && entity.noMelee == false && entity.useStyle == ItemUseStyleID.Swing;
+		}
+
+		public override GlobalItem Clone(Item item, Item itemClone) {
```
Chicken-Bones, line 26: overriding clone is not required anymore for value typed fields, they will be cloned correctly automatically.

ExampleMod/Common/GlobalItems/ExampleExperienceItem.cs
Chicken-Bones: Needs a comment about the mechanic... currently, spawns coins when upgraded? I dunno...

Chicken-Bones (changes requested):
I'm primarily unsure about the need to pass information from the 'on consumed' callback into the 'set result' callback. I'd rather if:
- You could modify the result item as each item is consumed or
- The 'OnCraft' callback contained a list of all the items which participated in the recipe (more preferable I think)

The second thing I'm not sure about is the need to register a callback on every affected recipe
1. For use case "transfer values from an ingredient to the result", per-recipe may make sense if it's just a special thing for a couple of recipes, but as an overall mechanic (like experience transfer) it's not really a property of the 'recipe', it's really applied to all recipes that produce a certain type of item.
2. For use case "make a certain item non consumable" that should be a property of the item (or even just a `ModItem/GlobalItem` hook
3. For use case like "bonus items on craft depending on inputs" it also makes sense to do this 'on craft' with the full list of ingredients available.

I'm interested to hear thoughts on this.
If you need help with the implementation, let me know.

Chicken-Bones:
Closed in favor of #2716

PR #2671, opened 2022-07-03
Toggle fullscreen when opening icon file picker, set default path

ghost:
### What is the bug?
On my machine, when clicking the icon in the publish mods UI and in fullscreen mode, the file picker appears behind tModLoader and I cannot interact with it. This softlocks my entire machine as I cannot close it or tModLoader.

Additionally, the file picker does not open with the default path set, which is a minor inconvenience.

### How did you fix the bug?
Toggle fullscreen temporarily if necessary when opening the file picker.

Implement the necessary logic to set the default path, and do it for this usage.

- The toggling fullscreen logic could be put as part of the file picker itself, but I didn't want to mess with this too much here.
- Instead of modifying the default Terraria file picker code, just have a TML version of it which is the same but with this change applied.

### Screenshot
Default path being set for ExampleMod, TML running in windowed mode temporarily:
tmlfix

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
@@ -119,11 +119,27 @@
  			};
  		}
  
-@@ -634,6 +_,7 @@
+@@ -634,10 +_,22 @@
  				new ExtensionFilter("Image files", "png", "jpg", "jpeg")
  			};
  
-+			// TODO: support default folder, set default folder to mod source folder for mods.
- 			string text = FileBrowser.OpenFilePanel("Open icon", extensions);
++			// Toggle fullscreen if neccessary to prevent file picker opening in background and locking user input
++			bool wasFullscreen = Main.graphics.IsFullScreen;
++			if (wasFullscreen) {
++				Main.ToggleFullScreen();
++			}
++
++			string openPath = Path.GetFullPath(Path.GetDirectoryName(_previewImagePath));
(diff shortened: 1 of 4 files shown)


direwolf420:
Can you check if the same softlock happens if you press any other buttons in tml UI (such as Open Logs when a crash happens)?

ghost:
> Can you check if the same softlock happens if you press any other buttons in tml UI (such as Open Logs when a crash happens)?

Open Logs, Open Sources, etc. are all fine. I guess because it is a separate process & window which gets opened, whereas the file picker kind of "belongs" to the TML process. I don't really know the exact terminology or reasons, probably partially just Windows (maybe even W11-specific) desktop environment behaviour, but it does happen 100% of the time for me.

If there are other areas that use the file picker I'm sure they'd have the same issue.

Chicken-Bones:
I'm not sure about the cross-platform effects of this. For windows specifically, there's probably a way to get the native handle of the file picker and bring to front. Especially because FNA doesn't support 'exclusive' fullscreen

ghost:
How much of an advantage would that actually be over just temporarily toggling fullscreen (which is probably more likely to work on other platforms if they happen to suffer from the same issue)?

For the record, disabling fullscreen causes the file picker to show in front on my machine, I'm not alt tabbing into it in that screenshot.

Chicken-Bones:
Well I just don't know. The fullscreen stuff and device resets have always been janky. I'd rather leave other platforms alone unless it can be verified on them.

ghost:
> Well I just don't know. The fullscreen stuff and device resets have always been janky. I'd rather leave other platforms alone unless it can be verified on them.

Fair enough

PR #2672, opened 2022-07-03
Fix #2644 Server Crash

Solxanich:

PR #2689, opened 2022-07-05
Better fix for docker in non-interactive mode

Chicken-Bones:
### What is the bug?

The fix for #2672 will cause a background thread to spin forever, wasting CPU and raising core contention.

### How did you fix the bug?

If input returns null, it's not coming back. Exit with a message.

Try multiple times? Sleep between retries? Doubt this would make any difference.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -831,14 +831,18 @@
  			}
  
  			YouCanSleepNow();
-@@ -4538,8 +_,15 @@
+@@ -4538,8 +_,19 @@
  
  		public static void startDedInputCallBack() {
  			while (!Netplay.Disconnect) {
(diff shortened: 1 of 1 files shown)


Solxanich (approved):
Approved, no comments

hkiang01:
doesn't seem to work per https://github.com/tModLoader/tModLoader/pull/2672#issuecomment-1175479908

Tamaized:
Confirmed working:
> `[17:54:14] [Server Input Thread/WARN] [tML]: Console input not connected, ignoring...`

Also works when running docker with `-i` flag in which case the above log line is not printed and CLI input works as expected.

Chicken-Bones:
@hkiang01 not sure what you experienced on your test, maybe you didn't get the right build, or had a different issue. Please send a `server.log` if further testing fails.

PR #2691, opened 2022-07-06
Fix some UPNP issues

ghost:
### What is the bug?
Current UPNP system is a bit of a mess and doesn't work on certain setups:
- Some devices will respond to all SSDP requests regardless of if they provide the correct service/match the correct device type (of course this is evil, but what can you do?). Terraria will still try and use them in this case, and end up with an unsupported gateway error. 
- The HOST field in the SSDP request was previously being set to the client's IP, which is incorrect (although I haven't seen this cause issues).
- The HTTP header in the SSDP request was using LF to break lines instead of CRLF, which is required (this does cause issues, it doesn't work on my router without this change).

### How did you fix the bug?
- Wait to receive multiple responses to SSDP requests (currently arbitrarily capped at 20), until we get one that actually supports the necessary service.
- Set the HOST field in the SSDP request correctly to `239.255.255.250:1900`.
- Break the lines in the SSDP request correctly with CRLF (`\r\n`).

With this applied, UPNP now works properly on my network (which was suffering from a combination of the CRLF issue and the misbehaving devices issue).

Tbh at some point this whole system may benefit from a refactor/rewrite.

IPv6 support in the future may be nice too.

### DIFF
--- a/patches/TerrariaNetCore/Terraria/UwUPnP/Gateway.cs
+++ b/patches/TerrariaNetCore/Terraria/UwUPnP/Gateway.cs
@@ -20,6 +20,15 @@ namespace UwUPnP
 		private readonly string serviceType = null;
 		private readonly string controlURL = null;
 
+		//   Allow user to set this env var in case they have some weird
+		// hardware that only accepts LF-seperated SSDP header lines (this
+		// breaks the spec, but it may well happen still).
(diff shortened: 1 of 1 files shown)


ghost:
Have changed slightly to allow reverting to old LF-separated SSDP behaviour with an env var, in case of weird hardware that this could break (I think it's highly unlikely)

ghost:
@Rartrin does this approach seem ok? I think hosts that only work with LF will be extremely rare, if they exist at all - all other UPnP libraries I looked at use only CRLF.

PR #2698, opened 2022-07-07
Fix #2697.

### LINKED ISSUE #2697: projectile.ContinuouslyUpdateDamage does not reset when a projectile is killed.
### Version

Latest 1.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

`projectile.ContinuouslyUpdateDamage` can be applied to a projectile in `SetDefaults()`
When set, this causes the projectile to constantly update damage. This is all good, but the projectile having this property does not go away when the projectile dies. Instead, it stays attached to the projectiles index in the `Main.projectile` array.

This can cause unexpected behavior for certain player-used items that may not have damage classes, but this is especially prevalent for boss, hostile, or NPC projectiles, which do classless damage. As such, the `projectile.ContinuouslyUpdateDamage` property will make them do no damage.

### Log File

There is no log file for this error.

### Steps to reproduce

1. Make a `ModProjectile` with the `projectile.ContinuouslyUpdateDamage` property set to true.
2. Put something like
`if(projectile.ContinuouslyUpdateDamage)`
`{`
`      Main.NewText("Projectile Type: " + projectile.type);`
`      Main.NewText("Projectile Index: " + projectile.whoAmI);`
`}` 
inside of a `GlobalProjectile.AI()` override.
3. Summon/Shoot the projectile many times, keep track of the indexes in chat that have the property set to true.
4. Despawn the summons/kill the projectiles (shoot them into a wall or something).
5. Use a different item and watch how its projectiles now have the property set to true.
6. Summon an NPC such as Retinazer and watch as the projectiles do zero damage (but only when their index appears in chat) 

### Expected Behavior

Projectiles should not inherit this attribute from projectiles that previously existed.

### Actual Behavior

The property stays stuck to the index of the projectile.

### Reproduction frequency

I reproduced this 3 times out of 3 attempts I tried to produce it.

### Additional Information

If the is a different way you are supposed to use this property, then I would like to be informed about it.

Antirhinnum:
### What is the bug?
#2697 

### How did you fix the bug?
Reset `Projectile.ContinuouslyUpdateDamage` in `Projectile.SetDefaults()`.

No.

### DIFF
--- a/patches/tModLoader/Terraria/Projectile.cs.patch
+++ b/patches/tModLoader/Terraria/Projectile.cs.patch
@@ -94,7 +94,7 @@
  				for (int j = 0; j < 200; j++) {
  					perIDStaticNPCImmunity[i][j] = 0u;
  				}
-@@ -221,7 +_,38 @@
+@@ -221,7 +_,39 @@
  			}
  		}
  
@@ -129,6 +129,7 @@
 +			ArmorPenetration = 0;
 +			CritChance = 0;
 +			WhipSettings = default;
++			ContinuouslyUpdateDamage = false;
 +			nameOverride = null;
  			ownerHitCheckDistance = 1000f;
  			counterweight = false;


Chicken-Bones:
Correct fix. Ty, @Solxanich maybe we should make a 'pending stable branch' to cherry-pick as you go. Either way, this one's approved.

PR #2707, opened 2022-07-11
Allow keybind text to use translated text

Cyrillya:
### What is the new feature?
The keybind name starts with "$" will be automatically translated.

It should be part of the localization of Mods.

Maybe remake the whole keybind system?

For ExampleMod: When English is selected, "Random Buff" is displayed. When Chinese is selected, "随机增益" is displayed (Chinese for Random Buff)
It will be good for mods that have multiple languages and keybinds.

### ExampleMod updates
- Add zh-Hans.hjson, which only contains the translated text of "Random Buff"
- Changed the registered keybind name from "Random Buff" to "$Mods.ExampleMod.Keybind.RandomBuff"
- Add a translation key for the random buff keybind


Configuration.targets
```diff
@@ -0,0 +1,18 @@
+<Project>
```
ghost: did you mean to include this?

Cyrillya: No I don't. I don't know why it was included. Deleted now.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIKeybindingListItem.cs.patch
```diff
@@ -1,6 +1,30 @@
...
++			_bindName = bindName;
+ 			_inputmode = mode;
+ 			_color = color;
+ 			base.OnClick += OnClickMethod;
+@@ -99,11 +_,19 @@
+ 		private string GetFriendlyName() {
+ 			switch (_keybind) {
+ 				default:
++					if (_modName != "Terraria" && _bindName.StartsWith('$'))
++						return $"{_modName}: {Language.GetTextValue(_bindName.Remove(0, 1))}";
```
ghost: `.Substring(1)`

ghost: is there an advantage/disadvantage to doing the localization here instead of in the `_keybind` getter, or in the `UIKeybindingListItem` constructor?

Configuration.targets
ghost: ✅ resolved

patches/tModLoader/Terraria/GameContent/UI/Elements/UIKeybindingListItem.cs.patch
Cyrillya: Terraria uses the `_keybind` text to dectect if a keybind is triggered or something. If I directly doing the localization in the `_keybind` getter, it crashes the game because the `KeyConfiguration.KeyStatus` doesn't contain a parameter for the translated text. And if I want to fix it I will have to make a lot of changes to the keybind system. It may also cause a problem that if you change your language, you'll have to re-bind the key again.

ghost: Sounds very reasonable, all good

ghost (approved):

ExampleMod/Common/Systems/KeybindSystem.cs
```diff
@@ -10,7 +10,7 @@ public class KeybindSystem : ModSystem
 
 		public override void Load() {
 			// Registers a new keybind
-			RandomBuffKeybind = KeybindLoader.RegisterKeybind(Mod, "Random Buff", "P");
+			RandomBuffKeybind = KeybindLoader.RegisterKeybind(Mod, "$Mods.ExampleMod.Keybind.RandomBuff", "P");
```
steviegt6: A comment should be added here explaining the use of the `$` character.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIKeybindingListItem.cs.patch
```diff
@@ -1,6 +1,30 @@
...
+ 	{
+ 		private InputMode _inputmode;
+ 		private Color _color;
+-		private string _keybind;
++		private readonly string _modName;
++		private readonly string _bindName;
++		private string _keybind => _modName == "Terraria" ? _bindName : $"{_modName}: {_bindName}";
+ 
+-		public UIKeybindingListItem(string bind, InputMode mode, Color color) {
++		public UIKeybindingListItem(string modName, string bindName, InputMode mode, Color color) {
```
steviegt6: Should probably keep the old constructor and make it internal, then have it call `: this("Terraria", bind, mode, color)`.

patches/tModLoader/Terraria/GameContent/UI/States/UIManageControls.cs.patch
```diff
@@ -175,7 +176,8 @@
 +						return container;
 +					}
 +
- 					return new UIKeybindingListItem(bind, currentInputMode, color);
+-					return new UIKeybindingListItem(bind, currentInputMode, color);
++					return new UIKeybindingListItem("Terraria", bind, currentInputMode, color);
```
steviegt6: This patch can be removed if the internal constructor suggestion is implemented.

steviegt6 (changes requested):
Very small stuff.

ExterminatorX99:
Should it match how lang keys are used in display name and tooltip?
`{$Key.Here}`

I also notice you can't mix text and key. You can only use one or the other

Cyrillya:
> Very small stuff.

Changes were made

ghost:
> Should it match how lang keys are used in display name and tooltip?
> `{$Key.Here}`
> 
> I also notice you can't mix text and key. You can only use one or the other

IMO the first follows from the second here. If it allowed interpolation, then I think the `{$x}` syntax makes sense, but with the current implementation the current syntax seems more fitting to me.

Wrt. to allowing mixing/interpolation, maybe it would be nice in the future, but how much use case is there for this? If it seems likely someone will want to implement it in the future as an improvement, it may be worth using `{$x}` syntax now to allow forward-compatibility. I see no reason why this patch has to include this functionality from the start though, nothing wrong with incremental improvements where/if someone finds interpolation necessary.

patches/tModLoader/Terraria/ModLoader/ModKeybind.cs
```diff
@@ -11,13 +12,16 @@ namespace Terraria.ModLoader
 	public class ModKeybind
```
Chicken-Bones: We should make this extend `ModType` and be autoloadable, but that is perhaps outside the scope of this PR.

patches/tModLoader/Terraria/ModLoader/ModKeybind.cs
```diff
@@ -11,13 +12,16 @@ namespace Terraria.ModLoader
...
 	{
 		internal readonly Mod mod;
-		internal readonly string name; // name from modder: "Random Buff"
-		internal readonly string uniqueName; // eg: "Example Mod: Random Buff" (currently also display name)
+		internal readonly string name; // name from modder: "RandomBuff"
+		internal readonly string uniqueName; // name saved to disk: "ExampleMod: RandomBuff"
 		internal readonly string defaultBinding; // from mod.Load
+		internal readonly string displayNameKey; // Mods.ExampleMod.Keybind.RandomBuff
+		internal string DisplayName => Language.Exists(displayNameKey) ? Language.GetTextValue(displayNameKey) : name; // "Random Buff"
```
Chicken-Bones: Given these are all `internal`, they should be safe to change. Can we make them consistent with other `ModType` classes?

Eg, `Mod`, `Name`, `FullName`, `ModTranslation DisplayName`

`DisplayName = LocalizationLoader.GetOrCreateTranslation(Mod, $"Keybind.{Name}");`

Chicken-Bones (changes requested):

DanYami:
so when can we fix the bug this has introduced

Chicken-Bones:
> so when can we fix the bug this has introduced

When you make an issue report detailing the bug (or better yet, a PR)

PR #2716, opened 2022-07-14
Add consumed items list to Recipe.OnCraft and RecipeCreationContext

andro951:
### What is the new feature?
Add a cloned list of consumed items to the `RecipeCreationContext` for `Mod/GlobalItem.OnCreate` and `Recipe.OnCraft` callbacks. 

Also moves `OnCreate` after `OnCraft` (previously it was before). 
- recipes can use `OnCraft` to adjust the crafted item, `OnCreate` gets the fully output item from the recipes
- @Chicken-Bones checked for public github usages of `OnCraft` and none of them would be affected by the re-order

Sample use cases:
-  Transfer `GlobalItem` values from ingredients to the result.
Example: True Night's edge with 500 experience + True Excalibur with 300 experience -> Terra Blade with 800 experience.

-  Allow other things to happen when consuming a specific item based on its `GlobalItem` fields or modified stats. An example from my mod: enchantment items are stored in an Item[] (max of 5) in the `GlobalItem`. When crafting items with multiple ingredients, they could each have up to 5 enchantments. I quickspawn the extra ones that don't fit.

A previous implementation was #2651 would work, but requires modders to maintain their own consumed item list with `OnCreate` and clear it at an appropriate time.

Note: This can be accomplished with IL edits (I have), but the IL edits get missed my other mods that affect crafting such as Magic Storage and Better Crafting, causing incompatibilities.

```cs
        public override void OnCreate(Item item, ItemCreationContext context) {
            if (context is RecipeCreationContext rContext) {
                foreach (Item ingredient in rContext.ConsumedItems) {
                    ...
```
or
```cs
recipe.AddOnCraftCallback((r, item, consumedItems) => ...);
```

### Porting Notes
- If you have a mod with custom crafting, `RecipeLoader.OnCraft` now calls `ItemLoader.OnCreate` so make sure you don't double-call the hooks.

### DIFF
--- a/ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
+++ b/ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
@@ -16,12 +16,7 @@ namespace ExampleMod.Common.GlobalItems
 	{
 		public int experience;
 		public int level;
-		public List<Item> consumedWeapons = new List<Item>();
 		public static int experiencePerLevel = 100;
-		private static int chosenPrefix = -1;
-		private static bool reforgeingItem = false;
-		private static int chosenRarity = -1;
-		private static bool skipAllowCheck = false;
 
 		public override bool InstancePerEntity => true;
 
@@ -33,13 +28,11 @@ namespace ExampleMod.Common.GlobalItems
 		public override void LoadData(Item item, TagCompound tag) {
 			experience = tag.Get<int>("experience");//Load experience tag
 			item.value += experience * 5;
-			consumedWeapons = tag.Get<List<Item>>("consumedItems");
 			UpdateLevel();
 		}
 
 		public override void SaveData(Item item, TagCompound tag) {
 			tag["experience"] = experience;//Save experience tag
-			tag["consumedItems"] = consumedWeapons;
 		}
 
 		public override void OnHitNPC(Item item, Player player, NPC target, int damage, float knockBack, bool crit) {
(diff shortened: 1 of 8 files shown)


ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		public override GlobalItem Clone(Item item, Item itemClone) {
+			//Overriding clone is required for GlobalItems that set InstancePerEntity to true.
+			//It is expremely important to include complex fields and properties such as arrays, lists and dictionaries in the Clone method.
+			//If they are not included, they will act like static fields or properties.
+			//It is not required to include simple fields and properties such as int, float, bool.
+			ItemWithGrowingDamage clone = (ItemWithGrowingDamage)base.Clone(item, itemClone);
+			return clone;
+		}
```
Chicken-Bones: As much as the comment is nice here, it's not a good example to override a method when you don't need to.

ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		public override void ModifyTooltips(Item item, List<TooltipLine> tooltips) {
+			if (experience > 0) {
+				UpdateLevel();
+				tooltips.Add(new TooltipLine(Mod, "level", $"Level: {level}") { OverrideColor = Color.LightGreen });
+				string levelString = $" ({(level + 1) * experiencePerLevel - experience} to next level)";
+				tooltips.Add(new TooltipLine(Mod, "experience", $"Experience: {experience}{levelString}") { OverrideColor = Color.White });
+			}
+		}
+		public override void OnCreate(Item item, ItemCreationContext context, List<Item> consumedItems) {
```
Chicken-Bones: The `List<Item> consumedItems` needs to go on the `RecipeCreationContext`.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -153,6 +153,22 @@
  				if (num <= 0)
  					continue;
  
+@@ -138,6 +_,7 @@
+ 						break;
+ 
+ 					if (item.IsTheSameAs(item2) || useWood(item.type, item2.type) || useSand(item.type, item2.type) || useFragment(item.type, item2.type) || useIronBar(item.type, item2.type) || usePressurePlate(item.type, item2.type) || AcceptedByItemGroups(item.type, item2.type)) {
++						RecipeLoader.ConsumedItems.Add(item.Clone());
```
Chicken-Bones: the clone should contain the stack size of how many were actually consumed.

Chicken-Bones (changes requested):

ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
andro951: I would like to try to keep the comments.  I’ll try to think of a simple way to use a list or dictionary.  Those are things I struggled with a huge amount when creating my GlobalItem class, so I’d like to pass it on to others.  I’ll probably just have it copy the consumed item list and use that somehow.

ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
andro951: Will do, thanks.

patches/tModLoader/Terraria/Recipe.cs.patch
andro951: Yup, that makes sense.  I missed that.

andro951:
Comment deleted.

ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
andro951: Removed

ExampleMod/Common/GlobalItems/ItemWithGrowingDamage.cs
andro951: Fixed.  Now part of the RecipeCreationContext

patches/tModLoader/Terraria/Recipe.cs.patch
andro951: Fixed.  I had to add a few extra lines to track the stack number, but I made it as clean as I could.

PR #2717, opened 2022-07-15
Grab Bag Loot Modernization

BasicallyIAmFox:
### What is the new feature?
This new feature replaces OpenVanillaBag, PreOpenVanillaBag and few other methods with new ModifyItemLoot and ModifyGlobalLoot in correspondent classes.
It could allow for mods such as Boss Checklist to check boss bag loot without need for other mod to add all loot for its boss manually in Boss Checklist.

I feel like this might be difficult if you want to add new item to vanilla pool, and NPC Loot way allows you to do that (even if its not best way read-wise), and so I had to IL edit vanilla [insert literally any method that can spawn loot] method to add just single item in the item pool.

No, I don't think so?

Same as ModifyNPCLoot pretty much.


BasicallyIAmFox:
TODO/REDO: Herb Bags, Presents and Goodie Bags dont work (needs fix)
After finishing: Example Mod changes

EDIT: for both done

BasicallyIAmFox:
image
kinda offtopic, but because of new way i managed to make tha thting in terraria

EDIT: percentages from wiki:
image
image

BasicallyIAmFox:
should i add new ItemDropRules, such as with ability to drop 2 items from options or more?

PR #2727, opened 2022-07-17
Mod Packs Revamp Part 1 of 2

Solxanich:
### What is the new feature?
A rework of the Mod Pack feature to archive a copy of all mods, the steam publish ids, the tml version, and their configs for later reuse or deployment to a server.

This change provisions for setting up servers/playthroughs that continue uninterrupted during month turnovers.

There are a few pieces not fully fleshed out yet:
1) Setting up a second instance to play tml/ copy over to a server
2) Syncing Server mods to an <IP ADDRESS> folder in ModPacks so that it is appropriately non-impacting
3) And probably a half dozen bug fixes to be made because of the changes such as refreshing the menu when a mod pack gets instanced / removed

Export Instance - Creates a new folder in the Install Folder called '<Mod Pack Name>' containing 
1) a copy of all mods
2) a copy of all configs
3) enabled.json
4) a CLI args text file to auto direct any game starts to use the provided mods and configs.
The instance can be run fully indepedent to regular steam install, making it ideal for getting files for a server or having a dedicated, indepedent playthrough.

Each Mod Pack now fully saves a copy of all .tmod and .cfg files for later use/preservation

The Import Mod Pack (Local) button - Adds the ModPack/Mods folder to the search path so that it gets loaded instead of Steam Workshop mods.

.... More information to be added before merge (I Hope I remember)

Added documentation to the wiki: https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-players#how-do-i-use-mod-packs-in-14

Sample Mod Packs for testing: 
[ModPacks.zip](https://github.com/tModLoader/tModLoader/files/9128630/ModPacks.zip)


patches/tModLoader/Terraria/ModLoader/UI/UIModPacks.cs
```diff
@@ -260,5 +212,117 @@ internal static bool IsValidModpackName(string name)
...
+				}
+
+				// Export non-workshop mods to the modpack
+				foreach (var mod in modsManualAll) {
+					if (!mod.Contains(enabled))
+						continue;
+
+					File.Copy(mod, Path.Combine(modsPath, Path.GetFileName(mod)), true);
+					break;
+				}
```
Chicken-Bones: you can use `ModLoader.Mods` and `Mod.File.path` rather than iterating load dirs and hoping you match the behavior of `ModOrganizer` correctly

PR #2729, opened 2022-07-19
Publicize WorldGen local variables & private variables

MutantWafflez:
### What is the new feature?

Converts all local variables within the `WorldGen.GenerateWorld` method to be `public static` variables of the same name in the WorldGen class, allowing modders to freely modify them. Also publicizes basically all `private` variables currently within the `WorldGen` class.

There are some variables within the `GenerateWorld` method that are helpful for modders (pyramid locations, number of pyramids, jungle X values, dungeon side, etc.), but they are completely trapped within the method's scope normally.

No; there is likely to be a WorldGen rework down the line, but for the time being, this will have to do.

During WorldGen, all you have to do is refer to the variables to change them or get their values, like you would for any other variable.


patches/tModLoader/Terraria/ModLoader/WorldGenLocals.cs
```diff
@@ -0,0 +1,135 @@
...
+	/// of the GenerateWorld method in the WorldGen class.
+	/// </summary>
+	public class WorldGenLocals
+	{
+
+		public int DungeonSide {
+			get;
+			internal set;
+		}
+
```
Chicken-Bones: Squish all these together. One line per property, no spacing between them.

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Terraria
+{
+	public partial class WorldGen
+	{
+		public static WorldGenLocals GenLocalValues;
+		internal static event Action UpdateGenLocals;
+
+		internal static void InvokeGenLocalUpdate() {
+			UpdateGenLocals?.Invoke();
+		}
```
Chicken-Bones: Remove the `event` modifier. Then:
- You can use `UpdateGenLocals =` rather than `+=`
- You don't need to unsubscribe (`-=`)
- You don't need this `InvokeGenLocalUpdate` method

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Terraria.ModLoader;
+
+namespace Terraria
+{
+	public partial class WorldGen
+	{
+		public static WorldGenLocals GenLocalValues;
```
Chicken-Bones: arguably, just put all the variables in this partial. Throw a `#region` around them if you want, or just a // local WorldGen.GenenerateWorld` values.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -361,11 +366,45 @@
...
+@@ -5731,6 +_,43 @@
  			int logX = -1;
  			int logY = -1;
  			int dungeonLocation = 0;
 +
++			GenLocalValues = new WorldGenLocals();
++			void SyncLocals() {
++				if (GenLocalValues is null) {
++					return;
++				}
```
Chicken-Bones: no need for this null check.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -361,11 +366,45 @@
...
++				GenLocalValues.SnowBottom = snowBottom;
++				GenLocalValues.SkyLakeCount = skyLakes;
++				GenLocalValues.LeftBeachEnd = leftBeachEnd;
++				GenLocalValues.RightBeachStart = rightBeachStart;
++				GenLocalValues.JungleOriginX = jungleOriginX;
++				GenLocalValues.SnowOriginLeft = snowOriginLeft;
++				GenLocalValues.SnowOriginRight = snowOriginRight;
++				GenLocalValues.FallenLogX = logX;
++				GenLocalValues.FallenLogY = logY;
++				GenLocalValues.DungeonLocation = dungeonLocation;
```
Chicken-Bones: This is quite smart, but you need to copy the values back the other way if it's a modded pass.

Chicken-Bones (commented):
So I reviewed this implementation buuut...

You'd be better off just declaring static variables with the same names as the vanilla locals, and then patching out the type decl on the locals so they reference the static vars instead.

PR #2735, opened 2022-07-20
update steamcmd setup to use full paths

starspritechippy:
### What is the bug?
For setting up a dedicated server, the `steamcmd` utility incorrectly uses paths relative to its usual install directory rather than the `Setup_tModLoaderServer.sh` script's path. Mods get installed into `/home/user/.steam/` rather than, e.g. `/home/user/.local/share/Steam/tmod`.

### How did you fix the bug?
Using ´pwd´ gets a full path to the working directory. Going up one level and then to the tmod directory ensures the mods are installed to the right directory.

- Uploading the mods manually
- Manually patching the script
- Using `-modpath /home/user/.steam/tmod` when launching the server


guillheu:
Hey it looks like there's still a slash `/` missing after the `$(pwd)`, making it so that the installation directory is something like `/current_directory../tmod` instead of `/current_directory/../tmod`.

Just changing
`steamcmd +force_install_dir $(pwd)../tmod +login anonymous +app_update 1281930 +quit`
to
`steamcmd +force_install_dir $(pwd)/../tmod +login anonymous +app_update 1281930 +quit`
should do the trick.

for reference, i'm having this issue on both a ubuntu-based container and my arch system.

PR #2740, opened 2022-07-21
Add hook ModifyCollisionData to NPC

Cyrillya:
### What is the new feature?
Add a hook ModifyCollisionData to NPC (Vanilla code in `NPC.GetMeleeCollisionData`)
There should be a way for modders to modify npc's hitbox. And if they want they can make it so when players hit a certain part of the npc, the damage will be modified.


JavidPack:
Looks right to me. Am I correct that these aren't weak spots, but parts that do more damage to the player? Maybe MinionBoss can do more damage with the mouth than a glancing collision, that would require some geometry

Tyfyter:
You are correct, and in vanilla this method is used primarily for enemies with weapons such as armed zombies and ogres

JavidPack:
It looks like the immunity cooldown slot does have some use, is there a reason it was removed?

PR #2768, opened 2022-07-28
Add Pre- and PostDrawHeldItem hooks to ModItem for custom item use/hold drawing

ghost:
### What is the new feature?

Added a `PreDrawHeldItem(PlayerDrawSet drawInfo)` and `PostDrawHeldItem(PlayerDrawSet drawInfo, List<DrawData> helItemDrawData)` hook to `ModItem`, allowing for finer control of drawing logic for used and held items. Vanilla drawing can be skipped in line with other `PreDraw` methods.

Custom hold or use animations and other drawing effects were previously either difficult to implement, or had a lot of overhead. This hook should streamline custom drawing so you no longer have to rely on continuously running condition checks, or strange IL edits.

Not that I can think of. The hook functions similar to other Pre- and PostDraw hooks. One design choice is to provide the hook with the entire `PlayerDrawSet`, instead of just the held item draw data. This opens the door to even more effects.

The ExampleMod item shows a basic example of what can be done with the hook.

### ExampleMod updates

Added `ExampleCustomDrawnSword`, showing off a custom use and hold style, as well as editing drawing data for any item.


patches/tModLoader/Terraria/DataStructures/PlayerDrawLayers.cs.patch
```diff
@@ -338,6 +338,30 @@
...
++
++			// Sets up draw data, only adds vanilla data if the (mod)item decides to draw
++			DrawPlayer_27_HeldItem(ref drawInfo);
++
++			// Restore original draw data and merge new (+vanilla) held item data
++			drawInfo.DrawDataCache = retainedDrawData;
++			ItemLoader.PostDrawHeldItem(drawInfo.heldItem, ref drawInfo, heldItemDrawData);
++			drawInfo.DrawDataCache.AddRange(heldItemDrawData);
++		}
++
```
ExterminatorX99: Can be moved into the TML partial

JavidPack:
Why did you close this? It was still being worked on.

Issue #2772, opened 2022-07-29
Add ModItem.MeleePrefix/RangedPrefix/etc. hooks


Tyfyter:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Hooks for the ItemLoader.MeleePrefix, ItemLoader.WeaponPrefix, ItemLoader.RangedPrefix, ItemLoader.MagicPrefix methods would allow weapons to receive prefixes other than the ones their damage type normally allows, such as allowing a modded whip to receive the same prefixes as vanilla whips

### What does this proposal attempt to solve or improve?

Currently making a whip receive the same prefixes as vanilla whips without remaking Item.Prefix requires code such as the following:

```cs
public override int ChoosePrefix(UnifiedRandom rand) {
	if (Item.noUseGraphic) {
		Item.DamageType = DamageClass.Melee;
		Item.noUseGraphic = false;
		Item.Prefix(-2);
		Item.DamageType = DamageClass.SummonMeleeSpeed;
		Item.noUseGraphic = true;
		return Item.prefix;
	}
	return -1;
}
```

### Which (other) solutions should be considered?

None that I can think of.

Chicken-Bones:
Completed

PR #2773, opened 2022-07-29
Add ModItem.MeleePrefix, ModItem.WeaponPrefix...

Tyfyter:
Adds virtual ModItem.MeleePrefix, ModItem.WeaponPrefix,
ModItem.RangedPrefix, and ModItem.MagicPrefix methods to replace the hardcoded conditions in their respective ItemLoader methods (#2772)

### What is the new feature?
The hardcoded conditions in ItemLoader.MeleePrefix, ItemLoader.WeaponPrefix,
ItemLoader.RangedPrefix, and ItemLoader.MagicPrefix have been moved to virtual methods in ModItem.

Currently the prefixes weapons can get are solely determined by damage type (and Item.noUseGraphic for melee weapons) requiring unintuitive workarounds for weapons that benefit from scale, velocity, or mana cost reduction but aren't the vanilla damage class which normally receives prefixes with such effects

The hooks could return null by default and the ItemLoader methods could retain the hardcoded defaults when null is returned;

The following code could be put in a sword to make it receive ranged prefixes instead of melee ones
```cs
public override bool MeleePrefix() {
	return false;
}
public override bool RangedPrefix() {
	return true;
}
```

### DIFF
--- a/ExampleMod/Content/Items/Weapons/ExampleWhip.cs
+++ b/ExampleMod/Content/Items/Weapons/ExampleWhip.cs
@@ -30,5 +30,10 @@ namespace ExampleMod.Content.Items.Weapons
 				.AddTile<Tiles.Furniture.ExampleWorkbench>()
 				.Register();
 		}
+		
+		// Makes the whip receive melee prefixes
+		public override bool MeleePrefix() {
+			return true;
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


direwolf420:
You mentioned in the linked issue that this should be used for whips. If that is the case, shouldn't this be added to ExampleWhip too?

Tyfyter:
It has now been added to ExampleWhip and ExampleWhipAdvanced

Tyfyter:
And that commit which I forgot to name just clarifies how they work together

Chicken-Bones:
Your implementation is good, though I don't like how 'prefix categories' can't be expanded to modded prefix groups. I'll look at the 1.4.4 change and see how this will mesh with them

Tyfyter:
The limitation of the categories is less than ideal, but any way to make it less limited would require the Prefix method to be reworked to actually support custom categories instead of having the categories be hardcodedby item type, and that's far too complicated for me to do until I can get my IDE to work more consistently.

PR #2781, opened 2022-07-30
Faster ReLogic.Content.AssetPathHelper.CleanPath

Cyrillya:
This is an optimization of `ReLogic.Content.AssetPathHelper.CleanPath`
It has been tested that using `StringComparison.Ordinal` as `comparisonType` (default is `StringComparison.CurrentCulture`) will be much faster than default if the enabled language is not English, or the path include characters other than English characters and numbers

Here's a Microsoft document: https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings
> Use comparisons with [StringComparison.Ordinal](https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison#system-stringcomparison-ordinal) or [StringComparison.OrdinalIgnoreCase](https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison#system-stringcomparison-ordinalignorecase) for better performance.

`CleanPath` is intended to give a legal path to use for loading assets and other things, so using `StringComparison.Ordinal` is probably the best choice for improving performance

### Using a non-English language
I called `CleanPath` 100000 times, using different `comparisonType` for string operating
Here is the output of my test

D59DEE`8U`{`N78{HCZE_K4

Game language: Chinese (`zh-Hans`) (The result have no difference between all 8 non-English languages Terraria supports)
Path: `Content/Items/ExampleSoul`

Apparently, `StringComparison.Ordinal` is the fastest `comparisonType`, almost 100 times faster than the default `StringComparison.CurrentCulture`.

### Using English
If the game uses English as the language. Using `StringComparison.Ordinal` is only 5 times faster for English-only paths
)}9}Q_YM 4EU3KX $T92OA8
However, if the path includes non-English characters (Russian here), it will be 100 times faster as well
image
(Of course I know nothing about Russian, I just copied a random text)

## Breakage
It has been tested that by using paths containing specific characters from all 9 languages Terraria supports (like `Content/Items/боеприпасы` and `Content/Items/Käfer`), the path returned with `StringComparison.Ordinal` is the same as the path returned with `StringComparison.CurrentCulture`

So it's certain that using `StringComparison.Ordinal` will not break paths in most languages, and it will improve performance a lot. (Also, why should paths include characters other than English characters and Arabic numerals? Using paths consisting of `a-z`, `A-Z` and `0-9` can avoid many problems)

### DIFF
--- a/patches/tModLoader/ReLogic/Content/AssetPathHelper.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/AssetPathHelper.cs.patch
@@ -1,0 +1,55 @@
+--- src/TerrariaNetCore/ReLogic/Content/AssetPathHelper.cs
++++ src/tModLoader/ReLogic/Content/AssetPathHelper.cs
+@@ -1,4 +_,5 @@
+ using System;
++using System.Diagnostics;
+ using System.IO;
+ 
+ namespace ReLogic.Content
+@@ -6,24 +_,24 @@
+ 	public static class AssetPathHelper
+ 	{
+ 		public static string CleanPath(string path) {
+-			path = path.Replace('/', '\\');
+-			path = path.Replace("\\.\\", "\\");
+-			while (path.StartsWith(".\\")) {
++			path = path.Replace("/", "\\", StringComparison.Ordinal);
++			path = path.Replace("\\.\\", "\\", StringComparison.Ordinal);
++			while (path.StartsWith(".\\", StringComparison.Ordinal)) {
+ 				path = path.Substring(".\\".Length);
+ 			}
+ 
+-			while (path.EndsWith("\\.")) {
++			while (path.EndsWith("\\.", StringComparison.Ordinal)) {
+ 				path = ((path.Length <= "\\.".Length) ? "\\" : path.Substring(0, path.Length - "\\.".Length));
+ 			}
+ 
+ 			int num;
+ 			for (num = 1; num < path.Length; num = CollapseParentDirectory(ref path, num, "\\..\\".Length)) {
+-				num = path.IndexOf("\\..\\", num);
++				num = path.IndexOf("\\..\\", num, StringComparison.Ordinal);
+ 				if (num < 0)
+ 					break;
+ 			}
+ 
+-			if (path.EndsWith("\\..")) {
++			if (path.EndsWith("\\..", StringComparison.Ordinal)) {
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
This is great, but one question. How much time toes path cleaning take in a typical game/mod load? Do you have some profiling results for me to show it's worth changing?

Cyrillya:
> This is great, but one question. How much time toes path cleaning take in a typical game/mod load? Do you have some profiling results for me to show it's worth changing?

Just did a test
```
(Mode: CurrentCulture)
Load Total Time: 22.8452688s
Clean Path Time: 0.1757388s
Percentage in loading: 0.7692568712520468%
AssetRepository.Request Time: 1.4061123s
Percentage in Request<T>: 12.498205157582364%
```
`CleanPath` take only `0.76%` in total loading time and `12%` in `AssetRepository.Request<T>` (`CleanPath` is mainly used in `Request<T>`, so I use it as a comparison.)

It doen't seem quite much. But what if I call `ModContent.Request<T>` while game running? (For example, I want to get the texture directly using `Request<T>` when drawing something)
```
(Mode: CurrentCulture)
Request Total Time: 336Ticks
Clean Path Time: 320Ticks
Percentage in Request<T>: 95.23809523809523%
```
([`Tick`](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.ticks?view=net-6.0) is a [`TimeSpan`](https://docs.microsoft.com/en-us/dotnet/api/system.timespan?view=net-6.0) property which means one ten-millionth of a second)
Since the asset has been loaded, `Request<T>` directly get the value in `AssetRepository._assets` using the key, path. Thus, `CleanPath` took up the major part of the time

The above is the case of using `CurrentCulture` mode. If you change it to `Ordinal` mode, most of the time will be saved. Here is the result using `Ordinal` mode
```
(Mode: Ordinal)
Request Total Time: 54Ticks
Clean Path Time: 29Ticks
Percentage in Request<T>: 53.70370370370371%
```

So what I want to notice is, this optimization mainly benefits the in-game asset requests.

Chicken-Bones:
Investigating this further, I can't see any reason for `.` or `..` to be in asset path names, nor leading or trailing `/`.

I think the only thing this method needs to do is normalise `/` vs `\` (and I'd just make them all use `/`, not even bother matching the environment path separator)

Issue #2779, opened 2022-07-30
[Feature Proposal] Download Mod Dependencies on Start-up


steviegt6:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder / Gameplay as a Player

### Description

tModLoader should download mod dependencies on start-up.

### What does this proposal attempt to solve or improve?

When downloading mods from the Steam Workshop, dependencies are not automatically downloaded (the in-game Mod Browser *does* automatically download dependencies). Additionally, if a mod updates or removes dependencies, these will not be updated.

### Which (other) solutions should be considered?

Needs discussion.

PR #2783, opened 2022-07-30
Download all missing dependencies on start-up.

steviegt6:
### What is the new feature?

Implements #2779.

A prompt for downloading dependencies is present as well (merged into the new/updated mods UI):
image
image
^ When no dependencies are available, it just displays the normal UI with an "OK" button.

See reasoning in #2779.

Needs discussion.


patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -126,6 +126,36 @@ private enum SearchFolders { }
 			return true;
 		}
 
+		internal static IEnumerable<ulong> DownloadWorkshopDependencies() {
```
Solxanich: Can we change the method name to be 'IdentifyWorkshopDepencies'? This method does no downloading

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -126,6 +126,36 @@ private enum SearchFolders { }
...
+			HashSet<ulong> dependencies = new HashSet<ulong>();
+
+			foreach (LocalMod mod in FindWorkshopMods()) {
+				// This shouldn't really ever fail, but better safe than sorry.
+				if (!TryReadManifest(GetParentDir(mod.modFile.path), out var manifest))
```
Solxanich, line 138: We should consider skip dependencies checks if localMod.modreferences contains no dependencies. 

One less query to make, thus better performance.
Not required, but would be nice

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -153,12 +155,33 @@ internal static class Interface
...
 
-					string info = ModOrganizer.DetectModChangesForInfoMessage();
-					if (info != null)
-						infoMessage.Show(Language.GetTextValue("tModLoader.ShowNewUpdatedModsInfoMessage") + info, Main.menuMode);
+					if (!skip) {
+						string newDownloads = showModChanges ? ModOrganizer.DetectModChangesForInfoMessage() : null;
+						string dependencies = promptDepDownloads ? ModOrganizer.ListDependenciesToDownload(deps) : null;
+
+						if (promptDepDownloads)
+							ModLoader.DependenciesToDownload = deps;
```
Solxanich: I'm not finding any reason why this is stored on ModLoader, nor needs to be. 
Remove?

Solxanich (commented):
Some code review. 
A couple of minor cleanups

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
steviegt6: Yeah, old name from before I merged with the download/update UI.

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
steviegt6: 👍

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
steviegt6: Was worried about local variable capturing, but it should probably be fine in hindsight.

Chicken-Bones (approved):

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -718,7 +719,9 @@ internal class AQueryInstance
 					ReleaseWorkshopQuery();
 				}
 
-				internal SteamUGCDetails_t FastQueryItem(ulong publishedId) {
+				// if dependencyCount > 0 then query, otherwise ignore
+				// default calls will ignore dependencies
+				internal SteamUGCDetails_t FastQueryItem(ulong publishedId, uint dependencyCount = 0) {
```
DarioDaF: Making "get dependencies" a bool and using the pDetails from the same query to get # of deps it way better

PR #2801, opened 2022-08-04
Delay steam init until needed (eg for mod browser) on dedicated servers

Chicken-Bones:
### What is the new feature?
When launching a server without steam, avoid loading the native `steam_api64` library until it is actually needed (eg, when requesting a mod from the mod browser).

Also includes a bunch of code cleanups and refactors I came across.

Makes dedicated servers more reliable.

The steam native libraries can't be loaded on every platform due to dependency resolution conflicts. In the example below, steam needs a version of glibc which isn't installed. 
```
Unable to load shared library '/home/tml/tModLoader/Libraries/Native/Linux/libsteam_api64.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: Error relocating /home/tml/tModLoader/Libraries/Native/Linux/libsteam_api64.so: __vsnprintf_chk: symbol not found (see server.log for full trace)
Server crash: 8/4/2022 3:24:05 AM
System.DllNotFoundException: Unable to load shared library '/home/tml/tModLoader/Libraries/Native/Linux/libsteam_api64.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: Error relocating /home/tml/tModLoader/Libraries/Native/Linux/libsteam_api64.so: __vsnprintf_chk: symbol not found
   at System.Runtime.InteropServices.NativeLibrary.LoadFromPath(String libraryName, Boolean throwOnError)
   at System.Runtime.InteropServices.NativeLibrary.Load(String libraryPath)
   at MonoLaunch.ResolveNativeLibrary(Assembly assembly, String name) in tModLoader\Terraria\MonoLaunch.cs:line 96
   at System.Runtime.Loader.AssemblyLoadContext.GetResolvedUnmanagedDll(Assembly assembly, String unmanagedDllName)
   at System.Runtime.Loader.AssemblyLoadContext.ResolveUnmanagedDllUsingEvent(String unmanagedDllName, Assembly assembly, IntPtr gchManagedAssemblyLoadContext)
   at Steamworks.NativeMethods.SteamInternal_GameServer_Init(UInt32 unIP, UInt16 usPort, UInt16 usGamePort, UInt16 usQueryPort, EServerMode eServerMode, UTF8StringHandle pchVersionString)
   at Steamworks.GameServer.Init(UInt32 unIP, UInt16 usGamePort, UInt16 usQueryPort, EServerMode eServerMode, String pchVersionString)
   at Terraria.Social.Steam.WorkshopHelper.ModManager.Initialize() in tModLoader\Terraria\Social\Steam\WorkshopHelper.TML.cs:line 111
   at Terraria.Social.SocialAPI.Initialize(Nullable`1 mode) in tModLoader\Terraria\Social\SocialAPI.cs:line 64
   at Terraria.Program.LaunchGame_(Boolean isServer) in tModLoader\Terraria\Program.cs:line 230
```

Could init the game server on entering the server mod browser menu, rather than on downloading an item, but it's a couple seconds on what's already a 15 second process anyway.

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -168,7 +168,7 @@ namespace Terraria.Social.Steam
 			/// <summary>
 			/// Downloads all UIModDownloadItems provided.
 			/// </summary>
-			public static void Download(List<ModDownloadItem> items) {
+			public static Task Download(List<ModDownloadItem> items) {
 				//Set UIWorkshopDownload
 				UIWorkshopDownload uiProgress = null;
 
@@ -192,7 +192,7 @@ namespace Terraria.Social.Steam
 
 				int counter = 0;
 
-				Task.Run(() => TaskDownload(counter, uiProgress, items));
+				return Task.Run(() => TaskDownload(counter, uiProgress, items));
 			}
 
 			internal static void DownloadBatch(string[] workshopIds, UI.UIState returnMenu) {
@@ -245,11 +245,11 @@ namespace Terraria.Social.Steam
 						foreach (var localMod in enabledItems) {
 							localMod.Enabled = true;
 						}
-					ModLoader.ModLoader.Reload();
-			}
+						ModLoader.ModLoader.Reload();
+					}
 				}
 				else
-					Task.Run(() => TaskDownload(counter, uiProgress, items, ids));
+					TaskDownload(counter, uiProgress, items, ids);
 			}
 
 			private EResult downloadResult;
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -109,36 +109,46 @@ namespace Terraria.Social.Steam
 		{
 			internal const uint thisApp = ModLoader.Engine.Steam.TMLAppID;
 
-			internal static bool SteamUser { get; set; } = false;
-			internal static bool SteamAvailable { get; set; } = true;
+			internal static bool SteamUser { get; set; }
+			internal static bool SteamAvailable { get; set; }
 
 			protected Callback<DownloadItemResult_t> m_DownloadItemResult;
 
 			private PublishedFileId_t itemID;
+			private string name;
 
 			internal static void Initialize() {
-				string apptxt = "steam_appid.txt";
-
-				// Non-steam tModLoader will use the SteamGameServer to perform Browsing & Downloading
-				if (SocialMode.Steam == SocialAPI.Mode) {
+				if (SocialAPI.Mode == SocialMode.Steam) {
+					SteamAvailable = true;
 					SteamUser = true;
-					if (File.Exists(apptxt))
-						File.Delete(apptxt);
-
 					return;
 				}
 
-				File.WriteAllText(apptxt, thisApp.ToString());
+				// On some systems without steam, the native dependencies required for steam fail to load (eg docker without requisite glibc)
+				// Thus, for dedicated servers we delay game-server init until someone tries to use steam features (eg mod browser)
 
-				if (GameServer.Init(0x7f000001, 7775, 7774, EServerMode.eServerModeNoAuthentication, "0.11.9.0")) {
+				// Non-steam tModLoader will use the SteamGameServer to perform Browsing & Downloading
+				if (!Main.dedServ && !TryInitViaGameServer())
+					Logging.tML.Error("Steam Game Server failed to Init. Steam Workshop downloading on GoG is unavailable. Make sure Steam is installed");
+			}
+
+			public static bool TryInitViaGameServer() {
+				ModLoader.Engine.Steam.SetAppId(ModLoader.Engine.Steam.TMLAppID_t);
+				try {
+					if (!GameServer.Init(0x7f000001, 7775, 7774, EServerMode.eServerModeNoAuthentication, "0.11.9.0"))
+						return false;
+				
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -109,36 +109,46 @@ internal class ModManager
...
+				// Thus, for dedicated servers we delay game-server init until someone tries to use steam features (eg mod browser)
+
+				// Non-steam tModLoader will use the SteamGameServer to perform Browsing & Downloading
+				if (!Main.dedServ && !TryInitViaGameServer())
+					Logging.tML.Error("Steam Game Server failed to Init. Steam Workshop downloading on GoG is unavailable. Make sure Steam is installed");
+			}
 
-				if (GameServer.Init(0x7f000001, 7775, 7774, EServerMode.eServerModeNoAuthentication, "0.11.9.0")) {
+			public static bool TryInitViaGameServer() {
+				ModLoader.Engine.Steam.SetAppId(ModLoader.Engine.Steam.TMLAppID_t);
```
Solxanich, line 135: The statics in this use steamworks prior to init which is not recommended. May fail for GoG. 
There was a reason the code was in the particular order it was- I can't recall the exact bug anymore, but I believe it lead to it trying to use SteamApi.init and messing it all up. Needs thorough testing for GoG

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -109,36 +109,46 @@ internal class ModManager
...
 					return;
 				}
 
-				File.WriteAllText(apptxt, thisApp.ToString());
+				// On some systems without steam, the native dependencies required for steam fail to load (eg docker without requisite glibc)
+				// Thus, for dedicated servers we delay game-server init until someone tries to use steam features (eg mod browser)
+
+				// Non-steam tModLoader will use the SteamGameServer to perform Browsing & Downloading
+				if (!Main.dedServ && !TryInitViaGameServer())
+					Logging.tML.Error("Steam Game Server failed to Init. Steam Workshop downloading on GoG is unavailable. Make sure Steam is installed");
```
Solxanich, line 131: I'd swear there was a better lang key available from the mod browser that should replace everything after 'failed to init'?

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -150,29 +160,21 @@ internal class ModManager
...
-				foreach (var item in items) {
-					if (item.Installed != null && item.Installed.Enabled) {
-						enabledItems.Add(item.Installed);
-						item.Installed.Enabled = false;
-					}
-				}
-
-				if (enabledItems.Count > 0) {
+				var needsReload = items.Any(item => item.Installed != null && item.Installed.Enabled);
+				if (needsReload) {
```
Solxanich: This code doesn't actually unload the mods.
How do you know which ones to reload later at that?

Player expectation would be updating a mod shouldn't also leave it disabled after- that is a hassle

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -181,66 +183,19 @@ internal class ModManager
 					Main.MenuUI.SetState(uiProgress);
 				}
 
-				int counter = 0;
-
-				Task.Run(() => TaskDownload(counter, uiProgress, items));
-			}
-
-			internal static void DownloadBatch(string[] workshopIds, UI.UIState returnMenu) {
```
Solxanich, line 189: This is actually needed and will be used. Do not remove

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -181,66 +183,19 @@ internal class ModManager
...
-					var id = ids[counter++];
-					mod = new ModManager(new PublishedFileId_t(ulong.Parse(id)));
-
-					hasUpdate = mod.NeedsUpdate() || !mod.IsInstalled();
-					name = id;
-					endCount = ids.Length;
+			private static void TaskDownload(UIWorkshopDownload uiProgress, List<ModDownloadItem> items, bool needsReload) {
+				foreach (var item in items) {
+					var mod = new ModManager(new PublishedFileId_t(ulong.Parse(item.PublishId)), item.DisplayName);
+					mod.InnerDownload(uiProgress, item.HasUpdate);
```
Solxanich: Each item has to be downloaded serially if we are to display UI that it downloaded. 
This will lead to a bunch of conflicting UIs when more than one (any mod with dependencies)

Solxanich (changes requested):
Far too many changes with problems. 

Should not be touching code outside of the target scope, significant extra effort for review.

The following is likely broken with proposed fluff:
- downloading mods with dependencies
- browsing mods on GoG
- updating mods that are enabled

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
Chicken-Bones: It does. I tested. `ModLoader.Unload` unloads all the mods. You don't need to disable to unload. When they're reloaded, the enabled set stays the same and they all come back.

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
Chicken-Bones: Downloads are still performed serially. There's no parallelisation here. This is a straight refactor of repeatedly re-queueing a task with the next index in an array, to a for loop, on the same thread.

PR #2817, opened 2022-08-10
Improve dedicated server management utils.

ppebb:
### What is the new feature?
Improved server management utils. A management script and a docker container. The script can manage github release or steamcmd installations along with workshop mods. 

The existing server utils are almost non-existent. These will make it much easier to setup a dedicated server, and now we won't have to redirect people to use an unofficial docker container.

No.

A dedicated server can be setup in a single command. `./manage-tModLoaderServer.sh -i -g` or `docker-compose build && docker-compose up -d`


guillheu:
Regarding your dockerfile I have a few remarks : 
- I don't understand why you *both* use the `steamcmd` image and *also* download the steamcmd binary. You seem to be copying both of them into your final image, but I don't understand the purpose of having 2 potentially different versions.
- In my experience, alpine images are most relevant when you don't have to update them or install any extra packages. Would be worth looking into the images size more deeply, because it's possible something like `debian:stable-slim`, if it already packs all the utilities you need, will likely be smaller than alpine with the extra packages.
- you apparently forgot to remove the temp steps on lines 72-76 and 84-86

Going to try out your scripts.

ppebb:
> You seem to be copying both of them into your final image, but I don't understand the purpose of having 2 potentially different versions.

I got that portion of the dockerfile from steamcmd's [alpine image](https://github.com/steamcmd/docker/blob/master/dockerfiles/alpine-3/Dockerfile). There aren't two versions of steamcmd installed. I believe the steamcmd:ubuntu container is just a verified working base image of ubuntu, not that the steamcmd binary and .sh script are preinstalled. If you're referring to copying both steamcmd.sh and the steamcmd binary, then they're both required as the script is just a wrapper. Regardless, all of the steps seem to be required for everything to function, although if you could figure out how to remove the ubuntu step entirely that would be quite welcome.

As for container size, I'll look into alternatives to alpine when I have time.

> you apparently forgot to remove the temp steps on lines 72-76 and 84-86

This was explained in the note at the top of the PR. The temp steps are required for the container to function until the next stable release cycle. The temp steps will be removed once everything is tested and everything is finalized.

guillheu:
Another recommendation, I think for this use case you might be better off using a ubuntu image instead of alpine.
Yes it's bigger, but it's also much more user friendly.

Alpine images are most impactful in production environments where you might have dozens of instances of the same stateless container. There, every megabyte makes a noticeable difference, and transates into lower operation costs.
But if this is meant as a user-facing container, having a more intuitive base image is, in my opinion, worth sacrificing a little bit of bandwidth and system memory. It'll allow the users to more easily tinker and explore the container, including at runtime, and makes troubleshooting easier for the dev team and users.

ppebb:
There isn't really anything less intuitive about alpine. Pretty much everything is identical to any other image from a user standpoint and shouldn't, except maybe that it uses musl (this wouldn't affect anything tML related though). There's not really anything specific to alpine that would inhibit someone from tinkering that comes to mind. Although if you could find some major issue caused by alpine, I'd be more than willing to change the base image.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```diff
@@ -0,0 +1,396 @@
...
+}
+
+function update_script {
+	local latest_script_version
+	latest_script_version=$(curl --silent "https://raw.githubusercontent.com/pollen00/tModLoader/serversetup/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh" | grep "script_version=" | cut -d '"' -f2)
+
+	if ! verlte $script_version "$latest_script_version"
+	then
+		echo "Updating from version v$script_version to v$latest_script_version"
+		curl --silent -O https://raw.githubusercontent.com/pollen00/tModLoader/serversetup/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```
Solxanich: update for final script

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```diff
@@ -0,0 +1,396 @@
...
+	fi
+	echo "Downloading version $1"
+	curl --silent -LJO "https://github.com/tModLoader/tModLoader/releases/download/$1/tModLoader.zip"
+}
+
+# Check $username is defined, exit if not
+function check_username {
+	if ! [[ -v username ]]
+	then
+		echo "Please enter a username to login with using --username"
```
Solxanich: Please clarify 'user name'. User name for what?
Why?
Unclear when running

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```diff
@@ -0,0 +1,396 @@
...
+		# Install tml from github, leave some file containing what version
+		local ver
+		ver=$(get_latest_release)
+		down_release "$ver"
+		echo "Unzipping tModLoader.zip"
+		unzip -q tModLoader.zip
+		rm tModLoader.zip
+		echo "$ver" > .ver
+
+		popd
```
Solxanich: Please put thus in to a function so can easily test it by itself

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```diff
@@ -0,0 +1,396 @@
...
+				echo "Removing old backups"
+				for file in ./v*.tar.gz;
+				do
+					if ! [[ "$file" == ./$oldver.tar.gz ]]
+					then
+						rm "$file"
+						echo "Removed $file"
+					fi
+				done
+			fi
```
Solxanich: split out functions, please

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
```diff
@@ -0,0 +1,396 @@
...
+ --tml-version       The version of tML to download. Only applies when using Github. This should be the exact tag off of Github (ex. v2022.06.96.4).
+ --update-script     Update the script to the latest version on Github.
+ --mods-only         Only install/update mods.
+ --no-mods           Don't install/update mods.
+ --check-dir         Directory to check for enabled.json, install.txt, and any .tmod files.
+
+When running --install and --update, enabled.json, install.txt, and any .tmod files will be checked for in the location of the script or in the directory specified by --check-dir."
+	exit
+}
+
```
Solxanich: Can I get a comment block indicating this is entry point of main script? 
Easy to get lost with functions listed first

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
```diff
@@ -1,25 +1,60 @@
...
+			launch_args="$launch_args -lobby friends"
+		else
+			launch_args="$launch_args -lobby private"
+		fi
 	fi
 fi
 
 chmod +x ./LaunchUtils/ScriptCaller.sh
-./LaunchUtils/ScriptCaller.sh $launch_args
+./LaunchUtils/ScriptCaller.sh "$launch_args"
```
Solxanich: How much has this been tested? have we supplied other args to verify it works fine still?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/Docker/Dockerfile
```diff
@@ -0,0 +1,90 @@
...
+
+# Download entrypoint script
+#NOTE: Update url before merging
+RUN curl -O https://raw.githubusercontent.com/pollen00/tModLoader/serversetup/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/Docker/Launch.sh \
+ && chmod u+x Launch.sh
+
+#NOTE: Temp step to download modified launch script. Remove before merging.
+RUN curl -O https://raw.githubusercontent.com/pollen00/tModLoader/serversetup/patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh \
+ && chmod u+x start-tModLoaderServer.sh \
+ && mv start-tModLoaderServer.sh $HOME/tModLoader/start-tModLoaderServer.sh
```
Solxanich: Update before merging

Solxanich (commented):
Some code cleanups that I could see.
Haven't had a chance to test - getting time to test is holding me back from merging.

Might need to get a linux user to help test instead.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/Docker/Dockerfile
ppebb: Once everything else is resolved I'll update this since it's needed for testing.

ppebb:
> Some code cleanups that I could see. Haven't had a chance to test - getting time to test is holding me back from merging.
> 
> Might need to get a linux user to help test instead.

Could also just use a live usb image in a vm. Would take maybe 10 minutes to setup.

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
ppebb: Accidental. Removed.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/ManageServer.sh
ppebb: Done.

Solxanich (approved):
Partial Approval. Happy with resolved feedback.
Spot Tested steamcmd install mod & github update tml functions.

Seems like it is going to work based on spot testing and code reads.

@Chicken-Bones - can you please take a look for any final thoughts?

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
```diff
@@ -1,23 +1,58 @@
...
+		*) # If it's not one of the args above, then pass it to ScriptCaller.sh
+			launch_args="$launch_args $1"
+			shift
+			;;
+	esac
+done
+
+# Use serverconfig.txt as config if not already specified
+if ! [[ "$launch_args" == *"-config"* ]]
+then
```
Chicken-Bones, line 26: Code style, can we put `then` on the same line as the `if`? Is there a reason this is inadvisable in bash?

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
```diff
@@ -1,23 +1,58 @@
...
+		-steam)
+			steam_server=true
+			shift
+			;;
+		-nosteam)
+			steam_server=false
+			shift
+			;;
+		*) # If it's not one of the args above, then pass it to ScriptCaller.sh
+			launch_args="$launch_args $1"
```
Chicken-Bones, line 18: question, do quoted args remain quoted?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `ManageServer.sh` script can be used to install tML either directly from the Github release or from SteamCMD. The script is made to run fully standalone, so just download it to your server and run it. No other files from the repo are needed.
+
+Github installations can be done using `./ManageServer.sh --install --github`. SteamCMD is the default so just leave out `--github` to use it instead. By default Github installations will be in `~/tModLoader`, and SteamCMD installations will be in `~/Steam/steamapps/common/tModLoader`.
+
+During installation, [enabled.json](#install.txt-and-enabled.json), [install.txt](#install.txt-and-enabled.json), and any `.tmod` files will be checked for in the directory of the script or in the directory specified by `--checkdir`.
+
+Note that for Github installations `unzip` must be on your `PATH`, and for SteamCMD installations `SteamCMD` must be on your `PATH`. Install it from your package manager or from [steam](https://developer.valvesoftware.com/wiki/SteamCMD).
+
+For additional information, run `./ManageServer.sh --help` to view the available arguments.
```
Chicken-Bones: This doesn't read well to a novice. Needs to be very simple 'quick start' style. A series of steps.
1. Installation
  Via steamcmd... (recommended)
  From github...
2. Installing mods
  Via modpack
  Individually
3. Launching

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Using the Docker container
+
+Before building and running the container first you need to prepare a directory to pass to the container. This applies to `docker` and `docker-compose` This directory will link to `~/.local/share/Terraria` inside of the container. Within the directory, place your [enabled.json](#install.txt-and-enabled.json), [install.txt](#install.txt-and-enabled.json), `serverconfig.txt`, and any `.tmod` files. This directory will be used in the `docker run` step. To manually add worlds, place them in `tModLoader/Worlds` inside of your directory (you may have to create the directory yourself).
+
```
Chicken-Bones: This needs an intro for people unfamiliar with docker. Should they be using docker? Why is it not an option as part of `ManageServer.sh`? Is it less fragile? Installing steam_cmd is hard.

There needs to be a section at the very top to help people figure out which method they're meant to use, why pick steamcmd vs docker vs github.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Docker Compose
+ * Copy `docker-compose.yml` and `Dockerfile` to the same directory
+ * `docker-compose build`
+ * `docker-compose up`
+
+To allow the container to persist after closing your terminal or ssh session, run with `docker-compose up -d`.
+
+
+### serverconfig.txt
+Because the container is meant to be headless, you can't select a world or any other options using the normal interactive method. They must be specified in `serverconfig.txt`. There are only three things required in your config, but you'll likely want to add more. Additional options can be found [on the Terraria wiki](https://terraria.wiki.gg/wiki/Server#Server_config_file)
```
Chicken-Bones: Please make the docker container work in interactive mode (`-i`)

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+# The world file to load. If you already have a world, set this to the filename. Only change the last segment of the path.
+world=/home/tml/.local/share/Terraria/tModLoader/Worlds/YourWorld.wld
+# Allow the world to be created if none are found. World size is specified by 1 (small), 2 (medium), and 3 (large).
+autocreate=1
+# Sets the name of the world when using autocreate
+worldname=YourWorld
+```
```
Chicken-Bones: We should ship a template serverconfig.txt instead.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Updating
+To update the script, run `./ManageServer.sh --updatescript`
```
Chicken-Bones: When? Why? How do I know it's out of date? What happens if I don't update?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+When using `ManageServer.sh`, updating can be done by running `./ManageServer.sh --update`. When using a Github install, use `--github` and `--folder` if your install is in a non-standard location.
```
Chicken-Bones: Install/Update should pick the latest stable. Please add a `--preview` option! (Not a blocker for approving this PR)

Chicken-Bones (commented):
Please call `ManageServer.sh` something like `manage-tModLoaderServer.sh` in keeping with our other script naming.

I didn't review `ManageServer.sh` in detail, it looks great and useful. 

My comments are mostly about the README

I'll get a friend to glance over docker, since I'm illiterate there.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
ppebb: Currently it is getting the latest stable (latest release from github releases page). Preview cannot be added because the build artifacts require a github account, unless there is some way around that.

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
ppebb: No reason not to, I just thought it looked nicer this way. Easy change though.

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
pollen00: Seems that they don't.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
ppebb: Will do once tML can handle SIGTERM and SIGINT itself.

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
ppebb: Explained it better and added an update check so the user will be informed if one is available.

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
ppebb: should be fine.

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
```diff
@@ -1,23 +1,53 @@
 #!/usr/bin/env bash
-cd "$(dirname "$0")"
+cd "$(dirname "$0")" || exit
```
Chicken-Bones, line 2: why this?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,65 @@
...
+## Docker vs Management Script
+Both the Docker container and the management script can install and update tModLoader and any mods. Docker is a service used to containerize, or isolate, a specific application from a host system. This is beneficial for both security and ease of use with more complicated services. If you would prefer more control over your server, then make use of `manage-tModLoaderServer.sh`.
+
+## Using The Script
+The `manage-tModLoaderServer.sh` script can be used to install tModLoader either directly from the GitHub release or from SteamCMD. The script is made to run fully standalone, so just download it to your server and run it. No other files from the repo are needed.
+
+### Installing tModLoader
+#### Via SteamCMD (recommended)
+* Ensure SteamCMD is installed and on your PATH. You can install SteamCMD from your package manager or [Valve's Wiki](https://developer.valvesoftware.com/wiki/SteamCMD).
+* Installation via SteamCMD can be performed using `./manage-tModLoaderServer.sh --install --username your_steam_username`.
```
Chicken-Bones: Replace "Installation via SteamCMD can be performed using" with just "Run"

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,65 @@
...
+* You will be prompted for your password (and your 2fa code if applicable).
+* By default, tModLoader will be installed to `~/Steam/steamapps/common/tModLoader`. To specify an installation directory, use `--folder /full/path/from/root`
+
+#### From GitHub
+* Installation from the latest GitHub release can be performed using `./manage-tModLoaderServer.sh --install --github`.
```
Chicken-Bones: As above, just "Run" is fine. A bullet point below can say "The script will automatically use the latest Github release, which will be the same as the version released on Steam."

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,65 @@
...
+* By default, tModLoader will be installed to `~/tModLoader`. To specify an installation directory, use `--folder /path/to/install`.
+
+Additionally, you can avoid updating or installing mods with the `--no-mods` argument.
```
Chicken-Bones: Is this a github only option? Should this go in the next section?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,65 @@
...
+#### Obtaining install.txt
+Because the steam workshop does not use mod names to identify mods, you must create a modpack to install mods from the workshop. To get an `install.txt` file and its accompanying `enabled.json`:
+* Go to Workshop
+* Go to Mod Packs
+* Click `Save Enabled as New Mod Pack`
+* Click `Open Mod Pack Folder`.
+* Enter the folder with the name of your modpack
+
+You can copy `enabled.json` and `install.txt` to your script directory and they will be used next time the script is run.
+
```
Chicken-Bones: What command do I need to run to install mods? Maybe remind people to run `manage-tModLoaderServer.sh --mods-only` after adding their mods/install.txt to the folder?

patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/README.md
```diff
@@ -0,0 +1,65 @@
...
+### Launching
+To run tModLoader, you just need to navigate to your install directory (`~/tModLoader` for GitHub, `~/Steam/steamapps/common/tModLoader` for SteamCMD, by default), and run `./start-tModLoaderServer.sh`.
```
Chicken-Bones: why not add a `-start` parameter to the manage script? Since that script knows where it gets installed to anyway

Chicken-Bones (changes requested):
Readme comments

patches/tModLoader/Terraria/release_extras/start-tModLoaderServer.sh
ppebb: If for some reason cd fails, it will exit rather than running the script in the wrong directory. Likely not needed, but it's safer.

Issue #1785, opened 2021-08-24
[1.4 MergedTesting] RGB Lighting Integration doesn't work 


Soulfurious:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Other (please specify elsewhere)

### OS

Windows

### Platform

Steam

### Description

Terraria seems unable to take control of various “RGB” enabled devices when TML is used for version 1.4 of the game. 

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/7039340/client.log)

### Steps to reproduce

1. Launch the game using the batch file located in the installations root directory.
2. Observe the multicolored LEDs on your mouse, keyboard, or any other supported peripheral. Observe your gaming software to see if profiles are switched.

### Expected Behavior

Upon launching the game, the peripherals profile should immediately switch to the profile created for the game. This should result in the multicolored LEDs on your peripherals dimming, starting the 'breathing' effect where the lights turn green and reach peak brightness as the sun reaches its zenith over the horizon in the main menu screen.

This behavior being present would obviously indicate that the feature of lighting integration is working as intended, which is not what's happening.

### Actual Behavior

The lighting integration does not work at all, and the gaming software does not switch to the Terraria profile which contains the integration for the original Terraria executable.  

### Reproduction frequency

This problem occurs every time the game is launched with TML 1.4's batch file on the MergedTesting branch.

### Additional Information

This feature appears to work normally in the vanilla game on my end. The option to use RGB is enabled in the JSON files in both installations.

I am using the MergedTesting branch (though results should be the same in the public alpha build, and .NET version 5.0.8 is installed. The vanilla game and the steam version of TML are installed in separate directories. The peripherals in use are the Logitech G213 and G502. G-Hub has been installed, migrating from LGS. I attached the screenshot of what my profile settings and paths look like within G-Hub.

![terraria lgs profile](https://user-images.githubusercontent.com/11460328/130621827-0764bbcd-0ad5-4dfc-a7cc-4e41c2ce1d11.png)

I have some speculation to bring to the table as well. Whilst troubleshooting this issue with Solxan and others, he surmised that something wrong within the startup code that is supposed to create RGB handlers. And that someone potentially disabled the error tossing for the scenario where RGB is unsuccessfully loaded, and that they also disabled it entirely by accident. I do not know whether or not this could be the case.

Armipotent:
**Temporary fix - Tested on `tModLoader v0.12 1.4 Alpha 02024925`**

[...]

PR #2821, opened 2022-08-11
Fix Razer Chroma on Windows x64 partially resolves issue #1785

LittleJay1985:
### What is the bug?
Razer Chroma RGB works in vanilla Terraria but not in tModLoader.

### How did you fix the bug?
Razer Chroma installs both x86 and x64 versions of the SDK (rzChromaSDK.dll + rzChromaSDK64.dll).
As Terraria runs as a x86 process it loads the x86 rzChromaSDK.dll just fine.
Whereas tModLoader will run as a x64 process and try to load a x64 version of rzChromaSDK.dll which doesn't exist.

I have modified the code to load the correct DLL - rzChromaSDK64.dll

I have tried to change as little as possible.

Alternatively I could add some functions to check the process architecture and load either the x86 or x64 dll. But as the install base of x86 Windows is miniscule I personally don't see the need for it.


Chicken-Bones:
Thankyou for your contribution! This looks good, but it should be done at the TerrariaNetCore patch level.
The easiest way to do this is to:
- Open TerrariaNetCore.sln
- Copy the files you changed from src/tModLoader/... to src/TerrariaNetCore/...
- Make sure it builds, and test the fix in the TerrariaNetCore.sln
- Click "Diff TerrariaNetCore" in the setup tool.
- Revert your patches/tModLoader changes 
  - Alternatively: click "Options > Reset Timestamp Optimisations" followed by "Diff tModLoader" in the setup tool
- Click "Regenerate Source" in the setup tool and make sure everything works properly

LittleJay1985:
Is that now in the correct place?

I originally wasn't sure ./src/TerrariaNetCore/.. was the correct place as the setup tool states its for fixing decompilation errors, hence applying the fix in ./src/tModLoader/...

So to clarify; is 

- ./src/Terraria/... for decomplication fixes?
- ./src/TerrariaNetCore/.. for .NET Framework to .NET6+ changes and Terraria bug fixes?
- ./src/tModLoader/... for tModLoader's framework additions?

Thanks

LittleJay1985:
Added code comments detailing changes

PR #2828, opened 2022-08-12
Armor Set Definition Revamp

ThomasThePencil:
## OVERVIEW
armor sets, as they're handled by tModLoader, have always been a massive pain. this PR aims to alleviate that trouble, and add plenty of ease of access for commonplace functionality, by revampin' the way that non-vanilla (and possibly vanilla) armor sets are defined, and their effects applied

this is a translation and modification of a system I *already* use and have workin' for Dissonant Duality, and with it, amongst the many smaller victories which this would squeeze out, I aim to make the followin' things easier:
- addin' non-conventional equipment sets and conditions for equipment set effect activation (as, currently, this is frustratin' at best to do)
- modifyin' the effects of existin' equipment sets and individual pieces, as doin' so currently is a massive nightmare due to tML's hooks for equipment effect modifications bein' painfully insufficient (there is no way to cancel the vanilla effects of an equipment item or set bonus by any intuitive means, and thus, no way to intuitively override them)
- generally keepin' armor sets in particular more organized, as havin' to shoehorn the set effects into an arbitrarily-chosen piece of the set runs counter to keepin' things nice and neat in one's codebase

## THIS IS A DRAFT
- [x] add ArmorSet with some entry-level functionality
- [ ] possibly convert vanilla sets to ArmorSet? high-maintenance and equally patch-intensive, more than likely, so ehhhh
- [ ] possibly add GlobalArmorSet? definitely discuss further
- [ ] explore approach vectors for easier individual armor piece/accessory modifications
- [ ] ExampleMod updates


Chicken-Bones:
You've explained the what, but not the why. As usual, please include 2+ concrete use-cases (mod features that are hard to make at the moment) and how this addresses them.

Why is there a whip use-time bug-fix in here?

ThomasThePencil:
**Q:** You've explained the what, but not the why. As usual, please include 2+ concrete use-cases (mod features that are hard to make at the moment) and how this addresses them.
**A:** fixed. the description now gives a better explanation of why tML's equipment hooks are painfully insufficient, although I'll gladly elaborate further if need be

**Q:** Why is there a whip use-time bug-fix in here?
**A:** that's part of the context of the label that's added just under that pre-existin' fix; said fix itself isn't changed at all

steviegt6:
I believe I fundamentally disagree with abstracting armors sets to this degree - especially in such a painfully object-oriented manner - but I would not mind a rework to the current system as it exists in tModLoader.
I understand that this is a draft PR, and everything is subject to change, but I felt that my criticism may aid you in your development.

> armor sets, as they're handled by tModLoader, have always been a massive pain.

Please elaborate. What is a pain and why? I can think of a couple of things, such as set bonus text modifications and more esoteric armor set definitions that don't conform with vanilla's rather basic idea of "three pieces of armor", but that's all relatively easy to work around with very little thought currently.

> [...] amongst the many smaller victories which this would squeeze out [...]

I am interested in hearing this, since no explicit examples have been provided in the PR's description yet.

> addin' non-conventional equipment sets and conditions for equipment set effect activation (as, currently, this is frustratin' at best to do)

What is frustrating in the existing implementation that does not carry over to this system? The only issues I see are that it's rather tiring to define and somewhat boilerplate-y, but this system just amplifies that.

> modifyin' the effects of existin' equipment sets and individual pieces, [...]

ModPlayer

> [...] tML's hooks for equipment effect modifications bein' painfully insufficient [...]

Does tModLoader need specialized hooks for everything? What's the justification?

> (there is no way to cancel the vanilla effects of an equipment item or set bonus by any intuitive means, and thus, no way to intuitively override them)

ModPlayer

> generally keepin' armor sets in particular more organized, as havin' to shoehorn the set effects into an arbitrarily-chosen piece of the set runs counter to keepin' things nice and neat in one's codebase

I do agree with this, it isn't pretty in its current form. I don't think making a specialized type is the right solution, though.

> possibly convert vanilla sets to ArmorSet? high-maintenance and equally patch-intensive, more than likely, so ehhhh

tModLoader has only been growing in terms of heavy-maintenance, bug-prone patches. I worry this will only serve to worsen it.

> possibly add GlobalArmorSet? definitely discuss further

What would this offer that cannot be accomplished with ease in a ModPlayer? How would these interact with other armor sets? Is it just manipulation of effects? If so, that should just be doable in a ModPlayer.

> explore approach vectors for easier individual armor piece/accessory modifications

*What* modifications? I've seen this thrown around several times without much of an explanation.

Chicken-Bones:
@steviegt6 Thankyou for your feedback. I 100% agree.

@ThomasThePencil when I say _concrete_ use cases, I mean:
- An actual piece of content (eg, the spectre armor set)
- A code snippet demonstrating how the content must currently be implemented
- A code snippet demonstrating how the content would be implemented with the implementation of this PR

It's okay to use some pseudo-code or descriptions, but as @steviegt6 has indicated, explicit code examples are the only way to reduce miscommunication.

patches/tModLoader/Terraria/ModLoader/ArmorSet.cs
```diff
@@ -0,0 +1,105 @@
...
+		/// Checks whether or not the player has the given armor set equipped in functional slots.<br/>
+		/// By default, checks for all pieces included in the set; if at least one is missing, the functional effects of this set are not applied.<br/>
+		/// Override this only if you need to make your own active check for this armor set's functional effects.<br/>
+		/// </summary>
+		/// <param name="player">
+		/// The player to check set bonus validity for.<br/>
+		/// </param>
+		/// <returns>
+		/// If the conditions to apply the functional armor set bonus are met, <see langword="true"/>; otherwise, <see langword="false"/>.</returns>
+		public virtual bool ActiveFunctional(Player player) {
```
Chicken-Bones, line 51: Don't put this method here, and don't make it `virtual`.

Instead, if we're going for this, it should be as simple as `ArmorSetLoader.FindMatchingSet(int head, int body, int legs)`

`FindMatchingSet` _can_ loop through all of the armor sets, and compare this, but we could also organise armor-sets into a tree and make the lookup much much faster.

Chicken-Bones (commented):
I actually don't mind this implementation at all.

Chicken-Bones:
`ArmorSetBonus` is an explicit thing in 1.4.5. tML will need a new API then

PR #2829, opened 2022-08-12
TML Dependency management overhaul

Mirsario:
This PR achieves the following:
- Replaces complicated XML 'code' for organizing TML's dependency libraries with C# task invocations, fixing a cursed `RemoveAt(-1)` error.
- The task works differently - instead of defining 'additional probing paths' in the `.runtimeconfig.dev.json` file, it modifies paths in the `.deps.json` one, giving the runtime direct paths to load dependencies from, hopefully minimizing cases for when libraries are loaded from an unwanted path, such as any system library dump folder that can be found on `PATH`.
- Gets rid of `NativeLibrary.SetDllImportResolver` calls and their custom resolvers, which were frankly not even implemented correctly, as they're not supposed to throw exceptions.
- Changes the native library layout to be based on .NET `rid`s, or 'runtime identifiers'. So, instead of `Windows`, we now have `win-x64`, and this will allow us to easily add 32-bit support, under `win-x86`.
- Completely cleans up and organizes all `.csproj` and `.targets` files under `/src`, on all patch stages. They're actually human readable now.


Mirsario:
Superseded #3965, a downsized recreation of this. Another PR will follow that one.

PR #2841, opened 2022-08-17
Add UpdateDamage hook to projectile

Capaldi12:
### What is the new feature?

UpdateDamage hook that allows user to modify how damage is updated for projectile.

At the moment, continuous damage update only takes in account originalDamage of projectile and corresponding damage StatModifier of player. With the new hook, user can make projectile damage also depend on other values, which allows for more elaborate projectile design.

For example, hook can be used to make projectile proportionally stronger when player is at low health:

```cs
// In class derived from ModProjectile
public override void UpdateDamage(ref StatModifier modifier, ref int originalDamage)
{
    Player owner = Main.player[Projectile.owner];
    modifier *= owner.statLifeMax / owner.statLife;
}
```

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalProjectile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalProjectile.cs
@@ -188,6 +188,15 @@ namespace Terraria.ModLoader
 		}
 
 		/// <summary>
+		/// Allows you to modify how damage of this projectile is updated. This hook is only called wehen minion, sentry or ContinuouslyUpdateDamage is set.
+		/// </summary>
+		/// <param name="projectile"></param>
+		/// <param name="modifier">Stat modifier that will be applied to damage</param>
+		/// <param name="originalDamage">Damage to which modifier will be applied</param>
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Is `ProjectileLoader.PreAI` insufficient?

direwolf420:
I confirm that the PreAI hook is sufficient for this, I use this in my own mod and I know another mod that does the same.

(Plus in your current hook design, passing the players damage modifier as ref is dangerous, it would be better if you instead make a copy of it (don't remember offhand what the best approach is, you can always make a new StatModifier instance) and pass that as ref instead, and that copy will be used for damage scaling)

Capaldi12:
> I confirm that the PreAI hook is sufficient for this, I use this in my own mod and I know another mod that does the same.
> 
> (Plus in your current hook design, passing the players damage modifier as ref is dangerous, it would be better if you instead make a copy of it (don't remember offhand what the best approach is, you can always make a new StatModifier instance) and pass that as ref instead, and that copy will be used for damage scaling)

I will take a note of this, thanks for feedback

PR #2847, opened 2022-08-18
Clarify where client.log can be found

LandarXT:
I fixed the bug report template so it clarifies better where the client.log file can be found (1.4 is the default branch and therefore should be mentioned equally, if not more to 1.3)


Chicken-Bones:
Put the 1.4 instructions at the top. Maybe with a note about "Right Click > Manage > Browse Local Files" on tModLoader in the list of games in your steam library

LandarXT:
Updated in [7453d6f](https://github.com/tModLoader/tModLoader/pull/2847/commits/7453d6ffa1df9348cb57a8934b50d8d611b23586) as per your comment

direwolf420:
missing a `)` in the last commit

PR #2868, opened 2022-08-22
Make GenPass OnBegin and OnComplete usable

Ishigh1:
For now, both of these methods are pretty unusable since each of them can contain only one method that gets overridden if you try to add one.
In this PR, I made both event based so multiple sources can add OnBegin/OnComplete actions, makeing it mod friendly
The usage of both are to do quick changes before/after a genpass instead of having to add new passes


ExterminatorX99:
Is `remove` needed? If it's not, this can be a smaller patch

Chicken-Bones:
OnBegin and OnComplete are going away in 1.4.4.
We'd prefer a more PlayerDrawLayer style approach to world-gen passes. Does this have any advantage over the current method of `InsertBefore/After` in the pass list?

PR #2869, opened 2022-08-22
OnStack Hook

andro951:
### What is the new feature?
``OnStack(Item increase, Item decrease, int amountBeingTransferred)`` method added to ``ModItem`` and ``GlobalItem`` to complement the existing ``CanStack()`` method.
``SplitStack(Item increase, Item decrease, int amountBeingTransferred)`` method added to ``ModItem`` and ``GlobalItem``.  
OnStack is called when stacks are merged.  SplitStack is called when a new item is created by splitting a stack.
These changes are only applicable to items with a maxStack > 1
``increase`` will always be the item with the stack being increased.
``decrease`` will always be the item with the stack being reduced (or removed if no remaining stack)
(OnStack includes 2 common functions to handle item stacking located in ``ItemLoader``, StackItems() and TryStackItems())

tml does not have a good way to manage or preserve changes to an item when stacked together with another item.
Stacking items results in copying or deleting data from one item to the other.

The current best option is to add checks into the ``CanStack()`` hook to try to only execute the desired code when the items will actually stack together.  This is not how ``CanStack()`` should be used and causes issues with other mods such as Magic Storage.

 The main option for a difference could be to have the ``OnStack()`` only call the ``ModItem.OnStack()`` and ``GlobalItem.OnStack()`` methods instead of managing the stacking of items.

public override void OnStack(Item increase, Item decrease, int numberToBeTransfered) {
	if (!increase.TryGetGlobalItem(out WeaponWithGrowingDamage weapon1) || !decrease.TryGetGlobalItem(out WeaponWithGrowingDamage weapon2))
		return;

	//Transfer experience and value to increase.
	weapon1.experience += weapon2.experience;
	weapon1.UpdateValue(increase, numberToBeTransfered);

	if (decrease.stack > numberToBeTransfered) {
		//Prevent duplicating the experience by clearing it on item2 if item2 will still exist.
		weapon2.experience = 0;
		weapon2.UpdateValue(decrease, -numberToBeTransfered);
	}
}

### ExampleMod updates
Unique variable example:
Modified WeaponWithGrowingDamage to apply to all weapons including stackable weapons.  It increases weapon damage based on item experience.  OnStack() is used to transfer all experience from one stack to another when stacking together to prevent duplication.

Multiple uses example:
ExampleMultiUseItem - A 4 use item that merges it's remaining uses when combined with another stack.

Averaging variable example:
ExampleStackableDurabilityItem - An item with a durability stat that is averaged between all items in a stack.

### Porting notes
Modified RecipeLoader.OnCraft(), OnCraftCallback() and RecipeCreationContext to include the original item (Item destinationStack) when crafting a stackable item with the same item in the mouseItem.
CanStack, CanStackInWorld, OnStack, and SplitStack are all instanced.  The first item passed (increase) is the instance used.  Consistent with the ModItem methods.

TODO:
Find where the split stack code is for equipping a bucket from a stack of buckets from left clicking on the player's helmet slot and add a SplitStack hook.


patches/tModLoader/Terraria/GameContent/ItemShopSellbackHelper.cs.patch
```diff
@@ -0,0 +1,11 @@
+--- src/TerrariaNetCore/Terraria/GameContent/ItemShopSellbackHelper.cs
++++ src/tModLoader/Terraria/GameContent/ItemShopSellbackHelper.cs
```
andro951: Don't know why the name changed

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -17,8 +17,21 @@ internal class ItemWithGrowingDamage : GlobalItem
...
+			switch (entity.type) {
+				case ItemID.CoinGun:
+					isWeapon = true;
+					break;
+				default:
+					isWeapon = entity.damage > 0 && entity.ammo == 0;
+					break;
+			}
+
+			return isWeapon;
```
Chicken-Bones: What's the motivation behind this change? Is it relevant to OnStack?

In general, I'd consider regular boolean logic to be preferable to this verbose `switch` style.
I don't think hard-coding to specific item ids is a good idea, since it promotes poor mod interop

patches/tModLoader/Terraria/GameContent/ItemDropRules/HerbBagDropsItemDropRule.cs
```diff
@@ -33,7 +33,7 @@ public class HerbBagDropsItemDropRule : IItemDropRule
 			for (int i = 0; i < amount; i++) {
 				int stack = Main.rand.Next(2, 5);
 				if (Main.rand.Next(3) == 0)
-					stack += Main.rand.Next(1, 5);
+					stack += Main.rand.Next(1, 5);//Planning to skip
```
Chicken-Bones: slipped in by accident?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1014,6 +1014,15 @@
  										flag2 = true;
  										break;
  									}
+@@ -22903,7 +_,7 @@
+ 								chest.item[num9].SetDefaults(3093);
+ 								chest.item[num9].stack = 1;
+ 								if (genRand.Next(5) == 0)
+-									chest.item[num9].stack += genRand.Next(2);
++									chest.item[num9].stack += genRand.Next(2);//Planning to ignore these
```
Chicken-Bones: Probably a leftover comment from when you were looking at places to change? No need to call OnStack when adjusting the stack of a `SetDefaults` call

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -686,6 +686,17 @@ public abstract class GlobalItem : GlobalType<Item, GlobalItem>
...
 		}
 
+		/// <summary>
+		/// Allows you to make things happen when items stack together.
+		/// </summary>
+		/// <param name="item1">The item that will be removed or have its stack reduced.</param>
+		/// <param name="item2">The item that will have its stack increased.</param>
+		/// <param name="numberToBeTransfered">The number to be transfered from item1 to item2.</param>
+		/// <param name="firstStackSplit">Only true when right clicking an item when the mouse item is air.  The clicked item is cloned and stack set to 0.</param>
+		public virtual void OnStack(Item item1, Item item2, int numberToBeTransfered, bool firstStackSplit) {
```
Chicken-Bones: This documentation is what I would expect, but we should call them `from` and `to` or `source` and `dest` to make it very clear which is which.

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -1428,6 +1428,62 @@ public static void OpenVanillaBag(string context, Player player, int arg)
...
+					return;
+
+				numTransfered = Math.Min(transferingItem.stack, numToMaxStack);
+			}
+			else {
+				numTransfered = numToTransfer;
+			}
+
+			foreach (var g in HookOnStack.Enumerate(globalItems)) {
+				g.OnStack(acceptingItem, transferingItem, numTransfered, firstStackSplit);
```
Chicken-Bones: This order `acceptingItem, transferingItem` is actually the opposite of what the documentation in `OnStack` says. I think the documentation is what people will expect, so you should swap the order _here_

Chicken-Bones (changes requested):
Mini review

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: updated

patches/tModLoader/Terraria/GameContent/ItemDropRules/HerbBagDropsItemDropRule.cs
andro951: removed

patches/tModLoader/Terraria/WorldGen.cs.patch
andro951: removed

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
andro951: changed them to increase and decrease

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
andro951: I kept it the same order and swapped the documentation so it's consistent with the CanStack() and CanStackInWorld() hooks.

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
```diff
@@ -0,0 +1,41 @@
...
+	public class ProjectileWithGrowingDamage : GlobalProjectile
+	{
+		//Related to GlobalItem: WeaponWithGrowingDamage
+
+		private Item sourceItem;
+
+		public override bool InstancePerEntity => true;
+		public override void OnSpawn(Projectile projectile, IEntitySource source) {
+			if (source is EntitySource_ItemUse itemSource) {
+				sourceItem = itemSource.Item;
```
Chicken-Bones: Unfortunately you cannot do this, unless you clone it (which I would not recommend).

Because `Item` instances are reused in Terraria (via `SetDefaults`) the item could change into air, be swapped with a different item, be replaced with one you picked up, etc before the projectile lands.

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
```diff
@@ -0,0 +1,41 @@
...
+			}
+			else if (source is EntitySource_ItemUse_WithAmmo itemWithUseAmmoSource) {
+				sourceItem = itemWithUseAmmoSource.Item;
+			}
+		}
+
+		public override void OnHitNPC(Projectile projectile, NPC target, int damage, float knockback, bool crit) {
+			if (sourceItem == null || !sourceItem.TryGetGlobalItem(out WeaponWithGrowingDamage weapon))
```
Chicken-Bones: The best you can do instead, is to check `plr.inventory[plr.selectedItem]`. I'd suggest saving the `type` of item that spawned the projectile, and checking that too.

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+		public override void OnStack(Item item1, Item item2, int numberToBeTransfered) {
+			if (!item1.TryGetGlobalItem(out WeaponWithGrowingDamage weapon1) || !item2.TryGetGlobalItem(out WeaponWithGrowingDamage weapon2))
+				return;
+
+			if (item1.stack == 0)
+				weapon1.experience = 0;
+			
+			//Transfer experience and value to item1.
+			weapon1.experience += weapon2.experience;//Works
+			//experience += weapon2.experience;//Doesn't work
```
Chicken-Bones: remove this comment and the `//Works` one above

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+			weapon1.UpdateValue(item1, numberToBeTransfered);
+
+			if (item2.stack > numberToBeTransfered) {
+				//Prevent duplicating the experience by clearing it on item2 if item2 will still exist.
+				weapon2.experience = 0;
+				weapon2.UpdateValue(item2, -numberToBeTransfered);
+			}
+		}
```
Chicken-Bones, line 142: good logic

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+	}
+	public class SlowBallShop : GlobalNPC
```
Chicken-Bones: `Snowball`?

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+	{
+		public override void SetupShop(int type, Chest shop, ref int nextSlot) {
+			if (type != ModContent.NPCType<ExamplePerson>())
+				return;
+
+			shop.item[nextSlot].SetDefaults(ItemID.Snowball);
+			if (shop.item[nextSlot].TryGetGlobalItem(out WeaponWithGrowingDamage weapon))
+				weapon.experience *= 2;
```
Chicken-Bones: Prefer to just set a value here.

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+		}
+		public override void SetDefaults(Item item) {
+			if (item.type == ItemID.Snowball) {
+				experience = 1;
+				UpdateValue(item);
+			}
+		}
+		public override void LoadData(Item item, TagCompound tag) {
+			int xp = tag.Get<int>("experience");//Load experience tag
+			GainExperience(item, xp, true);
```
Chicken-Bones: is there a need for the `setXp` parameter? All fields should be `0` in `LoadDAta` and `NetReceive`

Chicken-Bones: Otherwise, just set `experience` to 0 before calling `GainExperience`

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace ExampleMod.Common.GlobalItems
+{
+	public class WeaponWithGrowingDamage : GlobalItem
+	{
+		//Related to GlobalProjectile: ProjectileWithGrowingDamage
+
+		public int experience;
+		public static int experiencePerLevel = 100;
+		private int lastBonusValue;
+		private int stack = int.MinValue;
```
Chicken-Bones: it's definitely confusing that you store `stack` here and on `Item`

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+				foreach (Item ingredient in rContext.ConsumedItems) {
+					if (ingredient.TryGetGlobalItem(out WeaponWithGrowingDamage ingredientGlobal)) {
+						//Transfer all experience from consumed items to the crafted item.
+						GainExperience(item, ingredientGlobal.experience);
+					}
+				}
+			}
+		}
+
+		public override void OnStack(Item item1, Item item2, int numberToBeTransfered) {
```
Chicken-Bones: rename parameters to `increase` and `decrease`

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
```diff
@@ -682,10 +682,20 @@ public abstract class GlobalItem : GlobalType<Item, GlobalItem>
 		/// <br/>This is only called when two items of the same type attempt to stack.
 		/// </summary>
 		/// <returns>Whether or not the items are allowed to stack</returns>
-		public virtual bool CanStackInWorld(Item item1, Item item2) {
+		public virtual bool CanStackInWorld(Item increase, Item decrease) {
 			return true;
 		}
 
+		/// <summary>
+		/// Allows you to make things happen when items stack together.
```
Chicken-Bones, line 690: Needs an additional comment at the end with a space:
```
///
/// This method is not instanced.
```

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+
+namespace ExampleMod.Common.GlobalItems
+{
+	public class WeaponWithGrowingDamage : GlobalItem
+	{
+		//Related to GlobalProjectile: ProjectileWithGrowingDamage
+
+		public int experience;
+		public static int experiencePerLevel = 100;
+		private int lastBonusValue;
```
Chicken-Bones: if you make this `bonusValuePerItem` then you only need this one field.

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+using ExampleMod.Content.NPCs;
+
+namespace ExampleMod.Common.GlobalItems
+{
+	public class WeaponWithGrowingDamage : GlobalItem
+	{
+		//Related to GlobalProjectile: ProjectileWithGrowingDamage
+
+		public int experience;
+		public static int experiencePerLevel = 100;
```
Chicken-Bones, line 16: put this at the top please

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -0,0 +1,176 @@
...
+
+			int newBonusValue = experience * 5;//experience: 40, newBonusValue = 200
+			int newBonusValuePerItem = newBonusValue / stack;//newBonusValue: 200, stack: 100, newBonusValuePerItem = 2
+			item.value += newBonusValuePerItem - lastBonusValuePerItem;//newBonusValuePerItem: 2, lastBonusValuePerItem, 20, item.value += -18
+			lastBonusValue = newBonusValue;//newBonusValue: 200, lastBonusValue = 200
+		}
+
+		private bool CheckStackZero(Item item) {
+			if (stack <= 0) {
+				item.value = ContentSamples.ItemsByType[item.type].value;
```
Chicken-Bones: not a good idea, will overwrite value changes from other mods. Shouldn't be needed.
`UpdateValue` can just be
```
item.value -= lastBonusValuePerItem;

lastBonusValuePerItem = experience * 5 / item.stack;
item.value += lastBonusValuePerItem;
```

Chicken-Bones (changes requested):
Example changes requested. Will double check hooks tomorrow.

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
andro951: I added in the type checking.  Do you think that will be good enough?  It would just fail to gain xp if the item was changed, but the vast majority of the time, it would work.  This is the way I handle it in my mod and haven't noticed issues with it.  (I hadn't considered the game re-using the item instance after setting it as a source item, so I should add some additional checking there.)  It is missing the search method I use if the item has changed inventory slots.  Do you think I should add that in?

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
andro951: Is that the same as player.HeldItem?  You could check it that way, but there would be no way for minion weapons to ever gain xp unless you held your minion staff the whole time.  I would make a list on the GlobalItem that tracks all projectiles spawned by the associated item instead of a projectile tracking the source item.

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: removed

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: fixed

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: fixed

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
Chicken-Bones: I wouldn't worry about searching for the item if it's moved to a new slot, but I don't want to show storing the `Item` from `OnSpawn` in `ExampleMod` because even if you know how to safely guard against the item instance changing, others will not

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: Moved to just above the namespaces

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
Chicken-Bones: `public Item HeldItem => inventory[selectedItem];`

Apparently, `HeldItem` is definitely good then. If you wanted this to work for minion items and such, then inventory search is definitely the way to go.

Actually... on second thoughts, you can't cache the `Item` reference, but you definitely can cache the `WeaponWithGrowingDamage` instance!

Chicken-Bones: Just make sure you comment in `OnSpawn` about the trap

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: Updated

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
andro951: updated

patches/tModLoader/Terraria/ModLoader/GlobalItem.cs
andro951: I added this comment to CanStack, CanStackInWorld and OnStack.  Do you agree with the first two?  I set up OnStack the same way as them, so I'm assuming they are also not instanced.

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
andro951: done

ExampleMod/Common/GlobalProjectiles/ProjectileWithGrowingDamage.cs
andro951: updated

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -152,4 +155,80 @@ public class SnowBallShop : GlobalNPC
 			}
 		}
 	}
+
+
+	public class StackableTestItem : GlobalItem
```
Chicken-Bones: Maybe `[Autoload(false)]` this one?

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEFoodPlatter.cs.patch
```diff
@@ -53,25 +54,39 @@
...
+@@ -114,10 +_,8 @@
  			if (tEFoodPlatter.item.stack > 0)
  				tEFoodPlatter.DropItem();
  
 -			tEFoodPlatter.item = new Item();
-+			tEFoodPlatter.item = item.Clone();
 -			tEFoodPlatter.item.netDefaults(netid);
 -			tEFoodPlatter.item.Prefix(prefix);
- 			tEFoodPlatter.item.stack = stack;
+-			tEFoodPlatter.item.stack = stack;
```
Chicken-Bones: Comment rather than removing

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEWeaponsRack.cs.patch
```diff
@@ -53,27 +54,41 @@
...
-@@ -158,10 +_,14 @@
+@@ -158,10 +_,13 @@
  			if (tEWeaponsRack.item.stack > 0)
  				tEWeaponsRack.DropItem();
  
 +			/*
  			tEWeaponsRack.item = new Item();
  			tEWeaponsRack.item.netDefaults(netid);
  			tEWeaponsRack.item.Prefix(prefix);
+-			tEWeaponsRack.item.stack = stack;
```
Chicken-Bones: No need to remove line if it's commented

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEFoodPlatter.cs.patch
```diff
@@ -53,25 +54,42 @@
...
  
--			tEFoodPlatter.item = new Item();
-+			tEFoodPlatter.item = item.Clone();
--			tEFoodPlatter.item.netDefaults(netid);
--			tEFoodPlatter.item.Prefix(prefix);
++			/*
+ 			tEFoodPlatter.item = new Item();
+ 			tEFoodPlatter.item.netDefaults(netid);
+ 			tEFoodPlatter.item.Prefix(prefix);
  			tEFoodPlatter.item.stack = stack;
```
andro951, line 65: Not sure how to fix this

PR #2875, opened 2022-08-26
made most of Terraria.ModLoader.Config.UI public

Tyfyter:
### What is the new feature?
every class, method, and field in Terraria.ModLoader.Config.UI which was previously private or internal has been changed to public
currently using custom config elements is supported, but everything from displaying a tooltip to using preexisting config elements inside a custom one requires reflection, making actually creating custom config elements excessively difficult.

none which are as effective except changing the protected members as well

a config element which only shows settings which currently have multiple valid values


direwolf420:
This is just a "port" of this PR, right? #2173

Tyfyter:
Same goal and methodology, but this time without discriminating based on whether or not I could figure out how the class works

JavidPack:
I've been planning on fixing a bunch of things with these classes, if we make them public, it will be much harder to avoid breaking compatibility outright. I think we won't merge this PR until more work is done on the classes themselves to make sure they are ready for inheriting from. The whole design is probably going to change.

Issue #2874, opened 2022-08-24
Popup text from items using ModRarity won't disappear when its mod is disabled


Jadams505:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

The text that shows up when you pickup/reforge an item does not disappear if it has a custom rarity and the mod declaring that custom rarity is disabled.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/9419741/client.log)

### Steps to reproduce

1. Enable a mod that uses ModRarity (I used ExampleMod)
2. Drop an item that uses that rarity (I used ExampleSword)
3. Pick up that item
4. Quickly exit the world before the text has time to disappear
5. Disable the mod you enabled in step one
6. Login to the world
7. Return to the place where the item was picked up in step 3 and the text should still be there

### Expected Behavior

The text should fade away after a few seconds

### Actual Behavior

The text remains even between worlds and players

### Reproduction frequency

Everytime

Chicken-Bones:
Fixed in a741c55

PR #2877, opened 2022-08-27
Fixed popup text with custom rarities not disappearing

Jadams505:
### What is the bug?
See #2874 

### How did you fix the bug?
Only updates the color if the rarity is less than the number of rarities in `RarityLoader`


Chicken-Bones:
This might be an appropriate fix, but it indicates a flaw in the underlying design.
'unloaded' rarities should still have an entry in Rarity loader and an id in range (maybe this wasn't possible, I'll need to look a bit more).

Chicken-Bones:
Ok wow, so I read the log. Looks like the right thing to do is clear all popup texts on world exit

Chicken-Bones:
Fixed in a741c55

PR #2909, opened 2022-09-09
Max Health and Mana Manipulation API

absoluteAquarian:
### What is the new feature?
This PR aims to make modifying the player's maximum health and mana, as well as their displays, more friendly to mod developers.

Features include:
- a proper API for modifying `Player.statLifeMax` and `Player.statManaMax`
- being able to hook into the vanilla drawing for the hearts, stars and bars in the vanilla resource display sets via `ModResourceOverlay`
- allowing mod developers to define their own display sets, which can be chosen from the Settings menu, via `ModResourceDisplaySet`
  - choosing `ModResourceDisplaySet` displays is handled via the pre-existing option in the ingame Settings>Video menu
  - consequently, modded display sets are also saved to `config.json`
- removal of any code which limited the values of `Player.statLifeMax` and `Player.statManaMax`
- several new properties in `Player` for tracking how many Life Crystals, Life Fruit and Mana Crystals have been consumed
  - `Player.ConsumedLifeCrystals`
  - `Player.ConsumedLifeFruit`
  - `Player.ConsumedManaCrystals`
- a new helper method for increasing the player's health/mana and displaying the combat text via `Player.UseHealthMaxIncreasingItem()` and `Player.UseManaMaxIncreasingItem()`

Modifying the player's maximum health and mana currently requires using `Player.statLifeMax2` and `Player.statManaMax2`, which -- even though it works -- is clunky and goes against the purpose of those fields: "temporary" increases/decreases.

Furthermore, modifying how vanilla's hearts/stars/bars for health and mana are drawn is cumbersome, requiring a method detour or IL edit to be used, which can cause conflicts when multiple mods try to do so at once.

As a side note, any legacy usage of modifying `Player.statLifeMax2` and `Player.statManaMax2` will still be supported, but mod developers should be encouraged to move to the new API.

n/a

Modifying the player's maximum health:
```cs
// source:  Common/Players/ExampleStatIncreasePlayer.cs
// NOTE: The full source contains code for handling increasing the player's max mana as well.
//       It has been reduced for brevity's sake.

public int exampleLifeFruits;

public override void ModifyMaxStats(out StatModifier health, out StatModifier mana) {
	health = StatModifier.Default;
	health.Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit;
	// Alternatively:  health = StatModifier.Default with { Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit };
	mana = StatModifier.Default;
}
```

Drawing custom hearts over the hearts in the Classic resource display set:
```cs
// source:  Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
// NOTE: The full source contains code for handling the Fancy and Bars display sets.
//       It has been reduced for brevity's sake.

public override void PostDrawResource(ResourceOverlayDrawContext context) {
	// Hearts are replaced in groups of two
	if (context.resourceNumber > 2 * Main.LocalPlayer.GetModPlayer<ExampleLifeFruitPlayer>().exampleLifeFruits)
		return;

	Asset<Texture2D> asset = context.texture;

	// "context" contains information used to draw the resource
	// If you want to draw directly on top of the vanilla hearts, just replace the texture and have the context draw the new texture
	if (asset == TextureAssets.Heart || asset == TextureAssets.Heart2) {
		context.texture = ModContent.Request<Texture2D>("ExampleMod/Common/UI/ResourceOverlay/ClassicHeartOverlay", AssetRequestMode.ImmediateLoad);
		context.Draw();
	}
}
```

### ExampleMod updates
- `Content/Items/Consumables/ExampleLifeFruit.cs` was updated to use the new API
- added `Content/Items/Consumables/ExampleManaCrystal.cs`, which serves as an example mana increasing item
- added the `Common/UI/ResourceOverlay` folder, which contains examples of modifying vanilla's resource display sets
- added the `Common/UI/ExampleDisplaySets` folder, which contains an example of using `ModResourceDisplaySet`
- `ExampleLifeFruitPlayer` was renamed to `ExampleStatIncreasePlayer` and was moved to `Common/Players`
- updated the netcode handling in `ExampleMod.Networking.cs` to account for the new `ExampleManaCrystal` stats
- updated the `en-US.hjson` localization file to add an entry for `ExampleReversedBarsDisplay` (the `ModResourceDisplaySet` example)

### Additional Information
- the netcode handling in `ModLoader/Default/ModLoaderMod.cs` was modified to be more friendly to future additions
  - consequently, the netcode for `ModAccessorySlotPlayer` was also modified
- a lot of vanilla code that was checking `Player.statLifeMax` and/or `Player.statManaMax` were changed to check the `Player.ConsumedLifeCrystals`, `Player.ConsumedLifeFruit` and `Player.ConsumedManaCrystals` properties instead, since that's what was really happening in the first place
- added two new `AmountOfLifeHearts` and `AmountOfManaStars` properties to `PlayerStatsSnapshot` and changed the `LifePerSegment` and `ManaPerSegment` fields to be getter properties dependent on these new properties
- `PlayerResourceSetsManager` (aka `Main.ResourceSetsManager`) was expanded to include `ModResourceDisplaySet` instances
  - assuming that the `ModResourceDisplaySet` was loaded, you can get the instance via `Main.ResourceSetsManager.GetDisplaySet<T>()`
  - you can also get the instance via `Main.ResourceSetsManager.GetDisplaySet("ModName/Name")`
- usage of `IPlayerResourceDisplaySet.NameKey` was removed in favor of a `DisplayedName` getter property, though the property still exists for whoever might want to use it


ExampleMod/Common/Players/ExampleStatIncreasePlayer.cs
```diff
@@ -0,0 +1,42 @@
...
+			health.Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit;
+			// Alternatively:  health = StatModifier.Default with { Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit };
+			mana = StatModifier.Default;
+			mana.Base = exampleManaCrystals * ExampleManaCrystal.ManaPerCrystal;
+			// Alternatively:  mana = StatModifier.Default with { Base = exampleManaCrystals * ExampleManaCrystal.ManaPerCrystal };
+		}
+
+		public override void SyncPlayer(int toWho, int fromWho, bool newPlayer) {
+			ModPacket packet = Mod.GetPacket();
+			packet.Write((byte)ExampleMod.MessageType.ExamplePlayerSyncPlayer);
```
Chicken-Bones: rename to `ExampleStatIncreasePlayerSync`

ExampleMod/Common/Players/ExampleStatIncreasePlayer.cs
```diff
@@ -0,0 +1,42 @@
...
+		public override void ModifyMaxStats(out StatModifier health, out StatModifier mana) {
+			health = StatModifier.Default;
+			health.Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit;
+			// Alternatively:  health = StatModifier.Default with { Base = exampleLifeFruits * ExampleLifeFruit.LifePerFruit };
+			mana = StatModifier.Default;
+			mana.Base = exampleManaCrystals * ExampleManaCrystal.ManaPerCrystal;
+			// Alternatively:  mana = StatModifier.Default with { Base = exampleManaCrystals * ExampleManaCrystal.ManaPerCrystal };
+		}
+
+		public override void SyncPlayer(int toWho, int fromWho, bool newPlayer) {
```
Chicken-Bones, line 23: Also need to implement `clientClone` and `SendClientChanges` to avoid desync when a player eats some mid-game

ExampleMod/Common/Systems/ExampleStatIncreaseSystem.cs
```diff
@@ -0,0 +1,29 @@
...
+			// Due to one Example Mana Crystal increasing the player's maximum mana by only 10, some logic should be used to help make the increases seem fluid
+			int manaCrystals = player.GetModPlayer<ExampleStatIncreasePlayer>().exampleManaCrystals;
+
+			if (manaCrystals <= 0) {
+				// Nothing to do
+				return;
+			}
+
+			// Every odd number should display an extra star/bar compared to the number before it
+			int extraStarsToDraw = manaCrystals % 2 == 1 ? 1 : 0;
```
Chicken-Bones: I don't think this is a particularly good example, it seems confusing

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
```diff
@@ -0,0 +1,66 @@
...
+ 	{
+ 		public delegate void TextureGetter(int elementIndex, int firstElementIndex, int lastElementIndex, out Asset<Texture2D> texture, out Vector2 drawOffset, out float drawScale, out Rectangle? sourceRect);
+ 
++		public delegate void ResourceOverlayDelegate(int elementIndex, int firstElementIndex, int lastElementIndex, out PlayerStatsSnapshot snapshot, out IResourceDrawSource drawSource);
++
+ 		public Vector2 TopLeftAnchor;
+ 		public int ElementCount;
+ 		public int ElementIndexOffset;
+ 		public TextureGetter GetTextureMethod;
++		public ResourceOverlayDelegate GetResourceDrawContext;
```
Chicken-Bones: Why does this need to be a delegate?

patches/tModLoader/Terraria/ModLoader/ResourceDrawSources.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Terraria.GameContent.UI.ResourceSets;
+
+#nullable enable
+
+namespace Terraria.ModLoader
+{
+	public abstract class ResourceDrawSource : IResourceDrawSource
+	{
+		public string Context { get; protected set; }
```
Chicken-Bones: no point of a context param which is always null.

patches/tModLoader/Terraria/ModLoader/ResourceDrawSources.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Terraria.GameContent.UI.ResourceSets;
+
+#nullable enable
+
+namespace Terraria.ModLoader
+{
+	public abstract class ResourceDrawSource : IResourceDrawSource
```
Chicken-Bones: Do we need this class? What's wrong with just the `DisplaySet`?

Do we need all the sub-classes?

patches/tModLoader/Terraria/ModLoader/IResourceDrawSource.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Terraria.ModLoader
+{
+	public interface IResourceDrawSource
+	{
+		string? Context { get; }
+
+		IPlayerResourcesDisplaySet DisplaySet { get; }
+	}
+}
```
Chicken-Bones: don't make an interface when only one class implements it

patches/tModLoader/Terraria/ModLoader/ModResourceDisplaySet.cs
```diff
@@ -0,0 +1,573 @@
...
+		/// <param name="position">The position to draw the life at</param>
+		/// <param name="sourceFrame">The area within the life texture to draw.  Defaults to the entire texture</param>
+		/// <param name="color">The color to draw the life with.  Defaults to <see cref="Color.White"/></param>
+		/// <param name="rotation">The rotation to draw the life with.  Defaults to 0 radians</param>
+		/// <param name="origin">The relative center within the frame for rotation and scaling.  Defaults to the center of the frame</param>
+		/// <param name="scale">The scale to draw the life at.  Defaults to <see cref="Vector2.One"/></param>
+		/// <param name="pulse">Whether the life's scale should be modified by the "pulse" effect.  Defaults to <see langword="false"/></param>
+		/// <returns><see langword="null"/> if the parameters are invalid, otherwise the created context</returns>
+		public static ResourceOverlayDrawContext? GetClassicHeart(SpriteBatch spriteBatch, int heartNumber, Vector2 position, Rectangle? sourceFrame = null, Color? color = null, float rotation = 0, Vector2? origin = null, Vector2? scale = null, bool pulse = false) {
```
Chicken-Bones: These static helpers look really hard to maintain, and aren't used by `ExampleReversedBarsDisplay`

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
```diff
@@ -0,0 +1,222 @@
...
+			HorizontalBarsPlayerReosurcesDisplaySet.FillBarByValues(elementIndex, sprite, _hpSegmentsCount, _hpPercent, out offset, out drawScale, out sourceRect);
+		}
+
+		private void ManaFillingDrawer(int elementIndex, int firstElementIndex, int lastElementIndex, out Asset<Texture2D> sprite, out Vector2 offset, out float drawScale, out Rectangle? sourceRect) {
+			sprite = _mpFill;
+			HorizontalBarsPlayerReosurcesDisplaySet.FillBarByValues(elementIndex, sprite, _mpSegmentsCount, _mpPercent, out offset, out drawScale, out sourceRect);
+		}
+
+		private void LifePanelContext(int elementIndex, int firstElementIndex, int lastElementIndex, out PlayerStatsSnapshot snapshot, out IResourceDrawSource drawSource) {
+			snapshot = preparedSnapshot;
```
Chicken-Bones: Not a fan of this boilerplate. tML should handle this since it's exactly the same in every implementation

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
```diff
@@ -0,0 +1,75 @@
+--- src/TerrariaNetCore/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs
++++ src/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs
+@@ -1,3 +_,5 @@
++using Terraria.ModLoader;
++
+ namespace Terraria.GameContent.UI.ResourceSets
+ {
+ 	public struct PlayerStatsSnapshot
```
Chicken-Bones, line 8: I'm not convinced of the value of this feature

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
absoluteAquarian: `PlayerStatsSnapshot` is a vanilla type that i modified to make setting how many heart/stars/bars draw easier  
otherwise, modders would have to manually calculate the stars to draw instead of just doing something like `snapshot.AmountOfManaStars++;`

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
absoluteAquarian: it doesn't have to be, but it was the simplest way to implement it if a modder wanted to use `ResourceDrawSettings` in, say, their modded display sets

absoluteAquarian: edit: decided to add fields to the type instead

ExampleMod/Common/Players/ExampleStatIncreasePlayer.cs
```diff
@@ -0,0 +1,61 @@
...
+		public override void SendClientChanges(ModPlayer clientPlayer) {
+			ExampleStatIncreasePlayer clone = clientPlayer as ExampleStatIncreasePlayer;
+
+			if (exampleLifeFruits != clone.exampleLifeFruits || exampleManaCrystals != clone.exampleManaCrystals) {
+				ModPacket packet = Mod.GetPacket();
+				packet.Write((byte)ExampleMod.MessageType.ExampleStatIncreasePlayerSync);
+				packet.Write((byte)Player.whoAmI);
+				packet.Write((byte)exampleLifeFruits);
+				packet.Write((byte)exampleManaCrystals);
+				packet.Send(ignoreClient: Player.whoAmI);
```
Chicken-Bones: duplicate code with `SyncPlayer` above, consider common function, or just call `SyncPlayer` with `toWho: -1`

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
```diff
@@ -0,0 +1,203 @@
...
+			_panelLeft = Main.Assets.Request<Texture2D>(vanillaFolder + "Panel_Left", AssetRequestMode.ImmediateLoad);
+			_panelMiddleHP = Main.Assets.Request<Texture2D>(vanillaFolder + "HP_Panel_Middle", AssetRequestMode.ImmediateLoad);
+			_panelRightHP = ModContent.Request<Texture2D>(modFolder + "HP_Panel_Right", AssetRequestMode.ImmediateLoad);
+			_panelMiddleMP = Main.Assets.Request<Texture2D>(vanillaFolder + "MP_Panel_Middle", AssetRequestMode.ImmediateLoad);
+			_panelRightMP = ModContent.Request<Texture2D>(modFolder + "MP_Panel_Right", AssetRequestMode.ImmediateLoad);
+		}
+
+		// DrawLife runs before DrawMana, so keep that in mind
+		// In this case, the bars are swapped, so the order of drawing (top bar, then bottom bar) should be preserved
+		// This is why Mana resources are being drawn in DrawLife and Life resources are being drawn in DrawMana
```
Chicken-Bones: why must the order be preserved (top bar then bottom bar)?

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+			} else {
+				// None of the above cases applied, but this could be a modded resource that uses vanilla textures.
+				// So, it would be a good idea to check if certain vanilla resources are being drawn and then draw over them like usual
+				string name = context.texture.Name;
+
+				string fancyFolder = "Images/UI/PlayerResourceSets/FancyClassic/";
+				string barsFolder = "Images/UI/PlayerResourceSets/HorizontalBars/";
+
+				// NOTE: CompareAssetNames is defined below this method's body
+				if (name == TextureAssets.Heart.Name || name == TextureAssets.Heart2.Name) {
```
Chicken-Bones: compare assets directly using `==`, don't use `.Name`

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+				DrawClassicFancyOverlay(context);
+			} else if (context.IsSlot(FancyClassicPlayerResourcesDisplaySet.HeartPanelsFirstRow) || context.IsSlot(FancyClassicPlayerResourcesDisplaySet.HeartPanelsSecondRow)) {
+				// Draw over the Fancy heart panels
+				DrawFancyPanelOverlay(context);
+			} else if (context.IsSlot(HorizontalBarsPlayerReosurcesDisplaySet.LifePanels) && fillResourceNumber >= 1 && fillResourceNumber <= context.snapshot.AmountOfLifeHearts) {
+				// Draw over the Bars middle life panels
+				DrawBarsPanelOverlay(context);
+			} else if (context.IsSlot(HorizontalBarsPlayerReosurcesDisplaySet.LifeBars)) {
+				// Draw over the Bars life bars
+				DrawBarsOverlay(context);
```
Chicken-Bones: As nice as it is to showcase `IsSlot`, I think there's no point having two codepaths, when the fallback works for all. If you're just showcasing texture replacement for fills, then you should use the asset based codepath only.

A different example would be better for showcasing `IsSlot` (perhaps restyling the borders if you have full health or something?)

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+		public override void PostDrawResource(ResourceOverlayDrawContext context) {
+			// Life resources are drawn over in groups of two
+			// Bars panel drawing has two additional elements, so the "resource index" needs to be offset
+			int fillResourceNumber = context.IsSlot(HorizontalBarsPlayerReosurcesDisplaySet.LifePanels) ? context.resourceNumber - 1 : context.resourceNumber;
+			int exampleFruits = Main.LocalPlayer.GetModPlayer<ExampleStatIncreasePlayer>().exampleLifeFruits;
+
+			bool canReplaceResource;
+
+			if (context.IsSlot(HorizontalBarsPlayerReosurcesDisplaySet.LifePanels)) {
+				// Vanilla "replaces" the bars from right to left, so it would be a good idea to mimic that behaviour
```
Chicken-Bones: this is annoying, can we change the draw order in the vanilla and example mod bars so modders don't need to do this? It fails with the example bars, because we don't check that slot (and in general, may fail with modded resource UIs if they change draw order)

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.UI.ResourceDisplay
+{
+	public class VanillaLifeOverlay : ModResourceOverlay
+	{
+		public override void PostDrawResource(ResourceOverlayDrawContext context) {
+			// Life resources are drawn over in groups of two
+			// Bars panel drawing has two additional elements, so the "resource index" needs to be offset
+			int fillResourceNumber = context.IsSlot(HorizontalBarsPlayerReosurcesDisplaySet.LifePanels) ? context.resourceNumber - 1 : context.resourceNumber;
```
Chicken-Bones: is there no way to generalise this?

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+				} else if (CompareAssetNames(name, barsFolder + "HP_Panel_Middle")) {
+					// Draw over the Bars middle life panels
+					DrawBarsPanelOverlay(context);
+				}
+			}
+		}
+
+		private static bool CompareAssetNames(string existingName, string compareAssetPath) {
+			// This is a helper method for checking if a certain vanilla asset was drawn
+			return existingName == Main.Assets.Request<Texture2D>(compareAssetPath).Name;
```
Chicken-Bones: Not recommended to call `Request` on every frame. Cache instead

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.Xna.Framework.Graphics;
+using ReLogic.Content;
+using Terraria;
+using Terraria.GameContent;
+using Terraria.GameContent.UI.ResourceSets;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.UI.ResourceDisplay
+{
+	public class VanillaLifeOverlay : ModResourceOverlay
```
Chicken-Bones: Why does this example need to replace the panel, and not just filling? It seems particularly complicated as a result.

ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
```diff
@@ -1,56 +1,52 @@
...
+
+				// This method handles permanently increasing the player's max health and displaying the green heal text
+				player.UseHealthMaxIncreasingItem(LifePerFruit);
+
+				// This field tracks how many of the example fruit have been consumed
+				player.GetModPlayer<ExampleStatIncreasePlayer>().exampleLifeFruits++;
+				// This handles the 2 achievements related to using any life increasing item or reaching the max amount of consumed Life Crystals, Life Fruit and Mana Crystals.
+				// Ignored since our item is only useable after this achievement is reached
+				// AchievementsHelper.HandleSpecialEvent(player, 2);
+				//TODO re-add this when ModAchievement is merged?
```
Chicken-Bones: remove?

ExampleMod/Content/Items/Consumables/ExampleLifeFruit.cs
```diff
@@ -1,56 +1,52 @@
...
-
-			// This is very important. This is what makes it permanent.
-			player.GetModPlayer<ExampleLifeFruitPlayer>().exampleLifeFruits++;
-			// This handles the 2 achievements related to using any life increasing item or getting to exactly 500 hp and 200 mp.
-			// Ignored since our item is only useable after this achievement is reached
-			// AchievementsHelper.HandleSpecialEvent(player, 2);
-			//TODO re-add this when ModAchievement is merged?
-			return true;
+			// Returning null will make the item not be consumed
+			return canConsume ? true : null;
```
Chicken-Bones: just return `true` in the `if`. Better yet, invert the if and early return `null`

ExampleMod/Content/Items/Consumables/ExampleManaCrystal.cs
```diff
@@ -0,0 +1,62 @@
...
+				// This field tracks how many of the example crystals have been consumed
+				player.GetModPlayer<ExampleStatIncreasePlayer>().exampleManaCrystals++;
+				// This handles the 2 achievements related to using any mana increasing item or reaching the max amount consumed of Life Crystals, Life Fruit and Mana Crystals.
+				// Ignored since our item is only useable after this achievement is reached
+				// AchievementsHelper.HandleSpecialEvent(player, 1);
+				//TODO re-add this when ModAchievement is merged?
+			}
+
+			// Returning null will make the item not be consumed
+			return canConsume ? true : null;
```
Chicken-Bones: same comments as `ExampleLifeFruit`

ExampleMod/ExampleMod.Networking.cs
```diff
@@ -20,11 +20,12 @@ internal enum MessageType : byte
...
-				// This message syncs ExamplePlayer.exampleLifeFruits
-				case MessageType.ExamplePlayerSyncPlayer:
+				// This message syncs ExampleLifeIncreasePlayer.exampleLifeFruits and ExampleLifeIncreasePlayer.exampleManaCrystals
+				case MessageType.ExampleStatIncreasePlayerSync:
 					byte playernumber = reader.ReadByte();
-					ExampleLifeFruitPlayer examplePlayer = Main.player[playernumber].GetModPlayer<ExampleLifeFruitPlayer>();
-					examplePlayer.exampleLifeFruits = reader.ReadInt32();
+					ExampleStatIncreasePlayer examplePlayer = Main.player[playernumber].GetModPlayer<ExampleStatIncreasePlayer>();
+					examplePlayer.exampleLifeFruits = reader.ReadByte();
+					examplePlayer.exampleManaCrystals = reader.ReadByte();
```
Chicken-Bones: Please move into `ExampleStatIncreasePlayer` as an instance method.

ExampleMod/ExampleMod.Networking.cs
```diff
@@ -20,11 +20,12 @@ internal enum MessageType : byte
...
 					// SyncPlayer will be called automatically, so there is no need to forward this data to other clients.
```
Chicken-Bones: It will not as far as I can tell

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ClassicPlayerResourcesDisplaySet.cs.patch
```diff
@@ -0,0 +1,231 @@
...
+ 				}
+ 				else {
+-					float num7 = ((float)localPlayer.statLife - (float)(i - 1) * UIDisplay_LifePerHeart) / UIDisplay_LifePerHeart;
++					double num7 = ((float)snapshot.Life - (float)(i - 1) * UIDisplay_LifePerHeart) / UIDisplay_LifePerHeart;
+ 					num5 = (int)(30f + 225f * num7);
+ 					if (num5 < 30)
+ 						num5 = 30;
+ 
+-					num6 = num7 / 4f + 0.75f;
++					num6 = (float)(num7 / 4f + 0.75f);
```
Chicken-Bones: this is a lot of difficult patches to maintain :|

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
```diff
@@ -0,0 +1,75 @@
...
++			num += num4 / num2;
+ 			LifeFruitCount = num3;
++			/*
+ 			LifePerSegment = num;
+ 			ManaPerSegment = 20f;
++			*/
++			numLifeHearts = (int)(LifeMax / num);
++			numManaStars = (int)(ManaMax / 20d);
++
++			SystemLoader.ModifyStatSnapshot(player, ref this);
```
Chicken-Bones: Please remove `ModifyStatSnapshot`, we don't have good use-cases for it yet, and forcing the use of stat snapshot everywhere increases maintenance overhead

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
```diff
@@ -0,0 +1,61 @@
...
++				position = drawContext.position;
++				Vector2 origin = drawContext.origin;
++				Rectangle rectangle = drawContext.source ?? texture.Frame();
+ 				if (sourceRect.HasValue)
+ 					rectangle = sourceRect.Value;
+ 
++				/*
+ 				Vector2 position = topLeftAnchor + drawOffset;
+ 				Vector2 origin = OffsetSpriteAnchor + rectangle.Size() * OffsetSpriteAnchorByTexturePercentile;
++				*/
```
Chicken-Bones: why not move the `DrawResource` hook below these lines, rather than commenting them out.

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
```diff
@@ -0,0 +1,61 @@
...
++					origin = OffsetSpriteAnchor + frame.Size() * OffsetSpriteAnchorByTexturePercentile,
++					scale = new Vector2(drawScale),
++					SpriteBatch = spriteBatch
++				};
++
++				ResourceOverlayLoader.DrawResource(drawContext);
++
++				position = drawContext.position;
++				Vector2 origin = drawContext.origin;
++				Rectangle rectangle = drawContext.source ?? texture.Frame();
```
Chicken-Bones: `drawContext` is not passed by `ref` so there's no way for a mod to alter it, if that was your intent

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
```diff
@@ -0,0 +1,61 @@
...
+ 
+ namespace Terraria.GameContent.UI.ResourceSets
+ {
+@@ -16,6 +_,9 @@
+ 		public Vector2 OffsetPerDrawByTexturePercentile;
+ 		public Vector2 OffsetSpriteAnchor;
+ 		public Vector2 OffsetSpriteAnchorByTexturePercentile;
++		public PlayerStatsSnapshot StatsSnapshot;
++		public IPlayerResourcesDisplaySet DisplaySet;
++		public int ResourceSetSlot;
```
Chicken-Bones: `internal`?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4703,6 +4724,82 @@
...
++
+ 		private void ItemCheck_UseManaCrystal(Item sItem) {
+-			if (sItem.type == 109 && itemAnimation > 0 && statManaMax < 200 && ItemTimeIsZero) {
++			if (sItem.type == 109 && itemAnimation > 0 && ConsumedManaCrystals < ManaCrystalMax && ItemTimeIsZero) {
+ 				ApplyItemTime(sItem);
+-				statManaMax += 20;
+-				statManaMax2 += 20;
+-				statMana += 20;
+-				if (Main.myPlayer == whoAmI)
+-					ManaEffect(20);
```
Chicken-Bones: comment out please

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5827,6 +5928,31 @@
...
+ 							player.statMana = binaryReader.ReadInt32();
+ 							player.statManaMax = binaryReader.ReadInt32();
++							// Allow mana stat to exceed vanilla bounds (#HealthManaAPI)
++							/*
+ 							if (player.statManaMax > 200)
+ 								player.statManaMax = 200;
+ 
+ 							if (player.statMana > 400)
+ 								player.statMana = 400;
++							*/
```
Chicken-Bones, line 5957: Only one of these you need to comment out is the `statMana`, as the saved values here are vanilla max values anyway, so no problem with capping them to vanilla

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5690,10 +5787,14 @@
...
  				FileUtilities.Copy(path, path + ".bak", isCloudSave);
  
-@@ -38074,13 +_,20 @@
+@@ -38074,13 +_,24 @@
  			using (Stream stream = isCloudSave ? ((Stream)new MemoryStream(2000)) : ((Stream)new FileStream(path, FileMode.Create))) {
  				using (CryptoStream cryptoStream = new CryptoStream(stream, rijndaelManaged.CreateEncryptor(ENCRYPTION_KEY, ENCRYPTION_KEY), CryptoStreamMode.Write)) {
  					using (BinaryWriter binaryWriter = new BinaryWriter(cryptoStream)) {
++						// Keep the player state valid with vanilla (#HealthManaAPI)
++						// Any modifications to statLifeMax/statManaMax by mods can just be done during runtime
++						PlayerLoader.ResetMaxStatsToVanilla(player);
```
Chicken-Bones: This will leave the max stat values incorrect for the rest of the frame, better to just change the saved values in the save function

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -2464,7 +2464,16 @@
...
-@@ -13116,11 +_,19 @@
+@@ -13110,17 +_,28 @@
+ 			gravControl = false;
+ 			honeyCombItem = null;
+ 			gravControl2 = false;
++			PlayerLoader.ModifyMaxStats(this);
++			/*
+ 			statLifeMax2 = statLifeMax;
+ 			statManaMax2 = statManaMax;
++			*/
```
Chicken-Bones: No need to comment out these two lines.

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
```diff
@@ -78,6 +78,44 @@ public static class PlayerLoader
...
+
+				cumulativeHealth = cumulativeHealth.CombineWith(health);
+				cumulativeMana = cumulativeMana.CombineWith(mana);
+			}
+
+			player.statLifeMax = (int)cumulativeHealth.ApplyTo(player.statLifeMax);
+			player.statManaMax = (int)cumulativeMana.ApplyTo(player.statManaMax);
+
+			player.statLifeMax2 = player.statLifeMax;
+			player.statManaMax2 = player.statManaMax;
```
Chicken-Bones: Is there a good reason to have this in here, rather than just in `ResetEffects` like vanilla?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5827,6 +5928,31 @@
...
++							// Allow mana stat to exceed vanilla bounds (#HealthManaAPI)
++							/*
+ 							if (player.statManaMax > 200)
+ 								player.statManaMax = 200;
+ 
+ 							if (player.statMana > 400)
+ 								player.statMana = 400;
++							*/
++
++							PlayerLoader.ModifyMaxStats(player);
```
Chicken-Bones: why this call? `ModPlayer` data hasn't been loaded yet at this point

patches/tModLoader/Terraria/ModLoader/SystemLoader.cs
```diff
@@ -334,6 +335,12 @@ public static partial class SystemLoader
 			}
 		}
 
+		public static void ModifyStatSnapshot(Player player, ref PlayerStatsSnapshot snapshot) {
+			foreach (var system in HookModifyStatSnapshot.arr) {
+				system.ModifyStatSnapshot(player, ref snapshot);
+			}
+		}
```
Chicken-Bones: Remove hook, no good use case yet

patches/tModLoader/Terraria/Player.cs.patch
Chicken-Bones: In fact, this could mess with the vanilla fallback `ConsumedX` calculations if a mod just, for eg, adds a blanket 100hp.

Why not set the `ConsumedX` properties here, instead of using tML NBT data at all, since you'll have the vanilla data in all cases as that's what was saved.

Chicken-Bones (changes requested):

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
absoluteAquarian: otherwise, the bottom bar would clip into the top bar, which wouldn't be consistent with how the original display that this was "cloned" from works

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
absoluteAquarian: fair enough.  you had wanted me to use `.Name` beforehand, but this is likely to be more reliable

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
absoluteAquarian: they're included in case a modder calls `PlayerLoader.ModifyMaxStats()` for whatever reason

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
absoluteAquarian: if it would make sense for the life bar to still draw below the mana bar, then i'll change this to reflect that

patches/tModLoader/Terraria/Player.cs.patch
absoluteAquarian: good point

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
absoluteAquarian: as in, start with the rightmost bar and draw to the left instead of vice versa (the current logic)?

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
absoluteAquarian: possibly, but the only case where the "max resource number" isn't 20 is only the Bars panels

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
absoluteAquarian: if formerly was, but i forgot to alter this code after making it not be passed by `ref`

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ResourceDrawSettings.cs.patch
absoluteAquarian: ResourceDrawSettings is public, so if those fields were made internal, modders would be blocked from using the struct properly without resorting to Reflection

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ClassicPlayerResourcesDisplaySet.cs.patch
absoluteAquarian: `num6` appears to just be used for filling scale due to the "pulse" effect, so it's probably fine to just have `num7` be a `float` variable instead of `double`
same idea for the mana drawing method

ExampleMod/Common/UI/ResourceDisplay/VanillaLifeOverlay.cs
absoluteAquarian: good point.  i'll think it over

ExampleMod/ExampleMod.Networking.cs
absoluteAquarian: should a call to SyncPlayer or the like be done here, or should the comment just be removed?

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
Chicken-Bones: I would rather not have this, better to follow vanilla more closely, and only add this if a modder requests it

patches/tModLoader/Terraria/ModLoader/ModSystem.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using Terraria.GameContent.UI.ResourceSets;
```
Chicken-Bones: Unused import now

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
Chicken-Bones: I just.. can't conscience an example where `DrawLife` doesn't draw life :|

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+		// This field is used to cache vanilla assets used in the CompareAssets helper method further down in this file
+		private Dictionary<string, Asset<Texture2D>> vanillaAssetCache;
+
+		public override void Load() {
+			vanillaAssetCache = new();
+		}
+
+		public override void Unload() {
+			vanillaAssetCache = null;
+		}
```
Chicken-Bones: No need to do this, as it'll get garbage collected fine, the field isn't `static`

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace ExampleMod.Common.UI.ResourceOverlay
+{
+	public class VanillaLifeOverlay : ModResourceOverlay
+	{
+		// This field is used to cache vanilla assets used in the CompareAssets helper method further down in this file
+		private Dictionary<string, Asset<Texture2D>> vanillaAssetCache;
+
+		public override void Load() {
+			vanillaAssetCache = new();
+		}
```
Chicken-Bones: toss that in the field declaration

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+		public override void PostDrawResource(ResourceOverlayDrawContext context) {
+			Asset<Texture2D> asset = context.texture;
+
+			string fancyFolder = "Images/UI/PlayerResourceSets/FancyClassic/";
+			string barsFolder = "Images/UI/PlayerResourceSets/HorizontalBars/";
+
+			bool drawingBarsPanels = CompareAssets(asset, barsFolder + "HP_Panel_Middle");
+
+			// Bars panel drawing has two additional elements, so the "resource index" needs to be offset
+			int fillResourceNumber = drawingBarsPanels ? context.resourceNumber - 1 : context.resourceNumber;
```
Chicken-Bones: This needs to be provided by the `ResourceOverlayDrawContext`, can you showcase using `ResourceStat` instead?

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+				positionOffset = new Vector2(0, 4);
+			} else {
+				// Final panel in the first row
+				// Vanilla texture is "Heart_Right"
+				positionOffset = new Vector2(0, 4);
+			}
+
+			// "context" contains information used to draw the resource
+			// If you want to draw directly on top of the vanilla hearts, just replace the texture and have the context draw the new texture
+			context.texture = ModContent.Request<Texture2D>(texture, AssetRequestMode.ImmediateLoad);
```
Chicken-Bones: This one should be an `Asset<Texture2D>` field. 
Recommend using `??=`
No need for `ImmediateLoad` (if it doesn't draw for a frame or two, no big loss)

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+			// Draw over the Fancy heart panels
+			string texture = "ExampleMod/Common/UI/ResourceOverlay/FancyLifeOverlay_Panel";
+			// The original position refers to the entire panel slice.
+			// However, since this overlay only modifies the "inner" portion of the slice (aka the part behind the heart),
+			// the position should be modified to compensate for the sprite size difference
+			Vector2 positionOffset;
+
+			if (context.resourceNumber == context.snapshot.AmountOfLifeHearts) {
+				// Final panel to draw has a special "Fancy" variant.  Determine whether it has panels to the left of it
+				if (context.resourceNumber % 10 == 1) {
```
Chicken-Bones: You should check texture name here, rather than `resourceNumber` since that's what the framing info/position actually relates to, and then you wouldn't need to comment `// Vanilla texture is "Heart_Single_Fancy"` for eg

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+
+		private static void DrawClassicFancyOverlay(ResourceOverlayDrawContext context) {
+			// Draw over the Classic / Fancy hearts
+			// "context" contains information used to draw the resource
+			// If you want to draw directly on top of the vanilla hearts, just replace the texture and have the context draw the new texture
+			context.texture = ModContent.Request<Texture2D>("ExampleMod/Common/UI/ResourceOverlay/ClassicLifeOverlay", AssetRequestMode.ImmediateLoad);
+			context.Draw();
+		}
+
+		private static void DrawFancyPanelOverlay(ResourceOverlayDrawContext context) {
```
Chicken-Bones: Need a comment saying that this is not required, and that the example is of changing the background of the panel when the resource is low.

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,140 @@
...
+			context.Draw();
+		}
+
+		private static void DrawBarsPanelOverlay(ResourceOverlayDrawContext context) {
+			// Draw over the Bars middle life panels
+			// "context" contains information used to draw the resource
+			// If you want to draw directly on top of the vanilla bar panels, just replace the texture and have the context draw the new texture
+			context.texture = ModContent.Request<Texture2D>("ExampleMod/Common/UI/ResourceOverlay/BarsLifeOverlay_Panel", AssetRequestMode.ImmediateLoad);
+			// Due to the replacement texture and the vanilla texture having different heights, the source needs to also be modified
+			context.source = (Rectangle)context.source with { Height = context.texture.Height() };
```
Chicken-Bones: Can you use `context.source = context.texture.Frame();` here?

ExampleMod/ExampleMod.Networking.cs
Chicken-Bones: Please test this in multiplayer. I believe you need to forward the packet to all clients except the one it came from if `Main.netMode == NetModeID.Server`

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerResourceSetsManager.cs.patch
```diff
@@ -0,0 +1,48 @@
...
++	public partial class PlayerResourceSetsManager
+ 	{
+ 		private Dictionary<string, IPlayerResourcesDisplaySet> _sets = new Dictionary<string, IPlayerResourcesDisplaySet>();
+ 		private IPlayerResourcesDisplaySet _activeSet;
+@@ -23,7 +_,7 @@
+ 		}
+ 
+ 		private void Configuration_OnLoad(Preferences obj) {
+-			_activeSetConfigKey = Main.Configuration.Get("PlayerResourcesSet", "New");
++			_activeSetConfigKeyOriginal = _activeSetConfigKey = Main.Configuration.Get("PlayerResourcesSet", "New");
```
Chicken-Bones: Put `_activeSetConfigKeyOriginal = _activeSetConfigKey ` on a new line

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerResourceSetsManager.cs.patch
```diff
@@ -0,0 +1,48 @@
...
+ 			if (_loadedContent)
+ 				SetActiveFromLoadedConfigKey();
+ 		}
+@@ -35,8 +_,8 @@
+ 
+ 		public void LoadContent(AssetRequestMode mode) {
+ 			_sets["New"] = new FancyClassicPlayerResourcesDisplaySet("New", "New", "FancyClassic", mode);
+-			_sets["Default"] = new ClassicPlayerResourcesDisplaySet("Default", "Default");
+ 			_sets["HorizontalBars"] = new HorizontalBarsPlayerReosurcesDisplaySet("HorizontalBars", "HorizontalBars", "HorizontalBars", mode);
++			_sets["Default"] = new ClassicPlayerResourcesDisplaySet("Default", "Default");
```
Chicken-Bones: why re-order?

Chicken-Bones: The values of a dictionary aren't ordered anyway.

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
```diff
@@ -0,0 +1,73 @@
...
+ 		public int LifeMax;
+ 		public int LifeFruitCount;
+-		public float LifePerSegment;
++		// Made double to prevent bugs caused by precision limits (#HealthManaAPI)
++		public double LifePerSegment => LifeMax / (double)AmountOfLifeHearts;
+ 		public int Mana;
+ 		public int ManaMax;
+-		public float ManaPerSegment;
++		// Made double to prevent bugs caused by precision limits (#HealthManaAPI)
++		public double ManaPerSegment => ManaMax / (double)AmountOfManaStars;
```
Chicken-Bones: what was wrong with `float` here? Can you give an example?

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/HorizontalBarsPlayerReosurcesDisplaySet.cs.patch
```diff
@@ -0,0 +1,177 @@
...
++		public static void FillBarByValues(int elementIndex, Asset<Texture2D> sprite, int segmentsCount, float fillPercent, out Vector2 offset, out float drawScale, out Rectangle? sourceRect) {
+ 			sourceRect = null;
+ 			offset = Vector2.Zero;
+ 			float num = 1f;
+ 			float num2 = 1f / (float)segmentsCount;
+-			float t = 1f - fillPercent;
++			float t = fillPercent;
+ 			float lerpValue = Utils.GetLerpValue(num2 * (float)elementIndex, num2 * (float)(elementIndex + 1), t, clamped: true);
+-			num = 1f - lerpValue;
++			num = lerpValue;
```
Chicken-Bones: Please comment out the vanilla code here rather than removing it

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ClassicPlayerResourcesDisplaySet.cs.patch
```diff
@@ -0,0 +1,226 @@
...
+-			if (mouseScreen.X > (float)(500 + UI_ScreenAnchorX) && mouseScreen.X < (float)(500 + num + UI_ScreenAnchorX) && mouseScreen.Y > 32f && mouseScreen.Y < (float)(32 + TextureAssets.Heart.Height() + num2))
++			if (mouseScreen.X > (float)(500 + UI_ScreenAnchorX) && mouseScreen.X < (float)(500 + num + UI_ScreenAnchorX) && mouseScreen.Y > 32f && mouseScreen.Y < (float)(32 + TextureAssets.Heart.Height() + num2) && ResourceOverlayLoader.DisplayHoverText(snapshot, this, true))
+ 				CommonResourceBarMethods.DrawLifeMouseOver();
+ 
+ 			num = 24;
+-			num2 = 28 * localPlayer.statManaMax2 / UIDisplay_ManaPerStar;
+-			if (mouseScreen.X > (float)(762 + UI_ScreenAnchorX) && mouseScreen.X < (float)(762 + num + UI_ScreenAnchorX) && mouseScreen.Y > 30f && mouseScreen.Y < (float)(30 + num2))
++			num2 = 28 * snapshot.AmountOfManaStars;
++			if (mouseScreen.X > (float)(762 + UI_ScreenAnchorX) && mouseScreen.X < (float)(762 + num + UI_ScreenAnchorX) && mouseScreen.Y > 30f && mouseScreen.Y < (float)(30 + num2) && ResourceOverlayLoader.DisplayHoverText(snapshot, this, false))
+ 				CommonResourceBarMethods.DrawManaMouseOver();
```
Chicken-Bones, line 191: Is there any way we can reduce the patch size of this file? There's so much here that could break in an update

Chicken-Bones: I guess you don't need to use `snapshot` everywhere now that the hook is removed?

Chicken-Bones (changes requested):

ExampleMod/Common/UI/ExampleDisplaySets/ExampleReversedBarsDisplay.cs
absoluteAquarian: fair enough.  i'll swap the two method bodies

ExampleMod/ExampleMod.Networking.cs
absoluteAquarian: i look again in the code and SyncPlayer is only called automatically in `NetMessage.SyncOnePlayer()` and in MessageBuffer for `MessageID.PlayerInfo` and `MessageID.PlayerSpawn`

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
absoluteAquarian: 1) fair point.  this is a leftover from the old code that had multiple textures
2) this line replaces the texture variable (which won't be reflected outside of the method) then draws the new texture, hence `=` instead of `??=`

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
absoluteAquarian: oops, yes, this was mistakenly copied from when i was working on the bar filling drawing

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerResourceSetsManager.cs.patch
absoluteAquarian: they're iterated through in the order they're added at when using FirstOrDefault

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
absoluteAquarian: Let's take the example I used in the Discord thread:
1) `LifeMax` is set to `290` and `AmountOfLifeHearts` is `15` (e.g. a mod adds a "negative heart" which removes 10 max life)
2) `290 / 15d = 19.333333333333332d` and `290 / 15f = 19.333334f`
3) code that then tries to get the amount of hearts from `LifeMax / AmountOfLifeHearts` or `LifePerHeart` for whatever reason ends up with `15.000000000000002d` -- which gets correctly truncated to `15` -- instead of `14.999999f` -- which gets incorrectly truncated to `14`

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/ClassicPlayerResourcesDisplaySet.cs.patch
absoluteAquarian: yeah, i could safely revert the `snapshot` changes in the Classic display set due to the hook not being present anymore

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
absoluteAquarian: this is the method responsible for drawing the heart backgrounds

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
Chicken-Bones:
```cs
// field
Asset<Texture2D> panelTexture;

// in method later
context.texture = panelTexture ??= ModContent.Request<Texture2D>(texture);
```

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
Chicken-Bones: Yes, most people won't need to do custom backgrounds, just fore-grounds.

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
absoluteAquarian: ah, that's what you meant

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
absoluteAquarian: that's what i meant by "background" -- the part behind the hearts, stars and bar fillings

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerResourceSetsManager.cs.patch
Chicken-Bones: Well I looked at the reference source, and apparently it's true for this version of .NET

I don't like relying on it though, it's confusing to other developers who follow the typical advice (that dictionaries and hashmaps are unordered)

From MSDN
image

absoluteAquarian: fair point, though that's what the vanilla logic uses.  i should mention that `Dictionary<K, V>.Values` is what's being iterated

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
Chicken-Bones: This issue is not exclusive to `float` over double. This can happen any time you try to do `v = a / b` and then expect  `a / v == b`

```cs
for (int i = 200; i < 400; i++) { 
     double d = i / 15d;
     int k = (int)(i / d);
     if (k != 15) Console.WriteLine(i);
}
```
```
> 295/15d
19.666666666666668
> 295/(295/15d)
14.999999999999998
```

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerResourceSetsManager.cs.patch
Chicken-Bones: Yeah, the version of `.NET` we're using stores all the entries in an `Entry[]` in insertion order (at least when you don't remove and re-add stuff)

patches/tModLoader/Terraria/GameContent/UI/ResourceSets/PlayerStatsSnapshot.cs.patch
absoluteAquarian: right.  i'll revert the `float`->`double` changes

ExampleMod/Common/UI/ResourceOverlay/VanillaLifeOverlay.cs
```diff
@@ -0,0 +1,138 @@
...
+		public override void PostDrawResource(ResourceOverlayDrawContext context) {
+			Asset<Texture2D> asset = context.texture;
+
+			string fancyFolder = "Images/UI/PlayerResourceSets/FancyClassic/";
+			string barsFolder = "Images/UI/PlayerResourceSets/HorizontalBars/";
+
+			bool drawingBarsPanels = CompareAssets(asset, barsFolder + "HP_Panel_Middle");
+
+			// Bars panel drawing has two additional elements, so the "resource index" needs to be offset
+			int fillResourceNumber = drawingBarsPanels ? context.resourceNumber - 1 : context.resourceNumber;
```
Chicken-Bones: Alright, all we need is for this value to be provided by the context, rather than calculated here.

Chicken-Bones (changes requested):
Incredibly, I think this is the only change still required

PR #2910, opened 2022-09-09
Include Dotnet 64 win maybe??? Might decrease remaining reports of not opening

Solxanich:
This PR adds an included win64 6.0.0 dotnet zip that can be extracted without having to download it from the internet.

This should reduce teh number of issues with game not launching on Steam Windows, but the actual impact remains to be seen as to how much.

The github zips will have the dotnet separated, while steam will include it upfront.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -190,8 +190,12 @@ jobs:
             exit $err
           )
           
+          mkdir -p artifacts/dotnet
+          mv steam_build/LaunchUtils/dotnet-runtime-6.0.0-win-x64.zip artifacts/dotnet/dotnet-runtime-6.0.0-win-x64.zip
+          
           mkdir -p artifacts/Release
           cp -r steam_build/. artifacts/Release
+          
           echo ::endgroup::
           
           echo Cleaning up output directory in-between builds...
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Can we do this just for steam?

JavidPack:
This could rule out network issues like firewall or slow internet, I say lets do it for sure. It shouldn't be too hard to remove the file from the github release artifact.

Solxanich:
> Can we do this just for steam?

Why just for Steam?
Is there a reason we don't think GoG might need it under same failure mode as Steam?

If the ask is more so to delete the zip after it would have been deployed, thats probably more pragmatic change

Chicken-Bones:
> > Can we do this just for steam?
> 
> Why just for Steam? Is there a reason we don't think GoG might need it under same failure mode as Steam?
> 
> If the ask is more so to delete the zip after it would have been deployed, thats probably more pragmatic change

Mostly due to artifact file size. It's already pretty big :|

Solxanich:
BEcause this has yaml changes, can't test without merging. Will be merging soon

Chicken-Bones:
I really wanted a solution which doesn't add 30MB to our github repo size every time we change .NET, and doesn't add 30MB to every single build artifact, but I guess it's too late now. We can always change it later

Solxanich:
> I really wanted a solution which doesn't add 30MB to our github repo size every time we change .NET, and doesn't add 30MB to every single build artifact, but I guess it's too late now. We can always change it later

To be clear, I did separate out dotnet zip from the build outputs on artifacts - two 100 mb outputs and a common 30 mb output. 

You never voiced wanting our specific version of dotnet runtime to be fetched from outside of repo, so it doesn't happen unless mentioned. 

On topic, it is relatively small to change at this juncture, as the 'deploy' action is reasonably contained to whatever provides it.  Just wasn't testable for the github action without moving on to 1.4 at this point, regardless

Chicken-Bones:
I did miss the changes to the artifact. My initial: "Can we do this just for steam?" is still preferred. It's nice to check with other collaborators who have raised concerns on an issue before merging. I understand you couldn't test this without having it on 1.4 so I can see the urgency.

Chicken-Bones:
I'll ask jopo (or yourself) to prioritise a 'just for steam' solution this week.

Issue #2917, opened 2022-09-11
IO save more fine-grained


zzp198-old:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

tModLoader code as a Contributor/Maintainer

### Description

Currently, when saving mod data, if there is an error in saving mod data, all mod data will be lost.

### What does this proposal attempt to solve or improve?

Catch errors so that they do not affect the saving of other mod data.

Chicken-Bones:
Follow proposal #2935 for planned fix

PR #2919, opened 2022-09-11
[1.4] Add try catch guards when saving data

ExterminatorX99:
### What is the bug?
There is no handling of exceptions that occur during the saving of mod data.
The result of this is that no data at all is saved, even when most of the data is absolutely fine.
It is also possible for these exceptions to completely crash tML.
#2917 

### How did you fix the bug?
Add try catch and logging

#2920


Chicken-Bones:
This solution has a problem of silent data loss. Please consider contributing to https://github.com/tModLoader/tModLoader/issues/2935 instead

PR #2920, opened 2022-09-11
Fix #2917 

### LINKED ISSUE #2917: IO save more fine-grained

zzp198-old:
When PlayerIO/ItemIO throws an exception, only the data of the current mod / current item will be lost.

Test results: show that part of the data before the error is reported can be read and will not crash
%BF@1$TJ QUEYDE)@A4LEOH


Chicken-Bones:
This solution has a problem of silent data loss. Please consider contributing to https://github.com/tModLoader/tModLoader/issues/2935 instead

PR #2933, opened 2022-09-15
Update MonoMod to new Re-organized implementation.

Chicken-Bones:
### What is the new feature?

MonoMod Update:
- `NativeDetour` and `Detour` are gone, there was little need for modders to use these anyway. 
  - An alternative to `NativeDetour` based on an actual disassembler (iced) is in the works
- `MonoModHooks.RequestNativeAccess` is no longer needed, mods are free to use `new Hook` or `HookEndpointManager.Add`
- `Hook` will now undo itself on garbage collection, so if you're using this instead of `HookEndpointManager` for some reason, you will need to save them in static variables to prevent them getting garbage collected and undone.

Backwards compatible with many mods already (I've tested Calamity, RecipeBrowser, CheatSheet, MagicStorage)

Completely rewritten backend, more stable and performant, will be the future for ongoing support (eg Apple M1)

Nope

### ExampleMod updates
No changes!

### Porting Notes
- Remove any uses of `new NativeDetour` or `new Detour` from your code
  - Replace them with `new Hook` or `HookEndpointManager.Add`
- Make sure you store the result of `new Hook` in a static variable, to prevent it being garbage collected and undone immediately


Chicken-Bones:
Needs some more testing of major mod breakage before I'm happy to merge. Waiting for decision from MonoMod authors on whether/when `Detour` is going away

Issue #2403, opened 2022-05-21
[1.4] Non-Visible Constructor For `Mod` Type Throws Misleading Exception


steviegt6:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

If you make your mod's default constructor non-visibile (not `public`), the "no parameterless constructor" exception is thrown.

For example:
```cs
protected YourMod() { }
```

### Fixes

Either allow non-visible constructors, differentiate between the exceptions, or append an extra note to the exception.

Issue #2332, opened 2022-05-02
[1.4] [Feature Request] Update Mod Type Resolution


steviegt6:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

tModLoader only allows mod assemblies to contain one type which derives from `Terraria.ModLoader.Mod`.[^1]

[^1]: https://github.com/tModLoader/tModLoader/blob/d6c564c66d81fe7b1d5fac3e714d455501f3b4d6/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs#L178

### What does this proposal attempt to solve or improve?

While this is _normally_ fine, it poses an issue if you were to do something like offer an abstraction on top of `Terraria.ModLoader.Mod`. I develop and maintain a library where the default implementation of library features reside within a mod class. These default implementations should ideally be made available to users of the library.
You could argue that these should be delegated to `ModSystem`s or something similar, but a direct association is cleanest.

A demonstration of the issue as as follows:
```cs
public interface ILibraryMod
{
		ILibraryFeature Feature { get; }
}

// Used by both the Library AND serves as a *default implementation* for mods that wish to use the library.
public class LibraryMod : Mod, ILibraryMod
{
		public ILibraryFeature Feature { get; } = new DefaultLibraryFeature();
}
```
Here, you may observe how both a library and mods that use the library have a reason to use some sort of default implementation to reduce boilerplate. This is dangerous because you cannot have an abstract `Mod`-extending class that implements this behavior.

Here is how I currently solve the issue:
```cs
public class LibraryMod : Mod, ILibraryMod
{
		public ILibraryFeature Feature { get; } = new DefaultLibraryFeature();

		public override void Load()
		{
				ExecutePrivately(() => { /* some library-specific task */ });
		}

		private bool ExecutePrivately(Action task)
		{
				if (!GetType().Assembly.FullName?.StartsWith("LibraryName, ") ?? false)
						return false;

				task();
				return true;
		}
}
```
I have to wrap library-specific logic in a hacky method that checks if the type belongs to my library.
This entire hassle could be avoided if we were given the ability to define abstract `Mod` types.

### Which (other) solutions should be considered?

N/A

steviegt6:
Some internal discussions have made me realize this may require a rewording to make my reasoning for this change clear. Some questions have been brought up regarding the necessity of this implementation. I'll be making some edits tomorrow to reflect these developments.

[...]

PR #2937, opened 2022-09-16
[1.4] Allow usage of non public constructors

ExterminatorX99:
### What is the bug?
#2403 

### How did you fix the bug?
Allow non public constructors

Custom exception message saying we only support public constructors

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
@@ -273,7 +273,7 @@ namespace Terraria.ModLoader.Config
 		public static object AlternateCreateInstance(Type type) {
 			if (type == typeof(string))
 				return "";
-			return Activator.CreateInstance(type);
+			return Activator.CreateInstance(type, true);
 		}
 
 		// Gets an Attribute from a property or field. Attribute defined on Member has highest priority,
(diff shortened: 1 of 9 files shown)


JavidPack:
Reminder: We might need to also address #2332 (latest comments) at the same time.

Chicken-Bones:
We would like the alternative "custom error message saying that the default constructor must be public". We do this to reduce confusion about auto-loading. A `public` constructor is 'meant for public (outside) use', a `private` one may be assumed to only be used within the class, and autoloading comes in and calls it causing confusion. 

I see no benefit to allowing autoloading via non-public constructors.

direwolf420:
Just wanted to throw this in:
https://github.com/tModLoader/tModLoader/blob/ada8c2396bad92fd51abc37337fee1613037e1c0/patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs#L55-L64

It currently already uses `nonPublic` for `Activator.CreateInstance` here (which is valid as the check above also includes `NonPublic` flag)

Chicken-Bones:
@direwolf420 thanks for the update, ugh, regret... Something for the 1.4.4 breaking change list? or do we just say whatever

JavidPack:
I've kept many things internal in my mods so that other mods referencing it only see the public API that I provide. I think tmodloader accessing them is expected of a mod loader, but the access modifiers in my mods are for cross mod and in mod considerations only. Having to cater to tmodloader in this way would expose other classes I don't want mods referencing my mod to have to sift through. The Mod class could be special though, since it is the main way to access many things you'll need in a cross mod situation. I don't mind the apparent contradiction with that.

Issue #2408, opened 2022-05-21
[1.4] [Feature Proposal] Mod/GlobalWall.WallFrame hook


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

An analog of `Mod/GlobalTile.TileFrame` but for walls.

### What does this proposal attempt to solve or improve?

Allow for more custom framing for walls.

PR #2938, opened 2022-09-16
[1.4] `Mod/GlobalWall.WallFrame` hook

ExterminatorX99:
### What is the new feature?
`Mod/GlobalWall.WallFrame` hook #2408

Allow for more custom framing for walls #2408

No


patches/tModLoader/Terraria/ModLoader/GlobalWall.cs
```diff
@@ -24,5 +24,17 @@ public abstract class GlobalWall : GlobalBlockType
...
+
+		/// <summary>
+		/// Called for every wall that updates due to being placed or being next to a wall that is changed. Return false to stop the game from carrying out its default WallFrame operations. Returns true by default.
+		/// </summary>
+		/// <param name="i">The x position in tile coordinates.</param>
+		/// <param name="j">The y position in tile coordinates.</param>
+		/// <param name="type">Type of the wall being framed</param>
+		/// <param name="resetFrame">Whether this wall should try to change it's display frame</param>
+		/// <returns></returns>
+		public virtual bool WallFrame(int i, int j, int type, ref bool resetFrame) {
```
Chicken-Bones: I don't think `resetFrame` should be `ref`

The documentation for `resetFrame` should say "randomize frame" since that's what it's actually for.

Chicken-Bones (changes requested):
Needs two examples:
- Custom glass blocks. Wall framing needs to act as if the glass block has a wall behind it even if it doesn't. See `tile2.type == 54`
- Wall style randomization (perhaps in ExampleWall) when resetFrame is true

JavidPack:
I have some changes in mind for this.

ExampleMod/Common/GlobalWalls/ExampleGlobalWall.cs
```diff
@@ -0,0 +1,36 @@
...
+			/*
+			if (style < 15) {
+				if (Main.tile[i, j - 1].HasTile && Main.tile[i, j - 1].TileType == ModContent.TileType<ExampleGlassBlock>())
+					style |= 1;
+				if (Main.tile[i - 1, j].HasTile && Main.tile[i - 1, j].TileType == ModContent.TileType<ExampleGlassBlock>())
+					style |= 2;
+				if (Main.tile[i + 1, j].HasTile && Main.tile[i + 1, j].TileType == ModContent.TileType<ExampleGlassBlock>())
+					style |= 4;
+				if (Main.tile[i, j + 1].HasTile && Main.tile[i, j + 1].TileType == ModContent.TileType<ExampleGlassBlock>())
+					style |= 8;
```
Chicken-Bones: I would much prefer we just implement this via a `TileID.Sets` entry

patches/tModLoader/Terraria/Framing.cs.patch
```diff
@@ -8,6 +8,24 @@
...
++		public static Point16[][] wallFrameLookup;
+ 		private static Point16 frameSize8Way;
+ 		private static Point16 wallFrameSize;
+ 		private static BlockStyle[] blockStyleLookup;
+-		private static int[][] phlebasTileFrameNumberLookup;
++		public static int[][] phlebasTileFrameNumberLookup;
+-		private static int[][] lazureTileFrameNumberLookup;
++		public static int[][] lazureTileFrameNumberLookup;
+-		private static int[][] centerWallFrameLookup;
++		public static int[][] centerWallFrameLookup;
```
Chicken-Bones: do these lookups need to be made public?

patches/tModLoader/Terraria/Framing.cs.patch
```diff
@@ -17,3 +35,33 @@
...
+-
+-				tile.wallFrameNumber((byte)num2);
+ 			}
+ 			else {
+ 				num2 = tile.wallFrameNumber();
+@@ -370,6 +_,10 @@
+ 			if (num == 15)
+ 				num += centerWallFrameLookup[i % 3][j % 3];
+ 
++			if (!WallLoader.WallFrame(i, j, tile.wall, resetFrame, ref num, ref num2))
```
Chicken-Bones, line 72: should this go before the `centerWallFrameLookup`?

patches/tModLoader/Terraria/ModLoader/GlobalWall.cs
```diff
@@ -24,5 +24,19 @@ public abstract class GlobalWall : GlobalBlockType
...
+		/// Called for every wall that updates due to being placed or being next to a wall that is changed. Return false to stop the game from carrying out its default WallFrame operations. Returns true by default.
+		/// </summary>
+		/// <param name="i">The x position in tile coordinates.</param>
+		/// <param name="j">The y position in tile coordinates.</param>
+		/// <param name="type">Type of the wall being framed</param>
+		/// <param name="resetFrame">True if the calling code intends that the frameNumber be randomly changed, such as when placing the wall initially or loading the world, but not when updating due to nearby tile or wall placements</param>
+		/// <param name="style">The style or orientation that will be applied</param>
+		/// <param name="frameNumber">The random style that will be applied</param>
+		/// <returns></returns>
+		public virtual bool WallFrame(int i, int j, int type, bool resetFrame, ref int style, ref int frameNumber) {
```
Chicken-Bones: `resetFrame` -> `randomizeFrame`

patches/tModLoader/Terraria/Framing.cs.patch
Chicken-Bones: I don't think the hook should return, because vanilla provides no way to early return on wall framing

Chicken-Bones (changes requested):
Examples look good

patches/tModLoader/Terraria/Framing.cs.patch
JavidPack: The return is so if a user is implementing their own lookup, they can avoid the vanilla lookup being applied. 

centerWallFrameLookup changes the style, so I figure we should provide the chosen style to the modder so they can act on it reliably.

patches/tModLoader/Terraria/Framing.cs.patch
JavidPack: centerWallFrameLookup at least, unless we implement the TileID.Sets above. The others I thought might be useful but don't need to be.

JavidPack:
Thanks @ExterminatorX99

Issue #2855, opened 2022-08-20
Retain menu theme after mod reload


CrezyDud:
### Version

Latest 1.4-stable

### OS

Linux

### Platform

GOG

### This bug affects

Gameplay as a Player

### Description

The memu theme resets when reloading mods (at least the 1.3 menu theme gets reset)
I like music more of 1.3, so it kinda annoying

### Additional Information

not lots to say here

CrezyDud:
most annyoing when compilings mods

PR #2940, opened 2022-09-16
[1.4] Fix mod menu always resetting on mod reload

ExterminatorX99:
### What is the bug?
#2855

### How did you fix the bug?
Only reset the mod menu if it's a modded one

Keep current behaviour


patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
```diff
@@ -144,11 +144,12 @@ public static class MenuLoader
...
 		internal static void Unload() {
-			currentMenu = MenutML; // Prevent asset disposed exceptions by disallowing modded menus during the unload process.
-
 			loading = true;
+			// Prevent asset disposed exceptions by disallowing modded menus during the unload process.
 			if (menus.IndexOf(currentMenu) >= DefaultMenuCount) {
-				switchToMenu = MenutML;
+				currentMenu.OnDeselected();
+				currentMenu = MenutML;
+				currentMenu.OnSelected();
```
Chicken-Bones: We should only change `currentMenu` on the main thread, that's why `switchToMenu` exists.

I suggest
```cs
switchToMenu = MenutML;
while (currentMenu != MenutML) Thread.Yield();
```

ExterminatorX99: Then why is the current behaviour to directly change `currentMenu`?

Chicken-Bones: Someone probably added that as a patch that didn't get proper review. selectedMenu is used in the condition

Issue #2902, opened 2022-09-06
JofairdenArmorShaderLayer.Draw Crash to Desktop


ameberlan:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Game will randomly crash to desktop while playing, throwing an error pertaining to shaders.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/9493669/client.log)

### Steps to reproduce

Can't persistently reproduce and don't have a guaranteed cause. Have a suspicion, and haven't been able to get it to re-occur since making this change (see additional information).

### Expected Behavior

Game should not be crashing and exiting to desktop.

### Actual Behavior

Game is crashing and force closing completely.

### Reproduction frequency

Happens at least once per play session (two to three hour span), and have had it happen as many as four or five times. Happens both with minimal mods, and none at all.

### Additional Information

Upon reviewing the crash information and the following line from the crash section in the client.log:

Terraria.ModLoader.Default.Developer.Jofairden.JofairdenArmorShaderLayer.Draw(PlayerDrawSet& drawInfo) in tModLoader\Terraria\ModLoader\Default\Developer\Jofairden\Layers\JofairdenArmorShaderLayer.cs:line 43

I requested my friend remove his vanity armor as he was wearing the andromedon set (Jofairden's developer set) and see if it would recur. I have not been able to get this to happen again since. Unsure what is causing shaders to throw issues. We're housed in a "manufactured" above ground mushroom biome so not sure if this is having any effect on the vanity shaders (likely not, but it may be relevant).

ExterminatorX99:
Replication steps:
1. Enter world and wear any Andromedon armor piece
2. Reload mods
3. Enter world and wear any Andromedon armor piece

PR #2944, opened 2022-09-17
[1.4] Fully unload Jofairden armor assets

ExterminatorX99:
### What is the bug?
#2902 

### How did you fix the bug?
Fully unload Jofairden armor assets

Change `??=` to `if` and add a disposed check


Chicken-Bones:
This fix is correct, but, in my opinion, a better fix would be to remove the `static` modifier from these fields. The layers are re-created each time the mod is reloaded, so new assets will be procured correctly

PR #2953, opened 2022-09-20
[1.4.4] Change return type of Mod.AddContent to bool, add docs

direwolf420:
### What is the new feature?
Makes `Mod.AddContent` return true if the load succeeded, false if not.
`GoreLoader.AddGoreFromTexture` was also adjusted to reflect this as it's just a wrapper around `AddContent`.
Adds docs to some methods that were missing (but exist for similar methods in `ModContent`)
Unifies existing docs regarding terminology used for referring to the registered template instance.

This helps when the modder is adding content manually and wants to populate something post-AddContent (without having to use `ILoadable.Load`). There is currently no feedback of if the content actually loaded or not, aside from manually calling `IsLoadingEnabled` on it (which may not be backwards compatible if tml adds more conditions into the method).

The documentation adjustments are there to avoid confusion about what the returned instance actually represents. I have seen modders sometimes do `ItemLoader.GetItem(item.type)` when they should be doing `item.ModItem` instead.

This is pretty straightforward, many `AddX` methods all around the .NET codebase already have this behavior, such as `HashSet<T>.Add`.

```cs
if (Mod.AddContent(instance)) {
    someGeneralList.Add(instance);
}
```

### ExampleMod updates
Not required, EM doesn't use this

### Porting Notes
Recompile is required (but since this targets 1.4.4 it doesn't matter)


patches/tModLoader/Terraria/ModLoader/Mod.cs
```diff
@@ -125,21 +125,31 @@ public void AddConfig(string name, ModConfig mc)
...
+		public bool AddContent<T>() where T : ILoadable, new() => AddContent(new T());
+
+		/// <summary> Call this to manually add the given content instance to the game. </summary>
+		/// <param name="instance"> The content instance to add </param>
+		/// <returns> true if the instance was successfully added </returns>
+		public bool AddContent(ILoadable instance) {
 			if (!loading)
 				throw new Exception(Language.GetTextValue("tModLoader.LoadErrorNotLoading"));
 
 			if (instance.IsLoadingEnabled(this)) {
```
Chicken-Bones: invert condition for early return?

direwolf420: Makes sense

Chicken-Bones (approved):
The `GetContent` summaries are a little confusing, but I don't have a better idea how to word them.

direwolf420:
Last commit added a round of unifications for existing docs regarding terminology used:
* things using "content" now specify that it's a 'template' instance separately (or change it to "template" aswell for non-collection returns)
* things using "base" or nothing now use "template" when it's talked about retrieval for types that are non-singletons

I will be migrating this branch to 1.4.4 in a bit

Issue #2966, opened 2022-09-23
Fix/replace -join


Ved-s:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

-j/-join cmdline argument would be great for MP mod testing
I want to fix this and add a way to auto select player for this

Chicken-Bones:
Completed ion #2971

PR #2971, opened 2022-09-24
#2966 fix and improve -join

Ved-s:
Fixed -join and added -plr parameter for player to join with
-join argument may be with optional port (although, port can still be specified with -p)
-plr accepts player file path or player name

### DIFF
--- a/patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
+++ b/patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
@@ -11,7 +11,7 @@
  using Terraria.Social;
  
  namespace Terraria.Initializers
-@@ -26,12 +_,64 @@
+@@ -26,12 +_,16 @@
  			string s;
  			if ((s = TryParameter("-p", "-port")) != null && int.TryParse(s, out int result))
  				Netplay.ListenPort = result;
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
```diff
@@ -22,6 +22,59 @@
  		}
  
  		private static void LoadClientParameters(Main game) {
+-			string iP;
+-			if ((iP = TryParameter("-j", "-join")) != null)
+-				game.AutoJoin(iP);
```
Chicken-Bones: Please comment out original code instead of removing

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
```diff
@@ -22,6 +22,59 @@
...
++						}
++
++						Main.OpenPlayerSelect(fileData => {
++							fileData.SetAsActive();
++							Main.menuMode = 10;
++							Netplay.StartTcpClient();
++						});
++					};
++				}
++			}
```
Chicken-Bones: Could this not all be in `Main.AutoJoin`?

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
```diff
@@ -22,6 +22,59 @@
...
++
++				if (!Netplay.SetRemoteIP(ip)) {
++					ModLoader.Logging.tML.Error("Could not set remote IP: " + ip);
++				}
++				else {
++					if (port >= 0)
++						Netplay.ListenPort = port;
++
++					ModLoader.ModLoader.OnSuccessfulLoad += () => {
++						string player = TryParameter("-plr", "-player");
```
Chicken-Bones: If `-plr` or `-player` is not supplied we should keep the vanilla behaviour and go to the player select menu.

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
```diff
@@ -22,6 +22,59 @@
...
++			if ((ip = TryParameter("-j", "-join")) != null) {
++
++				int port = -1;
++				if (ip.Contains(':')) {
++					int colonInd = ip.LastIndexOf(':');
++					if (int.TryParse(ip.Substring(colonInd + 1), out int portNum)) {
++						port = portNum;
++						ip = ip.Substring(0, colonInd);
++					}
++				}
```
Chicken-Bones: This is helpful, but why not put it in `Netplay.SetRemoteIP`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
Ved-s: So edit `SetRemoteIP` to accept port?

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
Ved-s: That's what it does already

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
Ved-s: I can move code there. Should I add `string player` parameter to AutoJoin or make it parse -plr parameter?

Chicken-Bones: Add parameter

patches/tModLoader/Terraria/Initializers/LaunchInitializer.cs.patch
Chicken-Bones: Yes

Issue #2932, opened 2022-09-15
Figure out how to do an 'extended preview' for 1.4.4


Chicken-Bones:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

tModLoader code as a Contributor/Maintainer

### Description

We should consider how to run a multi-month preview.

Need to figure out:
- How to prevent automatic git stable rollover
  - Disable the action temporarily?
- Whether the preview version should reflect the month as it goes into the second month
  - I think yes
- Make sure that publishing on the second preview month doesn't delete stable (2 month old) versions of the mod from workshop
- No freeze in the first month?
- Anything else I haven't thought of

### What does this proposal attempt to solve or improve?

Terraria 1.4.4 will contain breaking changes, and mods may want more than 1 month to update. We should consider an extended preview. 

### Which (other) solutions should be considered?

- Moving stable to a legacy steam branch 
  - Probably not a good idea, break everyone's playthroughs with mods which haven't updated yet
- Publish a '1.4.4 stable' version to another steam beta branch and continue as normal
  - Still need to make sure that all 3 versions of a mod are kept in the steam folder (stable, 1.4.4 stable, preview)
  - Need to consider any text that directs people to 'preview' when their tML version is outdated
  - Does this help at all? Who would play on a 1.4.4 stable, when not enough mods have updated. If mods have updated, we would just make this the actual stable.
  - Would having two active 'stable' versions create issues? Probably not, no different from users who just haven't updated for a while.

Solxanich:
Closing as resolved under #2973

Note that all github actions scripts contain commented out lines of which are NOT dynamic- so we have to be conscious accordingly.

PR #2973, opened 2022-09-25
Introduce concept of LTS to support #2932

Solxanich:
This PR adds the long term service metadata to tmodloader builds to allow for extended preview periods.

This is in support of #2932 .

We will still need to disable the relevant portions of MonthlyMerge.yml & PReviewChanges.yml during such periods manually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ name: Build tModLoader
 on:
   # This workflow runs when a push to 1.4 branch happens, or when a label is added to a pull request targetting 1.4 branch.
   push:
-    branches: [1.4_mergedtesting, gh_actions_test, 1.4-preview, 1.4-stable, 1.4]
+    branches: [1.4_mergedtesting, gh_actions_test, 1.4-preview, 1.4-stable, 1.4, lts_tModLoader]
   pull_request_target:
     types: [labeled]
     branches: [1.4-preview, 1.4-stable, 1.4, 1.4_mergedtesting]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,19 +77,19 @@ jobs:
            custom_tag: ${{ steps.version.outputs.new_version }}
            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       
-      - name: Check out LTS branch
+      - name: Check out stable branch
         uses: actions/checkout@v2
         with:
           ref: "1.4-stable"
           persist-credentials: false
           fetch-depth: 0
       
-      - name: LTS Version Fetch
-        id: lts
+      - name: stable Version Fetch
+        id: stable
         if: github.event_name == 'push'
         shell: bash
         run: |
-          echo "Fetching the LTS version of tmodloader"
+          echo "Fetching the stable version of tmodloader"
           version=$(git describe --tags --abbrev=0)
           
           regex="^v([0-9]+).([0-9]+).([0-9]+).([0-9]+)?$"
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -226,6 +263,7 @@ jobs:
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
           TERRARIA_VERSION: 1436
           TMLVERSION: ${{ steps.version.outputs.new_version }}
+          LTSVERSION: ${{ steps.lts.outputs.lts_version }}
```
Chicken-Bones: Love this, I was just going to hard-code it, but this is great

patches/tModLoader/Terraria/ModLoader/BuildInfo.cs
```diff
@@ -37,23 +38,24 @@ public enum BuildPurpose
...
+			if (parts.Length>=5) {
+				CommitSHA = parts[4];
 			}
 			else {
 				CommitSHA = "unknown";
 			}
-			if (parts.Length>=5) {
-				BuildDate = DateTime.FromBinary(long.Parse(parts[4])).ToLocalTime();
+			if (parts.Length>=6) {
+				BuildDate = DateTime.FromBinary(long.Parse(parts[5])).ToLocalTime();
```
Chicken-Bones: Do we need all of these `parts.Length` checks? It looks like the number of separators is fixed now, so we could do fixed indices, or `parts[i++]` in all of them

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -535,40 +535,34 @@ private enum SearchFolders { }
...
-				File.Delete(location);
-			else
-				Directory.Delete(location, true);
-		}
-
-		internal static string FindOldest(string repo) {
-			string[] tmods = Directory.GetFiles(repo, "*.tmod", SearchOption.AllDirectories);
-			if (tmods.Length == 1)
-				return tmods[0];
+			var compareVersion = DecreaseDateVersion(BuildInfo.ltsVersion);
```
Chicken-Bones: Sure, keep all versions since the last `LTS`. Seems like a good start.

patches/tModLoader/Terraria/Terraria.csproj.patch
```diff
@@ -21,12 +21,13 @@
 -    <AssemblyName>$(AssemblyName)Release</AssemblyName>
 +  <PropertyGroup>
 +    <tMLVersion Condition="'$(TmlVersion)' == ''">9999.0</tMLVersion>
++	<LTSVersion Condition="'$(LTSVersion)' == ''">0.12</LTSVersion>
```
Chicken-Bones: Why 0.12? Because that's the 1.3 version? 1.3 really doesn't fit into this framework, so maybe 0.0 would be better?

Chicken-Bones (commented):
Looks good so far. Fortunately people can't publish from dev versions.

patches/tModLoader/Terraria/Terraria.csproj.patch
Solxanich: 0.12 was the Alpha version we first started uploading to Steam on. 
Kinda arbitrary.

Happy to set to 0.0 for a cleaner arbitrary

patches/tModLoader/Terraria/ModLoader/BuildInfo.cs
Solxanich: No we really don't. I probably should have removed them when I touched it; all the safety is in the csproj now.

Updated accordingly.

.github/workflows/build.yml
Solxanich: Yup, tried to make it more maintainable. Don't want to have to update this every month.

patches/tModLoader/Terraria/Terraria.csproj.patch
Chicken-Bones: Please do

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
Solxanich: To be precise - I'm keeping LTS minus 1 for rollover safety. 

For example, for Preview = 2022.9
LTS = 2022.8
LTS minus 1 = 2022.7

Only problem with this is if we ran two back to back 2+ month preview. Wouldn't be guaranteed for rollback to last LTS on player base.

PR #2977, opened 2022-09-25
Fix Item.DropItem Sharing ModItems

Antirhinnum:
### What is the bug?
Items dropped using `Item.DropItem()` would have their `ModItem` linked to two different `Item`s -- the one dropped in the world that the player can interact with and pick up, and the one passed into `DropItem()` that is ultimately lost. This would result in any modifications done to `ModItem.Item` not applying to the player's version of the item. For example, dual-use items that modify an item's fields on use would inexplicably break after being placed and retrieved from Plates or Item Frames.

### How did you fix the bug?
Reassigned `ModItem.Entity` for the newly-dropped `Item`.

Not that I'm aware of.

### DIFF
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -205,6 +205,8 @@ namespace Terraria
 
 			droppedItem.ModItem = item.ModItem;
 			droppedItem.globalItems = item.globalItems;
+			if (droppedItem.ModItem != null)
+				droppedItem.ModItem.Entity = droppedItem;
 
 			if (Main.netMode == NetmodeID.Server)
 				NetMessage.SendData(21, -1, -1, null, droppedItemId);


Chicken-Bones:
This needs a proper fix, that transfers the entire `Item` into the 'world', either via cloning or directly.
It looks like the current behaviour would also reset any changes that the `ModItem` or `GlobalItem` have made to `Item` fields.

JavidPack:
What we do in Player.QuickSpawnClonedItem might work.

Chicken-Bones:
`Player.QuickSpawnClonedItem` should not exist, it should be the default behaviour of `Player.QuickSpawnItem`, but yes it will work

JavidPack:
We fixed the underlying issue in [e2bb161](https://github.com/tModLoader/tModLoader/commit/e2bb161b4e46d63fba28b5bb76a471e8d8265b7f)

Thanks for bringing this to our attention.

PR #2981, opened 2022-09-27
Fix #2980 ChatManager.GetStringSize with newlines and format tags

### LINKED ISSUE #2980: Chat tags in multiline tooltips create empty space for no reason
### Version

Latest 1.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

mod capability as a modder **and** gameplay as a player. welcome to the real world

### Description

usin' color codes in multiline tooltips seems to make empty space in the tooltip for no reason, like so
image
this stacks with additional color codes and can actually cause the actual tooltip above it to be pushed off of the screen, like so
image

### Log File

the hell is a log file gonna do here?

### Steps to reproduce

**step the first:** make a multiline tooltip (the "multiline" aspect is critical, as this does not happen with separate lines)
**step the second:** use color codes within your newly-multline'd tooltip
**step the third:** ???
**step the fourth:** observe the empty space that's present in-game, and wonder why that tooltip revamp gets talked up so much when nobody has any concrete progress put forward towards doin' it

### Expected Behavior

there's no empty space at the bottom of the tooltip

### Actual Behavior

didn't I already tell you? there's empty space at the bottom of the tooltip where there shouldn't be

### Reproduction frequency

100% of the time, specifically when usin' chat tags in multiline tooltips. it doesn't seem to happen for anything else

### Additional Information

_No response_

Ved-s:
Fix for #2980

### DIFF
--- a/patches/tModLoader/Terraria/UI/Chat/ChatManager.cs.patch
+++ b/patches/tModLoader/Terraria/UI/Chat/ChatManager.cs.patch
@@ -1,10 +1,12 @@
 --- src/TerrariaNetCore/Terraria/UI/Chat/ChatManager.cs
 +++ src/tModLoader/Terraria/UI/Chat/ChatManager.cs
-@@ -147,7 +_,7 @@
+@@ -147,7 +_,9 @@
  						result.Y = Math.Max(result.Y, vector.Y + value.Y);
  					}
  
 -					if (array.Length > 1) {
++					// Fix spacing with \n and chat tags, #2981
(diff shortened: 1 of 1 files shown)


Tyndareus:
Is this on latest? Because I'm suddenly having an issue with newlines on tooltips.
image

The text string responsible is just `"The blood of a true ancient warrior courses through you.\nPerk 1: {0:P1} HP grants {1:P2} damage.\nPerk 2: All attacks drain 20% less Ki and deal 10% more damage during the night."`

And all I am doing is using that string in ModifyBuffTip

PR #2983, opened 2022-09-27
UI Fixes

Destructor-Ben:
2 things have been fixed / changed:
The mod info UI
Button's changing border colour when hovered over

The mod info UI was previously cramped and difficult to read in, so the panel width has been increased. Additionally, the homepage and workshop page buttons had scaled down text inside of them and were unaligned, so they now have correct spacing and the text inside of them doesn't have to be shrunk down to fit in them.

Modded buttons had previously only changed their background colour when they were moused over, which is different to the vanilla buttons. Now they will have their border changed when they are moused over to better fit them in with the vanilla UI.


JavidPack:
Looks good, thanks, I didn't even notice the vanilla buttons had that effect.

Issue #1616, opened 2021-07-18
[1.4] Unintended behavior regarding CombinedHooks.CanPlayerHitNPCWithItem


direwolf420:
### Description
Hooks called through `CombinedHooks.CanPlayerHitNPCWithItem` do not share the same logic as the 1.3 versions. Returning true ignores the melee hitbox collision, meaning a swung item will hit the NPC regardless of being in contact or not.
This is contrary to the projectile->NPC hooks where hitboxes are checked accordingly.

### Possible solutions
1. Pass `itemRectangle` into the hooks, modders would have to check `itemRectangle.Intersects(npc.getRect()) && (npc.noTileCollide || player.CanHit(npc))` if they want to return anything other than null
2. Change the order of operations to insert the hook in a similar place to mirror projectile->NPC hook logic to avoid modders having to check the above manually
3. Create an additional hook to allow more finetuning, quoting Mirsario: "some canhit hooks should likely be split into canhit and cancollidewith"

JavidPack:
Fixed in [d9eceb3](https://github.com/tModLoader/tModLoader/commit/d9eceb3249a19a4c5d03c62ebbfbe3ebd33882d2)

PR #2984, opened 2022-09-27
#1616 fix by separating collision check hook from melee hit hook

icreray:
### What is the new feature?
Hooks that called through `CombinedHooks.CanPlayerHitNPCWithItem` no longer ingores melee hitbox collision if methods returns true (for example already existing `ModPlayer.CanHitNPCWithProj` doesn't ignores the collision).
New methods works simular to `CombinedHooks.CanPlayerHitNPCWithItem` _(same way of combining hooks output, uses vanilla logic if returns null)_, but also have a hitbox of melee attack as additional parameter for custom hitbox check.

It adds much more predictable behavior of metioned hook methods, and more flexibility. (Personally i facing with this same issue when trying to make custom voodoo doll to my mod).

direwolf suggested a few in #1616.

Creating a custom voodoo doll:
```cs
public class VoodooDollPlayer : ModPlayer {

    private bool _canKillSomeNPC = false;
    
    public void AllowKillSomeNpc() { // This is called from SomeVoodooDollItem.UpdateEquip
        _canKillSomeNPC = true;
    }
    
    public override bool? CanHitNPC(Item item, NPC target) {
        if (_canKillSomeNPC && target.type == /* some npc id */) {
            return true; // This is no longer ignores collision (previously the player could hit NPCs from any distance)
        }
        
        return null;
    }
    
    // ... 
}
```

Modifying melee hit collision check with new hook:
```cs
public class CannotHitCloseNPCsPlayer : ModPlayer {

    public override bool? CanMeleeAttackCollideWithNPC(Item item, Rectangle meleeAttackHitbox, NPC target) {
        Rectangle playerHitbox = Player.getRect();
        Rectangle tagetNPCHitbox = target.getRect();
        
        if (playerHitbox.Intersects(targetNPCHitbox)) {
           return false; // Player can't damage npc
        }

        return null; // Vanilla collision check
    }
    
    // ...
}
```
### Porting notes
`ModPlayer.CanHitNPC`, `ModItem.CanHitNPC`, `GlobalItem.CanHitNPC`, `ModNPC.CanBeHitByItem` and `GlobalNPC.CanBeHitByItem` now won't ignore the hitbox collision checks when returning `true` in them. 
Collision checking happens after them. To disable collision checking return `true` from corresponding methods: `ModPlayer.CanMeleeAttackCollideWithNPC`, `ModItem.CanMeleeAttackCollideWithNPC`, `GlobalItem.CanMeleeAttackCollideWithNPC`, `ModNPC.CanBeCollidedWithPlayerMeleeAttack`, `GlobalNPC.CanBeCollidedWithPlayerMeleeAttack` _(Also be aware that the collision check also happens if CanHitNPC returns null and the npc can get hit via vanilla behavior)_.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
+++ b/patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
@@ -113,6 +113,32 @@ namespace Terraria.ModLoader
 			return result;
 		}
 
+		//TODO: Clean up code duplications in methods with same way of combining hooks result
+		public static bool? CanPlayerMeleeAttackCollideWithNPC(Player player, Item item, Rectangle meleeAttackHitbox, NPC target) {
+			bool? result = null;
+
+			bool ModifyResult(bool? nbool) {
+				if (nbool.HasValue) {
+					result = nbool.Value;
+				}
+				return result != false;
+			}
+
+			if (!ModifyResult(PlayerLoader.CanMeleeAttackCollideWithNPC(player, item, meleeAttackHitbox, target))) {
+				return false;
+			}
+
+			if (!ModifyResult(ItemLoader.CanMeleeAttackCollideWithNPC(item, meleeAttackHitbox, player, target))) {
+				return false;
+			}
+
+			if (!ModifyResult(NPCLoader.CanBeCollidedWithPlayerMeleeAttack(target, player, item, meleeAttackHitbox))) {
+				return false;
+			}
+
+			return result;
+		}
+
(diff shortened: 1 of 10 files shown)


icreray:
Oh, this commit also applies to pull request...

JavidPack:
K, I ported to 1.4.4 and added some comments. It was simpler to redo on 1.4.4 and push directly and mark you as author, so I'll close this one. 

Thanks for writing this up.

PR #2986, opened 2022-09-27
Fix #2743, show message to modders without .net 6 sdk installed

### LINKED ISSUE #2743: Display advisory message on Mod Sources menu requiring dotnet 6.0 sdk

Yes

Mod capability as a Modder

### Description

When opening the mod sources menu, tML should run `dotnet --list-sdks` and parse the output. If no compatible (6.0+) sdks are available, an info screen should be shown with a link to the download page, and a recommendation to install Visual Studio 2022 (if on windows).

There are plans #2472 to require the sdk in the future. And it has recently been reported that tModPorter can get stuck in an infinite loop when the user only has an older sdk (eg 5.0) on their system.

Unfortunately, I can't debug the tModPorter issue, because building tModPorter means having a 6.0 sdk installed... (and I haven't figured out how to remote debug another PC, or if it's worth the effort).

Maybe it shouldn't popup _every_ time, but really, people need to go get the sdk if they wanna make mods. It's not hard, and should work 100% of the time.

JavidPack:


patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -225,6 +230,57 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
 			modListViewPosition = _modList.ViewPosition;
 		}
 
+		private bool ShouldSkipPopulate() {
```
Chicken-Bones: Names like `Should` imply a lack of side-effects. Consider instead `ShowInfoMessages`

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -225,6 +230,57 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
...
+		private void ShowWelcomeMessage(string altButtonTextKey, string url, int gotoMenu = Interface.modSourcesID, UIState state = null) {
+			Interface.infoMessage.Show(Language.GetTextValue("tModLoader.MSFirstLaunchModderWelcomeMessage"), gotoMenu, state, Language.GetTextValue(altButtonTextKey),
+			() => {
+				SoundEngine.PlaySound(SoundID.MenuOpen);
+				Utils.OpenToURL(url);
+			});
+		}
+
+		private bool CheckDotnet() {
+			if (!dotnetSDKFound) {
```
Chicken-Bones: invert if to reduce nesting

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -225,6 +230,57 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
...
+				try {
+					string output = Process.Start(new ProcessStartInfo {
+						FileName = "dotnet",
+						Arguments = "--list-sdks",
+						UseShellExecute = false,
+						RedirectStandardOutput = true
+					}).StandardOutput.ReadToEnd();
+					Logging.tML.Info(output);
```
Chicken-Bones: `"\n" + output"` that way all output lines stack up neatly in the logs

Chicken-Bones (changes requested):
Looks good, minor code review

Issue #2972, opened 2022-09-24
Add new secret seed conditions for ItemDropRules ByCondition


Setnour6:
### Do you intend to personally contribute/program this feature?

Yes (but I can't do everything)

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

So far, there are two secret seed conditions implemented within the ItemDropRules.Conditions, and that being TenthAnniversaryIsUp and DontStarveIsUp.
![image](https://user-images.githubusercontent.com/70764728/192102494-d9428109-e974-4e27-a92b-72c6c5fad6f2.png)
This is one of the ItemDropRule Conditions for world seeds, being TenthAnniversaryIsUp.
Of course, there is also the inverse condition, "SecretSeed"Is**Not**Up

I noticed that there is no drop conditions for any other seed, like the "For The Worthy" seed world and Drunk World.
So why don't we add drop conditions for the other seeds?

Here is an example for the "For The Worthy" seed condition implementation:
```cs
public class ForTheWorthyIsUp : IItemDropRuleCondition, IProvideItemConditionDescription
        {
            public bool CanDrop(DropAttemptInfo info)
            {
                return Main.getGoodWorld;
            }

            public bool CanShowItemDropInUI()
            {
                return Main.getGoodWorld;
            }

            public string GetConditionDescription()
            {
                return Language.GetTextValue("Bestiary_ItemDropConditions.IsForTheWorthy");
            }
        }
````
Now obviously it's not perfect, and I do want something like this in the game for the potential to make it so loot can be dropped exclusively or increasingly on drunk or worthy worlds, or who knows what.

### What does this proposal attempt to solve or improve?

In a worthy world, players would join the world and defeat bosses as usual; but with an extra challenge, there should be a way for players to have a chance at receiving greater rewards, like an exclusive item.
There are already exclusive/alternative droprate items in the Tenth Anniversary seed worlds (not that I would know because I haven't tried the tenth anniversary world for droprule related stuff), and in the Constant seed world.

Say that a mod decides to make it so when a boss is defeated in a "Not The Bees" world, an extra drop is included, being a special item with bee-like properties on the item, such as releasing stingers or bees alongside their other properties (or other special properties).

### Which (other) solutions should be considered?

Alternatively, you can just use an if statement and check if the player is in a seed world, and then add your drop rule there like you normally would.

Chicken-Bones:
These seem like reasonable tML additions. Please open a PR

PR #2987, opened 2022-09-27
Add new secret seed conditions for ItemDropRules

Setnour6:
Condition descriptions are null but I want them to have values just like in the actual conditions.cs file, where GetTextValue gets the entry being in string "Bestiary_ItemDropConditions.Is10thAnniverary" for example. If possible, change the patch file to add the new seeds the way the conditions file has them. Also see #2972 for the issue.

### What is the new feature?
I noticed that there is no drop conditions for any other seed, like the "For The Worthy" seed world and Drunk World.
So why don't we add drop conditions for the other seeds?
See Issue #2972 
See Issue #2972 
Alternatively, you can just use an if statement and check if the player is in a seed world, and then add your drop rule there like you normally would.
See Issue #2972 
```cs
public class ForTheWorthyIsUp : IItemDropRuleCondition, IProvideItemConditionDescription
        {
            public bool CanDrop(DropAttemptInfo info)
            {
                return Main.getGoodWorld;
            }

            public bool CanShowItemDropInUI()
            {
                return Main.getGoodWorld;
            }

            public string GetConditionDescription()
            {
                return Language.GetTextValue("Bestiary_ItemDropConditions.IsForTheWorthy");
            }
        }
```
There is no actual Localization description for GetTextValue on IsForTheWorthy, and I am unable to provide that for tModLoader (if anyone knows how, please help provide it for me).
### ExampleMod updates
_N/A_

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/Conditions.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/Conditions.TML.cs
@@ -20,5 +20,41 @@
 			public bool CanShowItemDropInUI() => true;
 			public string GetConditionDescription() => null;
 		}
+		public class DrunkWorldIsUp : IItemDropRuleCondition, IProvideItemConditionDescription
+		{
+			public bool CanDrop(DropAttemptInfo info) => Main.drunkWorld;
+			public bool CanShowItemDropInUI() => true;
+			public string GetConditionDescription() => null;
+		}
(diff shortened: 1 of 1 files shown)


Setnour6:
This is everything I can do for now

Setnour6:
@Chicken-Bones do you still want to add the localization description for Bestiary_ItemDropConditions_IsForTheWorthy, or should I wait until tModLoader fully receives the 1.4.4 updates (assuming that they added the extra bestiary localization already)?

Setnour6:
Closing this PR due to inactivity and current non-requirement

PR #2990, opened 2022-09-28
ExpandAttribute

zzp198-old:
### What is the new feature?

Modders can now choose the default expand of the list/class.

It can be well managed when there are many config items.

no

[Expand(true/false)]
public List<string> arg;

[Expand(true/false)]
public Dictionary<int, string> arg;

[Expand(true/false)]
public HashSet<string> arg;

[Expand(true/false)]
public class SubConfig{}

### ExampleMod updates

Added an example in ModConfigShowcaseMisc


JavidPack:
Looks good, I added what we talked about earlier.

PR #2995, opened 2022-10-01
ExampleMod Example for Projectiles Moving in a Sine Wave

absoluteAquarian:
- added `Content/Projectiles/ExampleSineWaveMotionProjectile.cs`, which serves as an example of making projectiles move in a sine wave
- added `Content/Items/Weapons/ExampleMagicRod.cs`, the item that spawns the `ExampleSineWaveMotionProjectile` projectile

Not sure


direwolf420:
Overall agree that this should be added. There's even guides on the forums for such a thing: https://forums.terraria.org/index.php?threads/creating-a-projectile-that-moves-in-a-sive-wave-pattern.83245/#post-1821219

One small formatting thing I noticed:
```
} else {
```
should be split like

```
}
else {
```

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+		public override void AI() {
+			// Ensure that the "waveDirection = 1" projectile starts above the "waveDirection = -1" projectile not matter which direction they were fired from
+			int trueDirection = Projectile.direction * waveDirection;
+
+			// How many oscillations happen per second
+			// Higher value = more oscillations
+			float wavesPerSecond = 2;
+
+			// Math.Sin expects a radians angle instead of degrees
+			float radians = MathHelper.ToRadians(sineTimer * 6f * wavesPerSecond);
```
Chicken-Bones: Okay, sure, 'ticks' * 6 gives you 1 revolution per second, but uhh, not very clear. Instead I suggest:
```cs
float waveProgress = sineTimer / 60f * wavesPerSecond; // 1 for each full sine wave
float angle = waveProgress * 2 * MathF.PI; // angles in radians
```

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+			float sine = (float)Math.Sin(radians) * trueDirection;
```
Chicken-Bones: `MathF.Sin` exists

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+			offset *= sine * waveAmplitude;
+
+			// Update the position manually since ShouldUpdatePosition returns false
+			initialCenter += Projectile.velocity;
+			Projectile.Center = initialCenter + offset;
+
+			// Update the rotation used to draw the projectile
+			// This projectile should act as if it were moving along the sine wave
+			float cosine = (float)Math.Cos(radians) * trueDirection;
+			Projectile.rotation = Projectile.velocity.ToRotation() + MathHelper.PiOver4 * cosine * -1;
```
Chicken-Bones: Why add the `cosine` here? Not clear how this math works.

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+			initialCenter = Projectile.Center;
+		}
+
+		public override Color? GetAlpha(Color lightColor) {
+			return drawColor * 0.64f;
+		}
+
+		public override void AI() {
+			// Ensure that the "waveDirection = 1" projectile starts above the "waveDirection = -1" projectile not matter which direction they were fired from
+			int trueDirection = Projectile.direction * waveDirection;
```
Chicken-Bones: Why use `Projectile.direction` here? I don't see where it's set.

ExampleMod/Content/Items/Weapons/ExampleMagicRod.cs
```diff
@@ -0,0 +1,72 @@
...
+					drawColor = Color.Red;
+				} else {
+					waveDirection = -1;
+					drawColor = Color.Blue;
+				}
+
+				Projectile projectile = Projectile.NewProjectileDirect(source, position, velocity, type, damage, knockback, player.whoAmI);
+
+				ExampleSineWaveMotionProjectile modProjectile = projectile.ModProjectile as ExampleSineWaveMotionProjectile;
+				modProjectile.waveDirection = waveDirection;
```
Chicken-Bones: Suggest that instead of using `waveDirection` you use `waveOffset` (from 0-1, added to angle before multiplying by 2pi) that way if you had 4 projectiles you could do fractions.

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+					Projectile.frame = Projectile.direction == 1 ? 2 : 0;
+				}
+				else {
+					Projectile.frame = 1;
+				}
+			}
+			else {
+				// Offset is zero
+				Projectile.frame = 1;
+			}
```
Chicken-Bones: Comments are needed to explain more about the relationship between sprite and math here. My understanding is that you just want to change the sprite when you're at a 'peak' or 'trough' to indicate it'll be turning around.

You can detect those just by comparing `sine` though, and if I'm reading the math correctly, you currently have the 'curved' sprites (0 and 2) used when the projectile is travelling straight (0-45, 135-225, 315-... degrees)

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
```diff
@@ -0,0 +1,150 @@
...
+
+			// Spawn dusts
+			Dust.NewDustPerfect(Projectile.Center, ModContent.DustType<Dusts.ExampleSolution>(), Velocity: Vector2.Zero, newColor: drawColor, Scale: 0.5f);
+
+			Projectile.direction = (Projectile.velocity.X > 0).ToDirectionInt();
+
+			sineTimer++;
+		}
+
+		public override bool PreDraw(ref Color lightColor) {
```
Chicken-Bones, line 128: I'm pretty sure this shouldn't be required. Why is positioning and framing the texture not enough? By altering the draw here, you can create differences between draw and hitbox.

Chicken-Bones (changes requested):

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
absoluteAquarian: The derivative of `sin(x)` is `-cos(x)`.  I use this derivative to make the projectile "point" toward where it will be moving.

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
absoluteAquarian: Isn't it set automatically?

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
absoluteAquarian: i tried to do that via the DrawOffset variables, but it was annoying to make them work when the projectile was reversed

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
absoluteAquarian: I double checked, and it is.

absoluteAquarian: nevermind on the previous two comments.  `Projectile.direction` is set later in AI

ExampleMod/Content/Projectiles/ExampleSineWaveMotionProjectile.cs
absoluteAquarian: changed to only check `sine`

PR #3000, opened 2022-10-01
OnStartDay and OnStartNight hooks

Cyrillya:
### What is the new feature?
Added `OnStartDay` and `OnStartNight` hooks to ModSystem.
Useful if you want to start a invasion or spawn a boss when day/night starts.
Spawn a boss naturally at night like EoC.
### ExampleMod updates
Each night the Minion Boss has a 10% chance of being spawned naturally if it is not downed.

### DIFF
--- a/ExampleMod/Common/Systems/NaturalBossSpawnSystem.cs
+++ b/ExampleMod/Common/Systems/NaturalBossSpawnSystem.cs
@@ -1,0 +1,64 @@
+﻿using ExampleMod.Content.NPCs.MinionBoss;
+using Microsoft.Xna.Framework;
+using Terraria;
+using Terraria.Chat;
+using Terraria.ID;
+using Terraria.Localization;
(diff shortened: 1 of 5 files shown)


direwolf420:
So the main selling point of this PR is the ability to prevent natural boss/invasions from occuring? Because doing code on day start/end is already possible using existing ModSystem hooks by checking Main.time and Main.dayTime

Cyrillya:
I just realized that we can use `Main.time` and `Main.dayTime`. So these hooks may be useless

steviegt6:
These seem convenient enough to be justifiable, it at least reduces boilerplate on the modder's side.

JavidPack:
At the very least someone can PR an example of a nightly or morning event with random chance for ExampleMod, just to show the correct code. Even a simple Main.NewText("You think today will be really cool for no reason") would suffice.

PR #3020, opened 2022-10-06
Various checks for latest Terraria install, Fix #1979

### LINKED ISSUE #1979: 1.4 Alpha check Terraria version

JavidPack:
Still to do: 
Maybe we need some sort of checklist or task list token to remind us of all the places things need to update when a new version comes out.

### DIFF
--- a/patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
+++ b/patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
@@ -37,12 +37,33 @@
  			services.AddService(typeof(AssetReaderCollection), assetReaderCollection);
  			services.AddService(typeof(IAssetRepository), provider);
  		}
-@@ -597,6 +_,8 @@
+@@ -597,6 +_,29 @@
  			Main.ResourceSetsManager.LoadContent(mode);
  			Main.MinimapFrameManagerInstance.LoadContent(mode);
  			Main.AchievementAdvisor.LoadContent();
 +
 +			UICommon.LoadTextures(); // Added by TML.
++		}
++
++		private static bool canaryFilesChecked;
++		internal static void CheckCanaryFiles() {
++			if (canaryFilesChecked)
++				return;
++
++			// These checks ensure the latest Terraria is installed.
++			try {
++				Main.Assets.AssetLoadFailHandler = null;
++
++				LoadAsset<Texture2D>("Images/Projectile_981", AssetRequestMode.ImmediateLoad);
++			}
++			catch {
++				ModLoader.Engine.ErrorReporting.FatalExit("Terraria assets are missing, make sure to update Terraria to the latest version.");
++			}
++			finally {
++				Main.Assets.AssetLoadFailHandler = Main.OnceFailedLoadingAnAsset;
++			}
++
++			canaryFilesChecked = true;
  		}
  
  		private static Asset<T> LoadAsset<T>(string assetName, AssetRequestMode mode) where T : class => Main.Assets.Request<T>(assetName, mode);
--- a/patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
+++ b/patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
@@ -37,33 +37,12 @@
  			services.AddService(typeof(AssetReaderCollection), assetReaderCollection);
  			services.AddService(typeof(IAssetRepository), provider);
  		}
-@@ -597,6 +_,29 @@
+@@ -597,6 +_,8 @@
  			Main.ResourceSetsManager.LoadContent(mode);
  			Main.MinimapFrameManagerInstance.LoadContent(mode);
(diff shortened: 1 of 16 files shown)


patches/tModLoader/Terraria/ModLoader/Engine/TerrariaSteamClient.cs
```diff
@@ -140,6 +140,11 @@ public enum LaunchResult
 					return;
 				}
 
+				int TerrariaBuildID = SteamApps.GetAppBuildId();
+				Logger.Info("Terraria BuildID: " + TerrariaBuildID);
+				if (TerrariaBuildID < 9653812) // Currently v1.4.4.4. Update this when any Terraria update changes any asset.
```
Chicken-Bones: Make this a constant at the top of the file.

Chicken-Bones: Are these UUIDs or do they actually increment?

patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
```diff
@@ -37,12 +37,33 @@
...
 +
 +			UICommon.LoadTextures(); // Added by TML.
++		}
++
++		private static bool canaryFilesChecked;
++		internal static void CheckCanaryFiles() {
++			if (canaryFilesChecked)
++				return;
++
++			// These checks ensure the latest Terraria is installed.
```
Chicken-Bones: Is this necessary? I'd prefer a `File.Exists` in `InitTMLContentManager`, since we have the `vanillaContentFolder` there at the time, and it is mandatory

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Engine/TerrariaSteamClient.cs
JavidPack: They increment only.

patches/tModLoader/Terraria/Initializers/AssetInitializer.cs.patch
JavidPack: done

patches/tModLoader/Terraria/ModLoader/Engine/TerrariaSteamClient.cs
```diff
@@ -140,6 +141,11 @@ public enum LaunchResult
 					return;
 				}
 
+				int TerrariaBuildID = SteamApps.GetAppBuildId();
+				Logger.Info("Terraria BuildID: " + TerrariaBuildID);
+				if (TerrariaBuildID < LatestTerrariaBuildID)
```
Chicken-Bones: can you pipe this back to the main process via an error code? This should be the primary point of failure.

patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -190,6 +190,11 @@ public partial class Main
 				ErrorReporting.FatalExit(Language.GetTextValue("tModLoader.ContentFolderNotFound"));
 			}
 
+			// Canary file, ensures that Terraria has updated to at least the version this tModLoader was built for
+			if (!File.Exists(Path.Combine(vanillaContentFolder, "Images", "Projectile_981.xnb"))) {
+				ErrorReporting.FatalExit(Language.GetTextValue("tModLoader.TerrariaOutOfDateMessage"));
```
Chicken-Bones, line 195: Is this needed given the build check ID below? Is there a possibility of someone having an incomplete Terraria install? If so, maybe this should just tell people to verify their terraria local files instead.

Chicken-Bones (changes requested):

Chicken-Bones (approved):

PR #3023, opened 2022-10-09
Fix engine crash when failed to fetch local mod dependencies

Lama3L9R:
### What is the bug?
Game crashed when failed to fetch mod dependencies due to bad network connection
### How did you fix the bug?

Surround try catch around this method call and log the issue.
```
WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
```
Of course yes! Currently I just log the network issue not showing it to the user. The best way is to open a message box and tell the player about his poor network connection. However I'm not a game developer and I don't know how to draw stuff on screen. So I just catch it and log it.

I have tested my code in both development environment and release environment, both of them are working expectly.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -138,7 +138,17 @@ namespace Terraria.ModLoader.Core
 				if (!TryReadManifest(GetParentDir(mod.modFile.path), out var manifest))
 					continue;
 
-				WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				try {
+					WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				}
+				catch (OverflowException e) {
+					Utils.ShowFancyErrorMessage(Language.GetTextValue("tModLoader.WorkshopIrregularDependenciesFailure", mod.DisplayName, e.Message, mod.DisplayName), Interface.loadModsID);
+					Logging.tML.Warn($"Failed to fetch missing dependencies for local mod: {mod.DisplayName}. Irregular dependencies count detected! {e.Message}. Please report this issue to {mod.DisplayName}'s developers!", e);
+				}
+				catch (Exception e) {
+					Utils.ShowFancyErrorMessage(Language.GetTextValue("tModLoader.WorkshopFetchDependenciesFailure", mod.DisplayName), Interface.loadModsID);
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -138,7 +138,12 @@ private enum SearchFolders { }
 				if (!TryReadManifest(GetParentDir(mod.modFile.path), out var manifest))
 					continue;
 
-				WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				try {
+					WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				}
+				catch (Exception e) {
+					Logging.tML.Warn($"Failed to fetch dependencies for local mod: {mod.DisplayName}. Check your internet connection!", e);
```
Solxanich: Let's follow your advice to have this go to a UI indicating what happened; you are actually fairly close.

Use `Utils.ShowFancyErrorMessage("message", Interface.loadModsID)`.
We added this method a long time ago due to how often we need Error UI messages in the codebase.
Interface.loadModsID is the UI State when this gets called; resetting it will continue without this check (this time),

Update the message to mention that the 'dependencies' are 'missing dependencies', and that you can check the Steam Workshop to manually download them there.

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -138,7 +138,13 @@ private enum SearchFolders { }
 				if (!TryReadManifest(GetParentDir(mod.modFile.path), out var manifest))
 					continue;
 
-				WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				try {
+					WorkshopHelper.QueryHelper.GetDependenciesRecursive(manifest.workshopEntryId, ref dependencies);
+				}
+				catch (Exception e) {
+					Utils.ShowFancyErrorMessage($"Failed to fetch missing dependencies for local mod: {mod.DisplayName}. Check your internet connection or download manually from Steam Workshop!", Interface.loadModsID);
+					Logging.tML.Warn($"Failed to fetch missing dependencies for local mod: {mod.DisplayName}. Check your internet connection!", e);
```
steviegt6: Should probably be moved to a localization key, and the string value should be the same for both here.

Lama3L9R: What key should I use? `tModLoader.WorkshopFetchDependenciesFailure` ?

steviegt6: Sure.

Issue #2747, opened 2022-07-22
Item Chat Tags in ModConfigurations are off-center


SheepishShepherd:
### Version

Latest 1.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

When using item chat tags in Labels or Headers, the items are way off-center compared to other places it is used. Also, when using item chat tags in Tooltips, they do not function at all.

![unaligned](https://user-images.githubusercontent.com/36808587/180480356-5dff1895-0139-4ed4-871c-09783d76419d.png)

### Steps to reproduce

Add any item chat tag to a Mod's configurations.

### Expected Behavior

Item chat tags should appear within the bounds of the clickable config, and be more aligned with text.

### Actual Behavior

Item chat tags noticeably dip further down than they should.

JavidPack:
I looked into this, and there isn't much I can do to fix this without changing existing behavior in chat. I can move the header underline down, that's about it. I might look later.

tooltips now work, and I think items in labels now draw more correctly (the scale of them).

PR #3030, opened 2022-10-12
Scalable chat tags

Ved-s:
adds TextSnippet.UniqueDraw overload with Vector2 scale parameter
fixes ChatManager methods to use baseScale when rendering chat tags
fixes GetStringSize to work correctly with scaled chat tags

### Before

Scaling 0.7x
image

Scaling 2x
image

### After

Scaling 0.7x
image

Scaling 2x
image


patches/tModLoader/Terraria/UI/Chat/TextSnippet.cs.patch
```diff
@@ -0,0 +1,26 @@
...
++using System;
+ 
+ namespace Terraria.UI.Chat
+ {
+@@ -36,6 +_,15 @@
+ 
+ 		public virtual Color GetVisibleColor() => ChatManager.WaveColor(Color);
+ 
++		/// By default, calls UniqueDraw with float scale parameter for compatibility
++		public virtual bool UniqueDraw(bool justCheckingString, out Vector2 size, SpriteBatch spriteBatch, Vector2 position = default(Vector2), Color color = default(Color), Vector2? scale = null) {
```
direwolf420: should be //

direwolf420 (commented):
A before/after would be nice to see

patches/tModLoader/Terraria/UI/Chat/TextSnippet.cs.patch
Ved-s: done

JavidPack:
Looks like we arrived at some of the same conclusions. I was looking into this and found similar fixes, I'll do more testing later and see if we need to bring in some of the other changes I found when I was working on this earlier. 

Can you show how your current implementation works for the modconfig issue?
https://github.com/tModLoader/tModLoader/issues/2747

Ved-s:
image
It's still off-center but it's not clipping out

Ved-s:
Probably because there's a patting at the top because of wacky non-symmetrical font size

JavidPack:
Thanks.

PR #3040, opened 2022-10-16
Mod loading starts during splash screen.

ghasttear1:
### What is the new feature?

Mod loading starts during the splash screen.

Initially while mucking around, I implemented skipping the splash screen as soon as vanilla was done loading however I thought this made more sense. This feature makes it quicker for everyone to get to the main menu and get in game as the splash screen has dead time where it is doing nothing and is much better spent getting a head start on loading.

Please let me know if you think this would be better implemented in a different way or could use changes, I am more than happy to implement any feedback.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -337,7 +337,7 @@
  		public static int selectedPlayer = 0;
  		public static int selectedWorld;
 -		public static int menuMode;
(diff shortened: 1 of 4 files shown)


direwolf420:
How does this feature interact with various UIs that appear before mod loading takes place? I.e. the welcome screen on the first game launch, or the "mod updates outside of the game since last launch" screen

ghasttear1:
I did make sure to check with those and they looked to be working when I tried.

PR #3050, opened 2022-10-20
Added WallID.CanBeClearedDuringGeneration

Tyfyter:
### What is the new feature?
Adds WallID.CanBeClearedDuringGeneration to replace the hardcoding of which walls can't be replaced by WorldGen.Spread.Wall2, could probably be more accurately named

Currently any walled caves which happen to extend into areas which will have grass walls have all of their walls (within range) replaced

The set could be substantially more or less specific


Chicken-Bones:
@Tyfyter Looks like an appropriate name for this is actually `CannotBeReplacedByWallSpread`

Also, please add new sets to `WallID.TML.cs` and use id constants rather than magic numbers

direwolf420:
but magic numbers are easier to maintain because you can tell at a glance when something changed by comparing the original vanilla code and the tml set?

Chicken-Bones:
You'll only notice when reviewing patches in an update, in which case you have the diff in magic-number form, so you can update the set right then and there

Chicken-Bones:
Added to vanilla for a coming hotfix, probably 1.4.4.8

PR #3057, opened 2022-10-25
Added IsPlayingFilm property to CinematicManager

Kazzymodus:
### What is the new feature?

This PR simply adds a property to CinematicManager to expose whether any films are active at the moment.

Films can already be used in tModLoader, but there is no way to check whether a film is currently playing. If that were possible, we could allow for certain logic to be executed while films are running, such as locking the camera to a position or continuously modifying a player variable that would otherwise be reset every frame. This allows for much more robust cutscene control, as you would no longer be limited to only modifying state when a film starts or stops.

~~The use of Any is not strictly necessary as even in the case of multiple films only the first film in the collection should be active. The extra overhead of being thorough is completely minimal, however, especially considering in most cases there will be only one film in the collection to begin with, so I see no good reason to use something else.~~

Now the property only checks the first film, in order to properly reflect what the Manager is doing rather than what state the films are in.

Suppose that we want to lock the camera to a specific position while a film is playing, rather than being centered around the player. Without the property we can't do this: we can only change state at the beginning and end of a film, but camera position is recalculated every frame. With the new property, however, we could do this:

```cs
public class FilmSystem : ModSystem
{
    public override void ModifyScreenPosition()
    {
        if (!CinematicsManager.IsPlayingFilm) return;
        Main.screenPosition = new Vector2(61900, 5420);
    }
}
```

### ExampleMod updates

None included, as it would only make sense if there actually was an ExampleFilm, which doesn't rely on tModLoader functionality to begin with. If needs be, the above FilmSystem could be included, but it wouldn't ever do anything within ExampleMod itself.


Chicken-Bones:
As discussed in Discord, this PR needs to include an example film of some kind, with a way to play it (eg via command). `CinematicsManager` is unused in vanilla, so Example Mod will be required to test the effectiveness of any changes.

Kazzymodus:
Unfortunately I don't feel the extra work (which is justified) weighs up to the benefits of this change, so I'm closing this PR for now.

PR #3058, opened 2022-10-25
Aggressively check dotnet installed correctly.

JavidPack:
In discord support, a common issue is incomplete dotnet installs. Sometimes a dll is missing and the game crashes, sometimes dotnet.exe is missing. The cause of this is still unknown. Antivirus quarantines sounds like a reasonable reason, but users with this issue claim antivirus is off. Many users have success deleting the dotnet folder, or manually extracting the dotnet zip instead of letting the launch script do it.

This branch attempts to detect incomplete installs and force a reinstall if detected.

I'll get some #support people to test this script.


patches/tModLoader/Terraria/release_extras/LaunchUtils/InstallNetFramework.sh
```diff
@@ -30,6 +30,16 @@ if [ -d "$dotnet_dir" ]; then
 			rm -rf "$old_version"
 		fi
 	done
+	
+	# For some reason, sometimes the dotnet folder is missing some files, if so, delete it so it gets installed again
+	numberFilesInDotnetDir=`find "$install_dir" -type f | wc -l`
+	expectedFilesInDotnetDir=227
```
Solxanich, line 36: This number would be true strictly for windows x64 binaries. 

Example: Mac x64 binary is 185 files.

Do we want to restrict this to $_uname == _NT for windows only? 
Or do we want to make the fun switch case?

Rest of the code all looks ok, and seems to work fine.

Osjoud:
Pls help with #3342 for the love of god

PR #3059, opened 2022-10-27
Fixed vines disappearing when teleporting

Jadams505:
### What is the bug?
Vines and Seaweed disappear for a frame when teleporting in color mode
Easily reproducible by using a rod of discord near vines with lighting mode set to color

More technically:

This issue has to do with `Main.renderNow` being set to true. The positions for where vines are drawn are updated in `Main.RenderTiles2()` and only update when `TileDrawing._vineRootsPositions` does not contain that position. `_vineRootsPositions` is cleared when `solidLayer` is true which occurs when `Main.RenderTiles()` is called. Thus drawing vines relies on the fact that `RenderTiles()` is called before `RenderTiles2()`. However due to how tModLoader changes render order (see #1802) when the game "renders now" after a teleport `RenderTiles2()` is called and `renderCount` is reset to 0 then it gets incremented to 1 (no issue) then it gets incremented to 2 and `RenderTiles2()` is called again and since `RenderTiles()` was not called in-between `_vineRootsPositions` is never cleared so `_specialCount` never gets updated so no vines are drawn until the next call of `RenderTiles2()` where everything is back to normal.

### How did you fix the bug?
I made it so that `_vineRootsPositions` and `_reverseVineRootsPositions` clear when `solidLayer` is false. In other words when `RenderTiles2()` is called so it does not rely on `RenderTiles()` to be called. This is also because the only time they are updated is when `solidLayer` is false so clearing them right before they get updated makes the most sense.

Honestly the relationship between `TileDrawing._specialPositions` and `TileDrawing._vineRootsPositions` seems very redundant and it wouldn't be too hard to not use `_vineRootsPositions` all together. It seems like it's there just for convenience.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
@@ -330,7 +330,7 @@
  		}
  
  		private bool IsWindBlocked(int x, int y) {
-@@ -4440,16 +_,20 @@
+@@ -4440,16 +_,21 @@
  			if (solidLayer) {
  				_displayDollTileEntityPositions.Clear();
  				_hatRackTileEntityPositions.Clear();
@@ -338,6 +338,7 @@
 -				_reverseVineRootsPositions.Clear();
 +				return;
  			}
++			//TML: Moved so positions get cleared and added in the same frame
 +			_vineRootsPositions.Clear();
 +			_reverseVineRootsPositions.Clear();
  		}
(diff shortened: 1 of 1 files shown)


direwolf420:
Glad you found a solution to this issue which has been bugging me for a while but I never got around to make an issue for!
As a suggestion, seeing the patch you made by itself doesn't explain why the code was moved. Adding a short comment (similar to the one you see in the linked PR) would help contextualizing it
Have you tested if your fix works for all lighting modes?

Jadams505:
Comment added. Also I have tested and it and it does work in all lighting modes

PR #3060, opened 2022-10-28
Added ModNPC.NoTownNPCHappiness Property to Allow Town NPCs to Not Be Affected by Happiness.

Rijam:
### What is the new feature?

Added a new `ModNPC` property, `NoTownNPCHappiness`, that allows the modder to make their Town NPC not affect by Happiness.

Even if you do not set any Town NPC affection in your Town NPC, the Town NPC will still be affected by solitude/crowding, being homeless, and being in the Corruption/Crimson/Dungeon.

One way to remove Happiness for your Town NPC is to add it to the `ActsLikeTownNPC[]` set. However, that comes with other side effects like the Town NPC not having a map icon and despawning when far away.

You can hide the Happiness button if you manually set `Main.LocalPlayer.currentShoppingSettings.HappinessReport = ""` in `GetChat()`, but that only stops the button from being drawn. Price calculations are still done.

With this property, you can easily make it so your Town NPC does not have the Happiness button and their prices are always neutral.

An alternative design would be to use `NPCID.Sets`, but I chose a `ModNPC` property instead so the field could be changed during the game (instead of just mod load time). For example, a mod wants their Town NPC to not be affected by Happiness after meeting a certain requirement such as defeating a boss. They can change `NPC.ModNPC.NoTownNPCHappiness` to true in something like `PreAI()`.

Another alternative design would be to make it an `NPC` field instead of `ModNPC`.

Updated Example Traveling Merchant to use this new property. Previously, all of their prices were at maximum because they were homeless. They also had a Happiness button unlike the vanilla Traveling Merchant which does not.

### ExampleMod updates
Updated Example Traveling Merchant. The Example Bone Merchant is already not affected by Happiness because it uses the `ActsLikeTownNPC[]` set.

### DIFF
--- a/ExampleMod/Content/NPCs/ExampleTravelingMerchant.cs
+++ b/ExampleMod/Content/NPCs/ExampleTravelingMerchant.cs
@@ -185,6 +185,7 @@ namespace ExampleMod.Content.NPCs
 			NPC.knockBackResist = 0.5f;
 			AnimationType = NPCID.Guide;
 			TownNPCStayingHomeless = true;
+			NoTownNPCHappiness = true;
 			CreateNewShop();
 		}
 
--- a/patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
@@ -40,6 +40,15 @@
  			_database = new PersonalityDatabase();
  			new PersonalityDatabasePopulator().Populate(_database);
  		}
+@@ -86,7 +_,7 @@
+ 				if (npc.type == 656 || npc.type == 637 || npc.type == 638)
+ 					return;
+ 
+-				if (IsNotReallyTownNPC(npc)) {
++				if (IsNotReallyTownNPC(npc) || npc.ModNPC?.NoTownNPCHappiness == true) {
+ 					_currentPriceAdjustment = 1f;
+ 					return;
+ 				}
 @@ -129,7 +_,7 @@
  					_currentPriceAdjustment *= 0.95f;
  				}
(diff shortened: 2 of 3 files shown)


Chicken-Bones:
I'd prefer a `Set`. If you absolutely need the ability to change the value at runtime for your mod, a `virtual` property would also be okay. No fields please, state is messy.

Rijam:
After some discussion on Discord ([link](https://discord.com/channels/103110554649894912/711551818194485259/1045210352897507418)), we discussed creating a virtual hook that would not only let you set if the Town NPC is affected by happiness or not, but also modify the price scaling.

Some pseudocode:
```cs
public override bool TownNPCHappiness(ref float priceAdjustment)
{
    priceAdjustment *= 1.5f; // increase the price and keep happiness
    return true;
}
// and
public override bool TownNPCHappiness(ref float priceAdjustment) => false; // remove all happiness (prices are at 1f)
```
This would come with a GlobalNPC version of course. This would allow people to make an item or furniture piece that globally makes Town NPCs happier.  I will make something in Example Mod showcasing this.

I will probably close this PR and make a new one targeting 1.4.4 because there are some changes with the happiness in the Don't dig up and Get fix boi seeds.

Chicken-Bones:
I think a `ref float priceAdjustment` hook is good. But I'd also be happy with a regular `bool` on `NPC` that can just be modified in `PreAI` or `SetDefaults` as desired.

Rijam:
Mostly covered in 7f2f4fb3a02c5b9a0567eb638c65d3464a4aeae4

JavidPack:
Ah, I guess it will be, didn't remember this PR when I was working on that. Thanks for keeping an eye on this feature

Rijam:
No worries. This PR needed a redesign and wasn't even targeting 1.4.4 lol.

PR #3063, opened 2022-10-29
Fix quick heal & mana consuming non-consumable items

Cyrillya:
### What is the bug?
Quick heal and quick mana will consume the potion, regardless of what `item.consumable` is
You can only disable the consume by overriding `ConsumeItem()` and return false, which is kinda annoying and confusing
Vanilla don't have this check since there's no something heals but doesn't consume, but for mods we need it
### How did you fix the bug?
Add `item.consumable` check:
```diff
- if (ItemLoader.ConsumeItem(item, this))
+ if (item.consumable && ItemLoader.ConsumeItem(item, this))
```


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -572,7 +572,7 @@
 +					ManaEffect(healMana);
  			}
  
-+			if (ItemLoader.ConsumeItem(item, this))
++			if (item.consumable && ItemLoader.ConsumeItem(item, this))
```
Solxanich: Instead of changing both of these locations, 
Add the ' not item.consumable' to ItemLoader.ConsumeItem method itself

Ensures the fix is applied in all locations.

Chicken-Bones: `ItemLoader.ConsumeItem` has 34 references. For quick mana/health it makes sense to check `item.consumable`, because the intent is to consume a potion or edible. 

For many others (placeables, gold keys, paint, wire, etc etc) it would be breaking change to check `item.consumable`

Chicken-Bones:
This PR is approved, but is being added to the 1.4.4 list, since it may break some mods and we need porting notes

Solxanich:
Hi there, 
We are currently working on putting together a build on 1.4.4, and noticed this was queued.

If you don't mind, it would be appreciated if you would re-target this to 1.4.4 branch?

Thanks! Please let us know if cant re-target

Cyrillya:
I've retargeted and updated the patch to 1.4.4

PR #3070, opened 2022-10-31
Item.NewItem overloads that takes Item, fix small moditem<->item issuees from TEs and dying.

JavidPack:
Currently, we have to juggle `Item.NewItem`, item cloning, and manually sending network syncing packets to spawn an item in the world while preserving modded data.  

Player.QuickSpawnClonedItem did it almost right, but TileEntities dropping their items and mediumcode players dropping items when dying had bad logic.

The bad logic was this:
```
				itemDrop.ModItem = item.ModItem;
				itemDrop.globalItems = item.globalItems;
```

This code neglects to set ModItem.Item/Entity to the actual item, resulting in the 2 way binding of Item and ModItem to be wrong.

Player.QuickSpawnClonedItem used the approach of using Item.Clone and then tweaking Item.position, but this approach actually lost many of the changes set in Item.NewItem, such as Item.velocity

Using the Clone approach, we would have to make sure to assign all fields modified intentionally by Item.NewItem:
```
int droppedItemId = NewItem(source, rectangle, item.netID, 1, noBroadcast: true, prefixGiven: item.prefix);
var droppedItem = Main.item[droppedItemId];

Item clone = item.Clone();
clone.whoAmI = droppedItemId;
clone.position = droppedItem.position;
clone.wet = droppedItem.wet;
clone.velocity = droppedItem.velocity;
clone.active = droppedItem.active;
clone.timeSinceItemSpawned = droppedItem.timeSinceItemSpawned;
clone.newAndShiny = droppedItem.newAndShiny;
clone.playerIndexTheItemIsReservedFor = droppedItem.playerIndexTheItemIsReservedFor;
Main.item[droppedItemId] = clone;

if (Main.netMode == 1)
	NetMessage.SendData(21, -1, -1, null, itemDropId);
```

Another issue with this is ItemLoader.OnSpawn in Item.NewItem acts on an Item instance that is discarded.

Setting `newItem.ModItem = original.ModItem;` seems like a mistake waiting to happen. 

#2977 bandaid fixes the original issue by setting `droppedItem.ModItem.Entity = droppedItem;`, but this approach won't fix the bigger issues.

Adding Item.NewItem overloads that accept an Item instance will simplify tmod and modder code. The code for Item.NewItem is split between the code creating the item instance and the code applying NewItem changes to that item instance. This allows the Item to be constructed from itemType/stack/prefix or from a clone of the passed in item.

### DIFF
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -188,10 +188,7 @@ namespace Terraria
 			if (NPCLoader.blockLoot.Contains(Type))
 				return Main.maxItems;
 
-			if (Type > 0 && cachedItemSpawnsByType[Type] != -1) {
-				cachedItemSpawnsByType[Type] += item.stack;
-				return 400;
-			}
+			// cachedItemSpawnsByType feature deliberately omitted from this NewItem
 
 			Main.item[400] = new Item();
 			int num = 400;
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -1,13 +1,10 @@
 using Microsoft.Xna.Framework;
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Reflection;
 using Terraria.DataStructures;
 using Terraria.ID;
 using Terraria.ModLoader;
 using Terraria.ModLoader.IO;
-using Terraria.Utilities;
 
 namespace Terraria
 {
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -1,10 +1,12 @@
 using Microsoft.Xna.Framework;
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using Terraria.DataStructures;
 using Terraria.ID;
 using Terraria.ModLoader;
 using Terraria.ModLoader.IO;
+using Terraria.Utilities;
 
 namespace Terraria
 {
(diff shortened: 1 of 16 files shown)


patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -124,11 +124,8 @@ public partial class Player : IEntityWithInstances<ModPlayer>
 		/// <param name="item">The item you want to be cloned</param>
 		/// <param name="stack">The stack to give the item. Note that this will override maxStack if it's higher.</param>
 		public int QuickSpawnClonedItem(IEntitySource source, Item item, int stack = 1) {
```
Chicken-Bones, line 128: Now that all `NewItem` calls perform cloning, this should be renamed.

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -124,11 +124,8 @@ public partial class Player : IEntityWithInstances<ModPlayer>
...
-			int index = Item.NewItem(source, getRect(), item.type, stack, false, -1, false, false);
-			Item clone = Main.item[index] = item.Clone();
-			clone.whoAmI = index;
-			clone.position = position;
-			clone.stack = stack;
+			int index = Item.NewItem(source, getRect(), item, false, false, false);
+			Main.item[index].stack = stack;
 
 			// Sync the item for mp
 			if (Main.netMode == NetmodeID.MultiplayerClient)
```
Chicken-Bones, line 133: This screams bugs to me. Items should not be spawned client side, as the selected `index` may be wrong.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5797,44 +5795,57 @@
...
  
  				inventory[i].TurnToAir();
-@@ -37735,6 +_,10 @@
+ 				if (i < armor.Length) {
+ 					if (armor[i].stack > 0) {
+-						int num2 = Item.NewItem(itemSource_Death, (int)position.X, (int)position.Y, width, height, armor[i].type);
++						int num2 = Item.NewItem(itemSource_Death, (int)position.X, (int)position.Y, width, height, armor[i]);
+-						Main.item[num2].netDefaults(armor[i].netID);
+-						Main.item[num2].Prefix(armor[i].prefix);
+-						Main.item[num2].stack = armor[i].stack;
```
Chicken-Bones: Comment instead of removing please

Chicken-Bones: (And in 3 places below)

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -170,6 +172,68 @@ public static int NewItem(IEntitySource source, Rectangle rectangle, int Type, i
...
+				return 400;
+			}
+
+			Main.item[400] = new Item();
+			int num = 400;
+			if (Main.netMode != 1)
+				num = PickAnItemSlotToSpawnItemOn(reverseLookup, num);
+
+			Main.timeItemSlotCannotBeReusedFor[num] = 0;
+			Main.item[num] = item.Clone();
```
Chicken-Bones: I'm not in favor of duplicating this code. Can we reuse the vanilla code with more clever method splitting?

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -170,6 +172,68 @@ public static int NewItem(IEntitySource source, Rectangle rectangle, int Type, i
...
+			if (Main.rand == null)
+				Main.rand = new UnifiedRandom();
+
+			if (NPCLoader.blockLoot.Contains(Type))
+				return Main.maxItems;
+
+			if (Type > 0 && cachedItemSpawnsByType[Type] != -1) {
+				cachedItemSpawnsByType[Type] += item.stack;
+				return 400;
+			}
```
Chicken-Bones: cached spawns should probably not work with the `Item` variant, only the `type` variant.

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -170,6 +172,68 @@ public static int NewItem(IEntitySource source, Rectangle rectangle, int Type, i
...
+			int Type = item.type;
+
+			if (WorldGen.gen)
+				return 0;
+
+			if (Main.rand == null)
+				Main.rand = new UnifiedRandom();
+
+			if (NPCLoader.blockLoot.Contains(Type))
+				return Main.maxItems;
```
Chicken-Bones: Hmm, I guess `blockLoot` is a discussion for another time. May as well let it work with this...

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Player.TML.cs
JavidPack: Clients spawn items in slot 400, then when the server receives the item message, spawn the item themselves with the proper index and syncs that to clients. This is how the existing Player.QuickSpawnItem works as well.

patches/tModLoader/Terraria/Player.cs.patch
JavidPack: Will do.

patches/tModLoader/Terraria/Player.TML.cs
JavidPack: Only the NewItem calls that we change to use Item will use cloning. This method is used by modders currently. There is a QuickSpawnItem that takes an Item, but it just passes in item.type to the normal QuickSpawnItem. This will be a breaking change, but it is unfortunate that the QuickSpawnItem with Item was added before the QuickSpawnClonedItem version was made.

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -1015,6 +1019,36 @@
...
++			if (itemToClone == null) {
+-			item.SetDefaults(Type);
++				item.SetDefaults(Type);
+-			item.Prefix(pfix);
++				item.Prefix(pfix);
+-			item.stack = Stack;
++				item.stack = Stack;
++			}
++			else {
++				Type = itemToClone.type; // just to be safe if IL edits change type swaps and velocity tweaks.
```
Chicken-Bones: Can you patch this with a goto?
```
if (itemToClone != null) {
    Main.item[num] = itemToClone;
    goto skipItemCreation;
}

patches/tModLoader/Terraria/Player.TML.cs
Chicken-Bones: Sounds like a good breaking change to me. We should just rename this, and make it use the `Item` variant of `NewItem`

patches/tModLoader/Terraria/Player.TML.cs
Chicken-Bones: Ah, I understand

Chicken-Bones (changes requested):

Issue #3067, opened 2022-10-31
NPC Shop Rework/Modernization


BasicallyIAmFox:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

A slight change to `(Mod|Global)NPC.SetupShop`.
Look lower for potential design.

### What does this proposal attempt to solve or improve?

It's currently pretty difficult to, for example, put X item before/after Y item.

So with this new feature, it would be much simpler, and be like that:
```
public override void SetupShop(int type, Chest shop, ChestLoot chestLoot)
{
    chestLoot.Put(ModContent.ItemType<ExampleItem>()); // Puts an Example Item at the end.
    chestLoot.PutBefore(ModContent.ItemType<ExampleItem>(), ModContent.ItemType<ExampleLifeFruit>()); // Puts Example Life Fruit right before Example Item.
    chestLoot.Remove(ModContent.ItemType<ExampleBuffPotion>()); // Remove Example Buff Potion from the shop if present.
    chestLoot.PutIf(ModContent.ItemType<ExampleSoul>(), () => Main.bloodMoon); // Puts an item if its Blood Moon. Use that method if you want better mod compatibility with other mods.
}
```

PR #3072, opened 2022-11-01
NPC Shop Modernization

BasicallyIAmFox:
### What is the new feature?
Modernizes NPC shops. #3067.

It could allow checking what items NPC can sell or not, better mod compatibility, and also allows easily insert item before/after X item, as well as adding some conditions to already existing items. Also avoids possible IndexOutOfRangeException, if too many items were added in single shop.

I was thinking of some sort of Database (similarly how Item and NPC loots has one), so all shops could be added after everything is ready, similarly to ItemDropDatabase.
It would also introduce new hook, called something like: `(Mod|Global)Item.InShop`. It would get called foreach item in shop when shop is opened. Would be useful for items those change something depending on something. (ie Tavernkeeper's shop. One of items there change its price depending on progression)

```
public override void SetupShop(int type, ChestLoot chestLoot)
{
    chestLoot.Put(ModContent.ItemType<ExampleItem>()); // Puts an Example Item at the end.
    chestLoot.PutBefore(ModContent.ItemType<ExampleItem>(), ModContent.ItemType<ExampleLifeFruit>()); // Puts Example Life Fruit right before Example Item.
    chestLoot.Remove(ModContent.ItemType<ExampleBuffPotion>()); // Remove Example Buff Potion from the shop if present.
    chestLoot.Put(ModContent.ItemType<ExampleSoul>(), () => Main.bloodMoon); // Puts an item if its Blood Moon. Use that method if you want better mod compatibility with other mods.
}
```

### ExampleMod updates
None yet. Will do later for ExamplePerson's shop, I guess.


Rijam:
Will there be a way to specify a custom shop price?
Currently, I can do something like this to specify a custom price:
```cs
shop.item[nextSlot].SetDefaults(ItemID.IronPickaxe);
shop.item[nextSlot].shopCustomPrice = 50000; // 5 gold
nextSlot++;
```
 An optional parameter would work. Something like `Put(int item, Func<bool> condition = null, customPrice = Item.value)` (pseudocode) is good enough for me.

BasicallyIAmFox:
you can provide an Item instance.
```
Item item = new(ItemID.IronPickaxe);
item.shopCustomPrice = 50000;
chestLoot.Put(item, ...);

// alt:
chestLoot.Put(ItemID.IronPickaxe, ...);
chestLoot[^1].item.shopCustomPrice = 50000;

// alt2:
chestLoot.Put(ItemID.IronPickaxe, ...);
chestLoot[ItemID.IronPickaxe].item.shopCustomPrice = 50000;
```
I could add `ItemBuilder` class, or something, that would "build" an Item instance without requiring to use `Index`es or local variables, but I feel like it isn't needed for single feature, lol. Considering that you'll usually want just to set price.

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Terraria.ID;
+
+namespace Terraria.ModLoader
+{
+	public readonly struct ChestLoot {
+		public struct Group {
```
Chicken-Bones: Preferred term is `Entry`, `Group` implies multiple `Item`s

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
...
+			}
+
+			public bool GetCondition() {
+				return condition?.Invoke() ?? true;
+			}
+		}
+
+		private static readonly Item emptyInstance = new();
+
+		private static object cache;
```
Chicken-Bones: What's the purpose of this?

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
...
+			cache = item;
+			if (CheckInvalidness(item) || items.FindIndex(group => group.item.type.Equals(cache)) == -1) {
+				return false;
+			}
+
+			candidatesByType.Add(item);
+			return true;
+		}
+
+		public bool Remove(Predicate<Item> match) {
```
Chicken-Bones: tML should never allow for removal of content, only disabling/hiding. Otherwise other mods may have their insertion locations messed up

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
...
+
+		public bool PutAt(int index, int item, Func<bool> condition = null) {
+			return PutAt(index, ContentSamples.ItemsByType[item], condition);
+		}
+
+		public bool PutAt(Predicate<Item> match, int item, Func<bool> condition = null) {
+			return PutAt(match, ContentSamples.ItemsByType[item], condition);
+		}
+
+		public bool Put(Item item, Func<bool> condition = null) {
```
Chicken-Bones: Use standard collections terminology. The only methods should be `Add`, `InsertAfter`, `InsertBefore`

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
...
+			if (Has(item)) {
+				return false;
+			}
+
+			items.Add(new(item, condition));
+			return true;
+		}
+
+		public bool PutAt(int index, Item item, Func<bool> condition = null) {
```
Chicken-Bones: Modders should not have to deal with indexes. Especially since they change with other insertions and removals. I suggest removing all the `index` based overloads.

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,217 @@
...
+		private static bool CheckInvalidness(int item) {
+			return item < ItemID.None || item >= ItemLoader.ItemCount;
+		}
+
+		private static bool CheckInvalidness(Item item) {
+			ArgumentNullException.ThrowIfNull(item, nameof(item));
+			return CheckInvalidness(item.type);
+		}
+
+		public bool Has(int item) {
```
Chicken-Bones: I'm not convinced that we should automatically prevent duplicate types. Not all `Item` with the same `type` are equivalent, a mod may want to sell two of the same `ModItem` with different custom data.

Chicken-Bones (changes requested):
First review, just looking at `ChestLoot` a bit

direwolf420:
Screenshot_16
The changes made to files in these directories seem to have nothing to do with the PR.

BasicallyIAmFox:
Yes, that is correct, @direwolf420 . They're there just cause of me switching to 1.4.4 branch I believe... I prolly have to wipe this branch completely and readd everything I added once again

PR #3078, opened 2022-11-03
Move & rename ModItem.SacrificeTotal

Mirsario:
This tiny PR does what the title says, and also makes EM consistently use the moved property. The reason for the former is a bit obvious - the utility property doesn't have anything to do with the Mod part of its containing type. XML descriptions were also updated, and the old property is kept around in obsoleted state.

To decide:
- ~~Should the property be renamed to something that mentions Research?~~
Renamed to `ReserachUnlockCount`.
- Should the dictionary that the property uses be ~~internalized~~ or turned ReadOnly?

TODOs:
- [ ] Add tModPorter rules.
- [x] Everything else.


Chicken-Bones:
Suggest `ResearchUnlockCount/ResearchRequiredCount/ResearchCount`

PR #3081, opened 2022-11-04
Made server manage script curl/wget agnostic

Freedbot:
### What is the bug?
The manage script is designed as an accessory to the docker container.  The container installs curl as a prerequisite for smooth operation.  Those people not using docker, but instead using the script directly will face an error if they're using a Linux distro that's packaged with wget instead of curl (IE: most flavors of Ubuntu, which is what I use).

### How did you fix the bug?
This tweaked manage-tModLoaderServer.sh will silently attempt to use curl first.  In case curl fails, the error is silenced and wget is silently attempted.  Failing wget is not silent, so failing both will produce a wget error and the script will still fail for a system with neither package present.

Unrelated: check_update and copy_worlds functions are condensed for simplicity, and a minor comment update.
Since the functionality is basically unchanged, the version increment is minimal.

Attempting to use alternative packages that already exist is the distro is ideal for this purpose.  On the other hand, Docker by design cannot access existing packages on the system, which is why curl is installed automatically.  With that said, a better solution to my method to reduce the chances for the script to fail even further would be to add another failsafe after wget fails to run a function to install curl for the user and try again.  I didn't add that because I'm treating Linux systems with neither installed as an outlier case, and it's more work.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
+++ b/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
@@ -42,8 +42,7 @@ function update_script {
 }
 
 function get_latest_release {
-	local 
-release_url="https://api.github.com/repos/tModLoader/tModLoader/releases/latest"
+	local release_url="https://api.github.com/repos/tModLoader/tModLoader/releases/latest"
 	local latest_release=$({
 		curl -s "$release_url" 2>/dev/null || wget -q -O- "$release_url";
 	} | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/') # Get latest release from github's api
--- a/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
+++ b/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
@@ -2,7 +2,8 @@
 
 #shellcheck disable=2164
 
-script_version="1.0.0.0"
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
```diff
@@ -49,21 +42,25 @@ function update_script {
 }
 
 function get_latest_release {
-	local latest_release
-	latest_release=$(curl --silent "https://api.github.com/repos/tModLoader/tModLoader/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/') # Get latest release from github's api
+	local 
+release_url="https://api.github.com/repos/tModLoader/tModLoader/releases/latest"
```
Solxanich: Please fix the formatting - this doesn't need a new line.

Everything else looks good, and seems to run ok, 
Great work making a clean fix.

Solxanich:
```suggestion
	local release_url="https://api.github.com/repos/tModLoader/tModLoader/releases/latest"
```
Please fix the formatting - this doesn't need a new line.

Everything else looks good, and seems to run ok, 
Great work making a clean fix.

Freedbot: I think that was just an artifact left from the tidy up.  Changed.

Solxanich:
Perfect. Thanks for the contribution!

PR #3085, opened 2022-11-07
1.4.4.7: ExtractinatorUse changes, reimplementation, porter, and examples

JavidPack:


ExampleMod/Common/GlobalItems/TorchExtractinatorGlobalItem.cs
```diff
@@ -0,0 +1,45 @@
...
+				return;
+
+			// If it is, we set stack to 1 and return a random torch. If the user is using the Chlorophyte Extractinator, we return Ultrabright Torch 10% of the time.
+			resultStack = 1;
+			if (extractinatorBlockType == TileID.ChlorophyteExtractinator && Main.rand.NextBool(10)) {
+				resultType = ItemID.UltrabrightTorch;
+				return;
+			}
+
+			resultType = Main.rand.Next(TorchExtractinatorModSystem.TorchItems);
```
Chicken-Bones, line 23: I believe from discussion (not from actually looking) that the chlorophyte extractinator _cycles_ through item pools, rather than randomly rerolling.

Chicken-Bones (changes requested):

PR #3086, opened 2022-11-08
1.4.4.7: Recipe shimmer examples and fixes

JavidPack:
Subsequent work will need to be done on new hooks, specifically hooks determining which recipe to decraft into similar to existing vanilla conditions limiting some decrafting to post skeletron. This will also facilitate Crimson/Corruption-type decisions as well.

Eventually we'll need an example of ShimmerCountsAsItem as well I think.


patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -408,15 +423,53 @@
...
++				}
++			}
++
++			ItemID.Sets.IsAMaterial[71] = false;
++			ItemID.Sets.IsAMaterial[72] = false;
++			ItemID.Sets.IsAMaterial[73] = false;
++			ItemID.Sets.IsAMaterial[74] = false;
++			ItemID.Sets.IsAMaterial[4076] = false;
++			ItemID.Sets.IsAMaterial[4131] = false;
++			ItemID.Sets.IsAMaterial[5325] = false;
```
Chicken-Bones: Why do we have to add these? Can you comment where they come from in the vanilla source? Maybe make a patch note there that links back to here?

Can mods replicate this?

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -280,6 +281,19 @@ public static class ConsumptionRules
...
+			return this;
+		}
+		public Recipe SetCrimson(bool crimson) {
+			this.crimson = crimson;
+			return this;
+		}
+		public Recipe SetCorruption(bool corruption) {
+			this.corruption = corruption;
+			return this;
+		}
```
Chicken-Bones: Hmm, these probably shouldn't have parameters. Maybe just `NotDecraftable()`, `CrimsonOnly()`, `CorruptionOnly()`? Perhaps `DisableShimmer()` to match `AddCustomShimmerResult`?

I feel like it would be better to have some sort of condition/category system to replace the `crimson/corruption` sets with. Since it's not really clear what these do.

Maybe something like `AddShimmerCondition`?

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -918,6 +925,15 @@
  			if (ItemID.Sets.BossBag[type])
  				return Color.Lerp(newColor, Color.White, 0.4f);
  
+@@ -47734,7 +_,7 @@
+ 				int num11 = FindDecraftAmount();
+ 				Recipe recipe = Main.recipe[decraftingRecipeIndex];
+ 				int num12 = 0;
+-				bool flag = recipe.requiredItem[1].stack > 0;
++				bool flag = recipe.requiredItem.Count > 1 && recipe.requiredItem[1].stack > 0;
```
Chicken-Bones: I think just `recipe.requiredItem.Count > 1` is correct here. This flag looks to be checking for multiple ingredients.

ExampleMod/Content/Items/Weapons/ExampleFlail.cs
```diff
@@ -49,9 +49,23 @@ internal class ExampleFlail : ModItem
...
+				.AddCustomShimmerResult(ItemID.RottenEgg)
+				.AddCustomShimmerResult(ItemID.Chain, 3)
+				.Register();
+
+			// By default, the last added recipe will be used for shimmer decrafting unless crimson or corruption specific recipes are found. We can use SetDecraftable(false) to tell the game to ignore this recipe.
 			CreateRecipe()
 				.AddIngredient<ExampleItem>()
+				.AddIngredient(ItemID.PadThai)
 				.AddTile<Tiles.Furniture.ExampleWorkbench>()
+				.SetDecraftable(false)
```
Chicken-Bones: I'd rather we have separate examples for everything shimmer/decraftable related.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Recipe.cs.patch
JavidPack: Added comments. Modders can use PostSetupRecipes to change these.

PR #3088, opened 2022-11-09
1.4.4.7: DefaultToX usage in ExampleMod, Documentation

JavidPack:
Changes ExampleMod to use various helper methods wherever useful.

Removing repetitive lines focuses the example on the relevant information, and helps prevents typo accidents for modders making similar content.

image

Documentation for these methods is now very verbose.

Also changed links from gamepedia to https://terraria.wiki.gg/

### Existing concerns:
DefaultToMusicBox -- Assumes `createTile = 139;`. We could change this to take a createTile, or just change ExampleMusicBox to use it and then set createTile directly
DefaultToTorch -- Same as DefaultToMusicBox, assumes `createTile = 4;`
Many things still don't have a DefaultToX
- Potions -- Would help modders avoid the mistake of forgetting to set either buffType or buffTime
- Swords
- Tools

Code a bit verbose:
- `Item.DefaultToPlaceableTile(ModContent.TileType<Tiles.ExampleLamp>());` could benefit from a generic overload: `Item.DefaultToPlaceableTile<Tiles.ExampleLamp>();`. Unsure.


Chicken-Bones:
Do change `DefaultToMusicBox` and the like to take a defaulting `createTile` parameter. Seems helpful.

`Item.DefaultToPlaceableTile(ModContent.TileType<Tiles.ExampleLamp>());` is fine for now. Modders that care about verbosity can static import `ModContent` or make their own overloads

Potions / Swords / Tools
- May as well, why not?

JavidPack:
Surpassed by #3098

PR #3089, opened 2022-11-10
Added catch conditions for ExampleQuestFish

Sneakilli:
Added conditions and checks to allow ExampleQuestFish to be caught.


ExampleMod/Common/Players/ExampleFishingPlayer.cs
```diff
@@ -63,8 +63,26 @@ public class ExampleFishingPlayer : ModPlayer
...
 				if (!attempt.veryrare && !attempt.legendary && attempt.rare && Main.rand.NextBool()) {
 					itemDrop = ModContent.ItemType<Content.Items.Consumables.ExampleFishingCrate>();
+					return; // This is important so your code after this that rolls items will not run
 				}
 			}
+
+			// Here we will set the catch conditions for our ExampleQuestFish
+			int exampleQuestFish = ModContent.ItemType<Content.Items.ExampleQuestFish>(); // We'll store the type as a variable, since we'll be referencing it several times
+			// First we check if today's quest matches our quest fish
+			if (attempt.questFish == ModContent.ItemType<Content.Items.ExampleQuestFish>()) {
```
direwolf420:
```suggestion
			// Here we will set the catch conditions for our ExampleQuestFish
			int exampleQuestFish = ModContent.ItemType<Content.Items.ExampleQuestFish>(); // We'll store the type as a variable, since we'll be referencing it several times
			// First we check if today's quest matches our quest fish
			if (attempt.questFish == exampleQuestFish) {
```

ExampleMod/Common/Players/ExampleFishingPlayer.cs
```diff
@@ -63,8 +63,26 @@ public class ExampleFishingPlayer : ModPlayer
...
+				// Now we'll check if the player already has the fish, and if they've completed today's quest yet
+				if (!Player.HasItem(exampleQuestFish) && !Main.anglerQuestFinished) {
```
direwolf420: This seems to be a regression from 1.3... Does vanilla not check for these two latter conditions anymore?
1.3 code used to just be
```cs
if (questFish == exampleQuestFish /*&& other conditions*/) {
    itemType = exampleQuestFish;
}
```

direwolf420 (changes requested):
[tModLoader - Shortcut.lnk](https://github.com/tModLoader/tModLoader/pull/3089/files#diff-77782b75980aa72ec7dadc02525bc5d30cb0e7795d8015cabbcc1ab514addb2e)
What does this file do? Seems unrelated to the PR.

ExampleMod/Common/Players/ExampleFishingPlayer.cs
Sneakilli: You are right, vanilla does account for this. I misunderstood the code and have made a commit to remove the extra checks.
The 1.3 code was at some point removed. This PR exists because the catching of exampleQuestFish was impossible.

Sneakilli:
> [tModLoader - Shortcut.lnk](https://github.com/tModLoader/tModLoader/pull/3089/files#diff-77782b75980aa72ec7dadc02525bc5d30cb0e7795d8015cabbcc1ab514addb2e) What does this file do? Seems unrelated to the PR.

Not sure why it existed in my repo, or why I didn't notice it. Either way I've removed it.

Chicken-Bones:
Hm, forgot I need to retarget this to 1.4.4, I'll cherry pick

PR #3096, opened 2022-11-17
Fix Corsair RGB

### LINKED ISSUE #1785: [1.4 MergedTesting] RGB Lighting Integration doesn't work

Antirhinnum:
### What is the bug?
Corsair RGB would not function on any version of tModLoader. 

### How did you fix the bug?
Updated the iCUE SDK dll to a newer 64-bit version and updated the interop methods to use the new DLL.
Partially fixes #1785.

The fix uses the 2019 version of Corsair's iCUE SDK, while Terraria uses the 2015 version. There does exist a 64-bit version of the 2015 dll, which may be preferred over the 2019 dll.

### Other
Newer DLL from: https://github.com/CorsairOfficial/cue-sdk/releases/tag/v3.0.464
In my testing, this works perfectly fine on tModLoader. However, this does not run on TerrariaNetCore for the same reason an earlier Razer Chroma fix (#2821) doesn't. (Specifically, the Razer Chroma dll isn't found and crashes the game.) However, the [fix for that issue](https://github.com/tModLoader/tModLoader/commit/98f990168bcf7000f99948e85ea153c29f152c5b) (which also fixes this one) was committed to tModLoader's patches instead of TerrariaNetCore's and marked as temporary.


Chicken-Bones:
Looks good to me. @Mirsario can you make sure the right patches in the right steps get sorted out?

Issue #3074, opened 2022-11-01
Major Localization Changes


JavidPack:
# Localization Changes
This is a proposal for large localization changes that will make localization easier to manage and use.

# Goals
In order of importance:
1. Maintaining translation files is too hard. Translation files shouldn't require any effort to start using for normal content.
    - There is no process to generate them from existing content
    - There is no process to update them with new content
    - Migrating to using them to support localization is a large endeavor
    
2. Contributing translations or making a translation mod is hard:
    - Extracting translations from closed source mods will let translators translate mods. 
    - Testing translations needs to be easier, since custom translation keys sometimes require guesswork for translators (context of where the key is used is not immediately apparent)

3. Translation keys are verbose and inconsistent, they need to be more consistent and nested below the content they correspond to rather than nested below other concepts:
    -  Current key patterns follow vanilla patterns, which separates translations inherently bound together:
        * `ModItem` currently has `DisplayName` set to `Mods.ModName.ItemName.ClassName` and `Tooltip` set to `Mods.ModName.ItemTooltip.ClassName`
        * Town NPC have translations all over: `Mods.ModName.TownNPCMood.NPCName`, `Mods.ModName.Bestiary.NPCName`, `Mods.ModName.Dialogue.NPCName`

4. Too much indentation. Despite hjson being simple syntax, hjson files can look ugly and disorganized with multiline strings and inconsistent tabs and spaces.

5. Simplify backend code:
    - Remove `ModTranslation` class. 
      - All content now only know how to retrieve their associated translation key, they do not store the translation values themselves.
      - Use `LocalizedText` instead, use vanilla approach to loading translations (only load and keep the current language in memory)
    - Automatically handle `LocalizedText` key construction and default value generation.

# Proposed Implementation
Each item corresponds to the numbered goal above:

---

tldr: Code based translations are no-longer supported (`ModTranslation.SetDefault`), all localized text must have a key in an English `.hjson` file. All localization files will be automatically populated.

---

1. After building a mod, all localization files for all languages supported by the mod will be automatically populated with all values after mod loading finishes:
    - tModLoader will track a list of recently built mod names. After mod loading, if a mod contained in that list matches a recently built mod, the localization file updating process will occur for that mod. The name is then removed from that list.
[...]

PR #3101, opened 2022-11-21
1.4.4 localization overhaul

JavidPack:
This is a work in progress implementation of #3074


Linnnnnna:
@JavidPack 
Will it inherit translated text when Mod updated?
And will the inherited text also become English when the Mod's text changed?
And And will it support/allow one pack translate multiple mods?
"ModInfoExtractLocalization": "提取本地化",
"ModInfoExtracted": "已提取",
"ModInfoEnableModToExtractLocalizationFiles": "启用模组以提取本地化文件",

4J5FU8QV4}@@_)SP{B( 43

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
```diff
@@ -1,22 +1,62 @@
...
 public partial class GameCulture
 {
-	public static GameCulture FromPath(string path)
+	/// <summary>
+	/// Derives a culture and shared prefix from a localization file path. Prefix will be found after culture, either separated by an underscore or nested in the folder.
+	/// <br/> Some examples:<code>
+	/// Localization/en-US_Mods.ExampleMod.hjson
+	/// Localization/en-US/Mods.ExampleMod.hjson
+	/// en-US_Mods.ExampleMod.hjson
+	/// en-US/Mods.ExampleMod.hjson
```
Chicken-Bones: Did we want to support `en-US/Mods/ExampleMod.hjson`?

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
```diff
@@ -1,22 +1,62 @@
...
-		string[] split = path.Split("/");
+		GameCulture culture = null;
+		string prefix = null;
+
+		string[] splitByFolder = path.Split("/");
+		for (int folderSplitIndex = 0; folderSplitIndex < splitByFolder.Length; folderSplitIndex++) {
+			string pathPart = splitByFolder[folderSplitIndex];
+
+			string[] splitByUnderscore = pathPart.Split("_");
+			for (int underscoreSplitIndex = 0; underscoreSplitIndex < splitByUnderscore.Length; underscoreSplitIndex++) {
```
Chicken-Bones: `foreach` preferred since we're rewriting this method anyway.

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
```diff
@@ -1,22 +1,62 @@
...
+	/// en-US/Mods.ExampleMod.hjson
+	/// </code>
+	/// </summary>
+	/// <param name="path"></param>
+	/// <returns></returns>
+	public static (GameCulture culture, string prefix) FromPath(string path)
 	{
 		path = Path.ChangeExtension(path, null);
 
-		string[] split = path.Split("/");
```
Chicken-Bones, line 12: May as well put this with the commented out code below

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
Chicken-Bones: Another option is a regex `(?:^|/)([a-zA-Z-]+)(?=(?:[/_](\w+))?)` try it out in regex101. Don't necessarily think it's more readable, but possibly easier to adjust.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+					RecurseThrough(entry.Qo() as WscJsonObject, newPrefix);
+				}
+				else if (item.Value.JsonType == JsonType.String) {
+					var localizationValue = item.Value.Qs();
+					string key = string.IsNullOrWhiteSpace(prefix) ? item.Key : prefix + "." + item.Key;
+					int actualOrderIndex = index - 1;
+					string actualCommentKey = actualOrderIndex == -1 ? "" : original.Order[actualOrderIndex];
+					string comment = original.Comments[actualCommentKey];
+					existingKeys.Add(new(key, localizationValue, comment));
+				}
```
Chicken-Bones: A lot of common code between these branches.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+	}
+
+	private static List<LocalizationEntry> FlattenJsonObject(WscJsonObject jsonObjectEng, string prefix)
+	{
+		// TODO: How should "$parentVal" be handled?
+		// TODO: Which entry should this comment attach to in the result, if it ends up being expanded?
+		//       # Some Comment on ExampleMod.Common
+		//       ExampleMod.Common: {...}
+
+		var newJsonObject = new WscJsonObject();
```
Chicken-Bones: unused, presumably left-over from experimenting

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+		// Clean up orphaned non-default language files, if any.
+		foreach (var remainingLocalizationFiles in allLocalizationFilesAllLanguages) {
+			string originalPath = Path.Combine(sourceFolder, remainingLocalizationFiles);
+			string newPath = originalPath + ".legacy";
+
+			File.Move(originalPath, newPath);
+		}
+	}
+
+	private static List<LocalizationEntry> FlattenJsonObject(WscJsonObject jsonObjectEng, string prefix)
```
Chicken-Bones: perhaps a rename now, `ParseLocalizationEntries`?

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+
+	private static void UpdateLocalizationFilesForMod(Mod mod, string outputPath = null, GameCulture specificCulture = null)
+	{
+		string sourceFolder = outputPath ?? Path.Combine(ModCompile.ModSourcePath, mod.Name);
+		if (!Directory.Exists(sourceFolder))
+			return;
+
+		// TODO: Maybe optimize to only recently built?
+
+		var localizationFileEntries = new List<LocalizationFileEntry>();
```
Chicken-Bones: suggest `baseLocalizationFiles`

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+			(var culture, string prefix) = GameCulture.FromPath(translationFile.Name);
+
+			allLocalizationFilesAllLanguages.Add(translationFile.Name);
+			allLanguages.Add(culture);
+
+			// TODO: Support arbitrary default language?
+			// Default language hjson files loaded into memory to gather comments and modder intended ordering.
+			if (culture == GameCulture.FromCultureName(GameCulture.CultureName.English)) {
+				JsonValue jsonValueEng = HjsonValue.Parse(translationFileContents, new HjsonOptions() { KeepWsc = true });
+				JsonObject jsonObjectEng = jsonValueEng.Qo();
```
Chicken-Bones: Not necessary. Can cast `jsonValueEng` directly to `WscJsonObject`.
Also prefer cast over `as` because `FlattenJsonObject` does not accept nulls, and `ClassCastException` is more informative

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+		var existingKeys = new List<LocalizationEntry>();
+		// Collect known keys. These are potentially missing from the localization files
+		foreach (var translation in translations) {
+			if (translation.Key.StartsWith($"Mods.{mod.Name}.")) {
+				existingKeys.Add(new(translation.Key, translation.Value.GetDefault(), null));
+			}
+		}
+
+		// Merge collected keys into flattened in-memory model
+		foreach (var existingKey in existingKeys) {
```
Chicken-Bones: Merge these two loops?

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+			// TODO: detect common folder path and use that?
+			mostSuitableLocalizationFile = new("en-US.hjson", "", new List<LocalizationEntry>());
+			localizationFileEntries.Add(mostSuitableLocalizationFile);
+			//throw new Exception("Somehow there are no files for this key");
+		}
+
+		return mostSuitableLocalizationFile;
+	}
+
+	internal static int CheckPrefixExistsLevel(List<LocalizationEntry> localizationEntries, string key)
```
Chicken-Bones: Suggest `LongestMatchingPrefix`

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+			if (translation.Key.StartsWith($"Mods.{mod.Name}.")) {
+				existingKeys.Add(new(translation.Key, translation.Value.GetDefault(), null));
+			}
+		}
+
+		// Merge collected keys into flattened in-memory model
+		foreach (var existingKey in existingKeys) {
+			LocalizationFileEntry suitableHJSONFile = FindHJSONFileForKey(localizationFileEntries, existingKey.key);
+
+			if (!KeyExistsInHJSON(suitableHJSONFile, existingKey.key)) {
```
Chicken-Bones: prefer to instead build a `Set<string>` of all default translations, and check for existence in that set before `FindHJSONFileForKey`. That way we only do work for 'new' translations.

Chicken-Bones: It also looks like there's some shared code between `CheckPrefixExistsLevel` and `AddEntryToHJSON` in that returning the common prefix would be helpful.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+		if(specificCulture != null) {
+			allLanguages.Clear();
+			allLanguages.Add(specificCulture);
+		}
+
+		// Update all languages that have been found in the mod
+		foreach (var culture in allLanguages) {
+			// Save all localization files
+			foreach (var localizationFileEntry in localizationFileEntries) {
+				// TODO: Don't save to disk if unchange optimization
```
Chicken-Bones: Definitely worthwhile. Should be able to build a set of which files actually changed above.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+
+		if(specificCulture != null) {
+			allLanguages.Clear();
+			allLanguages.Add(specificCulture);
+		}
+
+		// Update all languages that have been found in the mod
+		foreach (var culture in allLanguages) {
+			// Save all localization files
+			foreach (var localizationFileEntry in localizationFileEntries) {
```
Chicken-Bones: Please extract the body of this loop to another method

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+					var entry = localizationFileEntry.LocalizationEntries[i];
+
+					CommentedWscJsonObject parent = rootObject;
+
+					string key = entry.key;
+					if (!string.IsNullOrWhiteSpace(localizationFileEntry.prefix)) {
+						key = key.Substring(localizationFileEntry.prefix.Length + 1);
+					}
+
+					// Find/Populate the parents of this translation entry
```
Chicken-Bones: Does this always create an object tree? No support for flattening/expansion? Do we want to auto-flatten parents with only one child as a post-pass?

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+					// Populate parent object with this translation, manipulating comments to appear above the entry.
+
+					if (entry.value == null && entry.type == JsonType.Object) {
+						if (parent.Count == 0) {
+							parent.Comments[""] = entry.comment;
+						}
+						else {
+							string actualCommentKey = parent.Keys.Last();
+							parent.Comments[actualCommentKey] = entry.comment;
+						}
```
Chicken-Bones: duplicate with below, helper method?

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+						parent.Add(key.Split(".").Last(), new CommentedWscJsonObject());
+					}
+					else {
+						// Add values
+
+						string value = translations[entry.key].GetTranslation(culture);
+						key = key.Split(".").Last();
```
Chicken-Bones: you actually have `splitKey` above. So you can use `splitKey[^1]`

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+		if (specificCulture != null)
+			return;
+
+		// Clean up orphaned non-default language files, if any.
+		foreach (var remainingLocalizationFiles in allLocalizationFilesAllLanguages) {
+			string originalPath = Path.Combine(sourceFolder, remainingLocalizationFiles);
+			string newPath = originalPath + ".legacy";
+
+			File.Move(originalPath, newPath);
+		}
```
Chicken-Bones: Hmm, this gets a bit tricker with the "don't update files which didn't change" feature, but I have an idea.

This only deletes files which don't have an English counterpart, so you could attempt to map back the other way (find an english file with the appropriate name), or map all the english files to each culture to generate the full set of known paths.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -286,4 +295,317 @@ private static LocalizedText SetLocalizedText(Dictionary<string, LocalizedText>
...
+					if (!string.IsNullOrWhiteSpace(localizationFileEntry.prefix)) {
+						key = key.Substring(localizationFileEntry.prefix.Length + 1);
+					}
+
+					// Find/Populate the parents of this translation entry 
+					string[] splitKey = key.Split(".");
+					for (int j = 0; j < splitKey.Length - 1; j++) {
+						string k = splitKey[j];
+						if (parent.ContainsKey(k))
+							parent = (CommentedWscJsonObject)parent[k];
```
Chicken-Bones: Calling back to slack discussion. What if one of the 'new' translations (from a new piece of content I just added) has a parent which already has a value? It will be a `JsonValue` rather than a `CommentedWscJsonObject`

We would need to support `"$parentVal"` or a preferred alternative

patches/tModLoader/Terraria/ModLoader/UI/UICreateMod.cs
```diff
@@ -291,6 +295,15 @@ internal string GetLaunchSettings()
 }}";
 	}
 
+	internal string GetLocalizationFile(string modNameTrimmed)
+	{
+		return $@"Mods: {{
```
Chicken-Bones: newline after return to match style of other block text

patches/tModLoader/Terraria/ModLoader/Utilities/HjsonExtensions.cs
```diff
@@ -71,23 +72,42 @@ private static void WriteFancyHjsonValue(TextWriter tw, JsonValue value, int lev
...
+						lastJsonType = val.JsonType;
+						
 						tw.Write(kwl);
 
-						NewLine(tw, level + 1);
+						NewLine(tw, level + (showBraces ? 1 : 0));
+
 						kwl = GetComments(commentedObject.Comments, key);
 
+						bool commentedOut = jObject is CommentedWscJsonObject commented && commented.CommentedOut.Contains(key);
```
Chicken-Bones, line 100: I don't love this approach, the code doesn't really go here, and the hjson object will not come back the same when deserialized again. 

It doesn't hurt because we only do it in generated files but...

Could we not generate the comments directly when we generate the files?

patches/tModLoader/Terraria/ModLoader/Utilities/HjsonExtensions.cs
```diff
@@ -71,23 +72,42 @@ private static void WriteFancyHjsonValue(TextWriter tw, JsonValue value, int lev
...
+						if (commentedOut)
+							tw.Write("/* ");
```
Chicken-Bones: Using block comments to distinguish from user comments?
Can I suggest using `//` (leaving `#` for actual user comments)

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Localization/GameCulture.TML.cs
JavidPack: I think that'll get ugly, repeating the modname like that. Lets leave this idea for later if requested.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
JavidPack: I'll leave a TODO for this one. We previously had the idea to specify the flatness per-modtype, but maybe just saying objects with 1 and 2 entries is sufficient.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
JavidPack: Implemented the optimization for existing elements, I know a lot of code in this area changed in the other branch so I'll look into not doing the same work twice once they are one.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
JavidPack: nevermind, I implemented this with a current check for just 1 entry, we can change to 2 if we want.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
JavidPack: I can't see a way around a special key like $parentVal, other than not allowing keys like that. I'm going to hold off on implementing this one while we focus on getting everything else working.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
JavidPack: This does that, the only files left in here will be those that didn't match up to an english file.

patches/tModLoader/Terraria/ModLoader/Utilities/HjsonExtensions.cs
JavidPack: I had to use block comments because many translations will be multiple lines. This ensures that they are read correctly when loading the mod.

patches/tModLoader/Terraria/ModLoader/Utilities/HjsonExtensions.cs
JavidPack: Guess we should talk about this one. This is where we are generating the files, doing it after this method would require writing parsing code

Chicken-Bones: I wanted to generate actual comments instead, before the writer, but I guess you can't because you need the `jObject` to be formatted by the writer...

What does an ExampleMod output look like for a new language?

Setnour6:
I assume you know this already but just in case, the newest tModLoader post link saying "Re-design Mod Browser & Back-end" directs to this PR, which is about the 1.4.4 localization overhaul.
Did you mean PR 3346, [Re design Mod Browser & Back-end](https://github.com/tModLoader/tModLoader/pull/3346)?

JavidPack:
Thanks, fixed

PR #3107, opened 2022-11-24
Fix tile framing (causing infinite UnloadedXTile loop and invisible tiles)

direwolf420:
### What is the bug?
As of the changes done to how unloaded tiles are handled back at the start of 1.4, unloaded tiles can corrupt when turning from valid mod tiles into unloaded tiles, causing them to recurse on `TileFrame`.
And in both 1.3 and 1.4, changes done to height related fields of a `TileObjectData` associated with a modded tile could "brick" the tile drawing if the previously placed tile is in screen range, causing many tiles to turn invisible. This PR fixes both issues.

#### Technical Breakdown
* Tile with atleast 2 height, where the non-bottom coordinate height is bigger than 16
* Place tile, remove tile from the mod/unload mod
* On world load, it converts to unloaded tile that has "conventional" 16 height + 2 padding
* `frameX/Y` persists 1:1, meaning that for the "row" below the one that has coordinate height bigger than 16, it causes partFrameY to be > 0
* Which makes it continue further down (into `CoordinateHeights`) which then gets out of bounds as the unloaded tile only has 1 height

### How did you fix the bug?
The fix is to use the code from tall gates (`WorldGen.CheckTallGate`), which doesn't do the "scan down" step if the resulting Y would be more than the tiles height.
This fix is also convenient for modders who change tiles post-release. Previously, if you changed `Height` or `CoordinateHeights` of a tile that is already placed in the world, the above issues will occur to that tile aswell.
Now, after the modder changes the tile, the old placed tile might look off, but it will not break the game anymore. Breaking and placing it back will make it work properly. Previously on 1.3 you were forced to tell people who experience issues with your tile to unload the mod, then break the unloaded tiles (and ironically, those instructions currently don't work on 1.4 for the reasons described above).

The vanilla code i referenced does not use `+ tileData.CoordinatePadding` in its calculations, but it was included in the modded code so I kept that.


Mirsario (approved):
Confirmed as working.

Solxanich:
Tested, validated the change, and no concerns

Issue #3109, opened 2022-11-27
Menu select button gets physically pushed to the left (if there's too much raw text?)


QUIDD60x:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Using a lot of text on the modded menu swap button will physically move the button to the left. This issue is also affected by the text in color codes ( i presume), meaning a moved menu button for no apparent reason to the naked eye.

### Expected Behavior

Menu swap button should remain in place, at the bottom center.

### Actual Behavior

Menu button is pushed to the left varying amounts, based on how much text you have.

### Reproduction frequency

yes.

### Additional Information

I would attach an image but I can't, so here's https://imgur.com/4HKuHiY as a refrence image of what I'm talking about.

QUIDD60x:
For any future refrence, I have `public override string DisplayName => "[c/5EAFF1:Y][c/74acda:o][c/82abcc:u][c/96a9b8:r] [c/c1a48d:T][c/c1a48d:a][c/e7a067:l][c/e7a067:e] menu";` as the menu name, which is probably why it gets pushed to the left.

MacaylaMarvelous81:
I think that the color codes are included when measuring the string but not actually displayed, causing the displacement.

QUIDD60x:
> I think that the color codes are included when measuring the string but not actually displayed, causing the displacement.

Well I thought that much, I just don't know if there's a possible fix or not. The menu select button ends up really messed up, but if there isn't a fix then oh well.

PR #3111, opened 2022-11-28
WIP: Parse text in ModMenu display name before measuring it

MacaylaMarvelous81:
### What is the bug?
#3109 - The raw text of the ModMenu's display name is measured, but the raw text might not be exactly what's displayed. For example, color codes are not displayed when the text is drawn, but they are included when measuring the string. This can cause an offset when the text is drawn.

### How did you fix the bug?
I parsed the text with `ChatManager.ParseMessage` before measuring it.

There's probably a better way to get the resulting text when parsing the string, or another approach entirely such as removing color codes from the string.

Patch in action (sorry for using your mod's display name @QUIDD60x !)
Before
4HKuHiY
After
parsedtext


JavidPack:
Can you change this to use `ChatManager.GetStringSize` instead, as we usually do, and make sure it still works?

MacaylaMarvelous81:
OK, this is how it looks:
parsedtext2

QUIDD60x:
that works well, thank you!

PR #3117, opened 2022-12-03
[1.4.4] Boss health bar update

direwolf420:
### What is the new feature?
* Implement `BigProgressBarInfo.showText` functionality
    * Before it was just set to true, and never accessed. Now, it's set to true every update, and mods can reset it. It's combined with `&& BigProgressBarSystem.ShowText` (a config setting) for the code that determines if text can draw.
    * The vanilla callsite of the latter was _not_ changed, only the tml roundabout.
    * Naturally, the API was expanded to accomodate for this functionality, with a bonus offset added (which required `DrawHealthText` to be cloned into separate method). This was done in a breaking fashion, since 1.4.4 is also breaking all mods.
* Updated the tml methods to match vanilla (besides the text, nothing changed)
* Update comments for proper formatting
    * Some comments were not having a space after the `//`
* Fix for ExampleMods GlobalBossBar text overlapping with the health text

1.4.4 added "life/lifeMax" display on health bars, they should be supported by tml and customized by mods.

`DrawHealthText` was copied and publicised in `BossBarLoader`. An alternative would be to keep it in the original class it came from. EDIT: This was done.
A more generic `DrawHealthText` can be added that takes in one or multiple strings (i.e. to include the bosses' name), and a color. It would be easy to implement.

The ExampleMod example showcases usage of `info.showText` and `BossBarLoader.DrawHealthText`,
Mods may want to offset, hide, or re-draw the text that displays on the boss. For the latter, more methods may be needed (see the above category).

### Porting Notes
(Not really worth mentioning due to 1.4.4 breaking everything anyway) `BossBarDrawParams` now has `ShowText` and `TextOffset` fields, the only constructor/deconstructor was modified for this, so you need to recompile.

### DIFF
--- a/ExampleMod/Common/GlobalBossBars/ExampleGlobalBossBar.cs
+++ b/ExampleMod/Common/GlobalBossBars/ExampleGlobalBossBar.cs
@@ -25,7 +25,7 @@ namespace ExampleMod.Common.GlobalBossBars
 				string text = "GlobalBossBar Showcase";
 				var font = FontAssets.MouseText.Value;
 				Vector2 size = font.MeasureString(text);
-				//Draw centered on the boss bar, offset upwards, otherwise it will overlap with the health text
+				// Draw centered on the boss bar, offset upwards, otherwise it will overlap with the health text
 				spriteBatch.DrawString(font, text, drawParams.BarCenter - size / 2 + new Vector2(0, -30), Color.White);
 			}
 		}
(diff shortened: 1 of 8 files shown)


Chicken-Bones:
Lets keep the `DrawHealthText` overload in the original class (with a `.TML.cs` `partial`). That's consistent with our other approaches to modder friendly vanilla overloads (see `Item.TML.cs`) and makes it easier to keep in sync.

Good stuff.

PR #3123, opened 2022-12-06
[1.4.4] Modded chest lock support

Jadams505:
### What is the new feature?
Added a `LockChest` hook so modded chests can be locked by the chest lock item added in 1.4.4

1.4.4 added the chest lock item so it would be helpful if modded chests could have support for it easily

I don't know I basically just mimicked `UnlockChest`
Included in `ExampleChest`

### ExampleMod updates
Implemented `LockChest` for `ExampleChest`

### Notes
There is a vanilla bug where because the chest lock item is used on tiles in `TileID.Sets.IsAContainer` it can be used on dressers as well. I added `ExampleChest` to that set, but depending on how vanilla fixes the bug (I'm assuming changing to `TileID.Sets.BasicChest`) might be unnecessary


JavidPack:
Looks great, thanks for this, I didn't even know the Chest Lock existed.

PR #3130, opened 2022-12-09
[1.4.4] Log worldgen seed & consistency with existing world loading messages

direwolf420:
### What is the new feature?
Vanilla 1.4.4 prints the world generation seed before the world is created in the server console. This PR changes it to a combined call to include both server console and logs. Also added a missing comma after the seed.
This PR also updates the existing "Loading/Entering World" messages with the new info (expert/world evil).

Seeing the seed in the log helps debugging world gen issues that users report (especially when they didn't deliberately choose a seed). The other stuff is for consistency with the vanilla info.

The issue with seeds currently is that special seeds (such as "notthebees") override it - by checking the original seed contents, then setting flags based on that, and making a new random seed. The latter is what is shown in the log. TML could include the original seed in that message aswell, but it would provide limited advantages becase you can't actually create the same world based on those two values as the game doesn't allow you to input both of those seeds on world creation. CB mentioned that separating these two is on Re-Logics radar, so this would need to be monitored accordingly.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -49,7 +49,7 @@
  
  		public static void EnterWorld(int playerIndex)
  		{
-+			Logging.Terraria.InfoFormat("Entering world with player: {0}, IsCloud={1}, Size={2}x{3}, Evil: {4}, IsExpert: {5}", Main.ActivePlayerFileData.Name, Main.ActivePlayerFileData.IsCloudSave, Main.maxTilesX, Main.maxTilesY, WorldGen.crimson.ToInt(), Main.expertMode);
++			Logging.Terraria.InfoFormat("Entering world with player: {0}, IsCloud={1}, Width: {2}, Height: {3}, Evil: {4}, GameMode: {5}", Main.ActivePlayerFileData.Name, Main.ActivePlayerFileData.IsCloudSave, Main.maxTilesX, Main.maxTilesY, WorldGen.crimson.ToInt(), Main.GameMode);
 +			Interface.ResetData();
 +
  			if (Hooks.OnEnterWorld != null)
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -49,7 +49,7 @@
  
  		public static void EnterWorld(int playerIndex)
  		{
-+			Logging.Terraria.InfoFormat("Entering world with player: {0}, IsCloud={1}, Size={2}x{3}", Main.ActivePlayerFileData.Name, Main.ActivePlayerFileData.IsCloudSave, Main.maxTilesX, Main.maxTilesY);
++			Logging.Terraria.InfoFormat("Entering world with player: {0}, IsCloud={1}, Size={2}x{3}, Evil: {4}, IsExpert: {5}", Main.ActivePlayerFileData.Name, Main.ActivePlayerFileData.IsCloudSave, Main.maxTilesX, Main.maxTilesY, WorldGen.crimson.ToInt(), Main.expertMode);
```
Chicken-Bones: Why log `expertMode` and not `master` or `journey`? Consider `GameMode` instead?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -410,7 +410,12 @@
  		remixWorldGen = tempRemixWorldGen;
  		tenthAnniversaryWorldGen = tempTenthAnniversaryWorldGen;
  		drunkWorldGen = false;
-@@ -6034,6 +_,9 @@
+@@ -6030,10 +_,13 @@
+ 		}
+ 
+ 		Main.zenithWorld = everythingWorldGen;
+-		Console.WriteLine("Creating world - Seed: {0} Width: {1}, Height: {2}, Evil: {3}, IsExpert: {4}", seed, Main.maxTilesX, Main.maxTilesY, WorldGenParam_Evil, Main.expertMode);
```
Chicken-Bones, line 417: Yeah, lets change this to logging the game mode properly.

Chicken-Bones (changes requested):

Issue #3138, opened 2022-12-15
1.4.4: Items with `shootsEveryUse = true` shoots every frame when entering world


Cyrillya:
### Version

Self-Compiled (please specify branch)

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Branch: newest `1.4.4`
This is what it does basically
![dotnet_n3MMrLkzC7](https://user-images.githubusercontent.com/35227653/207820683-b0c3d65b-040f-458b-87d6-9f9d3bac3115.gif)
Projectiles is spawned every frame with Terra Blade not swinging. If you use it once it returns to normal.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/10235595/client.log)

### Steps to reproduce

1. Put Terra Blade on the first hotbar
2. Rejoin the world
3. It happens

If you then use the item it returns to normal

### Expected Behavior

No projectiles should be spawned

### Actual Behavior

Projectiles are spawned every frame

### Reproduction frequency

Always

### Additional Information

In method `Player.ItemCheck_OwnerOnlyCode`
```CSharp
bool flag2 = itemAnimation > 0 && ItemTimeIsZero && flag;
if (sItem.shootsEveryUse)
	flag2 = ItemAnimationJustStarted; // This is where flag2 set to true
if (sItem.shoot > 0 && flag2)
	ItemCheck_Shoot(whoAmI, sItem, weaponDamage);
```
Here's `ItemAnimationJustStarted` for ease:
```CSharp
public bool ItemAnimationJustStarted => itemAnimation == itemAnimationMax
```
I found that both `itemAnimation` and `itemAnimationMax` are 0 every time I enter the world. So `ItemAnimationJustStarted` is `true` then projectiles are shoot. Maybe we should set itemAnimationMax to -1 on entry.

PR #3139, opened 2022-12-15
[1.4.4] Infinite shooting fix

Cyrillya:
### What is the bug?
Issue #3138 
### How did you fix the bug?
Simply added `itemAnimationMax = -1` to `Player.Spawn()`
I don't know the exact changes tModLoader makes to item usage. Maybe the contributor(s) of the change will have a better solution

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3509,6 +3509,15 @@
  		return num2;
  	}
  
+@@ -29131,6 +_,8 @@
+ 		_funkytownAchievementCheckCooldown = 100;
+ 		bool flag = false;
+ 		if (context == PlayerSpawnContext.SpawningIntoWorld) {
++			itemAnimationMax = -1; // Added by TML. Issue #3138
++
+ 			if (Main.netMode == 0 && unlockedBiomeTorches) {
+ 				NPC nPC = new NPC();
+ 				nPC.SetDefaults(664);
 @@ -29189,6 +_,10 @@
  			}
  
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Why did you set it to `-1` and not `0`? Why does it not get set correctly when you start using the Terra blade after respawning? Isn't it supposed to be set when the first item animation starts?

Cyrillya:
> Why did you set it to `-1` and not `0`? Why does it not get set correctly when you start using the Terra blade after respawning? Isn't it supposed to be set when the first item animation starts?

Both `itemAnimation` and `itemAnimationMax` are 0 every time I enter the world. So `ItemAnimationJustStarted` is `true` then projectiles are shoot.

**Vanilla** code:
```CSharp
public bool ItemAnimationJustStarted => itemAnimation == itemAnimationMax - 1;
```
After entering the world, both  `itemAnimation` and `itemAnimationMax` are 0. Then `0 == -1` returns `false` so nothing happens. But **tML changed** `itemAnimationMax - 1` to `itemAnimationMax`, which caused the bug.

Respawning **doesn't cause** the bug but **entering world does**.  
~~Maybe I should check `context == PlayerSpawnContext.SpawningIntoWorld` before setting.~~ Already did

Chicken-Bones:
Hi, after looking at the code and sleeping on this. I believe the correct fix is to add an `itemAnimation > 0 && ` check to `ItemAnimationJustStarted`. No other changes necessary.

`itemAnimation == 0` means that no item is currently being used. You can see in the shoot code, `flag2` normally checks `itemAnimation > 0` but `ItemAnimationJustStarted` overwrites it.

PR #3145, opened 2022-12-19
[1.4.4] Add ItemID.Sets.GeodeDrops

direwolf420:
### What is the new feature?
`ItemID.Sets.GeodeDrops` is a `Dictionary<int, (int minStack, int maxStack)>` which allows modders to add or change the drops of a broken Geode.

### Implementation
I simply rerouted the entire `Projectile.DropGeodeGems` into a public and static `Projectile.DropGeodeLoot` method, because otherwise, even with minimal patches it would cause problems when the dictionary has less than 2 entries. Thus I decided to opt with a while loop instead which is much safer. The drop code is copied from the rerouted vanilla code.

In vanilla, Geode drops 2 different items of equal drop chance with a fixed range stack size. Modifying or adding to this behavior requires IL editing as the list and the stack are not exposed via any TML hooks or suitable methods to detour.
The way `ItemID.Sets.GeodeDrops` is implemented lets the modder add their own loot (usually gems) combined with a custom stack size.
What is _not_ possible is changing how many different items drop from the Geode. It's fixed at 2, and will drop less or nothing if mods change the count of the dictionary to less than 2.

* Some way to specify how many _different_ items will drop, not just 2.
* Some way to specify "weight" of each item in the pool (i.e. a mod wants to make Diamond rarer without changing the stack size).
* The exceptions thrown at evaluation of the stack range could be moved to a place where they are checked only once (after loading is finished), but that would make it vulnerable for edits to the dictionary when the game is being played.
* The exceptions could include the existing stack range for verbosity.
* The `(3, 7)` could be a `static readonly` in `Item` so that if vanilla/tml change it in a later update, and mods use it for their own items, they will get the updated stack range.

```cs
ItemID.Sets.GeodeDrops.Remove(ItemID.Sapphire);
ItemID.Sets.GeodeDrops[Type] = (3, 7); //Inside a ModItem class
ItemID.Sets.GeodeDrops[ModContent.ItemType<ExampleGem>()] = (3, 7);
ItemID.Sets.GeodeDrops[ItemID.Sapphire] = (12, 34);
```

### ExampleMod updates
Adding an entire gem for this is out of scope.

### DIFF
--- a/patches/tModLoader/Terraria/Projectile.TML.cs
+++ b/patches/tModLoader/Terraria/Projectile.TML.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Xna.Framework;
 using Terraria.DataStructures;
+using Terraria.ID;
 using Terraria.ModLoader;
 
 namespace Terraria;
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -86,6 +88,37 @@ private static void HandlePlayerStatModifiers(IEntitySource spawnSource, Project
...
+		var dict = ItemID.Sets.GeodeDrops;
+		foreach (var pair in dict) {
+			string exceptionCommon = $"{Lang.GetItemNameValue(pair.Key)} registered in 'ItemID.Sets.{nameof(ItemID.Sets.GeodeDrops)}'";
+			if (pair.Value.minStack < 1)
+				throw new Exception($"{exceptionCommon} must have minStack bigger than 0");
+
+			if (pair.Value.maxStack <= pair.Value.minStack)
+				throw new Exception($"{exceptionCommon} must have maxStack bigger than minStack");
+		}
+		var list = dict.Keys.ToList();
```
Chicken-Bones: Would prefer if this part was done in an `ItemLoader` method, after content is added. Sure people don't open geodes that often, but it doesn't provide a good template to follow. Better to separate initialization, validation, and execution

patches/tModLoader/Terraria/Projectile.TML.cs
```diff
@@ -86,6 +88,37 @@ private static void HandlePlayerStatModifiers(IEntitySource spawnSource, Project
...
+
+		int attempts = 0;
+		while (attempts < 2 && list.Count > 0) {
+			attempts++;
+
+			int item = Main.rand.Next(list);
+			list.Remove(item);
```
Chicken-Bones, line 105: I don't love that you have to make a new list every time to do this, but I don't have a better idea

Chicken-Bones (changes requested):

PR #3147, opened 2022-12-19
1.4.4 Making WorldGen.heartCount public

Ishigh1:
heartCount is a field found in WorldGen that tracks where hearts are placed along with heartPos
Both were made private in 1.4.4 but as heartCount is initialized in the Reset pass that some modders may not want to happen, heartCount needs to be set to 0 by a different pass owned by a mod and making it public avoids the reflection

The fact that heartPos could also be made public would also make sense if a mod wanted to set the positions without changing much of the logic behind their creation


Chicken-Bones:
Hm, sorry. Too specific. Skipping the `Reset` pass seems quite dangerous to me, and `heartPos` exists solely to communicate between `CrimVein` both of which only happen in the `"Corruption"` gen pass, so there's not really an opportunity for a mod to get 'inbetween' them in a normal way.

PR #3151, opened 2022-12-23
[1.4.4] Call `ItemLoader.UseItem` for food quick buff

Cyrillya:
### What is the new feature?
Add (missing?) `ItemLoader.UseItem`  call for food items that were used by quick buff
I was about to do something when the food is used, but I found that `UseItem` is not called when the food is used in quick buff, which makes it more difficult to add quick buff support for the use of food items
Not sure if it is a "bug fix" or a "new feature" lol

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -539,9 +539,12 @@
  		if (CountBuffs() == maxBuffs)
  			return;
  
-@@ -3238,7 +_,7 @@
+@@ -3237,8 +_,10 @@
+ 			if (num == 0)
  				num = 3600;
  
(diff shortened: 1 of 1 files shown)


Cyrillya:
This should be based on `1.4` branch actually, I'll do it if it is really necessary.

Solxanich:
Regarding 1.4 vs 1.4.4 - it is generally recommended to PR to 1.4.4 unless a critical fix is needed to carry through to preview and stable builds 

Is this something that can wait for (like 2 to 3 months for 1.4.4 to move along, hopefully on preview in next few weeks) or do you need this on stable/default?

Cyrillya:
It can wait until 1.4.4, yes

PR #3162, opened 2023-01-02
[1.4.4] Add TileID.Sets.AvoidedByMeteorLanding

direwolf420:
### What is the new feature?
Adds a TileID set which marks a tile to be avoided by meteor generation.

Currently, only modded chests and dungeon tiles are avoided by a meteor. Tiles that either generate on the surface or should never be tampered with their surroundings by a meteor are hardcoded (includes pylons, hat rack/mannequin, lihzard brick etc.).
The spirit of this set is to avoid meteor landing near important structures or the player's bases.

~~This is as straightforward as it gets. Naming could be changed to `PreventsMeteorGeneration` or `AvoidedBy/PreventsMeteorLanding`.~~ Decided on `AvoidedByMeteorLanding`
This is also very different from the existing set called `GetsDestroyedForMeteors`, there is no overlap here.

For modded pylons or important tiles that generate on the surface.
Some mod could likely add Ebonstone into this set which would prevent meteors from landing in corruption chasms.

### ExampleMod updates
ExamplePylon was updated to use this set.

### Porting notes
Add `TileID.Sets.AvoidedByMeteorGeneration[Type] = true;` to your modded pylons.

### DIFF
--- a/ExampleMod/Content/Tiles/ExamplePylonTile.cs
+++ b/ExampleMod/Content/Tiles/ExamplePylonTile.cs
@@ -58,7 +58,7 @@ namespace ExampleMod.Content.Tiles
 
 			TileID.Sets.InteractibleByNPCs[Type] = true;
 			TileID.Sets.PreventsSandfall[Type] = true;
-			TileID.Sets.AvoidedByMeteorGeneration[Type] = true;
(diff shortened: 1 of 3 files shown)


Chicken-Bones:
Approved, change to `Landing` preferred

PR #3165, opened 2023-01-02
[1.4.4] Add new secret seed conditions for ItemDropRules

Setnour6:
_**This description was taken from #2987.**_

Condition descriptions are null but I want them to have values just like in the actual conditions.cs file, where `GetTextValue` gets the entry being in string "Bestiary_ItemDropConditions.Is10thAnniverary" for example. If possible, change the patch file to add the new seeds the way the conditions file has them. Also see #2972 for the issue.
If it's not possible or is too difficult for "Bestiary_ItemDropConditions" to be updated, that is okay, since `IsPreHardmode` class in Conditions.TML.cs doesn't have a condition description at all.

### What is the new feature?
I noticed that there is no drop conditions for any other seed, like the "For The Worthy" seed world and Drunk World.
So why don't we add drop conditions for the other seeds?
See Issue #2972 
See Issue #2972 
Alternatively, you can just use an if statement and check if the player is in a seed world, and then add your drop rule there like you normally would.
See Issue #2972 
```cs
public class ForTheWorthyIsUp : IItemDropRuleCondition, IProvideItemConditionDescription
        {
            public bool CanDrop(DropAttemptInfo info)
            {
                return Main.getGoodWorld;
            }

            public bool CanShowItemDropInUI()
            {
                return Main.getGoodWorld;
            }

            public string GetConditionDescription()
            {
                return Language.GetTextValue("Bestiary_ItemDropConditions.IsForTheWorthy");
            }
        }
```
There is no actual Localization description for GetTextValue on IsForTheWorthy, and I am unable to provide that for tModLoader without issues (if anyone knows how, please help provide it for me).
### Second sample usage for the new feature
The following is for `MyGlobalNPC : GlobalNPC`
```cs
public override void ModifyNPCLoot(NPC npc, NPCLoot npcLoot)
    {
        // First, we need to check the npc.type to see if the code is running for the vanilla NPC we want to change
        if (npc.type is NPCID.Lavabat or NPCID.Hellbat)
    	{
            // This is where we add item drop rules for Lavabat/Hellbat, here are some examples:
            
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.DrunkWorldIsUp(), ItemID.FlowingMagma, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.BeesSeed(), ItemID.CrispyHoneyBlock, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.ForTheWorthyIsUp(), ItemID.LivingFireBlock, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.TenthAnniversaryIsUp(), ItemID.BurningSpirit, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.DontStarveIsUp(), ItemID.SpicyPepper, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.RemixSeed(), ItemID.Flamelash, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.NoTrapsSeed(), ItemID.FlameTrap, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.ZenithSeedIsUp(), ItemID.LakeofFire, 5, 1, 1, 5));
            npcLoot.Add(ItemDropRule.ByCondition(new Conditions.ZenithSeedIsUp(), ItemID.LivingFlameDye, 5, 1, 1, 5));
        }
    }
```
### ExampleMod updates
_N/A_


Setnour6:
I did notice that the condition descriptions being null is okay, since the condition description for `IsPreHardmode` does not actually exist in the code, but still works in general, so I GUESS it's fine?
If anyone can review this PR and see whether or not condition descriptions should really be added (`public string GetConditionDescription() => Language.GetTextValue("Bestiary_ItemDropConditions.IsZenithSeed")` for example), that would be great.

Chicken-Bones:
While I'm generally in favour of condition descriptions everywhere. It looks like it is intended to return `null` if the condition isn't meant to be visible in the bestiary.

A better design would be a separate flag, and requiring all conditions to have a description, so mods could display more info, but this matches vanilla so it is fine at this point.

Setnour6:
> While I'm generally in favour of condition descriptions everywhere. It looks like it is intended to return `null` if the condition isn't meant to be visible in the bestiary.
> 
> A better design would be a separate flag, and requiring all conditions to have a description, so mods could display more info, but this matches vanilla so it is fine at this point.

I'm not sure about "separate flags", so I don't think I can do that. I'm more sure about the condition descriptions, in which I can make the patches necessary, since I managed to do it (but I changed my mind because the other added TML conditions didn't have any actual condition descriptions).

As for the secret seed conditions, I made a modification about it, demonstrating the seed effects for Hellbats and Lava Bats (except that I cannot publish it because I'm not allowed to do so on Beta versions of tModLoader)

Chicken-Bones:
What I meant by the flag thing is that we could add
`bool ShowInUI()` to `IProvideItemConditionDescription`

Setnour6:
> What I meant by the flag thing is that we could add `bool ShowInUI()` to `IProvideItemConditionDescription`

I might be interpreting this very differently, but I think that it would look something like
```cs
public interface IProvideItemConditionDescription
{
    string GetConditionDescription();
    bool ShowInUI();
}
```
You can show or write an example in how you perceive it using what I wrote here, or in an email. I can make a new branch on the forked repo and more secret seed condition changes can be made there

Chicken-Bones:
That is what I meant yes :). I don't think you need to add it. The only benefit is that all conditions would have a description, even if it didn't show in the bestiary UI, it could be shown in an info/analysis mod

Setnour6:
> That is what I meant yes :). I don't think you need to add it. The only benefit is that all conditions would have a description, even if it didn't show in the bestiary UI, it could be shown in an info/analysis mod

I suppose so. I was thinking that `bool ShowInUI();` should instead say `bool ShowDescriptionInUI();` just so it's clearer for more people to understand.

Chicken-Bones:
`ShowDescriptionInBeastiary` even

Issue #1399, opened 2021-03-29
Spreading biome additions


Quethed:
Tried to make a set of corruptable blocks, found 2 important code hardcoded.

Adding this support would allow the author to extend the corruption, crimson, or hallow. This would be great for all those custom biomes in mods. Adding support for vanilla blocks is good too.

WorldGen.Convert(int, int, int, int) should have ModWorld support to add extra entries.
WorldGen.GERunner(int, int, float, float, bool) should also have ModWorld support to add extra entries, likely even the same method.
Any more occurrences would help too.

ModTile could be an alternate, but it could make code unnecessarily lengthy and might not be able to support vanilla blocks. GlobalTile is a another good alternate.

I would use this for my newest in development mod that overhauls the spreading biomes. I own the slime biome mod and it could use the ability to convert into a spreading biome too. Thorium and Calamity have an ocean biome. Tremor when it was in development had an arctic glacier biome. Spirit mod when in development had the reach biome and I believe a spirit biome too. Crystillium had a crystal biome. The list is extensive, my point is that this could be a small change for a huge impact.

Example mod should have conversion variants for ExampleBlock and use around this code:
```cs
// ModTile.SetDefaults()
TileID.Sets.Corrupt[Type] = true;
TileID.Sets.Crimson[Type] = true;
TileID.Sets.Hallow[Type] = true;

// ModWorld
public override void TileCountsAvailable(int[] tileCounts) {
	Main.evilTiles += tileCounts[mod.TileType("PurpleExampleBlock")]; // Corrupt variant
	Main.bloodTiles += tileCounts[mod.TileType("RedExampleBlock")]; // Crimson variant
	Main.holyTiles += tileCounts[mod.TileType("PinkExampleBlock")]; // Hallow variant
}

// The new method idea
// This should handle all vanilla conversion methods and projectiles.
public override void ModifyTileConversionTable(int type, ref int corrupt, ref int crimson, ref int hallow) {
	// Starts with all numbers the type.
	// Ends with the conversion chart nesessary.
	// For multiple pure variants, add multiple entries.
	// Conversion toward pure could default to the first entry.
	if(type == mod.TileType("ExampleBlock")) {
		corrupt = mod.TileType("PurpleExampleBlock");
		crimson = mod.TileType("RedExampleBlock");
		hallow = mod.TileType("PinkExampleBlock");
	}
}

//Same but for walls
[...]

PR #3166, opened 2023-01-02
WIP: Reworked Biome Block Counting for Vanilla Biomes

PaulHause:
### What is the new feature?

This is an attempt to partially address the issue discussed in [Issue#1339](https://github.com/tModLoader/tModLoader/issues/1399).

This is a slight rework of the way Terraria handles counting biome related blocks with concern to vanilla biomes. Instead of using the hardcoded tile numbers, the newly added counting mechanism uses the tile ID sets to allow for modders to add their own tiles to said sets and easily make a **modded** block count as part of a **vanilla** biome.

This small addition allows modders to more easily make their own blocks count as part of vanilla biomes. If a modder wished to make a new stone type of block that counted to the game as being a part of the Hallow biome, for example, it was previously difficult to do so (perhaps even impossible). 

This is where I ask for feedback. This is my first time attempting contribution to a large-scale project such as tModLoader and I would love pointers and feedback from anyone more familiar with the codebase. While very simple in approach, the current way I do this feels dangerous to me, as it essentially loops through every single tile in the game for every single frame of gameplay and it feels this could get bloated with how many tiles are added by a large number of mods. 

I would also appreciate any advice on if my code feels logical in its placement and usage, as well as my adherence to the style guide, as I had a larger patch than actual changes due to trying to keep my code legible to at least myself. There may very well be a better way to do this, but I wanted to at least get the ball rolling as this is a feature of interest to me for my own modding endeavors. (I am on the Discord server if you wish to chat with my directly as well: Username is Golden Guardian). 

The example I used above is good, but one I created for the purposes of testing my changes was to make a block that counted as both Snow and Hallow. This was to test if this implementation did not mess up any of vanilla Terraria's usual biome considerations. The expected result is that a small amount of the block would count as a Hallowed biome, and a much larger amount would count as both Hallowed and Snow. 

This was in fact the result and my own usage of this would be to further expand a mod allowing for Snow to be a tile that the Crimson, Corruption, and Hallow would spread to, not just ice. This could also allow modders to create entirely new structures within biomes constructed entirely of mod tiles without fearing overwriting the biome.


patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -957,5 +957,34 @@ public static class TileLoader
...
 			return GetTile(type)?.UnlockChest(i, j, ref frameXAdjustment, ref dustType, ref manual) ?? false;
 		}
+
+		public static void RecountTiles(SceneMetrics metrics) {
+
+			// reset every tile count
+			metrics.HolyTileCount = metrics.EvilTileCount = metrics.BloodTileCount = metrics.SnowTileCount = metrics.JungleTileCount = metrics.MushroomTileCount = metrics.SandTileCount = 0;
+
+			for (int i = 0; i < TileCount; i++) {
+				if (TileID.Sets.HallowBiome[i] == true)
```
Chicken-Bones: No need for `== true`

patches/tModLoader/Terraria/SceneMetrics.cs.patch
```diff
@@ -177,6 +177,18 @@
  		}
  
  		private void ExportTileCountsToMain() {
+@@ -355,6 +_,11 @@
+ 			PartyMonolithCount = _tileCounts[455];
+ 			GraveyardTileCount = _tileCounts[85];
+ 			GraveyardTileCount -= _tileCounts[27] / 2;
++
++			TileLoader.RecountTiles(this);
```
Chicken-Bones: Better to comment out the above code, if it's going to be ignored anyway

Chicken-Bones: This currently doesn't handle the `Main.remixWorld` condition changes which tiles count as jungle/evil/blood

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -957,5 +957,34 @@ public static class TileLoader
 		public static bool UnlockChest(int i, int j, int type, ref short frameXAdjustment, ref int dustType, ref bool manual) {
 			return GetTile(type)?.UnlockChest(i, j, ref frameXAdjustment, ref dustType, ref manual) ?? false;
 		}
+
+		public static void RecountTiles(SceneMetrics metrics) {
+
+			// reset every tile count
+			metrics.HolyTileCount = metrics.EvilTileCount = metrics.BloodTileCount = metrics.SnowTileCount = metrics.JungleTileCount = metrics.MushroomTileCount = metrics.SandTileCount = 0;
+
+			for (int i = 0; i < TileCount; i++) {
```
Chicken-Bones: Should store the result of `metrics._tileCounts[i]` in a local variable, and `continue` the loop if it's 0.

patches/tModLoader/Terraria/ID/TileID.TML.cs
```diff
@@ -69,6 +69,17 @@ partial class Sets
 			/// Tiles that are interpreted as a wall by nearby walls during framing, causing them to frame as if merging with this adjacent tile. Prevents wall from drawing within bounds for transparant tiles.
 			/// </summary>
 			public static bool[] WallsMergeWith = Factory.CreateBoolSet(Glass);
+
+			/// <summary>
+			/// New created sets to facilitate vanilla biome block counting including modded blocks. To replace the current hardcoded counts in SceneMetrics.cs
+			/// </summary>
+			public static bool[] CorruptBiome = Factory.CreateBoolSet(23, 24, 25, 32, 112, 163, 400, 398);
```
Chicken-Bones: Suggest, using `IntSet` instead, and we could put the 'score' values in here (1 normally, but it would support when a tile counts for multiple, or negative)
Eg `-10` evil for sunflowers

Chicken-Bones (changes requested):
Thankyou for your first contribution, please see my feedback.

patches/tModLoader/Terraria/SceneMetrics.cs.patch
PaulHause: Oh, I was unaware of this condition. I'm not seeing it in the source code I have, is this in a newer version? (I forked from 1.4 default, if that helps anything)

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
PaulHause: For my own curiosity, any reason to store it in a local variable rather than making a direct check? This is what I do in my most recent commit. Is this considered bad practice?

Chicken-Bones: Generally a good idea not to duplicate code (Terraria is a very poor example of this practice). Looking up an array repeatedly costs more than doing it once and saving the result.

patches/tModLoader/Terraria/SceneMetrics.cs.patch
Chicken-Bones: Ah, my apologies, I missed that this isn't targeting 1.4.4.

Development for 1.4 has stopped, so please rebase this onto the 1.4.4 branch.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
Chicken-Bones: Also same goes for below, if you're going to check `TileID.Sets.HallowBiome[i] != 0` you may as well just do the multiplication and add directly.

PaulHause: OH, I understand what you mean now. I somehow entirely overlooked that I use it for every line that it is adding to the counts even though that is the entire point, ahah.

I can fix that right away.

patches/tModLoader/Terraria/SceneMetrics.cs.patch
PaulHause: Understood, I'll do that. Should be a simple enough switch.

PaulHause:
This is going to look far too messy as I fix my own repository to rebase to 1.4.4, going to close this and make a new request when I've rebased correctly.

Chicken-Bones:
Force push is always an option!

PR #3170, opened 2023-01-05
Freeze time fix

Jadams505:
### What is the bug?
Journey mode's freeze time has different behavior in tModLoader and vanilla. More specifically skies and invasions update in vanilla even when `dayRate` is 0 whereas in tModLoader they do not.

This mostly causes visual bugs like ambient objects in the sky staying frozen in place. However two more important differences caused by this bug is that in tModLoader the credits will not roll and invasions won't start when time is frozen.
### How did you fix the bug?
I reverted to the vanilla behavior of always updating at least once in these cases and added `Main.forceStopWorldEvents` as a way for modders to bypass this behavior if they still want to freeze the events

The intended use of this feature is to override `ModSystem.ModifyTimeRate` and set `eventUpdateRate` to 0 and `Main.forceStopWorldEvents` to true in order to actually freeze these events.

The naming could possibly be better. 
It might also be helpful if there was an additional parameter added to `ModSystem.ModifyTimeRate` for this purpose

### DIFF
--- a/patches/tModLoader/Terraria/Graphics/Effects/SkyManager.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Effects/SkyManager.cs.patch
@@ -13,7 +13,7 @@
  
  	public void Reset()
  	{
-@@ -17,15 +_,20 @@
+@@ -17,15 +_,27 @@
  		}
  
  		_activeSkies.Clear();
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
I think, that the tML approach is actually better. Though it could be confusing/annoying to journey mode players.

I dislike the idea of a discontinuity around 0, where an update rate of `0.1` (10x longer days) stalls out invasions and slows credits and skies and everything, but suddenly when you set it to 0, half the things start running again.

I could see a decision to unlink invasions and credits from the time rate adjustment completely.

Jadams505:
I can definitely see how the tML approach is more straightforward and that by allowing time to slow down the vanilla approach makes less sense. 
I just thought that tML always had to follow vanilla and didn't realize this behavior was intentional

Issue #3152, opened 2022-12-23
Allow SetupShop hook to work with AppliesToEntity hook


stormytuna:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently `SetupShop` will not take into account `AppliesToEntity` for the current NPC, so you can have a class like below and it will still add the Terra Blade to every npcs shop

```
    public class SomeGlobalNPC : GlobalNPC {
        public override bool AppliesToEntity(NPC entity, bool lateInstantiation) => entity.type == NPCID.Dryad;

        public override void SetupShop(int type, Chest shop, ref int nextSlot) {
            shop.item[nextSlot].SetDefaults(ItemID.TerraBlade);
            nextSlot++;
        }
    }
```

### What does this proposal attempt to solve or improve?

This would prevent issues where someone adds an item to a shop thinking it's only in that shop as they used `AppliesToEntity`. It would also prevent needing extra logic inside `SetupShop` if `AppliesToEntity` is being used for the rest of the code in that `GlobalNPC`

Chicken-Bones:
Hi. Firstly I'd like to apologise for the confusion, perhaps we could add this to the documentation.

Unfortunately we can't do this in the current design, because `AppliesToEntity` takes an `NPC` parameter, but `SetupShop` doesn't have a full `NPC` instance, only a `int type`. Getting the actual `NPC` instance for all shops would be a significantly difficult change given the Terraria code.

DavidF-Dev:
``NPC instance = Main.npc[Main.LocalPlayer.talkNPC]``
Would this be a possible solution for this issue? Depends if ``SetupShop`` is called in cases other than player interaction. I just helped a user in the Discord who wanted multiple instances of the same NPC type but with different shops; and this worked for them. They could access the specific instance being talked to.

If you're amending the documentation for ``GlobalNPC``, can I also request that you change the method description in ``ModNPC``; so that it's obvious it **isn't** being called on the instance of the npc being talked to.

[...]

PR #3172, opened 2023-01-05
[1.4.4] Updated Shimmered Example NPCs and Added Gore

Rijam:

Updates the shimmered Example Person, Example Traveling Merchant, and adds shimmered Example Bone Merchant. Gore for all three of those NPCs were added. I also added/clarified some comments in Example Person.

Remaining issues:
* **I've created 3 styles for creating the gore (one style for each NPC). Tell me which style is the best and I will update the other two to match.**
* Example Bone Merchant has their shimmer texture while in the Bestiary if they are shimmered.
* Not directly Terraria/tModLoader's fault, but trying to check for `NPC.IsShimmerVariant` in `SetupShop()` will always be false. Vanilla doesn't do anything with this, but I know some modders will want to have special shop items when their Town NPC is shimmered. One way around this is to just check something like `Main.npc[NPC.FindFirstNPC(ModContent.NPCType<ExamplePerson>())].IsShimmeredVariant` in the shop. That's not super elegant, though.

If anyone has opinions regarding naming and/or whether we should put the shimmered textures in a Shimmered folder, let me know. 

I could also add something like
```cs
public override void PostAI() => NPC.color = NPC.IsShimmerVariant ? Main.DiscoColor : default; // Make the color of the NPC rainbow when shimmered.
```
if we wanted to show off some different possibilities.

No, but I am planning on making a Town NPC guide in the future.

<details><summary>Old description</summary><blockquote>

Adds a shimmered form to the Example Person. I also added/clarified some comments in Example Person.

Big remaining issues:
* The head icon doesn't change when being shimmered, unlike vanilla. Setting multiple textures with conditions in `GetHeadTextureIndex()` will result in the head icon to be invisible and an index out of bounds error. Perhaps we need a `AutoloadShimmeredHead` attribute or something?
* The shimmer state doesn't save when exiting and rejoining the world. Perhaps we need to save it in `SaveData()` and `LoadData()` hooks? `NPC.IsShimmeredVariant` is read only I'm pretty sure, so we'd have to do something else.
* Not directly Terraria/tModLoader's fault, but trying to check for `NPC.IsShimmerVariant` in `SetupShop()` will always be false. Vanilla doesn't do anything with this, but I know some modders will want to have special shop items when their Town NPC is shimmered. One way around this is to just check something like `Main.npc[NPC.FindFirstNPC(ModContent.NPCType<ExamplePerson>())].IsShimmeredVariant` in the shop. That's not super elegant, though.

If anyone has opinions regarding naming and/or whether we should put the shimmered textures in a Shimmered folder, let me know. 

I could also add something like
```cs
public override void PostAI() => NPC.color = NPC.IsShimmerVariant ? Main.DiscoColor : default; // Make the color of the NPC rainbow when shimmered.
```
if we wanted to show off some different possibilities.

I also wanted to know if I should add gore to Example Person. 

No, unless we want to make a Town NPC Guide or something like that, which I would be ok with making.

</blockquote></details>


Chicken-Bones:
- Shimmered head icons need investigating.
- Variants not saving is known, and listed in the porting notes as a todo
- `SetupShop` is related to #3152

Rijam:
I added the Example Bone Merchant because the vanilla Skeleton Merchant can be shimmered. However, I noticed the Bestiary was showing the shimmered texture even when the ITownNPCProfile says not to.

I tried to fix the issue of modded Town NPCs not saving their shimmer state with (in `WorldIO.cs`):
```patch
	internal static List<TagCompound> SaveNPCs()
				...
				if (npc.townNPC) {
					tag["displayName"] = npc.GivenName;
					tag["homeless"] = npc.homeless;
					tag["homeTileX"] = npc.homeTileX;
					tag["homeTileY"] = npc.homeTileY;
+					tag["townNpcVariationIndex"] = npc.townNpcVariationIndex;
				}
	...
	internal static void LoadNPCs(IList<TagCompound> list)
				...
				if (npc.townNPC) {
					npc.GivenName = (string)tag["displayName"];
					npc.homeless = tag.GetBool("homeless");
					npc.homeTileX = (int)tag["homeTileX"];
					npc.homeTileY = (int)tag["homeTileY"];
+					npc.townNpcVariationIndex = tag.GetInt("townNpcVariationIndex");
				}
	...
```
and it worked! ... until the world sometimes decided not to load. So, somebody who actually knows what they are doing should probably take a look at it. `NPC.ShimmeredTownNPCs[]` is probably what should be used instead anyway.

Rijam:
I've created 3 styles for creating the gore (one style for each NPC). Tell me which style is the best and I will update the other two to match.

JavidPack:
Thanks a bunch!

PR #3173, opened 2023-01-05
Fix content folder path on Mac with GOG version

ilcheese2:
### What is the bug?
tModLoader fails to find the content folder on Mac if you have the GOG version of the game but works correctly with the Steam version.
### How did you fix the bug?
I changed vanillaContentFolder to the correct one on Mac.
Not that I know of

### DIFF
--- a/patches/tModLoader/Terraria/Main.TML.cs
+++ b/patches/tModLoader/Terraria/Main.TML.cs
@@ -179,10 +179,10 @@ namespace Terraria
 				vanillaContentFolder = Path.Combine(Steam.GetSteamTerrariaInstallDir(), "Content");
 			}
 			else {
-				vanillaContentFolder = "../Terraria/Content"; // Side-by-Side Manual Install
+				vanillaContentFolder = Platform.IsOSX ? "../Terraria/Terraria.app/Contents/Resources/Content" : "../Terraria/Content"; // Side-by-Side Manual Install
 
 				if (!Directory.Exists(vanillaContentFolder)) {
-					vanillaContentFolder = "../Content"; // Nested Manual Install
+					vanillaContentFolder = Platform.IsOSX ? "../Terraria.app/Contents/Resources/Content" : "../Content"; // Nested Manual Install
 				}
 				Logging.tML.Info("Content folder of Terraria GOG Install Location assumed to be: " + Path.GetFullPath(vanillaContentFolder));
 			}


Chicken-Bones:
Probably should add these folders in addition to the current ones. You never know where people end up putting their mac installs.

Solxanich:
Based on Terraria mac file structure - I'm guessing this was never looked at to start with. Our current is windows/Linux based. 

GoG Mac is rare as is, so really good find and happy to see take initiative to fix!

In terms of fix, this is a good initial step as is. 
Directory & path makes sense. I approve of if want to add this and initialize baseline.

However, moving forward, I'm thinking may want to introduce a catch-all fallback to specify where the 'directory' for Terraria is for non Steam users. 
Could use our new file select utility. 

Would be less headache then writing the 6 possibilities I can see for Mac alone, and add a fair bit of resiliency at no real risk.

ilcheese2:
Thanks for replying. I can't seem to find the file select utility you talk about. Could you point me towards where it is?

Solxanich:
Apologies, I should have given some more details. The core utility I was referring to is Native File Dialog

I'm thinking we use Terraria.Utilities.FileBrowser in a similar fashion to how it is used in AWorkshopPublisherInstance.OpenFileDialogueToSelectPreviewImage. 

By reading out the directory selection from the result of that function, we could then use that as the location of the Terraria content folder if we couldn't find ourselves.

I would then suggest saving the location of where the content folder was in to the game folder itself, under LaunchUtils, as "TerrariaFolderPath.txt" containing "ContentFolder={Path}". 
Main reasoning being that at this point the config file has not been loaded from the save directory, so need an alternate save source. Might as well put it in the install folder in case someone wants to configure it upfront. 
With respect to ID'ing content folder, I suspect there might be other paths kicking around where we try to guess for GoG, and might be good to add this fallback to those later if we find a need.

JavidPack:
Looks fine to me. I thought we already accounted for this, but I guess we only accounted for the install location.

Solxanich:
Thanks JavidPack, yes, I was looking at a future item that could be done. 
As stated earlier - this is good as is to solve immediate issue. 
I'll merge right away

Solxanich:
@ilcheese2  - thanks for the contribution!
Please ping me in discord if you want the Collaborators role.

PR #3183, opened 2023-01-09
[1.4.4] `ModPlayer.AddMaterialsForCrafting` Method

Cyrillya:
### What is the new feature?
Allow modders to use their own items for crafting by overriding the method `ModPlayer.AddMaterialsForCrafting(out PlayerLoader.ActionOnUsedForCrafting onUsedForCrafting)`
Modders may want to use their own items for crafting in certain situation
You can craft with items in the chest which you opened in vanilla
For mods they can add a custom void bag, which you can craft with its contents
### ExampleMod updates
You can now craft with items in the chest you stand by (if exists), despite not opening it

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
@@ -27,7 +27,7 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 	public override ModPlayer NewInstance(Player entity)
 	{
 		var inst = base.NewInstance(entity);
-
+		
 		inst.Index = Index;
 
 		return inst;
@@ -41,7 +41,7 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 	protected override void ValidateType()
 	{
 		base.ValidateType();
-
+		
 		LoaderUtils.MustOverrideTogether(this, p => SaveData, p => LoadData);
 		LoaderUtils.MustOverrideTogether(this, p => p.clientClone, p => p.SendClientChanges);
 	}
@@ -1120,9 +1120,9 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 	/// </summary>
 	/// <param name="onUsedForCrafting">The action that gets invoked when the item is consumed</param>
 	/// <returns>An list of the items that may be used as crafting materials. If you want to add nothing, return new List&lt;Item&gt;().</returns>
-	public virtual List<Item> AddMaterialsForCrafting(out PlayerLoader.ActionOnUsedForCrafting onUsedForCrafting)
+	public virtual IEnumerable<Item> AddMaterialsForCrafting(out PlayerLoader.ActionOnUsedForCrafting onUsedForCrafting)
 	{
 		onUsedForCrafting = null;
-		return new List<Item>();
+		return null;
 	}
 }
--- a/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
@@ -1115,12 +1115,19 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 	}
 
 	/// <summary>
+	/// An action to be invoked when an item is partially or fully consumed
+	/// </summary>
+	/// <param name="item">The item that has been consumed. May have been set to air if the item was fully consumed.</param>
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -27,7 +27,7 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 	public override ModPlayer NewInstance(Player entity)
 	{
 		var inst = base.NewInstance(entity);
-		
+
```
Chicken-Bones: Please avoid accidental whitespace changes

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -179,6 +179,8 @@
  	{
 +		//TML: Just in case something throws or breaks and the 'clear after craft' doesn't run.
 +		RecipeLoader.ConsumedItems.Clear();
++
++		Dictionary<List<Item>, PlayerLoader.ActionOnUsedForCrafting> moddedMaterials = PlayerLoader.GetModdedCraftingMaterials(Main.LocalPlayer);
```
Chicken-Bones: No need for local variable here, just retrieve it in the `foreach` below

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
```diff
@@ -1326,4 +1326,24 @@ public static List<Item> GetStartingItems(Player player, IEnumerable<Item> vanil
...
+	/// <summary>
+	/// The action that gets invoked when the item is consumed
+	/// </summary>
+	/// <param name="item">The item that is consumed for crafting</param>
+	/// <param name="index">The index of the item in the List&lt;Item&gt;</param>
+	public delegate void ActionOnUsedForCrafting(Item item, int index);
+	private delegate List<Item> DelegateFindMaterialsFrom(out ActionOnUsedForCrafting onUsedForCrafting);
+	private static HookList HookAddCraftingMaterials = AddHook<DelegateFindMaterialsFrom>(p => p.AddMaterialsForCrafting);
+
+	public static Dictionary<List<Item>, ActionOnUsedForCrafting> GetModdedCraftingMaterials(Player player)
```
Chicken-Bones: This should return `IEnumerable<(IEnuermable<Item> items, ActionOnUsedForCrafting onCraft)>`

You can then use `yield return` inside the `foreach` below and yield a value tuple.

patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
```diff
@@ -1113,4 +1113,16 @@ public virtual IEnumerable<Item> AddStartingItems(bool mediumCoreDeath)
...
 	{
 	}
+
+	/// <summary>
+	/// Called when Recipe.FindRecipes is called or the player is crafting an item
+	/// You can use this method to add items as the materials that may be used for crafting items
+	/// </summary>
+	/// <param name="onUsedForCrafting">The action that gets invoked when the item is consumed</param>
+	/// <returns>An list of the items that may be used as crafting materials. If you want to add nothing, return new List&lt;Item&gt;().</returns>
+	public virtual List<Item> AddMaterialsForCrafting(out PlayerLoader.ActionOnUsedForCrafting onUsedForCrafting)
```
Chicken-Bones: Please change returntype to `IEnumerable<Item>` and default return value to `null`

Chicken-Bones (changes requested):
This is a good addition, well thought-out. Just needs to be a bit more generic (`IEnumerable` instead of `List` and `Dictionary`)

Please try and remove any whitespace changes your auto-formatter may have caused.

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
Cyrillya: Just tried it and got the error: `'foreach' statement cannot operate on enumerators of type 'FilteredArrayEnumerator<ModPlayer>' in 'async' or iterator methods because FilteredArrayEnumerator<ModPlayer>' is a 'ref' struct`. I assume that there's something wrong

Chicken-Bones: Hmm, unfortunate, we'll need to be a bit smarter then. I'll whip something up

Cyrillya:
Done

Chicken-Bones:
Please give my changes a once-over to see if I've missed anything. Please give it another test, and ask any questions

Cyrillya:
It works fine 👍

PR #3184, opened 2023-01-09
Add ability to manually select content folder

ilcheese2:
### What is the new feature?
It allows you to select a content folder if one can't be found.
It provides a fallback for when the folder can't be automatically detected. 
It could replace the GOG portion entirely but that may be inconvenient for players. 

Continues off https://github.com/tModLoader/tModLoader/pull/3173.

### DIFF
--- a/patches/tModLoader/Terraria/Main.TML.cs
+++ b/patches/tModLoader/Terraria/Main.TML.cs
@@ -15,6 +15,7 @@ using Terraria.ModLoader;
 using Terraria.ModLoader.Engine;
 using Terraria.ModLoader.UI;
 using Terraria.Social;
+using Terraria.Utilities.FileBrowser;
 
 namespace Terraria
 {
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/nativefiledialog.cs.patch
```diff
@@ -9,3 +9,12 @@
  		return result;
  	}
  
+@@ -107,7 +_,7 @@
+ 		IntPtr outPath2;
+ 		nfdresult_t result = INTERNAL_NFD_PickFolder(intPtr, out outPath2);
+ 		Marshal.FreeHGlobal((IntPtr)(void*)intPtr);
+-		outPath = UTF8_ToManaged(outPath2, freePtr: true);
++		outPath = UTF8_ToManaged(outPath2, freePtr: false);
```
Chicken-Bones, line 17: What's the reason for this change? Needs a comment at least

patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -170,37 +171,62 @@ public partial class Main
...
+						vanillaContentFolder = Platform.IsOSX ? "../Terraria.app/Contents/Resources/Content" : "../Content"; // Nested Manual Install
+					}
 				}
-				Logging.tML.Info("Content folder of Terraria GOG Install Location assumed to be: " + Path.GetFullPath(vanillaContentFolder));
+
+				if (!IsValidContentFolder(vanillaContentFolder, out _)) {
+					nativefiledialog.NFD_PickFolder(null, out vanillaContentFolder);
+				}
+
+				File.WriteAllText("LaunchUtils/TerrariaFolderPath.txt", vanillaContentFolder);
```
Chicken-Bones, line 208: Recommend moving this down below, as a 'common retry' code. Allows you to simplify some of the conditions here.

Also recommend only writing the `LaunchUtils/TerrariaFolderPath.txt` file if it's user selected.

Also better not to have user configuration items in the `LaunchUtils` folder. Just save this in the root directory and call it `ContentFolderPath.txt`

Chicken-Bones (changes requested):
Overall, I'm not sure this feature is justified. `InstallVerifier` already has pathing checks for even allowing you to launch the game on GoG.

I think instead, we should replace all of the GoG pathing code to require the `Content` folder to be alongside the `vanillaExePath` from `InstallVerifier`

patches/tModLoader/Terraria/nativefiledialog.cs.patch
ilcheese2: It's the same reason the other patch exists. I will add a comment.

Chicken-Bones:
Closing as not worth it, can always be re-opened later

PR #3190, opened 2023-01-12
UIMessageBox Fix

Destructor-Ben:
Fixed an incredibly minor error with the UIMessageBox, where text was not cut off properly while scrolling.

Sorry for this being in 1.4.4 btw, I can redo it if needed.

https://user-images.githubusercontent.com/62684035/211964099-0688a8b0-27ea-496d-af59-59771334a41a.mp4

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/Elements/UIText.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/Elements/UIText.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/GameContent/UI/Elements/UIText.cs
 +++ src/tModLoader/Terraria/GameContent/UI/Elements/UIText.cs
-@@ -149,7 +_,10 @@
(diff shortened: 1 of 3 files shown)


steviegt6:
This has been a small thing that bugged me but I never gave it much thought. Looks good!

It escapes me, but does UIText implicitly support chat tags? And if so, does this mean the issue with long chat tags being cut off is also resolved?

Destructor-Ben:
I believe so

JavidPack:
Just a note, in case it comes up:

Long ago we added the current code to avoid major slowdowns, caused by the text calculation code: https://github.com/tModLoader/tModLoader/commit/3ed0aed2ba9fe64edf6ad339061f5dee886e379d 

I didn't experience the same slowdowns today while testing, I think since then the UIText code has changed, but if it does become an issue again we can go back to the old approach and fix the drawing code to draw if the height of the text would put it in view. The UIElement.OverflowHidden feature would prevent drawing outside bounds already, so it wouldn't be too hard.

Issue #3182, opened 2023-01-08
Wire Reactivity for Mod Music Boxes


TecmagDiams:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

Currently tModLoader has a function to register Music Boxs however it does not include the vanilla behavior that makes them reactive to wiring requiring each mod to implement it's own wire behavior. All major mods I have tested do not appear to individually add this behavior which makes them not act 1:1 like the vanilla music boxes in-game.

Rather than require each modder to create this (although simple) behavior it feels more correct that tModLoader's call to register a music box should automatically create this behavior so that all mod music boxes are consistent with vanilla music boxes.

### What does this proposal attempt to solve or improve?

This proposal attempts to solve inconsistencies between vanilla and modded music box behavior. Music boxes are already a commonly used feature inside mods and mod developers should not need to individually replicate a vanilla behavior on each music box.

### Which (other) solutions should be considered?

The alternative is leaving this behavior up to modders, which creates inconsistencies with vanilla behavior.

Kogsey:
This had already been done and shouldn't be an issue, in tModLoader's Wiring patch it has:

```
case139:
case int _ when TileLoader.IsModMusicBox(tile):
    WorldGen.SwitchMB(i, j);
    break;
```

So the case for normal Music Box (139) falls into the tModLoader and they both follow the same logic, I used copy-pasted example mod code the check this, no manual wiring needed. I would assume from this that any changes in music box behaviour from mods are intentional?

Edit: Thought I should also mention the same function is used to modify the vanilla Music Box player interactions

TecmagDiams:
Thanks for looking into this! Taking what you've said I've done some additional testing to try and limit down what might cause this issue (and possibly make it more complex and lower priority).

I've identified this issue specifically happens in multiplayer. When started singleplayer I am able to flip a switch to toggle music boxes, but loading the same world in "Host & Play" or as a dedicated server results in the mod music boxes not reacting to wires (where vanilla ones still do). I've attached a quick example of the issue below.

(For those unfamiliar who shouldn't need to ask: the music box on the left from Verdant should open/close in it's states, and the Mushroom from Calamity should light up and go back. I had each one in an opposite state for a more comprehensive test)

https://user-images.githubusercontent.com/48166478/211872427-3bc29fd9-5ec4-4d7b-b4ff-bf14f21b5c84.mp4

[...]

PR #3191, opened 2023-01-12
Fix for multiplayer modded music boxes not working with wiring

Kogsey:
### What is the bug?
Due to MusicLoader.AddMusicBox not working correctly for server-side, TileLoader.IsModMusicBox always returns false on server side leading to the server side wiring during multiplayer not triggering the box to open and close.

issue #3182 

### How did you fix the bug?
AddMusicBox returns when Main.audioSystem == null, I do not know if it is intended for it to return server side, I'm under the assumption this is deliberate to avoid unnecessarily loading audio server side. Instead of messing with the working stuff I added a bool[] tileIsModMusicBox to Main with the normal arrangement along side changing IsModMusicBox to || or return true if Main.tileIsModMusicBox[tile.type] returns true, sidestepping the issue.

Edits to MusicLoader or Main.audioSystem will probably be able to fix it, how much change will need to be made for the very minor fix ¯\_(ツ)_/¯

Modified the ExampleMusicBoxTile to use the new method.

unlikely?

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleMusicBoxTile.cs
+++ b/ExampleMod/Content/Tiles/ExampleMusicBoxTile.cs
@@ -2,7 +2,6 @@ using ExampleMod.Content.Items.Placeable;
 using Microsoft.Xna.Framework;
 using Terraria;
 using Terraria.DataStructures;
-using Terraria.ID;
 using Terraria.ModLoader;
 using Terraria.ObjectData;
 
@@ -14,7 +13,6 @@ namespace ExampleMod.Content.Tiles
 		public override void SetStaticDefaults() {
 			Main.tileFrameImportant[Type] = true;
 			Main.tileObsidianKill[Type] = true;
-			TileID.Sets.TileIsModMusicBox[Type] = true;
 			TileObjectData.newTile.CopyFrom(TileObjectData.Style2x2);
 			TileObjectData.newTile.Origin = new Point16(0, 1);
 			TileObjectData.newTile.LavaDeath = false;
--- a/ExampleMod/Content/Tiles/ExampleMusicBoxTile.cs
+++ b/ExampleMod/Content/Tiles/ExampleMusicBoxTile.cs
@@ -2,6 +2,7 @@ using ExampleMod.Content.Items.Placeable;
 using Microsoft.Xna.Framework;
 using Terraria;
 using Terraria.DataStructures;
+using Terraria.ID;
 using Terraria.ModLoader;
 using Terraria.ObjectData;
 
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -272,6 +272,17 @@
...
  	public static bool inventorySortMouseOver;
  	public static float GraveyardVisualIntensity;
+@@ -868,6 +_,10 @@
+ 	public static short[] tileGlowMask = new short[TileID.Count];
+ 	public static bool[] tileContainer = new bool[TileID.Count];
+ 	public static bool[] tileSign = new bool[TileID.Count];
++	/// <summary>
++	/// A check to fix multiplayer support, used in TileLoader.IsModMusicBox
++	/// </summary>
++	public static bool[] tileIsModMusicBox = new bool[TileID.Count];
```
direwolf420: I would suggest moving this to `TileID.TML.cs`, in the `Sets` class. initialize it similar to other default-to-false arrays. Then you also don't need `Array.Resize(ref Main.tileIsModMusicBox, nextTile);`.
And change the doc to something more descriptive for modders, you can turn the current summary into a single comment instead as it's mainly tml-internal stuff

Kogsey: 👍

direwolf420: tileIsModMusicBox -> TileIsModMusicBox

Kogsey: 👍

Chicken-Bones:
Is there any reason not to remove the `Main.audioSystem == null` check in `MusicLoader.AddMusicBox`?

I don't like having two ways to do the same thing, and adding an extra line a modder needs to put in their music box tiles is not as good as just making the current code work.

Kogsey:
I was under the impressiom that all that audio stuff was intentionally loaded client side only to save load time, I'll look into it in a couple hours and come back with snippets

Chicken-Bones:
There's good reason not to load the actual music tracks, but I see no reason for the id mappings not to be present server-side.

Kogsey:
Well the way we do ID mappings now requires MusicLoader.GetMusicSlot() which uses MusicLoader.musicByPath, assigned in MusicLoader.AddMusic(), called from MusicLoader.AutoLoadMusic() but requires access to Mod.RootContentSource. to find the asset location, Mod.RootContentSource's intialiser in TModContentSource contains a statement to return on Main.DedServ.

So loading all works fine when added manually and autoload isn't needed to find the path, this could be fixed by removing autoload I guess? But that's not really a satisfying solution.

The best solution I can think of is passing through the string instead of the music slot and running AddMusic() server side in AddMusicBox

Chicken-Bones:
Looking at it more closely, I think we can make `AddMusicBox` accept `0` for `musicSlot` on the server, and only set `musicToItem` and `itemToMusic` on the client

Kogsey:
Tested your fix, that commit works

Issue #3090, opened 2022-11-10
1.4.4: Generalize corruption/crimson shimmer decrafting conditions and allow custom ones


JavidPack:
Currently the game has is hard-coded to support corruption and crimson specific recipes for decrafting via shimmer. (For example, with 2 recipes made with Rotten Chunk and Vertebra respectively, the game will return the Vertebra version in a Crimson world and the Rotten Chunk version in a Corruption world)

We'll need to provide a suitable API to allow modders to conditionally override the default shimmer recipe that will be returned as items to the player.

It would be good to migrate crimson and corruption conditions to this new system. 

Modders might potentially want to return specific recipes, or maybe even non-recipe sets of items.

To implement this, tmodloader will probably want to maintain a mapping from Item to Recipe[], rather than the current system of using ItemID.Sets.IsCrafted[Crimson/Corruption]. The recipes will then be queried to find the first one that has conditions satisfied and use it.

We will gather feedback for how mods intend to use and customize the shimmer feature so that we can support the use cases in a compatible manner.

If you have any item decrafting specific plans for a mod you are involved in, please comment below. Other features of shimmer will be covered in other Issues.

PR #3193, opened 2023-01-13
 Shimmer Recipe.Condition support + internalisation/removal of vanilla corruption/crimson versions

Kogsey:
### What is the new feature?
- Added support for the recipe condition systems to shimmer decrafting
- Added EnableShimmer for the rare cases it is needed
- Added the ability to set a priority to shimmer crafting

To allow modders to add conditions to shimmer item transforms, like biome, world type, etc..

probably, a complete new custom condition system is an option, but I can't think of anything the condition class would find usefull that wouldn't also be usefull in Shimmer, using the Recipe implementation keeps all the already coded conditions in for the shimmer. Along side this a seperate Condition system could be implemented using Recipe.ICondition anyway making the change rather easy down the line so for now I think it's best to use what's existing.

```
CreateRecipe()
	.AddIngredient<ExampleItem>()
	.AddIngredient(ItemID.Vertebrae)
	.AddTile<Tiles.Furniture.ExampleWorkbench>()
	.AddDecraftCondition(Recipe.Condition.CrimsonWorld)
	.Register();

CreateRecipe()
	.AddIngredient<ExampleItem>()
	.AddIngredient(ItemID.Cactus)
	.AddTile<Tiles.Furniture.ExampleWorkbench>()
	.AddDecraftCondition(Recipe.Condition.InDesert)
	.SetShimmerPriority(-1)
	.Register();
```

In the above example, the desert would take priority over the crimson, decrafting to a cactus in the desert, and a vertibrae in the crimson, right now this priority works through mainly convenience.

### ExampleMod updates
yes I did

This is mainly here for feedback and updates as I work,  #3090

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -22,16 +22,19 @@
  			return -1;
  
  		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
-@@ -22,7 +_,15 @@
+@@ -22,7 +_,18 @@
  		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
  			return ItemID.Sets.IsCraftedCorruption[type];
  
-+		//Deliberately preserving the "last first" order of the previous implementation
++		//Deliberately preserving the "last first" order of the previous implementation, adding manual priority support
 +		int recipe = recipeIndecies[recipeIndecies.Length -1];
 +
 +		foreach (int posRec in recipeIndecies) {
-+			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec]))
-+				recipe = posRec;
++			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec])) {
++				if (Main.recipe[posRec].shimmerPriority >= Main.recipe[recipe].shimmerPriority)
++					recipe = posRec;
++
++			}
 +		}
 +
 -		return num;
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -21,7 +21,7 @@
 +		*/
 +
 +		foreach (int recipeIndex in ItemID.Sets.CraftingRecipeIndices[type]) {
-+			if (RecipeLoader.DecraftAvailable(Main.recipe[recipeIndex]) && recipeIndex >= 0)
++			if (RecipeLoader.DecraftAvailable(Main.recipe[recipeIndex]))
 +				return recipeIndex;
 +		}
 +
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -6,7 +6,7 @@
  
  namespace Terraria.GameContent;
  
-@@ -12,8 +_,15 @@
+@@ -12,15 +_,34 @@
  
  	public static int GetDecraftingRecipeIndex(int type)
  	{
@@ -21,11 +21,14 @@
 +		if (recipeIndecies[0] < 0)
  			return -1;
  
++		/*  automatically replaced with corresponding Decrafting Conditions
  		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
-@@ -22,6 +_,16 @@
+ 			return ItemID.Sets.IsCraftedCrimson[type];
+ 
  		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
  			return ItemID.Sets.IsCraftedCorruption[type];
- 
++		*/
++
 +		//Deliberately preserving the "last first" order of the previous implementation, adding manual priority support
 +		int num = recipeIndecies[recipeIndecies.Length -1];
 +
@@ -35,10 +38,9 @@
 +					num = posRec;
 +			}
 +		}
-+
+ 
  		return num;
  	}
- 
 @@ -35,6 +_,9 @@
  			return true;
  
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -1,0 +1,18 @@
+--- src/TerrariaNetCore/Terraria/GameContent/ShimmerTransforms.cs
++++ src/tModLoader/Terraria/GameContent/ShimmerTransforms.cs
+@@ -1,4 +_,5 @@
+ using Terraria.ID;
++using Terraria.ModLoader;
+ 
+ namespace Terraria.GameContent;
+ 
+@@ -35,6 +_,9 @@
+ 			return true;
+ 
+ 		if (!NPC.downedGolemBoss && RecipeSets.PostGolem[decraftingRecipeIndex])
++			return true;
++
++		if (!RecipeLoader.ShimmerAvailable(Main.recipe[decraftingRecipeIndex]))
+ 			return true;
+ 
+ 		return false;
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -6,41 +6,29 @@
  
  namespace Terraria.GameContent;
  
-@@ -12,15 +_,34 @@
+@@ -12,6 +_,7 @@
  
  	public static int GetDecraftingRecipeIndex(int type)
  	{
 +		/*
-+		vanilla:
  		int num = ItemID.Sets.IsCrafted[type];
  		if (num < 0)
-+		return false;
-+		*/
-+
-+		int[] recipeIndecies = ItemID.Sets.IsCrafted[type];
-+		if (recipeIndecies[0] < 0)
  			return -1;
- 
-+		/*  automatically replaced with corresponding Decrafting Conditions
- 		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
- 			return ItemID.Sets.IsCraftedCrimson[type];
- 
- 		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
+@@ -23,6 +_,14 @@
  			return ItemID.Sets.IsCraftedCorruption[type];
-+		*/
-+
-+		//Deliberately preserving the "last first" order of the previous implementation, adding manual priority support
-+		int num = recipeIndecies[recipeIndecies.Length -1];
-+
-+		foreach (int posRec in recipeIndecies) {
-+			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec])) {
-+				if (Main.recipe[posRec].shimmerPriority >= Main.recipe[num].shimmerPriority)
-+					num = posRec;
-+			}
-+		}
  
  		return num;
++		*/
++
++		foreach (int recipeIndex in ItemID.Sets.CraftingRecipeIndices[type]) {
++			if (RecipeLoader.DecraftAvailable(Main.recipe[recipeIndex]) && recipeIndex >= 0)
++				return recipeIndex;
++		}
++
++		return -1;
  	}
+ 
+ 	public static bool IsItemTransformLocked(int type)
 @@ -35,6 +_,9 @@
  			return true;
  
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -22,7 +22,7 @@
  			return -1;
  
  		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
-@@ -22,7 +_,18 @@
+@@ -22,7 +_,17 @@
  		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
  			return ItemID.Sets.IsCraftedCorruption[type];
  
@@ -33,7 +33,6 @@
 +			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec])) {
 +				if (Main.recipe[posRec].shimmerPriority >= Main.recipe[recipe].shimmerPriority)
 +					recipe = posRec;
-+
 +			}
 +		}
 +
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -30,7 +30,7 @@
 +		int recipe = recipeIndecies[recipeIndecies.Length -1];
 +
 +		foreach (int posRec in recipeIndecies) {
-+			if (!RecipeLoader.DecraftAvailable(Main.recipe[posRec]))
++			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec]))
 +				recipe = posRec;
 +		}
 +
--- a/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
@@ -6,13 +6,46 @@
  
  namespace Terraria.GameContent;
  
+@@ -12,8 +_,15 @@
+ 
+ 	public static int GetDecraftingRecipeIndex(int type)
+ 	{
++		/*
++		vanilla:
+ 		int num = ItemID.Sets.IsCrafted[type];
+ 		if (num < 0)
++		return false;
++		*/
++
++		int[] recipeIndecies = ItemID.Sets.IsCrafted[type];
++		if (recipeIndecies[0] < 0)
+ 			return -1;
+ 
+ 		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
+@@ -22,7 +_,15 @@
+ 		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
+ 			return ItemID.Sets.IsCraftedCorruption[type];
+ 
++		//Deliberately preserving the "last first" order of the previous implementation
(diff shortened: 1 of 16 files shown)


patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
```diff
@@ -0,0 +1,54 @@
...
++			if (RecipeLoader.DecraftAvailable(Main.recipe[posRec])) {
++				if (Main.recipe[posRec].shimmerPriority >= Main.recipe[recipe].shimmerPriority)
++					recipe = posRec;
++
++			}
++		}
++
+-		return num;
++		return recipe;
+ 	}
```
direwolf420: for patch reduction, change `recipe` to `num`, and get rid of the empty line after the last `posRec` reference

patches/tModLoader/Terraria/ModLoader/RecipeLoader.cs
```diff
@@ -159,6 +159,17 @@ public static bool RecipeAvailable(Recipe recipe)
...
+	/// <summary>
+	/// Returns whether or not the conditions are met for this recipe to be transformed/decrafted.
+	/// </summary>
+	/// <param name="recipe">The recipe to check.</param>
+	/// <returns>Whether or not the conditions are met for this recipe.</returns>
+	public static bool DecraftAvailable(Recipe recipe)
+	{
+		return recipe.DecraftConditions.All(c => c.RecipeAvailable(recipe));
+	}
+
```
direwolf420, line 171: Redundant empty line

patches/tModLoader/Terraria/Recipe.Extensions.cs
```diff
@@ -135,6 +137,7 @@ public bool TryGetIngredient(Mod mod, string itemName, out Item ingredient)
 	public bool RemoveRecipeGroup(int groupID) => acceptedGroups.Remove(groupID);
 
 	public bool RemoveCondition(Condition condition) => Conditions.Remove(condition);
+	public bool RemoveShimmerCondition(Condition condition) => DecraftConditions.Remove(condition);
```
direwolf420: Missing empty line here

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -80,6 +84,8 @@ public Condition(NetworkText description, Predicate<Recipe> predicate)
 		public bool RecipeAvailable(Recipe recipe) => Predicate(recipe);
 	}
 
+
+
```
direwolf420: Both empty lines are redundant

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -292,18 +299,74 @@ public Recipe AddCondition(IEnumerable<Condition> conditions)
...
+	/// <summary>
+	/// resets a check that is used during load to prevent shimmer being used for this recipe, not for on the fly use
+	/// </summary>
+	public Recipe EnableShimmer()
+	{
+		notDecraftable = false;
+		return this;
+	}
+
+	[Obsolete("Replaced by the conditions system")]
```
direwolf420: This and the other Obsolete attributes are unnecessary. The current iteration of shimmer code is 1.4-dev only. Simply remove them if they are unused.

direwolf420 (commented):
surface-level review

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Kogsey: 👍 Done the list + some other stuff

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -292,21 +296,63 @@ public Recipe AddCondition(IEnumerable<Condition> conditions)
...
+			}
+		}
+		return this;
+	}
+
+	/// <summary>
+	/// Sets a check that is used during load to prevent shimmer being used for this recipe, not for on the fly use
+	/// </summary>
+	public Recipe DisableShimmer()
+	{
```
Chicken-Bones, line 346: Please add
```cs
		if (!RecipeLoader.setupRecipes)
			throw new RecipeException("A Recipe can only be disabled inside recipe related methods");

File: patches/tModLoader/Terraria/Recipe.TML.cs [hunk 7]
```diff
@@ -292,21 +296,63 @@ public Recipe AddCondition(IEnumerable<Condition> conditions)
...
+		notDecraftable = true;
+		return this;
+	}
+
+	/// <summary>
+	/// resets a check that is used during load to prevent shimmer being used for this recipe, not for on the fly use
+	/// </summary>
+	public Recipe EnableShimmer()
```
Chicken-Bones: Normally, we these kinds of things, we only support disabling. If you want to enable, just add a new recipe with the same thing. Enabling shimmer for a recipe is not likely to be a common use-case.

Please remove for now.

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
```diff
@@ -0,0 +1,53 @@
...
+ 
++		/*  automatically replaced with corresponding Decrafting Conditions
+ 		if (WorldGen.crimson && ItemID.Sets.IsCraftedCrimson[type] >= 0)
+ 			return ItemID.Sets.IsCraftedCrimson[type];
+ 
+ 		if (!WorldGen.crimson && ItemID.Sets.IsCraftedCorruption[type] >= 0)
+ 			return ItemID.Sets.IsCraftedCorruption[type];
++		*/
++
++		//Deliberately preserving the "last first" order of the previous implementation, adding manual priority support
```
Chicken-Bones: Can you give me an example of when this matters in vanilla?

I'm thinking we should just use the first available shimmer applicable recipe.

patches/tModLoader/Terraria/ID/ItemID.cs.patch
```diff
@@ -25,20 +25,25 @@
...
+@@ -303,17 +_,26 @@
  				}
  			}
  		});
 +
 +		/// <summary> Indicates that an item should show the material tooltip. Typically this means that the item is used in at least 1 recipe, but some items such as coins and void bag hide the tooltip for aesthetic reasons. </summary>
  		public static bool[] IsAMaterial = Factory.CreateBoolSet();
-+		/// <summary> Indicates that this item is the result of at least 1 recipe that isn't disabled or set to not decraftable. The value corresponds to the index of the Recipe that results in this item in <see cref="Main.recipe"/> array. If multiple recipes craft the same item, the recipe added last will be returned. </summary>
- 		public static int[] IsCrafted = Factory.CreateIntSet();
+-		public static int[] IsCrafted = Factory.CreateIntSet();
```
Chicken-Bones, line 35: Since we're changing things anyway, lets change this to `CraftingRecipeIndices`. The default should be `Array.Empty<int>()`

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -292,21 +296,63 @@ public Recipe AddCondition(IEnumerable<Condition> conditions)
...
-		crimson = true;
+		ShimmerConditions.AddRange(conditions);
 		return this;
 	}
 
-	public Recipe CorruptionOnly()
+	/// <summary>
+	/// Adds every condition from Recipie.Condtions to Recipie.ShimmerContions, checking for duplicates.
+	/// </summary>
+	public Recipe CopyConditionsToShimmer()
```
Chicken-Bones: Needs a better name. Maybe `ApplyConditionsAsShimmerConditions`? 

Maybe instead there should be a single function call to add a condition for both crafting and decrafting? `AddCraftingAndDecraftingCondition`?

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -292,21 +296,63 @@ public Recipe AddCondition(IEnumerable<Condition> conditions)
...
+	/// </summary>
+	/// <param name="condition">The predicate delegate condition.</param>
+	/// <param name="description">A description of this condition. Use NetworkText.FromKey, or NetworkText.FromLiteral for this.</param>
+	public Recipe AddShimmerCondition(NetworkText description, Predicate<Recipe> condition) => AddShimmerCondition(new Condition(description, condition));
+
+	/// <summary>
+	/// Adds an array of conditions that will determine whether or not the recipe can be decrafted/changed in shimmer. The conditions can be unrelated to items or tiles (for example, biome or time).
+	/// </summary>
+	/// <param name="conditions">An array of conditions.</param>
+	public Recipe AddShimmerCondition(params Condition[] conditions) => AddShimmerCondition((IEnumerable<Condition>)conditions);
```
Chicken-Bones: I think we should replace `Shimmer` with `Decrafting` here... in the spirit of the original intent which is that there may be other things which decide to 'decraft' recipes, and shimmer doesn't only decraft. You already have `DecraftAvailable` above

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -150,14 +156,23 @@
...
  		item.stack = itemStack;
  		customShimmerResults.Add(item);
 -		return item;
 +		return this;
  	}
  
++	/// <summary>
++	/// Sets the priority of this recipe over other recipes for the same item, highest first, defaults to 0
++	/// </summary>
++	public Recipe SetShimmerPriority(int priority)
```
Chicken-Bones: TML does not accept numerical priorities, because we want to avoid priority fighting. I suggest we run this PR without the priority system, and just use recipe ordering to start with.

If specific decrafting priorities are required, they need to be specified in a relational manner with before/after constraints, similar to all other sorting in tML (including recipe sorting)

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -473,8 +488,29 @@
 +			if (Main.recipe[i].Disabled)
 +				continue;
 +
- 			if (!Main.recipe[i].notDecraftable)
- 				ItemID.Sets.IsCrafted[Main.recipe[i].createItem.type] = i;
++			/*Vanilla:
```
Chicken-Bones: No need for `Vanilla:`. Block comments are assumed to be vanilla code, particularly if the corresponding tML replacement code comes directly after.

Please note, when commenting code. We only put the `/*` and `*/` on lines surrounding the code. We don't add `* ` to every line in the body, to reduce patches.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -473,8 +488,29 @@
...
+-			if (!Main.recipe[i].notDecraftable)
++			 * if (!Main.recipe[i].notDecraftable)
+-				ItemID.Sets.IsCrafted[Main.recipe[i].createItem.type] = i;
++			 * ItemID.Sets.IsCrafted[Main.recipe[i].createItem.type] = i;
++			 */
+ 
++			if (!Main.recipe[i].notDecraftable) {
```
Chicken-Bones: Can you use a `List<int>[]` internally here, and then `ToArray` to create `ItemID.Sets.IsCrafted`? Better for performance and readability. Can use `null` lists for items which have no recipe initially. 

`(recipeLists[itemType] ??= new()).Add(i);`

Chicken-Bones (changes requested):
Good work on this PR!

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Kogsey: I have no idea when it might matter, but that's the way vannila + past implementation had it and I figured that it might save some trouble down the line to leave it

patches/tModLoader/Terraria/Recipe.TML.cs
Kogsey: I'll just do both then? Or is that unnecessary bloat since they both essentially do the same thing?

patches/tModLoader/Terraria/Recipe.TML.cs
Kogsey: Damn, I had it right the second time

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Chicken-Bones: Are there _any_ vanilla items with multiple recipes both set to decraftable where the order matters? You should be able to go through the entries in the `IsCrafted` set and see which ones have > 1 entry (and aren't corruption/crimson variants).

If it doesn't matter to vanilla, we're free to make our own decisions here

Kogsey: There are 23 items with multiple recipes, 12 of those are crim/corr stuff already tagged as such, 2 for coins, 5 for torches, corrupt, crimson, hallow, ice, desert, the yoyo bag, and the gentlemans beard.

None of those really matter, the remaining two are the battle potion and the mechanical worm, should I just add in the crimson-corruption functionality to those?

patches/tModLoader/Terraria/Recipe.TML.cs
Kogsey: 👍

patches/tModLoader/Terraria/Recipe.TML.cs
Kogsey: 👍

patches/tModLoader/Terraria/Recipe.cs.patch
Kogsey: 👍

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Chicken-Bones: Yes, add to those. May be oversights in Vanilla

patches/tModLoader/Terraria/Recipe.cs.patch
Kogsey: Aight so I might have what you're asking wrong, but is:

```
		List<int>[] recipeLists = new List<int>[ItemLoader.ItemCount];
		for (int i = 0; i < numRecipes; i++) {

			if (Main.recipe[i].Disabled)
				continue;

			if (!Main.recipe[i].notDecraftable)
				(recipeLists[Main.recipe[i].createItem.type] ??= new()).Add(i);

		}
		
		ItemID.Sets.CraftingRecipeIndeces = recipeLists.Select(l => l.ToArray()).ToArray();
```

What you mean? and if so do I remove the defualt state from the factory? since it should no longer be needed

patches/tModLoader/Terraria/ID/ItemID.cs.patch
Kogsey: 👍

patches/tModLoader/Terraria/Recipe.TML.cs
Kogsey: I've decided to just change the name as they do essentially the same thing except the new one is 4 extra functions to maintain parity with vanilla

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Kogsey: Aight, changed to use the first recipe + fixed the two recipes

patches/tModLoader/Terraria/Recipe.cs.patch
Kogsey: 👍

patches/tModLoader/Terraria/Recipe.cs.patch
Chicken-Bones: Looks good to me! Yes, remove the default from the factory, let it be null. Shouldn't be accessed prior to this calculation.

Also, typo in your snippet above. `Indices` rather than `Indeces`

Chicken-Bones: Also, please do `var r = Main.recipe[i]`. Repeated array access triggers me a little :')

Can combine the conditions into one then and it'll be nice and concise

Kogsey: 👍

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
```diff
@@ -0,0 +1,41 @@
...
+ 		if (num < 0)
+ 			return -1;
+@@ -23,6 +_,14 @@
+ 			return ItemID.Sets.IsCraftedCorruption[type];
+ 
+ 		return num;
++		*/
++
++		foreach (int recipeIndex in ItemID.Sets.CraftingRecipeIndices[type]) {
++			if (RecipeLoader.DecraftAvailable(Main.recipe[recipeIndex]) && recipeIndex >= 0)
```
Chicken-Bones: The `>= 0` check is not useful here, may as well remove it.

patches/tModLoader/Terraria/ID/ItemID.cs.patch
```diff
@@ -25,20 +25,25 @@
...
++		/// <summary> Indicates that this item is the result of at least 1 recipe that isn't disabled or set to not decraftable. The value corresponds to the index of the Recipe that results in this item in <see cref="Main.recipe"/> array. Inner array contains multiple results for multiple recipes in order of creation</summary>
++		public static int[][] CraftingRecipeIndices; // = Factory.CreateCustomSet<int[]>(Array.Empty<int>()); written over during load
```
Chicken-Bones: No need for comment here, the old code is kinda confusing left around.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -68,11 +68,15 @@
...
-+	internal bool notDecraftable;
++
++	// Following fields made internal, automatically replaced with corresponding Decrafting Conditions
 -	public bool crimson;
 +	internal bool crimson;
 -	public bool corruption;
 +	internal bool corruption;
++
++	// Following fields made internal, use DisableDecraft()
++	internal bool notDecraftable;
```
Chicken-Bones, line 75: Field as moved, please keep it in the original spot for the patch.

Chicken-Bones (changes requested):
Almost perfect!

patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Kogsey: 👍

patches/tModLoader/Terraria/ID/ItemID.cs.patch
Kogsey: 👍

Issue #2643, opened 2022-06-28
The chest in the Example Mod cannot be replaced normally


skywhale-zhi:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When I tried to replace the Example Chest with another original Chest, no Example Chest fell

### Steps to reproduce

1.Let's put down an example chest
2.Replace it with a wooden chest

### Expected Behavior

Example chest is replaced by wooden chest, and the replaced chest falls down

### Actual Behavior

The replacement was successful, but no example chest fell

### Reproduction frequency

always

### Additional Information

I don't know whether it was deliberately designed like this or something else. I found this problem when learning ExampleMod. Normal chests can be replaced normally. The chest in ExampleMod will not be returned after being replaced.

direwolf420:
For anyone who implements this fix, looks like `WorldGen.KillTile_GetItemDrops`, the if block with `includeLargeObjectDrops` needs to include modded chests and dressers somehow (and NOT use `GetChestItemDrop/GetDresserItemDrop`, those methods only work for vanilla (and their callsites have a type check to not include modded types))

skywhale-zhi:
Can you be more specific? I can't find what kind of method to use.

direwolf420:
Did you figure out a solution?

JavidPack:
This is still unsolved.

PR #3198, opened 2023-01-18
Chest and dresser swap fix

Jadams505:
### What is the bug?
See #2643 
### How did you fix the bug?
Allowed `TileLoader.Drop` to return true if the tile has a `ChestDrop` or `DresserDrop` so that `KillTile_GetItemDrops` actually gets called for modded chests and dressers. Then set `dropItem` accordingly.
A more general solution would be to have a hook that allows for a multi-tile drop. That can be used specifically for block swap since the only alternative is `ModTile.KillMultiTile` which doesn't really work since block swapping doesn't actually "kill" the tile.


JavidPack:
Thanks for this. We looked into this feature and found a lot more changes that would be beneficial to be made, so we'll be continuing with that approach. Feel free to test it out, I didn't test a dresser yet, so if you have one in your mod you can try that: https://github.com/tModLoader/tModLoader/pull/3199

PR #3207, opened 2023-01-24
1.4.4 localization overhaul file watcher

JavidPack:
Implement auto refresh from changed hjson file.


patches/tModLoader/Terraria/Localization/LanguageManager.tML.cs
```diff
@@ -148,4 +150,65 @@ private string ComputeBoundTextValue(TextBinding binding)
...
+		if (watcherCooldown > 0) {
+			watcherCooldown--;
+			if (watcherCooldown == 0) {
+				string newText = Language.GetTextValue("tModLoader.WatchLocalizationFileMessage", watchedFile);
+
+				if (!Main.dedServ)
+					Main.NewText(newText);
+				else
+					Console.WriteLine(newText);
+				Logging.tML.Info(newText);
```
Chicken-Bones: Consider a helper in `Utils.TML.cs`

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -180,6 +180,23 @@ private static List<(string key, string value)> LoadTranslations(Mod mod, GameCu
...
+						try {
+							translationFileContents = File.ReadAllText(path);
+						}
+						catch (Exception) {
+							string newText = $"Failed to read {path}";
+							if (!Main.dedServ)
+								Main.NewText(newText);
+							else
+								Console.WriteLine(newText);
+							Logging.tML.Info(newText);
```
Chicken-Bones: The exception will trigger the `FirstChanceExceptionHandler` anyway, so there's probably no reason to log here as well.

patches/tModLoader/Terraria/Localization/LanguageManager.tML.cs
```diff
@@ -148,4 +150,65 @@ private string ComputeBoundTextValue(TextBinding binding)
...
+	}
+
+	private void HandleFileChangedOrRenamed(string fileName)
+	{
+		// fileName is relative to watched folder path
+		string modName = fileName.Split(Path.DirectorySeparatorChar).First();
+		if (ModLoader.ModLoader.TryGetMod(modName, out _)) {
+			watchedMod = modName;
+			watcherCooldown = defaultWatcherCooldown;
+			watchedFile = fileName;
```
Chicken-Bones: Probably need some way to support multiple files being saved at the same time. Perhaps a `Set<string>` with a lock?

We could canonicalize the file path here to find out which localization file in a mod we actually need to override.

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -180,6 +180,23 @@ private static List<(string key, string value)> LoadTranslations(Mod mod, GameCu
 
 				string translationFileContents = streamReader.ReadToEnd();
 
+				if (LanguageManager.Instance.watchedMod == mod.Name) {
```
Chicken-Bones: It's not entirely unlikely that someone may edit files from multiple mods and be confused why it's not working. Perhaps a better idea would be to have a `Dictionary<string, string> overrides` which maps 'in mod path' to absolute on disk path, and is populated by the file watcher and cleared on reload?

Chicken-Bones: Ideally we'd use the asset system for translations, and just runtime replace the asset, but I'm not sure that it works on servers yet :| I have work to do there.

patches/tModLoader/Terraria/Localization/LanguageManager.tML.cs
```diff
@@ -148,4 +150,75 @@ private string ComputeBoundTextValue(TextBinding binding)
...
+			fileWatcher.EnableRaisingEvents = false;
+			fileWatcher.Dispose();
+		}
+		localizationFileWatchers.Clear();
+	}
+
+	private void HandleFileChangedOrRenamed(string modName, string fileName)
+	{
+		watcherCooldown = defaultWatcherCooldown;
+		pendingFiles.Add(Path.Combine(modName, fileName));
```
Chicken-Bones: I think you need to `lock` around accesses to this collection. If you breakpoint you'll find that this method and `Update` are called on different threads.

patches/tModLoader/Terraria/Localization/LanguageManager.tML.cs
```diff
@@ -148,4 +150,75 @@ private string ComputeBoundTextValue(TextBinding binding)
...
+	}
+
+	internal void Update()
+	{
+		// Saving a file in some programs trigger the file multiple times. A cooldown allows tmod to wait until file is finished being changed.
+		if (watcherCooldown > 0) {
```
Chicken-Bones: Suggest using early returns for both of the `if` statements here

patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -341,6 +341,7 @@ internal static void Load(CancellationToken token)
 		ContentSamples.RebuildItemCreativeSortingIDsAfterRecipesAreSetUp();
 		ItemSorting.SetupWhiteLists();
 		LocalizationLoader.UpdateLocalizationFiles();
+		LanguageManager.Instance.SetupFileWatchers();
```
Chicken-Bones: Please move all the file watcher code into `LocalizationLoader`, it's a more natural place.

Then make a single `FinishSetup` method which calls `UpdateLocalizationFiles` and `SetupFileWatchers`
and a single `Unload` method which calls `LanguageManager.Instance.UnloadModdedEntries` and `UnloadFileWatchers`

Chicken-Bones (changes requested):

Issue #2842, opened 2022-08-17
Edit field for Signs fills with "Ghost" cursors after the text


DaLynxx:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

If I place a Sign and then click on it to enter text. I get the normal blue window with Edit and Close.
If I click Edit there is a graphical glitch in that a number of "cursor" characters gets rendered. Starting just after the word I write then continuing of to the right.

I can still edit the text and it looks normal once I close the editing and hover above the sign.

I get the same result when disabling all mods and generating a new player in a new world with the difference that the rendering glitch resets itself quite often so there is only about 2 to 5 cursor characters shown.

![image](https://user-images.githubusercontent.com/1823630/185108292-51a49c6c-2b48-43ca-8553-20bc7400b267.png)

I tried vanilla minecraft as well and I don't get this effect there.

### Steps to reproduce

1. Generate a new character
2. Generate a new world
3. Cut trees to get wood
4. Build a workbench
5. Build a sign
6. Place sign
7. Click to edit.
8. Click "Edit"
9. Rendering Glitch shows.

### Expected Behavior

No long string of cursor characters after the word.

### Actual Behavior

A long sting of cursors are rendered to the right of the word.

### Reproduction frequency

100%

DaLynxx:
I **don't** get the effect when editing names of Chests.

direwolf420:
What fps is your game running at, and what refresh rate is your monitor set to?

DaLynxx:
Running in windowed mode. 144Hz (reports it as 143,857) on Monitor. FPS 60.

Same effect (cursor glitch) when running full window.
Tried my second monitor as well. (Running 60Hz). The glitch still there.

Changed to 1024x768 resolution. No change.
Tested to turn Frame skip off. No change.

(Gonna modify the title since it is the cursor, letter I (large i) or 1 nor l (small L) does not look the same.)

Chicken-Bones:
Please attach a `client.log` file (you can find this in your tModLoader steam install directory under tModLoader-Logs)

DaLynxx:
Here you go.

[client.log](https://github.com/tModLoader/tModLoader/files/9387062/client.log)

Chicken-Bones:
Can you confirm this happens with no mods enabled?

direwolf420:
[...]

PR #3208, opened 2023-01-24
Fix sign cursor spam

Jadams505:
### What is the bug?
See #2842 
### How did you fix the bug?
Remove the text snippet with the "|" after it is drawn
No


JavidPack:
Thanks for this.

PR #3210, opened 2023-01-24
GetItemDrops hook and redesigns to support Block Swap feature

JavidPack:
# What is the new feature?
Various features, fixes, and ports all related to how tiles drop items. 

Note: This PR description has been updated with further changes from #3420 because this description serves as the migration guide for 1.4.4 tile changes. The "Files changed" for #3420 should be consulted as well if a modder is looking for the sum of changes for these 2 PRs.

## Tile Item Drop Redesign
- Tiles now automatically drop the item that places it. Tile place style is taken into consideration.
  - After mod loading, all items are checked for `Item.createTile` and `Item.placeStyle`. A mapping, `TileLoader.tileTypeAndTileStyleToItemType`,  is created of `TileType+TileStyle` to `ItemType`.
  - Modders can use `ModTile.RegisterItemDrop` to manually register drops for specific styles, or to register a fallback item.
  - `ModTile.GetItemDrops` can be used to override or augment the automatic drop system, otherwise the drop will be determined by checking `tileTypeAndTileStyleToItemType[(TileType, TileStyle)]` and falling back to `tileTypeAndTileStyleToItemType[(TileType, -1)]` if `TileStyle` not found
- "Multitiles", tiles larger than 1x1, will now use the automatic tile drop system and won't drop X * Y copies of the `Item`. 
- `ModBlockType.ItemDrop` removed. The automatic system and `ModTile.RegisterItemDrop` replace it.
- `ModTile.Drop` removed. `ModTile.CanDrop` and `ModTile.GetItemDrops` replace it. 
- `GlobalTile.Drop` changed return value from bool to void, with `GlobalTile.CanDrop` added to handle the logic regarding if a tile can drop or not.
  -  `GlobalTile.Drop` will now reliably only be called once even for multitiles.
- Ported ExampleMod
  - All unnecessary `Item` spawns in `ModTile.KillMultiTile` removed
  - All unnecessary `Item` spawns in `ModTile.Drop` removed
  - All `ItemDrop` in `ModTile` removed, `RegisterItemDrop` added for special cases
  - `ExampleTrap` ported. Serves as an example of when `ModTile.GetItemDrops` is necessary. Showcases using `AddContent`.
  - `ExampleChest` updated. In conjunction with the `Chest Lock` feature support, also serves as an example of assigning a fallback item with `ModTile.RegisterItemDrop`.
  - `ExampleDoorOpen` serves as an example of a situation where `ModTile.RegisterItemDrop` is needed.

## Wall Item Drop Redesign
- Walls now automatically drop the item that places it.
  - After mod loading, all items are checked for `Item.createWall`. A mapping, `WallLoader.wallTypeToItemType`,  is created of `WallType` to `ItemType`.
  -  Modders can use `ModWall.RegisterItemDrop` to manually register the item drop.
  - When a `Wall` is dropping it's `Item`, the `Item` drop is calculated from the following, which each step overriding the previous:
    - `wallTypeToItemType[WallType]`
    - `ModWall.Drop` method
- Ported ExampleMod
  - All `ItemDrop` in `ModTile` removed, added `ExampleWallUnsafe` as an example of  where `ModWall.RegisterItemDrop` is needed.

## Block Swap
- Fix block swap feature for Torches, Chests, and Dressers. Previously modded tiles of these types would not swap.
- Ported `ExampleDresser` to test block swap and show a `ModifySmartInteractCoords` example
- `TileLoader/GlobalTile/ModTile.CanReplace` hook. Allows modders to dictate if a tile can be swapped dynamically according to conditions. `CanKillTile` methods are insufficient, as `CanKillTile` logic assumes the tile will be missing after.
- Documentation for `TileID.Sets.PreventsTileReplaceIfOnTopOfIt` and others related to tile mining or swapping prevention.

## Miscellaneous Fixes
- Fix divide by zero error for tiles with no anchors (backport)
- Implement many missing `TileLoader.Drop` call sites
- Standardize `TileLoader.Drop` calls. Previously `WorldGen.KillTile` was called before `TileLoader.Drop`, rendering the hook useless since the tile data was already gone.
- Clean up many patches for `TileLoader.Drop` calls via the `using var _disableNewItem = new Item.DisableNewItemMethod(!drop);` hack. Most are now a single patch.
- Remove `ModTile.ChestDrop` and `ModTile.DresserDrop`, the automatic drop system will determine the item drop.

# Why should this be part of tModLoader?
## Tile Item Drop Redesign
Tile drop code is unnecessarily verbose. Multitiles need to use `ModTile.KillMultiTile` for no reason other than `ModBlockType.ItemDrop` was never properly implemented. `ModBlockType.ItemDrop` is also yet another line of code a modder will need to remember to add to each `ModTile`. 

## Wall Item Drop Redesign
`ModBlockType.ItemDrop` is also yet another line of code a modder will need to remember to add to each `ModWall`. 

## Block Swap
Users will appreciate feature parity. `CanReplace` allows modders to prevent users from bypassing `CanKillTile` logic by tile swapping

## Miscellaneous Fixes
Modders will appreciate the hooks reliably doing what they claim to do.

# Are there alternative designs?
- `ModTile.GetItemDrops` could be different. There is a tradeoff between providing more features and the simplicity of the code.
- `TileLoader/GlobalTile/ModTile.CanReplace` hook might benefit from defaulting to calling `CanKillTile`.
- `Drop` might be too ambiguous, we might want to rename some methods.
- `GlobalTile.GetItemDrops` could be possible, but not worth the effort.

# Sample usage for the new feature
Aside from all the removals in ExampleMod, `ExampleHerb` and `ExampleChest` show examples of the new `GetItemDrops` approach. Click on the `Files changed` button to see how various ExampleMod tiles have been changed. (Also see https://github.com/tModLoader/tModLoader/pull/3420/files for additional changes layered on top of those changes.)

# ExampleMod updates
All `ModTile` in ExampleMod have been updated.

# Porting Notes
The **Required** section explains in broad terms changes modders **MUST** make. 
The **Optional** section explains in broad terms changes modders should make. 
The **Guide** section further below shows specific examples of the exact changes modders will need to follow for various common situations. 

Modders should read and understand all sections, run tModPorter, then use the **Guide** section to adjust their code.

If, after following these instructions, a tile does not drop when in a different "state", such as on/off, see the **Other Issues** section.

## Required
Remove all `Item.NewItem` spawns in `ModTile.KillMultitile`. Not doing this will result in duplicate drops:
```diff
-public override void KillMultiTile(int i, int j, int frameX, int frameY) {
-	Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 32, 32, ModContent.ItemType<Items.Consumables.ExampleFishingCrate>());
-}
```

If there was custom item drop logic in `KillMultiTile` that won't automatically be covered by the automatic item drop system, the modder will need to move that logic to `ModTile.GetItemDrops` and remove the `Item.NewItem` calls. `KillMultiTile` is still used for killing chests and tile entities.

Remove all `ModBlockType.ItemDrop` lines. If an item doesn't exists that places the `ModTile` in a specific style, modders need to use `ModTile/ModWall.RegisterItemDrop` to register a specific drop.

Modders should use `ItemID.Sets.DisableAutomaticPlaceableDrop[Type] = true;` in `ModItem.SetStaticDefaults` for any item that should be ignored by the automatic drop system. For example, seeds and tile placement wands are items that should be ignored, since they aren't intended to drop from the tile when killed. 

**tModPorter**
`GlobalTile.Drop` no longer returns `bool`. Use new `GlobalTile.CanDrop` hook or changed `GlobalTile.Drop` depending on logic.
`ModTile.ChestDrop` removed. Replace with `ItemDrop` or just remove completely.
`ModTile.DresserDrop` removed. Replace with `ItemDrop` or just remove completely.
`ModBlockType.ItemDrop` removed. Replace with `RegisterItemDrop` if needed.

## Optional
If an item shouldn't be recovered when placing, use `ItemID.Sets.DisableAutomaticPlaceableDrop[Type] = true;` in `ModItem.SetStaticDefaults` to ignore this item for the automatic tile to item calculations.

### Typical 1x1 Terrain Tile
Remove all `ModBlockType.ItemDrop` lines. If no item places the tile,  `ModTile.RegisterItemDrop` show be added to register the drop.  For 1x1 tiles that have style and are `Main.tileFrameImportant`, see **Typical 1x1 Furniture Tile** below.
```diff
-ItemDrop = ModContent.ItemType<Items.Placeable.ExampleOre>();
```
### Typical 1x1 Furniture Tile
Remove all `Item.NewItem` calls in `Drop`. If the method is empty, you can delete it altogether. If you have tile entity killing code in here, it should have been in `KillTile` in the first place. Note that if the tile isn't placed by an item or if some tile styles aren't placed by items, you'll want to use `ModTile.RegisterItemDrop` to register drops for the missing styles. If the tile is `Main.tileFrameImportant` but doesn't use normal `TileObjectData` approach, see **Special 1x1 Furniture Tile** below.
```diff
-public override bool Drop(int i, int j) {
-	Tile t = Main.tile[i, j];
-	int style = t.TileFrameX / 18;
-	if (style == 0) {
-		Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 16, 16, ModContent.ItemType<Items.Placeable.ExampleBar>());
-	}
-	return base.Drop(i, j);
-}
```
### Special 1x1 Furniture Tile
Since this tile doesn't use `TileObjectData`, the actual style can't be determined and consequently the correct drop can't be determined automatically. If the tile isn't placed by an item or if some tile styles aren't placed by items, you'll want to use `ModTile.RegisterItemDrop` to register drops for the missing styles. Replace `Drop` method with `GetItemDrops` method and adjust the code to lookup the registered item drop:
```diff
-public override bool Drop(int i, int j) {
+public override IEnumerable<Item> GetItemDrops(int i, int j) {
	Tile t = Main.tile[i, j];
	int style = t.frameY / 18;
+	int dropItem = TileLoader.GetItemDropFromTypeAndStyle(Type, style);
+	yield return new Item(dropItem);
-	if (style == 0)
-		Item.NewItem(i * 16, j * 16, 16, 16, mod.ItemType(Items.Placeable.ExampleTrap.ExampleTrapA));
-	if (style == 1)
-		Item.NewItem(i * 16, j * 16, 16, 16, mod.ItemType(Items.Placeable.ExampleTrap.ExampleTrapB));
-	return base.Drop(i, j);
}
```
### Typical Furniture Tile
Remove all `Item.NewItem` calls in `KillMultiTile`. If the method is empty, you can delete it altogether.  Use `ModTile.RegisterItemDrop` to register drops for any missing styles, or to register a fallback item. Open door tiles in particular will need to use `ModTile.RegisterItemDrop` to drop the item.
```diff
-public override void KillMultiTile(int i, int j, int frameX, int frameY) {
-	Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 48, 32, ModContent.ItemType<Items.Placeable.Furniture.ExampleClock>());
-}
```
### Typical Chest or TileEntity Tile
Same as above, but killing chests and tile entities still need to happen, so leave `KillMultiTile` method and non-`NewItem` code within.
```diff
public override void KillMultiTile(int i, int j, int frameX, int frameY) {
-	Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 32, 32, ChestDrop);
	Chest.DestroyChest(i, j);
}
```
### Typical Wall Item
Remove all `ModBlockType.ItemDrop` lines. If no item places the wall, such as with unsafe walls, use `ModWall.RegisterItemDrop` to register the item drop.
```diff
-ItemDrop = ModContent.ItemType<Items.Placeable.ExampleWallAdvanced>();
```

### Other issues
If you notice after porting that a tile isn't dropping in some situations, such as when placed facing right instead of left, the `TileObjectData` might not have the correct values. If you copied `MinionBossRelic`, for example, you'll need to compare your code to the fixed code.

If you copied `ExampleDoorClosed`, your door might not drop after being opened and closed. You'll need to add the `StyleLineSkip = 3` line, or you can use the `CopyFrom` approach instead. See the [fixed version](https://github.com/tModLoader/tModLoader/blob/1.4.4/ExampleMod/Content/Tiles/Furniture/ExampleDoorClosed.cs).

If using `RandomStyleRange`, make sure to use `RegisterItemDrop` to register a fallback to recover the other styles correctly.

If the tile has an "on" and "off" state, or something similar, it too will need to assign a value to `StyleLineSkip` to properly account for the extra lines of animation and states.  Set `StyleLineSkip` to the number of animation or state sprites the tile has. It only needs to be set if that number is greater than 1.

# Tile Drop Summary
`ModTile.CanDrop` and `GlobalTile.CanDrop` decide if an item drops.
`GlobalTile.Drop` manually drop additional items
`ModTile.GetItemDrops` decides the default item, manually drop additional items


ExampleMod/Content/Items/Placeable/ExampleTrap.cs
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			public void Unload() {
+			}
+		}
+
+		// Here I define some strings that will be used as the ModItem.Name, the internal name of the ModItem. 
+		// We use these in the ExampleMod.Content.Tiles.ExampleTrap.GetItemDrops rather than ModContent.ItemType<Items.Placeable.ExampleTrap>() to retrieve the correct ItemID.
+		public const string ExampleTrapA = "ExampleTrapA";
+		public const string ExampleTrapB = "ExampleTrapB";
```
Chicken-Bones: Probably better to just use `ExampleTrap0` and `ExampleTrap1`.

ExampleMod/Content/Tiles/ExampleTrap.cs
```diff
@@ -24,51 +26,54 @@ public class ExampleTrap : ModTile
...
-				Item.NewItem(i * 16, j * 16, 16, 16, mod.ItemType(Items.Placeable.ExampleTrap.ExampleTrapA));
-			if (style == 1)
-				Item.NewItem(i * 16, j * 16, 16, 16, mod.ItemType(Items.Placeable.ExampleTrap.ExampleTrapB));
-			return base.Drop(i, j);
+			if (style == 0) {
+				dropItem = Mod.Find<ModItem>(Items.Placeable.ExampleTrap.ExampleTrapA).Type;
+			}
+			if (style == 1) {
+				dropItem = Mod.Find<ModItem>(Items.Placeable.ExampleTrap.ExampleTrapB).Type;
+			}
```
Chicken-Bones: I would normally recommend one of:
- `ContentInstance<Items.Placeable.ExampleTrap>.Instances[style].Type`
- A `static ExampleTrap[]` in `ExampleTrap` containing the instances
- A `static GetItem(int style)` function in `ExampleTrap`

ExampleMod/Content/Items/Placeable/ExampleTrap.cs
```diff
@@ -0,0 +1,71 @@
...
+		// By default, the game creates new instances rather than clone. By forcing Clone, we can preserve fields per Item added by the mod while sharing the same class.
+		protected override bool CloneNewInstances => true;
+		private readonly int placeStyle;
+
+		// The internal name of each piece of content must be unique. This code ensures that each of the 2 ExampleTrap instances added have a unique name.
+		public override string Name => placeStyle switch {
+			0 => ExampleTrapA,
+			1 => ExampleTrapB,
+			_ => throw new System.NotImplementedException(),
+		};
```
Chicken-Bones: I guess it's okay to use a `switch` expression here, though we generally avoid them in Example Mod. Not relevant if the above suggestion is taken.

ExampleMod/Content/Items/Placeable/Furniture/ExampleDresser.cs
```diff
@@ -0,0 +1,31 @@
...
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Items.Placeable.Furniture
+{
+	public class ExampleDresser : ModItem
+	{
+		public override void SetStaticDefaults() {
+			Tooltip.SetDefault("This is a modded dresser.");
+
+			Item.ResearchUnlockCount = 1;
```
Chicken-Bones, line 10: Would be nice to have this be the default in a separate PR as you mentioned before. Then a lot of content won't even need a `SetStaticDefaults`

ExampleMod/Content/Tiles/Furniture/ExampleChest.cs
```diff
@@ -49,13 +53,26 @@ public class ExampleChest : ModTile
...
 			TileObjectData.addTile(Type);
 		}
 
+		public override void GetItemDrops(int i, int j, ref int dropItem, ref int dropItemStack) {
+			Tile tile = Main.tile[i, j];
+			int style = TileObjectData.GetTileStyle(tile);
+			if (style == 0)
+				dropItem = ModContent.ItemType<Items.Placeable.Furniture.ExampleChest>();
+		}
+
```
Chicken-Bones: Do we need this, given `ItemDrop` above? If so, comment why?

ExampleMod/Content/Tiles/Furniture/ExampleChest.cs
```diff
@@ -49,13 +53,26 @@ public class ExampleChest : ModTile
...
 			TileObjectData.newTile.HookCheckIfCanPlace = new PlacementHook(Chest.FindEmptyChest, -1, 0, true);
 			TileObjectData.newTile.HookPostPlaceMyPlayer = new PlacementHook(Chest.AfterPlacement_Hook, -1, 0, false);
-			TileObjectData.newTile.AnchorInvalidTiles = new int[] { TileID.MagicalIceBlock };
+			TileObjectData.newTile.AnchorInvalidTiles = new int[] {
+				TileID.MagicalIceBlock,
+				TileID.Boulder,
+				TileID.BouncyBoulder,
+				TileID.LifeCrystalBoulder,
+				TileID.RollingCactus
+			};
```
Chicken-Bones, line 56: The quantity of code here that is common to _all_ modded chests is a pretty strong motivator for `ModChest : ModTile` (worth doing as a followup PR)

ExampleMod/Content/Tiles/Furniture/MinionBossRelic.cs
```diff
@@ -73,22 +74,13 @@ public class MinionBossRelic : ModTile
 			AddMapEntry(new Color(233, 207, 94), Language.GetText("MapObject.Relic"));
 		}
 
-		public override void KillMultiTile(int i, int j, int frameX, int frameY) {
-			// This code here infers the placeStyle the tile was placed with. Only required if you go the Item.placeStyle approach. You just need Item.NewItem otherwise
+		public override void GetItemDrops(int i, int j, ref int dropItem, ref int dropItemStack) {
+			// This code here infers the placeStyle the tile was placed with. Only required if you go the Item.placeStyle approach. You just need ItemDrop otherwise
```
Chicken-Bones: Not particularly clear why `GetItemDrops` would be preferred here

patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -221,15 +215,31 @@ public virtual void DropCritterChance(int i, int j, ref int wormChance, ref int
...
+	/// <br/> This existing logic should cover 99% of use cases, meaning that overriding this method should only be necessary in extremely unique tiles.
+	/// <br/> If this tile drops additional items, those can be spawned here. If the main item drop needs to be customized, spawn the item and set <paramref name="dropItem"/> to 0 to prevent the drop.
+	/// <br/> Use <see cref="CanDrop"/> to prevent any item drops. Use <see cref="KillMultiTile(int, int, int, int)"/> or <see cref="KillTile(int, int, ref bool, ref bool, ref bool)"/> for other logic such as cleaning up TileEntities or killing chests or signs.
+	/// </summary>
+	/// <param name="i">The x position in tile coordinates.</param>
+	/// <param name="j">The y position in tile coordinates.</param>
+	/// <param name="dropItem">The Item type of the drop item.</param>
+	/// <param name="dropItemStack">The stack of the drop item. Defaults to 1.</param>
+	public virtual void GetItemDrops(int i, int j, ref int dropItem, ref int dropItemStack)
```
Chicken-Bones: I still don't love the `ref` approach here. It's completely incompatible with items that have custom data too. I'd prefer a solution with a clear approach to supporting multiple drops. Perhaps `Action<Item> dropFunction`?

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +495,74 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+		ModTile modTile = GetTile(type);
+		foreach (var hook in HookCanDrop) {
 			if (!hook(i, j, type)) {
-				return false;
+				dropItem = false;
 			}
 		}
+		if (modTile != null) {
+			dropItem &= modTile.CanDrop(i, j);
+		}
```
Chicken-Bones: Suggest
`bool dropItem = modTile?.CanDrop(i, j) ?? true;`
and
`dropItem &= hook(i, j, type)`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +495,74 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+		if (tileData == null) {
+			// Terrain tile
+			needDrops = true;
+		}
+		else if(tileData.Width == 1 && tileData.Height == 1) {
+			// 1x1 tile, includeAllModdedLargeObjectDrops prevents double spawns from framing code calling CheckModTile, which calls KillTile_DropItems and KillTile. (Bars)
+			needDrops = !includeAllModdedLargeObjectDrops;
+		}
+		else {
+			if (includeAllModdedLargeObjectDrops)
```
Chicken-Bones: `else if` can be condensed (no need for `{`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +495,74 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+				needDrops = true;
+			else if (includeLargeObjectDrops) {
+				if (TileID.Sets.BasicChest[tileCache.type] || TileID.Sets.BasicDresser[tileCache.type]) {
+					needDrops = true;
+				}
+			}
+		}
+		if (needDrops) {
```
Chicken-Bones: Suggest guard statement, `if (!needDrops) return;`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +495,74 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
 			if (modTile.ItemDrop > 0) {
-				Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 16, 16, modTile.ItemDrop, 1, false, -1);
+				dropItem = modTile.ItemDrop;
+			}
+			if (tileData != null) {
+				int style = TileObjectData.GetTileStyle(tileCache);
+				(int type, int style) key = (tileCache.type, style);
+				if (tileTypeAndTileStyleToItemType.ContainsKey(key)) {
```
Chicken-Bones: Please use `TryGetValue`

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -963,7 +963,7 @@
  				KillTile(num + 1, num6 + m);
  		}
  
-+		if (!TileLoader.Drop(num, j, type))
++		if (!TileLoader.Drop(i, j, type))
```
Chicken-Bones: Could we consider a `using var` hack which disables all drops until the end of the current scope? See usages of `Item.DisableCloneMethod` and `Main.CurrentPlayerOverride` for examples. Might reduce patch size and improve maintainability dramatically.

Chicken-Bones (changes requested):
Great work and comments.

ExampleMod/Content/Items/Placeable/Furniture/ExampleDresser.cs
JavidPack: Made a proposal.

patches/tModLoader/Terraria/ID/ContentSamples.cs.patch
```diff
@@ -70,6 +70,18 @@
...
  			ItemsByType[k] = item;
+ 			string name2 = ItemID.Search.GetName(item.netID);
+ 			ItemPersistentIdsByNetIds[k] = name2;
+ 			ItemNetIdsByPersistentIds[name2] = k;
++			if (!ItemID.Sets.DisableAutomaticPlaceableDrop[k]) {
++				if (item.createTile > -1)
++					TileLoader.tileTypeAndTileStyleToItemType[(item.createTile, item.placeStyle)] = item.type;
++				if (item.createWall > -1)
++					WallLoader.wallTypeToItemType[item.createWall] = item.type;
++			}
```
Chicken-Bones: I think this should go in a `FinishSetup` function in `TileLoader/WallLoader`. Is there any reason it needs to run here?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -929,68 +929,58 @@
...
  			return;
  
  		destroyObject = true;
-+		// Extra context.
++		bool drop = TileLoader.Drop(x, j, type);
  		for (int k = 0; k < num; k++) {
  			if (Main.tile[x, num5 + k].type == type)
  				KillTile(x, num5 + k);
  		}
++		using var _disableNewItem = new Item.DisableNewItemMethod(!drop);
```
Chicken-Bones, line 959: Is there a specific reason these should be separate? It'd be nice to have single line patches. 

I'm thinking of the use-case where people want to replace the vanilla drops for a tile with something else. Is this common? 

I suppose the right way to do that would be to run their own drop code in `CanDrop`? 
But `CanDrop` is meant to be 'pure'. If modders actually did drop items there, they wouldn't be compatible with another mod attempting to block drops. 
We often make `Can` hook short-circuit with this in mind.

If overriding vanilla drops is meant to be supported, we need a little thought (and maybe an example).

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -929,68 +929,58 @@
...
-+
  		if (type == 92)
  			Item.NewItem(GetItemSource_FromTileBreak(x, j), x * 16, j * 16, 32, 32, 341);
- 
-@@ -27126,11 +_,13 @@
+@@ -27126,6 +_,7 @@
  							break;
  					}
  					break;
 +					// Extra patch context.
```
Chicken-Bones: No longer needed

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2076,16 +2096,34 @@
  		for (int k = 0; k < num16; k++) {
  			KillTile_MakeTileDust(i, j, tile);
  		}
-@@ -45758,6 +_,9 @@
+@@ -45756,9 +_,16 @@
  
- 	private static void KillTile_DropItems(int x, int y, Tile tileCache, bool includeLargeObjectDrops = false)
+ 	private static Player GetPlayerForTile(int x, int y) => Main.player[Player.FindClosest(new Vector2(x, y) * 16f, 16, 16)];
+ 
+-	private static void KillTile_DropItems(int x, int y, Tile tileCache, bool includeLargeObjectDrops = false)
```
Chicken-Bones, line 2175: Please add a comment above the method `// made internal, added includeAllModdedLargeObjectDrops param`

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -1813,7 +1814,7 @@
  		if (Main.rand == null)
  			Main.rand = new UnifiedRandom();
  
-+		if (NPCLoader.blockLoot.Contains(Type))
++		if (NPCLoader.blockLoot.Contains(Type) || newItemDisabled)
```
Chicken-Bones, line 2032: Should we consider generalizing the `NPCLoader.blockLoot` hack in a followup PR? It could probably also benefit from a `using var` thing at least. We might be able to combine the two features, so that modders can also block vanilla tile drops individually in `GlobalItem.Drop` with something like `Item.BlockVanillaDropsOf(type)` or something

patches/tModLoader/Terraria/ModLoader/ModPylon.cs
```diff
@@ -254,6 +254,7 @@ public override bool TileFrame(int i, int j, ref bool resetFrame, ref bool noBre
 			return true;
 		}
 
+		WorldGen.KillTile_DropItems(i, j, tileSafely, includeLargeObjectDrops: true, includeAllModdedLargeObjectDrops: true); // include all drops.
```
Chicken-Bones, line 257: Do we need to recommend people do similar if they had tiles with a similar implementation?

Chicken-Bones (changes requested):
Great job on this, it's shaping up really well.

patches/tModLoader/Terraria/ID/ContentSamples.cs.patch
JavidPack: No need, just using an existing loop, moved.

patches/tModLoader/Terraria/WorldGen.cs.patch
JavidPack: Yes, KillTile might cause other tiles to drop. For example, mining a stack of Bars with a Table placed on top. NewItem needs to be disabled after KillTile to not mistakenly disable drops from other tiles that might be dropping from the KillTile->FrameTile recursion. 

We do not currently support replacing drops. To do that, I'd think the existing drop would need to be collected, but we aren't doing that currently.

patches/tModLoader/Terraria/ModLoader/ModPylon.cs
JavidPack: I don't think any modders will be using code like this. ModPylon is strange in how it does things.

patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -192,19 +192,28 @@ public virtual bool CanDrop(int i, int j)
...
 	/// <param name="j">The y position in tile coordinates.</param>
-	/// <param name="dropItem">The Item type of the drop item.</param>
-	/// <param name="dropItemStack">The stack of the drop item. Defaults to 1.</param>
-	public virtual void GetItemDrops(int i, int j, ref int dropItem, ref int dropItemStack)
-	{
+	public virtual IEnumerable<Item> GetItemDrops(int i, int j)
+	{
+		// Automatic item derived from item.createTile and item.placeStyle
+		Tile tile = Main.tile[i, j];
+		TileObjectData tileData = TileObjectData.GetTileData(tile.type, 0, 0);
```
Chicken-Bones: Do you need this, or can you just do `TileObjectData.GetTileStyle` and check for `-1`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +499,97 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+		var itemDrops = modTile.GetItemDrops(x, y);
+		foreach (var item in itemDrops) {
+			item.Prefix(-1); // Assign a random prefix, as expected
+			int num = Item.NewItem(WorldGen.GetItemSource_FromTileBreak(x, y), x * 16, y * 16, 16, 16, item, noBroadcast: false);
+			Main.item[num].TryCombiningIntoNearbyItems(num);
 		}
+	}
 
-		return true;
+	public static int GetItemDropFromTypeAndStyle(int type, int style = 0, bool allowFallbackToStyleZero = true)
```
Chicken-Bones: Is this ever called with `allowFallbackToStyleZero: false`? Is this just intended for modders?

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +499,97 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+			dropItem = value;
+		}
+		else if (allowFallbackToStyleZero && tileTypeAndTileStyleToItemType.TryGetValue((type, 0), out int fallbackValue)) {
+			dropItem = fallbackValue;
+		}
+		// Override
+		ModTile modTile = GetTile(type);
+		if (modTile?.ItemDrop > 0) {
+			dropItem = modTile.ItemDrop;
+		}
```
Chicken-Bones: Since this is a unilateral override, shouldn't it be checked first?

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -485,29 +499,97 @@ public static void DropCritterChance(int i, int j, int type, ref int wormChance,
...
+		foreach (var hook in HookDrop) {
+			hook(i, j, type);
+		}
+
+		return true;
+	}
 
-			if (modTile.ItemDrop > 0) {
-				Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 16, 16, modTile.ItemDrop, 1, false, -1);
+	public static void GetItemDrops(int x, int y, Tile tileCache, ref int dropItem, ref int dropItemStack, bool includeLargeObjectDrops = false, bool includeAllModdedLargeObjectDrops = false)
```
Chicken-Bones: Are the `ref` params here now useless?

patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -188,15 +182,40 @@ public virtual void DropCritterChance(int i, int j, ref int wormChance, ref int
...
-	public virtual bool Drop(int i, int j)
+	public virtual bool CanDrop(int i, int j)
 	{
 		return true;
 	}
 
+	/// <summary>
+	/// Allows you to customize the items the tile at the given coordinates drops.
+	/// <br/> By default, this method will intelligently decide on a single item drop based on <see cref="ModBlockType.ItemDrop"/> and associated <see cref="TileObjectData"/>.
+	/// <br/> For tiles with a <see cref="TileObjectData"/>, the dropped item will be item type of the loaded item with <see cref="Item.createTile"/> set to this Tile and <see cref="Item.placeStyle"/> matching the style of the Tile. If the specific <see cref="Item.placeStyle"/> is not found, the decision will fall back to a <see cref="Item.placeStyle"/> of 0.
```
Chicken-Bones: Doesn't mention `ItemDrop` as part of the default behavior.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
```diff
@@ -22,7 +22,12 @@ public abstract class ModBlockType : ModTexturedType, ILocalizedModType
 	/// <summary> The default type of dust made when this tile/wall is hit. Defaults to 0. </summary>
 	public int DustType { get; set; }
 
-	/// <summary> The default type of item dropped when this tile/wall is killed. Defaults to 0, which means no item. </summary>
+	/// <summary>
+	/// The default type of item dropped when this tile/wall is killed. Defaults to 0, which means no item.
+	/// <br/> Only necessary if there is no item which places this tile (or one of its styles). Otherwise, a matching placeable item will be dropped automatically
+	/// <br/> If this tile/wall drops multiple items, set this to the default item.
+	/// <br/> For ModTile, use <see cref="ModTile.GetItemDrops(int, int, ref int, ref int)"/> to customize which item drops depending on tile style or other conditions.
```
Chicken-Bones: signature in documentation ref hasn't been update

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
JavidPack: Yeah, just there in case it comes up.

Chicken-Bones: I can't really think of a use, especially because you still allow `ItemDrop` override. Prefer to add on request.

JavidPack:
TODO: https://github.com/SamsonAllen13/ClickerClass/blob/1.4.4/Tiles/Mice/TrappedMiceChest.cs doesn't call GetItemDrops, debug it.

PR #3219, opened 2023-01-27
Declarative Shops

BasicallyIAmFox:
### What is the new feature?
Modernizes NPC shops. #3067.
Replaces `ModNPC.SetupShop` hook with `ModNPC.AddShops` (see Porting Notes)
`ModNPC.ModifyActiveShop` allows manual shop creation in old style.

It could allow checking what items NPC can sell or not, better mod compatibility, and also allows easily insert item before/after X item, as well as adding some conditions to already existing items. Also avoids possible IndexOutOfRangeException, if too many items were added in single shop. Info mods can show how to obtain X item because conditions for them are localized.

This PR offers a good compromise between flexibility (via `AbstractNPCShop`) and standardisation. Some aspects are not ideal like the implementation of the 'fixed layout' bartender, but work pretty well overall.

```
public override void AddShops()
{
    var shop = new NPCShop(Type, "Shop"); // You can have different internal names for your shop
    shop.Add<ExampleItem>(); // Adds Example Item in the shop.

    shop.InsertBefore(ModContent.ItemType<ExampleItem>(), ModContent.ItemType<ExampleLifeFruit>()); // Puts Example Life Fruit right before Example Item.

    shop.GetEntry(ModContent.ItemType<ExampleBuffPotion>()).Disable(); // Hides Example Buff Potion from the shop. Will throw an exception if item not found.
    // Safely hides Example Buff Potion from the shop.
    if (shop.TryGetEntry(ModContent.ItemType<ExampleBuffPotion>(), out var entry)) {
        entry.Disable();
    }

    shop.Add(ModContent.ItemType<ExampleSoul>(), NPCShop.Condition.BloodMoon); // Puts an item if its Blood Moon.

    shop.Register();
}
```

### ExampleMod updates
`ExampleNPCShop`, `WeaponWithGrowingDamage`, `ExampleGlobalNPC` were updated.
And of course all town npcs. (`ExamplePerson`, `ExampleTravelingMerchant` and `ExampleBoneMerchant`)

### Porting Notes
`ModPylon.IsPylonForSale(int npcType, Player player, bool isNPCHappyEnough)` -> `ModPylon.GetNPCShopEntry()`
`ModNPC.SetupShop(Chest shop, ref int nextSlot)` -> `ModNPC.ModifyActiveShop()`
`ModNPC.OnChatButtonClicked(bool firstButton, ref bool shop)` -> `ModNPC.OnChatButtonClicked(bool firstButton, ref string shopName)`
`GlobalNPC.SetupShop(int type, Chest shop, ref int nextSlot)` -> `GlobalNPC.ModifyActiveShop(NPC npc, string shopName, Item[] items)`

#### FAQ
- Does this remove the 40 item per shop limit?
  - No, but it should be compatible with a future PR that does.
  - In the mean time, a warning message will be displayed to the user when the shop has too many items to fit.
- How can I randomize my shop items?
  - True randomization would mean contents change every time the shop is opened, and is probably not desirable. Better to use something like moon phases, or a static variable which you change periodically. 
  - Several moon phase related conditions have been added, mostly to support skeleton merchant, but you can make your own conditions with a localization like "Randomly, with 10% chance depending on moon phase" or similar.
- How can I explicitly lay out my shop like DD2 bartender?
  - The only way to do that would be through adding entries and to call `Entry.ReserveSlot()` method afterwards. If conditions are not met, then it will put 'air' item instead of simply not doing anything. Example:
  ```
  shop.Add(
    new NPCShop.Entry(
      ModContent.ItemType<ExampleItem>(),
      Condition.IsMoonFull)
    .ReserveSlot()
  )
  ```
- Can I use this to add items to travelling merchant's shop?
  - Not at this stage. no.
  - Please continue to use `ModifyActiveShop`
    - You can use `TravellingMerchantShop.AddInfoEntry` to add entries to the 'informational' list of the shop, so mods like recipe browser can show how to obtain the item.


Solxanich:
Internal note: previously #3072

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,831 @@
...
+using Terraria.ID;
+using Terraria.Localization;
+
+namespace Terraria.ModLoader
+{
+	// can and/or will be used for other tml loots later. (ie pots loot)
+	public class TMLLootDatabase
+	{
+		private Dictionary<int, ChestLoot> npcShopByType = new();
+		private List<ChestLoot.Entry> globalNpcShopEntries = new();
```
Solxanich: proposed is `private Dictionary<string, ChestLoot> npcShopByName = new();`
and when registering for a mod, we need `ModNPC.FullName + "/" + ShopTitle`
For vanilla, let's use Terraria/{NPCName}/ShopTitle

By default, when registering a new Shop, shop title is 'Shop'. Modder can change during setup shop step.

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,831 @@
...
+			RegisterCyborg();
+			RegisterPainter();
+			RegisterWitchDoctor();
+			RegisterPirate();
+			RegisterStylist();
+			RegisterSkeletonMerchant();
+			RegisterBartender();
+			RegisterGolfer();
+			RegisterZoologist();
+			RegisterPrincess();
```
Solxanich: Double check for updates to shops from 1.4.4
For example Painter has a few extra items I believe

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,321 @@
...
+		return PutAt(destination, item, false, condition);
+	}
+
+	public bool InsertAfter(int destination, int item, params ICondition[] condition) {
+		return PutAt(destination + 1, item, true, condition);
+	}
+
+	public bool InsertAfter(int destination, Item item, params ICondition[] condition) {
+		return PutAt(destination + 1, item, true, condition);
+	}
```
Solxanich: Is Destination referring to the index of the slot or the target Item to be before/after?

If the latter, please rename to targetItem.
If the former, might need some more work

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,321 @@
...
+		public static readonly Condition TenthAnniversary = new(NetworkText.Empty, () => Main.tenthAnniversaryWorld);
+		public static readonly Condition BirthdayPartyIsUp = new(NetworkText.Empty, () => BirthdayParty.PartyIsUp);
+		public static readonly Condition NightLanternsUp = new(NetworkText.Empty, () => LanternNight.LanternsUp);
+		public static readonly Condition IsMoonFull = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.Full);
+		public static readonly Condition IsMoonWaningGibbous = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.ThreeQuartersAtLeft);
+		public static readonly Condition IsMoonThirdQuarter = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.HalfAtLeft);
+		public static readonly Condition IsMoonWaningCrescent = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.QuarterAtLeft);
+		public static readonly Condition IsMoonNew = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.Empty);
+		public static readonly Condition IsMoonWaxingCrescent = new(NetworkText.Empty, () => Main.GetMoonPhase() == MoonPhase.QuarterAtRight);
```
Solxanich: We will want to replace the NetworkText.Empty with something meaningful before merge.

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,831 @@
...
+			shop.Add(1120);
+			shop.Add(1969, new ChestLoot.Condition(NetworkText.Empty, () => Main.netMode == NetmodeID.MultiplayerClient));
+			shop.Add(3248, ChestLoot.Condition.Halloween);
+			shop.Add(1741, ChestLoot.Condition.Halloween);
+			shop.Add(2871, ChestLoot.Condition.IsMoonFull);
+			shop.Add(2872, ChestLoot.Condition.IsMoonFull);
+			shop.Add(4663, ChestLoot.Condition.BloodMoon);
+			shop.Add(4662, ChestLoot.Condition.InGraveyard);
+			RegisterNpcShop(12, shop);
+		}
```
Solxanich: Can we get ExamplePerson to be updated with an example like this in Setup Shops?

Just a brief couple of Example items being added with some standard conditions, and setting the shop name as Examplium

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,321 @@
...
+
+		AddCandidates(oldEntries);
+		foreach (Entry group in oldEntries) {
+			group.TryAdd(array);
+		}
+
+		slots = array.Count;
+		if (array.Count < 40) {
+			array.AddRange(Enumerable.Repeat(emptyInstance, 40 - array.Count));
+		}
```
Solxanich: Add additional check for array.Count > 40, and if so, output warning to log and chat that all items could not fit in the shop

Solxanich (changes requested):
Initial Pass Through.

Great job getting this going, I want to say we are more than half way there.

Solxanich (approved):
Pulled, tested, seems to be working well.
Code at high level looks good and not seeing any major red flags.

SetupLootboxShop is the only thing somewhat iffy about. I feel like that could return delegate, or return entire array, and allow us to get it out of the SetupShop core. 
I'll leave that for a second opinion

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -144,15 +144,24 @@ public class WeaponWithGrowingDamage : GlobalItem
...
+		public override void ModifyActiveShop(NPC npc, string shopId, Item[] items) {
+			if (shopId != TMLLootDatabase.CalculateShopName(ModContent.NPCType<ExamplePerson>(), "ExampliumShop")) {
+				return;
+			}
+
+			for (int i = 0; i < items.Length; i++) {
+				Item item = items[i];
+				if (item.type == ItemID.Snowball && item.TryGetGlobalItem(out WeaponWithGrowingDamage weapon)) {
+					weapon.GainExperience(item, 2); // can buy snowballs with 2xp!
+				}
```
Chicken-Bones: This doesn't need to use `ModifyActiveShop`, just need to use the `Item` variant when adding the snowball to the shop.

ExampleMod/Common/GlobalNPCs/ExampleNPCShop.cs
```diff
@@ -7,40 +7,31 @@ namespace ExampleMod.Common.GlobalNPCs
 {
 	class ExampleNPCShop : GlobalNPC
 	{
-		public override void SetupShop(int type, Chest shop, ref int nextSlot) {
+		public override void SetupShop(string shopId, ChestLoot shopContents) {
 			// This example does not use the AppliesToEntity hook, as such, we can handle multiple npcs here by using if statements.
```
Chicken-Bones: Comment no-longer relevant.

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,1240 @@
...
+		public void RegisterGlobalNpcShop(ChestLoot.Entry entry) => globalNpcShopEntries.Add(entry);
+
+		public ChestLoot GetNpcShop(string fullName) {
+			if (npcShopByName.ContainsKey(fullName))
+				return npcShopByName[fullName];
+
+			return null;
+		}
+
+		public static string CalculateShopName(int npcId, string shopName = "Shop")
```
Chicken-Bones: Prefer `GetNPCShopName`

ExampleMod/Common/GlobalNPCs/ExampleNPCShop.cs
```diff
@@ -7,40 +7,31 @@ namespace ExampleMod.Common.GlobalNPCs
...
-			if (type == NPCID.Dryad) {
+			if (shopId == TMLLootDatabase.CalculateShopName(NPCID.Dryad, "Shop")) {
```
Chicken-Bones: Lets not pass the `"Shop"` arg as it's default

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+	public Entry this[Index index] {
+		get {
+			var ind2 = items.ElementAtOrDefault(index);
+			bool hasInNormal = ind2.item != defaultInstance;
+			if (hasInNormal)
+				return ind2;
+
+			return putCandidates2.ElementAtOrDefault(index);
+		}
+	}
```
Chicken-Bones: I see no reason to support indexing here.

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+
+	public ChestLoot InsertAfter(int targetItem, Item item, params ICondition[] condition) {
+		return PutAt(targetItem, item, true, condition);
+	}
+
+	public ChestLoot Hide(int item) {
+		this[item].Hide();
+		return this;
+	}
```
Chicken-Bones: As above, if people want to hide an entry, they should probably be searching for it, and then they have an `Entry` index they can call. I'd advise against hard-coding vanilla entry indices in mods

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+	public void RegisterShop(int npcId, string name = "Shop") {
+		this.name = name;
+		Main.TMLLootDB.RegisterNpcShop(npcId, this, name);
+	}
+
+	private void AddCandidates(List<Entry> entries) {
+		List<(int nextTo, bool after)> candidates = putCandidates;
+		List<Entry> candidates2 = putCandidates2;
+		candidates.Reverse();
+		candidates2.Reverse();
```
Chicken-Bones: This mutates the original lists... so calling `AddCandidates` multiple times will mess up

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+
+	public ChestLoot Add(Item item, params ICondition[] condition) {
+		return Add(new Entry(item, condition));
+	}
+
+	private ChestLoot PutAt(int destination, Item item, bool after, params ICondition[] condition) {
+		putCandidates.Add(new(destination, after));
+		putCandidates2.Add(new(item, condition));
+		return this;
+	}
```
Chicken-Bones: Please see `Recipe.Ordering` and `SetOrdering` for a good design on how to implement a simple ordering system. 
- Ordering should be specified using `Entry` references, not indices. 
- It's ok to provide a helper which takes `int targetItem` and searches for an entry with that item id, but not required
- Ordering loops should be reported as soon as they're formed

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+		this[item].Hide();
+		return this;
+	}
+
+	public Item[] Build(bool lastSlotEmpty = true) {
+		return Build(out _, lastSlotEmpty);
+	}
+
+	public Item[] Build(out int slots, bool lastSlotEmpty = true) {
+		List<Item> array = new();
```
Chicken-Bones: Any reason to use a list instead of an `Item[40]`?

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,156 @@
...
+		AddCandidates(oldEntries);
+		foreach (Entry group in oldEntries) {
+			group.TryAdd(array);
+		}
+
+		slots = array.Count;
+		if (array.Count > 39) {
+			Main.NewText("Not all items could fit in the shop!", 255, 0, 0);
+			Logging.tML.Warn($"Unable to fit all item in the shop {name}");
+			slots = 39;
```
Chicken-Bones: This doesn't account for `lastSlotEmpty`

patches/tModLoader/Terraria/ModLoader/ChestLoot.Condition.cs
```diff
@@ -0,0 +1,115 @@
...
+public partial class ChestLoot {
+	public interface ICondition {
+		string Description { get; }
+
+		bool IsAvailable();
+	}
+
+	public sealed class Condition : ICondition {
+		#region Common conditions
+		public static readonly Condition Christmas = new(NetworkText.FromLiteral("During Christmas"), () => Main.xMas);
```
Chicken-Bones: Need to add localizations for all of these. No `FromLiteral`

Intent is to change `Description` from `string` and `NetworkText` to `LocalizedText` soon

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		public Entry OnSuccess(Entry entry) {
+			ChainedEntries[true].Add(entry);
+			return this;
+		}
+
+		public Entry OnFail(int itemId, params ICondition[] condition) {
+			return OnFail(new Entry(ContentSamples.ItemsByType[itemId], condition));
+		}
```
Chicken-Bones: Is `OnFail` necessary? That seems like it would be more appropriate for things that were chance based.
It seems like it would be appropriate for, well, `ChestLoot` or full loot tables, but since this is just conditional visibility I think it might be better to have all conditions be 'top level' (like recipes). Needs discussion

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace Terraria.ModLoader;
+
+public partial class ChestLoot {
+	private static Item EmptyInstance => new();
+
+	public class Entry {
+		internal readonly Item item;
+		private readonly List<ICondition> conditions;
+		private readonly bool askingNicelyToNotAdd = false;
+		private Ref<bool> hide;
```
Chicken-Bones: why `Ref<bool>`?

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Terraria.ModLoader;
+
+public partial class ChestLoot {
+	private static Item EmptyInstance => new();
+
+	public class Entry {
+		internal readonly Item item;
+		private readonly List<ICondition> conditions;
+		private readonly bool askingNicelyToNotAdd = false;
```
Chicken-Bones: No need for this, just check if `item.IsAir()` in `TryAdd`

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,96 @@
...
+		public bool IsAvailable() {
+			foreach (ICondition condition in conditions) {
+				if (!condition.IsAvailable()) {
+					return false;
+				}
+			}
+			return true;
+		}
+
+		public void TryAdd(List<Item> items) {
```
Chicken-Bones: Generally not advisable to call it `Try` when it doesn't return a `bool` and success doesn't depend on the parameters.

Chicken-Bones (changes requested):

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -144,15 +144,29 @@ public class WeaponWithGrowingDamage : GlobalItem
...
+			shopContents.Add(snowball);
+		}
+
+		public override void ModifyActiveShop(NPC npc, string shopId, Item[] items) {
+			if (shopId != TMLLootDatabase.GetNPCShopName(ModContent.NPCType<ExamplePerson>(), "ExampliumShop")) {
+				return;
+			}
+
+			for (int i = 0; i < items.Length; i++) {
+				Item item = items[i];
```
Chicken-Bones: `foreach (var item in items)`

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -144,15 +144,29 @@ public class WeaponWithGrowingDamage : GlobalItem
...
+			if (shopId != TMLLootDatabase.GetNPCShopName(ModContent.NPCType<ExamplePerson>(), "ExampliumShop")) {
 				return;
 			}
 
-			Item item = shop.item[nextSlot++];
-			item.SetDefaults(ItemID.Snowball);
-			if (item.TryGetGlobalItem(out WeaponWithGrowingDamage weapon)) {
-				weapon.GainExperience(item, 2); // can buy snowballs with 2xp!
+			var snowball = new Item();
+			snowball.SetDefaults(ItemID.Snowball);
```
Chicken-Bones: `new Item(ItemID.Snowball)`

ExampleMod/Common/GlobalNPCs/ExampleNPCShop.cs
```diff
@@ -7,40 +7,30 @@ namespace ExampleMod.Common.GlobalNPCs
...
 				// Editing item.value in SetupShop is an incorrect approach.
 
 				// This shop entry sells for 2 Defenders Medals.
-				shop.item[nextSlot].SetDefaults(ModContent.ItemType<ExampleMountItem>());
-				shop.item[nextSlot].shopCustomPrice = 2;
-				shop.item[nextSlot].shopSpecialCurrency = CustomCurrencyID.DefenderMedals; // omit this line if shopCustomPrice should be in regular coins.
-				nextSlot++;
+				shopContents.Add(ModContent.ItemType<ExampleMountItem>());
+				shopContents.LastEntry.Item.shopCustomPrice = 2;
+				shopContents.LastEntry.Item.shopSpecialCurrency = CustomCurrencyID.DefenderMedals; // omit this line if shopCustomPrice should be in regular coins.
```
Chicken-Bones:
```cs
new Item(ModContent.ItemType<ExampleMountItem>()) {
    shopCustomPrice = 2,
    shopSpecialCurrency = CustomCurrencyID.DefenderMedals
}
```

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -700,16 +700,24 @@ public virtual void SetChatButtons(ref string button, ref string button2)
...
 	{
 	}
 
-	/// <summary>
-	/// Allows you to add items to this NPC's shop. Add an item by setting the defaults of shop.item[nextSlot] then incrementing nextSlot. In the end, nextSlot must have a value of 1 greater than the highest index in shop.item that contains an item.
-	/// </summary>
-	/// <param name="shop"></param>
-	/// <param name="nextSlot"></param>
-	public virtual void SetupShop(Chest shop, ref int nextSlot)
+	public virtual void SetupShop()
```
Chicken-Bones: This needs to be called `AddShops` akin to `AddRecipes`

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -700,16 +700,24 @@ public virtual void SetChatButtons(ref string button, ref string button2)
...
+	{
+	}
+
+	public virtual List<Item> SetupLootboxShop()
+	{
+		return null;
+	}
+
+	public virtual void PostSetupShop()
```
Chicken-Bones: There is no use-case for this on `ModNPC`

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Terraria.ID;
+
+namespace Terraria.ModLoader;
+
+public partial class ChestLoot {
```
Chicken-Bones: Needs to be called `NPCShop`

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,116 @@
...
+			bool hasInNormal = index != -1;
+			return items[index];
+		}
+	}
+
+	public ChestLoot() {
+		items = new();
+	}
+
+	public void RegisterShop(int npcId, string name = "Shop") {
```
Chicken-Bones: These arguments belong in the constructor, alongside the class rename. Then this can just be `Register`

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -700,16 +700,24 @@ public virtual void SetChatButtons(ref string button, ref string button2)
...
+	{
+	}
+
+	public virtual void ModifyActiveShop(string shopId, Item[] items)
```
Chicken-Bones: `shopId` -> `shopName`

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -626,13 +627,19 @@ public virtual void OnChatButtonClicked(NPC npc, bool firstButton)
...
 	}
 
-	/// <summary>
-	/// Allows you to add items to an NPC's shop. The type parameter is the type of the NPC that this shop belongs to. Add an item by setting the defaults of shop.item[nextSlot] then incrementing nextSlot. In the end, nextSlot must have a value of 1 greater than the highest index in shop.item that contains an item. If you want to remove an item, you will have to be familiar with programming.
-	/// </summary>
-	/// <param name="type"></param>
-	/// <param name="shop"></param>
-	/// <param name="nextSlot"></param>
-	public virtual void SetupShop(int type, Chest shop, ref int nextSlot)
+	public virtual void SetupShop(string shopId, ChestLoot shopContents)
```
Chicken-Bones: Should be called `ModifyShop`

No need for the `shopId` argument, as the name and NPC type of the shop should be obtainable from the `shopContents` parameter (which can just be called `shop`)

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -700,16 +700,24 @@ public virtual void SetChatButtons(ref string button, ref string button2)
 	/// </summary>
 	/// <param name="firstButton"></param>
 	/// <param name="shop"></param>
-	public virtual void OnChatButtonClicked(bool firstButton, ref bool shop)
+	public virtual void OnChatButtonClicked(bool firstButton, ref string shopName)
```
Chicken-Bones, line 702: Need to adjust documentation. Need to be very clear that `shopName` should be set to `"Shop"` for the default shop, and otherwise can be a custom identifier which matches the name of a registered `NPCShop` instance for this NPC.

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -626,13 +627,19 @@ public virtual void OnChatButtonClicked(NPC npc, bool firstButton)
...
+	{
+	}
+
+	public virtual void SetupLootboxShop(int type, List<Item> items)
+	{
+	}
+
+	public virtual void PostSetupShop(string shopId, ChestLoot shopContents)
```
Chicken-Bones: No use case for this. Other loot systems don't have a hook like this.

If, for some reason, there is a need to modify shops after other `GlobalNPC.ModifyShop` hooks have run, `ModSystem.AddRecipes` or `PostAddRecipes` should suffice

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,1083 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Terraria.GameContent;
+using Terraria.ID;
+using Terraria.Localization;
+
+namespace Terraria.ModLoader;
+
+// can and/or will be used for other tml loots later. (ie pots loot)
+public class TMLLootDatabase
```
Chicken-Bones: Lets keep this specific to shops for now. There's no reason to combine other types of loot tables at this stage, especially since shops and pots and chests all have significant differences.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -700,16 +700,24 @@ public virtual void SetChatButtons(ref string button, ref string button2)
...
-	/// Allows you to add items to this NPC's shop. Add an item by setting the defaults of shop.item[nextSlot] then incrementing nextSlot. In the end, nextSlot must have a value of 1 greater than the highest index in shop.item that contains an item.
-	/// </summary>
-	/// <param name="shop"></param>
-	/// <param name="nextSlot"></param>
-	public virtual void SetupShop(Chest shop, ref int nextSlot)
+	public virtual void SetupShop()
+	{
+	}
+
+	public virtual List<Item> SetupLootboxShop()
```
Chicken-Bones: It is unclear at this point how to use `SetupLootboxShop`. I understand it's intended to be an access point for travelling merchant style shops, but it doesn't currently have much (if any) advantage over `ModifyActiveShop` for that use case, and could be separated to keep this PR smaller and easier to review

patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -337,6 +337,8 @@ internal static void Load(CancellationToken token)
 		ItemSorting.SetupWhiteLists();
 		ItemLoader.ValidateGeodeDropsSet();
 
+		Main.TMLLootDB.Initialize();
```
Chicken-Bones: Should go next to `SetupBestiary`, before `SetupRecipes`

patches/tModLoader/Terraria/ModLoader/TMLLootDatabase.cs
```diff
@@ -0,0 +1,1083 @@
...
+{
+	private Dictionary<string, ChestLoot> npcShopByName = new();
+	private List<ChestLoot.Entry> globalNpcShopEntries = new();
+
+	public void RegisterNpcShop(int npcId, ChestLoot chestLoot, string shopName = "Shop") {
+		npcShopByName.Add(GetNPCShopName(npcId, shopName), chestLoot);
+	}
+
+	public void RegisterGlobalNpcShop(ChestLoot.Entry entry) => globalNpcShopEntries.Add(entry);
```
Chicken-Bones: I am unsure that this is good design. 
On the one hand, being able to see which entries for shops are 'global' could be useful for some mods? 
On the other, this can currently be achieved with a tML `GlobalNPC.ModifyShop` (the load time one, not `ModifyActiveShop`)

With this implementation, it is also unclear _when_ it is valid to register a global shop entry.

ExampleMod/Content/NPCs/ExampleGlobalNPC.cs
```diff
@@ -27,17 +31,54 @@ public class ExampleGlobalNPC : GlobalNPC
...
+			// Let's add an item that appears just during Windy day and when NPC is happy enough (can sell pylons)
+			// If condition is fulfilled, add an item to the shop.
+			shopContents.Add(ModContent.ItemType<ExampleItem>(), ChestLoot.Condition.HappyWindyDay, ChestLoot.Condition.HappyEnough);
+
+			// Otherwise, if condition is not fulfilled, then let's check if its For The Worthy world and then sell Red Potion.
+			shopContents.Add(ItemID.RedPotion, redPotCondition, ChestLoot.Condition.ForTheWorthy);
+		}
+
+		// PostSetupShop hook is best when it comes to modifying existing items.
+		public override void PostSetupShop(string shopId, ChestLoot shopContents) {
```
Chicken-Bones: This could just be merged with the above method. The rationale for the split only applies if you have a `GlobalNPC` which needs to modify items added by another `GlobalNPC`, in which case you probably have a dependency and mod load order is your friend.

ExampleMod/Content/NPCs/ExampleGlobalNPC.cs
```diff
@@ -27,17 +31,54 @@ public class ExampleGlobalNPC : GlobalNPC
...
+			// Style 1 check for application
+			if (shopId != TMLLootDatabase.GetNPCShopName(NPCID.Merchant, "Shop"))
+				return;
+
+			// Style 2 check for application
+			if (shopId != "Terraria/Merchant/Shop")
+				return;
+
+			// Creaate a condition for later use.
+			var redPotCondition = new ChestLoot.Condition(NetworkText.FromLiteral("Describe what your condition is!"), () => !ChestLoot.Condition.HappyWindyDay.IsAvailable() || !ChestLoot.Condition.HappyEnough.IsAvailable());
```
Chicken-Bones: Needs to be `NetworkText.FromKey`. Localization is a priority. I'll be changing recipe conditions to use `LocalizedText` soon

ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -234,64 +234,36 @@ public class ExamplePerson : ModNPC
 					return;
 				}
 
-				shop = true;
+				shop = "ExampliumShop"; // Name of the shop tab we want to open.
```
Chicken-Bones: It is unclear why this NPC has a non-default shop name. We should make the example include 2 shops if the intent is to showcase this feature properly. Otherwise it's misleading

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -189,6 +195,24 @@
...
++	public void SetupShop(string shop, int type)
+ 	{
+ 		bool flag = Main.LocalPlayer.currentShoppingSettings.PriceAdjustment <= 0.8999999761581421;
+ 		Item[] array = item;
+@@ -1175,7 +_,7 @@
+ 		}
+ 
+ 		int num = 0;
+-		switch (type) {
++		/*switch (type) {
```
Chicken-Bones: put the `/*` on the line above

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1115,6 +1115,35 @@
  		ClientInitialize();
  		base.Initialize();
  	}
+@@ -5179,6 +_,8 @@
+ 		ItemDropsDB = itemDropDatabase;
+ 		bestiaryDatabase.Merge(ItemDropsDB);
+ 		bestiaryDatabaseNPCsPopulator.AddDropOverrides(bestiaryDatabase);
++		TMLLootDB = new TMLLootDatabase();
++		TMLLootDB.Initialize();
```
Chicken-Bones: Why have this be an instance, rather than a static class?

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1115,6 +1115,35 @@
...
++				continue;
++
++			int npcType = NPCLoader.shopToNPC[m];
++
++			if (m == 25) { // Painter 2 Shop Special Case
++				shop[m].SetupShop(TMLLootDatabase.GetNPCShopName(npcType, "Decor"), m);
++				continue;
++			}
++
++			shop[m].SetupShop(TMLLootDatabase.GetNPCShopName(npcType), m);
```
Chicken-Bones: This lookup code is common in two places, so it should go in `NPCLoader` or the loot database. It should be a function that takes `int shopType` and returns `NPCShop`

patches/tModLoader/Terraria/ModLoader/ChestLoot.cs
```diff
@@ -0,0 +1,116 @@
...
+
+	public Entry LastEntry => items[^1];
+
+	public Entry this[int item] {
+		get {
+			int index = items.FindIndex(x => x.item.type.Equals(item));
+			bool hasInNormal = index != -1;
+			return items[index];
+		}
+	}
```
Chicken-Bones: This should not be a publicly indexable data structure

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,112 @@
...
+			return this;
+		}
+
+		public Entry AddCondition(ICondition condition) {
+			ArgumentNullException.ThrowIfNull(condition, nameof(condition));
+			conditions.Add(condition);
+			return this;
+		}
+
+		public Entry Hide() {
```
Chicken-Bones: Should be called `Disable`

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace Terraria.ModLoader;
+
+public partial class ChestLoot {
+	private static Item EmptyInstance => new();
+
+	public class Entry {
+		internal readonly Item item;
+		private readonly List<ICondition> conditions;
+		private bool hide;
```
Chicken-Bones: Should be `public bool Disabled { get; private set; }`

patches/tModLoader/Terraria/ModLoader/ChestLoot.Entry.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Linq;
+using System;
+using Terraria.ID;
+
+namespace Terraria.ModLoader;
+
+public partial class ChestLoot {
+	private static Item EmptyInstance => new();
+
+	public class Entry {
```
Chicken-Bones: With `Entry` being public, we need a review of the public API surface, to make sure that all `public` functions are meant to be used by modders, and that there aren't duplicates with the main class.

Chicken-Bones (changes requested):

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -144,15 +144,27 @@ public class WeaponWithGrowingDamage : GlobalItem
 	}
 	public class SnowBallShop : GlobalNPC
 	{
-		public override void SetupShop(int type, Chest shop, ref int nextSlot) {
-			if (type != ModContent.NPCType<ExamplePerson>()) {
+		public override void ModifyShop(NPCShop shop) {
+			if (shop.FullName != NPCShopDatabase.GetNPCShopName(ModContent.NPCType<ExamplePerson>(), "ExampliumShop")) {
```
Chicken-Bones: Should compare `shop.NpcID` and `Name` directly

ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
```diff
@@ -144,15 +144,27 @@ public class WeaponWithGrowingDamage : GlobalItem
...
+
+		public override void ModifyActiveShop(NPC npc, string shopName, Item[] items) {
+			if (shopName != NPCShopDatabase.GetNPCShopName(ModContent.NPCType<ExamplePerson>(), "ExampliumShop")) {
+				return;
+			}
+
+			foreach (Item item in items) {
+				if (item.type == ItemID.Snowball && item.TryGetGlobalItem(out WeaponWithGrowingDamage weapon)) {
+					weapon.GainExperience(item, 2); // can buy snowballs with 2xp!
+				}
```
Chicken-Bones: This code seems unnecessary? A bad port? You already add snowballs with 2xp to the shop in `ModifyShop`. This would increase their `xp` to 4

ExampleMod/Common/GlobalNPCs/ExampleNPCShop.cs
```diff
@@ -7,40 +7,32 @@ namespace ExampleMod.Common.GlobalNPCs
 {
 	class ExampleNPCShop : GlobalNPC
 	{
-		public override void SetupShop(int type, Chest shop, ref int nextSlot) {
-			// This example does not use the AppliesToEntity hook, as such, we can handle multiple npcs here by using if statements.
-			if (type == NPCID.Dryad) {
+		public override void ModifyShop(NPCShop shop) {
+			if (shop.FullName == NPCShopDatabase.GetNPCShopName(NPCID.Dryad)) {
```
Chicken-Bones: As above, for NPCs with only one shop, you can just check `shop.NpcID`

ExampleMod/Content/NPCs/ExampleGlobalNPC.cs
```diff
@@ -27,17 +31,53 @@ public class ExampleGlobalNPC : GlobalNPC
...
+			// Safer approach for disabling item
+			if (shop.TryGetEntry(ItemID.CopperPickaxe, out NPCShop.Entry entry)) {
+				entry.Disable();
+			}
+
+			// Adding new Condition to Blue Flare. Now it will appear just if player carries a Flare Gun in their inventory AND is in Snow biome
+			shop.GetEntry(ItemID.BlueFlare).AddCondition(NPCShop.Condition.InSnowBiome);
+
+			// Creaate a condition for later use.
+			var redPotCondition = new NPCShop.Condition(NetworkText.FromKey("Mods.ExampleMod.ShopConditions.RedPotCondition"), () => !NPCShop.Condition.HappyWindyDay.IsAvailable() || !NPCShop.Condition.HappyEnough.IsAvailable());
```
Chicken-Bones: Should go next to the usage. 
Comment should say 'custom condition, opposite of conditions for `ExampleItem` above'

ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -234,64 +234,36 @@ public class ExamplePerson : ModNPC
...
-		// 	// 		shop.item[nextSlot].SetDefaults(Item.Type);
-		// 	// 		shop.item[nextSlot].shopCustomPrice = 0;
-		// 	// 		shop.item[nextSlot].GetGlobalItem<ExampleInstancedGlobalItem>().examplePersonFreeGift = true;
-		// 	// 		nextSlot++;
-		// 	// 		//TODO: Have tModLoader handle index issues.
-		// 	// 	}
-		// 	// }
-		// }
+		public override void AddShops() {
+			var exampleGunBulletCondition = new NPCShop.Condition(NetworkText.FromLiteral("If its third quarter or waning crescent moon"), () => Main.moonPhase >= 2 && Main.moonPhase < 4);
```
Chicken-Bones: No `FromLiteral`!

ExampleMod/Content/NPCs/ExamplePerson.cs
```diff
@@ -234,64 +234,36 @@ public class ExamplePerson : ModNPC
...
+			var chestLoot = new NPCShop(Type, "ExampliumShop")
```
Chicken-Bones: needs a rename from `chestLoot`

ExampleMod/Content/NPCs/ExampleTravelingMerchant.cs
```diff
@@ -248,22 +248,16 @@ class ExampleTravelingMerchant : ModNPC
...
-					continue;
-
-				shop.item[nextSlot].SetDefaults(item.type);
-				nextSlot++;
-			}
-		}
+		// TODO: add that method
+		/*public override List<Item> SetupLootboxShop() {
+			return shopItems;
+		}*/
```
Chicken-Bones: Hm? Still need the old code, to fill out the shop the old fashioned way

ExampleMod/Localization/en-US.hjson
```diff
@@ -186,5 +186,10 @@ Mods: {
 			ExampleTip2: "This is yet another tip added by Example Mod!"
 			DisabledExampleTip: "This tip will be added but then disabled by Example Mod."
 		}
+
+		#This is used for shop conditions.
+		ShopConditions: {
+			RedPotCondition: "Describe what your condition is!"
```
Chicken-Bones: Proper localization please. `"When not happy enough or not a happy windy day"` or something...

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -189,6 +195,24 @@
...
++	public void SetupShop(string shop, int type)
+ 	{
+ 		bool flag = Main.LocalPlayer.currentShoppingSettings.PriceAdjustment <= 0.8999999761581421;
+ 		Item[] array = item;
+@@ -1174,7 +_,7 @@
+ 			array[i] = new Item();
+ 		}
+ 
+-		int num = 0;
++		/*int num = 0;
```
Chicken-Bones: Start block comment on line above

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -204,16 +228,44 @@
...
 +			// Extra patch context.
  			if (Main.player[Main.myPlayer].ZoneGlowshroom && (!Main.remixWorld || Main.player[Main.myPlayer].Center.Y / 16f < (float)(Main.maxTilesY - 200)) && num < 39)
  				array[num++].SetDefaults(4921);
  		}
-+
-+		NPCLoader.SetupShop(type, this, ref num);
++		*/
  
++		int num = 0;
++		if (type == 19) // Travelling Merchant, move in to the LootboxShop setup method
```
Chicken-Bones: comment not accurate anymore, remove

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1115,6 +1115,19 @@
...
+@@ -5242,7 +_,11 @@
+ 		Chest.SetupTravelShop();
+ 		for (int m = 1; m < 100; m++) {
+ 			shop[m] = new Chest();
+-			shop[m].SetupShop(m);
++
++			if (m >= NPCLoader.shopToNPC.Length)
++				continue;
++
++			shop[m].SetupShop(NPCShopDatabase.GetNpcShopByIndex(m), m);
```
Chicken-Bones: Is this necessary? Can we just leave `shop` full of blank `new Chest()` values until the player opens one of them, since we rebuild every time anyway?

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -626,13 +627,11 @@ public virtual void OnChatButtonClicked(NPC npc, bool firstButton)
...
-	/// </summary>
-	/// <param name="type"></param>
-	/// <param name="shop"></param>
-	/// <param name="nextSlot"></param>
-	public virtual void SetupShop(int type, Chest shop, ref int nextSlot)
+	public virtual void ModifyShop(NPCShop shop)
+	{
+	}
+
+	public virtual void ModifyActiveShop(NPC npc, string shopName, Item[] items)
```
Chicken-Bones: Documentation required

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1149,19 +1150,28 @@ public static void OnChatButtonClicked(bool firstButton)
 		}
 	}
 
-	private delegate void DelegateSetupShop(int type, Chest shop, ref int nextSlot);
-	private static HookList HookSetupShop = AddHook<DelegateSetupShop>(g => g.SetupShop);
+	private delegate void DelegateModifyShop(NPCShop shop);
+	private static HookList HookModifyShop = AddHook<DelegateModifyShop>(g => g.ModifyShop);
```
Chicken-Bones: Should go below `AddShops` next to `ModifyShop`

patches/tModLoader/Terraria/ModLoader/NPCShop.cs
```diff
@@ -0,0 +1,125 @@
...
+	public string Name => name;
+	public string FullName => NPCShopDatabase.GetNPCShopName(NpcID, Name);
+
+	public Entry this[int item] {
+		get {
+			int index = items.FindIndex(x => x.item.type.Equals(item));
+			bool hasInNormal = index != -1;
+			return items[index];
+		}
+	}
```
Chicken-Bones: Please remove, since we now have `GetEntry`

patches/tModLoader/Terraria/ModLoader/NPCShop.cs
```diff
@@ -0,0 +1,125 @@
...
+
+	public NPCShop InsertAfter(int targetItem, Item item, params ICondition[] condition) {
+		return InsertAt(targetItem, item, true, condition);
+	}
+
+	public Item[] Build() {
+		return Build(out _);
+	}
+
+	public Item[] Build(out int slots) {
```
Chicken-Bones: Is there a reason not to pass in the `Item[]` here? I don't like that this method is hard-coded to 40 slots with a lot of bits and pieces

patches/tModLoader/Terraria/ModLoader/NPCShop.cs
```diff
@@ -0,0 +1,125 @@
...
+		List<Item> array = new();
+		List<Entry> oldEntries = new(items);
+
+		SortBeforeAfter(oldEntries, r => r.Ordering);
+		foreach (Entry group in oldEntries) {
+			if (group.Disabled || !group.ConditionsMet()) {
+				continue;
+			}
+			array.Add(group.Item);
```
Chicken-Bones: could fill into the array directly, and break when hitting 39 items

patches/tModLoader/Terraria/ModLoader/NPCShop.cs
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		slots = array.Count;
+		int limit = 39;
+		if (array.Count > limit) {
+			Main.NewText("Not all items could fit in the shop!", 255, 0, 0);
+			Logging.tML.Warn($"Unable to fit all item in the shop {name}");
+			slots = limit;
+		}
```
Chicken-Bones: This doesn't belong in here. Not nice to have methods which seem 'pure' on the outside, to be calling game functions. Better to just have a `bool` return, or an `out bool overflow`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/NPCShopDatabase.cs
```diff
@@ -0,0 +1,1091 @@
...
+
+	public static Dictionary<string, NPCShop> GetNpcShopsOf(int npcId) {
+		return npcShopByName.Where(x => {
+			var split = x.Key.Split('/');
+			return split[0] == (npcId < NPCID.Count ? "Terraria" : NPCLoader.GetNPC(npcId).Mod.Name)
+			&& split[1] == (npcId < NPCID.Count ? NPCID.Search.GetName(npcId) : NPCLoader.GetNPC(npcId).Name);
+		}).ToDictionary(x => x.Key, x => x.Value);
+	}
+
+	public static string GetNpcShopByIndex(int index)
```
Chicken-Bones: This should be called `GetVanillaShop`

patches/tModLoader/Terraria/ModLoader/NPCShopDatabase.cs
```diff
@@ -0,0 +1,1091 @@
...
+		return $"{(npcId < NPCID.Count ? $"Terraria/{NPCID.Search.GetName(npcId)}" : NPCLoader.GetNPC(npcId).FullName)}/{shopName}";
+	}
+
+	public static Dictionary<string, NPCShop> GetNpcShopsOf(int npcId) {
+		return npcShopByName.Where(x => {
+			var split = x.Key.Split('/');
+			return split[0] == (npcId < NPCID.Count ? "Terraria" : NPCLoader.GetNPC(npcId).Mod.Name)
+			&& split[1] == (npcId < NPCID.Count ? NPCID.Search.GetName(npcId) : NPCLoader.GetNPC(npcId).Name);
+		}).ToDictionary(x => x.Key, x => x.Value);
+	}
```
Chicken-Bones: No need for this method, can just allow people to enumerate all shops and then they can check `NPCType`

Chicken-Bones (changes requested):

PR #3222, opened 2023-01-29
Support paint spraying for any sized tile

Jadams505:
### What is the bug?
The paint sprayer can not auto paint certain modded tiles (depending on their `TileObjectData`) due to vanilla's 11x11 hard coded bounds around the players target tile. This is a particular problem when using certain combinations of `TileObjectData.Origin` and `TileObjecData.AnchorData` as well as tiles larger than 11x11. This bug even affects vanilla since the Christmas tree tile is 4x8 and anchored at the bottom so the top 4x2 tiles do not get painted.

This issue also affects the pressurator and its auto actuation feature in the same way.
### How did you fix the bug?
Reworked `Player.PlaceThing_Tiles_GetAutoAccessoryCache` to return a 2d array of the exact size of the tile to place/block swap  as opposed to the aforementioned 11x11 box around the player's tile target position. In order to do this I had to add parameters for the `TileObjectData` and the top left tile position where the tile will be placed.

Additionally `Player.PlaceThing_Tiles_PlaceIt_AutoPaintAndActuate` also needed the top left tile position parameter added to it since `tileTargetX` and `tileTargetY` are not reliable due to `TileObjectData.Origin` and `TileObjectData.AnchorData` .
The current alternative is to use the `ModTile.PlaceInWorld` hook.
### Notes
Vanilla's logic makes it so that in a block swap the new tile is only painted if the tile type is changed. This means that in the case of swapping vanilla and modded chests and dressers the new tiles will always be painted. Whereas swapping two vanilla dressers will not paint the new tile. This funnily enough leads to vanilla inconsistencies since chests are split between two tile ids (21 and 467). Thus swapping a gold chest (id 21) and a frozen chest (id 21) won't paint the tile. But swapping a gold chest (id 21) and a nebula chest (id 467) will paint the tile.


JavidPack:
Thanks. Sorry this fell by the wayside.

Issue #3110, opened 2022-11-27
Make World Generation Passes Public Methods Instead Of Anonymous Methods


Destructor-Ben:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

The anonymous methods used for world generation passes in vanilla should be changed to be public methods, or at least named methods.

### What does this proposal attempt to solve or improve?

Currently, IL editing or On hooking world generation passes is very difficult, since the methods are anonymous. If they were manually changed from anonymous methods to public methods, it would allow modders to have more flexibility to change world generation code without having to replace entire generation passes. It would also allow multiple mods to modify the same generation pass without conflicting. An example of a mod that could end up using this feature is the Remnants Mod, since it changes a lot of world generation, and if it migrated to this new system, it would have the potential of being compatible with more mods, like Calamity.

### Which (other) solutions should be considered?

Making a guide on how to find the anonymous methods.
Making it so the anonymous methods used as parameters in the LegacyGenPass can be modified in ModifyWorldGenTasks (I'm not sure if this is even possible, since I'm not an expert)

Issue #3213, opened 2023-01-26
IL Editing QoL Additions


Destructor-Ben:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

1) Generate text files of IL before and after a method is edited, and create a diff of them. This could be done either automatically or a modder manually calling a method like `WriteILToFile(Mod mod, ILContext il)`, or `CreateILDiff(string filePath1, string filePath2)`.

2) A method that logs an error indicating that a mods IL editing has failed, e.g. "IL Patch Error: Mod 'XX' failed to patch method 'XX' in '{Class that did the patching}'". This isn't that important, but it would allow for consistent and detailed error messages between mods.

### What does this proposal attempt to solve or improve?

Currently, IL editing is painful because there isn't an easy way to actually tell what went wrong. With this, there is a consistent error message when IL editing fails and modders can easily check what changes their code actually made.

### Which (other) solutions should be considered?

Allowing modders to implement this themselves. This would require examples of how to do this in EM instead of being built into tML.

EDIT:

This should also include dumping a list of registered IL hooks and detours (moved from https://github.com/tModLoader/tModLoader/pull/3240)

Chicken-Bones:
For suggestion 2, what happens currently? What kind of failures do you want to protect against. Can you provide sample code which throws an unhelpful error?

Destructor-Ben:
> what happens currently?

I think most mods just say something like "IL edit failed" or don't say anything when a patch fails, though some have detailed error messages. I know at least 2 that don't have patch fail messages.


> What kind of failures do you want to protect against.

When the game updates some mods IL edits might not get updated and will fail, so it can help modders that aren't actively working their mods to fix them without hassle. It also is consistent, so you can always expect to get a lot of information no matter each mods implementation of the error message.

> Can you provide sample code which throws an unhelpful error?

This is the worst I've found so far:
`mod.Logger.Error("Do you want a cake?");`
This probably only exists so the mod developer knows that there is an error, but it doesn't say where.

One of the best I've found is Calamity, which has its own LogFaliure method for IL editing fails, which looks like this:
```
public static void LogFailure(string name, string reason)
{
	CalamityMod.Instance.Logger.Warn("IL edit \"" + name + "\" failed! " + reason);
}
```
Magic storage is also great but I can't be bothered pasting a snippet of code here, though one thing it does is track how many patches it has been able to do successfully until an error occurs.

[...]

PR #3240, opened 2023-02-05
IL Editing GenPasses

Destructor-Ben:
### What is the new feature?

Implementation of #3110 

Adds `WorldGen.ModifyPass` and `WorldGen.DetourPass` to allow easily IL hooking legacy world gen passes

Obtaining the `MethodBase` for the world gen delegates previously required using a temporary read-only IL hook and scanning the method for the reference. Now it can be pulled directly from the delegate.

```cs
// All of the registration should take place in load
// Generation pass hooks are unloaded manually, so no Unload method is needed
public override void Load() {
	// IL editing the pyramids pass
	WorldGen.ModifyPass((PassLegacy)WorldGen.VanillaGenPasses["Pyramids"], Modify_Pyramids);

	// Detouring the shinies pass (generates ore)
	WorldGen.DetourPass((PassLegacy)WorldGen.VanillaGenPasses["Shinies"], Detour_Shinies);
}
```

### ExampleMod updates
See `ExampleWorldGenHookingSystem`


patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -7,4 +19,80 @@ public partial class WorldGen
...
+			HookEndpointManager.Unmodify(GetGenPassMethod(ilEdit.Item1), ilEdit.Item2);
+		}
+
+		foreach (var detour in genPassDetours) {
+			HookEndpointManager.Remove(GetGenPassMethod(detour.Item1), detour.Item2);
+		}
+	}
+
+	// TODO: tidy up
+	internal static MethodInfo GetGenPassMethod(GenPass task)
```
Chicken-Bones: Recommend this and `DetourTask` above take a `PassLegacy`. Let the compiler push the requirement onto the user so they're aware.

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -7,4 +19,80 @@ public partial class WorldGen
...
+			}
+		}
+		else {
+			throw new Exception("Task is not a PassLegacy");
+			return null;
+		}
+	}
+
+	// TODO: Remove after testing and create DumpHooks for detours
+	internal static void DumpILHooks()
```
Chicken-Bones: This could be useful in a separate PR. A better place to access this info is in `MonoModHooks.cs`?

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -7,4 +19,80 @@ public partial class WorldGen
...
+	/// <summary>
+	/// IL edits a PassLegacy during world generation. Use in ModifyWorldGenerationTasks or ModifyHardmodeTasks.
+	/// </summary>
+	/// <param name="task">The task to be IL edited</param>
+	/// <param name="callback">The method that IL edits the pass</param>
+	// TODO: Add support for hardmode tasks
+	public static void ModifyTask(GenPass task, ILContext.Manipulator callback)
+	{
+		HookEndpointManager.Modify(GetGenPassMethod(task), callback);
+		genPassILEdits.Add(Tuple.Create(task, callback));
```
Chicken-Bones: I don't like the idea of adding and removing IL hooks after load time. It 'should' be safe, but it's a fair bit of overhead, repeat logging, more risk of errors etc.

Fortunately, 1.4.4 made it so that none of the legacy gen pass delegates access any local variables anymore! That means you can do what we call a 'method split'

Break the `WorldGen.GenerateWorld` method into 3 methods, `GenerateWorld`, `AddGenPasses` and `GenerateWorld_RunTasksAndFinish`. 

```cs
		AddGenPasses();
		GenerateWorld_RunTasksAndFinish(seed, customProgressObject, generationStopwatch);
	}

	internal static void AddGenPasses() {
		AddGenerationPass("Reset", delegate (GenerationProgress progress, GameConfiguration passConfig) {
```

later
```cs
	private static void RunTasks(int seed, GenerationProgress customProgressObject, Stopwatch generationStopwatch) {
		SystemLoader.ModifyWorldGenTasks(_generator._passes, ref _generator._totalLoadWeight);
	
	...
```

Then you can call something like:
```cs
_generator = new();
AddGenPasses();
SystemLoader.HookGenPasses(_generator._passes)
```

Honestly, it doesn't look like any of the gen passes are conditionally added, so you could probably just store them all in a list at load, and expose it as `IReadOnlyList<GenPass> VanillaGenPasses` in `WorldGen.TML.cs`

Then your example could just use `WorldGen.EditGenPass("Living Trees", il=> ...`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/WorldGen.TML.cs
Destructor-Ben: This was mainly so I could tell that the hooks were being registered and unregistered properly, but I can move it to https://github.com/tModLoader/tModLoader/issues/3213

patches/tModLoader/Terraria/WorldGen.TML.cs
Destructor-Ben: Would IL editing other mods generation passes be important? I'm guessing not, since it would probably be a smaller group of people wanting it, and I'm guessing most mods would use GenPasses instead of PassLegacys, but in case it is, should all world generation passes get moved to being added at load time (including modded ones)?

Chicken-Bones: IL editing requires intimate knowledge of the target code. Presumably, modded gen passes would have their code in one or more methods, not just a lambda.

Either way, I wouldn't bother considering mods IL editing mods when making tML decisions

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1,5 +1,13 @@
 --- src/TerrariaNetCore/Terraria/WorldGen.cs
 +++ src/tModLoader/Terraria/WorldGen.cs
+@@ -4,6 +_,7 @@
+ using System.Threading;
+ using System.Threading.Tasks;
+ using Microsoft.Xna.Framework;
++using rail;
```
Chicken-Bones: accidental import?

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -7,4 +23,34 @@ public partial class WorldGen
 	private static double _timePass = 0.0; // Used to account for more precise time rates.
 
 	internal static void ClearGenerationPasses() => _generator?._passes.Clear();
+
+	internal static List<GenPass> vanillaGenPasses_internal = new();
```
Chicken-Bones: Preferred name is `_vanillaGenPasses`

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -7,4 +23,34 @@ public partial class WorldGen
...
+	public static ReadOnlyCollection<GenPass> VanillaGenPasses = vanillaGenPasses_internal.AsReadOnly();
```
Chicken-Bones: prefer to return `IReadOnlyCollection`. Sure someone can cast it back to `List` but we're not here stop that.

ExampleMod/Common/Systems/ExampleWorldGenHookingSystem.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+	// This ModSystem will demonstrate how to IL edit and Detour world generation passes
+	// Since world generation passes are anonymous methods (they don't have a name), they can't be edited the standard way (using IL_xx or On_xx)
+	public class ExampleWorldGenHookingSystem : ModSystem
+	{
+		// All of the registration should take place in load
+		// Generation pass hooks are unloaded manually, so no Unload method is needed
+		public override void Load() {
+			// IL editing the pyramids pass
+			GenPass pyramidsPass = WorldGen.VanillaGenPasses.First(t => t.Name == "Pyramids");// Locate the generation pass
```
Chicken-Bones: Need to call `FirstOrDefault` here if you want it to be error free

Chicken-Bones: Honestly, I'd prefer using `Single` and a cast, with no warning. Since no-one should be modifying the vanilla pass list anyway.

Chicken-Bones (changes requested):
Looking good

ExampleMod/Common/Systems/ExampleWorldGenHookingSystem.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+	// This ModSystem will demonstrate how to IL edit and Detour world generation passes
+	// Since world generation passes are anonymous methods (they don't have a name), they can't be edited the standard way (using IL_xx or On_xx)
+	public class ExampleWorldGenHookingSystem : ModSystem
+	{
+		// All of the registration should take place in load
+		// Generation pass hooks are unloaded manually, so no Unload method is needed
+		public override void Load() {
+			// IL editing the pyramids pass
+			GenPass pyramidsPass = WorldGen.VanillaGenPasses.FirstOrDefault(t => t.Name == "Pyramids");// Locate the generation pass
```
Chicken-Bones: Final feedback, on thinking about this more. 
- `VanillaGenPasses` should be an `IDictionary<string, GenPass>`
- Examples should just access the value, expecting it to exist, since nothing can (or should) remove vanilla gen passes.

Chicken-Bones:
I think the only minor change remaining on this is my `try catch` recommendation for the examples. I'm happy to contribute it and finish this if you'd like. 

Also, please send me a ping on the tModLoader Discord, I'd love to chat with you there!

Destructor-Ben:
> I'm happy to contribute it and finish this if you'd like

That would be really nice, thanks

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -1,4 +1,12 @@
+using MonoMod.Cil;
+using MonoMod.RuntimeDetour.HookGen;
+using System.Collections.Generic;
+using System.Reflection;
+using System;
+using Terraria.GameContent.Generation;
 using Terraria.WorldBuilding;
+using Terraria.IO;
+using System.Collections.ObjectModel;
```
Chicken-Bones: Imports cleanup?

PR #3243, opened 2023-02-10
PostSetupTileMerge hook, fixes related to tile merging

direwolf420:
### What is the new feature?
`Mod/GlobalTile.PostSetupTileMerge` hook to go with a restructuring of `Main.SetupTileMerge` to cover modded tiles (fixing a bug with `TileID.Sets.BlockMergesWithMergeAllBlock` not being automatically assigned to modded tiles (**i.e. closed doors, implicit tml support added here**), and the set being reset without a reassignment, causing Smooth Marble Block to not merge with certain vanilla tiles), and allow safer dynamic changes to tile merging.
To implement behavior similar to Echo Block and closed trap doors without overriding the hook, a new `BlockMergesWithMergeAllBlockOverride` set is added, allowing changing this specific behavior all through the commonly used `SetStaticDefaults` before vanilla automatically assigns it to the formerly mentioned set.
For implicit modded closed door support, an overload for `TileLoader.IsClosedDoor` was required.

* Calling the code used in `Main.SetupTileMerge` by tml manually after mods finished loading is required for compatibility with vanilla/other mods.
* Safe timing (not "undefined" in the case of `SetStaticDefaults`) where it's guaranteed that most mods have set tile related properties for all their tiles so other mods can poll for this state and do dynamic checks (similar to what vanilla does with `TileID.Sets.BlockMergesWithMergeAllBlock`) based on that data.

The split done to `Main.SetupTileMerge` could be handled differently, I chose the parameter appoach as it keeps the method names the same. Another design would be to rename the internal one to "VanillaSetupTileMerge", and copy (and or redirect to) the method that uses fixed `TileLoader.TileCount` for the kept `Main.SetupTileMerge` (without parameter).
The main reason for the split is so that no patch should be made on the callsites, and so that `Main.tileMerge` isn't reset again.

Since vanilla doors don't merge with `tileBlendAll` tiles, tml could automatically support it using `TileLoader.IsClosedDoor`. This would not account for trap doors though, which tml has no support for.

Bugfix showcase with Example Block having `Main.tileBlendAll[Type] = true`:
Before the PR:
Screenshot_3

After the PR:
Screenshot_4

Set showcase (disabling wood merging with the previously set "merge all" Example Block):
```cs
public class asd : GlobalTile
{
		TileID.Sets.BlockMergesWithMergeAllBlockOverride[TileID.WoodBlock] = false;
	}
}
```
Screenshot_5

### ExampleMod updates
ExampleDoorClosed is implicitely not merging, just like vanilla doors. No code changes required.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -448,7 +448,7 @@
  	{
  		int count = TileID.Count;
  		tileMerge = new bool[count][];
-@@ -2842,12 +_,23 @@
+@@ -2842,12 +_,29 @@
  			tileMerge[i] = new bool[count];
  		}
  
@@ -459,7 +459,13 @@
 +	{
 -		for (int j = 0; j < TileID.Count; j++) {
 +		for (int j = 0; j < count; j++) {
++			// TML: moved these cases into BlockMergesWithMergeAllBlockOverride
  			bool flag = j == 10 || j == 387 || j == 541;
++			if (TileID.Sets.BlockMergesWithMergeAllBlockOverride[j] is bool mergeOverride) {
++				TileID.Sets.BlockMergesWithMergeAllBlock[j] = mergeOverride;
++				continue;
++			}
++
  			TileID.Sets.BlockMergesWithMergeAllBlock[j] = !flag && tileSolid[j] && !tileSolidTop[j];
  		}
  	}
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -433,7 +433,7 @@
  
  	public static string playerPathName => ActivePlayerFileData.Path;
  
-@@ -2831,6 +_,8 @@
+@@ -2831,10 +_,13 @@
  		}
  
  		result.X = num;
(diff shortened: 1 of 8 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -442,7 +442,31 @@
...
+ 		int count = TileID.Count;
+ 		tileMerge = new bool[count][];
+@@ -2842,12 +_,29 @@
+ 			tileMerge[i] = new bool[count];
+ 		}
+ 
++		SetupTileMerge(count);
++	}
++
++	public static void SetupTileMerge(int count)
```
Chicken-Bones: `internal SetupAllBlockMerge` would be a better name

patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
```diff
@@ -274,4 +275,12 @@ public virtual void FloorVisuals(int type, Player player)
 	public virtual void ChangeWaterfallStyle(int type, ref int style)
 	{
 	}
+
+	/// <summary>
+	/// Allows you to set <see cref="TileID.Sets.BlockMergesWithMergeAllBlock"/> to false if you don't want a tile to merge with other tiles in <see cref="Main.tileBlendAll"/> (Echo Block does this).
+	/// <br/>Can be used to adjust other tile merge related things that are not possible to do in <see cref="ModBlockType.SetStaticDefaults"/> due to timing.
```
Chicken-Bones: Doc edit, second line is fine. First lien is not intended use anymore.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -1009,6 +1017,18 @@ public static void PlaceInWorld(int i, int j, Item item)
 		GetTile(type)?.PlaceInWorld(i, j, item);
 	}
 
+	public static void PostSetupTileMerge()
+	{
+		foreach (var hook in HookPostSetupTileMerge) {
+			hook();
+		}
+
+		for (int i = 0; i < tiles.Count; i++) {
```
Chicken-Bones: `foreach` please

Chicken-Bones (changes requested):

PR #3246, opened 2023-02-10
Fix 1.4.4 specific bugs caused by vanilla ID count restrictions

direwolf420:
### What is the bug?
Many of these fixes are 1.4.4 specific because it added alot of bound checks to array accesses where there haven't been any in 1.4.3.6, for whatever reason (Can be seen by comparing the Sea Oats (ID 529) `TileFrame` code in 1.4.3.6 and 1.4.4(.9)).
@Chicken-Bones could you please elaborate on this odd decision? At first I thought it's something TEdit related (tiles/walls loading with invalid types) but it seems that most of these bound checks are referring to new 1.4.4 content or codebase changes, or aren't tile related at all.
Vanilla also did these bound checks wrong in certain places, mainly in `WorldGen.Convert`: it would let a type of `Count` (for tile and wall respectively) through to the arrays, which would IOOB.

#### List of bugs:
Things explicitely tested and confirmed are checkmarked, the rest is derived from the code itself and not tested.
* wrong type check on modded torch detection (highly unlikely that this ever happens, but assuming a modded torch got the ID of `TileID.Count`, it would not count as one)
* [x] Toxic flask projectile wouldn't count modded platforms as solid, and would stop spreading upon touching them instead of going through
* [x] Dirt Bomb would not properly spread dirt when modded platforms are in the way
* Dead Man's Chest would ignore modded containers when deciding where to generate
* modded boulders would ignore other players standing in the way when trying to place them with smart cursor
* modded foliage would block Pumpkin Seed placement
* modded dirt/mud would not get replaced by "Dirtiest Block" during worldgen, and wouldn't merge with most tiles
* certain ambient objects (Small Piles, Sea Oats) would not properly destroy themselves if anchored to (suitable) modded tiles and TileFrame is called
* custom biome conversions that call `WorldGen.TryKillingTreesAboveIfTheyWouldBecomeInvalid` would not work with a modded tile type specified
* biome conversions back to normal (snow, sand, dirt) would not work for modded tiles despite being listed in the Conversions sets
* [x] Non-solid modded tiles would not cause sand above them to get converted to hardened sand of the specific biome (if for some reason a sand tile ended up resting on a non-solid tile), which would end up collapsing the sand pillar above them
* [x] Modded tiles in `TileID.Sets.CanGrowCrystalShards` would not grow Crystal Shards
* Modded ropes don't interact properly with railtracks
* Abigails Flower and Glow Tulip would only grow on vanilla grasses
* [x] vanilla foliage and trees would not grow infront of modded walls explicitely marked as `WallID.Sets.AllowsPlantsToGrow`
* [x] vanilla UG desert enemies would not spawn infront of modded walls explicitely marked as `WallID.Sets.AllowsUndergroundDesertEnemiesToSpawn`
* modded slimes that can transform into the copper town slime will never transform
* modded critters do not have the 1.5 second immunity when spawned
* [x] smart cursor would not work with modded blocks

**Some changes were omitted (please comment on if these should stay as is or get changed):**
* [x] Fertilizer currently doesn't grow modded trees, its projectile checks for `TileID.Count` before accessing `TileID.Sets.CommonSapling`, then calls `WorldGen.AttemptToGrowTreeFromSapling`, the latter has no support for modded to my knowledge, so I left it as is.
* [x] Pumpkins only grow on specific, hardcoded grasses (normal and hallow + mowed variants). If they end up on modded tiles they don't break automatically via TileFrame. For some reason there is an upper bounds check even though there are no arrays being accessed?
* `ProjectileID.Sets.StormTiger` has a count check in one place. I didn't change this as setting modded NPCs to this set would not make sense (Even though `ProjectileID.Sets.StardustDragon` is used in a similar way but does not have any checks)
* The code that enables info accessories from the Void Bag is guarded by a count check, but that code has no mod compatibility at all, so  I left it as is.

If there is no real reason why these boundary checks were added in the first place, they can simply be omitted (as done already in many places). I'll await CBs reply.

### DIFF
--- a/patches/tModLoader/Terraria/Collision.cs.patch
+++ b/patches/tModLoader/Terraria/Collision.cs.patch
@@ -1,5 +1,22 @@
 --- src/TerrariaNetCore/Terraria/Collision.cs
 +++ src/tModLoader/Terraria/Collision.cs
+@@ -3,6 +_,7 @@
+ using Microsoft.Xna.Framework;
+ using Terraria.DataStructures;
+ using Terraria.ID;
++using Terraria.ModLoader;
+ 
+ namespace Terraria;
+ 
+@@ -1030,7 +_,7 @@
+ 		if (tile == null || !tile.active() || tile.inActive() || !Main.tileSolid[tile.type])
+ 			return false;
+ 
+-		if (treatPlatformsAsNonSolid && tile.type > 0 && tile.type <= TileID.Count && (TileID.Sets.Platforms[tile.type] || tile.type == 380))
++		if (treatPlatformsAsNonSolid && tile.type > 0 && tile.type <= TileLoader.TileCount && (TileID.Sets.Platforms[tile.type] || tile.type == 380))
+ 			return false;
+ 
+ 		int num = tile.blockType();
 @@ -2614,7 +_,10 @@
  			tile = Main.tile[num2, num3];
  			tile2 = Main.tile[num2, num3 - 1];
--- a/patches/tModLoader/Terraria/Collision.cs.patch
+++ b/patches/tModLoader/Terraria/Collision.cs.patch
@@ -13,7 +13,7 @@
  			return false;
  
 -		if (treatPlatformsAsNonSolid && tile.type > 0 && tile.type <= TileID.Count && (TileID.Sets.Platforms[tile.type] || tile.type == 380))
-+		if (treatPlatformsAsNonSolid && tile.type > 0 && tile.type <= TileLoader.TileCount && (TileID.Sets.Platforms[tile.type] || tile.type == 380))
++		if (treatPlatformsAsNonSolid && tile.type > 0 && (TileID.Sets.Platforms[tile.type] || tile.type == 380))
  			return false;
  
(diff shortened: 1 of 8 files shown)


Chicken-Bones:
Sorry for taking so long to review this. 

There is no reason for these checks so they can be removed. I have no problem with the game crashing if a `Tile/Projectile/NPC/Item` somehow gets assigned an out of range `type`. A world with out of range tile types would fail to load via exception anyway, there are many places we access the arrays without bounds checks.

We don't have an internal policy against them, so some developers add them unnecessarily. Sorry for the problems this has caused. The only case where it makes sense to check is if the data is coming from an untrusted source (such as chat) or a function where `-1` may be a valid return value

direwolf420:
Removed the Count checks. I left them in the Worldgen.cs file where biome conversions are occuring, as the vanilla checks there were not using the correct comparison, ignoring whatever tile/wall would end up with the `Count` type
There are also places in vanilla where lower bound checks don't make sense, usually related to tiles or walls because their type can never be below 0. I didn't remove these since it doesn't affect anything.

PR #3248, opened 2023-02-11
GrappleCanLatchOnTo hook

direwolf420:
### What is the new feature?
New `Mod/GlobalProjectile.GrappleCanLatchOnTo` hook that takes in a tile position and lets the modder customize when and what the grappling hook should latch (without manual AI modifications).

This makes it possible to implement custom grappling conditions akin to Squirrel Hook.

I chose to change the signature of the vanilla method to use tile coordinates instead of the tile instance. This allows for a wider range of use cases.
I also chose to move the `Tile.HasUnactuatedTile` condition into the vanilla method that handles the latching conditions, to let modders be able to make their hooks latch onto actuated tiles or even air. This should not mess with modders adding their own tiles to it, but certain checks like `Tile.HasTile` may be needed for more advanced usages. This is documented in the docs and in the ExampleMod example.

* See ExampleMod
* Hooks that latch onto other entities or arbitrary locations alltogether (i.e. a force field)

### ExampleMod updates
Made Example Hook latch onto trees like Squirrel Hook does.

### DIFF
--- a/patches/tModLoader/Terraria/Projectile.cs.patch
+++ b/patches/tModLoader/Terraria/Projectile.cs.patch
@@ -1305,7 +1305,7 @@
  				flag = false;
  
  			if (flag) {
-@@ -36476,7 +_,24 @@
+@@ -36476,7 +_,23 @@
  		}
  	}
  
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1285,6 +1295,41 @@
...
++		// TML: Vanilla check modified by moving Tile.nactive() from its callsites into here to allow actuated or air to be grappled by mods
++		vanilla &= theTile.nactive();
++
++		bool? modded = ProjectileLoader.GrappleCanLatchOnTo(this, Main.player[owner], x, y);
++
++		if (modded.HasValue)
++			return modded.Value;
++
++		return vanilla;
++	}
```
Chicken-Bones, line 1328: This patch is less maintainable than it could be. I think since there's only 2 call-sites. It's better to put the `ProjectileLoader.GrappleCanLatchOnTo` check at the call sites.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1285,6 +1295,41 @@
...
  				Kill();
  
+@@ -36463,7 +_,8 @@
+ 				Main.tile[point4.X, point4.Y] = new Tile();
+ 
+ 			bool flag = true;
+-			if (Main.tile[point4.X, point4.Y].nactive() && AI_007_GrapplingHooks_CanTileBeLatchedOnTo(Main.tile[point4.X, point4.Y]))
++			// TML: Moved nactive check into method, changed from tile to coords
++			if (AI_007_GrapplingHooks_CanTileBeLatchedOnTo(point4.X, point4.Y))
+ 				flag = false;
```
Chicken-Bones, line 1305: (See next comment) Here, you can do `flag = ProjectileLoader.GrappleCanLatchOnTo(this, point4.X, point4.Y) ?? flag`

patches/tModLoader/Terraria/Projectile.cs.patch
Chicken-Bones: Also please use `is bool modGrappleOverride` style pattern matching instead of `HasValue` and `Value`

Chicken-Bones (changes requested):
Hook looks good, just notes on patching

Chicken-Bones:
Note for 1.4.5, hopefully the relevant code paths will be simplified.

PR #3251, opened 2023-02-13
[1.4.4] A Little* Bit of Documentation

Antirhinnum:
### What is the new feature?
Documentation for most ID sets and dependent structs (`WingStats`, for example); almost all of `Chest`, `FishingAttempt`, `PopupText`, `ItemVariant`, and `ContentSamples`; and some scattered fields across the project, such as `Main.worldSurface`.
Also addresses some internal comments about some TML code changes (`AmmoID.IsArrow` and similar; `Player.ItemCheck_PayMana()`).
A spiritual successor to #2670.

Most of the documented symbols either have many effects that aren't obvious from their names or have no obvious connection to what fields they work in tandem with. There are also many fields missing units (world/tile coordinates) that aren't immediately obvious.

Wiki pages would be the most obvious alternate choice, as they would reduce patch size and allow for both much more in-depth descriptions of fields' purposes and meaningful examples.

N/A

### ExampleMod updates
N/A


Rijam:
Sort of related, I created a gist a bit ago that has explanations for most of the `ItemID.Sets` https://gist.github.com/Rijam/84206c1bf1094933ff79231315ce7b71
Not saying that you need to change any of yours, I'm just letting you know that the gist I made is like a wiki page and has some examples as you said in the alternate designs section.

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -138,16 +368,104 @@
  		SoundEngine.PlaySound(22, X * 16, Y * 16);
  		for (int i = X; i <= X + 1; i++) {
  			for (int j = Y; j <= Y + 1; j++) {
- 				Tile tileSafely2 = Framing.GetTileSafely(i, j);
- 				if (tileSafely2.type == type) {
- 					tileSafely2.frameX += num;
-+					/*
- 					Main.tile[i, j] = tileSafely2;
-+					*/
- 				}
```
JavidPack, line 147: It looks like a recent fix was reverted here

Antirhinnum: Fixed in aa37d0e37094e124ba4ea6c00609ffd6a0c0193d.

patches/tModLoader/Terraria/DataStructures/FishingAttempt.cs.patch
```diff
@@ -0,0 +1,125 @@
...
+ 	public bool CanFishInLava;
++
++	/// <summary>
++	/// How high in the sky this attempt takes place, in the range [<c>0.25f</c>, <c>1f</c>]. Any value below <c>1f</c> takes place approximately in the top 10% of the world.
++	/// <br/> The lower this value, the smaller <see cref="waterNeededToFish"/> will be, which is automatically applied.
++	/// </summary>
+ 	public float atmo;
++
++	/// <summary>
++	/// The <see cref="ID.ItemID"/> of the quest fish the Angler wants, or <c>-1</c> if this player can't catch that fish today.
```
JavidPack: This is the main thing that stands out to me as an issue. You refer to item types as ItemIDs, but I think what we want to teach is the opposite. Change these sort of things to "item type" instead of pointing to the vanilla ID classes. (and the same for any other type that you refer to as TypeID)

Modders will know from the getting started guide how "X type" refers to "XID.cs" for vanilla type values and "ModContent.XType<>()" for modded type value.

JavidPack:
Sorry, bit busy lately. See that note I made, I think that is the approach we want to take for referring to various content types rather than linking to the vanilla typeID classes. 

After changing to that, I'll give it another read.

patches/tModLoader/Terraria/DataStructures/FishingAttempt.cs.patch
Antirhinnum: Before I change all of these, would the plain text "<x> type" or link to the associated `type` field be preferred?
Ex: `item type` vs. `<see cref="Terraria.Item.type"/>`.

I saw the discussion in #collaborators a week ago about doc. standardization (https://discord.com/channels/103110554649894912/445276626352209920/1074844954381258813), but it didn't seem to reach a consensus.

JavidPack: That's true, there wasn't much feedback from my question there. I think referencing Item.type is probably good, it is the primary place item types will exist. I'll need to reword some existing docs later to match our consistent terminology.

Lets go with `The item type (<see cref="Terraria.Item.type"/>) of the...` for now. I think that communicates that "item type" is the idea and Item.type is where it typically is used. The Item.type docs have more info on modded vs vanilla, so a modder should be able to click through if they forgot completely.

Antirhinnum: Most ID references have been changed now, but I've come onto a problem with `BuffID` and `ArmorIDs`.
`ArmorIDs` has two possible vanilla fields it could point to: `Item.headSlot` or `Player.head`. I think it'd be best to use the `Item` fields (since `ModItem`s are where most modders will be using any `ArmorIDs` sets) and would like to confirm with you that that fits the established pattern.
`BuffID` doesn't seem to have any good analogue to `Item.type`. `Player.buffType` and `NPC.buffType` come to mind, but it seems clunky using them -- it'd be like referring to item types using the player's inventory.

JavidPack:
Amazing, thanks so much.

PR #3252, opened 2023-02-13
IL Editing QoL Stuff

Destructor-Ben:
https://github.com/tModLoader/tModLoader/issues/3213

Added 3 methods and 1 exception
`DumpILHooks()` will dump the list of currently registered IL hooks to the console (useful for manual hook registering with reflection)
`DumpOnHooks()` will do the same but for detours
`DumpIL(Mod mod, ILContext il)` will dump the `ILContext` to` Logs/ILDumps/{Mod Name}/{Method Name}.txt`
`ILPatchFailureException(Mod mod, ILContext il, Exception innerException)` is a convenience exception which will call `DumpIL` and wrap the inner exception with a better message

### Sample Usage for the New Feature
```c#
try {
    ILCursor c = new ILCursor(il);
    c.GotoNext(i => i.MatchLdcI4(566));
    // IL editing stuff...
});
}
catch (Exception e) {
    throw new ILPatchFailureException(ModContent.GetInstance<ExampleMod>(), il, e);
    // or
    MonoModHooks.DumpIL(ModContent.GetInstance<ExampleMod>(), il);
}
```

#### Note
Some fixes were made in followup commit 179db727e9734ad151761fe9c3060fc5c4d382ac

### DIFF
--- a/ExampleMod/Content/Items/Accessories/WaspNest.cs
+++ b/ExampleMod/Content/Items/Accessories/WaspNest.cs
@@ -17,28 +17,30 @@ namespace ExampleMod.Content.Items.Accessories
 
 		// This IL editing (Intermediate Language editing) example is walked through in the guide: https://github.com/tModLoader/tModLoader/wiki/Expert-IL-Editing#example---hive-pack-upgrade
 		private static void HookBeeType(ILContext il) {
-			ILCursor c = new ILCursor(il);
+			try {
+				ILCursor c = new ILCursor(il);
 
-			// Try to find where 566 is placed onto the stack
-			if (!c.TryGotoNext(i => i.MatchLdcI4(566))) {
-				// Patch unable to be applied, so we let people know with this method and exit the method
-				MonoModHooks.LogILPatchFailure(ModContent.GetInstance<ExampleMod>(), il, "Unable to locate 566 on the stack");
-				return;
+				// Try to find where 566 is placed onto the stack
+				c.GotoNext(i => i.MatchLdcI4(566));
+
+				// Move the cursor after 566 and onto the ret op.
+				c.Index++;
+				// Push the Player instance onto the stack
+				c.Emit(OpCodes.Ldarg_0);
+				// Call a delegate using the int and Player from the stack.
+				c.EmitDelegate<Func<int, Player, int>>((returnValue, player) => {
+					// Regular c# code
+					if (player.GetModPlayer<WaspNestPlayer>().strongBeesUpgrade && Main.rand.NextBool(10) && Main.ProjectileUpdateLoopIndex == -1) {
+						return ProjectileID.Beenade;
+					}
+
+					return returnValue;
+				});
 			}
-
-			// Move the cursor after 566 and onto the ret op.
-			c.Index++;
-			// Push the Player instance onto the stack
-			c.Emit(OpCodes.Ldarg_0);
-			// Call a delegate using the int and Player from the stack.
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
```diff
@@ -113,4 +115,57 @@ internal static void RemoveAll(Mod mod)
...
+	/// A helper method that logs to the console that an IL patch failed.
+	/// </summary>
+	/// <param name="mod"></param>
+	/// <param name="il"></param>
+	/// <param name="reason"></param>
+	public static void LogILPatchFailure(Mod mod, ILContext il, string reason)
+	{
+		// TODO: use StringRep function but for MethodDefinition instead of MethodBase
+		// TODO: should this also include a patching class where the hooks are made and a feature name?
+		mod.Logger.Warn($"Failed to IL edit method \"{il.Method.Name}\", because \"{reason}\"");
```
Chicken-Bones: Could dump the entire `il.ToString` to a text file somewhere (`Logs/ILDumps`) while you're at it maybe?

ExampleMod/Content/Items/Accessories/WaspNest.cs
```diff
@@ -21,7 +21,9 @@ public class WaspNest : ModItem
 
 			// Try to find where 566 is placed onto the stack
 			if (!c.TryGotoNext(i => i.MatchLdcI4(566))) {
-				return; // Patch unable to be applied
+				// Patch unable to be applied, so we let people know with this method and exit the method
+				MonoModHooks.LogILPatchFailure(ModContent.GetInstance<ExampleMod>(), il, "Unable to locate 566 on the stack");
```
Chicken-Bones: Hm... I think that having to make error messages for every try-get which can fail will annoy some people, and that a better approach would be to wrap the entire method in a `try catch` and do a `throw new Exception("msg")`

Exceptions have the benefit of giving exact line numbers too

Chicken-Bones: Can call `LogILPatchFailure` in the `catch` handler

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
Destructor-Ben: Should I also add public methods for dumping IL and diffing it, or will that be done in MonoMod?

Chicken-Bones: dumping sure. Diffing is best left to the user's preferred tools

ExampleMod/Content/Items/Accessories/WaspNest.cs
```diff
@@ -17,26 +17,30 @@ public class WaspNest : ModItem
...
 
-				return returnValue;
-			});
+					return returnValue;
+				});
+			}
+			catch (Exception e) {
+				// If there are any failures with the IL editing, this custom exception will be thrown that dumps the IL to Logs/ILDumps/{Mod Name}/{Method Name}.txt
+				throw new ILPatchFailureException(ModContent.GetInstance<ExampleMod>(), il, e);
+			}
```
Chicken-Bones, line 43: It appears I may have been too hasty. Do we want to throw here? Or just log and dump?
The exception doesn't actually log.

PR #3257, opened 2023-02-15
Fix pet drawing in player select screen for invalid items + ExampleMod showcase

direwolf420:
### What is the bug?
Unloaded Item in the vanity pet slot causes the game to display the player as if he has an invisible pet in the player select screen. Same applies to items that end up in that slot even if they aren't valid anymore (i.e. if a modder changes a pet item to be something else)

### How did you fix the bug?
Copy the conditions from `ItemSlot` that restrict only vanity pet items going into that slot.

Early return instead of changing the vanilla line, but the IsAir check would have to be duplicated probably?

### ExampleMod updates
Added support for ExampleMod vanity pets for this feature, as by default they don't animate and usually display offset from the player/clip into the player below. The examples cover a basic one (paper airplane), and an advanced one (minion pet).


Chicken-Bones:
Looks great, thanks for this

PR #3258, opened 2023-02-15
InfoDisplay text coloring support

direwolf420:
### What is the new feature?
Changes `InfoDisplay.DisplayValue` signature to enable coloring the text.
Adds `GlobalInfoDisplay.ModifyDisplayColor` to change or override the coloring of the text.

Adds support for the 1.4.4 added feature which goes as follows:
"_Some info accessory tabs are grayed out when inactive or display N/A, and will light back up when active again. [Gold Critters](https://terraria.wiki.gg/wiki/Gold_Critters) will appear in gold on the [Lifeform Analyzer](https://terraria.wiki.gg/wiki/Lifeform_Analyzer)._"

Probably? The current way of giving both the text color and inactive color to the modder allows for more freedom for choosing the color (i.e. Lerp instead of assign when inactive, or dynamic color in general). An argument could be made to take this option away from the modder by setting a flag which then lets TML change the color instead if inactive. However, that would take away the customizability.

Lifeform Analyzer-like coloring.

### ExampleMod updates
ExampleMod was updated to utilize this for its minion display.
Screenshot_6

### Porting notes
(Handled by tModPorter)
`InfoDisplay.DisplayValue()` -> `InfoDisplay.DisplayValue(ref Color displayColor, Color inactiveColor)`


Chicken-Bones:
I don't like passing `inactiveColor` as a parameter, it'd be better as a static property. 
Maybe it could be a helper in `InfoDisplay.InactiveInfoTextColor`?

Otherwise looks great.

PR #3271, opened 2023-02-26
Example Drill

Blockaroz:

Adds the Example Drill, which introduces the concept of held projectiles and the drill variants for pickaxes

No, but comments may need tweaking


direwolf420:
The projectile AI does not look like you adapted it from aiStyle 20. Most of it is not accurate (and your current code does not work in multiplayer).
You should mention in a comment what impact setting `ItemID.Sets.IsDrill` has on modded items (namely nothing, contrary to what its purpose is of automatically adjusting useTime, useAnimation, and tileBoost, making the source code values in SetDefaults not entirely accurate if the modder would want to copy them from vanilla for balancing reasons)
Maybe also showcase `Item.tileBoost`.

Blockaroz:
Made some changes that hopefully reflect what you asked for

direwolf420:
Much better, just refactor some obvious vanilla stuff like player.inventory[player.selectedItem] into player.HeldItem, and the whole MouseX/Y stuff with gravity can be compacted to just Main.MouseWorld which takes gravity into account. Give the num123 variables proper names too.

and single line if/else statements like these
```cs
if (something)
    action;
```
should be changed to
```cs
if (something) {
    action;
}
```
for consistency with other EM stuff

direwolf420:
Missed the `Projectile.netUpdate = true;` single line, and lastly, all comments should have a space `// like this`

Blockaroz:
should be all good now?

edit: missed one

edit 2: accidentally left in another line, please excuse my sloppiness

direwolf420 (approved):

JavidPack:
Did you test this in multiplayer? I noticed `if (Main.myPlayer == Projectile.owner) {` was removed from the vanilla code, but from the looks of it it seems pretty important.

(I'm fixing things at the moment, no need to update the PR, just curious if you tested multiplayer or not)

PR #3273, opened 2023-02-26
Fix ProjectileLoader.ReceiveExtraAI not being called (consistently)

direwolf420:
### What is the bug?
1.4.4 introduced `Projectile.ai[2]`. Netcode wise, this now overlaps with the previous implementation of `XExtraAI`, causing ReceiveExtraAI to only be called for projectiles that have `ai[2] != 0f`.

### How did you fix the bug?
On the **sender** side:
* `bitsByte23[2]` was originally unused before 1.4.4 so TML used that for flagging if `extraAI` exists. Now, this flag is used for continuation of sending more flags (and by extension, data attached to those flags). Vanilla sets `bitsByte23[2]` to true if `bitsByte24` contains anything (vanilla populates `bitsByte24[0]` with `ai[2] != 0f`)
* TML now utilizes the mostly free `bitsByte24`, and flags `bitsByte24[1]` with `hasExtraAI`. Vanilla then takes care of setting `bitsByte23[2]` as required.

On the **receiver** side:
* TML accesses `bitsByte3[1]` to read the sent `hasExtraAI` (which corresponds to `bitsByte24[1]`)

This way, it adds no additional overhead while not modifying vanilla netcode order.

This seems like the cleanest solution to me. TML could optionally start with the flags from 7 instead of the first free one, this avoids compatibility issues in the long run until vanilla maxes out the BitsByte.

### DIFF
--- a/patches/tModLoader/Terraria/MessageBuffer.cs.patch
+++ b/patches/tModLoader/Terraria/MessageBuffer.cs.patch
@@ -262,7 +262,7 @@
  					num37 = -1;
  
  				array[2] = (bitsByte3[0] ? reader.ReadSingle() : 0f);
-+				byte[] extraAI = bitsByte3[2] ? ProjectileLoader.ReadExtraAI(reader) : null;
++				byte[] extraAI = bitsByte3[1] ? ProjectileLoader.ReadExtraAI(reader) : null;
 +
  				if (Main.netMode == 2) {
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Thanks, looks like this was partially implemented and untested during porting, and I made a couple mistakes.

Issue #3276, opened 2023-02-26
[1.4.4]Add “NearShimmer” Recipe Condition just like near other three liquids


gugudao:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

1.4.4 Update brings a fourth new liquid: Shimmer; Adding this liquid as recipe condition obviously helps to broaden the mods  easily create recipe related to shimmer

### What does this proposal attempt to solve or improve?

![@_LPH(F797)H4%8}06G JBS](https://user-images.githubusercontent.com/79732257/221423679-c3bbd6e4-b571-4525-a0f6-98b56482ee88.png)

PR #3277, opened 2023-02-27
Add `NearShimmer` recipe condition

Setnour6:
### What is the new feature?

See issue #3276

This addition allows the use of shimmer for crafting (or "decrafting") without actually using shimmer, particularly, for example, with mods that use the same item to decraft into other items, which may cause conflicts; this allows mods to add "craftable decrafting" recipes so that all players need to do is "decraft" without throwing anything in (in tandem with the example). Also see issue #3276

No, but continuing the description, there could be an automatic function done to make it so when two mods change a single item for decrafting, the recipe for the item "decrafting" can appear instead.

In example recipes:
```cs
 .AddCondition(Recipe.Condition.NearShimmer)
```

### ExampleMod updates

_n/a_

### Additional notes

I think the rest of the liquid-related things in the code need to also utilize shimmer. `NeedLava` can be used as a search reference.
Can someone please help provide translations for "needs shimmer" or simply "shimmer"?

### DIFF
--- a/patches/tModLoader/Terraria/Recipe.cs.patch
+++ b/patches/tModLoader/Terraria/Recipe.cs.patch
@@ -729,7 +729,7 @@
  	{
  		for (int i = 0; i < tileIDs.Length; i++) {
  			requiredTile[i] = tileIDs[i];
-@@ -15821,12 +_,77 @@
+@@ -15821,12 +_,78 @@
  
  	private static void AddRecipe()
  	{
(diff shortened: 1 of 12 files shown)


patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -49,6 +49,7 @@
 +	internal bool needWater;
 -	public bool needLava;
 +	internal bool needLava;
++	internal bool needShimmer;
```
Chicken-Bones: There is no need to make any changes to `Recipe.cs` here. The `needX` family of conditions exist solely to reduce the number of vanilla patches, and will always be `false` after vanilla recipes have loaded (they're replaced with `Condition` entries).

Please revert all changes to this file.

Chicken-Bones (changes requested):
Overall I think this is helpful and I have no problems with it. See the comment about `Recipe.cs` patches.

patches/tModLoader/Terraria/Recipe.cs.patch
Setnour6: I just undid the `recipe.cs` changes. This should be okay now

PR #3290, opened 2023-03-07
Modded Loadout IO

MutantWafflez:
Adds in proper loading/saving of 1.4.4's new Equipment Loadouts with modded data. 

Not much else to say in this description, oops.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
@@ -353,7 +353,7 @@ internal static class PlayerIO
 	{
 		TagCompound loadouts = new();
 
-		for (int i = 0; i < 3; i++) {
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
```diff
@@ -347,6 +349,31 @@ internal static string SaveUsedModPack(Player player)
...
+
+		return loadouts;
+	}
+
+	internal static void LoadLoadouts(EquipmentLoadout[] loadouts, TagCompound fullSavedCompound)
+	{
+		if (!fullSavedCompound.TryGet("loadouts", out TagCompound loadoutTag)) {
+			// This indicates that this player was created before 1.4.4, and thus has no loadouts; so nothing to load here.
+			return;
+		}
```
Chicken-Bones: This actually isn't necessary, you can just use `tag.GetCompound("loadouts")` in the caller

PR #3292, opened 2023-03-07
New 1.4.4 Info Accessory Hooks

MutantWafflez:
### What is the new feature?

Added 3 new hooks: `ModPlayer.ResetInfoAccessories()`, `ModPlayer.RefreshInfoAccessoriesFromTeamPlayers()` and `Mod/GlobalItem.UpdateInfoAccessory()`.

Takes advantage of the new 1.4.4 feature/fix that keeps info accessories updating, even when the game is paused. Furthermore, allows modders to let players of the same teams share info accessory if they are close enough; a feature that vanilla has for vanilla info accessories, but not yet a feature for modded ones.

Probably, but this was the design as advised by Chicken Bones.

Check ExampleMod's `ExampleInfoDisplay`, `ExampleInfoDisplayPlayer`, and `ExampleInfoAccessory`.


JavidPack:
Would it be useful to mention ItemID.Sets.WorksInVoidBag, commented in the EM example? I imagine someone might want to disable that for some specific item

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1350,36 +1350,80 @@
...
++	}
++
++	// Added by tML.
++	public void RefreshInfoAccsFromItemType(Item item)
++	{
++		RefreshInfoAccsFromItemType(item.type);
++		ItemLoader.UpdateInfoAccessory(item, this);
+ 	}
+ 
+ 	public void RefreshInfoAccsFromItemType(int accType)
```
Chicken-Bones: make `private` please, since it doesn't call the hook. Alternatively, no need to actually split the method. Just declare `int accType`

Chicken-Bones (changes requested):

Issue #3283, opened 2023-03-01
Add ItemID.Sets.SlimeDrops


BasicallyIAmFox:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

`ItemID.Sets.SlimeDrops` is a `Dictionary<int, (int minStack, int maxStack)>` which allows modders to add or change the ores that can appear and drop from slimes.

### What does this proposal attempt to solve or improve?

This new feature would allow easier implementation of alternative ores.

Chicken-Bones:
This seems fine to go ahead with, in the same vein as geodes

PR #3295, opened 2023-03-07
[1.4.4] Add ItemID.Sets.OreDropsFromSlime

BasicallyIAmFox:
Implements #3283 

### What is the new feature?
`ItemID.Sets.OreDropsFromSlime` is a `Dictionary<int, (int minStack, int maxStack)>` which allows modders to add or change the ores that can appear and drop from slimes.

In vanilla, slimes can generate ores inside their bodies with equal drop chance and with fixed range stack size. Modifying or adding to this behavior requires IL editing and/or detouring.
The way `ItemID.Sets.OreDropsFromSlime` is implemented lets the modder add their own loot (usually pre-boss ores) combined with a custom stack size.

- Some way to specify "weight" of each item in the pool (i.e. a mod wants to make Platinum Ore rarer without changing the stack size).
- The `(3, 13)` could be a `static readonly` in `Item` so that if vanilla/tml change it in a later update, and mods use it for their own items, they will get the updated stack range.

```
ItemID.Sets.OreDropsFromSlime.Remove(ItemID.SilverOre);
ItemID.Sets.OreDropsFromSlime[Type] = (3, 13); //Inside a ModItem class
ItemID.Sets.OreDropsFromSlime[ModContent.ItemType<ExampleOre>()] = (3, 13);
ItemID.Sets.OreDropsFromSlime[ItemID.SilverOre] = (12, 34);
```

### ExampleMod updates
The only change in ExampleMod is making ExampleOre to be able to drop from slimes.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
@@ -18,6 +18,24 @@
  		ItemDropAttemptResult result = default(ItemDropAttemptResult);
  		result.State = ItemDropAttemptResultState.Success;
  		return result;
+@@ -47,7 +_,7 @@
+ 				amountDroppedMinimum = 20;
+ 				amountDroppedMaximum = 45;
+ 				break;
+-			case 11:
++			/*case 11:
+ 			case 12:
+ 			case 13:
+ 			case 14:
+@@ -57,7 +_,7 @@
+ 			case 702:
+ 				amountDroppedMinimum = 3;
+ 				amountDroppedMaximum = 13;
+-				break;
++				break;*/
+ 			case 71:
+ 				amountDroppedMinimum = 50;
+ 				amountDroppedMaximum = 99;
 @@ -75,6 +_,10 @@
  				amountDroppedMinimum = 2;
  				amountDroppedMaximum = 5;
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
@@ -1,5 +1,14 @@
 --- src/TerrariaNetCore/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs
 +++ src/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs
+@@ -15,7 +_,7 @@
+ 	public bool CanDrop(DropAttemptInfo info)
+ 	{
+ 		if (info.npc.type == 1 && info.npc.ai[1] > 0f)
+-			return info.npc.ai[1] < (float)ItemID.Count;
++			return info.npc.ai[1] < (float)Terraria.ModLoader.ItemLoader.ItemCount;
+ 
+ 		return false;
+ 	}
 @@ -24,7 +_,7 @@
  	{
  		int itemId = (int)info.npc.ai[1];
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
```diff
@@ -9,3 +18,14 @@
  		ItemDropAttemptResult result = default(ItemDropAttemptResult);
  		result.State = ItemDropAttemptResultState.Success;
  		return result;
+@@ -75,6 +_,10 @@
+ 				amountDroppedMinimum = 2;
+ 				amountDroppedMaximum = 5;
+ 				break;
++		}
```
Chicken-Bones, line 41: vanilla code above should be commented out

patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -159,6 +160,18 @@ internal static void ValidateGeodeDropsSet()
 		}
 	}
 
+	internal static void ValidateSlimeOreDropsSet()
```
Chicken-Bones: May as well make these two into a common method. Like `ValidateDropsSet()`

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -1172,6 +1172,14 @@
  								bool flag26 = WorldGen.OpenDoor(num194, num195 - 1, direction);
  								if (!flag26) {
  									ai[3] = num56;
+@@ -45741,6 +_,7 @@
+ 						return 58;
+ 				}
+ 			case 2:
++				return Main.rand.Next(System.Linq.Enumerable.ToList(ItemID.Sets.OreDropsFromSlime.Keys)); // TML
+ 				if (Main.rand.Next(2) == 0)
```
Chicken-Bones, line 1181: comment out vanilla code which can't run due to `return`

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -1399,6 +1407,17 @@
  			if (Main.netMode == 2)
  				NetMessage.SendData(7);
  		}
+@@ -59957,8 +_,8 @@
+ 		else if (item == 965) {
+ 			num = Main.rand.Next(20, 46);
+ 		}
+-		else if ((item >= 11 && item <= 14) || (item >= 699 && item <= 702)) {
+-			num = Main.rand.Next(3, 9);
++		else if (ItemID.Sets.OreDropsFromSlime.TryGetValue(item, out var minMaxStack) || (item >= 11 && item <= 14) || (item >= 699 && item <= 702)) { // ore drops check from TML
```
Chicken-Bones: Leave the vanilla code commented out, and just have the tML code which replaces it

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ID/ItemID.TML.cs
```diff
@@ -74,5 +74,20 @@ partial class Sets
 			{ Diamond, (3, 7) },
 			{ Amber, (3, 7) }
 		};
+
+		/// <summary>
+		/// Dictionary for defining what ores can spawn as bonus drop inside slime body. All items in this dictionary are equally likely to roll, and will drop with a stack size between minStack and maxStack (exclusive).
```
JavidPack, line 79: This isn't exclusive, I don't think

BasicallyIAmFox: > This isn't exclusive, I don't think

it is exclusive. vanilla code in slime ores drop rule is like that:
```
minStack = minValue;
maxStack = maxValue;
stack = Main.rand.Next(minStack, maxStack + 1);
```

JavidPack: Yes, I know. So since 1 is added, a drop of 13 is possible, so the upper bound is inclusive, right? The wiki also says that they drop 3 to 13. not 3 to 12

BasicallyIAmFox: Wiki is doesn't always has correct or up to date info. But yeah, it could be me putting wrong number. Not entirely sure. Right now it should drop in amount of 3-13

PR #3296, opened 2023-03-07
1.4.4 dedserv mod config

Ishigh1:
### What is the new feature?
The ability to change the mod config while running a dedicated server

For now, the only way to change the config is to have set it with a client, which requires importing the file if both the dedicated server doesn't have a graphic interface

There is the possibility to identify the "host" of a server and copy the configuration in-game automatically, but this is convoluted and requires to create dummy worlds if the config the user wants to change affects the worldgen

Once the server runs, type "m" to access the mod list as usual, then "c <mod number>" to access its config, then the user is in a menu where configs can be edited with "c <config number> <new value>" and reset to the default value with 
"d"


Ishigh1:
Here is how it looks for now (by taking only 25 configs and allowing the clientside configs for the example) from ExampleMod.
Most of them are not supported, but I think that's not that important since Examplemod is full of weird config unlike most of regular mods.
image

JavidPack:
Ok, I'm done with my changes. It now works for all Types and handles more user error.

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -426,29 +427,27 @@ internal static void ServerModMenu(out bool reloadMods)
...
+							WriteColoredLine(ConsoleColor.Yellow, Language.GetTextValue("tModLoader.DedErrorNotEnabled"));
 						}
 					}
 					else {
-						Console.ForegroundColor = ConsoleColor.Yellow;
-						Console.WriteLine(Language.GetTextValue("tModLoader.DedErrorNotEnabled"));
-						Console.ResetColor();
+						WriteColoredLine(ConsoleColor.Yellow, Language.GetTextValue("tModLoader.DedErrorModOOB"));
 					}
 				}
```
Chicken-Bones: This style of code needs to be flattened. Error cases should come first, and fail fast. You may be able to achieve this with an `else if` chain and pattern matching. (Eg `Convert.ToInt32(match.Groups[1].Value) - 1 is int modIndex && ...`), otherwise, extract it to a separate method and use early `return`

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -482,115 +481,142 @@ internal static void ConfigureMod(Mod mod, List<ModConfig> configs)
...
-					if (variable.Type.IsAssignableTo(typeof(IConvertible)))
-						properties.Add(index++, (variable, config));
-					else
-						properties.Add(index++, (variable, null)); // Not convertible aren't supported, this is remembered by a null config
+					properties.Add(index++, (variable, config));
 				}
 			}
 		}
 
 		while (true) {
```
Chicken-Bones, line 488: Please factor the contents of this loop into multiple methods

direwolf420:
Feedback:
Pros:
* get-only properties get displayed
* enum fields show their enum name alongside the raw numeric value

Cons:
* no headers. Modders use them to explain concepts or split the config into sections which makes things more readable
* chat tags aren't supported, not sure what to do here (colors, item icons, etc). I.e. it will directly show [i:Heart] or [i:1234] if used like that
* there is no validation of the saved values. This is just like when manually editing the json file. I.e. it's possible to write SomeEnumValue : 2 when the enum only has underlying values of 0 and 1. It would be up to the modder to correct that in their code (using OnDeserialized)
* **can't edit configs after the world is selected**. Normally servers just have a serverconfig and suitable parameters which auto-launches so you never get to see the screen with the "server menu", I bet it would be difficult to make it work depending on the hosting site you use (I have no experience with them myself). I think in this case it's actually easier for the host person to edit the jsons than to temporarily turn off the serverconfig, edit configs using the console, and then turn the serverconfig on again

Feature requests:
* Instead of having to copy an entire collection, and then insert an element into it, then save it, have an extra command that allows one to add to a collection (be it a dictionary or a list, would require different syntax)
* An undo command that reverts the last edit. Does not have to be a full history, so just 1 depth

JavidPack:
Cons above mostly addressed in recent commit. I'll leave actual data validation for later if it is highly requested. As it is now it is functional for someone needing to directly adjust some simple config values on a server they only have console access to. 

Editing while the server is running is possible to implement in the future, it is just much more complicated to implement.


Chat tags show up as their `.Text`. I could have omitted them, but that looked weird for the examples I had available. Headers in yellow.

image

PR #3298, opened 2023-03-07
Fixes #3221 by adding a check to player and world names to prevent reserved keywords from being used. 

### LINKED ISSUE #3221: tModLoader has the capability of making corrupted files in Windows.
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Windows has some names what should never be set. As in, player character names, player mod pack names or mod names restricted names for Windows, Example, naming your character. ........, r....., CON, PRN, AUX, NUL, COM0, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. And many, many more reserved names. tModLoader will make a corrupted file by letting you name them as those special reserved names. You seriously need to block special names from being set. I accidentally named a character ........ Had to use a shredder tool to remove it.

### Log File

_No response_

### Steps to reproduce

First step, name your character or mod pack ........ or r.....
Second step, realized as you goofed.
Third step, pull your hair out while you are trying to delete the file. 

### Expected Behavior

Makes corrupted files what are extremely hard to remove.

### Actual Behavior

Making a headache.

### Reproduction frequency

On windows every single time.

### Additional Information

Be silly and make a mod pack, a character, or mod in general, those special characters/names.

Exiled1:
### What is the bug?
The bug is as stated in issue https://github.com/tModLoader/tModLoader/issues/3221 where files can be named after reserved filenames in windows, causing a potentially annoying cleanup process.

### How did you fix the bug?
I added a check to GetPlayerPathFromName and GetWorldPathFromName to sanitize against these specific reserved filenames.

I can't think of any alternatives at the moment. Other than not having used a reserved filename in the first place. Or using WSL or some external tool to remove these files.


Exiled1:
WOOOOOOOOOOOOOOOOOOOO ACTIVE CONTRIBUTION BABY

PR #3302, opened 2023-03-10
Mod config auto localization (Fix #3259)

### LINKED ISSUE #3259: Automatic localization keys for ModConfig members

Yes

Mod capability as a Modder

### Description

Proposed in discord chat (https://discord.com/channels/103110554649894912/1037051160705634369/1075545571923861604), this feature would automatically populate hjson files with entries for modconfig members.
```
		[Label("$Mods.ClickerClass.ClickerConfigClient.ShowCustomCursors.Label")]
		[Tooltip("$Mods.ClickerClass.ClickerConfigClient.ShowCustomCursors.Tooltip")]
		[DefaultValue(true)]
		public bool ShowCustomCursors;
```

Instead of adding Label and Tooltip attributes, and filling in the translation key, entries will be automatically generated as with other content. Either all entries are populated or only members annotated with `[Localized]`. To be decided.

_No response_

_No response_

JavidPack:
Fixes #3259 

If you are coming here from the migration guide or from the "Open Web Help" button, the Porting Notes section explains how to update your mod.

## Existing Behavior
Mods can annotate `ModConfig` fields with `TooltipAttribute` and `LabelAttribute` to assign labels and tooltips. The provided string is interpreted as a localization key if it starts with `$`, otherwise it is displayed as is.

`HeaderAttribute` adds a header above the entry and works in a similar manner.

## PR Implementation

- Modders can no longer provide display text to `ModConfig` members through the attributes. Localization files are used instead.
- All strings that refer to localization keys (in `LabelKey`, `TooltipKey`, `Header`, `LabelArgs`, `TooltipArgs`) must be prefixed with $. For example: `[LabelKey("$Mods.ExampleMod.Configs.Common.SampleEnumLabels.Strange")]`
- `[Label]` is obsolete, replaced with `[LabelKey]` if needed
  - Modders localize labels through the localization entries, which will automatically be generated. 
  - Autogenerated key: `Mods.{ModName}.Configs.{ClassName}.{MemberName}.Label`
  - Modders can use `[LabelKey]` to specify a specific localization key
  - Translation value defaults to the member name with spaces between capital letters: `someIntField` -> `some Int Field` 
  - Existing `[Label]` annotations will override that default during porting.
  - Modders using `[Label]` to force show config members in the UI when using `[JsonIgnore]` should now use `[ShowDespiteJsonIgnore]` instead.
  - If used previously to annotate `ModConfig` class, override `DisplayName` instead or just use autogenerated key. 
- `[Tooltip]` is obsolete, replaced with `[TooltipKey]` if needed
  - Modders localize tooltips through the localization entries, which will automatically be generated. 
  - Autogenerated key: `Mods.{ModName}.Configs.{ClassName}.{MemberName}.Tooltip`
  - Modders can use `[TooltipKey]` to specify a specific localization key
  - Translation value defaults `""`
  - Existing `[Tooltip]` annotations will override that default during porting.
  - For object references, if corresponding tooltip value is `""`, tooltip on the class will be shown as a fallback.
- `[Header]` has changed. 
  - `[Header]` supports manually provided key or an automatic key generated from a provided identifier.
    - `[Header("$Mods.ExampleMod.Configs.Common.LocalizedHeader")]` will use the provided key for the header text.
    - `[Header("HeaderIdentifier")]` will use an autogenerated key derived from the modname, class name, and provided identifier. For example, if this header was in the `ModConfigShowcaseLabels` class, the key would end up being `"Mods.ExampleMod.Configs.ModConfigShowcaseLabels.Headers.HeaderIdentifier"`
  - Identifier can't have spaces.
  - Modders will need to change existing `[Header]` by providing an identifier rather than the display text.
  - Autogenerated key: `Mods.{ModName}.Configs.{ClassName}.Headers.{Identifier}`
  - Translation value defaults to identifier (or member name if key was provided) with spaces between capital letters, followed by " Header": `VisualEffects` -> `Visual Effects Header` 
- Labels and Tooltips support substitutions through usage of `[LabelArgs]` and `[TooltipArgs]`. 
  - Strings provided are interpreted as a localization key if beginning with "$".
  - Scope simplification can be used to shorten the keys provided.
  - There are several extensive examples of this in `ModConfigShowcaseLabels`

```
ExampleModConfig: {
	DisplayName: Example Mod Config

	Headers.Items: Items

	ExampleWingsToggle: {
		Label: Example Wings Toggle
		Tooltip: Enables or disables the ExampleWings item
	}

	WeaponWithGrowingDamageToggle.Label: Weapon With Growing Damage Toggle
}
```

## Feedback Topics

1. ~~Should Label or Tooltip be opt-in? All fields should benefit from Label, but Tooltip isn't required~~
2. ~~Should the autogenerated keys for Headers be different? Their current Key isn't really logical.~~
3. ~~Should non-localization key Label/Tooltip/Header be allowed?~~
4. ~~What should be done about classes referenced in ModConfig, should they have automatic localization entries?~~

**New Feedback Topics**    

5. ~~Does it make sense to require `[Tooltip]`? It does prevent a disconnect between code and behavior, but a translation mod might want to add a tooltip to an entry that previously didn't have one.~~
6. ~~Topic 2 above is still a concern. Current behavior matches the current approach, but isn't easy to follow. `[Header("HeaderIdentifier")]` and "Mods.ModName.Configs.ConfigName.Headers.HeaderIdentifier" and so on could be useful.~~
7. ~~Topic 4 above is still a concern. Current ExampleMod shows them all in "Mods.ExampleMod.Configs.Common" for now. They might be rare enough that any automatic key solution would be hard to discover and use.~~

**Newer Feedback Topics**

8. Anything else?

## Porting Notes
1. Modders will need to find all `[Header]` annotations and change them to use either a specific key or an identifier. Removing spaces is the simplest approach.
    * Read the "`[Header]` has changed." bullet in the "PR Implementation" section for details on using an identifier.
2. For every config member with both `[JsonIgnore]` and `[Label]`, add `[ShowDespiteJsonIgnore]`
3. Save, build, and reload. If there are errors with headers, fix them and repeat.
4. The `.hjson` file should now be updated with entries for every config member. Each should have a default value. If `[Label]` and `[Tooltip]` were used previously, their values should be seen in the resulting `.hjson` entries.
5. Edit the `.hjson` file and fix any autogenerated label or header translations.
6. Test changes in game by saving the `.hjson` file and then navigating away and back to the `ModConfig` menu to refresh.
7. Keep testing changes until everything works as it did before. (For `[Header]`, you might need to remove " Header" from the auto-generated text.)
8. Delete all `[Label]` and `[Tooltip]` annotations, they have no effect anymore.
9. Optionally: Utilize the new features to clean up and organize, such as sharing localization keys or formatting the translated text with substitutions. See InterpolatedTextA/B/C in ModConfigShowcaseLabels.cs to see how that is done.


patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +74,113 @@ internal static void Add(ModConfig config)
...
+	{
+		foreach (PropertyFieldWrapper variable in ConfigManager.GetFieldsAndProperties(type)) {
+#pragma warning disable CS0618
+			if (Attribute.IsDefined(variable.MemberInfo, typeof(JsonIgnoreAttribute)) && !(Attribute.IsDefined(variable.MemberInfo, typeof(LabelAttribute)) || Attribute.IsDefined(variable.MemberInfo, typeof(ShowDespiteJsonIgnoreAttribute))))
+				continue;
+#pragma warning restore CS0618
+
+			try {
+				// TODO: Classes used in Generic types and Enums
+				// TODO: throw error after updating hjson somehow, so auto keys are populated?
```
Chicken-Bones: What does this mean?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +74,113 @@ internal static void Add(ModConfig config)
...
+
+				// Label and Tooltip will always exist. Header is optional, need to be used to exist.
+				var header = GetLocalizedHeader(variable, config, throwErrors: true);
+				if (header != null) {
+					string identifier = header.IsIdentifier ? header.identifier : variable.Name;
+					Language.GetOrRegister(header.key, () => $"{Regex.Replace(identifier, "([A-Z])", " $1").Trim()} Header");
+				}
+
+				string labelKey = GetLabelKey(variable, config, fallbackToClass: false, throwErrors: true);
+				if (labelKey.StartsWith($"Mods.{config.Mod.Name}")) {
```
Chicken-Bones: Hidden requirement?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +74,113 @@ internal static void Add(ModConfig config)
...
+					string labelValue = Regex.Replace(variable.Name, "([A-Z])", " $1").Trim();
+					if (labelObsolete != null)
+						labelValue = labelObsolete.LocalizationEntry;
```
Chicken-Bones: null coalescing? put inside the `() => ` so it only runs when necessary. Not actually harmful for performance, but is the intended pattern

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +74,113 @@ internal static void Add(ModConfig config)
...
+						string typeLabelKey = config.Mod.GetLocalizationKey($"Configs.{variable.Type.Name}.Label");
+						string typeLabelValue = Regex.Replace(variable.Type.Name, "([A-Z])", " $1").Trim();
+						if (typeLabelObsolete != null)
+							typeLabelValue = typeLabelObsolete.LocalizationEntry;
+						Language.GetOrRegister(typeLabelKey, () => typeLabelValue);
+					}
+
+					var typeTooltip = (TooltipKeyAttribute)Attribute.GetCustomAttribute(variable.Type, typeof(TooltipKeyAttribute));
+					if (typeTooltip != null) {
+						if (typeTooltip.malformed)
```
Chicken-Bones: Fair bit of duplicate code here and elsewhere. Can `LabelKeyAttribute` and `TooltipKeyAttribute` share a common base class?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +478,206 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		return Language.GetText(key).Format(args);
+	}
+
+	// Used to determine which key to register, based only on field/property, not class, not necessarily which key to use in UI.
+	internal static string GetLabelKey(PropertyFieldWrapper memberInfo, ModConfig config, bool fallbackToClass, bool throwErrors)
+	{
+		var label = (LabelKeyAttribute)Attribute.GetCustomAttribute(memberInfo.MemberInfo, typeof(LabelKeyAttribute));
+		if (label != null) {
+			if (label.malformed && throwErrors)
+				throw new ValueNotTranslationKeyException($"{nameof(LabelKeyAttribute)} only accepts localization keys for the 'key' parameter.");
```
Chicken-Bones: An alternative to `malformed` has been provided in Discord, but on code-style, please refactor to reduce duplicate code as much as possible. It's hard to read when differences in behaviour are mixed in with common copy paste.

In this case 'get and validate' should be a method.

You can even simplify things like `(LabelKeyAttribute)Attribute.GetCustomAttribute(memberInfo.Type, typeof(LabelKeyAttribute))` if they start getting too long, like `GetAttr<LabelKeyAttribute>(memberInfo.Type)`

Ideally, a reader should be able to read your method and quickly understand exactly _what_ it does at a reasonable level of detail, thanks to good method names, without needing to understand _how_ it does it. If I want to know exactly _how_ you check attributes for malformed translations, I'll look into the `GetAndValidate` method.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +478,206 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+				return label.key;
+		}
+
+		// Autokey. If member belongs to a Type from this mod...
+		Type typeMemberBelongsTo = memberInfo.MemberInfo.DeclaringType;
+		if (typeMemberBelongsTo.Assembly == config.GetType().Assembly) { 
+			string keyName = config.Name;
+			// Use Type name if not ModConfig instance.
+			if (typeMemberBelongsTo != config.GetType())
+				keyName = typeMemberBelongsTo.Name;
```
Chicken-Bones: Ternary preferred

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +478,206 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+			if (modName == "tModLoader" || modName == "FNA")
+				return $"Config.{typeMemberBelongsTo.Name}.{memberInfo.Name}.Label";
+			// Type is from some other mod.
+			return $"Mods.{modName}.Configs.{typeMemberBelongsTo.Name}.{memberInfo.Name}.Label";
+		}
+	}
+
+	internal static string GetLocalizedLabel(LabelKeyAttribute labelAttribute, PropertyFieldWrapper memberInfo) {
+		// Priority: Provided/Auto Key on member -> Key on class if member translation is empty string -> member name
+		var config = Interface.modConfig.pendingConfig;
```
Chicken-Bones: Not that I'm completely familiar with config loading, but using UI user state during loading code looks super confusing. I'm assuming you set some state in a `try finally` during load but...

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +478,206 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		Type typeMemberBelongsTo = memberInfo.MemberInfo.DeclaringType;
+		if (typeMemberBelongsTo.Assembly == config.GetType().Assembly) { 
+			string keyName = config.Name;
+			// Use Type name if not ModConfig instance.
+			if (typeMemberBelongsTo != config.GetType())
+				keyName = typeMemberBelongsTo.Name;
+
+			return config.Mod.GetLocalizationKey($"Configs.{keyName}.{memberInfo.Name}.Label");
+		}
+		else {
```
Chicken-Bones: something about this `if/else` feels fishy. Since there's no registration in this code, and this is some sort of fallback, why is it different for our mod vs other mods?

patches/tModLoader/Terraria/ModLoader/Config/UI/ConfigElement.cs
```diff
@@ -72,17 +73,15 @@ public void Bind(PropertyFieldWrapper memberInfo, object item, IList array, int
...
-			TextDisplayFunction = () => LabelAttribute.Label;
-		}
-
-		TooltipAttribute = ConfigManager.GetCustomAttribute<TooltipAttribute>(MemberInfo, Item, List);
-
-		if (TooltipAttribute != null) {
-			TooltipFunction = () => TooltipAttribute.Tooltip;
+		LabelAttribute = ConfigManager.GetCustomAttribute<LabelKeyAttribute>(MemberInfo, Item, List);
+		Label = ConfigManager.GetLocalizedLabel(LabelAttribute, MemberInfo);
+		// TODO: Support interpolating value?
```
Chicken-Bones: TODO?

patches/tModLoader/Terraria/ModLoader/Config/UI/UIModConfig.cs
```diff
@@ -437,7 +437,7 @@ public override void OnActivate()
 
 		SetMessage("", Color.White);
 
-		string configDisplayName = ((LabelAttribute)Attribute.GetCustomAttribute(modConfig.GetType(), typeof(LabelAttribute)))?.Label ?? modConfig.Name;
+		string configDisplayName = ConfigManager.GetModConfigDisplayName(modConfig);
```
Chicken-Bones: Good refactors in this file

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -621,6 +621,13 @@ private static LocalizationFile FindHJSONFileForKey(List<LocalizationFile> files
 				levelFound = level;
 				best = file;
 			}
+			if (!string.IsNullOrWhiteSpace(file.prefix)) {
+				level = file.prefix.Split(".").Length;
+				if (level > levelFound) {
+					levelFound = level;
+					best = file;
+				}
+			}
```
Chicken-Bones: would be cleaner to factor this into `LongestMatchingPrefix`, it could just take the `LocalizationFile` as an arg instead of `entries` and start with for loop at `file.Prefix.Split(".").Length`

Chicken-Bones (changes requested):
Code quality improvement suggestions.

Didn't make it all the way through `ConfigManager` due to time taken to read. Will re-review `ConfigManager` when some refactoring has been done to make it more readable.

Chicken-Bones (changes requested):
Code quality improvement suggestions.

Didn't make it all the way through `ConfigManager` due to time taken to read. Will re-review `ConfigManager` when some refactoring has been done to make it more readable.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: Hm, this meant the modder would see a half fixed hjson, even if they had a malformed attribute. It's not needed, modders should just follow the porting instructions I posted and do it in 1 pass. I'll remove the comment

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: Fixed this

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: Well, it only makes sense to register keys that will be exported in the first place. We don't have any way of knowing which mod's hjson to put translation keys that aren't `Mod.{ModName}`.

patches/tModLoader/Terraria/ModLoader/Config/ConfigAttributes.cs
```diff
@@ -40,16 +41,61 @@ public class ReloadRequiredAttribute : Attribute
...
+}
+
+[Obsolete("Labels are now automatically localized in localization files. Use LabelKeyAttribute to customize the key or use the autogenerated translation key instead. The functionality for showing this field or property in the ModConfig UI despite using [JsonIgnore] is now done through [ShowDespiteJsonIgnoreAttribute]")]
 [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Class)]
 public class LabelAttribute : Attribute
 {
 	private readonly string label;
 
 	public string Label => label.StartsWith("$") ? Localization.Language.GetTextValue(label.Substring(1)) : label;
+	public string LocalizationEntry => label.StartsWith("$") ? $"{{{label}}}" : label;
```
Chicken-Bones, line 99: Double `{{`?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +74,94 @@ internal static void Add(ModConfig config)
 		configList2.Add(GeneratePopulatedClone(config));
 	}
 
+	internal static void FinishSetup()
+	{
+		// Register localization for all fields and properties that should show
+		foreach (var activeConfigs in ConfigManager.Configs) {
+			foreach (var config in activeConfigs.Value) {
+				try {
+					string modConfigLabelKey = GetModConfigLabelKey(config, throwErrors: true);
```
Chicken-Bones: Should `ModConfig` not just have a `LocalizedText DisplayName` property?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,138 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+			return key;
+		}
+	}
+
+	private static T GetAndValidate<T>(MemberInfo memberInfo, bool throwErrors = true, bool errorOnType = false) where T : ConfigKeyAttribute
+	{
+		var configKeyAttribute = (T)Attribute.GetCustomAttribute(memberInfo, typeof(T));
+		if (configKeyAttribute != null) {
+			if (configKeyAttribute.malformed && throwErrors)
+				throw new ValueNotTranslationKeyException($"{nameof(T)} only accepts localization keys for the 'key' parameter.", errorOnType: true);
```
Chicken-Bones: You don't need the `errorOnType` param, because you can check `memberInfo is Type`

Chicken-Bones: Additionally, you use `errorOnType` like
```cs
				if (e.errorOnType) {
					e.additional = $"The class '{variable.Type.FullName}' caused this exception.";
				}
				else {
					e.additional = $"The member '{variable.Name}' found in the '{variable.MemberInfo.DeclaringType}' class caused this exception.";
				}
```

but you have `memberInfo` in `GetAndValidate` _is_ either `variable.Type` or `variable.MemberInfo` and `variable.Name` is `memberInfo.Name`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,138 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+				return configKeyAttribute.key;
+		}
+
+		// Autokey: Determine key from the Type the member belongs to.
+		Type typeMemberBelongsTo = memberInfo.MemberInfo.DeclaringType; 
+		string modName = typeMemberBelongsTo.Assembly.GetName().Name;
+		string className = typeMemberBelongsTo == config.GetType() ? config.Name : typeMemberBelongsTo.Name;
+		if (modName == "tModLoader" || modName == "FNA") // tModLoader keys handle translations for existing classes
+			return $"Config.{className}.{memberInfo.Name}.{dataName}";
+		return $"Mods.{modName}.Configs.{className}.{memberInfo.Name}.{dataName}";
```
Chicken-Bones: I notice a lot of complexity coming from needing to thread the `ModConfig` instance through a lot of methods where it is ultimately only used just for `{config.Mod.Name}` or `config.Name` to build paths.

In these code paths, you have to handle both members on a `ModConfig` instance, and members on a custom `Type` used as a config property (recursively).

I notice `ModConfig.Autoload` has a `// TODO: Does non-autoloaded ModConfigs have a use-case?`

The only time when it would be useful for `config.Name` to diverge from `config.GetType().Name` would be if there were two configs with the same type but different names (which is why `Name` is `virtual` in `ModType`)

Now whether we want to ditch manual calls to `Mod.AddConfig` is a different matter, but for localization purposes, I think it's actually _harmful_ to have the generated localization keys depend on `config.Name` rather than just `config.GetType().Name`. The reason being that if you _did_ have two configs with the same type but different names, all the localizations would be duplicated!

If you only have one config instance per type (assume they're singletons), then there is no reason for `config.Name` to not be `config.GetType().Name`

This is all by means of saying, we should do all localization registration on a `Type` basis, and consider `ModConfig` instances by nothing more than their `Type`. This will bring simplicity, consistency and reduce weird behavior in untested cases above.

Also, you can have a `HashSet<Type>` of types you've already registered.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Config/ConfigAttributes.cs
JavidPack: Yes, we are string interpolating and outputting a literal `{`. The result is that the translation entry will be `newkey: {$originalkey}`

JavidPack: `{{` is the escape sequence for a literal `{`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: I see, I missed passing in the argument there, but I can replace it anyway

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: implemented.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +81,82 @@ internal static void Add(ModConfig config)
 		configList2.Add(GeneratePopulatedClone(config));
 	}
 
+	internal static void FinishSetup()
+	{
+		typesWithLocalizationRegistered.Clear();
```
Chicken-Bones: Best to do this in `Unload`, or register a callback to `TypeCaching.OnClear` (which is just called from `Unload`)

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +81,84 @@ internal static void Add(ModConfig config)
...
+	{
+		AssemblyManager.GetAssemblyOwner(type.Assembly, out var modName);
+		foreach (PropertyFieldWrapper variable in ConfigManager.GetFieldsAndProperties(type)) {
+#pragma warning disable CS0618
+			if (Attribute.IsDefined(variable.MemberInfo, typeof(JsonIgnoreAttribute)) && !(Attribute.IsDefined(variable.MemberInfo, typeof(LabelAttribute)) || Attribute.IsDefined(variable.MemberInfo, typeof(ShowDespiteJsonIgnoreAttribute))))
+				continue;
+#pragma warning restore CS0618
+
+			try {
+				RegisterLocalizationKeysForMemberType(variable.Type, type.Assembly);
```
Chicken-Bones: not 100% actionable yet, but I believe enums would be better handled in ^. `[Header]` also shouldn't be supported for enums correct? It's just labels for their values.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +81,84 @@ internal static void Add(ModConfig config)
...
+			}
+			catch (ValueNotTranslationKeyException e) when (!e.handled) {
+				if (e.errorOnType) {
+					e.additional = $"The class '{variable.Type.FullName}' caused this exception.";
+				}
+				else {
+					e.additional = $"The member '{variable.Name}' found in the '{variable.MemberInfo.DeclaringType}' class caused this exception.";
+				}
+				e.Data["mod"] = modName;
+				e.handled = true; // Recursion will cause this to be called higher in the stack, causing issues.
```
Chicken-Bones: I have the impression that we can get this info, or just as good info at the place we throw the exception. Which would remove the need for the recursion bailout. 

`e.Data["mod"] = modName;` is pretty standard, but that could go in the `FinishSetup` loop (per mod)

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -72,6 +81,84 @@ internal static void Add(ModConfig config)
...
+				RegisterLocalizationKeysForMemberType(variable.Type, type.Assembly);
+
+				// Handle obsolete attributes. Use them to populate value of key, if present, to ease porting.
+#pragma warning disable CS0618 // Type or member is obsolete
+				var labelObsolete = (LabelAttribute)Attribute.GetCustomAttribute(variable.MemberInfo, typeof(LabelAttribute));
+				var tooltipObsolete = (TooltipAttribute)Attribute.GetCustomAttribute(variable.MemberInfo, typeof(TooltipAttribute));
+#pragma warning restore CS0618 // Type or member is obsolete
+
+				// Label and Tooltip will always exist. Header is optional, need to be used to exist.
+				var header = GetLocalizedHeader(variable, throwErrors: true);
```
Chicken-Bones: Can we remove `throwOnErrors`?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+
+			return key;
+		}
+	}
+
+	private static T GetAndValidate<T>(MemberInfo memberInfo) where T : ConfigKeyAttribute
+	{
+		var configKeyAttribute = (T)Attribute.GetCustomAttribute(memberInfo, typeof(T));
+		if (configKeyAttribute != null) {
+			if (configKeyAttribute.malformed)
```
Chicken-Bones: `&&`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+				throw new ValueNotTranslationKeyException($"{nameof(T)} only accepts localization keys for the 'key' parameter.", errorOnType: memberInfo is Type);
+		}
+		return configKeyAttribute;
+	}
+
+	// Used to determine which key to register, based only on field/property, not class, not necessarily which key to use in UI.
+	internal static string GetConfigKey<T>(MemberInfo memberInfo, string dataName) where T : ConfigKeyAttribute
+	{
+		var configKeyAttribute = GetAndValidate<T>(memberInfo);
+		if (configKeyAttribute != null) {
```
Chicken-Bones: null coalesce? Whole method could be one line.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		var configKeyAttribute = (T)Attribute.GetCustomAttribute(memberInfo, typeof(T));
+		if (configKeyAttribute != null) {
+			if (configKeyAttribute.malformed)
+				throw new ValueNotTranslationKeyException($"{nameof(T)} only accepts localization keys for the 'key' parameter.", errorOnType: memberInfo is Type);
+		}
+		return configKeyAttribute;
+	}
+
+	// Used to determine which key to register, based only on field/property, not class, not necessarily which key to use in UI.
+	internal static string GetConfigKey<T>(MemberInfo memberInfo, string dataName) where T : ConfigKeyAttribute
```
Chicken-Bones: `private`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		var configKeyAttribute = GetAndValidate<T>(memberInfo);
+		if (configKeyAttribute != null) {
+			return configKeyAttribute.key;
+		}
+
+		// Autokey: Determine key from the Type the member belongs to.
+		return GetDefaultLocalizationKey(memberInfo, dataName);
+	}
+
+	internal static string GetDefaultLocalizationKey(MemberInfo member, string dataName)
```
Chicken-Bones: `private`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		string memberKey = member is Type ? member.Name : $"{member.DeclaringType.Name}.{member.Name}";
+		return $"{groupKey}.{memberKey}.{dataName}";
+	}
+
+	internal static string GetLocalizedText<T, TArgs>(PropertyFieldWrapper memberInfo, string dataName, bool defaultToNull) where T : ConfigKeyAttribute where TArgs : ConfigArgsAttribute
+	{
+		// Priority: Provided/AutoKey on member -> Provided/AutoKey on class if member translation is empty string -> member name or null
+		var args = (TArgs)Attribute.GetCustomAttribute(memberInfo.MemberInfo, typeof(TArgs));
+		string configKey = GetConfigKey<T>(memberInfo.MemberInfo, dataName: dataName);
+		if (configKey != null && Language.Exists(configKey)) {
```
Chicken-Bones: Suggest `#nullable enable` for this file. `configKey` here cannot be null. There's also no way for `Language.Exists` to fail (because we register all of them). This means you can go straight to `Language.GetTextValue`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+			string configLocalization = Language.GetTextValue(configKey);
+			if (!string.IsNullOrEmpty(configLocalization))
+				return FormatTextAttribute(configKey, configLocalization, args?.args);
+		}
+
+		var typeConfigKey = (T)Attribute.GetCustomAttribute(memberInfo.Type, typeof(T));
+		if (typeConfigKey != null && Language.Exists(typeConfigKey.key)) {
+			string typeConfigLocalization = Language.GetTextValue(typeConfigKey.key);
+			if (!string.IsNullOrEmpty(typeConfigLocalization))
```
Chicken-Bones: `Exists` check is good here, we don't want `Label` to fallback to the type. _but_ the only way to get here would be if the entry in the localization file for `configKey` (`member.Label`) was `""`. In this case, we should just return `""`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		// Priority: Provided/AutoKey on member -> Provided/AutoKey on class if member translation is empty string -> member name or null
+		var args = (TArgs)Attribute.GetCustomAttribute(memberInfo.MemberInfo, typeof(TArgs));
+		string configKey = GetConfigKey<T>(memberInfo.MemberInfo, dataName: dataName);
+		if (configKey != null && Language.Exists(configKey)) {
+			string configLocalization = Language.GetTextValue(configKey);
+			if (!string.IsNullOrEmpty(configLocalization))
+				return FormatTextAttribute(configKey, configLocalization, args?.args);
+		}
+
+		var typeConfigKey = (T)Attribute.GetCustomAttribute(memberInfo.Type, typeof(T));
```
Chicken-Bones: No `GetConfigKey`? `[attr]` and `GetDefaultLocalizationKey(memberInfo.Type, dataName)` are split into two separate if statements?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		if (typeConfigKey != null && Language.Exists(typeConfigKey.key)) {
+			string typeConfigLocalization = Language.GetTextValue(typeConfigKey.key);
+			if (!string.IsNullOrEmpty(typeConfigLocalization))
+				return FormatTextAttribute(typeConfigKey.key, typeConfigLocalization, args?.args);
```
Chicken-Bones: does it make sense to combine `args` from member, with `[Tooltip]` from the type? Maybe yes! But what if people put `[TooltipArgs]` on the type?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		else if (memberInfo.Type.IsClass || memberInfo.Type.IsEnum) {
+			string typeKey = GetDefaultLocalizationKey(memberInfo.Type, dataName);
+			if (Language.Exists(typeKey)) {
+				string typeConfigLocalization = Language.GetTextValue(typeKey);
+				if (!string.IsNullOrEmpty(typeConfigLocalization))
+					return FormatTextAttribute(typeKey, typeConfigLocalization, args?.args);
+			}
+		}
+
+		return defaultToNull ? null : memberInfo.Name;
```
Chicken-Bones: As mentioned above, don't think this fallback behaviour makes sense anymore/is wise

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +459,119 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+	}
+
+	internal static HeaderAttribute GetLocalizedHeader(PropertyFieldWrapper memberInfo, bool throwErrors)
+	{
+		// Priority: Provided Key or key derived from identifier on member
+		var header = GetCustomAttribute<HeaderAttribute>(memberInfo, null, null);
+		if (header != null) {
```
Chicken-Bones: early return?

patches/tModLoader/Terraria/ModLoader/Config/ModConfig.cs
```diff
@@ -19,6 +21,12 @@ public abstract class ModConfig
 	[JsonIgnore]
 	public string Name { get; internal set; }
 
+	[JsonIgnore]
+	public string LocalizationCategory => "Configs";
```
Chicken-Bones, line 28: ~Can't implement `ILocalizedModType` I suppose, hmm, cause they don't autoload. I think they should extend `ILocalizedModType` but just add `[Autoload(false)]` to `ModConfig`, should work~

~Might be confusing though? Since `[Autoload(false)]` would imply they don't autoload, but they do, just... before other `ILoadable`. Bleh I think actually `IModType` and `ILoadable` should be split. I'll do that.~

See https://github.com/tModLoader/tModLoader/commit/5a4a96eca31aa647265f38dda2de261e17ba3a5d, can now extend from `ILocalizedModType` without issues. Should probably slap a `ModTypeLookup.Register` call in there for them too.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: Language.Exists can fail, for example a vanilla class without a registered localization. (EquipType, for one.)

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: We still need this for some vanilla classes that might not have a key registered.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -292,33 +391,29 @@ public static object AlternateCreateInstance(Type type)
 
 	// Gets an Attribute from a property or field. Attribute defined on Member has highest priority,
 	// followed by the containing data structure, followed by attribute defined on the Class.
-	public static T GetCustomAttribute<T>(PropertyFieldWrapper memberInfo, object item, object array) where T : Attribute
+	public static T? GetCustomAttribute<T>(PropertyFieldWrapper memberInfo, object? item, object? array) where T : Attribute
```
Chicken-Bones: Can we clean this method up so it has well defined behavior?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -292,33 +391,29 @@ public static object AlternateCreateInstance(Type type)
...
-
-		attribute = (T)Attribute.GetCustomAttribute(type, typeof(T), true) ?? attribute;
-
-		// Member
-		attribute = (T)Attribute.GetCustomAttribute(memberInfo.MemberInfo, typeof(T)) ?? attribute;
-		return attribute;
+		return
+			(T?)Attribute.GetCustomAttribute(memberInfo.MemberInfo, typeof(T)) ?? // on the member itself
+			(T?)Attribute.GetCustomAttribute(type, typeof(T), true) ?? // on a provided fallback type
+			(T?)Attribute.GetCustomAttribute(memberInfo.Type, typeof(T), true); // on the member type. TODO is there ever a reason to use this?
```
Chicken-Bones: See my todo here. I don't think this last fallback makes sense. `type` should probably be called `fallbackType`. Then you can pass in `MemberInfo` rather than `PropertyFieldWrapper`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +451,102 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		string memberKey = member is Type ? member.Name : $"{member.DeclaringType!.Name}.{member.Name}";
+		return $"{groupKey}.{memberKey}.{dataName}";
+	}
+
+	internal static string? GetLocalizedText<T, TArgs>(PropertyFieldWrapper memberInfo, string dataName) where T : ConfigKeyAttribute where TArgs : ConfigArgsAttribute
+	{
+		bool isTooltip = typeof(T) == typeof(TooltipKeyAttribute);
+
+		// Priority: Provided/AutoKey on member -> Provided/AutoKey on class if member translation is empty string and T is Tooltip -> member name or null
+		var args = GetCustomAttribute<TArgs>(memberInfo, memberInfo.Type);
```
Chicken-Bones: do we want to pull args from the `memberInfo.Type` if the label is coming from the `memberInfo`? Shouldn't do anything bad because there should be no substitution placeholders in the label, but still seems strange.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +451,102 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+		if (Language.Exists(configKey)) {
+			string configLocalization = Language.GetTextValue(configKey);
+			if (!(isTooltip && string.IsNullOrEmpty(configLocalization)))
+				return FormatTextAttribute(configKey, configLocalization, args?.args);
+		}
+
+		if (memberInfo.Type.IsClass || memberInfo.Type.IsEnum) {
+			string typeConfigKey = GetConfigKey<T>(memberInfo.Type, dataName);
+			if (Language.Exists(typeConfigKey))
+				return FormatTextAttribute(typeConfigKey, Language.GetTextValue(typeConfigKey), args?.args);
```
Chicken-Bones: Here we apply args which could be from the member, to a tooltip on the type. Might be intended in this case?

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +451,102 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+	{
+		bool isTooltip = typeof(T) == typeof(TooltipKeyAttribute);
+
+		// Priority: Provided/AutoKey on member -> Provided/AutoKey on class if member translation is empty string and T is Tooltip -> member name or null
+		var args = GetCustomAttribute<TArgs>(memberInfo, memberInfo.Type);
+		string configKey = GetConfigKey<T>(memberInfo.MemberInfo, dataName);
+		if (Language.Exists(configKey)) {
+			string configLocalization = Language.GetTextValue(configKey);
+			if (!(isTooltip && string.IsNullOrEmpty(configLocalization)))
+				return FormatTextAttribute(configKey, configLocalization, args?.args);
```
Chicken-Bones: If you use `Language.GetText` instead of `GetTextValue`, and pass the `LocalizedText` to `FormatTextAttribute` then you can access `.key` and `.value` for the scoping

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +451,102 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+	}
+
+	internal static string? GetLocalizedText<T, TArgs>(PropertyFieldWrapper memberInfo, string dataName) where T : ConfigKeyAttribute where TArgs : ConfigArgsAttribute
+	{
+		bool isTooltip = typeof(T) == typeof(TooltipKeyAttribute);
+
+		// Priority: Provided/AutoKey on member -> Provided/AutoKey on class if member translation is empty string and T is Tooltip -> member name or null
+		var args = GetCustomAttribute<TArgs>(memberInfo, memberInfo.Type);
+		string configKey = GetConfigKey<T>(memberInfo.MemberInfo, dataName);
+		if (Language.Exists(configKey)) {
```
Chicken-Bones: if `configKey` doesn't exist, we should go straight to returning `null`. Because that means that the member isn't part of a mod, no point attempting type fallback. 

I don't like the `isTooltip` check here, consider two overloads `GetLocalizedLabel` and `GetLocalizedTooltip` both of which return `string` rather than `string?`. Tooltip can null coalesce to `""` and label to `memberInfo.Name`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -356,6 +451,102 @@ public static bool EnumerableEquals(IEnumerable a, IEnumerable b)
...
+				return FormatTextAttribute(typeConfigKey, Language.GetTextValue(typeConfigKey), args?.args);
+		}
+
+		return isTooltip ? null : memberInfo.Name;
+	}
+
+	internal static HeaderAttribute? GetLocalizedHeader(PropertyFieldWrapper memberInfo)
+	{
+		// Priority: Provided Key or key derived from identifier on member
+		var header = GetCustomAttribute<HeaderAttribute>(memberInfo, null, null);
```
Chicken-Bones: Does it make sense to use `GetCustomAttribute` here? Rather than just pulling it straight from `memberInfo.MemberInfo`? `Headers` aren't allowed to be on anything but the member. Then you can change the parameter type from `PropertyFieldWrapper` to `MemberInfo`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: I think it can't hurt, and I can imagine situations where it could be used, so I'm fine with it.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: Same, I can imagine situations where it could be useful, and it shouldn't cause issues.

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
JavidPack: I don't understand why we wouldn't want to fallback to the Type for non-mod classes, the Type might have a localization attribute on it.

JavidPack: Ok, I missed that returning null would use the member.Name. We don't have any vanilla types with fields of vanilla types, so we don't need to think about this. If we do, we can just localize them.

Chicken-Bones (approved):

PR #3304, opened 2023-03-12
ModTypeLookup.RegisterLegacyNames method

direwolf420:
### What is the new feature?
An alternative for `LegacyNamesAttribute` for cases where manual loading is used. Also adjusted an error message when duplicate content is registered to account for these legacy names, and added docs for this whole legacy/alias feature (it had none previously).

Currently, you can't programmatically define legacy/alias names for content. You can however for the actual name of the content (as it's an overridable property). This now makes things possible where you are refactoring manually loaded content and want it to be backwards compatible with its "older" version.

~~Yes. I don't like the fact that I have to go a similar route to `IModType.FullName` and having to implement that in `ModTileEntity` and `ModType` (no default interface properties :cry:). This forces modders who have custom IModTypes to implement this now aswell, which is breaking (?).
The backend could be handled better, instead of split between attribute and property some centralized system within ModTypeLookup could be set up to store the legacy names.~~
Has been rewritten to use the new method.

```cs
public class ManuallyLoadedItemLoader : ILoadable
{
	public void Load(Mod mod)
	{
		AddContentWithLegacyNames(mod, new ManuallyLoadedItem("SomeCoolerItem"), "SomeItem", "SomeOlderItem");
		AddContentWithLegacyNames(mod, new ManuallyLoadedItem("OtherItemToImplyThatAttributeHereIsInsufficientAsBothAreTheSameClass"), "NotSomeItem");
	}

	//Convenience method, modders may opt to make a Mod extension method for this instead
	private static void AddContentWithLegacyNames(Mod mod, ModItem instance, params string[] legacyNames) {
		mod.AddContent(instance);
		ModTypeLookup<ModItem>.RegisterLegacyNames(instance, legacyNames);
	}

	public void Unload()
	{

	}
}

[Autoload(false)]
public class ManuallyLoadedItem : ModItem
{
	private readonly string name;
	public override string Name => name;

	public ManuallyLoadedItem(string name)
	{
		this.name = name;
	}

	protected override bool CloneNewInstances => true;
}
```

### ExampleMod updates
Nothing in ExampleMod uses this.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
@@ -6,6 +6,9 @@ using Terraria.ModLoader.Core;
 
 namespace Terraria.ModLoader;
 
+/// <summary>
+/// Enables content instances to register with <see cref="Register(T)"/> towards retrieval via <see cref="ModContent.Find{T}(string)"/> and similar methods.
+/// </summary>
 public static class ModTypeLookup<T> where T : IModType
 {
 	private static readonly Dictionary<string, T> dict = new Dictionary<string, T>();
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/ModLoader/ModTypeLookup.cs
```diff
@@ -24,8 +25,8 @@ public static void Register(T instance)
 	{
 		RegisterWithName(instance, instance.Name, instance.FullName);
 
-		//Add legacy aliases, if the type has any.
-		foreach (string legacyName in LegacyNameAttribute.GetLegacyNamesOfType(instance.GetType())) {
+		// Add legacy aliases, if it has any.
+		foreach (string legacyName in GetLegacyNames(instance)) {
 			RegisterWithName(instance, legacyName, $"{instance.Mod?.Name ?? "Terraria"}/{legacyName}");
```
Chicken-Bones, line 46: A better option would just be to make a public method:
```cs
public void RegisterLegacyName(T instance, string legacyName) => RegisterWithName(instance, legacyName, $"{instance.Mod?.Name ?? "Terraria"}/{legacyName}")
```

Chicken-Bones (changes requested):

direwolf420:
Updated PR title and text to reflect changes

Issue #3307, opened 2023-03-14
Add a buffName parameter to ModifyBuffTip hooks


stormytuna:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Adding a `buffName` parameter to `ModifyBuffTip` hooks would allow modders to easily change a buff's display name dynamically. I propose adding a `ref string buffName` parameter, similar to how buff tooltips are changed using the same hook.

### What does this proposal attempt to solve or improve?

It's currently difficult to dynamically change a buff's display name, requiring advanced methods like IL or reflection (unless I'm missing something obvious)

One example could be for animating a buff's display name, if a Modder wanted something like Minecraft's glitched/scrambled text (see the *k* part of [this gif](https://static.wikia.nocookie.net/minecraft_gamepedia/images/7/7e/Minecraft_Formatting.gif/revision/latest?cb=20200828001454))

### Which (other) solutions should be considered?

It would be possible to separate out modifying the tooltip and modifying the display name into two different hooks, possibly `ModifyBuffTip` and `ModifyBuffName`

JavidPack:
This seems reasonable. I think the easiest solution is rename `ModifyBuffTip` to `ModifyBuffText` and add the `buffName` parameter.

Whoever implements this should also make sure to add the tModPorter rule and test cases.

PR #3309, opened 2023-03-15
Changes ModifyBuffTip hook into ModifyBuffText hook (#3307)

stormytuna:
Implements #3307

### What is the new feature?

Extends functionality of `ModifyBuffTip` hook to allow modders to modify a buff's display name. Also renames it to `ModifyBuffText` to better fit with what it actually does.

Allows modders to dynamically change a buff's display name, something that was only possible through more advanced methods like reflection.

Could've separated it into two hooks, one for the tooltip and one for display name.

```csharp
// In a ModBuff class, a simple example of how a stacking buff could utilise this hook
public override void ModifyBuffText(ref string buffName, ref int rare, ref string tip) {
  StackingBuffPlayer modPlayer = Main.LocalPlayer.GetModPlayer<StackingBuffPlayer>(); // A ModPlayer class that tracks our buff's stacks
  buffName = $"Stacking buff (x{modPlayer.BuffStacks})"; // Changes with our player's buff stacks
  tip = $"Dealing ${modPlayer.BuffStacks}% more damage!"; // Also changes with our player's buff stacks
}
```

### ExampleMod updates

Updated `ExampleGlobalBuff` to use the new hook, no actual example of changing a buff's display name though.

### Other notes

I didn't change the naming conventions from `ModifyBuffTip`, so `tip` and `rare` are still called `tip` and `rare`, but I think they should be called `buffTooltip` and `buffNameRarity`/`buffRare`/`somethingLikeThat`.

I did change the ordering, so it goes `buffName`, `rare` then `tip` instead of `buffName`, `tip` then `rare` so that it groups the "buff display name" parameters but that can be changed back, not sure how big of a deal it is.

And finally I'm not sure if there needs to be an example of this in ExampleMod


tModPorter/tModPorter.Tests/TestData/GlobalBuffTest.Expected.cs
```diff
@@ -0,0 +1,6 @@
+﻿using Terraria.ModLoader;
+
+public class GlobalBuffTest : GlobalBuff
+{
+	public override void ModifyBuffText(int buff, ref string buffName, ref int rare, ref string tip) { }
```
JavidPack: I don't think you ran the tests. This should end up as `type` not `buff`. Change GlobalBuffTest.cs to type as well.

ExampleMod/Common/GlobalBuffs/ExampleGlobalBuff.cs
```diff
@@ -48,7 +48,22 @@ public class ExampleGlobalBuff : GlobalBuff
 			return true;
 		}
 
-		public override void ModifyBuffTip(int type, ref string tip, ref int rare) {
+		//public override void ModifyBuffTip(int type, ref string tip, ref int rare) {
```
JavidPack: You've commented out old code. Please delete all the old code instead.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3914,7 +3914,8 @@
 +					if (meleeBuff[num34])
 +						rare = -10;
 +
-+					BuffLoader.ModifyBuffTip(num34, ref buffTooltip, ref rare);
++					// BuffLoader.ModifyBuffTip(num34, ref buffTooltip, ref rare);
++					BuffLoader.ModifyBuffText(num34, ref buffName, ref rare, ref buffTooltip);
```
JavidPack: Lets go with `num34, ref buffName, ref buffTooltip, ref rare` instead.

patches/tModLoader/Terraria/ModLoader/ModBuff.cs
```diff
@@ -87,7 +87,14 @@ public virtual bool ReApply(NPC npc, int time, int buffIndex)
...
 	/// </summary>
-	public virtual void ModifyBuffTip(ref string tip, ref int rare)
+	//public virtual void ModifyBuffTip(ref string tip, ref int rare)
+	//{
+	//}
+
+	/// <summary>
+	/// Allows you to modify the name and tooltip that displays when the mouse hovers over the buff icon, as well as the color the buff's name is drawn in.
+	/// </summary>
+	public virtual void ModifyBuffText(ref string buffName, ref int rare, ref string tip)
```
JavidPack: Same here, buffName, tip, rare

JavidPack (changes requested):
Looking good, please make the changes requested and I think it should be good.

JavidPack:
Good job, thanks. I added an example I thought of and merged it, thanks for contributing.

Issue #2667, opened 2022-07-02
Mod Accessory Slot Unloaded Bug


Solxanich:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Ishigh
I'm getting issues with an unloaded accessory slot, seems that new UnloadedAccessorySlot doesn't increase the size of exHideAccessory, triggering an issue immediately in ModAccessorySlotPlayer.LoadData
(lines 78-89)

Occurs in player select screen when one character has more unloaded slots than another

Solxan:
I think the expectation is supposed to be:
Size = Current Unloaded allocations + this character save data length.

Solxan:
but if there is a new unloaded slot introduced on that player that wasn't there on previous players, than you end up with save data - last unloaded slots + current unloaed slots = 1
Which gives an error 

Ishigh:
Well, based on what you just said, I tried ResizeAccesoryArrays(UnloadedSlotCount + order.Count);
It seems to work
That or resize the array every time you add a new unloadded slot instead

PR #3310, opened 2023-03-15
Fix Unloaded Mod Accessory Slot #2667

Solxanich:
### What is the bug?
#2667 

### How did you fix the bug?
Updated the logic to move all unloaded slots to only exist on the ModPlayer instead of the Loader.
There is no server requirement to know about unloaded slots on a ModPLayer, and as such, no need for it to be in the Loader.

By eliminating Loader involvement, this ensures any unloaded nuances are kept contained within a single player character.

A complete redesign that includes consideration to this? But not really something to jump in to easily.


patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -268,6 +258,8 @@ public static void SendVisualState(int toWho, int plr, int slot, bool hideVisual
 
 		private static void HandleVisualState(BinaryReader r, int fromWho)
 		{
+			// Does this receiving need to factor in difference between LoadedCount and Total?
+			// Shouldn't?
```
Chicken-Bones: No, only loaded slots are sent via `SendVisualState`

PR #3329, opened 2023-03-19
Fix Town NPC Melee Swinging Source Rectangle

Rijam:
### What is the bug?

Setting any texture for a Town NPC's melee swinging will always draw as a 36x36 sprite. Even if the sprite is not that size. Smaller sprites will have their edges "extended" and larger sprites will be cut off. The only way for a modder to fix this is to IL edit (or detour the entirety of `Main.DrawNPCExtras()`).

The bug is that the source rectangle for the drawing of the item is always set as a 36x36 rectangle and does not update the size to match the texture that the modder set.

The Exotic Scimitar has a resolution of 40x48, so it gets cut off.
Capture

The following is the current code. Comments have been added and variables have been renamed to make it easier to understand.
```cs
// In Main.DrawNPCExtras()
// If the Town NPC has the melee attack type and is about to attack.
if (NPCID.Sets.AttackType[n.type] == 3 && n.ai[0] == 15f) {
	int itemSize = 32;
	float scale = 0f;
	Vector2 offset = Vector2.Zero;
	short itemType = 4; // The default item is an Iron Broadsword. It has a resolution of 36x36.
	
	// Set values for vanilla NPCs. Modded NPCs don't apply so skipping this section.
	...
	// After the values for the vanilla NPCs have been set...

	// Get the item frame. This also loads the texture with Main.instance.LoadItem(itemType).
	GetItemDrawFrame(itemType, out var itemTexture2, out var rectangle3);
	// rectangle3 = (0, 0, 36, 36)

	// Load the values for the Modded NPCs...
	NPCLoader.DrawTownAttackSwing(n, ref itemTexture2, ref itemSize, ref scale, ref offset); 
	// Modder loads their new texture, sets the itemSize, etc.
	// Uh oh, rectangle3 is never updated to the new texture!!!!

	Tuple<Vector2, float> swingStats = n.GetSwingStats(NPCID.Sets.AttackTime[n.type] * 2, (int)n.ai[1], n.spriteDirection, itemSize, itemSize);
	Vector2 drawPos = swingStats.Item1 + (swingStats.Item1 - n.Center) * scale + offset;
	Vector2 origin = rectangle3.Size() * new Vector2((n.spriteDirection != 1) ? 1 : 0, 1f);
	
	spriteBatch.Draw(itemTexture2,	// Texture
		new Vector2((int)(drawPos.X - screenPosition.X), (int)(drawPos.Y - screenPosition.Y)), // position
		rectangle3, // source rectangle. Wait, this is still the source rectangle for the Iron Broadsword!!!
		n.GetAlpha(npcColor), // color
		swingStats.Item2, // rotation
		origin, // origin. This is now also incorrect because it used the source rectangle for the Iron Broadsword.
		n.scale, // scale
		npcSpriteEffect ^ SpriteEffects.FlipHorizontally, // sprite effects
		0f);
}
```

### How did you fix the bug?

I fixed the bug by swapping `GetItemDrawFrame()` and `NPCLoader.DrawTownAttackSwing()`. I changed `NPCLoader.DrawTownAttackSwing()`'s `ref Texture2D item` to `ref int item` and had it modify the item type instead of the item texture.

`DrawTownAttackSwing(ref Texture2D item, ref int itemSize, ref float scale, ref Vector2 offset)` -> `DrawTownAttackSwing(ref int item, ref int itemSize, ref float scale, ref Vector2 offset)`

`GetItemDrawFrame()` already loads the new texture with `Main.instance.LoadItem(itemType)`, so there is no need for the modder to do it themselves.

This change not only fixes the bug, but it also make it easier for modders and more in line with [`DrawTownAttackGun()`](https://docs.tmodloader.net/docs/preview/class_mod_n_p_c.html#a6420c889d630d0cd51591ee4706829bb). Instead of the modder needing to load an item's texture and then passing the Texture2D, they can just pass the item type.

I also added it to tModPorter and it worked when I tested it on one of my mods (Tell me if I did it wrong, though).

The fix in action:
Capture 2 Capture 4

Recalculate the source rectangle instead of modifying the item type itself. I didn't choose this way because the Texture2D variable is from the `out` of `GetItemDrawFrame`. So, `NPCLoader.DrawTownAttackSwing` couldn't be called before then.

There is still the oddity of `itemSize` only being 1D instead of 2D. `GetSwingStats()` can take both an `itemWidth` and `itemHeight`, but `itemSize` is passed for both. This is how it is in vanilla, and it doesn't make that much of a difference, so I don't think it really matters.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -3181,6 +3181,15 @@
  			Vector2 vector4 = DrawPlayerItemPos(1f, itemtype);
  			GetItemDrawFrame(itemtype, out var itemTexture, out var value2);
  			int num12 = (int)vector4.X - num11;
+@@ -22027,7 +_,7 @@
+ 			int num13 = 32;
+ 			float num14 = 0f;
+ 			Vector2 zero = Vector2.Zero;
+-			short num15 = 4;
++			int num15 = 4; // TML: Changed short to int.
+ 			if (n.type == 207) {
+ 				num15 = 3349;
+ 				num14 = 0.15f;
 @@ -22048,6 +_,7 @@
  			}
  			else if (n.type == 441) {
@@ -3192,9 +3201,9 @@
 @@ -22058,6 +_,9 @@
  			}
  
- 			GetItemDrawFrame(num15, out var itemTexture2, out var rectangle3);
++			NPCLoader.DrawTownAttackSwing(n, ref num15, ref num13, ref num14, ref zero);
 +
-+			NPCLoader.DrawTownAttackSwing(n, ref itemTexture2, ref num13, ref num14, ref zero);
+ 			GetItemDrawFrame(num15, out var itemTexture2, out var rectangle3);
 +
  			Tuple<Vector2, float> swingStats = n.GetSwingStats(NPCID.Sets.AttackTime[n.type] * 2, (int)n.ai[1], n.spriteDirection, num13, num13);
  			Vector2 vector6 = swingStats.Item1 + (swingStats.Item1 - n.Center) * num14 + zero;
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -3181,15 +3181,6 @@
  			Vector2 vector4 = DrawPlayerItemPos(1f, itemtype);
  			GetItemDrawFrame(itemtype, out var itemTexture, out var value2);
  			int num12 = (int)vector4.X - num11;
-@@ -22027,7 +_,7 @@
- 			int num13 = 32;
- 			float num14 = 0f;
- 			Vector2 zero = Vector2.Zero;
--			short num15 = 4;
-+			int num15 = 4; // TML: Changed short to int.
- 			if (n.type == 207) {
- 				num15 = 3349;
- 				num14 = 0.15f;
 @@ -22048,6 +_,7 @@
  			}
  			else if (n.type == 441) {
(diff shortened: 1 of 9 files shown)


Rijam:
I forgot to update the documentation for the hook. I'll do that soon.

JavidPack:
I'll look more into this tomorrow probably, but I think removing the ability to assign any texture2d will be disappointing to many existing mods. Not every modded townnpc uses an actual item texture I assume.

Just fixing the rectangle seems more correct. Does the 1.3 code have the same issue?

Rijam:
With that logic we should change `DrawTownAttackGun()` to a Texture2D :P

Ok it looks like this issue is unique to 1.4.4. It was changed for vanilla Town NPCs to edit the item type instead of Texture2D. All previous version just set a Texture2D. The source rectangle is set to `null` and the origin just uses the `Texture2D.Size()` instead of the `rectangle.Size()`.

We could leave everything the same and just change the origin and spriteBatch.Draw()
```cs
// rectangle3.Size() changed to itemTexture2.Size()
Vector2 origin4 = itemTexture2.Size() * new Vector2((n.spriteDirection != 1) ? 1 : 0, 1f);
// rectangle3 sourceRectangle change to itemTexture2.Bounds (or we could just put null like previous versions).
spriteBatch.Draw(itemTexture2, new Vector2((int)(vector6.X - screenPosition.X), (int)(vector6.Y - screenPosition.Y)), itemTexture2.Bounds, n.GetAlpha(npcColor), swingStats.Item2, origin4, n.scale, npcSpriteEffect ^ SpriteEffects.FlipHorizontally, 0f);
```

JavidPack:
Didn't see your comment before working on this. The sourceRectangle is useful for animated textures, so I added it instead of assuming the full texture should be drawn.

Also, I asked and a mod author confirmed they were using arbitrary textures for some town npc . 

Yes, the DrawTownAttackGun could probably also have texture parameters, but there is some code there for holdout offset that I didn't want to work through at the moment, and I don't think anyone has asked for it anyway....Maybe I'll add that before merging.

Issue #3237, opened 2023-02-04
[1.4.4.9] @linux Does not find dotnet sdk 6 (in game at develop mods button) but it is installed


Unftf:
### Version

Latest 1.4-preview

### OS

Linux

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

While testing the 1.4.4.9 beta branch the game did not found dotnet sdk 6 which I installed on the system.
I was unable to build my mods in 'Develop Mods' menu.

But if I type that command into bash I get:
```
~$ dotnet --list-sdks
6.0.405 [/usr/share/dotnet/sdk]`
```

I installed it with
```
sudo apt-get install dotnet-sdk-6.0 
from rep: deb [arch=amd64] https://packages.microsoft.com/ubuntu/20.04/prod focal main
```

### Log File

On clicking onto it I get this error message in bash:
```
[Main Thread/DEBUG] [tML]: Process.Start (UseShellExecute = False): "dotnet" --list-sdks
[Main Thread/INFO] [tML]: 

[Main Thread/WARN] [tML]: Silently Caught Exception: 
System.ArgumentException: Version string portion was too short or too long. (Parameter 'input')
   at System.Version.ParseVersion(ReadOnlySpan\1 input, Boolean throwOnFailure)
   at System.Version.Parse(String input)
   at System.Version..ctor(String version)
   at Terraria.ModLoader.UI.UIModSources.CheckDotnet() in tModLoader\Terraria\ModLoader\UI\UIModSources.cs:line 284
   at Terraria.ModLoader.UI.UIModSources.ShowInfoMessages() in tModLoader\Terraria\ModLoader\UI\UIModSources.cs:line 252
   at Terraria.ModLoader.UI.UIModSources.OnActivate() in tModLoader\Terraria\ModLoader\UI\UIModSources.cs:line 230
   at Terraria.UI.UIElement.Activate() in tModLoader\Terraria\UI\UIElement.cs:line 452
   at Terraria.UI.UserInterface.SetState(UIState state) in tModLoader\Terraria\UI\UserInterface.cs:line 262
   at Terraria.GameContent.UI.States.UIWorkshopHub.Click_OpenModSourcesMenu(UIMouseEvent evt, UIElement listeningElement) in tModLoader\Terraria\GameContent\UI\States\UIWorkshopHub.TML.cs:line 83
   ..

[Main Thread/DEBUG] [tML]: 'dotnet --list-sdks' check failed: 
[...]

PR #3334, opened 2023-03-22
Fixed problem with detecting dotnet on linux

Fili-pk:
What is the bug?
https://github.com/tModLoader/tModLoader/issues/3237
After entering the Develop Mods tab on linux a message is shown to install dotnet 6 even though it is already installed.
The bug is caused by name collision between system dotnet and dotnet build in tmodloader.

How did you fix the bug?
I managed to fix the bug by finding dotnet in PATH and checking available sdk by running dotnet using absolute path.

I think that one potential alternative is to rename build-in dotnet.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
@@ -251,7 +251,7 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
 			return true;
 		}
 
-		if (!IsCompatibleDotnetSdkAviable()) {
+		if (!IsCompatibleDotnetSdkAvailable()) {
 			ShowWelcomeMessage("tModLoader.DownloadNetSDK", "https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-developers#developing-with-tmodloader", 888, PreviousUIState);
 			return true;
 		}
@@ -311,7 +311,7 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
 		return null;
 	}
 
-	bool IsCompatibleDotnetSdkAviable()
+	bool IsCompatibleDotnetSdkAvailable()
 	{
 		if (dotnetSDKFound)
 			return true;
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
@@ -309,14 +309,9 @@ internal class UIModSources : UIState, IHaveBackButtonCommand
 		if (dotnetSDKFound)
 			return true;
 
-		string dotnetPath = GetSystemDotnetPath();
-		if (dotnetPath == null)
-			return false;
-
-
 		try {
 			string output = Process.Start(new ProcessStartInfo {
-				FileName = dotnetPath,
+				FileName = GetSystemDotnetPath() ?? "dotnet",
 				Arguments = "--list-sdks",
 				UseShellExecute = false,
 				RedirectStandardOutput = true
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -266,14 +268,56 @@ private void ShowWelcomeMessage(string altButtonTextKey, string url, int gotoMen
...
+	}
+
+	bool IsCompatibleDotnetSdkAviable()
 	{
 		if (dotnetSDKFound)
 			return true;
 
+		string dotnetPath = GetSystemDotnetPath();
+		if (dotnetPath == null)
+			return false;
```
Chicken-Bones: Just change this to `GetSystemDotnetPath() ?? "dotnet"` so at least we get the old behavior as fallback

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -266,14 +268,56 @@ private void ShowWelcomeMessage(string altButtonTextKey, string url, int gotoMen
...
+		if (commandToFindPathOfExecutable == null)
+			return null;
+
+		try {
+			return Process.Start(new ProcessStartInfo {
+				FileName = commandToFindPathOfExecutable,
+				Arguments = "dotnet",
+				UseShellExecute = false,
+				RedirectStandardOutput = true
+			}).StandardOutput.ReadToEnd().Trim();
```
Chicken-Bones, line 298: Please add a `File.Exists` check on the result of this command and return `null` if failed

Chicken-Bones (changes requested):
Thanks for this. I'm assuming even though we haven't added our dotnet to `PATH`, the location of the current process (not even the working directory) is always searched first?

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -266,14 +268,56 @@ private void ShowWelcomeMessage(string altButtonTextKey, string url, int gotoMen
...
+		}
+		catch (Exception e) {
+			Logging.tML.Debug("Finding SystemDotnetPath failed: ", e);
+		}
+
+		return null;
+	}
+
+	bool IsCompatibleDotnetSdkAviable()
```
direwolf420:
```suggestion
	bool IsCompatibleDotnetSdkAvailable()
```

direwolf420: Same with where it's called

Fili-pk:
> Thanks for this. I'm assuming even though we haven't added our dotnet to `PATH`, the location of the current process (not even the working directory) is always searched first?

Yes, the location of current process is searched first.

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
Fili-pk: https://github.com/tModLoader/tModLoader/pull/3334/commits/478eeee7c8d34c016806e230afe17b446c54b481

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
Fili-pk: https://github.com/tModLoader/tModLoader/pull/3334/commits/d9c4ce39cbf5c12b4f9be13ccb13285c27f29338

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
Fili-pk: https://github.com/tModLoader/tModLoader/pull/3334/commits/a2c7311b5bc3bb5a03b708acef0ab80a2a953193

PR #3336, opened 2023-03-22
Fixed bug problem with opening folders on linux

Fili-pk:
What is the bug?
After clicking Open Sources button tmodloder opens browser.
image

How did you fix the bug?
I changed "open-folder" command to "xdg-open".

Probably you can use "gnome-open" command but it only work on gnome.

### DIFF
--- a/patches/tModLoader/Terraria/Utils.cs.patch
+++ b/patches/tModLoader/Terraria/Utils.cs.patch
@@ -34,6 +34,39 @@
  						if (num3 < 0)
  							num3 = 0;
  
+@@ -644,17 +_,32 @@
+ 			return false;
+ 		}
+ 	}
++	private static void OpenFolderXdg(string folderPath)
++	{
++		ProcessStartInfo processStartInfo = new ProcessStartInfo {
++			FileName = "xdg-open",
++			Arguments = folderPath,
++			UseShellExecute = true,
++			CreateNoWindow = true
++		};
++
++		processStartInfo.EnvironmentVariables["LD_LIBRARY_PATH"] = "/usr/lib:/lib";
++		Process.Start(processStartInfo);
++	}
+ 
+ 	public static void OpenFolder(string folderPath)
+ 	{
+ 		if (TryCreatingDirectory(folderPath)) {
+ 			if (Platform.IsLinux) {
++				/*
+ 				Process.Start(new ProcessStartInfo(folderPath) {
+ 					FileName = "open-folder",
+ 					Arguments = folderPath,
+ 					UseShellExecute = true,
+ 					CreateNoWindow = true
+ 				});
++				*/
++				OpenFolderXdg(folderPath);
+ 			}
+ 			else {
+ 				Process.Start(new ProcessStartInfo(folderPath) { UseShellExecute = true });
 @@ -669,6 +_,11 @@
  		return array;
  	}


JavidPack:
I know we've gone back and forth many times with the folder opening code in the past. I'm surprised this hasn't been reported, did this work on 1.4 stable?

Solxanich:
While I see it - yes, 1.4-stable is different than 1.4.4. Very curious if 1.4-stable works.

If not, this should be fixed at Terraria.Netcore patch level

Relevant stable code
image

Chicken-Bones:
Hi, this code comes straight from vanilla. The reason it doesn't work, is because we haven't copied the `open-folder` script from the vanilla Terraria directory.

We could easily include this script and call it `open-folder-linux`. Otherwise we could replace the `FileName` with `xdg-open` and alter the `LD_LIBRARY_PATH` env var for the process we're about to launch.

```
#!/bin/bash
export LD_LIBRARY_PATH="/usr/lib:/lib"
xdg-open $@
```

Fili-pk:
> Hi, this code comes straight from vanilla. The reason it doesn't work, is because we haven't copied the `open-folder` script from the vanilla Terraria directory.
> 
> We could easily include this script and call it `open-folder-linux`. Otherwise we could replace the `FileName` with `xdg-open` and alter the `LD_LIBRARY_PATH` env var for the process we're about to launch.
> 
> ```
> #!/bin/bash
> export LD_LIBRARY_PATH="/usr/lib:/lib"
> xdg-open $@
> ```

Ok, now it should create a process with environment variables set.

patches/tModLoader/Terraria/Utils.cs.patch
```diff
@@ -34,6 +50,49 @@
...
++		if (!TryCreatingDirectory(folderPath))
++			return;
++
++		if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
++			RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) {
++			Process.Start(new ProcessStartInfo(folderPath) { UseShellExecute = true });
++			return;
+ 		}
++
++		OpenFolderXdg(folderPath);
```
Chicken-Bones: Please try and minimise patch size. Comment out the vanilla `Process.Start` lines under the linux conditional, and insert `OpenFolderXdg(folderPath)` directly after the comment. 

https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#comment-out-vanilla-code-rather-than-removing-it

I've noticed in your other recent commit that you've been considering the `FreeBSD` platform. While I appreciate the forethinking, tML doesn't currently run on FreeBSD and we haven't had any requests for it, so it's better for maintainability to leave the structure of code as-is. 

`ReLogic.Platform.IsLinux` also covers FreeBSD at the moment

Chicken-Bones (changes requested):
Implementation looks great and is approved, just some feedback on code style! Let me know if it's not clear _exactly_ what I want, and I can make the changes directly, or we can discuss in Discord.

patches/tModLoader/Terraria/Utils.cs.patch
Fili-pk: Sorry for the formatting errors. I was not aware of these guidelines.

Chicken-Bones:
Thankyou for your contribution, and taking the time to learn the tModLoader process and style.

PR #3340, opened 2023-03-23
Attempt to fix #3035

### LINKED ISSUE #3035: Memory leak with repeated mod reloads
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

**Tested on both stable and preview, same behavior**

After about 15 (forced) mod reloads, the game starts to hang at the "Unloading tModLoader v2022.XXX" step for several seconds, with task manager showing drastic increases of memory usage on dotnet.exe, which presist after a mod reload.

### Log File

Nothing peculiar in the log.

### Steps to reproduce

1 Make sure enabled.json is empty, to eliminate mods causing this
2. open task manager to monitor dotnet.exe memory usage
3. start the game
4. go to workshop, mods, forced reload, repeat 15 times (or until you notice slowdowns/memory increase)
5. the game stays progressively longer on "Unloading tModLoader" step, with memory visibly increasing
6. This effect gets exponentially worse with subsequent reloads to the point where the game hangs for 30+ seconds and/or memory is in the 4+ GB

### Expected Behavior

Reloads should not cause a memory leak

### Actual Behavior

Reload causes a memory leak

### Reproduction frequency

100%.

### Additional Information

 This is not a clean/fresh computer so I have mods in my workshop/Mods folders, meaning they (try to) get displayed in the Mods menu where I go to every time I want to reload mods, maybe it has something to do with that. I have not tested it with 0 "existing" mods

After running the game through a blank mod in VS, and testing this, I found that this is the culprit, it keeps multiplying this count every reload:
Screenshot_6
Screenshot_7
Screenshot_8

Pannoniae:
### What is the bug?
(see #3035)
GlobalInstantiationArrayPool<T> leaks memory in registering TypeCaching.OnClear actions.
### How did you fix the bug?
This does not fully fix it since it doesn't solve the underlying problem of the ArrayPools not being reset but it turns the problem from a multiplying growth to a linear growth since only the rented memory is leaked.
LoaderUtils.ResetStaticMembers is not appropriate for resetting the ArrayPools since from my understanding, it creates them again but somehow keeps the references alive. This needs to be further investigated.
Yes, fixing the root issue. This is a mitigation against ~15 reloads making the game use gigabytes of leaked memory.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/LoaderUtils.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/LoaderUtils.cs
@@ -197,7 +197,6 @@ internal class GlobalInstantiationArrayPool<T>
 
 	static GlobalInstantiationArrayPool()
 	{
-		// honestly, this should go in 'OnResizeArrays'
-		TypeCaching.OnClear += () => LoaderUtils.ResetStaticMembers(typeof(GlobalInstantiationArrayPool<T>), false);
+
 	}
 }


Chicken-Bones:
Thanks for pinpointing the issue. I implemented a proper fix in 40eb3d9214cd3376cfd39f89f8ba894898e911fe

I used `ResetStaticMembers` thinking it was a nice way to de-duplicate the assignment of the `Pool` variable. Silly me didn't realise that would re-run the actual class constructor, therefore re-registering the `OnClear` callback.

Pannoniae:
Thank you very much!

PR #3343, opened 2023-03-24
More UI Fixes

Destructor-Ben:
### What is the bug?
Various small UI issues that haven't been fixed yet.
- `IngameOptions.cs` doesn't include the options for achievements and has a spacing issue.
- Info accessory display pages don't have localized tooltips
- Spelling mistake in "Steam Workshop page", should be "Steam Workshop Page"
- `Utils.DrawBorderStringBig()` has an incorrect spread parameter

Also added `UICommon.TooltipMouseText(string text)` for easy access to mouse text with a hover background seen in tooltips and used it in config element tooltips and the memory bar in the manage mods menu.

### How did you fix the bug?
`IngameOptions.cs` - Removed skipping the achievements option when drawing the menu
Info display pages - Added localization
Spelling mistake - Fixed it in lang files
`Utils.DrawBorderStringBig()` - Changed the spread parameter from 2f to 1.5f

No


patches/tModLoader/Terraria/ModLoader/Config/UI/UIModConfig.cs
```diff
@@ -395,7 +395,7 @@ public override void Draw(SpriteBatch spriteBatch)
 		base.Draw(spriteBatch);
 
 		if (!string.IsNullOrEmpty(Tooltip)) {
-			UICommon.DrawHoverStringInBounds(spriteBatch, Tooltip, GetDimensions().ToRectangle());
+			UICommon.MouseText(Tooltip, true);
```
Chicken-Bones: Always use named parameters (`hoverBackground: true`) when passing raw bool or integer values to methods where the meaning of the param is not obvious

patches/tModLoader/Terraria/ModLoader/UI/UICommon.cs
```diff
@@ -87,6 +88,25 @@ public static void DrawHoverStringInBounds(SpriteBatch spriteBatch, string text,
 		Terraria.UI.Chat.ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.MouseText.Value, text, vector, color, 0f, Vector2.Zero, Vector2.One);
 	}
 
+	public static void MouseText(string text, bool hoverBackground = false)
```
Chicken-Bones: Please add some simple documentation for this method.

Chicken-Bones: Is there any reason to use this method without passing `true` to `hoverBackground`? If so, are there places in tML that should use it? If not, should we just call this method `TooltipMouseText` or something? Given the relation to the `SettingsEnabled_OpaqueBoxBehindTooltips`

patches/tModLoader/Terraria/Utils.cs.patch
```diff
@@ -62,7 +62,7 @@
 +		*/
 +
 +		Vector2 vector = ChatManager.GetStringSize(value, text, Vector2.One);
-+		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, value, text, pos, color, 0f, new Vector2(anchorx, anchory) * vector, new Vector2(scale), -1f, 2f);
++		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, value, text, pos, color, 0f, new Vector2(anchorx, anchory) * vector, new Vector2(scale), -1f, 1.5f);
```
Chicken-Bones: Based on the commented code above, it looks like the `spread` parameter here should be `1f`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Utils.cs.patch
Destructor-Ben: I think I did try 1f at first, but the border looked way too thin. When I used hot reload to compare vanilla to modified code, 1.5f seemed to have no change. I still have no clue why it is like this.

patches/tModLoader/Terraria/ModLoader/UI/UICommon.cs
Destructor-Ben: I was more intending it to be for general MouseText, since I was planning on adding bounds and maximum width, but I didn't want to modify vanilla methods. I'll change the name and make it solely for MouseText with a background

patches/tModLoader/Terraria/Utils.cs.patch
Chicken-Bones: given that this is a pixel offset parameter for the shadow text in the 4 axial directions, and Terraria only uses `NEAREST` filtering, sub-pixel values have no effect. So it should be either `1` or `2`, vanilla uses `1`

Destructor-Ben: Vanilla
vanilla

Spread 1.5f
1 5 spread

Spread 2f
spread 2

Spread 1f
1 spread

PR #3344, opened 2023-03-24
Better Mod Development Menu

Destructor-Ben:
### What is the new feature?
Changes in the Mod Sources Menu
- Made the search bar take up the width of the UIPanel
- Made the UIPanel taller since previously it wasn't using all of the space it could have been
- Added more links, so there are now links for the discord server (specifically for #mod-programming-help), wiki, documentation, mod migration, and example mod.

Before
Old Meny

After
New meny

It will help new modders since it will give them all of the necessary links for getting help, and it also makes the UI slightly nicer to use

It is UI, so yes.

N/A

### ExampleMod updates
N/A


direwolf420:
A before/after would be appreciated

Chicken-Bones:
LGTM, assigning @JavidPack

Issue #3254, opened 2023-02-14
Damage Class name has an extra space between damage number and damage class name


JavidPack:
### Version

1.4.4-Preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description
From Nil256 and direwolf420 on Discord

`289  melee damage` 
should be 
`289 melee damage`

![2023-02-09_190707](https://user-images.githubusercontent.com/4522492/218824351-f9cf8ffd-740e-4e10-ad96-9c1b4bd6e096.png)

Note: The fix for this might be complicated, as these keys are used in many places, we'll want to check them all. We'll want to ensure compatibility for other languages too, if the word order is relevant.

PR #3349, opened 2023-03-26
Fixed double damage space

Fili-pk:
What is the bug?
https://github.com/tModLoader/tModLoader/issues/3254
The bug is caused by this code
```c#
item.DamageType.DisplayName
```
returning damage type name with an extra space at the beginning.

How did you fix the bug?
To fix this bug i simply added trimming to start of damage type name.

This problem may be probably solved by changing this line
```c#
tip = new LocalizedText("", " " + item.DamageType.DisplayName);
```
to
```c#
tip = new LocalizedText("", item.DamageType.DisplayName.ToString());
```
but for safety i used trim.

Other solution is to change 
```c#
item.DamageType.DisplayName
```
behavior to not include extra space at beginning.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2308,7 +2308,7 @@
  					if (diff == 1)
  						black = new Microsoft.Xna.Framework.Color((byte)((float)(int)mcColor.R * num4), (byte)((float)(int)mcColor.G * num4), (byte)((float)(int)mcColor.B * num4), a);
  
-@@ -16311,47 +_,90 @@
+@@ -16311,47 +_,91 @@
  
  					if (hoverItem.expert || rare == -12)
  						black = new Microsoft.Xna.Framework.Color((byte)((float)DiscoR * num4), (byte)((float)DiscoG * num4), (byte)((float)DiscoB * num4), a);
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Yeah, not sure whether it might be better to have the localization always trimmed. Including leading spaces in a localization is... misleading

Fili-pk:
> Yeah, not sure whether it might be better to have the localization always trimmed. Including leading spaces in a localization is... misleading

After analyzing differences between 1.4 and 1.4.4 branches i found that 1.4 code removes leading space.
```c#
internal protected override string DisplayNameInternal => Language.GetTextValue(LangKey).Substring(1);
```
And 1.4.4 code leaves space
```c#
public override LocalizedText DisplayName => Language.GetText(LangKey);
```

I can fix this problem by removing space in translation files.
Or by using old solution of removing space.
Or by changing this line
```c#
tip = new LocalizedText("", " " + item.DamageType.DisplayName);
```
to this
```c#
tip = new LocalizedText("", item.DamageType.DisplayName.ToString());
```

So which solution i should choose?

JavidPack:
Thanks for doing the research and detailing the options. I went with an approach of just handling vanilla classes separately. I don't want modders to have to add a space to hjson files, and I don't want to maintain patches for all the vanilla damage class translations, but I also don't want to prevent modders from using extra whitespace for whatever reason they might think of. 

Frankly, any solution would work, but I think what I ended up with after reviewing your implementation and suggestions is good and doesn't require mangling too much.

Issue #3299, opened 2023-03-08
[1.4.4] Catchable Town NPCs Are Not Friendly


Rijam:
### Version

Other (please specify elsewhere)

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

When setting a Town NPC to be catchable with `Main.npcCatchable[npc.type] = true;`, the Town NPC becomes not friendly. That means the player will take contact damage from them and their weapons will hurt the Town NPC. Other Town NPCs (who are not also catchable) will target and run from the catchable Town NPC because they see it as hostile. This happens with both modded and vanilla Town NPCs.

This bug was introduced very recently. I suspect it was introduced from #3246 as that is the only recent PR that has anything to do with catching.

Video demonstrating the bug. The Angler was set to be catchable in GlobalNPC. The reason the bug net didn't do anything is because a catchItem wasn't defined. The only mods enabled were Cheat Sheet, HERO's Mod, and the test mod with the change to the Angler.

https://user-images.githubusercontent.com/11262234/223593260-e2216bfc-4526-49f0-97ed-66fd10abf459.mp4

### Log File

I don't believe the logs have anything related to this, but here it is anyway.

[client.log](https://github.com/tModLoader/tModLoader/files/10915464/client.log)

### Steps to reproduce

Make a Town NPC catchable in SetDefaults. Code for the mod used in the video:
```cs
using Terraria;
using Terraria.ID;
using Terraria.ModLoader;

namespace HostileAngler
{
	public class HostileAngler : Mod
	{
	}

	public class GlobalNPCs : GlobalNPC
	{
		public override void SetDefaults(NPC npc)
		{
			if (npc.type == NPCID.Angler)
			{
				Main.npcCatchable[npc.type] = true;
			}
		}
		public override void GetChat(NPC npc, ref string chat)
		{
			// For debugging
			Main.NewText("[c/999999:NPC.friendly] " + npc.friendly + " [c/999999:Main.npcCatchable[Type]] " + Main.npcCatchable[npc.type] + " [c/999999:npc.catchItem] " + npc.catchItem + " [c/999999:NPCID.Sets.CountsAsCritter[Type]] " + NPCID.Sets.CountsAsCritter[npc.type] + " [c/999999:npc.townNPC] " + npc.townNPC);
		}
	}
}
```

Uploaded is the mod source code and compiled `tmod` file.
[...]

PR #3353, opened 2023-03-27
Fixes setting npcCatchable true forcing NPC.friendly to falce

Kogsey:
### What is the bug?
As stated in #3299, NPCs that have Main.npcCatchable[Type] = true are forcebly changed to NPC.friendly = false.

### How did you fix the bug?
The issue was a system for making the player immune to damage from a released enemy that would otherwise damage them setting every NPC to unfriendly after the timer. I'm pretty sure it's a vannila issue because the function is mostly the same there but there might be some other catch for it that was removed. fixed by modifying the system to only run when the npc was unfriendly to begin with, leving the NPC as is.

N/A

Another small note because I'm quite confused about it, I may have made un unintentional change? Or just fixed a bug that was so hidden by this bug?
I changed:
` 			if (releaseOwner != 255 || SpawnedFromStatue)`
to:
`+			if (releaseOwner == 255 || SpawnedFromStatue || catchableNPCTempImmunityCounter < 0)`

What I'm not sure about is releaseOwner != 255 in the original, this was the tag for not running the timer system (but still set friendly to false), which would be checking if the npc *was* released not if it wasn't, I'm not sure if I'm interpretting something wrong or if it's just the whole function is a mess.

Anyway, now it just works by waiting 90 frames where a released npc is friendly if it was not otherwise, which I'm pretty sure was the intention.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -364,7 +364,7 @@
  	public static bool[] tileFrameImportant = new bool[TileID.Count];
  	public static bool[] tilePile = new bool[TileID.Count];
  	public static bool[] tileBlendAll = new bool[TileID.Count];
-@@ -942,6 +_,12 @@
+@@ -942,6 +_,13 @@
  	public static int[] grasshopperCageFrameCounter = new int[cageFrames];
  	public static bool[] tileSand = new bool[TileID.Count];
  	public static bool[] tileFlame = new bool[TileID.Count];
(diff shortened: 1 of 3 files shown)


JavidPack:
Ok, this was a bit of a confusing trip, but I think what I ended up with fixed all the issues.

The original code had an issue with some vanilla critter that had friendly = true.

I asked around and player releases of critters should not enact the immunity, so that code was correct as it was originally written. 

Thanks a bunch for getting this worked on. One thing that will help in future PRs is to not refactor the vanilla code too much, your initial changes to the code layout made it really hard to identify the changes in logic from looking at the diff.

Thanks again.

PR #3355, opened 2023-03-28
Followup Improvements to StrikeNPC, Hit and Hurt hooks

Chicken-Bones:
### What is the new feature?

- Adds `SetMaxDamage` to `NPC.HitModifiers` and `Player.HurtModifiers` to allow limiting the maximum damage the calculation can produce.
- Adds `NPC.SimpleStrikeNPC` helper which combines the simple damage calculation, strike, and net syncing.
- Adds `NPC.StrikeInstantKill` helper method, to make instant killing NPCs easier. Mostly equivalent to the vanilla code `StrikeNPCNoInteraction(9999, 0, 0)` and also sends the packet in multiplayer.

```cs
// in AI
if (Main.netMode != NetmodeID.MultiplayerClient)
    npc.StrikeInstantKill();

// when an NPC takes damage from a custom effect
npc.SimpleStrikeNPC(damage, hitDirection);
```

If the effect comes from a `Player` (accessory etc), use `Player.ApplyDamageToNPC` or `Player.StrikeNPCDirect` instead.

### ExampleMod updates
None required

### Porting Notes
- `NPC.SimpleStrike` renamed to `NPC.CalculateHitInfo`

### DIFF
--- a/patches/tModLoader/Terraria/MessageBuffer.cs.patch
+++ b/patches/tModLoader/Terraria/MessageBuffer.cs.patch
@@ -308,7 +308,7 @@
 +						SourceDamage = reader.Read7BitEncodedInt(),
 +						DamageType = DamageClassLoader.DamageClasses[reader.Read7BitEncodedInt()],
 +						HitDirection = reader.ReadSByte(),
-+						KnockBack = reader.ReadSingle()
++						Knockback = reader.ReadSingle()
 +					};
 +					BitsByte flags = reader.ReadByte();
 +					hit.Crit = flags[0];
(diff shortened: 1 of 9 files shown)


ScalarVector1:
Is there a reason the netMode check should occur outside of StrikeInstantKill? Is there a case where you might want to call it without that check? (AKA can this check just be integrated into that method itself?)

Chicken-Bones:
> Is there a reason the netMode check should occur outside of StrikeInstantKill? Is there a case where you might want to call it without that check? (AKA can this check just be integrated into that method itself?)

This is answered in the documentation on the method. `Not recommended for use on multiplayer clients because the net packet will trigger <see cref="PlayerInteraction"/> on the server.`

The method is intended to be called from the server (or in single player). There is no way to avoid triggering `PlayerInteraction` if the instant kill originates from a 'client' without sending a different (custom) packet which executes the strike serverside.

PR #3362, opened 2023-03-29
Config Elements state if they required a reload

Destructor-Ben:
### What is the new feature?
Made config elements say in their tooltip if they require a reload. They also say it in their label if they have been changed

Currently it isn't obvious if changing a config element will require a reload, which can be annoying while in game, since it prevents saving.

It's UI, so yes.

N/A

### ExampleMod updates
N/A


JavidPack:
Note to myself, test config.: 
```cs
	public class AReloadTests : ModConfig
	{
		public override ConfigScope Mode => ConfigScope.ServerSide;

		[DefaultValue(true)]
		[ReloadRequired]
		public bool DefaultsToTrueRR;

		[ReloadRequired]
		public bool DefaultsToFalseRR;

		[ReloadRequired]
		public float FloatRR;

		//public float Float;

		[ReloadRequired]
		public Color ColorRR;

		//public Color Color;

		[ReloadRequired]
		public List<int> ListInitialRR = new List<int>() { 1, 3, 5 };
		//public List<int> ListInitial = new List<int>() { 2, 4, 6 };

		[ReloadRequired]
		public List<int> ListNullRR;
		//public List<int> ListNull;

		[ReloadRequired]
		public List<int> ListEmptyRR = new List<int>();
		//public List<int> ListEmpty = new List<int>();

		[ReloadRequired]
		public Pair pairRR = new Pair();
		[ReloadRequired]
		public List<Pair> pairListRR = new List<Pair>();

		public override bool NeedsReload(ModConfig pendingConfig) {
			return base.NeedsReload(pendingConfig);
		}
	}
```

PR #3372, opened 2023-04-01
Fix console commands on Docker servers

turtletowerz:
### What is the bug?
On Docker servers when running or attaching to a running server no console commands can be used, presumably because STDIN isn't being kept open.

### How did you fix the bug?
Added the equivalent of `-it` to the docker-compose file.

You can `docker run -it ...` to run the container with the same parameters but there's no other way for compose files to handle this.


Chicken-Bones:
I need to ask the original developers of the docker container what side-effects this may have. Servers running in completely uninteractive headless mode where the game is launched by a webpanel or cron job or similar may get stuck if `stdin` is connected.

ppebb:
This should have no side effects. As long as the container is launched with `-d` it is just as headless as before.

Issue #3316, opened 2023-03-16
DamageClass documentation


JavidPack:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Many methods relating to DamageClass are not well documented. Modders frequently misuse DamageClass and should be taught to use CountsAsClass when relevant. StatModifier.ApplyTo and Player.GetTotalDamage also need documentation that explains proper usage. 

If needed, a new wiki page could also explain the main concepts, but that can be done later as a separate effort.

### What does this proposal attempt to solve or improve?

Teach modders proper damageclass usage.

PR #3374, opened 2023-04-02
Updated some documentation

mtkloepf:
### What is the new feature?
Resolves issue  #3316 

Documentation updated for:
* DamageClass and all implementations of CountsAsClass( )
* StatModifier.ApplyTo( )
* GetTotalDamage( ) functions

Updates add more clarity for function usage - especially APIs important to modders

Documentation is always a subjective effort :)

n/a

### ExampleMod updates

### DIFF
--- a/patches/tModLoader/Terraria/Item.TML.cs
+++ b/patches/tModLoader/Terraria/Item.TML.cs
@@ -128,19 +128,15 @@ public partial class Item : TagSerializable, IEntityWithGlobals<GlobalItem>
 
 	public TagCompound SerializeData() => ItemIO.Save(this);
 
-	/// <summary>
-	/// This is used to check if the item has a specific <see cref="DamageClass"/>.
-	/// </summary>
-	/// <seealso cref="CountsAsClass(DamageClass)"/>
+	/// <inheritdoc cref="CountsAsClass"/>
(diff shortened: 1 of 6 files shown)


noctisatrae:
> Documentation is always a subjective effort :)

Love this ❤️! Good PR, I think you're being pretty clear stating what you want to do!  I'm also rewriting a very specific part of the documentation, could you review it at #3378 ?

mtkloepf:
@JavidPack Please review at your leisure.

JavidPack:
Looks good, thanks. In the future, `<seealso>` and `<example>` are not useful, since they don't show up in the intellisense box.

mtkloepf:
> Looks good, thanks. In the future, `<seealso>` and `<example>` are not useful, since they don't show up in the intellisense box.

I suppose there's an argument for using `<see>` instead, but example and SA will show in the online documentation if we upload the new version. That was my main target. Good to know your thoughts though.

JavidPack:
Our online documentation will probably do something with them, but that's definitely not where our modders would be looking for relevant information.

noctisatrae:
:D I look at it a lot actually to see the arguments needed for a function & class.

JavidPack:
Yes, it is an option, but the expectation is your IDE should be showing you all the information already.

PR #3378, opened 2023-04-03
Improving the documentation (Recipe.Create)

noctisatrae:
I recently saw in the documentation that it was not documented even though it is commonly used. However, I'm not sure I have been clear enough and that I understood what stack does. Can you help me? 

This PR has been motivated by a *[discussion on Terraria's forums](https://forums.terraria.org/index.php?threads/how-to-set-quantity-of-items-for-recipe.118289/#post-2920946)* ! Looking forward to hear back from y'all!

Note: I would have used a PR template but couldn't find one suitable for documentation issues. Maybe something to look into ;) !

### DIFF
--- a/patches/tModLoader/Terraria/Recipe.TML.cs
+++ b/patches/tModLoader/Terraria/Recipe.TML.cs
@@ -10,6 +10,13 @@ using Terraria.ModLoader.Exceptions;
 
 namespace Terraria;
 
+/// <summary>
+/// A Recipe is a collection of ingredients, tiles, and a resulting Item. This is how players can craft items in the game.<br/>
+/// The <see href="https://github.com/tModLoader/tModLoader/wiki/Basic-Recipes">Basic Recipes Guide</see> teaches how to add new recipes to the game and how to manipulate existing recipes.<br/>
+/// Use <see cref="Recipe.Create(int, int)"/> to create a Recipe instance resulting in the specified item. Use <see cref="AddIngredient(int, int)"/> to add ingredients and <see cref="AddTile(int)"/> to add crafting stations. Finally, use <see cref="Register"/> to complete the recipe and register it to the game.<br/>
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -460,6 +460,12 @@ public Recipe Register()
 		return this;
 	}
 
+	/// <summary>
+	/// Create the recipe but does not yet register it into the game. Call this at the very beginning when creating a new craft.
+	/// </summary>
+	/// <param name="result">What item will be given when the craft has been completed</param>
```
mtkloepf, line 473: I think it might be important to note here that the result variable is actually indicating the ItemID of the "resulting" item of the recipe.

To expand on that a bit: the name result is confusing because it is not a return result or a calculation. I think a rename to "itemID" or something similar is warranted to not confuse developers.

@Chicken-Bones @JavidPack any thoughts?

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -460,6 +460,12 @@ public Recipe Register()
...
+	/// <param name="stack">The stack -> how many result items given when the craft is completed. (eg. 1 wood -> 4 wood platform)</param>
```
mtkloepf: The param name is actually "amount" here. It could also be worth pointing out the default of 1, so you only need to pass a value if you want something other than 1 of the item per craft.

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -460,6 +460,12 @@ public Recipe Register()
...
+	/// <exception cref="RecipeException">A Recipe can only be created inside recipe related methods</exception>
 	public static Recipe Create(int result, int amount = 1)
```
mtkloepf, line 476: You can probably add a "returns" tag and fill that out too.

`<returns> </returns>`

mtkloepf:
You were pretty on point about the "stack." It is indeed how many are created per craft of the item. Good updates!

noctisatrae:
Yay! First commit here :D !

Issue #3384, opened 2023-04-04
Field to control if an item shooting turns the player


Destructor-Ben:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Adding a field to items something along the lines of `shootingTurnsPlayer` which is used in `Player.ItemCheck_Shoot`.

### What does this proposal attempt to solve or improve?

Currently you have to use IL editing to prevent the player from facing a projectile when it is shot, and the vanilla behaviour that prevents this is hardcoded to the Grand Design and Beam Sword.
Vanilla Code:
```cs
bool flag = true;
int type = sItem.type;
if (type == 723 || type == 3611)
    flag = false;

Vector2 value = Vector2.UnitX.RotatedBy(fullRotation);
Vector2 vector = Main.MouseWorld - pointPoisition;
Vector2 v = itemRotation.ToRotationVector2() * direction;
if (sItem.type == 3852 && !ItemAnimationJustStarted)
    vector = (v.ToRotation() + fullRotation).ToRotationVector2();

if (vector != Vector2.Zero)
    vector.Normalize();

float num = Vector2.Dot(value, vector);
if (flag) {
    if (num > 0f)
        ChangeDir(1);
    else
        ChangeDir(-1);
}
```

```cs
if (type == 723 || type == 3611)
    flag = false;
```
Should be changed to
```cs
if (sItem.shootingTurnsPlayer)
    flag = false;
```

Chicken-Bones:
Normally I advocate for `ItemID.Sets` entries, but I think this makes sense as an `Item` field.

Destructor-Ben:
Close?

PR #3385, opened 2023-04-05
Added field for items to not change player direction on shoot

Destructor-Ben:
https://github.com/tModLoader/tModLoader/issues/3384

### What is the new feature?
Added a field to items that can allow items to prevent shooting a projectile make the player face the projectile
`item.ChangePlayerDirectionOnShoot`

Currently IL editing is required to prevent a shot projectile from making the player turn towards it, since the vanilla implementation that excludes the Beam Sword and Grand Design from this behaviour is hardcoded.

Using a set in ItemID.Sets

In ModItem.SetDefaults
```cs
public override void SetDefaults()
{
    item.ChangePlayerDirectionOnShoot = false;
}
```

### ExampleMod updates
I will do one if requested

### DIFF
--- a/ExampleMod/Content/Items/Weapons/ExampleShootingSword.cs
+++ b/ExampleMod/Content/Items/Weapons/ExampleShootingSword.cs
@@ -36,6 +36,9 @@ namespace ExampleMod.Content.Items.Weapons
 
 			// If you want melee speed to only affect the swing speed of the weapon and not the shoot speed (not recommended)
 			// Item.attackSpeedOnlyAffectsWeaponAnimation = true;
+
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Yes, an example is required. Please carefully consider naming and documentation. Can you explain how this is related to and differs from `useTurn`?

Ideally the default value of the member would be `false`, if there is no way to have a sensible name, then make sure you put an `= true` after the property declaration, and that you specify the default in the docs.

`ChangePlayerDirectionOnShoot = false` additions to `SetDefaults` for vanilla items should have a `// Added by TML` appended to the end, just so it's clear to a reader.

PR #3390, opened 2023-04-08
Wrap many ModPlayer hooks in exception suppressors to prevent cascading failures

Ouroel:
### What is the bug?

Because `Player.Update` is wrapped in an exception supressing `try/catch`, it is often the case that a mod throwing an exception at an earlier point during player updating will cause half the update method to silently not run, and users will perceive it as an error with a mod with hooks later in the method (because they didn't look at logs).

As an example
> An exception thrown in `ProcessTriggers` will cause minions to despawn as `Player.slotsMinions` is not reset.

### How did you fix the bug?

Placed many hooks in a `try { ... } catch { }`

We could wrap more hooks in handlers, less hooks in handlers, or no hooks in handlers.
Here I (@Chicken-Bones) have tried to minimise excessive boilerplate and potential performance loss from exception handling in every hook, by focusing only on methods which are called every frame, and which are likely to have cascading effects.

I haven't bothered with eg `Shoot` hooks or item update hooks, since it's reasonable to expect people will just look at the logs for those, or realise that it occurs with every item not just one. There's always the risk that users say 'This item from Calamity is broken' when it's not Calamity's fault for eg, but I'm not convinced we should `try/catch` every hook ever.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
@@ -672,12 +672,7 @@ public static class PlayerLoader
 	{
 		foreach (var modPlayer in HookProcessTriggers.Enumerate(player))
 		{
-			try
-			{
-				modPlayer.ProcessTriggers(triggersSet);
-			}
-			catch (Exception e)
-			{ }
+			try { modPlayer.ProcessTriggers(triggersSet); }	catch (Exception e)	{ }
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
The implementation here is fine, though please check the patch files and avoid adding newlines where not necessary.

I'm concerned about setting a precedent where _every_ `ModPlayer` hook needs a `try catch` around it. Maybe that wouldn't be a bad idea... but given this PR focuses on `ProcessTriggers`, can you give me a reason why exceptions are much more likely to be thrown in this method as opposed to any others?

Chicken-Bones:
Bumping this since it's been over a week.

Ouroel:
I just made this PR because I'm tired of people whining about errors in ProcessTriggers despawning minions and then the player blaming Summoner's Shine for it.

I can probably trycatch more stuff but I haven't found the need to do so.

PR #3392, opened 2023-04-09
BuyItem and PreReforge split into more granular methods, Fixed savings in void bag not being considered when reforging

Jadams505:
### What is the bug?
Coins in the void bag would not be considered when reforging
### How did you fix the bug?
Count the void bag coins in `Player.CanBuyItem`
It might be advantageous to make the method internal or remove the `customCurrency` parameter because it calls 
`CustomCurrencyManager.BuyItem` for custom currencies which consumes items


Chicken-Bones (approved):
Perfect, great msthod splits

PR #3399, opened 2023-04-10
Allows the fusion of multiple entity sources

Ouroel:
### What is the new feature?

Allowing mods to combine entity sources into one entity source. This is useful for combining multiple mod-specific entity sources together, which is useful for feeding in spawn-specific data from and to two different mods.

There is no current way to combine two entity sources together, which means there is no way to feed entity source data to two different mods at once. For instance, Summoner's Shine minions are spawned with IEntitySource data read to determine attack speed and crit values, but Calamity's minion respawn minions are spawned with special IEntitySource data which cannot be read by any other mods - hence Calamity's minion respawn feature does not load Summoner's Shine data.

This allows mods who desire better cross-mod support to combine two entity sources together in order for different mods to load them individually in their OnSpawn hooks.

However, existing mods have to adopt this new feature manually if they desire to benefit from it.

This is the most streamlined implementation I can think of.

[Your IEntitySource]
.AddEntitySource([Second Entity Source])
.AddEntitySource([Third Entity Source]); //...etc

[Desired IEntitySource] Variable = [Input IEntitySource].AttemptGet<[Desired IEntitySource]>(); //will be null if it failed or isn't a collection. will return the desired IEntitySource if it's already the correct IEntitySource or the correct IEntitySource is in the collection.

### ExampleMod updates
None

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/CombinedEntitySource.cs
+++ b/patches/tModLoader/Terraria/ModLoader/CombinedEntitySource.cs
@@ -8,13 +8,13 @@ public class CombinedEntitySource : IEntitySource
     public string Context => "tModLoader_CombinedEntitySource";
     List<IEntitySource> entitySources = new();
 
-    public EntitySourceTest AddEntitySource(IEntitySource second)
+    public CombinedEntitySource AddEntitySource(IEntitySource second)
     {
         entitySources.Add(second);
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
The work in #3500 should make entity sources better, and reduce the _need_ for combining sources. Where it is required, interfaces are the way to go, and some have been provided

PR #3403, opened 2023-04-11
ExampleMod Example for a Custom Swing Sword

Midnight417:
- added `ExampleMod/Content/Items/Weapons/ExampleCustomSwingSword.cs` and a corresponding `ExampleMod/Content/Projectiles/ExampleCustomSwingSword.cs`, which serves as an example of a sword weapon with a custom swinging animation made via a held projectile

Most probably not


JavidPack:
Great job.

JavidPack:
Just so you know, in case you were using a similar projectile in one of your mods, there is a sync issue with projectile.spritedirection, it is fixed in https://github.com/tModLoader/tModLoader/commit/ba3750cbb8b0ab0675e1e8ed123c618edbc26362

PR #3408, opened 2023-04-13
Fix AnyWeapon mod prefixes not being loaded on items

Jadams505:
### What is the bug?
Modded prefixes that apply to weapons of any type would fail to load (when joining back into a world) onto items that only accept one damage type.

Reproduction: 
1) Obtain an item with a prefix category of `AnyWeapon` (I used ExamplePrefix on a copper shortsword)
2) Leave the world
3) Join back into the world and the prefix will be missing
### How did you fix the bug?
Check for this case in `PrefixLoader.CanRoll`
No


Chicken-Bones:
I think the logic in `CanRoll` is just backwards. `AnyWeapon` means the prefix can go on any weapon yes, but `AnyWeapon` for an item means the item _only_ accepts prefixes which can go on `AnyWeapon` and not class-specific ones

PR #3414, opened 2023-04-16
Re-order Program startup so error messages are relevant

Solxanich:
Currently, errors in logging startup are shown as "Failed to Handle Save Data", which is really misleading.
As well, FNALogging also included a dependency check buried inside it, which would make the problem worse.

This PR moves the FAUDIO dep check in to a separately wrapped method, handles the BadFormat error with another Catch, and moves Logging.Startup to occur after HighDPI environment changes,

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Engine/NativeLibraries.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Engine/NativeLibraries.cs
@@ -21,10 +21,15 @@ internal class NativeLibraries
 
 		try {
 			NativeLibrary.Load("mfplat.dll", Assembly.GetExecutingAssembly(), DllImportSearchPath.System32);
+			throw new BadImageFormatException();
 		}
 		catch (DllNotFoundException e) {
 			e.HelpLink = "https://support.microsoft.com/en-us/topic/media-feature-pack-list-for-windows-n-editions-c1c6fffa-d052-8338-7a79-a4bb980a700a";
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Why should log startup run after `AttemptSupportHighDPI`?

A significant issue I have with this is that logging init should be as early as possible, so as much information as possible actually ends up in the logs.

I'm happy for `LogStartup` to be separated from the save directory search, this is good, but `CheckDependencies` and fna version logging should run after `LogStartup`

patches/tModLoader/Terraria/ModLoader/Engine/NativeLibraries.cs
```diff
@@ -21,10 +24,15 @@ internal static void CheckNativeFAudioDependencies()
 
 		try {
 			NativeLibrary.Load("mfplat.dll", Assembly.GetExecutingAssembly(), DllImportSearchPath.System32);
+			throw new BadImageFormatException();
```
Chicken-Bones, line 27: leftover test code, game is unlaunchable

Solxanich:
AttemptSupportHighDPI -> Changes Environment variables and we don't currently update environment.log when that is done. Arguably simplest was to have it run before LogStartup, but I can see an argument going the other way. 

Logging Init -> we don't change the init. Ionly change the Log startup.
Startup log being client.log OS info being most pertinent information that would be lost if the CheckDependencies fails. I definitely changed the order when shouldn't have there.

FNA version logging runs at the tailend of Log Startup. That's unchanged.

Definitely got trigger happy here

PR #3415, opened 2023-04-16
Various NPC Gravity Changes

Kogsey:
### What is the new feature?
Allows modders to affect NPC gravity and max fall speed without interfering with other mods, properties to use are: `GravityMultiplier` and `MaxFallSpeedMultiplier`

New public stuff in `NPC` for modders to play with:
- `gravity`: Previously private field now a readonly property.
- `maxFallSpeed`: The counterpart to `NPC.gravity` for fall speed.
- `GravityMultiplier`: New property to be used to modify gravity. Made a `MultipliableFloat` to prevent "haha `NPC.gravity = 0.3f`" ruining mod compatibility
- `MaxFallSpeedMultiplier`: The as for gravity, so for max fall speed.
- `GravityIgnoresType`: Stops the NPC changing its gravity due to either it's `type` or `aiStyle`
- `GravityIgnoresSpace`: Stops the NPC changing its gravity due to space.
- `GravityIgnoresLiquid`: Stops the NPC changing its gravity due to liquids.
- Two public readonly `float[]` `GravityWetMultipliers` and `MaxFallSpeedWetMultipliers`, give in code values for the changes liquids make to gravity.

Private changes in NPC.tml.cs:
- `vanillaGravity` and `vanillaMaxFallSpeed` used by `gravity` and `maxFallSpeed` alongside `GravityMultiplier` and `MaxFallSpeedMultiplier` to combine vanilla accessor logic and new logic

Private changes in NPC.cs:
- Moved `UpdateNPC_UpdateGravity()` to be called before buff updates, so gravity can be modified in buff code without being overriden later
- Multiple changes to `UpdateNPC_UpdateGravity()` to support public variables

The current ways to modify NPC gravity involve reflection, IL editing or manually adjusting `velocity`. They are not cross-mod compatible.

`gravity` and `maxFallSpeed` don't need to be public, but it can make it easier for modders to understand what's going on, or implemented special behavior.

Debuffs that affect gravity, modifying vanilla NPC's interactions with space, bosses that feel gravity more dynamically during a fight or affect player gravity based on distance (especially interesting might explore in my mod).

Increased control of liquid gravity means more customisability in enemies' water/lava/honey/shimmer interactions, like making crabs or water enemies "swim" better without much hassle

### ExampleMod updates
- Added logic in `PartyZombie.cs` that makes the NPC fall like water in honey and vice versa
- Added, a buff `ExampleGravityDebuff` and applying bullet with projectile `ExampleGravityDebuffBullet` and `ExampleGravityDebuffBullet` that show off simple gravity affecting behaviour


ExampleMod/Content/NPCs/PartyZombie.cs
```diff
@@ -68,6 +68,19 @@ public class PartyZombie : ModNPC
...
 		}
 
+		public override void AI() {
+			if (NPC.wet) {
+				if (NPC.honeyWet) { // Removes the effects of honey's fall rate making the NPC fall normally in honey
+					NPC.Gravity /= NPC.GravityWetMultipliers[1];
+					NPC.MaxFallSpeed /= NPC.MaxFallSpeedWetMultipliers[1];
+				}
+				else { // Removes water falls speed effects, then adds honey falls speed effects, making the NPC fall at the honey rate in water
+					NPC.Gravity *= NPC.GravityWetMultipliers[1] / NPC.GravityWetMultipliers[0];
```
Chicken-Bones: These should use `LiquidID` references. It also seems like you have honey as 1 when it should be 2 (might want to re-order your `GravityWetMultipliers` to match `LiquidID` order

ExampleMod/Content/Projectiles/ExampleGravityDebuffBullet.cs
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		public override void Kill(int timeLeft) {
+			// This code and the similar code above in OnTileCollide spawn dust from the tiles collided with. SoundID.Item10 is the bounce sound you hear.
+			Collision.HitTiles(Projectile.position + Projectile.velocity, Projectile.velocity, Projectile.width, Projectile.height);
+			SoundEngine.PlaySound(SoundID.Item10, Projectile.position);
+		}
+
+		public override void OnHitNPC(NPC target, NPC.HitInfo hit, int damageDone) {
+			target.AddBuff(ModContent.BuffType<ExampleGravityDebuff>(), 1800);
```
Chicken-Bones: maybe 10 seconds would be more reasonable

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -51,8 +51,84 @@
...
++
++	/// <summary>
++	/// Used to keep vanilla and modded gravity effects working neatly
++	/// </summary>
++	private float internalGravityBase = 0.3f;
++
++	/// <summary>
++	/// Modify this value in order to change the NPCs active gravity, this can be done in AI as gravity values are reset slightly beforehand, and used slightly after.
++	/// </summary>
++	public MultipliableFloat Gravity = MultipliableFloat.One;//TML: #AccessibleNPCGravity, added
```
Chicken-Bones: Having both `Gravity` and `gravity` could be confusing. Suggestions:
- Call this `GravityMultiplier`
or
- make `gravity` an internal wrapper around `Gravity.Value`

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -51,8 +51,84 @@
...
++
++	/// <summary>
++	/// Used to keep vanilla and modded gravity effects working neatly
++	/// </summary>
++	private float internalMaxFallSpeedBase = 10f; //TML: #AccessibleNPCGravity, added
++
++	/// <summary>
++	/// Modify this value in order to change the NPCs active maxFallSpeed, this can be done in AI as gravity values are reset slightly beforehand
++	/// </summary>
++	public MultipliableFloat MaxFallSpeed = MultipliableFloat.One; //TML: #AccessibleNPCGravity, added
```
Chicken-Bones: Same feedback as above for `Gravity` and `gravity`

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -51,8 +51,84 @@
...
++	/// <br> These are mutually exclusive and in reverse order priority where lava is the same as water. </br>
++	///	</summary>
++	public static float[] GravityWetMultipliers => new float[] { 2f / 3f, 1f / 3f, 1.5f / 3f }; //TML: #AccessibleNPCGravity, added
++
++	/// <summary>
++	/// The effect of the types of wet on the NPCs maxFallSpeed, here for reference when modifying maxFallSpeed:
++	/// <br> wet, honetWet, and shimmerWet.</br>
++	/// <br> These are mutually exclusive and in reverse order priority where lava is the same as water. </br>
++	///	</summary>
++	public static float[] MaxFallSpeedWetMultipliers => new float[] { 0.7f, 0.4f, 0.55f }; //TML: #AccessibleNPCGravity, added
```
Chicken-Bones: As noted earlier, I think these should line up with `LiquidID`.

Also, I know you don't want people to modify them, but eh, allocating a new array on every access is not the way to go. Better to just advise mods not to modify. You could consider making the field type `RefReadOnlyArray<float>` which will prevent easy modification

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -2261,9 +2340,18 @@
...
 +		NPCLoader.ResetEffects(this);
++		UpdateNPC_UpdateGravity();
  		UpdateNPC_BuffSetFlags();
  		UpdateNPC_SoulDrainDebuff();
  		UpdateNPC_BuffClearExpiredBuffs();
+ 		UpdateNPC_BuffApplyDOTs();
+ 		UpdateNPC_BuffApplyVFX();
+ 		UpdateNPC_BloodMoonTransformations();
+-		UpdateNPC_UpdateGravity(out var maxFallSpeed);
++		//UpdateNPC_UpdateGravity(out var maxFallSpeed) // TML: #AccessibleNPCGravity, removed out parameter
```
Chicken-Bones: Use block comment around the line, and note that it was moved to run with `NPCLoader.ResetEffects`

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -51,8 +51,84 @@
...
++	/// </summary>
++	public bool GravityIgnoresType = false; //TML: #AccessibleNPCGravity, added
++
++	/// <summary>
++	/// Set true for the NPC's gravity to be unaffected by space's lowered gravity.
++	/// </summary>
++	public bool GravityIgnoresSpace = false; //TML: #AccessibleNPCGravity, added
++
++	/// <summary>
++	/// Set true for the NPC's gravity to be unaffected by liquid's gravity.
```
Chicken-Bones: Remove the final `gravity`. Liquid doesn't really have gravity. Feel free to explain it some other way (viscosity, slowdown, etc)

Chicken-Bones (changes requested):
Love the new example bullet. Implementation is on the right track. Feedback is about making sure the new API is easy to use and doesn't have 'gotchas'

patches/tModLoader/Terraria/NPC.cs.patch
Kogsey: How about this?
```
private static float[] _maxFallSpeedWetMultipliers = new float[4] { 0.7f, 0.7f, 0.4f, 0.55f };
public static float[] MaxFallSpeedWetMultipliers => _maxFallSpeedWetMultipliers;
```

patches/tModLoader/Terraria/NPC.cs.patch
Kogsey: Changed to GravityMultiplier

ExampleMod/Content/NPCs/PartyZombie.cs
Kogsey: Added lava into the array, noted that in code they are treated the same in the comment, now work with LiquidID, and changed the example

Kogsey:
Commit includes proposed:
```
private static float[] _maxFallSpeedWetMultipliers = new float[4] { 0.7f, 0.7f, 0.4f, 0.55f };
public static float[] MaxFallSpeedWetMultipliers => _maxFallSpeedWetMultipliers;
```
just to get something in before work

patches/tModLoader/Terraria/NPC.cs.patch
Chicken-Bones: Hm, lets do 
```cs
public static float[] MaxFallSpeedWetMultipliers { get; } = new float[] { ... };
```

Kogsey: Yeah, dunno why I wrote in the backing fields manually, this is why I shouldn't code when dead tired 🤦‍♂️

Kogsey: Fixed in commit

Chicken-Bones (approved):

PR #3418, opened 2023-04-17
Correct LogStartup order & Int.MaxValue UnloadedItem

Solxanich:
Correct LogStartup order & Int.MaxValue UnloadedItem
From #3414 & ea0150d5e5609b7413d119c6c56ab9412964c453

Notes:
Probably should refactor all these custom calls and try/catch in to a Program.StartupTmlSequence method and reduce patch size. 
Could also move the DeveloperMode check as well then so that Startup logging is earliest in chain

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -127,7 +127,7 @@ public static partial class Program
 			Logging.Init(isServer ? Logging.LogFile.Server : Logging.LogFile.Client);
 
 			if (Platform.Current.Type == PlatformType.Windows && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X64)
-				throw new Exception("The current Windows Architecture of your System is CURRENTLY unsupported. Aborting...");
+				ErrorReporting.FatalExit("The current Windows Architecture of your System is CURRENTLY unsupported. Aborting...");
 		}
 		catch (Exception e) {
 			ErrorReporting.FatalExit("Failed to init logging", e);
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
It is my vague recollection that the reason for the dep check being right before the FAudio version log is because you can't actually load the `FAudio` dll to get the version from it if those deps aren't present. This means that we need to split the FNA logging setup from the rest of the logging, (or just put it back the way it was)

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -120,6 +120,44 @@ private static void SetSavePath()
...
 	}
 
+	private static void StartupSequenceTml(bool isServer)
+	{
+		try {
+			ControlledFolderAccessSupport.CheckFileSystemAccess();
+			Logging.Init(isServer ? Logging.LogFile.Server : Logging.LogFile.Client);
+
+			if (Platform.Current.Type == PlatformType.Windows && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X64)
+				throw new Exception("The current Windows Architecture of your System is CURRENTLY unsupported. Aborting...");
```
Chicken-Bones: Can just be a `FatalExit`, rather than exception that's immediately caught. We could probably split `LogStartup` to at least add env info to the logs, but probably not necessary, this is fine.

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -120,6 +120,44 @@ private static void SetSavePath()
...
+		Logging.LogStartup(isServer); // Should run as early as is possible. Want as complete a log file as possible
+
+		try {
+			SetSavePath(); 
+		}
+		catch (Exception e) {
+			ErrorReporting.FatalExit("Failed to establish a save location", e);
+		}
+				
+		if (ModLoader.Core.ModCompile.DeveloperMode) // Needs to run after SetSavePath
```
Chicken-Bones: Not just needs to run after `SetSavePath`, needs to have static ctor called after `SetSavePath`, so might need to check that via breakpoint

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -120,6 +120,44 @@ private static void SetSavePath()
...
+			Logging.tML.Info("Developer mode enabled");
+
+		AttemptSupportHighDPI(isServer); // Can run anytime
+
+		try {
+			CheckDependencies(); // Should run after LogStartup
```
Chicken-Bones: May as well just call `NativeLibraries.CheckNativeFAudioDependencies`, no need for a springboard method. Honestly, no need to run that on server either, so just batch it with `FNALogging.RedirectLogs`

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -120,6 +120,44 @@ private static void SetSavePath()
...
+		}
+		catch (Exception e) {
+			ErrorReporting.FatalExit("Unexpected failure in verifying dependencies. Please reach out in the tModLoader Discord for support", e);
```
Chicken-Bones: No need for `try catch` here, should never be hit, the handlers are all in `CheckNativeFAudioDependencies`

Chicken-Bones: If you still want a fallback handler, a blanket `try-catch` around the entire `StartupSequenceTml` method would be more appropriate

Chicken-Bones (changes requested):
Happy with this direction, only a few comments

patches/tModLoader/Terraria/Program.TML.cs
Solxanich: Tested and validated that the ModComiple static ctor was only invoked by this in LogStartup. 
This ordering works ok.

Chicken-Bones (approved):

Issue #3293, opened 2023-03-07
Better Solutions


BasicallyIAmFox:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Adds a new ModType that is called ModConversion. As name says, it allows to add custom conversions. Maybe there also could be GlobalConversion or something like that for modifying existing conversions. This could also add optimized TileScanner class, which as name says, would scan nearby tiles.
Code for custom conversion types would look like that in practice:
```
using Terraria.ID;
using Terraria.ModLoader;
using TheConfectionRebirth.Tiles;
using TheConfectionRebirth.Walls;

public class ConfectionConversion : ModConversion {
    public override void SetStaticDefaults() {
        Conversion
            // Replace moss and all stone convertibles into Creamstone
            .From(Main.tileMoss)
            .From(TileID.Sets.Conversion.Stone)
            .To<Creamstone>()
            .OnConversion(TryKillingTreesAboveIfTheyWouldBecomeInvalid)
            .RegisterTile()
            
            .From(TileID.Sets.Conversion.Ice)
            .To<BlueIce>()
            .OnConversion(TryKillingTreesAboveIfTheyWouldBecomeInvalid)
            .RegisterTile()
            
            .From(TileID.Mud)
            .To<CookieBlock>()
            .BeforeConversion((Tile tile, int i, int j) => {
                if (TileScanner.ScanCircle((tile, i, j) => tile.TileType == ModContent.TileType<CreamGrass>(), i, j, 1)) {
                    return ConversionRunCodeValues.Run;
                }
                // Can't convert if no cream grass nearby!
                return ConversionRunCodeValues.DontRun;
            })
            .RegisterTile()
            
            // Break all thorns
            .From(TileID.Sets.Conversion.Thorn)
            .To(ConversionHandler.Break)
            .RegisterTile()
            
            // Make all grass walls be convertible into cream grass wall
            .From(WallID.Sets.Conversion.Grass)
            .To<CreamGrassWall>()
            .RegisterWall();
    }
}
```
And they would call conversion tile to other tile like that: `WorldGen.Convert<ConfectionConversion>(i, j)` - to convert one tile respectively. Putting a number after coordinates will convert all tiles in X radius using the TileScanner.

### What does this proposal attempt to solve or improve?

[...]

PR #3428, opened 2023-04-23
Better Solutions

BasicallyIAmFox:
#3293 

### What is the new feature?
Adds new class to be used by modders: `Conversion`.
This attempts to increase readability as well as simplify all the code used by conversions/solutions. (possibly even do some optimizations!)

Currently every single conversion type is just a static method called "Convert" in their custom solution projectile with generally too much repeative code (just like vanilla, I guess) and barely any readability.

To Be Discussed.

```
var conversion = new Conversion(FullName);

// Turn all dirt to stone
conversion.ConvertTile(TIleID.Dirt, TileID.Stone).OnConversion(TryKillingTreesAboveIfTheyWouldBecomeInvalid);

conversion.Register();
```

### ExampleMod updates
`ExampleSolution` will be update it to use this new feature.


ilija04:
ExampleSolution does exist, and it's right here: https://github.com/tModLoader/tModLoader/blob/1.4.4/ExampleMod/Content/Items/Ammo/ExampleSolution.cs

BasicallyIAmFox:
Alright, PR is fully functional I believe. Should be ready for reviews.

BasicallyIAmFox:
Nevermind, this PR slows down hardmode biome gen infinitely.

Ouroel:
> Nevermind, this PR slows down hardmode biome gen infinitely.

It doesn't anymore I believe

PR #3433, opened 2023-04-24
ModHair API

steviegt6:
### What is the new feature?

Adds `ModHair`, an API for registering custom player hairstyles.

This is not easily doable without the assistance of an API. Attempting to do so by yourself will inevitably clash with others' implementations.

Not for the basic features added by this PR.

https://github.com/steviegt6/aomoji-vanity/blob/dev/tml-1.4.4-modhair/src/AomojiVanity/Content/Hair/HairAutoloader.cs


JavidPack:
I took a quick look, seems to save and restore fine. I'm thinking IsMale should be abstract, so the modder has to implement it. There is also an issue with the player instance in the main menu accessing modded biomes in Conditions, it throws an error because I think that player doesn't have the ModPlayers on it for some reason. I didn't have time to fully investigate, might be a simple fix or something really complex.

The condition added isn't useful at all, since it combines the 2 most commonly used conditions, chicken bones might be able to tell you exactly how those should be split up and named.

steviegt6:
The condition was combined since in vanilla it's *either*, which is what most modders would generally care about. Just realized I forgot to add a localization key for it anyway.

I'll rectify these tomorrow; out sick today.

patches/tModLoader/Terraria/ModLoader/ModHair.cs
```diff
@@ -0,0 +1,53 @@
...
+
+	/// <summary>
+	/// The path to the alternative texture used for when the hair is covered by a hat.
+	/// </summary>
+	public virtual string AltTexture => Texture + "_Alt";
+
+	/// <summary>
+	/// Determines whether this hairstyle is considered male, which influences what gender is picked when randomizing a character.
+	/// </summary>
+	public virtual bool IsMale => true;
```
Chicken-Bones: Male bias here is not preferable (yes, Terraria isn't super good about this thing). Suggest:
```cs
	enum Gender { Unspecified, Male, Female }

	/// <summary>
	/// Used to set the character gender based on hair style when randomizing a new character. <br/>
	/// If Unspecified, the gender will be randomly rolled. <br/>
	/// Note that all hair styles can be selected with either gender. This is just a default for quick randomization
	/// </summary>
	public virtual Gender TypicalCharacterGender => Gender.Unspecified;
```

ExampleMod/Content/Hairs/ExampleHair.cs
```diff
@@ -0,0 +1,25 @@
...
+		public override bool IsMale => false;
+		
+		public override void SetStaticDefaults() {
+			HairID.Sets.DrawBackHair[Type] = true;
+		}
+
+		public override IEnumerable<Condition> GetUnlockConditions() {
+			return Enumerable.Empty<Condition>();
+			// TODO: throws exception on player creation menu, biome flags not available on player instance?
+			//yield return ExampleConditions.InExampleBiome;
```
Chicken-Bones: `UnlockConditions` should only be called in dresser/stylist UI. There should be a separate enum for where a hair style can be chosen, maybe with flags
```cs
CharacterCreation = 1
Dresser = 2
Stylist = 4
CharacterCreationOrStylist = 5
All = 7
```

Chicken-Bones (changes requested):
Looks great, just a few small feedback items

patches/tModLoader/Terraria/ModLoader/ModHair.cs
ScalarVector1: Not sure this is a great idea, it becomes inconsistent with ``isMale`` on Player.

ScalarVector1: or perhaps we should just go with a name which represents what it is actually used for such as ``createsMaleCharacterOnRandomization``

Chicken-Bones: How would it be inconsistent? Aside from `Unspecified` which is just an improvement, because modded hairstyles may not actually have an associated gender. Yes, name could be `RandomizedCharacterCreationGender` instead.

Chicken-Bones:
Merging this because it's good to go now and my apologies for the delay. One thing that you could consider in a followup PR is unloaded hair. 

I'd suggest:
- Adding a property to `ModHair` which lets modders optionally specify a fallback vanilla hair to be saved to the vanilla file.
- When `ModContent.TryFind` fails in `LoadHair`, save the hair name to `UnloadedPlayer` instead
- When a hairstyle is selected for a character, clear the saved `UnloadedPlayer.hair` if any. Make this a public method so mods can call it when they set the player's hair style if they do for any reason.
- When saving hair, save the `UnloadedPlayer.hair` to the file if there is one.

PR #3445, opened 2023-05-01
Add ItemID.Sets.CoinLuckValue

direwolf420:
### What is the new feature?
`ItemID.Sets.CoinLuckValue` that defines how much ["coin luck"](https://terraria.wiki.gg/wiki/Luck#Coins) the item gives to nearby players when it gets shimmered.
Removed some docs on `ItemID.Sets.CommonCoin` as coin luck now uses the new set.

Expose the system vanilla coins use to the modder via a simple one-liner.

Introduction of more generic hooks for `Item.CanShimmer` and `Item.GetShimmered` + some more code written by the modder (calling `Main.LocalPlayer.AddCoinLuck`). Should be a separate PR since that would cover more specific modder usecases.
The set is not suitable when an item does more than just giving coin luck, in that case the above needs to be used.

See ExampleMod

### ExampleMod updates
ExampleMod has a custom currency (`ExampleItem`) so it would naturally be suitable for this coin luck mechanic. The code is
`ItemID.Sets.CoinLuckValue[Type] = Item.value;`
However it's commented out since the same item is present in shimmer decrafting aswell, which is a feature clash for the sake of example. Ideally ExampleMod should get it's own item for a currency.


Rijam:
Here is a sprite for an ExampleCustomCurrency item if you wanted to add it.
ExampleCustomCurrency

PR #3446, opened 2023-05-01
Automatically register TownNPCMood localizations

JavidPack:
(The information in this description has been updated with follow-up changes from [PR #3494](https://github.com/tModLoader/tModLoader/pull/3494))

Automatically adds localization keys for town npc mood text. Note that the "[Love/Like/Dislike/Hate]Biome" and "[Love/Like/Dislike/Hate]NPC" keys will only be added to the localization file if there are registered biome or NPC preferences that would use the translation.

Uses the existing "TownNPCMoodBiomes.Key" substitutions as default values.

```
ExamplePerson: {
	DisplayName: Example Person

	TownNPCMood: {
		Content: I am content.
		NoHome: I hate not having a home.
		FarFromHome: I am too far from home.
		LoveSpace: I love having so much space.
		DislikeCrowded: I dislike how crowded I am.
		HateCrowded: I hate how crowded I am.
		LoveBiome: I love {BiomeName}.
		LikeBiome: I like {BiomeName}.
		DislikeBiome: I dislike {BiomeName}.
		HateBiome: I hate {BiomeName}.
		LoveNPC: I love {NPCName}.
		LikeNPC: I like {NPCName}.
		DislikeNPC: I dislike {NPCName}.
		HateNPC: I hate {NPCName}.
		LikeNPC_Princess: I like {NPCName}.
		Princess_LovesNPC: I like {NPCName}.
	}
}
```

If existing localizations keys are present in the mod before this change, they will be used as substitution keys for the value of the new keys. Modders will want to eventually move the text around in their .hjson file. tModLoader can't do this for the modder:

```diff
NPCs: {
	ExamplePerson: {
		DisplayName: Example Person

		TownNPCMood: {
-			Content: "{$Mods.ExampleMod.TownNPCMood.ExamplePerson.Content}"
+			Content: This is the original text.
		}
	}
}

-TownNPCMood: {
-	ExamplePerson: {
-		Content: This is the original text.
-	}
-}
```

`NPCID.Sets.NoTownNPCHappiness` added to control the happiness button.

TownNPCDialogueName are also automatically registered. These are used in town npc mood text to replace `{BiomeName}`. English examples all seem to have the word "the" at the beginning, suggesting that these need their own localization separate from Biome.BiomeName.DisplayName to work properly in all languages.

```
Biomes: {
	ExampleSurfaceBiome: {
		DisplayName: Example Surface
		TownNPCDialogueName: the Example Surface Biome
	}

	ExampleUndergroundBiome: {
		DisplayName: Example Underground
		TownNPCDialogueName: the Example Underground Biome
	}
}
```

~The princess also has special text for loving each npc. For modded content, I'm not sure if these should keep the same "TownNPCMood_Princess.LoveNPC_FullName" key or not. As it is, the key won't autopopulate the hjson, but it follows the existing pattern. Modders can add the key to their hjson if they really care, but it is a very small corner case. The default text with this change will be `"I love {NPCName}."` rather than `"$TownNPCMood_Princess.LoveNPC_FullName"`, so it is an improvement. In the future, I imagine the system could be updated either in Terraria or through a mod to give special like text for each npc, so keeping the existing key pattern could help.~

# Short Summary
TownNPCMood localization entries will automatically populate .hjson for town npc. Only used keys will be populated.
TownNPCDialogueName localization entries will automatically populate .hjson for biomes.
`NPCID.Sets.NoTownNPCHappiness` added for town npc that don't use happiness.

# Porting Notes
Move `Mods.{ModName}.TownNPCMood.{NPCName}.*` entries to `Mods.{ModName}.NPCs.{NPCName}.TownNPCMood.*`. Do this after a build/reload to make this process easier.
Move `TownNPCMood_Princess.LoveNPC_{ModNPCFullName}` to `Mods.{ModName}.NPCs.{NPCName}.TownNPCMood.Princess_LovesNPC`, if present.
Populate `Mods.{ModName}.Biomes.{BiomeName}.TownNPCDialogueName`. Usually this is similar to DisplayName, but with a lowercase "the" at the front. Used in happiness text.
Fill out any other TownNPCMood entries that might be missing still.


patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -182,14 +182,35 @@
...
  		if (type == 37 || type == 368 || type == 453)
 +		*/
-+		if (type == 37 || type == 368 || NPCID.Sets.ActsLikeTownNPC[type])
++		if (NPCID.Sets.IsNotReallyTownNPC[type])
  			return true;
  
  		return false;
++	}
++
++	internal void RegisterTownNPCMoodLocalizations(ModNPC npc)
```
Chicken-Bones: I get that happiness and shopping are pretty intertwined, but I think this code is better off in `NPCLoader`

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -182,14 +182,35 @@
...
++		if (npc.NPC.townNPC && !NPCID.Sets.IsTownPet[npc.NPC.type] && !Main.ShopHelper.IsNotReallyTownNPC(npc.NPC)) {
++
++			string prefix = $"Mods.{npc.Mod.Name}.TownNPCMood.{npc.Name}";
++			string[] keys = new string[] {
++				"Content", "NoHome", "FarFromHome", "LoveSpace", "DislikeCrowded", "HateCrowded", "LoveBiome", "LikeBiome", "DislikeBiome", "HateBiome", "LoveNPC", "LikeNPC", "DislikeNPC", "HateNPC", "LikeNPC_Princess"
++			};
++			foreach (var key in keys) {
++				string fullKey = $"{prefix}.{key}";
++				string defaultValueKey = "TownNPCMood." + key;
++				Language.GetOrRegister(fullKey, () => Language.GetTextValue(defaultValueKey));
```
Chicken-Bones: Should use a key reference to the default value, rather than `GetTextValue`

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -182,14 +182,35 @@
...
++				//Language.GetOrRegister(fullKey, () => $"Translation missing for {fullKey}");
++			}
++			// TODO: Won't show up automatically in corresponding hjson since no Mods.ModName
++			Language.GetOrRegister($"TownNPCMood_Princess.LoveNPC_{npc.FullName}", () => Language.GetTextValue("TownNPCMood_Princess.LoveNPC"));
```
Chicken-Bones: For mods, we should just add another key like `Princess_LoveNPC` for the Princess -> NPC text. Keep it all under the `Mods.ModName.TownNPCMood.ModNPCName` group

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -182,14 +182,35 @@
...
++	{
++		if (npc.NPC.townNPC && !NPCID.Sets.IsTownPet[npc.NPC.type] && !Main.ShopHelper.IsNotReallyTownNPC(npc.NPC)) {
++
++			string prefix = $"Mods.{npc.Mod.Name}.TownNPCMood.{npc.Name}";
++			string[] keys = new string[] {
++				"Content", "NoHome", "FarFromHome", "LoveSpace", "DislikeCrowded", "HateCrowded", "LoveBiome", "LikeBiome", "DislikeBiome", "HateBiome", "LoveNPC", "LikeNPC", "DislikeNPC", "HateNPC", "LikeNPC_Princess"
++			};
++			foreach (var key in keys) {
++				string fullKey = $"{prefix}.{key}";
++				string defaultValueKey = "TownNPCMood." + key;
```
Chicken-Bones: Why have a separate `TownNPCMood.NPCName` category, rather than putting them all with the NPC themselves (via `npc.GetLocalization`)

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
```diff
@@ -44,7 +44,7 @@ public abstract class ModBiome : ModSceneEffect, IShoppingBiome, ILocalizedModTy
 
 	public GameContent.Bestiary.ModBiomeBestiaryInfoElement ModBiomeBestiaryInfoElement { get; internal set; }
 
-	string IShoppingBiome.NameKey => Name;
+	string IShoppingBiome.NameKey => $"Mods.{Mod.Name}.TownNPCMoodBiomes.{Name}";
```
Chicken-Bones: Is there a reason for this to be different to `DisplayName`?

I looked up references to `NameKey` and they're pretty scares, it seems like it would be better to replace it with `LocalizedText DisplayName` (on `IShoppingBiome`) and pass the `LocalizedText` everywhere.

It is only ever used to create the localization substitutions object via 
`new { BiomeName = BiomeNameByKey(...)`

but we could just do `new { BiomeName = DisplayName }` since `LocalizedText` instances work fine as substitution args

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -15,11 +15,16 @@ public static partial class Sets
...
-		/// on the minimap, like the bone merchant in vanilla.
+		/// Whether or not a given NPC will act like a town NPC in terms of AI, animations, and attacks, but not in other regards, such as appearing on the minimap, like the bone merchant in vanilla.
 		/// </summary>
 		public static bool[] ActsLikeTownNPC = Factory.CreateBoolSet(453);
 
+		//Default ID is the skeleton merchant, traveling merchant, old man
+		/// <summary>
+		/// If true, the given NPC will not count towards town NPC happiness and won't have a happiness button.
+		/// </summary>
+		public static bool[] IsNotReallyTownNPC = Factory.CreateBoolSet(37, 368, 453);
```
Chicken-Bones: Yeah, just call this `NoTownNPCHappiness`, docs should note that it is not necessary to set this for pets `IsTownPet`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
direwolf420: I think this code is supposed to port pre-1.4.4 way of writing happiness to 1.4.4 so the modder doesn't have to manually copypaste it

JavidPack: My thought was the modder would see the entries in the .hjson and immediately be able to edit them to fit the voice and characterization of their town npc. If the modder just sees `"{$TownNPCMood.DislikeBiome}"`, they wouldn't know to use `{BiomeName}` in the text that they eventually write without looking at the game's source code.

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
JavidPack: Well, they are different in ways in vanilla that seem to suggest they need to be different. In vanilla, `TownNPCMoodBiomes.Snow` is `"the Snow"`, whereas `Bestiary_Biomes.Snow` is `"Snow"`.

Others are a difference between "The" and "the" ("The Dungeon" vs "the Dungeon"). TownNPCMoodBiomes entries are used within a sentence, `"I love {BiomeName}."`, but Bestiary_Biomes are used in the bestiary in the search filter as button labels, so capitalization is used for the word "The".

We could make some assumptions, but I assume there might be issues later on.

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
Chicken-Bones: I agree, my misunderstanding. I'm not sure what this will do on non-english languages, but should be okay.

patches/tModLoader/Terraria/ModLoader/ModBiome.cs
Chicken-Bones: Alright, but same feedback here as for NPC mood in general. Better to use `GetLocalizationKey("TownNPCDialogueName")`

patches/tModLoader/Terraria/Localization/Content/en-US/Town.json.patch
```diff
@@ -0,0 +1,11 @@
+--- src/TerrariaNetCore/Terraria/Localization/Content/en-US/Town.json
++++ src/tModLoader/Terraria/Localization/Content/en-US/Town.json
+@@ -1836,6 +_,8 @@
+ 		"LoveNPC": "I love {NPCName}.",
+ 		"DislikeNPC": "I dislike {NPCName}.",
+ 		"HateNPC": "I hate {NPCName}.",
++		"LikeNPC_Princess": "I like {NPCName}.",
++		"Princess_LovesNPC": "I like {NPCName}.",
```
Chicken-Bones: Better to add these to `tModLoader.json`

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -69,26 +69,28 @@
...
++			text = ModContent.GetModNPC(npcType).GetLocalizationKey("TownNPCMood");
++		else if (_currentNPCBeingTalkedTo.type < NPCID.Count)
 -		string text = "TownNPCMood_" + NPCID.Search.GetName(_currentNPCBeingTalkedTo.netID);
 +			text = "TownNPCMood_" + NPCID.Search.GetName(_currentNPCBeingTalkedTo.netID);
 +		else {
 +			var modNpc = _currentNPCBeingTalkedTo.ModNPC;
-+
-+			text = $"Mods.{modNpc.Mod.Name}.TownNPCMood.{modNpc.Name}";
++			text = modNpc.GetLocalizationKey("TownNPCMood");
 +		}
```
Chicken-Bones: Can you work this patch without the deletion? 

```cs
string text = "TownNPCMood_" + NPCID.Search.GetName(_currentNPCBeingTalkedTo.netID);
if (textKeyInCategory == "Princess_LovesNPC")
    ...
else if (_currentNPCBeingTalkedTo.ModNPC is ModNPC modNPC)
    text = modNpc.GetLocalizationKey("TownNPCMood");
```

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -69,26 +69,28 @@
...
+-	private static string BiomeNameByKey(string biomeNameKey) => Language.GetTextValue("TownNPCMoodBiomes." + biomeNameKey);
 +	//TML: Turned public.
 +	//TODO: Anything else?
--	private static string BiomeNameByKey(string biomeNameKey) => Language.GetTextValue("TownNPCMoodBiomes." + biomeNameKey);
-+	public static string BiomeNameByKey(string biomeNameKey) => Language.GetTextValue("TownNPCMoodBiomes." + biomeNameKey);
++	public static string BiomeNameByKey(string biomeNameKey) => Language.GetTextValue(biomeNameKey.StartsWith("Mods.") ? biomeNameKey : "TownNPCMoodBiomes." + biomeNameKey);
  
- 	private void AddHappinessReportText(string textKeyInCategory, object substitutes = null)
+-	private void AddHappinessReportText(string textKeyInCategory, object substitutes = null)
++	private void AddHappinessReportText(string textKeyInCategory, object substitutes = null, int npcType = 0)
```
Chicken-Bones: `npcType -> otherNPCType`

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -69,26 +69,28 @@
...
  	{
 +		string text;
 +
-+		if (_currentNPCBeingTalkedTo.type < NPCID.Count)
++		if(textKeyInCategory == "Princess_LovesNPC")
```
Chicken-Bones: Something to think about in the future, is there any way for mods to add these sort of relationships between their own NPCs and another mod's?

Feature 1: `ModTownNPC` from another mod adds a special localization for liking a specific NPC, rather than just the general substitution based one.

Feature 2: A way for another mod to make `ExamplePerson` like `ModTownNPC`, with the localization defined in `ModTownNPC` (like `ExamplePerson_LovesNPC`)

Feel free to convert to a feature request issue

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -140,9 +142,19 @@
...
- 		AddHappinessReportText("LoveNPC_" + NPCID.Search.GetName(npcType), new {
- 			NPCName = NPC.GetFullnameByID(npcType)
- 		});
++		if (npcType < NPCID.Count)
+-		AddHappinessReportText("LoveNPC_" + NPCID.Search.GetName(npcType), new {
++			AddHappinessReportText("LoveNPC_" + NPCID.Search.GetName(npcType), new {
+-			NPCName = NPC.GetFullnameByID(npcType)
++				NPCName = NPC.GetFullnameByID(npcType)
+-		});
++			});
```
Chicken-Bones: `goto` or early return?

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
```diff
@@ -140,9 +142,19 @@
 -	public void LoveNPCByTypeName(int npcType)
 +	internal void LoveNPCByTypeName(int npcType)
  	{
- 		AddHappinessReportText("LoveNPC_" + NPCID.Search.GetName(npcType), new {
- 			NPCName = NPC.GetFullnameByID(npcType)
- 		});
++		if (npcType < NPCID.Count)
```
Chicken-Bones: needs a type check on this NPC. What if a mod is using this to love another modded NPC? May not be princess.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/ShopHelper.cs.patch
JavidPack: I haven't seen any modder care enough about this to mention it. Mostly modders have been confused that princess even has those npc specific texts. I'll make an issue though.

JavidPack: https://github.com/tModLoader/tModLoader/issues/3493

Chicken-Bones (approved):

PR #3453, opened 2023-05-04
Rework NPCID.Sets.DebuffImmunitySets

### LINKED ISSUE #3358: Buff Immunities (NPCID.Sets.DebuffImmunitySets) too difficult to use

Yes

Mod capability as a Modder

### Description

Ever since 1.4, `NPCID.Sets.DebuffImmunitySets` contains default buff immunity data for every NPC. During `NPC.SetDefaults`, the `NPC.buffImmune` array is filled out in accordance with the data contained within. If no entry is found, the NPC is set to be immune to no buffs. Importantly, this runs after `NPCLoader.SetDefaults(this);`, meaning that there is no sensible place to set immunities in the way taught in 1.3 code.

Currently, mods can use `ModNPC.SetStaticDefaults` to register a `NPCDebuffImmunityData` to `NPCID.Sets.DebuffImmunitySets`. This covers the use case of setting specific immunities for a specific npc, but there are still situations that are difficult.

- Buff that acts like or "inherits" the behavior of another buff:
  - For example, Meteor Head is immune to `OnFire` and `OnFire3` ("Hellfire"), which is stronger version of a fire based debuff. 
  - A modded fire based debuff might want to inherit the immunity, but the code to do so is very verbose due to the dictionary and int[] elements.
  - In 1.3, modders would simply add `NPC.buffImmune[ModContent.BuffType<MySuperOnFire>()] = NPC.buffImmune[BuffID.OnFire];` in `GlobalNPC.SetDefaults`, but the current design doesn't allow that.
 
- Adding buff immunities to specific vanilla NPC:
  - Adding specific immunities is also hard, for the same reasons.

## Workaround
Some mods do the work to edit `NPCID.Sets.DebuffImmunitySets` directly, but it is very verbose code. Some mods bypass the system altogether and run their own 1.3 style code in PreAI and edit `NPC.buffImmune` directly. I think it is useful for default npc immunities to be data, so it would be nice to facilitate the use cases with some helper methods.

## Dynamic immunities
Immunity changes resulting from AI code are all still possible and work the same.

Discord discussion link: https://discord.com/channels/103110554649894912/445276626352209920/1090218856028581898

Thorium approach
[LAEZPTXu.txt](https://github.com/tModLoader/tModLoader/files/11092836/LAEZPTXu.txt)

_No response_

_No response_

JavidPack:
Fixes #3358

The 1.4 change to `NPCID.Sets.DebuffImmunitySets` has caused a lot of unnecessary frustration. This PR attempts to fix that, allowing debuff immunities to be easily editable but also keeping the approach of having `NPCID.Set`s to store a static reference of default npc immunities.

# Existing Behavior
Currently `NPCID.Sets.DebuffImmunitySets` is used to define default NPC immunities. `DebuffImmunitySets` maps NPCType to `NPCDebuffImmunityData`. `NPCDebuffImmunityData` is hard to edit, as the `SpecificallyImmuneTo` field is an array of BuffTypes that is sometimes null. Additionally, `DebuffImmunitySets` doesn't have entries for all NPCTypes. Together, these make editing the data contained frustrating and code intensive. (Checking for null, initializing if null, running code to change the array and then assigning the array back)

The data contained in `DebuffImmunitySets` is used to initialize `NPC.buffImmune`. Crucially, it runs after `NPCLoader.SetDefaults`, meaning that any modder wishing to tweak a vanilla NPC's default immunity would have to write lots of code to change `DebuffImmunitySets` or make a way to run code only once in `GlobalNPC.AI`. This code can be moved, but modders would still have to wrangle `DebuffImmunitySets`. 

Buff inheritance is implicitly implied in Terraria by the consistency that several buffs are applied in tandem with other buffs. For example, all NPC immune to `OnFire` are immune to `OnFire3`, and the same for `Frostburn` and `Frostburn2`. Mods adding buffs of similar "attributes" can implement the buff immunity themselves, but there is no way for multiple mods to add buffs and NPC and result in a consistent experience for players. (For example, the player would expect FireImpA added by ModA to be immune to OnFireC added by ModC because FireImpA is immune to OnFire and OnFire3)

# Changes

- `NPCID.Sets.DebuffImmunitySets` removed, replaced by `NPCID.Sets.SpecificDebuffImmunity`, `NPCID.Sets.ImmuneToAllBuffs`, and `NPCID.Sets.ImmuneToRegularBuffs`.
- `NPCID.Sets.ShimmerImmunity` removed, replaced by `NPCID.Sets.SpecificDebuffImmunity`.
- Buff immunity inheritance has been made explicit and has been made modder editable.
  - `BuffID.Sets.GrantImmunityWith` set added to facilitate buff immunity inheritance. `OnFire3->OnFire`, `Frostburn2->Frostburn`, and `Poisoned->Venom` are the initial values. This means that any NPC immune to `OnFire` will automatically be immune to `OnFire3`
  - `NPC.BecomeImmuneTo` method added. Allows adjusting NPC immunity while honoring immunity inheritance. A more compatible approach to dynamic immunity than adjusting `NPC.immune` directly.
  - `NPC.ClearImmuneToBuffs` method added. Used to clear any buffs that the npc has recently become immune to.
- `NPC.buffImmune` is now populated before `ModNPC/GlobalNPC.SetDefaults`. This allows modders to make advanced changes to buff immunities if needed.

## Buff immunity inheritance algorithm

1. NPC defaults to being vulnerable to all buffs: `NPC.buffImmune` all false
2. If `NPCID.Sets.ImmuneToAllBuffs`, `NPC.buffImmune` is filled with true
3. If `NPCID.Sets.ImmuneToAllBuffs` or `NPCID.Sets.ImmuneToRegularBuffs`
    - If `NPCID.Sets.ImmuneToRegularBuffs`, immune to all non-tag buffs
    - Become vulnerable to all `NPCID.Sets.SpecificDebuffImmunity` entries that are `false`
    - Apply `BuffID.Sets.GrantImmunityWith`: NPC becomes vulnerable to a buff if the buff inherits from other buffs and the NPC is vulnerable to all of those buffs.
    - Become immune to all `NPCID.Sets.SpecificDebuffImmunity` entries that are `true`
4. If neither `NPCID.Sets.ImmuneToAllBuffs` nor `NPCID.Sets.ImmuneToRegularBuffs`
    - Become immune to all `NPCID.Sets.SpecificDebuffImmunity` entries that are `true`
    - Apply `BuffID.Sets.GrantImmunityWith`: NPC becomes immune to a buff if the buff inherits from other buffs and the NPC is immune to any of those buffs.
    - Become vulnerable to all `NPCID.Sets.SpecificDebuffImmunity` entries that are `false`
 5. `NPCLoader.SetDefaults`. Any advanced logic can be done by modder code.

### Example
Imagine a `ModNPC` named `WaterElemental`. The author specifies that it is immune to `OnFire` (TODO)

# Porting Notes
Modders will need to rework all lines involving `NPCID.Sets.DebuffImmunitySets`, `BuffID.Sets.IsAnNPCWhipDebuff`, `NPCID.Sets.ShimmerImmunity`, and likely `NPC.buffImmune`. 

Running `tModPorter` will change `IsAnNPCWhipDebuff` to `IsATagBuff` and mark `DebuffImmunitySets` with notes. The other changes require manual work and possibly code restructuring.

## NPCID.Sets.DebuffImmunitySets
`NPCID.Sets.DebuffImmunitySets` is now split up into several "normal" sets for easier manipulation.

1. Replace `NPCDebuffImmunityData.SpecificallyImmuneTo` with `NPCID.Sets.SpecificDebuffImmunity`

```diff
// Specify the debuffs it is immune to
-NPCID.Sets.DebuffImmunitySets.Add(Type, new NPCDebuffImmunityData {
-	SpecificallyImmuneTo = new int[] {
-		BuffID.Poisoned // This NPC will be immune to the Poisoned debuff.
-	}
-});
+// This NPC will be immune to the Poisoned debuff.
+NPCID.Sets.SpecificDebuffImmunity[Type][BuffID.Poisoned] = true;
```

Also remove any code doing a similar approach for other NPC by editing DebuffImmunitySets directly.

2. Replace `NPCDebuffImmunityData.ImmuneToWhips`  and `NPCDebuffImmunityData.ImmuneToAllBuffsThatAreNotWhips` with `NPCID.Sets.ImmuneToAllBuffs` and `NPCID.Sets.ImmuneToRegularBuffs`

**IMPORTANT:** The meaning of the new sets is different from the original sets. Set `ImmuneToAllBuffs` only if `ImmuneToWhips` and `ImmuneToAllBuffsThatAreNotWhips` were set originally. If only `ImmuneToAllBuffsThatAreNotWhips` was set, set `ImmuneToRegularBuffs`. If just `ImmuneToWhips` was set, that doesn't make sense, you likely misunderstood it.

```diff
// Immune to all buffs
-NPCID.Sets.DebuffImmunitySets.Add(Type, new NPCDebuffImmunityData {
-	new NPCDebuffImmunityData {
-		ImmuneToAllBuffsThatAreNotWhips = true,
-		ImmuneToWhips = true
-	}
-};
+NPCID.Sets.ImmuneToAllBuffs[Type] = true;

// Immune to all non-tag buffs
-NPCID.Sets.DebuffImmunitySets.Add(Type, new NPCDebuffImmunityData {
-	new NPCDebuffImmunityData {
-		ImmuneToAllBuffsThatAreNotWhips = true,
-	}
-};
+NPCID.Sets.ImmuneToRegularBuffs[Type] = true;
```

## NPCID.Sets.ShimmerImmunity
 `NPCID.Sets.ShimmerImmunity` removed, replace with `NPCID.Sets.SpecificDebuffImmunity`:
```diff
-NPCID.Sets.ShimmerImmunity[Type] = true;
+NPCID.Sets.SpecificDebuffImmunity[Type][BuffID.Shimmer] = true;
```

## Buff "inheritance" or editing NPC.buffImmune
tModLoader has never specifically supported buff immunity inheritance, but many modders have implemented this independently in their own mods. Modders should adjust their code to use the tModLoader provided buff inheritance for compatibility with other mods. Remove all code adjusting `NPCID.Sets.DebuffImmunitySets.DebuffImmunitySets` directly and replace it with suitable alternative.

Use `BuffID.Sets.GrantImmunityWith[buffType].Add(BuffID.Ichor);` to indicate that an NPC should become immune to this buff if it is immune to the specified buff (in this example, Ichor).

```diff
// in ModNPC.AI or similar methods
-npc.buffImmune[ModContent.BuffType<ExampleJavelinDebuff>()] = npc.buffImmune[BuffID.BoneJavelin]; 
// in ModBuff.SetStaticDefaults
+BuffID.Sets.GrantImmunityWith[Type].Add(BuffID.BoneJavelin); 
```

Code can also be simplified in anticipation of buff inheritance. Code setting immunity to both `OnFire3` and `OnFire` for an NPC, for example, could just set an immunity to `OnFire`. Leaving it in is no issue, however.

Do note that if you require a specific buff immunity to not be inherited by an NPC, you'll need to add code specifying that the NPC is specifically vulnerable to that buff: `NPCID.Sets.SpecificDebuffImmunity[Type][BuffID.OnFire3] = false;`

Manually editing `NPC.buffImmune` on the 1st frame an NPC is active or writing convoluted code to adjust `NPCID.Sets.DebuffImmunitySets` should no longer be necessary.

## Dynamic NPC.buffImmune
Manually changing `NPC.buffImmune` during `AI` is still fine, but using `NPC.BecomeImmuneTo` and `NPC.ClearImmuneToBuffs` instead will honor buff immunity inheritance. See `MinionBossBody.ApplySecondStageBuffImmunities` and the docs to learn more.


ExampleMod/Common/GlobalNPCs/DamageOverTimeGlobalNPC.cs
```diff
@@ -15,11 +15,13 @@ internal class DamageOverTimeGlobalNPC : GlobalNPC
...
-		public override void SetDefaults(NPC npc) {
-			// TODO: This doesn't work currently. tModLoader needs to make a fix to allow changing buffImmune
-
-			// We want our ExampleJavelin buff to follow the same immunities as BoneJavelin
-			npc.buffImmune[ModContent.BuffType<ExampleJavelinDebuff>()] = npc.buffImmune[BuffID.BoneJavelin];
+		public override void SetStaticDefaults() {
+			// We want our ExampleJavelin buff to follow the same immunities as BoneJavelin. SkeletronHead and SkeletronPrime, for example, are immune to BoneJavelin
+			for (int i = 0; i < NPCLoader.NPCCount; i++) {
+				if (NPCID.Sets.ImmuneToDebuffs[i].Contains(BuffID.BoneJavelin)) {
+					NPCID.Sets.ImmuneToDebuffs[i].Add(ModContent.BuffType<ExampleJavelinDebuff>());
```
Chicken-Bones: This seems order dependent wrt to other mod entities :|
This might be okay in `PostSetupContent`

patches/tModLoader/Terraria/DataStructures/NPCDebuffImmunityData.cs.patch
```diff
@@ -1,12 +1,13 @@
...
 +++ src/tModLoader/Terraria/DataStructures/NPCDebuffImmunityData.cs
-@@ -1,17 +_,36 @@
+@@ -1,17 +_,37 @@
  using Terraria.ID;
 +using Terraria.ModLoader;
  
  namespace Terraria.DataStructures;
  
 +/// <summary>
++/// <b>Unused:</b> Replaced by <see cref="NPCID.Sets.ImmuneToDebuffs"/>, <see cref="NPCID.Sets.ImmuneToAllDebuffsByDefault"/>, and <see cref="NPCID.Sets.ImmuneToWhipDebuffs"/><br/><br/>
```
Chicken-Bones: Need some examples of using these two

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -50,5 +53,45 @@ public static partial class Sets
...
+			ImmuneToDebuffs = Factory.CreateCustomSet<List<int>>(null);
+			for (int type = 0; type < NPCLoader.NPCCount; type++) {
+				if (DebuffImmunitySets.TryGetValue(type, out var data) && data != null) {
+					// Meaning slightly changed: ImmuneToWhipDebuffs now bool? and ImmuneToAllDebuffsByDefault doesn't account care about whips, hence the logic.
+					if (data.ImmuneToWhips) {
+						ImmuneToWhipDebuffs[type] = !data.ImmuneToAllBuffsThatAreNotWhips ? true : null;
+					}
+					else {
+						ImmuneToWhipDebuffs[type] = data.ImmuneToAllBuffsThatAreNotWhips ? false : null;
+					}
```
Chicken-Bones: might be more readable as a ternary
```
ImmuneToWhipDebuffs[type] =
    data.ImmuneToWhips && !data.ImmuneToAllBuffsThatAreNotWhips ? true :
    !data.ImmuneToWhips && data.ImmuneToAllBuffsThatAreNotWhips ? false :
    null;
```

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -774,12 +774,25 @@
...
-@@ -10793,13 +_,18 @@
+@@ -10792,14 +_,31 @@
+ 			DeathSound = SoundID.NPCDeath6;
  			catchItem = 2121;
  		}
- 
-+		NPCLoader.SetDefaults(this);
++		// TML: Moved before NPCLoader.SetDefaults to allow changing. DebuffImmunitySets broken into separate sets for easier manipulation.  
++		if (NPCID.Sets.ImmuneToAllDebuffsByDefault[type])
++			Array.Fill(buffImmune, true);
```
Chicken-Bones, line 837: Do you need to fill the array regardless? Where does it get cleared?
`Array.Fill(buffImmune, NPCID.Sets.ImmuneToAllDebuffsByDefault[type]);`

Chicken-Bones (commented):
This design looks good!

The PR description or ExampleMod needs to give some code snippets indicating how modders could replicate the `ImmuneToWhipDebuffs` pattern.

One potential downside that may need to be made clear in documentation, is that you can't actually get the default debuff immunity of an NPC statically, because we don't 'bake' the sets, and modders may have their own things like `ImmuneToWhipDebuffs` or `ShimmerImmunity`

The example of making an enemy immune to `ExampleJavelinDebuff` is somewhat concerning because it wouldn't generalise well to situations where the target buff isn't in `ImmuneToDebuffs`

patches/tModLoader/Terraria/NPC.cs.patch
JavidPack: It gets filled with false at the start of NPC.SetDefaults, so we don't need to add code for that.

ExampleMod/Common/GlobalNPCs/BuffImmuneGlobalNPC.cs
```diff
@@ -0,0 +1,55 @@
...
+			// Firstly, it links together existing debuffs with modded debuffs. These links ensure that any NPC immune to an existing debuff will automatically be immune to the modded debuff. See ExampleJavelinDebuff.cs for another examples of this concept.
+			// Secondly, it changes some existing NPC immunities.
+
+			// This code causes any NPC immune to Ichor to automatically be immune to BetsysCurse and ExampleDefenseDebuff.
+			foreach (var defenseReductionDebuff in IsADefenseReductionDebuff) {
+				if(defenseReductionDebuff == BuffID.Ichor) {
+					continue;
+				}
+				BuffID.Sets.GrantImmunityWith[defenseReductionDebuff].Add(BuffID.Ichor);
+			}
```
Chicken-Bones: This would have to run in `PostSetupContent` since there's no guarantee that everything has been added to `IsADefenseReductionDebuff`. Unless `IsADefenseReductionDebuff` is used for some runtime mechanic, a better solution would be to make a function like `SetDefenseDebuffStaticDefaults` which could have the `GrantImmunityWith` implementation.

ExampleMod/Common/GlobalNPCs/BuffImmuneGlobalNPC.cs
```diff
@@ -0,0 +1,55 @@
...
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.GlobalNPCs
+{
+	// Showcases various debuff immunity changes. Both individual debuff immunity changes and inherited immunity changes are shown.
+
+	// For mod compatibility, it is important to put code adjusting debuffs in the correct methods:
+	// Adjust BuffID.Sets.GrantImmunityWith and NPCID.Sets.SpecificDebuffImmunity in ModNPC/GlobalNPC.SetStaticDefaults.
```
Chicken-Bones, line 10: It is acceptable to adjust `SpecificDebuffImmunity` in `PostSetupContent` if there is a need to change the immunities of content from other mods and `GrantImmunityWith` is insufficient.

ExampleMod/Common/GlobalNPCs/BuffImmuneGlobalNPC.cs
```diff
@@ -0,0 +1,55 @@
...
+	// Only adjust NPC.buffImmune[] directly in ModNPC/GlobalNPC.SetDefaults for complex or world state specific logic.
+	// Adjusting NPC.buffImmune[] during AI methods is possible, but be aware that debuff inheritance will not apply.
```
Chicken-Bones: debuff inheritance won't apply to manual changes made in either `AI` or `SetDefaults`.

ExampleMod/Common/GlobalNPCs/BuffImmuneGlobalNPC.cs
```diff
@@ -0,0 +1,55 @@
...
+		public static HashSet<int> IsADefenseReductionDebuff = new HashSet<int>() { BuffID.Ichor, BuffID.BetsysCurse };
+
+		public override void SetStaticDefaults() {
+			// DONT apply to NPCID.Sets.ImmuneToDebuffs, that's not the point.
+
+			// This example does 2 things.
+			// Firstly, it links together existing debuffs with modded debuffs. These links ensure that any NPC immune to an existing debuff will automatically be immune to the modded debuff. See ExampleJavelinDebuff.cs for another examples of this concept.
+			// Secondly, it changes some existing NPC immunities.
+
+			// This code causes any NPC immune to Ichor to automatically be immune to BetsysCurse and ExampleDefenseDebuff.
```
Chicken-Bones: Should note that in vanilla, there are several enemies which are immune to ichor who are _not_ immune to `BetsysCurse` (and maybe list them)

ExampleMod/Content/NPCs/MinionBoss/MinionBossBody.cs
```diff
@@ -149,9 +143,8 @@ public class MinionBossBody : ModNPC
 			NPC.boss = true;
 			NPC.npcSlots = 10f; // Take up open spawn slots, preventing random NPCs from spawning during the fight
 
-			// Don't set immunities like this as of 1.4:
-			// NPC.buffImmune[BuffID.Confused] = true;
-			// immunities are handled via dictionaries through NPCID.Sets.DebuffImmunitySets
+			// Default buff immunities should be set in SetStaticDefaults through the NPCID.Sets.ImmuneTo{X} arrays.
+			// To dynamically adjust immunities of an active NPC, NPC.buffImmune[] can be changed in AI: NPC.buffImmune[BuffID.OnFire] = true;
```
Chicken-Bones, line 148: It would be nice if we had our own `SuperFire` debuff as an example, along with minion boss purging all fire (and frostburn) debuffs as part of AI for a time (setting `buffImmune` and also going through the inheritances to see which other buffs to clear)

Chicken-Bones: Might need some helpers for efficiently reframing the immunity question to be asked the other way (list of buffs to also be immune to if immune to X)

patches/tModLoader/Terraria/ID/BuffID.TML.cs
```diff
@@ -13,5 +15,25 @@ public partial class Sets
...
+
+		/// <summary>
+		/// An NPC will automatically become immune (<see cref="NPC.buffImmune"/>) to the indexed buff type if it is already immune to any buff in the corresponding List. <br/>
+		/// For NPCs with <see cref="NPCID.Sets.FullyImmuneToBuffs"/> or <see cref="NPCID.Sets.DefaultDebuffImmunity"/> set to true, the reverse of that logic is used to automatically set the NPC as un-immune: The NPC will automatically become un-immune to the indexed buff type if it is already un-immune to all of the buffs in the corresponding List.<br/><br/>
+		/// This set helps standardize several buff immunity inheritance rules that are effectively present in Terraria while providing modders the ability to do the same. For example, all NPC with immunity to <see cref="OnFire"/> are also immune to <see cref="OnFire3"/>.<br/>
+		/// Modders can use this set to create buffs that act as variants of existing buffs, and should inherit the immunity value of those existing buffs. NPC from Terraria and other mods will automatically have these immunity rules applied, regardless of mod load order.
+		/// </summary>
```
Chicken-Bones, line 24: Excellent comments here. `un-immune` -> `vulnerable`, `already un-immune` -> `specifically vulnerable`

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
 		/// Note: This set DOES NOT DO ANYTHING if your NPC doesn't use the Vanilla TownNPC aiStyle (aiStyle == 7).
 		/// </remarks>
 		public static bool[] AllowDoorInteraction = Factory.CreateBoolSet();
+
+		/// <summary>
+		/// If <see langword="true"/>, this NPC type (<see cref="NPC.type"/>) will be immune to all debuffs by default.<br/><br/>
```
Chicken-Bones: All debuffs and 'tag' buffs

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
...
+		/// Use this for special NPC that cannot be hit at all, such as fairy critters, container NPCs like Martian Saucer and Pirate Ship, bound town slimes, and Blazing Wheel. Dungeon Guardian also is in this set to prevent the bonus damage from whip debuffs.<br/><br/>
+		/// If the NPC should be attacked, it is recommended to set <see cref="DefaultDebuffImmunity"/> to <see langword="true"/> instead. This will prevent all debuffs except whip debuffs (<see cref="BuffID.Sets.IsAnNPCWhipDebuff"/>), which are intended to affect enemies typically seen as immune to all debuffs.<br/><br/>
```
Chicken-Bones: We should consider renaming `IsAnNPCWhipDebuff` to `IsAnNPCTagBuff` with appropriate documentation, that modders can use it for any sort of time limited interaction between NPCs and attacks. Noting specifically that the purpose of it is to work with `DefaultDebuffImmunity` in that `TagBuffs` are not considered _debuffs_ in that they're a combat mechanic, not something adversely affecting an NPC.

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
...
+		/// Modders can specify specific buffs to not be immune to by assigning <see cref="SpecificDebuffImmunity"/> to false.
```
Chicken-Bones: `not be immune` -> `be vulnerable`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
 		/// Note: This set DOES NOT DO ANYTHING if your NPC doesn't use the Vanilla TownNPC aiStyle (aiStyle == 7).
 		/// </remarks>
 		public static bool[] AllowDoorInteraction = Factory.CreateBoolSet();
+
+		/// <summary>
+		/// If <see langword="true"/>, this NPC type (<see cref="NPC.type"/>) will be immune to all debuffs and "tag" buffs by default.<br/><br/>
+		/// Use this for special NPC that cannot be hit at all, such as fairy critters, container NPCs like Martian Saucer and Pirate Ship, bound town slimes, and Blazing Wheel. Dungeon Guardian also is in this set to prevent the bonus damage from "tag" buffs.<br/><br/>
```
direwolf420:
```suggestion
		/// Use this for special NPCs that cannot be hit at all, such as fairy critters, container NPCs like Martian Saucer and Pirate Ship, bound town slimes, and Blazing Wheel. Dungeon Guardian also is in this set to prevent the bonus damage from "tag" buffs.<br/><br/>
```

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
...
+		/// Modders can specify specific buffs to be vulnerable to by assigning <see cref="SpecificDebuffImmunity"/> to false.
+		/// </summary>
+		public static bool[] FullyImmuneToBuffs; // derived from DebuffImmunitySets
+
+		/// <summary>
+		/// If <see langword="true"/>, this NPC type (<see cref="NPC.type"/>) will be immune to all debuffs except tag debuffs (<see cref="BuffID.Sets.IsATagBuff"/>) by default.<br/><br/>
+		/// Use this for NPC that can be attacked that should be immune to all normal debuffs. Tag debuffs are special debuffs that facilitate combat mechanics, such as the "summon tag damage" applied by whip weapons. Wraith, Reaper, CultistBoss, the Celestial Pillars, The Destroyer, and the Martian Saucer Turret/Cannon/Core are examples of NPC that use this setting.<br/><br/>
```
direwolf420:
```suggestion
		/// Use this for NPCs that can be attacked that should be immune to all normal debuffs. Tag debuffs are special debuffs that facilitate combat mechanics, such as the "summon tag damage" applied by whip weapons. Wraith, Reaper, Lunatic Cultist, the Celestial Pillars, The Destroyer, and the Martian Saucer Turret/Cannon/Core are examples of NPCs that use this setting.<br/><br/>
```

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -55,5 +59,48 @@ public static partial class Sets
...
+		/// Modders can specify specific buffs to be vulnerable to by assigning <see cref="SpecificDebuffImmunity"/> to false.
+		/// </summary>
+		public static bool[] DefaultDebuffImmunity; // derived from DebuffImmunitySets
+
+		/// <summary>
+		/// Indexed by NPC type and then Buff type. If <see langword="true"/>, this NPC type (<see cref="NPC.type"/>) will be immune (<see cref="NPC.buffImmune"/>) to the specified buff type. If <see langword="false"/>, the NPC will not be immune.<br/><br/>
+		/// By default, NPC aren't immune to any buffs, but <see cref="DefaultDebuffImmunity"/> or <see cref="FullyImmuneToBuffs"/> can make an NPC immune to all buffs. The values in this set override those settings.<br/><br/>
```
direwolf420:
```suggestion
		/// By default, NPCs aren't immune to any buffs, but <see cref="DefaultDebuffImmunity"/> or <see cref="FullyImmuneToBuffs"/> can make an NPC immune to all buffs. The values in this set override those settings.<br/><br/>
```

patches/tModLoader/Terraria/ID/NPCID.cs.patch
```diff
@@ -138,12 +138,15 @@
  		public static Dictionary<int, NPCBestiaryDrawModifiers> NPCBestiaryDrawOffset = NPCBestiaryDrawOffsetCreation();
 +
 +		/// <summary>
++		/// <b>Unused:</b> Replaced by <see cref="SpecificDebuffImmunity"/>, <see cref="FullyImmuneToBuffs "/>, and <see cref="DefaultDebuffImmunity"/><br/><br/>
 +		/// The default debuff immunities for a given NPC type (<see cref="NPC.type"/>).
 +		/// </summary>
 +		/// <remarks><see cref="NPC.buffImmune"/> can still be manually changed to grant temporary immunity.</remarks>
- 		public static Dictionary<int, NPCDebuffImmunityData> DebuffImmunitySets = new Dictionary<int, NPCDebuffImmunityData> {
+-		public static Dictionary<int, NPCDebuffImmunityData> DebuffImmunitySets = new Dictionary<int, NPCDebuffImmunityData> {
++		private static Dictionary<int, NPCDebuffImmunityData> DebuffImmunitySets = new Dictionary<int, NPCDebuffImmunityData> {
```
direwolf420, line 146: Do there need to be docs for something that is private?

direwolf420: Same for ShimmerImmunity set

patches/tModLoader/Terraria/NPC.TML.cs
```diff
@@ -197,4 +197,52 @@ public int GetPartyHatGore()
...
+	/// Adjusts <see cref="buffImmune"/> to make this NPC immune to the provided buff as well as all other buffs that inherit the immunity of that buff (via <see cref="BuffID.Sets.GrantImmunityWith"/>). This method can be followed by <see cref="ClearImmuneToBuffs(out bool)"/> if the NPC should clear any buff it currently has that it is now immune to.
+	/// </summary>
+	/// <param name="buffType"></param>
+	public void BecomeImmuneTo(int buffType)
+	{
+		buffImmune[buffType] = true; 
+
+		for (int i = 0; i < BuffID.Sets.GrantImmunityWith.Length; i++) {
+			var buffsToInherit = BuffID.Sets.GrantImmunityWith[i];
+			if (buffsToInherit.Contains(buffType)) {
```
direwolf420, line 286: Could this be sped up perhaps? Or would that be premature optimization, since its unlikely that it will be called every tick, assuming modders follow the ExampleMod example.

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -774,12 +781,78 @@
...
++			// If not immune to most buffs...
++			// Apply all `SpecificDebuffImmunity` `true` overrides
++			for (int i = 0; i < NPCID.Sets.SpecificDebuffImmunity[type].Length; i++) {
++				bool? buffImmunity = NPCID.Sets.SpecificDebuffImmunity[type][i];
++				if (buffImmunity is true) {
++					buffImmune[i] = true;
++				}
++			}
++
++			// Apply all `GrantImmunityWith`  (if immune to OnFire, add OnFire3, keep in mind this is _any_ of the immunities in `[OnFire3]` list)
```
direwolf420:
```suggestion
+			// Apply all `GrantImmunityWith` (if immune to OnFire, add OnFire3, keep in mind this is _any_ of the immunities in `[OnFire3]` list)
```

direwolf420 (commented):
Small review, mostly doc fixes

patches/tModLoader/Terraria/ID/NPCID.cs.patch
JavidPack: It's fine, they were already there and it could help with code maintenance being able to consult with original code docs

patches/tModLoader/Terraria/NPC.TML.cs
JavidPack: I'll make a note, but probably not needed for now.

PR #3465, opened 2023-05-09
Update NetClientSocialModule.cs.patch

### LINKED ISSUE #3456: Tmodloader crashes, unable to load into menu even WITHOUT any mods.
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

2 weeks ago I started playing Tmodloader with my friends in the 1.4-stable version. Everything was fine until today that I was not able to get the game to load into main menu. The fatal error message (image provided) suggested that it was due to a mod's fault so I removed all the mods and still it's showing the same message. Since I wasn't able to load into menu to remove the mods, I had to remove them from steam workshop (image provided)[The mod folder in my Document is also empty as seen (images provided)]. Even then the game still says that it's due to a mod's fault for crashing. I have yet to seen any similar situation so I have no idea what to do. Please help.

The mods I was using in Tmodloader 1.4-stable were (image provided)
image
image
image
image
image
image

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/11412736/client.log)

### Steps to reproduce

- Click "play" Tmodloader on steam
- Game crashes

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

Every time I try to open the game

### Additional Information

I can run my Terraria without any problem whatsoever but every time I try to launch Tmodloader it crashes before being able to  load into main menu.

JavidPack:
Fixes #3456

Will need to be ported to stable as well, once finalized.

The underlying issue, a steam client beta change, has not been addressed.

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/NetClientSocialModule.cs.patch
+++ b/patches/tModLoader/Terraria/Social/Steam/NetClientSocialModule.cs.patch
@@ -37,11 +37,12 @@
  			Main.OpenPlayerSelect(delegate (PlayerFileData playerData) {
  				Main.ServerSideCharacter = false;
  				playerData.SetAsActive();
-@@ -141,6 +_,14 @@
+@@ -141,6 +_,15 @@
  				Main.statusText = Language.GetTextValue("Social.JoiningFriend", friendName);
  				_lobby.Join(result.m_steamIDLobby, OnLobbyEntered);
  			});
 +
++			// TODO: Consider adding this logic to Main.OpenPlayerSelect as well. Also add checks for any other async task we don't want interrupted.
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Suggest applying the fix to `Main.OpenPlayerSelect` instead. Looking at some history on the Re-Logic side, the guards in `OpenPlayerSelect` make some effort to prevent a join interrupting something important. We should consider if there's any modded menus with async tasks we also want to block join requests in. 

Delaying the join request till the end of mod loading is the right thing to do, so well done there.

Looking at the uses of `OnSuccessfulLoad`, I can see some potential issues if multiple callbacks get registered.

PR #3466, opened 2023-05-09
ModPrefix.GetTooltipLines, examples, plus inheritable localization example

JavidPack:
Prefixes typically affect stats, many of those stats are automatically handled, but custom stats are not. Modders currently need to use ModifyTooltipLines in GlobalItem to add tooltip lines for every prefix added in the mod. 

This approach can get messy, it would be nicer to have a way to declare the tooltip lines in the modprefix class itself, this PR does that.

This PR also shows an example of https://github.com/tModLoader/tModLoader/wiki/Localization#inheritable-localized-properties and an example of sharing the same key between inherited classes. These examples are lacking from ExampleMod since only ExamplePrefix uses inheritance for content.


Chicken-Bones (approved):
LGTM

Issue #3438, opened 2023-04-27
Example Minecart?


tomiscool850:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Just like a Mount, except it only activates on tracks and dosent have the buff

direwolf420:
I would be willing to contribute this (in a few weeks). Just need to look into the new Minecart Upgrade Kit item, and find someone to make the sprites

Chicken-Bones:
Desaturating a vanilla sprite and increasing contrast is a perfectly acceptable method of making example sprites :)

JavidPack:
I noticed that vanilla minecarts have a left and right buff, I imagine it is necessary for some multiplayer syncing. Make sure to look into that, try to figure out why it is necessary.

tomiscool850:
epic, also ik i said with no buff, i didnt know minecarts have buffs

PR #3477, opened 2023-05-14
ExampleMod: Example Minecart

direwolf420:
Implements #3438.

Showcases some advanced localization retrieval (all cart descriptions are the same).

Added docs for `Item.DefaultToMount`, but it's currently a misnomer. 1.4.5 will likely rename it as only carts use it.

Publicised `Mount.SetAsHorse` and `Mount.SetAsMinecart` as the latter is very useful here. The latter was also adjusted to match future 1.4.5 signature of only 1 buff as opposed to 2 buffs.

Made `Mount.MinecartDirectional` internal as it will not be used in 1.4.5 any more and to make porting easier.

Moved ExampleMountItem to the new Mounts folder as there's now two mounts in ExampleMod.

Interaction with the [Minecart Upgrade Kit](https://terraria.wiki.gg/wiki/Minecart_Upgrade_Kit) is awkward because that item _overwrites_ any changes made to movement related properties. If you want to have your changes scale properly, you need to use `ModPlayer.PostUpdateRunSpeeds`, which I didn't showcase here.

I could add more fields to edit for minecarts, or make some of the helper methods vanilla uses public (dust spawn, delegates).

### Porting Notes
If you use `BuffID.Sets.BasicMountData` you now don't have to also set `Main.buffNoTimeDisplay`/`buffNoSave`.

### Old PR Text
<details>
  <summary>Old PR Text not made with known changes to 1.4.5 in mind</summary>
  
Implements #3438.

Introduced a pattern for buff loading similar to `ExampleTrap` as minecarts require two buffs ("Left" and "Right"). _Why_ it is necessary (and wasn't necessary before on 1.3), I don't know. Hoping that a Re-Logic employee might shed some light on this.
The split gives an opportunity to showcase some advanced name/texture/localization retrieval, which is showcased there.

Added docs for `Item.DefaultToMount`. Thing is that not all vanilla mounts are using this. This method doesn't make the item itself useable, so half the mounts can be used as an item, and the other half can't (the latter includes minecarts, so it's used in this example, but it's not used on the other mount example) 

Publicised `Mount.SetAsHorse` and `Mount.SetAsMinecart` as the latter is very useful here.

Moved ExampleMountItem to the new Mounts folder as there's now two mounts in ExampleMod.

Interaction with the [Minecart Upgrade Kit](https://terraria.wiki.gg/wiki/Minecart_Upgrade_Kit) is awkward because that item _overwrites_ any changes made to movement related properties. If you want to have your changes scale properly, you need to use `ModPlayer.PostUpdateRunSpeeds`, which I didn't showcase here.

I referred to `ExampleTrap` for the buff (loader) as that explains everything in detail. I could add more fields to edit for minecarts, or make some of the helper methods vanilla uses public (dust spawn, delegates).
  
</details>


Chicken-Bones:
There's two buffs because... it's a way to net sync cart sprite direction (which is based on grapple hook angle). Does it need to be this way? No. I'll see about simplifying it for 1.4.5

JavidPack:
CB, should we merge this now and simplify later, or just wait?

Chicken-Bones:
~~I think that it's not likely to be 'fixed' in vanilla, but there's no reason not to do it in tML. It's very simplifiable~~

Fix coming in 1.4.5

Chicken-Bones:
Notes from last time I looked at double'd up minecart buffs:
- It looks to me that they could be eliminated if `Player.minecartLeft` was synced in the `PlayerControls` message
- All they do is flip the minecart sprite draw based on initial grappling hook direction. Bumpers don't even flip the cart sprite

direwolf420:
Implemented suggested changes as per discord https://discord.com/channels/103110554649894912/445276626352209920/1125311655052591165
I left a few //TODO 1.4.5 to check when that arrives as those things are likely gonna break

ExampleMod/Content/Buffs/ExampleMinecartBuff.cs
```diff
@@ -0,0 +1,31 @@
...
+		public override void SetStaticDefaults() {
+			Main.buffNoTimeDisplay[Type] = true;
+			Main.buffNoSave[Type] = true;
+		}
+
+		public override void Update(Player player, ref int buffIndex) {
+			//TODO 1.4.5: review new Set for buff -> mount mapping, ExampleMountBuff too
+			player.mount.SetMount(ModContent.MountType<ExampleMinecartMount>(), player);
+			player.buffTime[buffIndex] = 10;
+		}
```
Chicken-Bones: Is this preferred over setting `BasicMountData` in `SetStaticDefaults`?

ExampleMod/Content/Buffs/ExampleMinecartBuff.cs
```diff
@@ -0,0 +1,31 @@
...
+		//public override LocalizedText DisplayName => Language.GetText("BuffName.MinecartLeft");
+		// But for the sake of example, we want to reuse the item name
+		public override LocalizedText DisplayName => ModContent.GetInstance<ExampleMinecart>().DisplayName;
+
+		// Use the vanilla Description
+		public override LocalizedText Description => Language.GetText("BuffDescription.MinecartLeft");
+
+		public override void SetStaticDefaults() {
+			Main.buffNoTimeDisplay[Type] = true;
+			Main.buffNoSave[Type] = true;
```
Chicken-Bones: In vanilla, these are automatically set via `BasicMountData` in `Main.Initialize_TileAndNPCData1`. Should we add a tML method to `BuffLoader` which copies these values across in the same manner in-between `SetupContent` and `PostSetupContent`?

To be clear, `BasicMountData` is what's being replaced with `BuffID.Sets.MountType` in 1.4.5 which will have the same behavior (just minus the direction flag).

ExampleMod/Content/Items/Mounts/ExampleMinecart.cs
```diff
@@ -0,0 +1,27 @@
...
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Items.Mounts
+{
+	public class ExampleMinecart : ModItem
+	{
+		public override void SetDefaults() {
+			Item.DefaultToMount(ModContent.MountType<ExampleMinecartMount>());
```
Chicken-Bones: Is there really any point calling this if you're going to override `width` and `height` anyway? Just set `mountType = ` directly

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -1463,6 +1463,22 @@
...
++	/// <summary>
++	/// This method sets a variety of Item values common to mount items.<br/>
++	/// Specifically: <code>
++	/// width = 36;
++	/// height = 26;
++	/// mountType = mount;
++	/// </code>
++	/// </summary>
++	/// <param name="mount">The mount type</param>
+ 	public void DefaultToMount(int mount)
```
Chicken-Bones: Not sure this should be public, or should be called `DefaultToMinecart`, should we refactor+tModPorter?

Chicken-Bones (changes requested):

ExampleMod/Content/Items/Mounts/ExampleMinecart.cs
direwolf420: I would wait until 1.4.5 vanilla has decided what to do with the method. Best bet is just `mountType =` and update the example in 1.4.5 if needed.

patches/tModLoader/Terraria/Item.cs.patch
direwolf420: Same as https://github.com/tModLoader/tModLoader/pull/3477#discussion_r1250996981

ExampleMod/Content/Items/Mounts/ExampleMinecart.cs
Chicken-Bones: Please change to `mountType = `

PR #3480, opened 2023-05-17
Minor Example Mod Changes

Rijam:

- Resprited Example Spear
  - The previous sprite was ugly, had mixels, looked too much like Example Javelin, and didn't match the new Example Mod Sprites.
- Removed Example Spear and Example Robe from Old (Robe was ilija04's suggestion).
  - A new version already exists, so the old one can be removed.
- Added some comments to `ExampleYoyo`, `ExampleYoyoProjectile`, `ExampleWhip`, `ExampleWhipAdvanced`, and `ExampleHook`.
  - For `ExampleYoyo`, comments were added for what the Sets do. Most useful for figuring out what to set `GamepadExtraRange` to.
  - For `ExampleYoyoProjectile`, a comment was added letting people know that some of the vanilla yoyos are slightly larger.
  - For `ExampleWhip` and `ExampleWhipAdvanced`, comments were added better explaining what the measurements meant. That way it is more clear on what each of the numbers mean and how to find them for your sprite.
  - For `ExampleHook`, added comments stating to set certain values if CloneDefaults isn't used. (direwolf's suggestion)
- The Example Quest Fish can now be placed in a weapon rack.
  - All vanilla fish can be placed in a weapon rack, and it was missing from Example Quest Fish.
- Simplified `ExampleMagicMirror`'s grappling hook removal code (direwolf's suggestion).
- Simplified `MinionBossBag` by removing all of the drawing code. That is already handled automatically by the BossBag set. (direwolf's suggestion).
  - I also updated the name of the bag to include the name of the boss in it. This change was made to all vanilla bags in 1.4.3
- Added the `HidesHands[]` Set to `ExampleRobe` with a comment saying the set doesn't actually have an effect unless `HideArms[]` is also true.
- Removed `includePDB = true` in the `build.txt` (direwolf's suggestion).
- Updated the summary/docs of `ItemID.Sets.BossBag[]` to say that it also creates a glow around the item when dropped in thw orld. (direwolf's suggestion).
- Fixed several typos in `ItemID.Sets`, `NPCID.Sets`, `ProjectileID.Sets` and `ModHair`. This could be considered outside of the scope of this PR, so let me know if these changes should be their own PR.

No

Let me know if I should split these changes into multiple PRs.


direwolf420:
Some more EM changes I couldn't be bothered to make a PR for:
replace
https://github.com/tModLoader/tModLoader/blob/0e4e99d86abf1a8353ebf080f75c643944d7a796/ExampleMod/Content/Items/Tools/ExampleMagicMirror.cs#L47-L54
with
`player.RemoveAllGrapplingHooks();`

remove includePDB line in build.txt, it's unused

ilija04:
Remove ExampleRobe from the Old directory, because it's already implemented

Chicken-Bones (approved):

JavidPack:
ExampleRobe in the old folder set `drawHands` to true in the `DrawHands` hook. According to tmodporter, `"After registering this as EquipType.Body, use ArmorIDs.Body.Sets.HidesHands[slot] = false if you had drawHands set to true. If you had drawArms set to true, you don't need to do anything"`, which suggests we should be setting `ArmorIDs.Body.Sets.HidesHands[slot] = false`. I'm not sure if that is still the case, if you could test it, that would help. It looks like other robes set it to false.

Rijam:
Even though `ArmorIDs.Body.Sets.HidesHands[Item.bodyslot]` is true by default, it doesn't actually hide the hands unless `ArmorIDs.Body.Sets.HidesArms[Item.bodySlot]` is also true. `ArmorIDs.Body.Sets.HidesArms[]` defaults to false, so the hands are still shown. It just happened to work out for Example Robe. Should we still add the set with a comment about HidesArms too?

steviegt6:
> Even though `ArmorIDs.Body.Sets.HidesHands[Item.bodyslot]` is true by default, it doesn't actually hide the hands unless `ArmorIDs.Body.Sets.HidesArms[Item.bodySlot]` is also true. `ArmorIDs.Body.Sets.HidesArms[]` defaults to false, so the hands are still shown. It just happened to work out for Example Robe. Should we still add the set with a comment about HidesArms too?

I see no reason not to; the more information, the merrier (in cases like this anyway).

PR #3494, opened 2023-05-23
Implement selective TownNPCMood key registration.

JavidPack:
As requested, only register TownNPCMood keys that will be used.

Also, auto migrates legacy TownNPCMood keys to new key pattern. Resulting entry will be a substitution pointing to original key. Modders will still want to eventually move the text, but they don't need to immediately to get the mod working. (Can't put actual translation text because that would only affect the current language.)

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
@@ -159,9 +159,13 @@ public static class NPCLoader
 			keys.Add("Princess_LovesNPC");
 
 			foreach (var key in keys) {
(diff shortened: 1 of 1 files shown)


Chicken-Bones (approved):
This is the best you can do without a rework to the whole system. I know thomas started a rework draft, but shop price preferences seem like a pretty minor thing overall. 

I don't _like_ how separate this is from `ShopHelper` but everything in the shop/preference system is `internal/private`, so mods can't extend from it to add their own preference factors/localizations anyway.

Also, I guess mods might want `Transformed` (shimmered) custom affection text too? But that's currently hardcoded to vanilla bestiary girl.

PR #3501, opened 2023-05-27
Example Rockets

Rijam:
### What is the new feature?

Improvements to modded rocket ammo.

* Examples of using `AmmoID.Sets.SpecificLauncherAmmoProjectileMatches` to provide weapon specific projectiles for specific ammo items, typical of rocket launchers but useable by any weapon.
* `AmmoID.Sets.SpecificLauncherAmmoProjectileFallback` allows for weapons to "inherit" ammo specific projectiles from other weapons, facilitating mod compatibility for these types of weapons.
* `ProjectileID.Sets.Explosive[]` to handle all of the things that Projecitle AI Style 16 (Explosive) did.
  * Sets the timeLeft to 3 and the projectile direction when colliding with an NPC or player in PVP (so the explosive can detonate).
  * Shimmer interaction: bounce off of the top of Shimmer and explode with no blast damage when hitting the sides or bottom.
  * Damage other players in the blast radius in For the Worthy worlds.
*  Expose `Projectile.PrepareBombToBlow` and add Mod/GlobalProjectile.PrepareBombToBlow hooks to facilitate modded explosives, allowing explosives to properly adjust properties when exploding.
* `Projectile.HurtPlayer(Rectangle hitbox)`. A simplified copy of `Projectile.BombsHurtPlayers()` which damages the local player if they intersect with the hitbox.

Rocket ammo is very strange and isn't as intuitive as arrow and bullet ammo. Rocket ammo was quite difficult to deal with because each launcher in vanilla has its own behavior depending on the ammo type. Interaction between mods with custom rocket ammo types were also questionable.

Added Example Rocket based on Rocket I which includes `ExampleRocket` item, `ExampleRocketLauncher` item, `ExampleRocketProjectile`, `ExampleGrenadeProjectile`, `ExampleProximityMineProjectile`, and `ExampleSnowmanRocketProjectile`.

Even though this is based on Rocket I, I included a few comments about things from Rocket III, Mini Nuke I, and how to destroy tiles like Rocket II.

This PR also updates ExampleExplosive: Updating its sprite, adding Shimmer interaction, adding the new `ProjectileID.Sets.PlayerHurtDamageIgnoresDifficultyScaling` set, and making the projectile damage the player.

- I thought the `ProjectileID.Sets.Explosive` would handle damaging other players in For the Worthy worlds, but it seems like it doesn't because the vanilla `Kill()` isn't called on ModProjectiles? Might need some more investigation. So, the parts in the projectiles that resize the projectile and damage the player in For the Worthy worlds may or may not be necessary.

<details><summary>(old PR things)</summary><blockquote>

* ~Default projectile type for rocket ammo. Setting `Item.shoot = ModContent.ProjectileType<T>()` in the rocket ammo will now make that projectile the default project to use if `ProjectileID.Sets.SpecificLauncherAmmoProjectileMatches` isn't defined for this ammo type. Previously, setting `Item.shoot` would result in the game trying to grab the wrong projectile type, likely one that was greater than the number of loaded projectiles.~

- ~~If you check the vanilla `Shimmer()` method, you might see Explosive AI projectiles also have this code:~~
 ```cs
if (timeLeft > 3000)
	timeLeft -= 60;

velocity.Y -= 0.5f;
```
~~The `velocity.Y -= 0.5f` is supposed to make the projectile "fall upwards", but vanilla rockets don't do this. If you shoot the rocket at the top layer of some Shimmer where it doesn't hit the surface of the Shimmer but is above the other full Shimmer tiles below it, the rocket (and grenades, and mines) will continue on their normal path and don't "fall upwards". Idk if this is a bug or if I am misunderstanding the source code.~~ Vanilla bug where the second check for aiStyle 16 is never hit (pointed out by direwolf). May or may not be fixed in future Terraria versions.

<details><summary>Gif showcase of Shimmer</summary><blockquote>

Vanilla Rocket I. It bounces off of the top, explodes if it hits a full tile, but in between it just flies straight.
VanillaRocketIShimmer2

Here is an example of what it would've looked like if the projectiles did "fall upwards".
ExampleFallingUpShimmer
</blockquote></details>

- ~~PVP is weird. Direct hits work fine, but hits from the blast damage from the right are weird. If the shooter shoots to the right of victim hitting a tile or NPC, the blast damage will cause the victim to be knocked back to the right in the shooter's view. The victim experiences knockback correctly. After a second or so, the victim's correct location is synced to the shooter causing them to teleport. Before the position is synced, though, the shooter can still shoot the victim at the position on their (shooter's) screen. To the victim it looks like the shooter is completely missing, but they (victim) is still being hit. Interestingly, this same issue doesn't happen in For the Worthy and Get Fixed Boi worlds where explosives can damage friendly players (it only happens when PVP is enabled). I also noticed that damage numbers are different for the shooter and victim. Is there something that I missed or should we just chalk this up to "PVP is unsupported".~~ The knockback seems to work correctly now.

<details><summary>Gif showcase of PVP (old)</summary><blockquote>

Shooter on the left, victim on the right.
https://i.imgur.com/Hp1qlgj.png
</blockquote></details>

- Just a comment style question: is putting comments in between if else statements ok?
```cs
if (thisBool) {
    //... do stuff
}
// Is a comment right here ok?
else {
    //... do other stuff
}
```

</details>

Shouldn't contain any breakage. Any projectile using Projectile AI Style 16 will still work as before. Using `PickAmmo()` instead of `ProjectileID.Sets.SpecificLauncherAmmoProjectileMatches` should still work.

#### Short Summary
* `ProjectileID.Sets.Explosive[]` to match the behavior of vanilla explosives without having to use aiStyle 16
* New `Projectile.HurtPlayer(Rectangle hitbox)` method which will damage the local player if they intersect the hitbox.
* New `PrepareBombToBlow` hook to implement explosives correctly.
* A full set of rocket projectiles as well as rocket launcher and rocket ammo items, showing how they all connect to each other.

#### Porting Notes
* If you were using `Projectile.aiStyle = ProjAIStyleID.Explosive` (`16`) to match the vanilla behavior of explosives, you can now use `ProjectileID.Sets.Explosive[]`. Also override `PrepareBombToBlow` and add the explosion resizing logic there.
* Rocket ammo should have `ProjectileID.Sets.SpecificLauncherAmmoProjectileMatches` defined for all applicable vanilla launchers.
* Rocket launchers should set `AmmoID.Sets.SpecificLauncherAmmoProjectileFallback` to the closest vanilla launcher to inherit ammo-specific projectiles from.
* If you were using `PickAmmo()` or similar to fix your rocket ammo shooting the wrong projectile for the Snowman Cannon and Celebration Mk2, double check that because it is probably not necessary anymore.


JavidPack:
Well done, thanks for keeping this alive, it certainly was quite complicated and the examples are great.

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1601,6 +1601,32 @@
...
++	/// Used to adjust projectile properties immediately before the projectile becomes an explosion. This is called on projectiles using the <see cref="ProjAIStyleID.Explosive"/> aiStyle or projectiles that are contained in the <see cref="ProjectileID.Sets.Explosive"/> set. By defaults tileCollide is set to false and alpha is set to 255.
++	/// <para/> Calls the <see cref="ProjectileLoader.PrepareBombToBlow(Projectile)"/> method.
++	/// </summary>
+-	private void PrepareBombToBlow()
++	public void PrepareBombToBlow()
+ 	{
+ 		tileCollide = false;
++		if(type >= ProjectileID.Count) // This seems unused and could cause issues for modded projectiles.
+-		ai[1] = 0f;
++			ai[1] = 0f;
```
Chicken-Bones, line 1618: @JavidPack is this logic inverted?

JavidPack: Yep, thanks, fixed.

Issue #3537, opened 2023-06-12
Update FNA3D to latest


BAFAT76:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

hello I can't launch it, each time i try it open cmd and after that an error, i don't have any mods on it.
please i really want to play on it again. 

### Log File

[16:32:45.049] [Main Thread/INFO] [tML]: Starting tModLoader client 1.4.4.9+2023.06.23.0|2022.09|preview|Preview|5952f4eb397287e8ac3f246f6945c1a5348de3ad|5249906444923372851 built 11/06/2023 03:10
[16:32:45.072] [Main Thread/INFO] [tML]: Log date: 12/06/2023
[16:32:45.072] [Main Thread/INFO] [tML]: Running on Windows (v10.0.19044.0) X64 NetCore 6.0.0
[16:32:45.073] [Main Thread/INFO] [tML]: FrameworkDescription: .NET 6.0.0-rtm.21522.10
[16:32:45.073] [Main Thread/INFO] [tML]: Executable: C:\Program Files (x86)\Steam\steamapps\common\tModLoader\tModLoader.dll
[16:32:45.073] [Main Thread/INFO] [tML]: Working Directory: C:\Program Files (x86)\Steam\steamapps\common\tModLoader
[16:32:45.074] [Main Thread/INFO] [tML]: Launch Parameters:  /gldevice:openGL
[16:32:45.076] [Main Thread/INFO] [tML]: Parsed Launch Parameters: 
[16:32:45.972] [Main Thread/INFO] [tML]: Saves Are Located At: C:\Users\Boby_le_bob\Documents\My Games\Terraria\tModLoader-preview
[16:32:46.053] [Main Thread/INFO] [tML]: Display DPI: Diagonal DPI is 96. Vertical DPI is 96. Horizontal DPI is 96
[16:32:46.053] [Main Thread/INFO] [tML]: High DPI Display detected: setting FNA to highdpi mode
[16:32:46.056] [Main Thread/DEBUG] [FNA]: Querying linked library versions...
[16:32:46.056] [Main Thread/DEBUG] [FNA]: SDL v2.24.0
[16:32:46.070] [Main Thread/DEBUG] [FNA]: FNA3D v22.8.0
[16:32:46.087] [Main Thread/DEBUG] [FNA]: FAudio v22.9.1
[...]

PR #3540, opened 2023-06-12
Fix for #3537

LittleJay1985:
**What is the bug?**
#3537 - System.InvalidOperationException: No supported FNA3D driver found!
Crash on launch when unsupported CmdLineArg '/gldriver:opengl' is used on windows.

**How did you fix the bug?**
Added a check for '/gldriver:opengl' to ScriptCaller.sh when launched under Windows.
If found, removes the argument from the CommandLineArgs whilst keeping any others.

**Are there alternatives to your fix?**
Current fix in FNAFixes.cs acts on Environment.GetEnvironmentVariable(SDL_VIDEODRIVER) correctly.
But as FNA uses Enviroment.GetCommandLineArgs() Method directly and is called after TModLoaders FNAFixes.Init(), the fix would have to be in FNA's codebase.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
@@ -86,11 +86,30 @@ else
 	fi
 fi
 
-if [[ -f "$install_dir/dotnet" || -f "$install_dir/dotnet.exe" ]]; then
-	echo "Launched Using Local Dotnet"  2>&1 | tee -a "$LogFile"
-	[[ -f "$install_dir/dotnet" ]] && chmod a+x "$install_dir/dotnet"
-	exec "$install_dir/dotnet" tModLoader.dll "$customargs" "$@" 2>"$NativeLog"
+if [[ "$_uname" == *"_NT"* ]]; then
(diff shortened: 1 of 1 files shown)


LittleJay1985:
This could just be a temporary fix as FNA3D apparently supports OpenGL and Vulkan on windows.
I'll look into it further when i have the chance.

Solxanich:
Has to be xplicitly `/gldevice:OpenGL`
As FNA is crossplatform, it is also case sensitive. 

More info can be found in the FNA docs on that case sensitivity.

I don't think we need to fix this, might be better to educate around?

I'll let ya ponder if want to keep going

LittleJay1985:
Looks to be an issue in FNA3D using SDL_strcmp. If the argument is not capitalised exactly as '/gldriver:OpenGL' it will not find any render drivers.

Please hold off on a merge for now.

LittleJay1985:
Pull request sent https://github.com/FNA-XNA/FNA3D/pull/170
And Merged https://github.com/FNA-XNA/FNA3D/commit/d30147935f4b93f59e73952b713fc6116b263d6d

I'll create a fix for when FNA is updated.

LittleJay1985:
Closing as not needed

PR #3543, opened 2023-06-13
Update DedicatedServerUtils to match Main serverconfig.txt

Drackrath:


Chicken-Bones:
We're considering removing the copy to prevent this sort of desync happening, and instead updating the instructions to tell people where to grab it. Please provide input on #3555

PR #3545, opened 2023-06-13
Migrate users to 1.4.3-legacy in preparation for 1.4.4 default

JavidPack:
Still a lot todo...

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -513,36 +513,23 @@ namespace Terraria.ModLoader.Core
 			// TODO: This logic is incorrect. I think sometimes we want to know every mod, and othertimes we want to know every mod with a version we can load.
 			// The tmods.Length == 1 check and the logic below contradict each other, sometimes returning null when no loadable mods and othertimes returning a mod path.
 			Version tmodVersion = new Version(BuildInfo.tMLVersion.Major, BuildInfo.tMLVersion.Minor);
-			string[] tmods = Directory.GetFiles(repo, "*.tmod", SearchOption.AllDirectories);
-			if (tmods.Length == 1)
-				return tmods[0];
-
-			string val = null;
-			Version currVersion = null;
-			foreach (string fileName in tmods) {
-				var match = PublishFolderMetadata.Match(fileName);
-
-				if (match.Success) {
-					Version testVers = new Version(match.Groups[1].Value);
-					if (testVers > tmodVersion) {
-						if(testVers >= new Version(2023, 4))
-							modsReadyFor144.Add(Path.GetFileNameWithoutExtension(fileName));
-						continue;
-					}
-					else if (testVers == currVersion) {
-						val = fileName;
-					}
-					else if (testVers > currVersion) {
-						currVersion = testVers;
-						val = fileName;
-					}
-				}
-				else if (val == null) {
-					val = fileName;
-					currVersion = new Version(0, 12);
-				}
+			var information = AnalyzeWorkshopTmods(repo);
+			if (information == null) {
+				Logging.tML.Warn($"Unexpectedly missing .tMods in Workshop Folder {repo}");
+				return null;
 			}
-			return val;
+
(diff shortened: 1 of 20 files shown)


patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -31,6 +31,7 @@ public static class ModLoader
 		public static string LastLaunchedTModLoaderAlphaSha;
 		public static bool ShowWhatsNew;
 		public static bool AlphaWelcomed;
+		public static bool SeenMigrateTo143BranchMessage;
```
Chicken-Bones, line 34: We want to show it on _every_ launch for people who aren't on the branch. GoG players don't matter, but you should use the result from `SteamApps.GetCurrentBetaName` to control the visibility of the message

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -451,5 +467,23 @@ internal static class Interface
...
+			bool OnBetaBranch = Steamworks.SteamApps.GetCurrentBetaName(out string branchName, 1000);
+			if (OnBetaBranch)
+				return false;
+
+			// User is on Stable. We need to check if this program is the actual one in the install dir or not.
+			string currentDirectory = Path.GetFullPath(Directory.GetCurrentDirectory());
+			Steamworks.SteamApps.GetAppInstallDir(Engine.Steam.TMLAppID_t, out string tModLoaderInstallLocation, 1000);
+
+			if(!string.Equals(currentDirectory, tModLoaderInstallLocation)) // TODO: Is there a better way?
+				return false;
```
Solxanich, line 484: Use case?
We want all 1.4.3 users, independent of if they installed via default branch on steam or manually to use 1.4.3-legacy builds going forward.
Especially since it involves save files.
Would recommend we instead check if the tML version is greater than 2022.09.47.62 and return false

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -508,6 +510,8 @@ private enum SearchFolders { }
 		private static readonly Regex PublishFolderMetadata = new Regex(@"[/|\\]([0-9]{4}[.][0-9]{1,2})[/|\\]");
 
 		internal static string GetActiveTmodInRepo(string repo) {
+			// TODO: This logic is incorrect. I think sometimes we want to know every mod, and othertimes we want to know every mod with a version we can load.
+			// The tmods.Length == 1 check and the logic below contradict each other, sometimes returning null when no loadable mods and othertimes returning a mod path.
 			Version tmodVersion = new Version(BuildInfo.tMLVersion.Major, BuildInfo.tMLVersion.Minor);
```
Solxanich, line 515: This logic should probably be rewritten and refactored with the CleanupPublishFolder logic.

Specs on this function:
- tml version 0.12 is a direct file path to .tmod (no folder). Is a legacy case
- for purposes of mod loading, we want the latest .tmod for the tml version running
- if there is no .tmod for this version or earlier, we return null (ie no mod exists that can be loaded). Should be logged

patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
```diff
@@ -100,20 +102,45 @@ internal class UIModItem : UIPanel
 			_uiModStateText.OnClick += ToggleEnabled;
 			Append(_uiModStateText);
 
+			if (!BuildInfo.IsDev && BuildInfo.tMLVersion < _mod.tModLoaderVersion) {
+				string updateVersion = $"v{_mod.tModLoaderVersion}";
+				string updateURL = "https://github.com/tModLoader/tModLoader/releases/latest";
+				bool lastMonth = BuildInfo.tMLVersion.Minor == 12;
+				if (BuildInfo.IsStable && new Version(_mod.tModLoaderVersion.Major, _mod.tModLoaderVersion.Minor) == new Version(BuildInfo.tMLVersion.Major + (lastMonth ? 1 : 0), BuildInfo.tMLVersion.Minor + (lastMonth ? 0 : 1)))
```
Solxanich: use `> BuildInfo.StableVersion`

Solxanich (changes requested):
A few changes to simplify some items

Solxanich (approved):

PR #3552, opened 2023-06-16
Extra Jump API

absoluteAquarian:
### What is the new feature?
This pull request adds an actual API for adding modded extra jumps.  
The changes and additions that this API implements are as follows:
- `ExtraJumpState`
  - Stores the state of an extra jump for a player
  - Contains three properties:
    - `bool Enabled { get; }`
      - Set by `ExtraJumpState.Enable()`, cleared by `ExtraJumpState.Disable()`
    - `Available { get; set; }`
      - Indicates if an extra jump is ready to use (`true`) or has already been consumed (`false`). Jumps are consumed when they start
    - `Active { get; }`
      - Indicates whether the extra jump is currently being performed
    - `void Enable()`
      - Sets the extra jump as available to use, typically in equipment/accessory logic
      - `Disable()` overrides `Enable()` regardless of call order
    - `void Disable()`
      - Forcibly prevents using the extra jump for the current game tick. Consumes it if it was `Available`, and stops it if it was `Active`
- `ExtraJump`
  - A singleton `ModType` containing hooks related to performing visual effects for the jump and manipulating the player
  - `ExtraJump.Position GetDefaultPosition()`
    - Used to place the extra jump between **vanilla** extra jumps
    - If a non-vanilla extra jump is not returned here, an exception is thrown
    - There are a few predefined positions for modder convenience: `BeforeMountJumps`, `MountJumpPosition`, `BeforeBottleJumps` and `AfterBottleJumps`
  - `IEnumerable<ExtraJump.Position> GetModdedConstraints()`
    - Used to place the extra jump before or after other **modded** extra jumps that were sorted to the same "location" from `GetDefaultPosition()`
  - `void Visuals(Player player)`
    - Used to create effects while the extra jump is in progress, e.g. the dusts from the Blizzard in a Bottle's extra jump
  - `float GetDurationMultiplier(Player)`
    - Used to set the duration – or "strength" – of the extra jump
  - `void OnStarted(Player)`
    - Executes at the start of the extra jump
  - `void OnEnded(Player)`
    - Executes at the end of the extra jump, either from its duration expiring or when it is forcibly cancelled
  - `void UpdateHorizontalSpeeds(Player)`
    - Runs after `ModPlayer.PostUpdateMiscEffects()`, but before `ModPlayer.PostUpdateRunSpeeds()`
    - Used to adjust the horizontal speed variables (typically `runAcceleration` and `maxRunSpeed`) while the extra jump is in progress
  - `void OnRefreshed(Player)`
    - Executes whenever the extra jump is refreshed, typically when landing or being reeled in by a grappling hook
- `ExtraJumpLoader`
  - The corresponding loader for `ExtraJump`
- Changes to `Terraria.Player`:
  - All fields related to vanilla extra jumps have been integrated into a new `private ExtraJumpState[] extraJumps` field
    - Can be accessed via the `public Span<ExtraJumpState> ExtraJumps` property
    - Accessing the underlying array is similar to how `DamageClass` works:
      - `public ref ExtraJumpState GetJumpState<T>(T baseInstance) where T : ExtraJump`
      - `public ref ExtraJumpState GetJumpState<T>() where T : ExtraJump`
  - Flipper swimming has been reimplemented as an "extra jump" in this new API
    - The `accFlipper` field has been turned into a property which references `GetJumpState(ExtraJump.Flipper)`
  - The `sandStorm` field was turned into a property which indirectly depends on `GetJumpState(ExtraJump.SandstormInABottle).Active`
    - The reason for this change was due to [a gist](https://gist.github.com/absoluteAquarian/8a76bba28848ea5918bda5a26acf2f9a) documenting its purpose, which was essentially the same as whether the _Sandstorm in a Bottle_ jump is currently being performed (aka `Active`)
  - New methods/fields for facilitating mass checking and manipulation of extra jumps:
    - `bool AnyExtraJumpUsable()`
      - Returns `true` if any extra jump's `Available` property returns `true`
      - Setting `blockExtraJumps` (see below) to `true` will force this method to return `false`
    - `void ConsumeAllExtraJumps()`
      - Sets the `Available` property in all extra jumps to `false`
    - `bool blockExtraJumps`
      - Prevents extra jumps from being performed when set to `true` but does not consume them, or cancel an in progress jump
    - `void StopExtraJumpInProgress()`
      - Stops the current extra jump if one is active
  - `void RefreshAllExtraJumps()`
    - Previously was a `private` method named `RefreshDoubleJumps()`
    - Sets the `Available` property to `true` in any extra jumps that are `Enabled`
    - **NOTE:** disabled jumps via `ExtraJumpState.Disable()` cannot be refreshed
- Several new hooks to `ModPlayer` which are invoked in tandem with the corresponding hooks in `ExtraJump`

Currently, adding an extra jump takes at minimum 3 IL edits and 3 method detours in order to properly implement, and doing so can easily clash with other mods.

None that I can think of.

Indicating that a `ExtraJump` is available for use in `ModItem::UpdateAccessory()`:
```cs
public override void UpdateAccessory(Player player, bool hideVisual) {
	player.GetJumpState<SimpleExtraJump>().Enable();
}
```
Forcing the _Sandstorm in a Bottle_ extra jump to be disabled in `ModPlayer::PostUpdateMiscEffects()`:
```cs
public override void PostUpdateMiscEffects() {
	player.GetJumpState(ExtraJump.SandstormInABottle).Disable();
}
```
Preventing extra jumps while the player has a debuff
```cs
public virtual void Update(Player player, ref int buffIndex) {
	player.blockExtraJumps = true;
}
```

### ExampleMod updates
ExampleMod now includes the following:
- `Content/Items/Accessories/ExampleExtraJumpAccessory.cs`
  - showcases a basic extra jump
- `Content/Items/Accessories/ExampleMultiExtraJumpAccessory.cs`
  - showcases an advanced jump using a `ModPlayer` to track how many jumps the extra jump has remaining
- `Common/Players/ExampleExtraJumpModificationPlayer.cs`
  - showcases an example of modifying an extra jump

### Additional Notes
If a modder wants to modify the `ExtraJump.UpdateHorizonalSpeeds()` changes from an ExtraJump, they should use the following in ModPlayer:
```cs
public override void PostUpdateRunSpeeds() {
		if (Player.GetJumpState<TheJump>().Active && !Player.Mount.Active) {
				// Revert speed changes here
				// ... then apply custom changes
		}
		// Additional logic goes here
}
```

### Bug Fixes
The extra jump for the Basilisk mount was meant to have the highest priority, but due to a small code error, it only has priority over the Cloud in a Bottle extra jump _**and**_ it ends up being consumed, even if it wasn't actually used.

This PR fixes the bug by restoring the mount's proper priority order (basilisk > goat) and preventing it from being always consumed.


ExampleMod/Content/Items/Accessories/ExampleExtraJumpAccessory.cs
```diff
@@ -0,0 +1,123 @@
...
+			}
+
+			for (int i = 0; i < 3; i++) {
+				Dust dust = Dust.NewDustDirect(spawnPos, player.width, 12, DustID.Snow, player.velocity.X * 0.3f, player.velocity.Y * 0.3f, newColor: Color.Gray);
+				dust.fadeIn = 1.5f;
+				dust.velocity *= 0.6f;
+				dust.velocity -= player.velocity * 0.8f;
+				dust.noGravity = true;
+				dust.noLight = true;
+			}
```
Chicken-Bones: don't like playing spot the difference, copy paste bad, please extract to a method with relevant parameters (`velocityMultiplier, playerVelocityMultiplier`)

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -594,4 +594,40 @@ public void AddImmuneTime(int cooldownCounterId, int immuneTime)
 			hurtCooldowns[cooldownCounterId] += immuneTime;
 		}
 	}
+
+	// Extra jumps
+	private ExtraJumpData[] extraJumps = new ExtraJumpData[ExtraJumpLoader.ExtraJumpCount];
+
+	public ref ExtraJumpData GetExtraJump<T>(T baseInstance) where T : ExtraJump => ref extraJumps[baseInstance.Type];
+
+	public ref ExtraJumpData GetExtraJump<T>() where T : ExtraJump => ref GetExtraJump(ModContent.GetInstance<T>());
```
Chicken-Bones: Should we call the class `State` instead of `Data`? 

Not a big fan of the method returning a different type to what the name implies. `GetJumpState` or `GetJumpData` better perhaps?

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -594,4 +594,40 @@ public void AddImmuneTime(int cooldownCounterId, int immuneTime)
...
+
+	public Span<ExtraJumpData> ExtraJumps => extraJumps.AsSpan();
+
+	internal Span<ExtraJumpData> ModdedExtraJumps => extraJumps[ExtraJumpLoader.DefaultExtraJumpCount..];
+
+	public bool AnyExtraJumpAvailable()
+	{
+		for (int i = 0; i < extraJumps.Length; i++)
+		{
+			ExtraJumpData data = extraJumps[i];
```
Chicken-Bones: `foreach (var data in extraJumps)`

Use `ref var` if you need to modify the value in-loop

patches/tModLoader/Terraria/ModLoader/VanillaExtraJumps.cs
```diff
@@ -0,0 +1,401 @@
...
+		Main.gore[num25].velocity.Y = Main.gore[num25].velocity.Y * 0.1f - player.velocity.Y * 0.05f;
+		num25 = Gore.NewGore(new Vector2(player.position.X - 36f, player.position.Y + (float)num22 - 16f), new Vector2(0f - player.velocity.X, 0f - player.velocity.Y), Main.rand.Next(11, 14));
+		Main.gore[num25].velocity.X = Main.gore[num25].velocity.X * 0.1f - player.velocity.X * 0.1f;
+		Main.gore[num25].velocity.Y = Main.gore[num25].velocity.Y * 0.1f - player.velocity.Y * 0.05f;
+		num25 = Gore.NewGore(new Vector2(player.position.X + (float)player.width + 4f, player.position.Y + (float)num22 - 16f), new Vector2(0f - player.velocity.X, 0f - player.velocity.Y), Main.rand.Next(11, 14));
+		Main.gore[num25].velocity.X = Main.gore[num25].velocity.X * 0.1f - player.velocity.X * 0.1f;
+		Main.gore[num25].velocity.Y = Main.gore[num25].velocity.Y * 0.1f - player.velocity.Y * 0.05f;
+	}
+
+	public override bool PreVisuals(Player player) => (player.canJumpAgain_Sandstorm || !player.hasJumpOption_Sandstorm) && base.PreVisuals(player);
```
Chicken-Bones: comment, what does this do?

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
```diff
@@ -0,0 +1,20 @@
...
+public struct ExtraJumpData
+{
+	/// <summary>
+	/// Whether the extra jump can be used. This field is what should be set by equipment
+	/// </summary>
+	public bool Active;
+	/// <summary>
+	/// Whether the extra jump has been consumed
+	/// </summary>
+	public bool JumpAvailable;
```
Chicken-Bones: Names `Active` and `JumpAvailable` are unclear. Documentation here actually implies inverse of correct use.
Consider 
`Active` -> `Equipped` or `Enabled`
`JumpAvailable` -> `Consumed` (and inverting the associated logic)

Documentation also needs to be clear about _when_ the flags are set. Does the `Consumed` get set immediately on pressing space?

What happens if a jump is unequipped while `PerformingJump`?

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
```diff
@@ -0,0 +1,20 @@
...
+	/// <summary>
+	/// Whether any effects (e.g. spawning dusts) should be performed
+	/// </summary>
+	public bool PerformingJump;
```
Chicken-Bones: If `Active` is the only one meant to be set by modders, the rest should be `{ get; internal set; }`. Consider making them all properties for uniformity/future change proofing.

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Terraria.DataStructures;
+
+/// <summary>
+/// A structure containing fields used to manage extra jumps
+/// </summary>
+public struct ExtraJumpData
+{
+	/// <summary>
+	/// Whether the extra jump can be used. This field is what should be set by equipment
```
Chicken-Bones: Need to document _when_ it should be set. The fact it's automatically cleared in `ResetStats`. If mods are only meant to enable it, then make a method `Enable()`. 

Don't want mods fighting over enable/disable order.

Where are mods meant to disable jumps. I see two use cases:
- Conflict (a modder doesn't want a certain jump to work when another accessory is equipped or action is being performed)
- Temporary lockout to all jumps (debuff, or a state which prevents jumping entirely like being in a tractor beam or something). If the lockout ends while the player still has air-time, they should be able to continue using jumps they haven't yet used.

Ideally need examples of the above/documentation on how to correctly acheive

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
```diff
@@ -0,0 +1,224 @@
...
+			jump.Type = i++;
+			ContentInstance.Register(jump);
+			ModTypeLookup<ExtraJump>.Register(jump);
+		}
+	}
+
+	public static void ModifyPlayerHorizontalSpeeds(Player player)
+	{
+		foreach (ExtraJump moddedExtraJump in orderedJumps) {
+			// Special case: Sandstorm in a Bottle uses a separate flag
```
Chicken-Bones: anything we can do about this?

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
```diff
@@ -0,0 +1,224 @@
...
+			if (data.PerformingJump && data.Active && !data.JumpAvailable && jump.PreVisuals(player) && PlayerLoader.PreExtraJumpVisuals(jump, player)) {
+				jump.Visuals(player);
+				PlayerLoader.ExtraJumpVisuals(jump, player);
+			}
+		}
+	}
+
+	public static void ProcessJumps(Player player, bool flipperOrSlimeMountSwimming)
+	{
+		if (!flipperOrSlimeMountSwimming) {
```
Chicken-Bones: invert if, early return. 

Does this condition belong in here, or outside? Is this an example of one of the the 'all jump lockout' reasons I mentioned above?

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
```diff
@@ -0,0 +1,224 @@
...
+	{
+		foreach (ExtraJump jump in ExtraJumps)
+			player.GetExtraJump(jump).Active = false;
+	}
+
+	internal static void ClearUnavailableExtraJumps(Player player)
+	{
+		foreach (ExtraJump jump in ExtraJumps) {
+			ref ExtraJumpData data = ref player.GetExtraJump(jump);
+			if (!data.Active) {
```
Chicken-Bones: invert if, early return

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
```diff
@@ -0,0 +1,224 @@
...
+			ref ExtraJumpData data = ref player.GetExtraJump(jump);
+			if (data.PerformingJump) {
+				jump.OnEnded(player);
+				PlayerLoader.OnExtraJumpEnded(jump, player);
+				data.PerformingJump = false;
+			}
+		}
+	}
+
+	public static void RefreshExtraJumps(Player player)
```
Chicken-Bones: what if a jump gets equipped mid flight? Should it become available? Probably not I suppose.

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
```diff
@@ -0,0 +1,157 @@
...
+	/// </summary>
+	/// <param name="player">The player performing the jump</param>
+	public virtual void Visuals(Player player) { }
+
+	/// <summary>
+	/// Return <see langword="false"/> to prevent <see cref="Visuals(Player)"/> from executing.<br/>
+	/// By default, this hook returns whether the player is moving upwards with respect to <see cref="Player.gravDir"/>
+	/// </summary>
+	/// <param name="player">The player performing the jump</param>
+	public virtual bool PreVisuals(Player player) {
```
Chicken-Bones: Consider `CanShowVisuals`?

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
```diff
@@ -0,0 +1,157 @@
...
+	/// This occurs at the start of the grounded jump and while the player is grounded.
+	/// </summary>
+	/// <param name="player">The player instance</param>
+	public virtual void OnRefreshed(Player player) { }
+
+	/// <summary>
+	/// This hook runs before the <see cref="ExtraJumpData.JumpAvailable"/> flag for this extra jump is set to <see langword="false"/> in <see cref="Player.Update(int)"/> due to the jump being unavailable or when calling <see cref="Player.ClearAllExtraJumps"/> (vanilla calls it when a mount that blocks jumps is active).
+	/// </summary>
+	/// <param name="player">The player instance</param>
+	public virtual void OnCleared(Player player) { }
```
Chicken-Bones: Please provide a reason to use the hook in the documentation

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
```diff
@@ -0,0 +1,157 @@
...
+	/// Fart in a Jar: runAcceleration *= 3; maxRunSpeed *= 1.75;<br/>
+	/// Goat mount: runAcceleration *= 3; maxRunSpeed *= 1.5;<br/>
+	/// Sandstorm in a Bottle: runAcceleration *= 1.5; maxRunSpeed *= 2;<br/>
+	/// Santank mount: runAcceleration *= 3; maxRunSpeed *= 1.5;<br/>
+	/// Tsunami in a Bottle: runAcceleration *= 1.5; maxRunSpeed *= 1.25;<br/>
+	/// Unicorn mount: runAcceleration *= 3; maxRunSpeed *= 1.5;
+	/// </para>
+	/// </summary>
+	/// <param name="player">The player performing the jump</param>
+	public virtual void ModifyHorizontalSpeeds(Player player) { }
```
Chicken-Bones: Prefer a more general name, perhaps `Update`

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
```diff
@@ -0,0 +1,157 @@
...
+	/// Goat mount: 2<br/>
+	/// Sandstorm in a Bottle: 3<br/>
+	/// Santank mount: 2<br/>
+	/// Tsunami in a Bottle: 1.25<br/>
+	/// Unicorn mount: 2
+	/// </para>
+	/// </summary>
+	/// <param name="player">The player performing the jump</param>
+	/// <returns>A modifier to the player's jump height, which when combined effectively acts as the duration for the extra jump</returns>
+	public abstract float GetDuration(Player player);
```
Chicken-Bones: `GetJumpHeightMultiplier`?

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -594,4 +594,40 @@ public void AddImmuneTime(int cooldownCounterId, int immuneTime)
...
+			if (data.Active && data.JumpAvailable)
+				return true;
+		}
+
+		return false;
+	}
+
+	// An alternative more in line with the vanilla behavior
```
Chicken-Bones: Don't like the need for this, hard to reason about.

Chicken-Bones (changes requested):

ExampleMod/Content/Items/Accessories/ExampleExtraJumpAccessory.cs
absoluteAquarian: noted, i'll abstract it to a method

patches/tModLoader/Terraria/Player.TML.cs
absoluteAquarian: renamed to `ExtraJumpState` and `GetJumpState()`

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: vanilla has one case where they can be publicly refreshed (`Player::RefreshMovementAbilities()`).  every other case is either from standing on the ground or jumping out of a grapple.
should that Player method be made `internal` along with this ExtraJumpLoader method?

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
absoluteAquarian: most users may be able to understand it setting a "duration" easier, but i'm willing to change it if saying it's a "multiplier" would be better

patches/tModLoader/Terraria/Player.TML.cs
absoluteAquarian: this is just to keep the vanilla logic as it was in order to future proof against any inconsistencies (this is also why other quirks were left as-is)

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
absoluteAquarian: removed due to not being able to think of a good use case

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: i don't think that you understand what this method is doing.
the vanilla logic had several blocks of
```cs
if (!hasJumpOption_X)
    canJumpAgain_X = false;
```
which this method just abstracts to a loop

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: leftover from when there was a property to ignore that parameter; there's no need for the parameter anymore

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: frustratingly, no
for whatever reason, `sandStorm` has separate and unique checks and assignments apart from `isPerformingJump_Sandstorm`

patches/tModLoader/Terraria/ModLoader/VanillaExtraJumps.cs
absoluteAquarian: allows modders to override/expand the requirements for making the jump visuals appear

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
absoluteAquarian: renamed `Active` to `Enabled`

`JumpAvailable` is mimicking the vanilla behaviour, which uses this flag for saying whether the jump can be used... aka it's "available" for consumption.  inverting the logic would be a headache for not much gain

`JumpAvailable` gets cleared immediately when the extra jumps starts (aside from contexts where all jumps are cleared)

unequipping an extra jump while `PerformingJump` is true just makes the visuals not appear.  re-equipping the jump while the "jump duration" is still in effect just makes the visuals appear again (this is the vanilla behaviour)

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
absoluteAquarian: all of the corresponding vanilla flags are `public`, so i don't see why `JumpAvailable` and `PerformingJump` should be given limited set accessibility

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
absoluteAquarian: documentation added.

considering adding a `Disable()` method which uses a hidden backing field to make the "value" of `Enabled` (formerly `Active`) appear as `false` even when it's set to `true` by making it a property

this should remedy any conflicts for disabling jumps (setting `Active` to `false` can still disable it, calling `Disable()` forcibly disables it)

patches/tModLoader/Terraria/ModLoader/VanillaExtraJumps.cs
Chicken-Bones: More specifically, why does sandstorm have this special case?

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
Chicken-Bones: Needs to be included in documentation here. `Whether the extra jump has been consumed` implies that it's true if consumed. I don't think people will actually make that mistake, because the variable name is clear, but consider 
```
True if the jump has not yet been used. Will be set to false when a jump starts.
```
also:
```
Remember to also check Enabled
```

patches/tModLoader/Terraria/ModLoader/VanillaExtraJumps.cs
absoluteAquarian: this is from `CloudInABottleJump`, and it's present because the vanilla code in `Player::DoubleJumpVisuals()` also had that for some reason

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
absoluteAquarian: noted.  i'll change the summary

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
Chicken-Bones: duration is fine, but `Multiplier` needs to be in the name. Since `1.0` isn't a duration

patches/tModLoader/Terraria/Player.TML.cs
Chicken-Bones: To keep this maintainable, we need to know why you would want to use this method rather than the one above. I understand "that's what vanilla had" but we're expanding on it here, so we need to be able to fully understand the system (vanilla isn't known for easy to understand code). 

I would expect that checking if jumps are available without an active check would result in potential bugs. Maybe the active check isn't _needed_ in the vanilla calls but it wouldn't hurt? I haven't looked at the logic deeply, but it seems spooky to me.

patches/tModLoader/Terraria/ModLoader/ExtraJump.cs
absoluteAquarian: good point

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
Chicken-Bones: Vanilla doesn't set good precedents for an API. The point of limiting set accessibility is to guide modders away from bugs. If these flags are adequately controlled by tML, modders don't have to think about why or when to change them. They also don't have to think about what may happen if they get changed at a weird time.

absoluteAquarian: i can see why `PerformingJump` should be limited, but imo modders should still be able to modify `JumpAvailable`

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
Chicken-Bones: Yeah, enabled overrides are good, we use this design with `SetCrit()` and `DisableCrit()` in hit modifiers for example, providing a clear override precedence.

You've explained to me that setting `Enabled = false` doesn't clear `JumpAvailable`, so that when the thing disabling the jump expires, the jumps can continue where they left off. This is good, but should be clear to a modder (from documentation on `JumpState`)

I would also recommend adding a variable to `Player` which can disable all double jumps (and maybe another one for flight), as a convenience so a modder doesn't have to loop over all jumps every frame to disable them.

We could use such a flag to implement the flipper swimming jump lockout for example.

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
Chicken-Bones: Can you do some research, and make a write-up? What's special about sandstorm, aside from perhaps janky code. What does it do differently to blizzard/tsunami? How would a mod replicate those effects if they wanted to.

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
Chicken-Bones: The `Active` check here means that if something is suppressing a double-jump when the player leaves the ground, but then expires, the jump will not be considered available :(

patches/tModLoader/Terraria/Player.TML.cs
absoluteAquarian: in the vanilla logic, it's specifically checking if all extra jumps have been consumed (`JumpAvailable` would be `false` for all jumps).  furthermore, an extra jump is "consumed" by default

so the `Enabled` check would likely just end up with the same result?  after all, a disabled jump is always "consumed" and enabled jumps have to be used to be consumed... so checking `Enabled` would be pointless in a pure vanilla context

however, if modders can set `JumpAvailable` to true, then this would throw a wrench in the plan when the extra jump's `Enabled` flag is cleared

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: sure, i'll add another comment here — or a gist — with my findings (and a mention in the discord thread) once i get around to that

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: alas, this is a mimicry of vanilla behavior

granted, vanilla never had anything except for a hardcoded flipper check for disabling jumps.  this could probably be remedied by the `Disable()` method stuff mentioned in another conversation

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
absoluteAquarian: decided to add a `ExtraJumpState::Disable()` method which works as a forced disabling of the extra jump (prevents `PerformJump` from executing, does not affect `Enabled`) as well as `Player::blockExtraJumps` for easily disabling all extra jumps (it's checked alongside the flipper check)

patches/tModLoader/Terraria/ModLoader/VanillaExtraJumps.cs
absoluteAquarian: removed due to being obscure and not having any meaningful impact on gameplay other than blocking the Cloud jump's dust particles for no reason

patches/tModLoader/Terraria/DataStructures/ExtraJumpData.cs
Chicken-Bones: I recall one of the uses was marking a jump as still having uses left. Just need to note in the docs _why_ you would consider setting it to true or false, and what happens if you set it at a specific time (with respect to the other flags and logic)

absoluteAquarian: added documentation

patches/tModLoader/Terraria/ModLoader/ExtraJumpLoader.cs
absoluteAquarian: the gist in question: https://gist.github.com/absoluteAquarian/8a76bba28848ea5918bda5a26acf2f9a

absoluteAquarian: resolved due to making `Player.sandStorm` rely on `Player.isPerformingJump_Sandstorm`

patches/tModLoader/Terraria/Player.TML.cs
absoluteAquarian: resolved by removing the `NoActiveCheck` method in favor of making vanilla check `Enabled` as well via the `public` method

Chicken-Bones (approved):

Issue #3551, opened 2023-06-15
AcceptClientChanges should return a NetworkText instead of string.


JavidPack:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### What does this proposal attempt to solve or improve?

The sever doesn't know the clients selected language, sending a NetworkText allows the receiving end to calculate the final localized text.

Issue #2762, opened 2022-07-26
Filter mods list for mods with modconfig configs


Xetaxheb:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

Essentially another button or hidden inside mod-sidedness filter button, a filter for mods that have the gear that leads to config.

### What does this proposal attempt to solve or improve?

Making it easier to go through configs quickly with lots of mods installed, and reducing the incidence of scrolling past one by mistake.

### Which (other) solutions should be considered?

Putting as an option in all/enabled/disabled filter could make more sense because if a mod has the gear than by necessity it is **_loaded_**, though that doesn't always mean it is **_enabled_**.  This 'edge case' (disabling mods as you go through configuring) will probably in practice crop up way more than someone wanting to filter simultaneously for gear and for mod-sidedness, even though at face value it may seem more valuable to be able to filter sidedness and configurability... so I don't know which solution is better. 
Adding a 4th button adds more clutter to the UI, obviously. I don't think putting it under sorting is a good choice, because it can be desirable to sort alphabetically or by recent changes.

JavidPack:
This has been addressed by now. In current tmodloader, all mods with configs are easily accessible within the Mod Configuration menu

PR #3556, opened 2023-06-17
Config UI Rework [Old]

Destructor-Ben:
### What is the new feature?
- Completely reworked config list UI
  - Can now be accessed in the main menu as well as in game
  - Has a list for mods and a list for configs
  - Selecting a mod will show all it's available configs
  - Clicking a config will open it
  - Each config has a red or blue dot indicating whether it is client or server side
    - Red is server, blue is client
    - Hovering over the dot will also say if it is a server side or client side config
- Reworked mod config UI
  - Removed the next and previous buttons, since they aren't needed with the new config list
  - Moved the header panel
  - Moved the search bar
  - Changed the main panel sizing
  - Changed the notification panel to a popup/modal
  - Fixed spacing issues in the bottom buttons caused by the text size changing the button sizes (shortened the button names but gave them tooltips to compensate)
  - Changed revert button sound to menu close instead of menu open
  - Made scrolling anywhere scroll the main config list
  - Made all expandable elements default to not expanded
  - Fixed config UI buttons not refreshing after saving in multiplayer
  - Made boolean config elements play a click sound when changed
  - Changed header elements height to remove unused space and made their baseline 2 pixels instead of 1
  - Fixed a lot of cases with floats having annoying rounding errors like `4.199999` in the element labels (casted it to a decimal before it is displayed)
- Mod Config
  - Added `ModConfig.Open` to allow mods to open configs from their own UI
  - ~~Implemented recursive checking for whether a reload is required (previously only top level fields would trigger it)~~
    - ~~Also has the option to only recursively check certain types of fields~~
  - Won't be implementing recursive reload because it is buggy and complicated, see Important Note
- Made the config button on UIModItem in the mods menu to open the config list with the mod selected
- Added a button to open the mod config list from the mods menu
- Localized a lot of the config UI properly (some wasn't localized at all and some wouldn't allow language changes in game)

<details>
<summary><h3>Before</h3></summary>
<br>

Screenshot_20230617_011615

Screenshot_20230617_011546

</details>

<details>
<summary><h3>After</h3></summary>
<br>

<details>
<summary><h4>Old</h4></summary>
<br>

Screenshot_20230617_120038
</details>

Screenshot_20230726_083352

Screenshot_20230726_083406

Screenshot_20230617_120026

Screenshot_20230617_120126

Screenshot_20230617_120153

</details>

Currently the UI for the configs is ugly and confusing and it has been stated many times on the tML discord server that it should be changed. It is also difficult for users to find configs. In the main menu, the only ways to access configs was through the config button (but that only took you to the first one a mod registered), and the rest through the difficult to notice next and previous buttons. In game, there was the config list, which is how I usually changed configs, but you couldn't change the values that would require a reload, so you had to deal with the main menus config UI. In many cases I have never even realized that a mod has multiple configs because of this.

- ~For the notification popup, `NetworkText` should be used instead of localization keys, however this would be a breaking change, since it would change the parameters of the `AcceptClientChanges` method. Instead, I have allowed a localization key to be passed in, since `Language.GetTextValue()` will return the value passed in if a key isn't found, so mods that pass in the value won't be broken. The `NetworkText` change will probably be handled in https://github.com/tModLoader/tModLoader/issues/3551, but I can do it here if needed (I just wanted to avoid breaking changes for UI)~ Handled in https://github.com/tModLoader/tModLoader/pull/3750 and already merged into this
- A filter for the mods menu that only shows mods that have a config could be implemented like https://github.com/tModLoader/tModLoader/issues/2762 suggested, but it is mostly irrelevant now with the new UI
- The "Pending Changes" notification lasts for a very small amount of time, which causes an annoying flicker when the server receives the responses, so it could be removed, but I don't know how long it lasts for on a slow internet connection. It could always be changed to an animated image in the corner of the screen like the workshop publishing progress.

ModConfig.Open (works in game and in menus)
```csharp
ModContent.GetInstance<YourConfigName>().Open();
```

### ExampleMod updates
None, but could demonstrate the sample usages

### Important Note
There will most likely be a follow up PR that handles breaking UI changes and probably a rework to `ConfigElement` and `ModConfig`, with features such as moving attributes to a `ConfigField<T>` wrapper and recursive reload required.


Cyrillya:
Sounds good though I don't see the actual effect. Can you please add some screenshots of the new config UI to this pr?

direwolf420:
> Sounds good though I don't see the actual effect. Can you please add some screenshots of the new config UI to this pr?

The images are in the `>Before` and `>After` spoilers. I didn't notice them the first time I read through the PR though so maybe adding a header there should highlight them better

direwolf420:
I welcome all of these changes, especially the option to open the config through code

Destructor-Ben:
Only just occurred to me that this should be a draft since I still have a lot to do.

Destructor-Ben:
Closing, will be succeeded.

PR #3563, opened 2023-06-20
Tags for each language

JavidPack:
Adds tags for each language indicating if a mod is translated to a specific language. (We can also get rid of "Localization Support", it actually wasn't ever used on the workshop page)

image

Tag will automatically be set based on counts of non-trivial localization entries. Current math checks if there are at least 75% of entries translated. Modders can still click to set or unset a tag, but the system is automatic and will override previous choices each time the mod is published. This is to account for the localization status of a mod changing due to new content or new localizations.

Currently, since English is the default language, technically all mods will set the English tag. This might change later if we ever decide to support non-English default language for mods.

Also added scripts to facilitate syncing steam tag localization with tmodloader source tag localization entries. Previously we thought the system didn't work, but it does work after several minutes. 

chrome_2023-06-20_16-24-56

After this is merged and live, I'll update steam directly with the new Language tag category. It'll be a dropdown like mod side or tmodloader version. In game support will happen after the mod browser updates are merged.


JavidPack:
dotnet_2023-07-05_18-40-53

JavidPack:
Steam side of things are live.

image

PR #3564, opened 2023-06-21
Fix Character Creation Window Hair Draw Ordering Bug

### LINKED ISSUE #3550: Char creation hair window has incorrect draw order
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

image
as can be seen above, every texture in the hairstyle tab of the character creation menu is drawn after the hairstyle options themselves are, causin' them to obscure the actual hairstyles in a number of fashions (partial pseudo-transparency for hairstyles not moused over, and complete obscuration of a hairstyle that's bein' hovered over)

### Log File

_No response_

### Steps to reproduce

1. start to make a character
b. go into the hairstyle tab

### Expected Behavior

the hairstyles draw last, not causin' the above issue

### Actual Behavior

the hairstyles draw first out of the things that are drawn here

### Reproduction frequency

always, with no mods

### Additional Information

_No response_

Antirhinnum:
### What is the bug?
Fixes #3550.

### How did you fix the bug?
Enabled `UIElement.UseImmediateMode` in `UIHairStyleButton`. This follows usage from other UI elements that draw entities (`UICharacter`, `UIBestiaryEntryIcon`), as well as discussion in #collaborators about how players always draw using `SpriteSortMode.Immediate` (https://discord.com/channels/103110554649894912/445276626352209920/1092236989383839805).

This bug only happens on the `tModLoader` solution: The `Terraria` and `TerrariaNetCore` solutions do not have this bug, nor does vanilla *Terraria*. Thus, this bug was most likely caused by a tModLoader change.

It might be better to track down and fix the underlying issue, but I'm nowhere near well-versed enough in tModLoader's player drawing changes (or XNA/FNA graphics) to feel comfortable doing so.
The bug first appears after this commit: d85ea870730eb3178469e139a683236d18dec96b

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/Elements/UIHairStyleButton.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/Elements/UIHairStyleButton.cs.patch
@@ -1,0 +1,11 @@
+--- src/TerrariaNetCore/Terraria/GameContent/UI/Elements/UIHairStyleButton.cs
++++ src/tModLoader/Terraria/GameContent/UI/Elements/UIHairStyleButton.cs
+@@ -25,6 +_,8 @@
+ 		Height = StyleDimension.FromPixels(44f);
+ 		_selectedBorderTexture = Main.Assets.Request<Texture2D>("Images/UI/CharCreation/CategoryPanelHighlight");
+ 		_hoveredBorderTexture = Main.Assets.Request<Texture2D>("Images/UI/CharCreation/CategoryPanelBorder");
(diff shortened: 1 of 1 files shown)


JavidPack:
Thanks for looking into this. I worry the issue causing this might be affecting other things, so I'll try to figure out why the issue has appeared.

Chicken-Bones (approved):
Correct solution. Alternate solution would be to `Begin/End` the current sprite-batch prior to drawing the head, but that's what `UseImmediateMode` does here in practical terms anyway.

PR #3565, opened 2023-06-21
Fix `ModPlayer.PreKill` Not Stopping Kill Sounds

### LINKED ISSUE #3554: Unable to disable vanilla DEATH sound
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

On the new version setting playsound ref bool to false in override of a PreKill method in modplayer extending class doesn't seem to do anything. I haven't changed anything since 1.4.3 in the code there, but on 1.4.3 death sound can be disabled to be replaced and on 1.4.4 it doesn't seem to work, both sounds are played, is it the new issue or are there some implicit changes in the method behaviour, which I need to know?

### Log File

_No response_

### Steps to reproduce

set playsound ref bool to false in override of a PreKill method in modplayer extending class

### Expected Behavior

Death sound is disabled like on 1.4.3

### Actual Behavior

death sound still is played.

### Reproduction frequency

always

### Additional Information

two mods are extending modplayer (library mod and my mod using it), trying to set playsound to false in both doesn't seem to have an effect, originally library modplayer doesn't override prekill.

Antirhinnum:
### What is the bug?
Fixes #3554.

### How did you fix the bug?
Re-added the missing patch.

No.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -4878,6 +4878,21 @@
  		if (pvp)
  			pvpDeath = true;
  
+@@ -30004,10 +_,14 @@
+ 			}
+ 		}
+ 
++		if (!playSound)
++			goto postSound;
++
+ 		if (Main.dontStarveWorld || Main.tenthAnniversaryWorld)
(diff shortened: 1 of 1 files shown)


JavidPack:
Works, thanks for this.

PR #3568, opened 2023-06-21
Rubblemaker support: safe unloading and example usage

GeorgeFeldy:
### What is the new feature?
~~A very basic rubblemaker GlobalItem hook support. The user has the GlobalItem hook:
`AddToRubblemaker(FlexibleTileWand rubblemakerSmall, FlexibleTileWand rubblemakerMedium, FlexibleTileWand rubblemakerLarge)` in which they can call `AddVariation`, `AddVariations`, `AddVariations_ByRow` just like vanilla does.~~
Edit: Adds safe reloading of the Rubblemaker tables to the vanilla state, on mod unload, to avoid issues if the modder extends the tables with entries.

Rubblemaker is a feature that helps builders tremendously, it's a good idea to have modded support for that.
~~Directly modifying the rubblemaker tables or detouring could cause issues when the mod is unloaded, so best to provide a hook for this purpose.~~

### Are there alternatives?
Modders could simply take the rubblemaker instance and add items to it on mod load.
`FlexibleTileWand.RubblePlacementSmall.AddVariations(...)`
This would mean that they have to be very careful to reset the rubblemaker tables to the vanilla state on mod unload:
```
RubblePlacementSmall = CreateRubblePlacerSmall();
RubblePlacementMedium = CreateRubblePlacerMedium();
RubblePlacementLarge = CreateRubblePlacerLarge();
```
The reloading added in this PR will remove the need for that.

### Possible issues
~~Hook calling is currently done on ItemLoader.FinishSetup, and~~ reloading to the vanilla tables is done in ItemLoader.Unload. Might not be the best way to do that.

See:
- ExampleMod\Content\Tiles\ExampleChunkSmall.cs
- ExampleMod\Content\Tiles\ExampleChunkMedium.cs
- ExampleMod\Content\Tiles\ExampleChunkLarge.cs

### ExampleMod updates
Added ExampleChunkSmall, ExampleChunkMedium, ExampleChunkLarge ModTiles, each with a base type which configures the tile, a child type which is added to the Rubblemaker table, and another child type which is intented for use as a "natural" tile (i.e. placed on WorldGen), with example item and critter drops exclusive to it.
Possible extra examples:
- ~~Different item to place same tileID as example, but different styles~~
- ~~Another ModItem to place a vanilla style~~
- ~~Vanilla item to place ModTile~~
(Edit: the examples provided might be enough)
- **AddVariations_ByRow** example


JavidPack:
Does this need a hook? It seems like since the rubblemakers objects are all public static, the modder could just register in ModTile.SetStaticDefaults. We will still need to have the unload code of course, to reset them on mod unload, but I don't think the hook is needed. Can you try that and see if it works fine?

Also, is there something fundamentally different about rubble tiles compared to their regular tile counterpart? Do they not drop insects, do they drop different items, should modders create separate rubble tiles for their content to match Terraria behavior, etc? It would be good to add a comment to the ModTile explaining what makes a rubble tile special so modders know what they need to do. 

It might be possible to show an example of a normal and the rubble variant sharing the same texture and code, to minimize core repetition, but you can leave that to me if you want. 

You also have some accidental newlines that would be good to remove.

GeorgeFeldy:
Thanks for the feedback! No, it actually does not need a specific hook. I will remove it and update EM accordingly.

Rubble-made tiles are entirely different tileIDs, without any item drops or NPCs spawned on kill. 

I shall add to EM "natural" tile counterparts, that do have item/NPC drops, without code repetition.

Chicken-Bones:
Great work on this

ExampleMod/Content/Tiles/ExampleChunkLarge.cs
```diff
@@ -0,0 +1,63 @@
...
+using Terraria.DataStructures;
+using Terraria.GameContent;
+using Terraria.ID;
+using Terraria.ModLoader;
+using Terraria.ObjectData;
+
+namespace ExampleMod.Content.Tiles
+{
+	public abstract class ExampleChunkLargeBase : ModTile {
+		public override void SetStaticDefaults() {
```
direwolf420: as per [Style guide](https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide), opening braces after classes should be on a new line. Same goes for all the other files in this PR, I won't highlight them separately.
Also, since it's ExampleMod, might be necessary to add a comment on the abstract classes that says something like "implementing classes further down the file"

ExampleMod/Content/Tiles/ExampleChunkLarge.cs
```diff
@@ -0,0 +1,63 @@
...
+	// Natural ambient tile (i.e. created on WorldGen, not by Rubblemaker) 
+	public class ExampleChunkLargeNatural : ExampleChunkLargeBase{
+
+		//Override texture path, shared with rubble variant
+		public override string Texture => "ExampleMod/Content/Tiles/ExampleChunkLarge";
+
+		// Drops 3 ExampleItems on tile break
+		public override void KillMultiTile(int i, int j, int frameX, int frameY) {
+			Item.NewItem(new EntitySource_TileBreak(i, j), i * 16, j * 16, 32, 32, ModContent.ItemType<ExampleItem>(), Stack: 3);
+		}
```
direwolf420: This and other Kill(Multi)Tile usages:
* if it's only 1 item dropped, in `SetStaticDefaults`:`RegisterItemDrop(ModContent.ItemType<ExampleItem>());`
* If it's multiple (3 for example), use the `GetItemDrops` hook and return `yield return new Item(ModContent.ItemType<ExampleItem>(), 3);`

ExampleMod/Content/Tiles/ExampleChunkLarge.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		//Override texture path, shared with natural variant
+		public override string Texture => "ExampleMod/Content/Tiles/ExampleChunkLarge";
+
+		public override void SetStaticDefaults() {
+			// Call to base SetStaticDefaults. Must inherit static defaults from base type 
+			base.SetStaticDefaults();
+
+			// Add rubble variant, all existing styles, to Rubblemaker, allowing to place this tile by consuming ExampleItems
+			FlexibleTileWand.RubblePlacementLarge.AddVariations(ModContent.ItemType<ExampleItem>(), Type, 0, 1, 2, 3, 4, 5);
```
direwolf420: For modder convenience, you should add a new `AddVariations` overload that has no style parameters, which will then ignore tile styles (same item for every style basically). This might require changing the logic of the rubblemaker code slightly so that it will ignore tile style check if this override was used

direwolf420:
Some more comments on what exactly "Small"/"Medium"/"Large" mean/what tile dimensions they represent would be very important to have. I also know there are modded rubbles that don't match the size of any example rubbles in this PR, is it possible to add support for those? I.e. a 1x2 stalagmite/tite

JavidPack:
Just so you know I'm already in the process of doing the things mentioned by others, no need to do those since I already have local changes

JavidPack:
Thanks again for this.

direwolf420:
I don't see any explanation of what exactly "Small"/"Medium"/"Large" mean/what tile dimensions they represent. I think this is very important to note, either in the docs of the respective vanilla objects, and/or in examplemod.

Issue #3504, opened 2023-05-28
ProcessHitAgainstNPC does not respect DamageClass.UseStandardCritCalcs


jjohnsnaill:
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

Damage classes have a UseStandardCritCalcs property to disable random critical hits on weapons of that class. This works for projectiles, but not for hitting NPCs with the item directly.

### Steps to reproduce

Swing a weapon with a damage type that sets UseStandardCritCalcs to false at an NPC.

### Expected Behavior

The weapon never critically hits.

### Actual Behavior

UseStandardCritCalcs isn't checked, and the weapon is capable of critically hitting.

Chicken-Bones:
Seems like an easy fix

PR #3576, opened 2023-06-23
Check UseStandardCritCalcs in ProcessHitAgainstNPC

Cerbrus:
### What is the bug?

ProcessHitAgainstNPC does not respect DamageClass.UseStandardCritCalcs, 

reported here: #3504

### How did you fix the bug?
Add a condition around the crit damage calculation, so crit is only applied if the weapon can crit

Not as far as I'm aware.


Chicken-Bones:
Please use a `goto` to avoid indenting the block and reduce patch size

Cerbrus:
> Please use a `goto` to avoid indenting the block and reduce patch size

Done! :D

Chicken-Bones:
Thanks!

Issue #2923, opened 2022-09-13
Support for taller head armor sprites like the wizard's hat


Photonic0:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Vanilla features are often very hard to re-implement and having an item property or hook to allow for drawing of taller hats like badger's hat, wizard's hat and bunny perch would be more convenient.

### What does this proposal attempt to solve or improve?

It is currently impossible to use a "default" spritesheet format for the aformentioned hats, as they will get "cropped"(see image)
The code, item sprite and spritesheet are also included.
![image](https://user-images.githubusercontent.com/62021912/189794902-5dc8bf8b-0b80-4527-bcaf-3ac4d4cc4cbb.png)
![image](https://user-images.githubusercontent.com/62021912/189794938-7edd536f-8e7c-49e4-b0ed-b236302724ed.png)

![DevHat](https://user-images.githubusercontent.com/62021912/189795215-3835f905-820c-4de1-adf9-69ca1cd58ce5.png)
![DevHat_Head](https://user-images.githubusercontent.com/62021912/189795317-ae908288-8438-4c1e-9673-b9cdd515a0ce.png)
https://hst.sh/amilokemux.cs

JavidPack:
More info: In `DrawPlayer_21_Head`, vanilla hats like this are checked
```
bool flag = drawinfo.drawPlayer.head == 14 || drawinfo.drawPlayer.head == 56 || drawinfo.drawPlayer.head == 114 || drawinfo.drawPlayer.head == 158 || drawinfo.drawPlayer.head == 69 || drawinfo.drawPlayer.head == 180;
```

This is later used to adjust the coordinates of the `DrawData` added to the player render.

We can fix this with a ArmorIDs.Heads.Sets. Or we can use this as an example of finding and editing DrawData manually, but this is something many would want so a set sounds right.

cozykoko:
done and closable?

PR #3577, opened 2023-06-23
Add mod support for taller head armor sprites

Cerbrus:
### What is the new feature?
Support for taller head armor sprites like the wizard's hat
Reported at: #2923

This makes it possible to add head armor in mods with taller head armor sprites, like wizard's hats.

This solution was suggested in the reported issue.

I renamed the `flag` boolean to the more descriptive `isTallHat`.

If you're adding a new head armor item, you can enable taller sprite rendering with the following line:

```
ArmorIDs.Head.Sets.IsTallHat[Item.headSlot] = true;
```

### ExampleMod updates

I've added a `ExampleTallHelmet`  (ingame "Example Tall Helmet") to show it functioning.


JavidPack:
Thanks for this. I wasn't sure if the textures from the issue were free to use, so I swapped them out. I also fixed a bug when drawinfo.drawPlayer.head is -1, meaning no helmet is equipped.

JavidPack:
Some Notes: 
Also, I decided against adding support for extra wide helmets, because it wouldn't be easy to support even taller helmets without enforcing some specific arbitrary layout choices. I'll leave that to modders.

Cerbrus:
> Thanks for this. I wasn't sure if the textures from the issue were free to use, so I swapped them out. I also fixed a bug when drawinfo.drawPlayer.head is -1, meaning no helmet is equipped.  

You're welcome!

I'm new to contributing here, so your changes are welcome 😃 

I was curious, I see you changed the flag name back to `flag`, why is that?

JavidPack:
> I was curious, I see you changed the flag name back to `flag`, why is that?

I changed it back because we try to minimize the patches. 

When Terraria updates, there is a process of migrating patches to the new decompiled source code, that process requires us to review any patch in methods that changed since last version. It's easier to maintain if it's just a single change, if it's 2 changes then there is a possibility that we make a mistake and what used to be `flag` is now `flag3` and we miss updating one or the other patch, resulting in bugs. 

We don't always minimize to this extent, such as with confusing variables, but the intent of `flag` is easy enough to follow that it's easier to leave it as flag.

Cerbrus:
> > I was curious, I see you changed the flag name back to `flag`, why is that?
> 
> I changed it back because we try to minimize the patches. <....>


Ah, that makes a lot of sense! Thanks for the explanation!

PR #3578, opened 2023-06-23
Hovering Wings

### LINKED ISSUE #3441: Hovering Wing Deficiencies

Yes

Mod capability as a Modder

### Description

Some way to mark that a modded pair of wings is able to hover, similar to vanilla's Hoverboard. Additionally, some way to specify that a pair of wings should consume less flight time when hovering and not moving. Both of these are hardcoded for specific vanilla wings.

In short, modders would (ideally) be able to specify the following for modded wings:
* Whether or not the wings should allow hovering.
* The slowdown factor of entering the hovering state (see "Other solutions").
* The flight time usage per frame while hovering in place.

These would be implemented using sets in `ArmorIDs.Wing.Sets`:
* `ArmorIDs.Wing.Sets.CanHover` (default: false)
* `HoverSlowdown` (default: 0.9f)
* `HoverFlightTimeCost` (default: 0.5f)

Currently, there is no reasonable way to make a modded pair of wings hover, despite the functionality being seen in many vanilla wings (Hoverboard, Vortex Booster, Celestial Starboard, etc.).

`ArmorIDs.Wing.Sets.Stats` and the associated `WingStats` struct allow specifying how fast a player should move when trying to hover. This behavior works fine, but does not actually make the player hover: They just speed up if holding down (or equivalent).

The code that handles hovering is right in the middle of `Player.Update()`, so it's unreasonable to IL edit in hovering logic, especially for something so common in vanilla wings. Additionally, there are no nearby tModLoader hooks where a modder could handle hovering logic manually.

It *is* currently reasonable to make a pair of wings consume less flight time when "hovering" using `Mod/GlobalItem.VerticalWingSpeeds()` by directly modifying `Player.wingTime` to offset the default `-1f` per frame. However, this approach is unintuitive and must be done defensively to avoid the modder increasing the player's wing time infinitely. It also doesn't work for the Jetpack's "boost" ability, but that's its own can of worms.

The blocks of code in vanilla that handle hovering and flight time use are as follows:
```cs
// void Player.Update()
if ((wingsLogic == 22 || wingsLogic == 28 || wingsLogic == 30 || wingsLogic == 31 || wingsLogic == 33 || wingsLogic == 35 || wingsLogic == 37 || wingsLogic == 45) && TryingToHoverDown && controlJump && wingTime > 0f && !merman) {
	float num71 = 0.9f;
	if (wingsLogic == 45)
		num71 = 0.8f;

	velocity.Y *= num71;
	if (velocity.Y > -2f && velocity.Y < 1f)
		velocity.Y = 1E-05f;
}

if (wingsLogic == 37 && TryingToHoverDown && controlJump && wingTime > 0f && !merman) {
	velocity.Y *= 0.92f;
	if (velocity.Y > -2f && velocity.Y < 1f)
		velocity.Y = 1E-05f;
}
```
```cs
// Player.WingMovement()
if ((wingsLogic == 22 || wingsLogic == 28 || wingsLogic == 30 || wingsLogic == 31 || wingsLogic == 37 || wingsLogic == 45) && TryingToHoverDown && !controlLeft && !controlRight)
	wingTime -= 0.5f;
else
	wingTime -= 1f;
```
The "slowdown factor" mentioned in the main proposal are the `0.9f`, `0.8f`, and `0.92f` values in `Player.Update()` that determine how quickly the player slows down when entering a hovering state.

As is always the case with new sets, these will require maintenance if any future *Terraria* updates add more wings, which not be desired. However, only keeping track of these stats for modded wings would make it incredibly difficult for modders to programmatically check these fields on vanilla wings, an issue that already exists with vanilla grappling hooks.

Raw sets for the proposed fields could work, but may be cumbersome compared to `WingStats`. Turning these into fields or properties on `Item` or `ModItem` seems excessive, as these fields have no relevance to non-wing items.

The `WingStats` struct is a candidate for implementation, but not preferred. The struct already has a `HasHoverDownStats` property, which decides if a pair of wings changes speed while hovering. Amending this would require either reassigning the value to all vanilla wings (not easily maintainable) or making a second property (might be confusing for modders due to similar name). The `HasHoverDownStats` bool could also be renamed to something less confusing, likely through hiding and a property (like how tModLoader handles `Item.melee`).

Antirhinnum:
### What is the new feature?
Resolves #3441.
Allows modders to mark that a certain `wingSlot` should be allowed to hover in place, similarly to vanilla's Hoverboard.
Also allows configuration of this hovering.
All added sets apply to both vanilla and modded wings.

See #3441.

See #3441.
This approach adds new sets because no vanilla wing changes its hovering stats at runtime. A hook similar to `VerticalWingSpeeds` could be used to allow changing stats at run time, but it seems excessive for this. Plus, changing `Item.wingSlot` or `Player.wingsLogic` would work just as well.

```cs
// In ModItem.SetStaticDefaults, with [AutoloadEquips(EquipType.Wing)]
// All values set to default value.
ArmorIDs.Wings.Sets.CanHover[Item.wingSlot] = true; // Allows hovering. Default: false
ArmorIDs.Wings.Sets.InUseWhenAttemptingHover[Item.wingSlot] = null; // Used by Solar, Stardust, and Nebula wings, included for compatibility. Default: null (vanilla decision)
ArmorIDs.Wings.Sets.HoverSlowdown[Item.wingSlot] = 0.8f; // How fast the player downs down to a stable hover. Default: 0.9f
ArmorIDs.Wings.Sets.StationaryHoverUsage[Item.wingSlot] = 0.25f; // How much wing time to use per frame when trying to hover. Default: 1f - Hoverboard: 0.5f
```

### DIFF
--- a/patches/tModLoader/Terraria/DataStructures/WingStats.cs.patch
+++ b/patches/tModLoader/Terraria/DataStructures/WingStats.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/DataStructures/WingStats.cs
 +++ src/tModLoader/Terraria/DataStructures/WingStats.cs
-@@ -1,15 +_,47 @@
+@@ -1,15 +_,48 @@
  namespace Terraria.DataStructures;
  
 +/// <summary>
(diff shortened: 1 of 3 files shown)


JackerAcid:
Any update on this? 
This looks like a very useful feature to add instead of having to manually add the logic in yourself to the wing.
Could also present interesting cross-mod possibilities.

PR #3583, opened 2023-06-26
Initial Code for Steam Cloud Save Migrations

Solxanich:
In order to support the transition to 1.4.3-legacy branch and having both 1.4.4 and 1.3 and 1.4.3 available, we have to implement some changes in save data structures.

This takes the changes made previously to have a 1.4.3-legacy folder created, and adds the equivalent code for Steam Cloud.

Outstanding:

- [x] Cleanup Port143FilesFromStable to not crash due to file inaccessibility by changing it's design.
- [x] DEtermine if better to just write all new code specifically for steam cloud saves instead of trying to reuse code haphazardly

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -89,12 +89,13 @@ namespace Terraria
 		private static void Port143FilesFromStable(string savePath, bool isCloud) {
 			// Copy all current stable player files to 1.4.3-legacy during transition period. Skip ModSources & Workshop shared folders
 
-			string newFolderPath = Path.Combine(savePath, ReleaseFolder);
+			string newFolderPath = Path.Combine(savePath, Legacy143Folder);
 			string oldFolderPath = Path.Combine(savePath, StableFolder);
 
 			if (!Directory.Exists(newFolderPath) && Directory.Exists(oldFolderPath)) {
 				Utilities.FileUtilities.CopyFolder(oldFolderPath, newFolderPath, isCloud,
-					excludeFilter: new System.Text.RegularExpressions.Regex("(Workshop)|(ModSources)"));
+					// Exclude the ModSources folder that exists only on Stable, and exclude the temporary 'Workshop' folder created during first time Mod Publishing
+					excludeFilter: new System.Text.RegularExpressions.Regex("((/|\\)Workshop(/|\\))|((/|\\)ModSources(/|\\))"));
 			}
 		}
 
@@ -104,7 +105,7 @@ namespace Terraria
 			Port143FilesFromStable(savePath, isCloud);
 		}
 
-		public const string ReleaseFolder = "tModLoader-1.4.3";
+		public const string Legacy143Folder = "tModLoader-1.4.3";
 		public const string StableFolder = "tModLoader";
 		public const string DevFolder = "tModLoader-dev";
 		public const string PreviewFolder = "tModLoader-preview";
@@ -123,7 +124,7 @@ namespace Terraria
 			}
 
 			// 1.4.3 legacy custom statement - the legacy 143 folder
-			var fileFolder = ReleaseFolder;
+			var fileFolder = Legacy143Folder;
 
 			SavePath = Path.Combine(SavePath, fileFolder);
 
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -39,18 +39,18 @@ namespace Terraria
 			}
 		}
 
-		private static void PortOldSaveDirectories() {
+		private static void PortOldSaveDirectories(string savePath) {
 			// PortOldSaveDirectories should only run once no matter which branch is run first.
 
 			// Port old file format users
-			var oldBetas = Path.Combine(SavePath, "ModLoader", "Beta");
+			var oldBetas = Path.Combine(savePath, "ModLoader", "Beta");
 
 			if (!Directory.Exists(oldBetas))
 				return;
 
 			Logging.tML.Info($"Old tModLoader alpha folder \"{oldBetas}\" found, attempting folder migration");
 
-			var newPath = Path.Combine(SavePath, ReleaseFolder);
+			var newPath = Path.Combine(savePath, StableFolder);
 			if (Directory.Exists(newPath)){
 				Logging.tML.Warn($"Both \"{oldBetas}\" and \"{newPath}\" exist, assuming user launched old tModLoader alpha, aborting migration");
 				return;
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -87,24 +87,14 @@ public static partial class Program
...
-				// Especially if OneDrive is involved. This is not a good solution and will fail as is
-				string modSources = Path.Combine(newFolderPath, "ModSources");
-				if (Directory.Exists(modSources))
-					Directory.Delete(modSources, true);
-
-				string workshopTemp = Path.Combine(newFolderPath, "Workshop");
-				if (Directory.Exists(workshopTemp))
-					Directory.Delete(workshopTemp, true);
+				Utilities.FileUtilities.CopyFolder(oldFolderPath, newFolderPath, isCloud,
+					excludeFilter: new System.Text.RegularExpressions.Regex("(Workshop)|(ModSources)"));
```
Chicken-Bones: I'm concerned about an exclude filter which doesn't have locators. `Directory.GetDirectories` returns full paths, and `IsMatch` with this regex will find the specified string at any place in the path. Something like `WorkshopInfo.json` could end up getting excluded.

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -86,18 +84,30 @@ public static partial class Program
...
 			}
 		}
 
+		internal static void PortFilesMaster(string savePath, bool isCloud) {
+			PortOldSaveDirectories(savePath);
+			PortCommonFilesToStagingBranches(savePath);
+			Port143FilesFromStable(savePath, isCloud);
+		}
 
 		public const string ReleaseFolder = "tModLoader-1.4.3";
```
Chicken-Bones: Is `ReleaseFolder` a misnomer here? `LegacyFolder` better?

patches/tModLoader/Terraria/Program.TML.cs
Solxanich: Make it more strict - Looks for /Workshop/ or /ModSources/.
This will ensure it only excludes items deriving those intermediate folders

patches/tModLoader/Terraria/Program.TML.cs
Solxanich: Selected Legacy143Folder for best clarity

patches/tModLoader/Terraria/Utilities/FileUtilities.cs.patch
```diff
@@ -17,3 +17,72 @@
...
+ 				File.Copy(source, destination, overwrite);
+-			else if (SocialAPI.Cloud != null && (overwrite || !SocialAPI.Cloud.HasFile(destination)))
+-				SocialAPI.Cloud.Write(destination, SocialAPI.Cloud.Read(source));
++				return;
++			}
++
++			// Sanitize the paths for Steam calls
++			string cloudPath = Social.Steam.CoreSocialModule.GetCloudSaveLocation();
++			destination = ConvertToRelativePath(cloudPath, destination);
++			destination = ConvertToRelativePath(cloudPath, source);
```
Chicken-Bones: typo?

patches/tModLoader/Terraria/Utilities/FileUtilities.cs.patch
```diff
@@ -17,3 +17,72 @@
...
+@@ -154,16 +_,24 @@
+ 			return match.Groups["path"].Value;
+ 		}
+ 
+-		public static void CopyFolder(string sourcePath, string destinationPath) {
++		public static void CopyFolder(string sourcePath, string destinationPath, bool isCloud = false, Regex excludeFilter = null) {
+ 			Directory.CreateDirectory(destinationPath);
+ 			string[] directories = Directory.GetDirectories(sourcePath, "*", SearchOption.AllDirectories);
+ 			for (int i = 0; i < directories.Length; i++) {
++				string testStr = ConvertToRelativePath(sourcePath, directories[i]);
```
Chicken-Bones: `relPath`?

patches/tModLoader/Terraria/Utilities/FileUtilities.cs.patch
```diff
@@ -17,3 +17,72 @@
...
++				if (excludeFilter != null && excludeFilter.IsMatch(testStr))
++					continue;
++
+ 				Directory.CreateDirectory(directories[i].Replace(sourcePath, destinationPath));
+ 			}
+ 
+ 			directories = Directory.GetFiles(sourcePath, "*.*", SearchOption.AllDirectories);
+ 			foreach (string obj in directories) {
++				string testStr = ConvertToRelativePath(sourcePath, obj);
```
Chicken-Bones: As above

patches/tModLoader/Terraria/Utilities/FileUtilities.cs.patch
```diff
@@ -17,3 +17,72 @@
...
++		}
++
++		/// <summary>
++		/// Converts the full 'path' to remove the base path component.
++		/// Example: C://My Documents//Help Me I'm Hungry.txt is full 'path'
++		///		basePath is C://My Documents
++		///		Thus returns 'Help Me I'm Hungry.txt'
++		/// </summary>
++		public static string ConvertToRelativePath(string basePath, string fullPath) {
++			return fullPath.Replace(basePath + Path.DirectorySeparatorChar, "");
```
Chicken-Bones: Would rather not rely on `Path.DirectorySeparatorChar` here, the base path may also be repeated in the path string on some platforms or in cloud paths. Better to use `StartsWith` and `SubString`

patches/tModLoader/Terraria/Social/Steam/CoreSocialModule.TML.cs
```diff
@@ -0,0 +1,25 @@
...
+			Program.PortFilesMaster(GetCloudSaveLocation(), isCloud: true);
+		}
+
+		internal static string GetCloudSaveLocation() {
+			// 512 is Windows Path max length
+			Steamworks.SteamUser.GetUserDataFolder(out string steamCloudFolder, 512);
+
+			// current steamCloudFolder looks like: C:\\Program Files (x86)\\Steam\\userdata\\SteamID OLD\\1281930\\local
+			// So we have to do some path manipulation to get to C:\\Program Files (x86)\\Steam\\userdata\\SteamID OLD\\\\105600\\local
+			return Path.Combine(Directory.GetParent(Directory.GetParent(steamCloudFolder).ToString()).ToString(), "105600", "remote");
```
Chicken-Bones: `Path.GetDirectoryName` is the stringy version of `Directory.GetParent` which is a bit easier to compose here

patches/tModLoader/Terraria/Social/Steam/CoreSocialModule.TML.cs
```diff
@@ -0,0 +1,25 @@
+using System.IO;
+
+namespace Terraria.Social.Steam
+{
+	public partial class CoreSocialModule : ISocialModule
+	{
+		public const string WindowsSteamCloudPath = "C:\\Program Files (x86)\\Steam\\userdata";
+		public const string LinuxSteamCloudPath = "~/.local/share/Steam/userdata";
+		public const string MacSteamCloudPath = "~/Library/Application Support/Steam/userdata";
```
Chicken-Bones: Is this really necessary? I can easily imagine people with different drive letters or install locations. None of the other steam cloud code uses them.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Social/Steam/CoreSocialModule.TML.cs
Solxanich: This code doesn't even use it. 
I was thinking might be nice to have for supportability, but it's not like you can actually move the files around yourself - only copy them away from there to local manually.
Not really useful at this juncture, will remove.

PR #3588, opened 2023-06-27
PaintedFallingLeaf gore set, fix ExampleTreeLeaf

direwolf420:
### What is the bug?
`ExampleTreeLeaf` would glitch out when spawned
[Image](https://cdn.discordapp.com/attachments/445276626352209920/1123173112377249896/Screenshot_1.png)

### How did you fix the bug?
It was not utilizing the same logic as other tree leaves. Introduced `GoreID.Sets.PaintedFallingLeaf` which controls the `Gore.Frame`. Also, for consistency with vanilla, the gore now appears regardless of the "Blood and Gore" setting.

I added the caveat "it's coming from a tree _via wind effects_" to the docs because in vanilla it's not consistent; leaves spawned on growth or on shake are not "painted" and default to the regular variant. Probably a vanilla bug or oversight? @Chicken-Bones


JavidPack:
Thanks. Can confirm that the shake tree leaves are a vanilla bug, so we'll leave that for Terraria to fix.

Issue #3404, opened 2023-04-11
Conditions for Shimmer Transforming


PrimeEmperor:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

There must be some kind of feature (For example, boolean variables, like "CanUseItem(Player player)" for ModItem or "CanBeCaughtBy(Item item, Player player)" for ModNPC), which will decide a shimmer transform conditions. 
This is meant for NPCs and items without recipes (Items with those already have ".AddDecraftCondition").

### What does this proposal attempt to solve or improve?

This feature will make an easy way to create a conditions for shimmer transformations.

Vanilla already has some examples of this: Rod of Harmony and Terraformer - for Rod of Discord and Clentaminator respectively, both are normally uncraftable. They can be transformed after defeating Moon Lord.

Chicken-Bones:
For anyone looking to take a crack at this, the shimmer transform system should be as declarative as possible. By declarative I mean, registering transformations and conditions with static methods/sets. There may still be a need for general purpose hooks, but I think most key examples can be implemented without.

Kogsey:
I did the initial shimmer stuff so I may as well do this one, I think? Maybe I setup some stuff for item transformations with whatever types but that might've slipped out, was a while ago so memory is a little off, NPC stuff should be pretty easy.

PR #3594, opened 2023-06-30
Full Shimmer API

Kogsey:
### What is the new feature?
ref #3404 

More flexible Shimmer API disconnected from `Recipe`, abstract class `ShimmerTransformation` and class `ShimmerTransformation<T : IModShimmerable>` acts as controller for transformations, all ShimmerTransformation transformations take priority over normal transformations but fall through to vanilla if conditions all fail.

`ShimmerTransformation` is designed to operate similar to `Recipe` from the modder's perspective:
 - `CreateShimmerTransformation()` can be used to instantiate a new `ShimmerTransformation<T>` from `ModItem`, and `ModProjectile` like recipe
 - `AddCondition(Condition condition)` Same `Condition` as `Recipe`
 - `AddResult(ModShimmerResult result)` modders can derive from `ModShimmerResult` for custom spawn behaviour, all following wrap this
   - `AddItemResult(int type, int stack)`
     - `AddModItemResult<T>(int stack) where T : ModItem`
   - `AddNPCResult(int type, int count)`
     - `AddModNPCResult<T>(int count) where T : ModNPC`
   - `AddCoinLuckResult(int coinLuck)`
 - `DisableVanillaItemConstraints()` For shimmers with results Bone or Lihzard brick vanilla disables shimmering, using this re-enables it
 - `AddCanShimmerCallBack(CanShimmerCallBack)`
 - `AddModifyShimmerCallBack(ModifyShimmerCallBack)`
 - `AddOnShimmerCallBack(OnShimmerCallBack)`
 - `SortBefore(ShimmerTransformation transformation, int forType)` Same as recipe
 - `SortAfter(ShimmerTransformation transformation, int forType)`
 - `Register()`
   - `RegisterRegister(params int[] types)` Used when instantiated outside of ModNPC or ModItem, following wrap this
   - `Register(IEnumerable<int> types)`
   - `Register(int type)`

ShimmerTransformation also contains logic to "redirect" transformations. System is very simple due to no reference to the shimmering type being stored in a ShimmerTransformation instance, wraps `ItemID.Sets.ShimmerCountsAsItem` as well. All static methods must be called on ShimmerTransformation<T> and are restrcited to T by type:
- `AddRedirect(int typeFrom, int typeTo)`
- `GetRedirectedType(int type)`

Adds identical hooks to `GlobalNPC`, `GlobalItem`, `GlobalProjectile`, `ModNPC`, `ModProjectile`, and `ModItem`
 - `OnShimmer(NPC/Item/Projectile)` called when the entity shimmers in either vanilla or modded
 - `CanShimmer(NPC/Item/Projectile)` checks if the entity can shimmer, prevents both vanilla and modded if false

Adds `NPCID.Sets.IgnoreNPCSpawnedFromStatue[int]`
- Allows modders to turn off statue NPCs despawning in shimmer

Can shimmer an entity on modder command with `DoShimmer(IModShimmerable)` or `TryModShimmer(IModShimmerable)`

- I thought about adding in support for droprule tables but I can't think of much that can't be done with the modify shimmer callback, but an implementation with that or something similar would make read-access better if something like recipe browser or magic storage wanted to support shimmer or a mod wanted to expand the bestiary with shimmer info
- It may be useful to make ShimmerTransformation a read-only record, or at least internally use one, with ShimmerTransformation as a logic class to prevent a billion "with" mutations when creating. Currently multiple keys can index the same ShimmerTransformation index, and they are mutable, this may cause unexpected behaviour for unaware/inexperienced modders but this is really only a cross mod problem and most people doing that stuff understand reference types vs value types
- For the bugfix #3722 I've simply added a counterpart to the recipe consumtion rule, shimmer consumption, this is non-breaking but sort of unnoticeable, another non-breaking way would to add a bool condition for shimmer at the end of the delegate and cast the current delegate when adding it to the list but this doesn't solve the problem. I'm in favour of adding the bool in and letting it error but that just be because I don't use this in my mod

I think this system is robust for all the common use cases and makes the weird stuff doable:
- Calamity can late-game RoH with globals
- If people want to make transformations carry prefixes/other data it can be done with the same, or by using custom result types

Created to be quite close to recipe in function, register can be done in SetStaticDefaults() or AddRecipes().

Example for creating an item with similar behaviour to the rod of harmony:
```
CreateShimmerTransformation()
	.AddCondition(Condition.DownedMoonLord)
	.AddItemResult(ItemID.RodOfHarmony)
	.Register();
```

Example for adding a transformation to vanilla:
```
public class ExampleShimmerGlobalNPC : GlobalNPC
{
		int[] transformationFor = new int[] {
			NPCID.Goldfish,
			NPCID.Crab,
			NPCID.Penguin,
			NPCID.PartyBunny,
		};

		new ShimmerTransformation<NPC>() // Since we're not in ModNPC we instantiate manually
			.AddNPCResult(NPCID.Frog, 1) // Vanilla frog
			.AddCondition(Condition.InBeach) // On the beach
			.Register(transformationFor); // Registers for every entity passed within the array

		NPCID.Sets.ShimmerIgnoreNPCSpawnedFromStatue[NPCID.Goldfish] = true; // The goldfish statue spawns can undergo any shimmer operation
	}
}
```

Example for removing an item's ability to transform:
```
public class ExampleRemoveShimmer : GlobalItem
{
	public override bool AppliesToEntity(Item entity, bool lateInstantiation) {
		return entity.type == ItemID.RodofDiscord;
	}

	public override bool CanShimmer(Item item) {
		return false; // The rod of discord can no longer shimmer.
	}
}
```

Example for a generic transformation:
```
CreateShimmerTransformation()
	.AddCanShimmerCallBack((ModShimmer transformation, IModShimmerable target) => <bool>)
	.AddItemResult(ItemID.ExplosiveBunny, 30)
	.AddNPCResult(NPCID.Skeleton, 3)
	.AddOnShimmerCallBack((ModShimmer transformation, IModShimmerable target, IEnumerable<IModShimmerable> spawned) => <void>)
	.Register();
```

### ExampleMod updates
Changed `ShimmerShowcase.cs` to `ItemShimmerShowcase.cs` + examples for new shimmer system
Added `NPCShimmerShowcase.cs`
Added `ExampleComplexCustomShimmerable.cs`
Added `ExampleShimmerGlobalNPC.cs`
Added `ExampleRemoveShimmer.cs`


patches/tModLoader/Terraria/GameContent/ShimmerTransforms.cs.patch
Kogsey, line 1: I added this on my last shimmer PR, it's uneeded as when getting `decraftingRecipeIndex` this is already checked, just noting here because it's semi-unrelated to PR

Issue #3507, opened 2023-05-30
IsTileBiomeSightable hook


Lion8cake:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

A new hook for mod tiles called IsTileBiomeSightable that would allow for the tile to be picked up by the biome sight potion and its buff.

### What does this proposal attempt to solve or improve?

This would allow for any spreadable biome or any biome intended to be clensed similar to Calamity's Astral Infection to be picked up by the biome sight potion to be clensed easier.

Chicken-Bones:
Will a `TileID.Sets` entry work for you, or do you need a hook?

Lion8cake:
I do not mind personally, some people way wish to have checks in place to enable biome sight at a point in progression so it might be better to have it be a hook.

PR #3595, opened 2023-06-30
Add `IsTileBiomeSightable` hook to `ModTile`

Cerbrus:
### What is the new feature?
A new hook for mod tiles called `IsTileBiomeSightable` that allows for the tile to be picked up by the biome sight potion and its buff.

Reported here: #3507

The PR got larger than I expected, as these hooks seem to touch quite a few files.
I looked at `IsTileSpelunkable` and `IsTileDangerous` for examples.

This adds mod support so modded tiles can be highlighted from the biome sight potion.

Another proposed suggestion was to use a `TileID.Sets` property, but a hook seemed preferable (more flexible in use)

In a `ModTile`, override the `IsTileBiomeSightable` hook:

```csharp
public override bool IsTileBiomeSightable(int i, int j) => true;
```

### ExampleMod updates

I made the `ExampleOre` in `ExampleMod` biome sightable

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleOre.cs
+++ b/ExampleMod/Content/Tiles/ExampleOre.cs
@@ -31,7 +31,9 @@ namespace ExampleMod.Content.Tiles
 		}
 
 		// Example of how to enable the Biome Sight buff to highlight this tile
-		public override bool IsTileBiomeSightable(int i, int j) => true;
+		public override bool IsTileBiomeSightable(int i, int j) {
+			return true;
+		}
 	}
 
 	public class ExampleOreSystem : ModSystem
--- a/ExampleMod/Content/Tiles/ExampleOre.cs
+++ b/ExampleMod/Content/Tiles/ExampleOre.cs
@@ -30,7 +30,7 @@ namespace ExampleMod.Content.Tiles
 			// MinPick = 200;
 		}
 
-		// Example of how to enable the Biome Sight buff to highlight this tile
+		// Example of how to enable the Biome Sight buff to highlight this tile. Biome Sight is technically intended to show "infected" tiles, so this example is purely for demonstration purposes.
 		public override bool IsTileBiomeSightable(int i, int j, ref Color sightColor) {
 			sightColor = Color.Blue;
 			return true;
--- a/ExampleMod/Content/Tiles/ExampleOre.cs
+++ b/ExampleMod/Content/Tiles/ExampleOre.cs
@@ -30,6 +30,7 @@ namespace ExampleMod.Content.Tiles
 			// MinPick = 200;
 		}
 
+		// Example of how to enable the Biome Sight buff to highlight this tile
 		public override bool IsTileBiomeSightable(int i, int j) => true;
 	}
 
--- a/ExampleMod/Content/Tiles/ExampleOre.cs
+++ b/ExampleMod/Content/Tiles/ExampleOre.cs
@@ -29,6 +29,8 @@ namespace ExampleMod.Content.Tiles
 			// MineResist = 4f;
 			// MinPick = 200;
 		}
+
+		public override bool IsTileBiomeSightable(int i, int j) => true;
 	}
 
 	public class ExampleOreSystem : ModSystem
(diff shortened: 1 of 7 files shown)


ExampleMod/Content/Tiles/ExampleOre.cs
```diff
@@ -29,6 +29,9 @@ public class ExampleOre : ModTile
 			// MineResist = 4f;
 			// MinPick = 200;
 		}
+
+		// Example of how to enable the Biome Sight buff to highlight this tile
+		public override bool IsTileBiomeSightable(int i, int j) => true;
```
direwolf420: ExampleMod convention:
```suggestion
		public override bool IsTileBiomeSightable(int i, int j) {
			return true;
		}
```

patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -568,3 +586,12 @@
  	{
  		_windGrid.GetWindTime(i, j, pushAnimationTimeTotal, out var windTimeLeft, out var directionX, out var _);
  		float num = (float)windTimeLeft / (float)pushAnimationTimeTotal;
+@@ -6903,7 +_,7 @@
+ 			float num6 = (float)num2 * num3 * windCycle + num4;
+ 			if (_localPlayer.biomeSight) {
+ 				Color sightColor = Color.White;
+-				if (Main.IsTileBiomeSightable(type, tileFrameX, tileFrameY, ref sightColor)) {
++				if (Main.IsTileBiomeSightable(x, i, type, tileFrameX, tileFrameY, ref sightColor)) {
```
direwolf420, line 594: Just for other reviewers: `x, i` are indeed the correct coordinates

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3169,6 +3169,38 @@
...
++		IsTileBiomeSightable(tileX, tileY, t.type, t.frameX, t.frameY, ref sightColor);
++
++	internal static bool IsTileBiomeSightable(int tileX, int tileY, ushort typeCache, short tileFrameX, short tileFrameY, ref Microsoft.Xna.Framework.Color sightColor)
+ 	{
++		bool? modded = TileLoader.IsTileBiomeSightable(tileX, tileY, typeCache);
++		if (modded.HasValue)
++			return modded.Value;
++
+-		if (TileID.Sets.CorruptBiomeSight[type] || (remixWorld && type == 474)) {
++		if (TileID.Sets.CorruptBiomeSight[typeCache] || (remixWorld && typeCache == 474)) {
```
direwolf420: I suggest changing the parameter name back to `type`, this avoids patches like these

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -3169,6 +3169,38 @@
  		if (tileSpelunker[typeCache])
  			return true;
  
+@@ -17293,19 +_,27 @@
+ 		return false;
+ 	}
+ 
+-	public static bool IsTileBiomeSightable(ushort type, short tileFrameX, short tileFrameY, ref Microsoft.Xna.Framework.Color sightColor)
++	//TML: Added x/y to IsTileBiomeSightable methods for hook compatibility, made internal. Public variant is in Main.TML.cs.
```
direwolf420: space after a comment, as per conventions:
```suggestion
+	// TML: Added x/y to IsTileBiomeSightable methods for hook compatibility, made internal. Public variant is in Main.TML.cs.
```

direwolf420 (commented):
Good PR, surface level review here.

Referenced style conventions: <https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide>

Cerbrus:
Thanks for your review, @direwolf420. I've applied the suggested changes!

JavidPack:
If I'm reading this right, it doesn't look like sightColor is passed into the new hooks. Isn't that the whole point of the hook, to assign a color? I don't think modders would want to be limited to just White.

Cerbrus:
> If I'm reading this right, it doesn't look like sightColor is passed into the new hooks. Isn't that the whole point of the hook, to assign a color? I don't think modders would want to be limited to just White.

There was no mention of that in the issue, but I'll see if I can add that.

Cerbrus:
I've added `sightColor`, @JavidPack :D

direwolf420:
Should mention its default value/behavior in the docs

Cerbrus:
Like that, @direwolf420 ?

JavidPack:
Thanks, good job.

Cerbrus:
Glad to be of help!

PR #3604, opened 2023-07-06
Info Display Fixes

Destructor-Ben:
> Note: The changes in this PR have been superseded by the changes in https://github.com/tModLoader/tModLoader/commit/3ca9bf16a3722c35ea86377b7a28f4456b072743

### What is the new feature?
Allowed mods to modify an info displays shadow colour and allowed info displays to have a custom hover texture.
Currently IL edits are required to change an info displays hover texture or text shadow colour. In EM, the info display is square, which doesn't match up with the vanilla hover texture, which is circular. It is also impossible to use a golden colour for an info display because the shadow colour can't be customized.
~~The vanilla hover texture could be the default value instead of a custom one, and this would also make slightly less breakage for mods.~~
#### Shadow Colour
```csharp
public override string DisplayValue(ref Color displayColor, ref Color displayShadowColor) {
	displayColor = GoldInfoTextColor;
	displayShadowColor = GoldInfoTextShadowColor;

	return "Display value";
}
```
#### Hover Texture
```csharp
public override string HoverTexture => Texture + "_Hover";// By default is VanillaHoverTexture;
```
### ExampleMod updates
Made ExampleInfoDisplay use a custom hover texture since it is square.
Also showcases using a custom color in some situations.

### What is the bug?
In vanilla, when hovering over an info displays icon, it makes the info display below it have a transparent label to help see the mouse text. In tML, this is messed up because it has to handle display paging and other things, which mess with the 'i' variable that vanilla uses
### How did you fix the bug?
I fixed this by changing the remaining 'i's with 'drawnCounts' in the drawing loop.
No

## Porting Notes
- Change `GlobalInfoDisplay.ModifyDisplayColor` in 1.4.4 mods to have the shadow colour parameter (tModPorter)
- Change `InfoDisplay.DisplayValue` to have the shadow colour parameter (tModPorter)
- Override `InfoDisplay.HoverTexture` to supply your own hover texture named `{InfoDisplayName}_Hover` if suitable.


JavidPack:
Have you ever done tModPorter work? It would be great if we had tModPorter work with this, I think it should also add a comment that explains that either a _Highlight texture is now needed or to override HoverTexture and return VanillaHoverTexture.

The actual tModporter code would just be a ChangeHookSignature line for the hooks that changed, but there would also be associated tests. 

If you know how to do that, great, otherwise I'll prioritize making a guide for that, since I don't think we explain that anywhere.

Destructor-Ben:
I haven't done anything with tModPorter before so a guide would be great.

JavidPack:
Here is the guide: https://github.com/tModLoader/tModLoader/wiki/Contributing-to-tModPorter

Let me know if anything is confusing or if you need any other help.

JavidPack:
Good job. I went ahead and made the vanilla texture the default, I figure that will be what most modders will want if they just following vanilla textures as a guide.

PR #3605, opened 2023-07-06
Add Exception handling for NativeLibrary.TryLoad in RGB Devices

Solxanich:
NativeLibrary.TryLoad will still throw exceptions such as:
1) BadFormat
2) Dll Not Found
And a few others, making it unsafe to call without having an additional try-catch wrapping.

This PR adds Try-Catch Wrapping to eliminate surprise crashes when one of these happens due to unexpected circumstances on a particular users device.

Example: https://discord.com/channels/103110554649894912/1126562463601672223
EDIT for Discord context:
This particular example suggests CUESDK fails to load, despite being present in the folder with sufficient priviledges, due to assumed cascading native dependency errors from a dependency in a restricted folder.

Due to the nature of this bug being difficult to manage from our end, and the rare occurrence of RGB users and this bug, it is proposed to try-catch the exception.


Chicken-Bones:
Hmmmm. This should be done as a wrapper. A good place is `NativeLibraries` (our own class). 

Note that this code was previously considered exception safe, as the only case when `TryLoad` will fail is if one of the native libararies in the _tML natives folder_ fails to load, which is normally fatal. 

The Discord thread you linked the user had a `CUESDK.x64_2019.dll` in their `Libraries\Native\Windows` folder, ~~which we don't ship... so how did it get there??~~ which I guess is an optional dep. I wonder why it's giving `E_ACCESSDENIED`

PR #3606, opened 2023-07-07
Remix Hallow worldgen is now in the "Hardmode Good" genpass

Lion8cake:
### What is the new feature?
Remix hallow worldgen now is properly included in the Hardmode Good genpass that tmod adds. This also makes the Hardmode Evil also run in remix seeds aswell

This allows worldgen that relies on either Hardmode Good or Hardmode Evil to be ran in remix seeds.

Not that I can identify, the current patch for the remix seed was a bad fix when porting tmodloader to 1.4.4

Make a new mod, make worldgen that relies on the Hardmode Good pass to run. Results on the remix seed


Chicken-Bones:
Please do this in a way which minimises patches. We cannot be changing indentation or re-ordering large amounts of vanilla code.

Should the hardmode evil task run on remix at all? It doesn't generate evil.

Since the remix hardmode good task is completely different to the regular one, does it make sense to give it the same name? Why not give it a different task name, like `Hardmode Good Remix`?

Chicken-Bones:
Please send me a DM in Discord or post in `#collaborators` if you need any help with the feedback

Lion8cake:
Sorry for the late comment, been on holidays for the past 9-10 days
1. I've tried to not move as much vanilla code this time, im unsure if there is more i can minimise
2. I do think the hardmode evil should run in the remix seed, this will give the option for modders using hardmodeevil to also have their own worldgen in the remix seed without needing to change genpasses
3. I do believe the remix hallow worldgen should stay in the same HardmodeGood genpass both for consistancy with vanilla's normal world genpasses and also to again not limit those using the HardmodeGood genpass into being locked out in the remix seed.

Chicken-Bones:
@direwolf420 @Ozzatron @JavidPack requesting a weigh in here.

Context:
- In 1.4.3/1.4.4 currently, the hallow and evil 'strips' that get generated in hardmode are wrapped in `new PassLegacy("Hardmode Good", HardmodeGoodTask)` and `new PassLegacy("Hardmode Evil", HardmodeEvilTask)`, and passed to `SystemLoader.ModifyHardmodeTasks`
- In remix worlds, no extra evil is generated, and hallow is generated all throughout the underground in the left or right-most 1/7th of the world.
  - It effectively uses the clentaminator (`WorldGen.Convert`), as opposed to the more hard-coded `GERunner` that regular hm evil uses.
- In 1.4.4 remix seeds _currently_ there is no `"Hardmode Good/Evil"` task, and the remix code runs without being passed to `ModifyHardmodeTasks`. 
- This PR puts the remix hardmode hallow gen code into a `PassLegacy` task.

The _question_ is whether this task should be called `"Hardmode Good"` or `"Hardmode Good Remix"`

If labelled `"Hardmode Good"`:
- Mod code which modifies or inserts before/after the `"Hardmode Good"` task will run on remix.
- This could be either good (automatic compatibility) 
- Or bad because a mod expects `"Hardmode Good"` to mean the standard `GERunner` strip worldgen

If we choose to call it `"Hardmode Good"`, should we add an _empty_ `"Hardmode Evil"` task to remix seed worldgen?
- This could be good, because mods which want to locate a task after it can still do so.
- Or bad, because it doesn't do anything so removing/disabling it is pointless

direwolf420:
I favor `"Hardmode Good Remix"`, modders need to update their code for 1.4.4 anyway. The other options all have some hidden implications which are not clear to modders. The availability of the new pass should be communicated via docs

Chicken-Bones:
With #3642 we should probably include all passes and disable the ones that don't apply in remix/regular by default. That should be a workable solution for everyone

Chicken-Bones (approved):

JavidPack:
Thanks a bunch.

PR #3615, opened 2023-07-10
Fix compilation after PR #3595

direwolf420:
### What is the bug?
#3595 had bad code for `BuildGlobalHook` as during its development a `ref` paramater was added.
Why it's a thing in the first place: <https://github.com/dotnet/roslyn/issues/57517>

### How did you fix the bug?
Add generic constraints

No


Chicken-Bones:
Certainly this is fixed in VS2022 no?

direwolf420:
I'm using VS 2022, my latest sdk is 6.0.406

Issue #3619, opened 2023-07-11
enabling a mod doesn't enable its dependency


cozykoko:
Credit goes to solxan

### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

 For example enabling Mod of Redemption doesn't enable the dependencies automatically too. It would be nice to have it enabled automatically for unaware players too, since you can't play the mod without it.

PR #3628, opened 2023-07-14
Automatically enable depencencies

Cerbrus:
When enabling a mod, automatically enable available dependencies.

### What is the new feature?

When enabling a mod, the mod's dependencies (if installed) will automatically be enabled as well.
When disabling a mod, it will not disable the dependencies. This change only affects disabled -> enabled.

This was requested in issue #3619

It's a small quality of life improvement.

Possibly?

I updated `Modloader.cs` to enable the dependencies, and `UIModItem.cs` to make sure the UI updates for the toggled dependencies.

- Install some mods that have (weak) dependencies, and their dependencies.
- Disable all mods.
- Enable a mod that has dependencies, and you should see the dependencies enable as well.

### ExampleMod updates

An easy test is to make `ExampleMod` dependant on something in its `buid.txt`:
```
modReferences = CheatSheet
```

I thought it'd be overkill to actually make `ExampleMod` dependant on something else, as that could complicate development unnecessarily.

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
@@ -842,7 +842,6 @@
 		// "Vector2.Y.Label": "Y"
 	},
 	"Mods": {
-		// "ModLoader.Unloaded": "Unloaded",
 		"ModLoader.Items.AprilFools.DisplayName": "Aprilscherz",
 		"ModLoader.Items.StartBag.DisplayName": "Starterpaket",
 		"ModLoader.Items.StartBag.Tooltip": "Einige Starter-Items haben nicht in dein Inventar gepasst",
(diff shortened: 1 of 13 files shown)


Chicken-Bones:
`ModOrganizer.FindMods` is not meant to be frequently called, putting it in `SetModEnabled` will significantly slow down performance when enabling mod packs or large numbers of mods and spam the logs. 

I think this change only needs to be applied to the `UIModItem` and the server mods menu at the moment.
`UIModItem` already has a `_modReferenceIcon` and an `EnableDependencies` function, so for this implementation that icon and tooltip may need changing/removing.

Assigning to @JavidPack

JavidPack:
New behavior. As discussed, we only enable, not disable dependencies. 

https://github.com/tModLoader/tModLoader/assets/4522492/7e252b2e-65ee-44e8-84a4-ef90f7227eee

JavidPack:
Thanks a lot for taking the initiative with this.

PR #3636, opened 2023-07-17
add DamageClassLoader.GetDamageClass

direwolf420:
### What is the new feature?
new `DamageClassLoader.GetDamageClass` method for retrieving a `DamageClass` instance from its type.

Modders have currently no way to actually use `DamageClass.Type` as a lookup for the `DamageClass` instance. TML itself does it in i.e. netcode, modders might need it too.

No. This mirrors existing `XLoader.GetX` methods.

```cs
binaryWriter.Write7BitEncodedInt(damageClass.Type);
//
DamageClass damageClass = DamageClassLoader.GetDamageClass(binaryReader.Read7BitEncodedInt());
```

### ExampleMod updates
None

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


JavidPack:
Seems reasonable.

Issue #3640, opened 2023-07-20
Hide tooltip lines instead of removing them


BasicallyIAmFox:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Add `TooltipLine.IsHidden` property.

### What does this proposal attempt to solve or improve?

Currently, modders are forced to remove tooltip line if they want to hide it.
Removing potentially leads to mod conflicts. (In fact there were cases of that in past)

JavidPack:
Related: #2072

Chicken-Bones:
I think this is a simple subset of the larger and more controversial proposal that is easy to implement. Go ahead and PR. Follow the `PlayerDrawLayer` example and do
```cs
public bool Visible { get; private set; } = true;
public void Hide() => Visible = false;
```

Chicken-Bones:
You may also consider doing the same thing to `GenPass` (`Enabled` and `Disable()`)

PR #3641, opened 2023-07-20
Now TooltipLine instances can be hidden.

BasicallyIAmFox:
### What is the new feature?
#3640 

#3640 

None that I'm aware of, no.

```
int index = tooltips.FindIndex(x => x.Mod == "Terraria" && x.Name == "ItemName");
// Hide Item's Name tooltip line. This realistically should be always safe to do since every item has a name
tooltips[index].Hide();
```

### ExampleMod updates
`ExampleTooltipsItem` was changed to encourage modders use `Hide` instead of `Remove` methods.


Chicken-Bones:
Don't recommend `FindIndex` just use `FirstOrDefault()` and `?.Hide()`. Please update the commented example to do that

PR #3642, opened 2023-07-20
Made GenPass instances hideable.

BasicallyIAmFox:
### What is the new feature?
Makes GenPassses hideable.

Same reason as #3640 

None that I'm aware of, no.

```
int index = tasks.FindIndex(genpass => genpass.Name.Equals("Reset"));
// No mod should realistically should remove genpasses.
tasks[index].Disable();
```

### ExampleMod updates
Should be none, I believe.

### DIFF
--- a/patches/tModLoader/Terraria/WorldGen.cs.patch
+++ b/patches/tModLoader/Terraria/WorldGen.cs.patch
@@ -566,7 +566,7 @@
  					bool flag2 = TileID.Sets.Dirt[tile2.type];
  					if (notTheBees)
  						flag2 = flag2 || TileID.Sets.Mud[tile2.type];
-@@ -13849,12 +_,35 @@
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -582,6 +582,8 @@
 +	private static void GenerateWorld_RunTasksAndFinish(int seed, Stopwatch generationStopwatch, GenerationProgress customProgressObject)
 +	{
 +		SystemLoader.ModifyWorldGenTasks(_generator._passes, ref _generator._totalLoadWeight);
++
++		_generator._passes.RemoveAll(x => !x.Enabled);
```
Chicken-Bones: Should go in the `Modify` method

PR #3644, opened 2023-07-21
Fix spelling error

thatrobotdev:
Sorry, this doesn't really fit with any of the PR templates, so hopefully this is in a manageable form!

I was reading the terms of use and found a spelling mistake with the word "complay" in place of "comply", so this just fixes the error.


Solxanich:
Thanks for catching that!

PR #3646, opened 2023-07-21
Example Last Prism Port

Vaema:
Ported over Example Last Prism and its projectiles.
I did test it beforehand to make sure it functions correctly, and it did.


JavidPack:
Thanks. I noticed the mod didn't actually load since you messed up the hjson by trying to manually populate it. You should make sure the mod loads when you make a PR, don't just assume you copied over everything correctly.

Thanks a bunch.

PR #3659, opened 2023-07-25
SChinese small update

IBRS-4Ever:
Added new strings, currently not localized. (I'm not the original translator, so I leave them un-touched, wait for @Dierney to translate them.)


JavidPack:
It looks like you uncommented a bunch but didn't translate them. Did you forget them? Uncommenting things but not translating them only hinders, it doesn't help.

IBRS-4Ever:
It's my fault, I have reset the tModLoader.json now

JavidPack:
Ok, I'm thinking there is some confusion. Adding the commented out entries to the `tModLoader.json` files are done automatically to all languages by running the scripts, which we run periodically, we don't need PRs for that. 

Everything in the SteamWorkshopLocalization folder is populated automatically from the `tModLoader.json` files by other scripts that we run periodically, so those shouldn't be manually edited either. 

I ran these scripts earlier today, so there isn't anything in this PR that should be merged, so I'm going to close it.

PR #3684, opened 2023-08-02
TileID.Sets To Let Tiles Be Placed Next to Non-Solid tiles

Rijam:
### What is the new feature?

Added a new tile ID set `TileID.Sets.CanPlaceNextToNonSolidTile[]` which allows players to place tiles next to non-solid tiles.

By default, you cannot place tiles on non-solid tiles. In vanilla, `Player.PlaceThing_Tiles_BlockPlacementForAssortedThings()` has an exception for Cobwebs, Coin Piles, Living Fire Blocks, Smoke Blocks, and Bubble Blocks which lets you place them on other non-solid tiles. Since the values are hardcoded, it is extremely difficult to make a non-solid mod tile that behaves in the same way.

Using `TileID.Sets.IsBeam[]` comes close, but there are still placement differences. Beams can be placed on each other, but they can't be placed on other non-solid tiles. Here is a gif example between Beams and Living Fire:

BeamVsLivingFire

* As far as trying to get a mod tile with this placement style, you could mess around with `TileObjectData.newTile.HookCheckIfCanPlace`, `TileObjectData.newTile.UsesCustomCanPlace = true`, and making a CanPlace hook. I've tried it and gotten it to do the placement correctly, but it was causing an object reference not set to an instance of an object error on the player when denying the placement (which causes the player's vanity and animations to break). You could instead allow the player to place the tile anywhere (midair), but that is a compromise rather than a real solution.

<details><summary>Old bullet points</summary><blockquote>

* If anyone thinks of a better name for the set, suggest it. Maybe `NonSolidCanPlaceOnNonSolid` or `NonSolidCanBePlacedOnNonSolid`?

* For the patch. I used `/* */` to comment out the vanilla code. I thought putting a `//` directly on the line made an uglier looking patch. The first example is the patch I committed. Tell me if one is more preferred:

```diff
@@ -31729,7 +_,10 @@
 		else if (inventory[selectedItem].createTile == 275 || inventory[selectedItem].createTile == 276 || inventory[selectedItem].createTile == 277) {
 			canPlace = true;
 		}
+		/*
 		else if (inventory[selectedItem].createTile == 51 || inventory[selectedItem].createTile == 330 || inventory[selectedItem].createTile == 331 || inventory[selectedItem].createTile == 332 || inventory[selectedItem].createTile == 333 || inventory[selectedItem].createTile == 336 || inventory[selectedItem].createTile == 340 || inventory[selectedItem].createTile == 342 || inventory[selectedItem].createTile == 341 || inventory[selectedItem].createTile == 343 || inventory[selectedItem].createTile == 344 || inventory[selectedItem].createTile == 379 || inventory[selectedItem].createTile == 351) {
+		*/
+		else if (TileID.Sets.NonSolidCanPlaceOnEachOther[inventory[selectedItem].createTile]) {
 			if (Main.tile[tileTargetX + 1, tileTargetY].active() || Main.tile[tileTargetX + 1, tileTargetY].wall > 0 || Main.tile[tileTargetX - 1, tileTargetY].active() || Main.tile[tileTargetX - 1, tileTargetY].wall > 0 || Main.tile[tileTargetX, tileTargetY + 1].active() || Main.tile[tileTargetX, tileTargetY + 1].wall > 0 || Main.tile[tileTargetX, tileTargetY - 1].active() || Main.tile[tileTargetX, tileTargetY - 1].wall > 0)
 				canPlace = true;
 		}
@@ -32318,7 +_,9 @@
```

```diff
@@ -31729,7 +_,8 @@
 		else if (inventory[selectedItem].createTile == 275 || inventory[selectedItem].createTile == 276 || inventory[selectedItem].createTile == 277) {
 			canPlace = true;
 		}
-		else if (inventory[selectedItem].createTile == 51 || inventory[selectedItem].createTile == 330 || inventory[selectedItem].createTile == 331 || inventory[selectedItem].createTile == 332 || inventory[selectedItem].createTile == 333 || inventory[selectedItem].createTile == 336 || inventory[selectedItem].createTile == 340 || inventory[selectedItem].createTile == 342 || inventory[selectedItem].createTile == 341 || inventory[selectedItem].createTile == 343 || inventory[selectedItem].createTile == 344 || inventory[selectedItem].createTile == 379 || inventory[selectedItem].createTile == 351) {
+		// else if (inventory[selectedItem].createTile == 51 || inventory[selectedItem].createTile == 330 || inventory[selectedItem].createTile == 331 || inventory[selectedItem].createTile == 332 || inventory[selectedItem].createTile == 333 || inventory[selectedItem].createTile == 336 || inventory[selectedItem].createTile == 340 || inventory[selectedItem].createTile == 342 || inventory[selectedItem].createTile == 341 || inventory[selectedItem].createTile == 343 || inventory[selectedItem].createTile == 344 || inventory[selectedItem].createTile == 379 || inventory[selectedItem].createTile == 351) {
+		else if (TileID.Sets.NonSolidCanPlaceOnEachOther[inventory[selectedItem].createTile]) {
 			if (Main.tile[tileTargetX + 1, tileTargetY].active() || Main.tile[tileTargetX + 1, tileTargetY].wall > 0 || Main.tile[tileTargetX - 1, tileTargetY].active() || Main.tile[tileTargetX - 1, tileTargetY].wall > 0 || Main.tile[tileTargetX, tileTargetY + 1].active() || Main.tile[tileTargetX, tileTargetY + 1].wall > 0 || Main.tile[tileTargetX, tileTargetY - 1].active() || Main.tile[tileTargetX, tileTargetY - 1].wall > 0)
 				canPlace = true;
 		}
@@ -32318,7 +_,9 @@
```
</blockquote></details>

Added ExampleLivingFire and ExampleLivingFireTile. See below.

### ExampleMod updates

Added ExampleLivingFire and ExampleLivingFireTile. 
* The item defines a static Vector3 for the light color which is used several times instead of re-typing the color values for each spot.
* The tile showcases the new set as well as moving the drawing down 2 pixels in `SetDrawPositions()` and animating the tiles with `AnimateTile()`. (The only difference between ExampleLivingFireTile and the vanilla Living Fire tiles is that the vanilla Living File has a different color for the map and doesn't reuse the light color.)


ExampleMod/Content/Items/Placeable/ExampleLivingFire.cs
```diff
@@ -0,0 +1,44 @@
...
+			Item.maxStack = Item.CommonMaxStack;
+			Item.consumable = true;
+			Item.createTile = ModContent.TileType<Tiles.ExampleLivingFireTile>(); // Set this to the tile that we want to create.
+			Item.width = 12;
+			Item.height = 12;
+		}
+		public override void PostUpdate() {
+			// Add some lighting when the item is dropped in the world.
+			// Curiously, only the regular Living Fire Block creates light.
+			Lighting.AddLight((int)((Item.position.X + (Item.width / 2)) / 16f), (int)((Item.position.Y + (Item.height / 2)) / 16f), LightColor.X, LightColor.Y, LightColor.Z);
```
Rijam: Simplify by using Item.Center instead of position + width/height /2

JavidPack:
Very good. I ended up renaming it since it doesn't just affect non-solid tiles themselves. 

Also, your AnimateIndividualTile code does not animate the tile more randomly than the AnimateTile approach, it just runs the same logic for every tile on screen rather than once for the tile type. To animate more randomly you would need to adjust the frameYOffset using the tile coordinates to have the animation desync from the normal tile animation cycle.

Issue #3315, opened 2023-03-16
Tile documentation


JavidPack:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

It would be beneficial if the Tile struct were documented. The Tile class itself should mention the concept of how vanilla code and tmod code differs for the sake of clarity.

Of particular note is how HasTile is important and should be checked before other properties relating to the tile. Also, modders should be reminded to use HasUnactuatedTile for collision checks rather than HasTile. 

Other interesting things is how walls can exist despite HasTile being false.

Finally, each doc summary should have a section explaining the corresponding vanilla code approach, so modders can easily verify that they didn't make any mistake while porting. 

For example, the docs for `HasActuator` could mention `The getter corresponds to <c>tile.actuator()</c> and the setter corresponds to <c>tile.actuator(value)</c>`

### What does this proposal attempt to solve or improve?

Help modders understand and corroborate vanilla decompiled code.

PR #3685, opened 2023-08-02
Tile Documentation

Destructor-Ben:
### What is the new feature?
Added XML documentation for the `Tile` struct.
Issue: https://github.com/tModLoader/tModLoader/issues/3315

Note: the documentation for internal methods showed for Example Mod so I assumed it would show for all mods, however I'm stupid and that's not the case.

See issue

It's documentation so yes. I would much appreciate any feedback.

N/A

### ExampleMod updates
N/A


JavidPack:
Have you checked if docs for internal fields show for modders? I'm not sure if it shows or if it just says it's not accessible, would be good just to make sure.

Destructor-Ben:
Ah crap, it worked in ExampleMod so I went ahead with it but it doesn't work with any other mod.

Solxanich:
The 'cell' based naming is a bit confusing. 

I think we've generally used Location and Position the most throughout the code.
I think it might be best to stick with that?

Example: Copy's locational data from this location to new location.

PR #3703, opened 2023-08-08
Mod Icon Bug Fixes

Destructor-Ben:
### What is the bug?
- When publishing a mod, there is a weird border around the mod icon
- Mod icons can sometimes disappear in the manage mods menu because they aren't 80x80 pixels in size

### How did you fix the bug?
- Removed the weird border if a mod is being published
- Added a fallback texture in case a mod icon can't be found, and otherwise forcibly scales all mod icons to 80x80

Unless there is code that forces a mod to have an 80x80 icon (I haven't checked), then mods can now have non 80x80 icons. This could have many implications, and I don't know if there is a reason there was a forced 80x80 size apart from displaying in the mods menu if `ScaleToFit` didn't exist. Additionally, the mod browser scales the icons since it uses the steam workshop icons. This means that `icon_workshop.png` will most likely become redundant with `icon.png` becoming larger and used for workshop and in-game.

For the mod icon, it could be forced for mods to have an 80x80 icon in the mod building instead.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
@@ -84,6 +84,18 @@
  		uiList.Add(CreatePreviewImageSelectionPanel("image"));
  		uiList.Add(CreatePublicSettingsRow(0f, 44f, "public"));
  		uiList.Add(CreateTagOptionsPanel(0f, 44, "tags"));
+@@ -124,7 +_,10 @@
+ 		uIElement.SetPadding(0f);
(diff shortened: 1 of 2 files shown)


Solxanich:
@JavidPack  - I'm not familiar with the original decision making around the 80x80 forced mod icon size for local mods.
Would you please review this PR with those historical context?

I'm good with the WorkshopPublishInfoState change to remove Achievement_Borders being overlaid.

JavidPack:
I decided that since 80x80 is in the docs and mods could be loading and displaying them, it is good to keep it so we don't introduce issues for those mods.

We can revisit the idea later if we want to get rid of the 2 png files, but right now I don't think we should mess with what is working.

PR #3714, opened 2023-08-09
Change SplitStack() favorited behavior to match OnStack()

andro951:
### What is the bug?
OnStack() causes the favorited status to follow the item to the destination item, where SplitStack() removes favorited status from the destination item.   
Example:
Stack of 4 throwing knives in your inventory that is favorited. Right click the stack once, now 1 in the mouse, 3 in your inventory.  The stack in your inventory is favorited and the on in the mouse is not.  Right click a 2nd time, and the inventory stack looses favorited and the mouse stack gains favorited.

### How did you fix the bug?
Replacing "destination.favorited = false;" in SplitStack() with the same code from OnStack() for swapping favorited.

Create a separate function for SwapFavorited() called by SplitStack() and OnStack() to prevent duplicated code.  Let me know if you would like me to switch it to that.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
@@ -1566,7 +1566,10 @@ public static class ItemLoader
 	public static void SplitStack(Item destination, Item source, int numToTransfer)
 	{
 		destination.stack = 0;
-		destination.favorited = false;
+		if (source.favorited) {
+			destination.favorited = true;
+			source.favorited = false;
+		}
 
 		foreach (var g in HookSplitStack.Enumerate(destination)) {
 			g.SplitStack(destination, source, numToTransfer);


JavidPack:
I don't think this fix works, it seems to result in the opposite behavior of Terraria.

In vanilla, right clicking a favorited item leaves the inventory item favorited. With this fix, it makes the mouse item favorited and removes the favorited flag from the inventory item. It seems like StackItems is the culprit, it removes the favorited from the source item. I don't know if that is intentional, StackItems is called in other situations.

Chicken-Bones:
Need to be careful and consider all the possible uses of item stacking. I believe stacking needs some context on whether you're transferring to or from the 'hand'. Favourited item status is meant to be attracted to inventory slots.

Currently the from-hand behavior works correctly. If I pick up a stack of 8000 favorited bullets, and put them onto a stack of 4000 unfavorited bullets, I should end up with a stack of 9999 favorited in the inventory, and 2001 left in my hand without the favorite status.

Chicken-Bones:
Fixed in e64e561f4a4a9b890df8330dadee366882c1493c

Issue #2795, opened 2022-08-02
Add icon to indicate mod side


NotLe0n:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

Mod cards in the mod-list and mod-browser should have a small icon - like the Steam Workshop icon, to indicate the mod side.

![image](https://user-images.githubusercontent.com/26361108/182416509-f01e15bd-d74a-45db-a2e4-839b689cc523.png)
(see the red icon on the right, next to the steam icon)

The icon should be designed by someone qualified. The image is just a mockup to show a possible location for the icon.

### What does this proposal attempt to solve or improve?

Currently you have to go to the steam workshop page to find out the mod side.
![image](https://user-images.githubusercontent.com/26361108/182417084-fc7e2c0c-a018-4c9d-a789-c8be70782dd0.png)

 There is no way to see that information in game.

emyhrberg:
I made a mod for this: https://steamcommunity.com/sharedfiles/filedetails/?id=3438106830

PR #3718, opened 2023-08-11
Add mod side icon

stormytuna:
Implements #2795 

### What is the new feature?

Adds an icon to Mods in the Mods List which denotes their side (both, client, server, or nosync)

Icon on a non-workshop mod
image

Icon on a workshop mod
image

Icon for both sides
ModSideBothIcon

Icon for client side
ModSideClientIcon

Icon for server side 
ModSideServerIcon

Icon for no sync
ModSideNoSyncIcon

Icon for an unknown side
ModSideUnknownIcon

See issue

As with any UI/UX there are many different ways to display it, I just implemented what the issue suggested


JavidPack:
This would definitely need some hover text explaining the icons. Note that there already is a server side mod icon that shows up for server side mods, so we can get rid of that. Also it would be useful to implement the same for the workshop entries. 

image

JavidPack:
Looks like after some discussion in the Discord collaborators channel chat, we've decided that an icon doesn't really solve anything and would just be visual clutter, so this PR will go **on hold** for now. Don't worry about implementing the things I just mentioned for now, we can revisit this later if the decision changes. 

Thanks for putting in the work, it really helps.

PR #3725, opened 2023-08-14
Re-enable Auto-Reload and Enable Mods Leaving ModBrowser

Solxanich:
In 1.3 tModLoader, we had a feature that would auto-enable newly downloaded mods from the Mod Browser, and reload on exit.

This behavior was opt-out via an in-game tModLoader Settings option.

This PR enables the last bit of code to bring that back; question is should we.

ALTERNATIVE:
We modify the "New Mods Detected" on Launch UI to be generalized to be usable with Mod Browser, ensure dependencies are downloaded regardless of choice, and make it a choice on that UI whether they want to auto-enable the mods or "I will Enable My Mods"


JavidPack:
As a note, I recently removed the option from displaying in the settings menu https://github.com/tModLoader/tModLoader/commit/2118314b5c8f436c0433428ae721887bd36067fd . 

I think it's fine if we close this, I don't think users have requested or suggested this behavior at all recently.

PR #3726, opened 2023-08-14
Auto Enable Dependency Mods on Reload

Solxanich:
Adjusts the reload logic to ensure that all dependencies, if on the system, are enabled during mod reload.

Otherwise tosses an error within the try-catch in SelectAndSortMods.

Also some minor refactors of existing dependencies checking code to minimize code duplication.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -328,7 +328,6 @@ internal static class ModOrganizer
 		var modsToLoad = availableMods.Where(mod => mod.Enabled && LoadSide(mod.properties.side)).ToList();
 		VerifyNames(modsToLoad);
 
-		EnsureDependenciesExist(modsToLoad, false);
 		foreach (var mod in modsToLoad) {
 			EnableWithDeps(mod, availableMods);
 		}
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -325,17 +325,12 @@ internal static class ModOrganizer
 		//}
 		Interface.loadMods.SetLoadStage("tModLoader.MSFinding");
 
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -329,7 +312,12 @@ internal static List<LocalMod> SelectAndSortMods(IEnumerable<LocalMod> mods, Can
 		VerifyNames(modsToLoad);
 
 		try {
-			EnsureDependenciesExist(modsToLoad, false);
+			// Guarantee that all non-weak dependencies are enabled when going to reload
+			if (!EnsureAllDependenciesEnabled(modsToLoad, false)) {
+				modsToLoad = mods.Where(mod => mod.Enabled && LoadSide(mod.properties.side)).ToList();
+				VerifyNames(modsToLoad);
+			}
+
```
Chicken-Bones: Aside from the duplication of the above logic. The main issue with the implementation is that it doesn't correctly handle recursive/transitive/nested dependencies. The code for recursively enabling dependencies should go before the `var modsToLoad` line.

I also think (haven't tested) that it won't actually work correctly, because you pass `modsToLoad` rather than `mods` to `EnsureAllDependenciesEnabled` which means there aren't any disabled mods in the list to enable.

On reflection, I don't think it's necessary to combine the logic for missing or invalid deps, with the 'enable deps' logic. In the current form it doesn't make it simpler. 

I think it could be as simple as:
```cs
		void EnableWithDeps(LocalMod mod)
		{
			mod.Enabled = true;

			foreach (var depName in mod.properties.RefNames(includeWeak: false)) {
				if (mods.SingleOrDefault(m => m.Name == depName) is LocalMod { Enabled: false } dep)
					EnableWithDeps(dep);
			}
		}

		foreach (var mod in mods.Where(mod => mod.Enabled))
			EnableWithDeps(mod);
```

Chicken-Bones (approved):
LGTM. @Solxanich feel free to merge if you've tested this and it works

PR #3734, opened 2023-08-17
Update ru-RU tModLoader.hjson

RazzSG:


patches/tModLoader/Terraria/Localization/Content/ru-RU/tModLoader.json
```diff
@@ -551,8 +551,8 @@
 		"TagsLanguage_French": "Французский",
 		"TagsLanguage_Spanish": "Испанский",
 		"TagsLanguage_Russian": "Русский",
-		"TagsLanguage_Chinese": "Китайский упрощённый",
-		"TagsLanguage_Portuguese": "Бразильский португальский",
+		"TagsLanguage_Chinese": "Китайский",
+		"TagsLanguage_Portuguese": "Португальский",
```
JavidPack: I took бразильский португальский and китайский упрощенный from the Language select page in the Settings menu, are you sure this is correct? If you are just trying to adjust how long they are, we can fix that elsewhere, but it looks like you are removing the "brazillian" and "simplified" aspects of those language names. I don't like introducing an inconsistency, if I'm reading this right.

RazzSG: I adjusted the text because it was too long. I'd appreciate it if you could fix this issue. If you disagree with this change, I can revert it.
image

Blueberryy: You can just short those like: "Китай. упрощ." and "Бразил. порт."

RazzSG: Abbreviating words is considered impolite in translations. It's better to use full words for clarity and respect.

JavidPack:
I've made a note in #3769 for myself to fix the text overflow.  Thanks

Issue #3733, opened 2023-08-17
Add a Disabled field to IItemDropRule or similar


Jacklz:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently if a modder wants to prevent a loot drop from an item bag or an npc they have to completely remove the loot from the loot table. Is it possible to take the same route for IItemDropRule that was taken for recipes and NPC Shop items? Where a Disabled field was given to determine if the recipe/shop item was "removed" from the pool, but still accessible. Loot can also be removed/added while in game unlike recipes that can only be added on a mod reload. If a mod depended on a specific droprule that was then removed during the game, it could cause some problems?

### What does this proposal attempt to solve or improve?

This could allow for mods to see/access loot rules that were "removed" from the loot table, and allow for further editing and re-enabling

### Which (other) solutions should be considered?

Not sure about alternatives

Chicken-Bones:
Approved, someone do it this month please!

BasicallyIAmFox:
> Approved, someone do it this month please!

atp could have Conditions list

Issue #2759, opened 2022-07-25
Simplify/Rework drop rules


Chicken-Bones:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

The current drop rules system is really hard to modify and work with. This is due to a couple factors:
- every rule gets its own 'wrapper chance', even if it's a 'selection' rule (like `OneFromRulesRule` or `FewFromRulesRule` etc
  - There only needs to be one rule which wraps another rule in a 'chance'
  - Stuff like `OneFromRulesRule` and `ItemDropWithConditionRule` don't need their own chance.
  - Luck scaling could just be one parameter of a 'chance' rule rather than luck and noluck variants for every rule
- there's no way to inspect the actual chance of a rule returning 'success' (and thus triggering a chain). If it is a 'wrapper rule'
  - Eg, one-from-rules with chance 1 always succeeds, but it the wrapped rule has its own chance, then it that may fail
- 'drop style' rules like `DropLocalPerClientAndResetsMoneyTo0` and `DropOneByOne` and `DropPerPlayerOnThePlayer` have nothing todo with the chance, but require yet another rule type, making pattern matching harder for modders
- `DropBasedOnExpert/MasterMode` can't be generalised to modded difficulties. They should just be conditional rules.
- There's no need for `ItemDropWithConditionRule` it could just be a chained rule on `LeadingConditionRule`

All this makes rules a pain to modify by modders. We should provide some helper methods at least.
Though finding the right rule-chain to modify is also likely to be a problem, potentially there should be a more compatible version of modify loot hooks where modders can modify the rules by adding to tables or altering rates, but not change the overall structure, and then a second pass where the structure can be changed (to add more conditions or such)

ThomasThePencil:
I like what you're goin' for here, but as someone who's had to mess with vanilla rules extensively for DD, I...can't help but feel that the initial pitch misses the mark a little bit on why so many, myself included, have found this new system to be a pain despite its many superiorities over the pre-1.4 approach. as such, I'd like to throw my hat into the ring here

[...]

PR #3743, opened 2023-08-20
Disabeable item drop rules

BasicallyIAmFox:
### What is the new feature?
Adds `IItemDropRule.Disabled` property

See #3733 

The one and patch heavy is adding common, base item drop rule class and make every drop rule inherit it. Additionally commenting out most item drop rule fields and properties.

See ExampleMod updates.

### ExampleMod updates
Added example of disabling Reaper's Death Sickle drop rule.


Jacklz:
Should adding/disabling loot be prevented during the game and only permitted during reload? Like with recipes.

BasicallyIAmFox:
> Should adding/disabling loot be prevented during the game and only permitted during reload? Like with recipes.

i dont see any realistic reason why anyone would do that tbh

Jacklz:
> > Should adding/disabling loot be prevented during the game and only permitted during reload? Like with recipes.
> 
> i dont see any realistic reason why anyone would do that tbh

i dont see why either, but its more about consistency and taking precautions. Why were recipes locked to reload only?

Chicken-Bones:
> > > Should adding/disabling loot be prevented during the game and only permitted during reload? Like with recipes.
> > 
> > 
> > i dont see any realistic reason why anyone would do that tbh
> 
> i dont see why either, but its more about consistency and taking precautions. Why were recipes locked to reload only?

So that mods can perform computations at load time like whether an item is a material, or which recipes can be used to craft an item, without re-checking the enabled conditions every frame/whenever

direwolf420:
I think it would be nicer if `npcLoot.Get()` would have a new parameter that by default only returns non-disabled rules, that way modders don't have to add `IsDisabled` checks anywhere

BasicallyIAmFox:
> I think it would be nicer if `npcLoot.Get()` would have a new parameter that by default only returns non-disabled rules, that way modders don't have to add `IsDisabled` checks anywhere

Err, that's very likely to be breaking change then... But if that dotnetbuild pr gets merged this month then its fine i think. Its only breaking change pr so far i know?

Jacklz:
I noticed in the ExampleMod examples for disabling loot, you have to explicitly define the type of loot you're disabling with the class name. If you're trying to disable loot from other mods, you're most likely not going to know the class name or the fields within that class without decompiling that mod or looking at its source if its even open.

Also some items like loot bags can be a single IItemDropRule type with multiple items defined to drop but have no friendly way to remove or add a single item to/from that individual drop rule, especially if a mod is using a custom class inheriting the IItemDropRule interface.

Is there a way to make this easier?

BasicallyIAmFox:
> I noticed in the ExampleMod examples for disabling loot, you have to explicitly define the type of loot you're disabling with the class name. If you're trying to disable loot from other mods, you're most likely not going to know the class name or the fields within that class without decompiling that mod or looking at its source if its even open.
> 
> Also some items like loot bags can be a single IItemDropRule type with multiple items defined to drop but have no friendly way to remove or add a single item to/from that individual drop rule, especially if a mod is using a custom class inheriting the IItemDropRule interface.
> 
> Is there a way to make this easier?

sounds like something out of the scope of this pr. (#2759)

Jacklz:
> > I noticed in the ExampleMod examples for disabling loot, you have to explicitly define the type of loot you're disabling with the class name. If you're trying to disable loot from other mods, you're most likely not going to know the class name or the fields within that class without decompiling that mod or looking at its source if its even open.
> > Also some items like loot bags can be a single IItemDropRule type with multiple items defined to drop but have no friendly way to remove or add a single item to/from that individual drop rule, especially if a mod is using a custom class inheriting the IItemDropRule interface.
> > Is there a way to make this easier?
> 
> sounds like something out of the scope of this pr. (#2759)

Oh i forgot about this PR. Thanks for linking. I have some crude ideas for it that i may post about there when i get around to finishing my ideas.

BasicallyIAmFox:
@Chicken-Bones can you review this any time soon?

Chicken-Bones:
We've stalled out a bit on this because we need to decide whether disabled rules should be returned by various calls to the loot system aside from actual rolling. We need to come up with a clear set of guidelines for how to _use_ the loot system, and without these it's hard to know whether this feature will create more confusion.

When mods are searching for places to add their loot to existing pools, should disabled rules be returned? What if mods want to remove an item from a `OneFromOptionsDropRule`?

We really need to do a review of current modded usages of the loot system and identify all the times that destructive changes are made (removing/restructuring rules) and determine if we can support these use cases explicitly. Then we can reason about how to make simple changes (such adding or removing an item from a loot pool) compatible with more drastic ones.

Issue #3697, opened 2023-08-06
Easier Access to Modded World Data in the World Selection Screen


Pure-Epic:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

A new class could be made. It can retrieve .twld file data, allowing modders to check for tag values. This class is client-sided and runs in the world selection screen only as the modder can already load world data ingame with `ModSystem`.

### What does this proposal attempt to solve or improve?

Allowing easier access to the .twld file allows modders to create custom world icons and show custom difficulties in the world selection screen without making the screen freeze for 5 seconds or making a config that would crash the game because it can't read tags correctly.

### Which (other) solutions should be considered?

Alternatively, part of ModSystem could run in the world selection screen, just as how part of ModPlayer runs in the player selection screen (the latter made efficient custom player icons possible).

Issue #1704, opened 2021-08-05
[feature proposal] Header for twld data


Ishigh1:
### Is there an existing suggestion for this?

- [X] I have searched the existing issues

### Do you intend to personally contribute/program this feature?

No

### Description

Some mods need to get info from a world before it is even loaded, to change the world selection UI
My proposal is to provide ways to save info outside of the main twld save to provide quick way to access info

### What does this proposal attempt to solve or improve?

For now the only way to have such info is to load the whole twld of each world, this can easily take few seconds even with few low modded worlds

### Which (other) solutions should be considered?

I'd say either a separate file or a special part of the twld file that can be fetched separately without having to load everything, or few methods to directly use it so every mod doesn't have to load it

Chicken-Bones:
If you provide me some `.twld` files which take the said 2s to load, and a list of the tags you'd like to get, I'll look at both profiling the current deserializer, and making a 'quick access' one which can search for a specific tag path in a file and only deserialize that portion.

Ishigh1:
With 5 worlds with this data, it requires the 2 seconds to load the .twlds
The data I need is at `modData.[mod = AdvancedWorldGen].data.Options` (and even by commenting out this search the load is still slow)
[Fragile_Mesa_of_Care.zip](https://github.com/tModLoader/tModLoader/files/6949248/Fragile_Mesa_of_Care.zip)

Ishigh1:
Well it seems it went faster somehow, only 1/10th of a second for every twld

JavidPack:
This would facilitate #1416

Chicken-Bones:
Annoyingly this would require another file alongside `.twld`, maybe `.twldheader`? The NBT format provides no way to 'skip' entries you don't care about

absoluteAquarian:
A `.twldheader` file would be useful.  The bottleneck in loading the `.twld` file is having to load the `tiles.tileData` and `tiles.wallData` entries, which are easily in the millions of bytes.

If you only wanted to check the `modData` in the world, having to load `tiles.tileData` and `tiles.wallData` would be unnecessary.

Chicken-Bones:
I'm reluctant to add an extra file.

[...]

Issue #1416, opened 2021-04-01
Add indicator of changed mods for Worlds similar to characters


Monkeyatgunpoint:
### Description
there is an icon that says what mods are new and/or missing when selecting a character, why doesn't this apply to worlds too? I also think it should prompt you when you try to load up a world with mods missing and ask if you wish to continue, furthermore, it should be more obvious when there are new mods or missing mods than a tiny box with an exclamation mark, maybe make the entire row orange or red.

### What does this proposal attempt to solve or improve?
this helps prevent players from clicking the play button on their favourite world, only to find everything irreversibly broken (this happened to me twice in case you haven't guessed, took me hours to replace everything)

Solxanich:
A couple of items:
A) have you tried using Mod Helpers lock mods for world feature?  It may be helpful for you in interim that is 1.3
B) in terms of 'irreversibly broken' could you please be more specific? Our current plans for 1.4 include preserving walls and chests in addition to 1.3s tiles, so if something else is breaking, it would be great to know
C) I do like the idea of worlds showing if mods added/removed, so good suggestion independent

Hotmud9:
Why do you think world backups are a thing?

PR #3746, opened 2023-08-22
Add modded world header data

Chicken-Bones:
### What is the new feature?

Modded world data can be now be saved into a 'header' in the .twld file. The header can be read without deserializing the entire .twld file, and the modded data is accessible in the world select menu and during vanilla world loading.

- Use `ModSystem.SaveWorldHeader` to add data to the header.
- Use `WorldFileData.TryGetHeaderData<ModSystem>(out TagCompound data)` to access the data.
- Data for the current world is in `Main.ActiveWorldFileData`
-  `UIWorldListItem.Data` and `UICharacterListItem.Data` getters have been added for mods wishing to modify the selection menus.
- Use `WorldFileData.WorldGenModsRecorded` and `TryGetModVersionGeneratedWith(string mod, out Version modVersion)` to detect when the world was generated without your mod, or with an old version.

#### Implementation details
Unfortunately the NBT file format is not seekable so it is not possible to skip the reading of some tags in order to find a specific path. However, by relying on the implementation of `TagCompound` saving tags in insertion order, we are able to ensure the header tag is always the first one in the file, and initialize a reader at the correct position to only read the header.

Enables:
- #3697
- #1704
- #1416

A separate `.twldheader` file would reduce the reliance on potentially fragile insertion order saving (which in turn depends on the implementation of `Dictionary`), but at the cost of significant additional file maintenance and desync risk.

To save some custom data:
```cs
		public override void SaveWorldHeader(TagCompound tag) {
			tag["isSpecialModdedWorld"] = ...
		}
```

To read some custom data:
```cs
UIWorldListItem uiItem = ...
if (uiItem.Data.TryGetHeaderData<MyModSystem>(out TagCompound data) && data.GetBool("isSpecialModdedWorld"))
    ...
```
or 
```cs
if (Main.ActiveWorldFileData.TryGetHeaderData<ExampleModSystem>(out TagCompound data) && data.GetBool("isSpecialModdedWorld"))
    ...
```

To check which version of your mod was enabled during world-gen on entering a world
```cs
public override void OnEnterWorld() {
	if (Main.netMode == NetmodeID.SinglePlayer && Main.ActiveWorldFileData.WorldGenModsRecorded 
			&& Main.ActiveWorldFileData.TryGetModVersionGeneratedWith("ExampleMod", out Version modVersion)) {
		// check version or display a message about a biome missing from the world
		// consider saving a marker in the world file or header indicating that the message has been displayed
	}
}
```
### ExampleMod updates

Added `ExampleWorldHeaderSystem`

### DIFF
--- a/patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
+++ b/patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Terraria.ModLoader;
 using Terraria.ModLoader.IO;
 
@@ -6,6 +7,13 @@ namespace Terraria.IO;
 
 public partial class WorldFileData
 {
+	internal IList<string> usedMods; // log this?
+	internal string modPack;
+	/// <summary>
+	/// Contains the mods (and their version) that were enabled when this world was generated.<para/>
+	/// The feature tracking which mods were used to generate a world was added in v2023.8, so if <see cref="modsGeneratedWith"/> is <see langword="null"/>, this world was generated before then and can't be used to determine if a specific mod was enabled when the world was generated.
+	/// </summary>
+	public IList<(string modName, Version modVersion)> modsGeneratedWith; // TODO: Log this when loading world.
 	internal Dictionary<string, TagCompound> ModHeaders { get; set; } = new Dictionary<string, TagCompound>();
 
 	public bool TryGetHeaderData<T>(out TagCompound data) where T : ModSystem => TryGetHeaderData(ModContent.GetInstance<T>(), out data);
--- a/patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
+++ b/patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using Terraria.GameContent.UI;
 using Terraria.ModLoader;
 using Terraria.ModLoader.IO;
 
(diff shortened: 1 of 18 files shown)


Pure-Epic:
Finally, a dream feature came true!

patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using Terraria.ModLoader;
+using Terraria.ModLoader.IO;
+
+namespace Terraria.IO;
+
+public partial class WorldFileData
+{
+	internal IList<string> usedMods; // log this?
```
Chicken-Bones: Only if you also do so for players. Remove comment and make a new issue/PR if desired.

patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
```diff
@@ -0,0 +1,21 @@
...
+	internal string modPack;
+	/// <summary>
+	/// Contains the mods (and their version) that were enabled when this world was generated.<para/>
+	/// The feature tracking which mods were used to generate a world was added in v2023.8, so if <see cref="modsGeneratedWith"/> is <see langword="null"/>, this world was generated before then and can't be used to determine if a specific mod was enabled when the world was generated.
+	/// </summary>
+	public IList<(string modName, Version modVersion)> modsGeneratedWith; // TODO: Log this when loading world.
```
Chicken-Bones: This should be `Dictionary<string, Version> modVersionsDuringWorldGen`

Chicken-Bones: Should not be public. Expose a `TryGet` method. Consider adding a second, `bool WorldGenModsRecorded` property which checks the `modVersionsDuringWorldGen == null` so modders can distinguish 'unknown' case

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -631,4 +637,123 @@ internal static void EraseWorld(string path, bool cloudSave)
...
+	}
+
+	internal static List<TagCompound> SaveGeneratedWithMods()
+	{
+		if (Main.ActiveWorldFileData.modsGeneratedWith == null)
+			return null;
+		return Main.ActiveWorldFileData.modsGeneratedWith
+			.Select(x => new TagCompound {
+				["mod"] = x.modName,
+				["version"] = x.modVersion.ToString()
```
Chicken-Bones: As mentioned above, in the 'use dictionary' feedback. Consider adding a tag serializer for `Version` if we don't have one instead of manually doing to/from string

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -631,4 +637,123 @@ internal static void EraseWorld(string path, bool cloudSave)
...
+			string fullname = entry.Key;
+
+			if (ModContent.TryFind<ModSystem>(fullname, out var system)) // handle legacy renames
+				fullname = system.FullName;
+
+			data.ModHeaders[fullname] = (TagCompound)entry.Value;
+		}
+		LoadUsedMods(data, tag.GetList<string>("usedMods"));
+		LoadUsedModPack(data, tag.GetString("usedModPack"));
+		if(tag.ContainsKey("generatedWithMods")) // GetList will return an empty List instead of null. null and empty list have different meaning for this data.
```
Chicken-Bones: missing whitespace

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -631,4 +637,123 @@ internal static void EraseWorld(string path, bool cloudSave)
...
+	{
+		data.ModHeaders = new Dictionary<string, TagCompound>();
+		foreach (var entry in tag.GetCompound("modHeaders")) {
+			string fullname = entry.Key;
+
+			if (ModContent.TryFind<ModSystem>(fullname, out var system)) // handle legacy renames
+				fullname = system.FullName;
+
+			data.ModHeaders[fullname] = (TagCompound)entry.Value;
+		}
```
Chicken-Bones, line 726: Split this out to `LoadModHeaders` since the other load/save data methods are also split

patches/tModLoader/Terraria/IO/WorldFileData.tML.cs
JavidPack: #3768 will be that pr.

PR #3751, opened 2023-08-23
GlobalNPC.AddShops

ThomasThePencil:
### What is the new feature?
adds an `AddShops` hook to `GlobalNPC`, similar to the one introduced for `ModNPC` back when shops were originally overhauled

there currently isn't a way to **sensibly** add shops to existin' NPCs without detours. I am an avid fan of detour abuse and still didn't agree with that, and so we're here

note that this doesn't add an actual shop button to NPCs on its own, as vanilla's button not-really-a-system doesn't check for an NPC havin' an applicable shop. this is moreso a proper piece of infrastructure for people to add their own NPC shops to existin' NPCs in sensible locations; how they add the button to access said shop is an entirely separate ordeal

no, not really

```
public override void AddShops(int type)
{
	NPCShop nurseShop = new NPCShop(NPCID.Nurse);
	nurseShop.Add(ItemID.HealingPotion);
	nurseShop.Register();
}
```

### ExampleMod updates
none, currently. one can be added if desired, but I don't think it's necessary


Chicken-Bones:
I think the global call should go in `NPCLoader.AddShops(int type)` and supply the `int type` to the `GlobalNPC`. That way you can use the `Enumerate(int type)` call on the hook-list and `GlobalType.AppliesToEntity` will work with the hook. Effectively, copy the code in `NPCLoader.ModifyShop`

Please make sure the documentation is clear about when the hook is called (`Be aware that this hook is only called once per NPC during loading."`)

PR #3752, opened 2023-08-23
Porting ExampleCritter and related (+ a bugfix somehow)

Kogsey:
Please update on merge: #1403
Some things to do with the ExampleMod Example:
- Not exactly a port. ExampleCritter was "Lava Snail" in the old version which was a little close to "Magma Snail" from vanilla so here it's "Lava Frog" because I like frogs.
  - Original sprites left bc I didn't want to delete it
- The IL code from the last version has been updated to work for the frog statue instead of the snail
- A Couple vanilla functions were made public static for modder use in relation to critter cages

### Bugfix?
Related to #3353 Ran across it completely by accident. (For the record my fix did work the first time ;) )
Since `NPC.SetDefaults` is called twice if `NPC.CloneDefaults` is used (once for each type), there were two (sort of) issues when the the type passed to `CloneDefaults` is `Main.npcCatchable`
1. If the "true" NPC type is `Main.npcCatchable`, then the two lines `catchableNPCOriginallyFriendly = friendly;`, and `friendly = true;` run twice meaning friendly is always true. This is fixed with an easy check.
2. If the "true" NPC type is _not_ `Main.npcCatchable` then when the code to reset friendly is reached, it never runs. This is fixed by removing the check for `Main.npcCatchable` here as another check in the function will always cover it.

Shouldn't do


JavidPack:
Thanks for this. Just a note in case you do any other ExampleMod work in the future, we've decided to avoid `using static Terraria.ModLoader.ModContent;` in ExampleMod for 1.4. We also won't be using the "File Scoped Namespaces" feature of C# in ExampleMod itself since it is still quite a new concept and can cause confusion.

PR #3757, opened 2023-08-24
Added AddBackgroundTexture overload

BasicallyIAmFox:
### What is the bug?
Currently, `BackgroundTextureLoader.AddBackgroundTexture(Mod, string)` takes _full_ texture path.

For example, `BackgroundTextureLoader.AddBackgroundTexture(Mod, "Assets/Textures/Backgrounds/ExampleBiomeSurfaceClose")` would throw but if you add mod name to texture path then it will be just fine.

### How did you fix the bug?
Added `BackgroundTextureLoader.AddBackgroundTexture(string)` overload that takes full texture path and adjusted `BackgroundTextureLoader.AddBackgroundTexture(Mod, string)` to call it instead.

No.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/BackgroundLoaders.cs
+++ b/patches/tModLoader/Terraria/ModLoader/BackgroundLoaders.cs
@@ -38,6 +38,22 @@ public sealed class BackgroundTextureLoader : Loader
 	/// <summary> Safely attempts to output the slot/ID of the background texture with the given mod and path. </summary>
 	public static bool TryGetBackgroundSlot(Mod mod, string texture, out int slot) => TryGetBackgroundSlot($"{mod.Name}/{texture}", out slot);
 
+	/// <inheritdoc cref="AddBackgroundTexture(Mod, string)"/>
+	public static void AddBackgroundTexture(string texture)
+	{
+		if (texture == null)
+			throw new ArgumentNullException(nameof(texture));
+
+		ModContent.SplitName(texture, out string modName, out _);
+		if (!ModLoader.TryGetMod(modName, out Mod mod) || !mod.loading) {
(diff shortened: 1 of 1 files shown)


JavidPack:
This will break all mods calling the existing AddBackgroundTexture, right? Just making a note of this.

BasicallyIAmFox:
> This will break all mods calling the existing AddBackgroundTexture, right? Just making a note of this.

yeah, unfortunately. don't think there's any way to avoid that

Chicken-Bones:
> This will break all mods calling the existing AddBackgroundTexture, right? Just making a note of this.

You could always autodetect if a mod was using old variant by attempting `SplitName` and testing the `mod`

Chicken-Bones:
This API is consistent with other texture path APIs:
- `GoreLoader.AddGoreFromTexture`
- `Mod.AddBossHeadTexture`
- `Mod.AddNPCHeadTexture`
- `EquipLoader.AddEquipTexture`

A proposed change would have to include all of these endpoints as well as sufficient rationale for changing them.

BasicallyIAmFox:
> This API is consistent with other texture path APIs:
> 
>     * `GoreLoader.AddGoreFromTexture`
> 
>     * `Mod.AddBossHeadTexture`
> 
>     * `Mod.AddNPCHeadTexture`
> 
>     * `EquipLoader.AddEquipTexture`
> 
> 
> A proposed change would have to include all of these endpoints as well as sufficient rationale for changing them.

I can understand the ones in `Mod`, can't understand the others. It doesn't makes much sense to me that it takes a `Mod` argument, but uses it just for checking if its is loading or not. Many will be confused why their `BackgroundTextureLoader.AddBackgroundTexture(Mod, "Assets/Textures/Backgrounds/ExampleBiomeSurfaceClose")` throws an error.

PR #3760, opened 2023-08-25
Added ExampleCustomUseStyleWeapon, an example of a custom useStyle

Stardust9681:
Added `ExampleCustomUseStyleWeapon`, showcasing an implementation of `UseStyle` and `UseItemHitbox`, and `UseItemFrame` to implement a custom item use style. The actual code is in `ExampleCustomUseStyleGlobalItem` and the custom use style is applied to the vanilla `Cutlass` item as well.

An example of a custom use style have been a frequent request, this example should satisfy those who wish to implement custom weapon behavior in an item directly. Note that `ExampleCustomSwingSword` also exists and showcases another approach, using a held projectile to drive a custom animation. Both approaches have pros and cons and both are acceptable options to use.

This PR also fixes `ExampleCustomSwingSword` to work in reversed gravity. It also fixes the `Item.useStyle` docs which misleadingly suggested that setting `useStyle` to 0 would allow using `ModItem.UseStyle` to implement a custom use style. The value of 0 will actually cause use style code to be skipped altogether.

This PR also adds `ItemLoader.RegisterUseStyle` to allow mods to safely determine a unique custom use style ID. Prior to this it was up to mods to decide on a random number that no other mod happened to be using if they wanted to implement a custom use style through a `GlobalItem` and rely on the value of `Item.useStyle` directly.

Note that this particular example needs multiplayer syncing (see `ExampleCustomUseStylePlayer`), a simpler example that does not require multiplayer syncing can be achieved using this as a template by removing the ability to target a specific angle. This should be a simple matter of removing the `ExampleCustomUseStylePlayer` class and removing the mouse-derived angle calculations in `ExampleCustomUseStyleWeapon.UseStyle`.

ExampleCustomUseStyleWeapon:

https://github.com/user-attachments/assets/b233cbb9-00d6-421e-a541-b3d35fbdac2e

Cutlass (with melee hitbox visualization active):

https://github.com/user-attachments/assets/ea5c6021-27f8-47c9-8ba5-555cccb70771

# Discussion
- ~~Right now the use of `Item.useStyle` is in limbo. Some vanilla code checks `useStyle != 0`, such as the code to determine if an item should be used or not. There is a `if (item.useStyle > 0)` in `ContentSamples.GetItemGroup` that wouldn't work as expected if modders use negative values. Modders seeking completely custom use style can technically use whatever number they want. Even if they use an existing number, their logic can overwrite the changes anyway.~~ This PR changes the `if (item.useStyle > 0)` check to be `if (item.useStyle != 0)` to match the logic everywhere else.
- ~~I was thinking we probably want to support registering use style ids, for example if a `GlobalItem` were used to hold code for a new use style, many `ModItem` could use it without writing any boilerplate code (even crossmod). We could recommend `-1` for a `ModItem` that is completely custom maybe.~~ Implemented via `ItemLoader.RegisterUseStyle`
- ~~I remember seeing some mod that used arbitrary numbers, it's rare, but it is possible that eventually 2 mods will choose the same number, so it will be better to solve this directly.~~ See `ItemLoader.RegisterUseStyle`.

# Additional Changes
`Utils.CornerRectangle`, `Utils.BoundingRectangle`, and `Utils.Including` have been added to facilitate working with points and rectangles. See their documentation for more info.

# Porting Notes
* Replace magic numbers (made up numbers) for custom `Item.useStyle` values with the result of `ItemLoader.RegisterUseStyle`. This will allow greater compatibility and will avoid the potential of conflicting values between mods.


patches/tModLoader/Terraria/ModLoader/ItemLoader.cs
```diff
@@ -26,6 +26,7 @@ namespace Terraria.ModLoader;
 public static class ItemLoader
 {
 	public static int ItemCount { get; private set; } = ItemID.Count;
+	public static int UseStyleCount { get; private set; } = ItemUseStyleID.RaiseLamp + 1;
```
Chicken-Bones: I'd rather see a `Count` field added to `ItemUseStyleID`. It's more correct and we're more likely to pick it up in patch reviewer.

patches/tModLoader/Terraria/ID/ContentSamples.cs.patch
```diff
@@ -56,6 +56,15 @@
  					if (item.createTile == 178) {
  						orderInGroup = 5;
  					}
+@@ -627,7 +_,7 @@
+ 			}
+ 
+ 			orderInGroup = -item.rare;
+-			if (item.useStyle > 0)
++			if (item.useStyle != 0) // #3760 - Allow negative useStyle values since all the other useStyle checks allow it
```
Chicken-Bones, line 64: Upstreamed, and fixed some incoming 1.4.5 checks :)

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
```diff
@@ -0,0 +1,205 @@
...
+				useStylePlayer.SyncDirection(player);
+			}
+		}
+	}
+
+	// This class contains the actual custom UseStyle logic.
+	public class ExampleCustomUseStyleGlobalItem : GlobalItem
+	{
+		public static int ExampleCustomUseStyle;
+		public static Dictionary<int, int> HandOffsets; // Stores custom hold offsets if needed
```
Chicken-Bones: `CreateIntSet` please

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
```diff
@@ -0,0 +1,205 @@
...
+			float itemLength = (itemSize * player.GetAdjustedItemScale(item)).Length(); // We don't use Item.Size/width/height because those values are for the item hitbox when dropped.
+
+			// Calculate the handle and tip positions
+			Vector2 handlePosition = handDirection * (HandOffsets.TryGetValue(item.type, out int offset) ? new Vector2(offset) : Vector2.Zero) + player.MountedCenter;
+			Vector2 tipPosition = handlePosition + handDirection * itemLength;
+
+			// Now we use those values to create the item hitbox
+			(int X1, int X2) = Order(handlePosition.X, tipPosition.X);
+			(int Y1, int Y2) = Order(handlePosition.Y, tipPosition.Y);
+			hitbox = new Rectangle(X1, Y1, X2 - X1, Y2 - Y1);
```
Chicken-Bones: This is kinda sad, would be nice if we had `Utils.CornerRect`, or `BoundingRect(p1, p2)`

In 1.4.5 I do it like this:
```cs
			var r = Utils.CenteredRectangle(points[0], Vector2.Zero);
			foreach (var point in points)
				r = r.Including(point.ToPoint());

			BoundingRect = r;
```
			
Where `Including` is an extension method in `Utils`
```cs
		public static Rectangle Including(this Rectangle rect, Point point)
		{
			int l = Math.Min(rect.Left, point.X);
			int r = Math.Max(rect.Right, point.X);
			int t = Math.Min(rect.Top, point.Y);
			int b = Math.Max(rect.Bottom, point.Y);
			return new Rectangle(l, t, r - l, b - t);
		}
```

Chicken-Bones (approved):

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
JavidPack: Added.

patches/tModLoader/Terraria/Utils.TML.cs
```diff
@@ -275,4 +275,50 @@ private static void AddArgToDictionary(string text, ref string text2, ref Dictio
...
+		if (points.Length == 0)
+			return new Rectangle();
+		var rectangle = new Rectangle(points[0].X, points[0].Y, 0, 0);
+		foreach (var point in points)
+			rectangle = rectangle.Including(point);
+		return rectangle;
+	}
+
+	/// <inheritdoc cref="BoundingRectangle(Point[])"/>
+	public static Rectangle BoundingRectangle(params Vector2[] vectors) => BoundingRectangle(vectors.Select(ToPoint).ToArray());
```
Chicken-Bones: Since we don't have a value linq system imported, I recommend duplicating the method above

patches/tModLoader/Terraria/Utils.TML.cs
```diff
@@ -275,4 +275,50 @@ private static void AddArgToDictionary(string text, ref string text2, ref Dictio
...
+		return new Rectangle(left, top, width, height);
+	}
+
+	/// <inheritdoc cref="CornerRectangle(Point, Point)"/>
+	public static Rectangle CornerRectangle(Vector2 pointA, Vector2 pointB) => CornerRectangle(pointA.ToPoint(), pointB.ToPoint());
+
+	/// <summary>
+	/// Creates a <see cref="Rectangle"/> containing all of the provided points.
+	/// </summary>
+	public static Rectangle BoundingRectangle(params Point[] points)
```
Chicken-Bones: I'm not happy making this `params` until we have `params Span`

patches/tModLoader/Terraria/Utils.TML.cs
```diff
@@ -275,4 +275,50 @@ private static void AddArgToDictionary(string text, ref string text2, ref Dictio
...
+	{
+		if (points.Length == 0)
+			return new Rectangle();
+		var rectangle = new Rectangle(points[0].X, points[0].Y, 0, 0);
+		foreach (var point in points)
```
Chicken-Bones: A regular for-i loop could start at 1

Chicken-Bones (changes requested):

Chicken-Bones (approved):

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
```diff
@@ -0,0 +1,202 @@
...
+			Item.width = 58;
+			Item.height = 58;
+			Item.knockBack = 2f;
+			Item.value = Item.sellPrice(gold: 1);
+			Item.rare = ItemRarityID.Green;
+			Item.autoReuse = true;
+			Item.UseSound = SoundID.Item1;
+
+			// This controls how far out the weapon should be held from the hand. This weapon uses 0 (so we don't actually need to set it) but the logic in ExampleCustomUseStyleGlobalItem works for other values as well. We can set it here or in the CreateIntSet method, both work.
+			//ExampleCustomUseStyleItemSets.HandOffsets[Type] = 0;
```
direwolf420, line 39: Shouldn't this be in SetStaticDefaults instead?

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
```diff
@@ -0,0 +1,202 @@
...
+		// Checking ItemID.Cutlass is necessary here because SetDefaults below won't run for Cutlass to change its Item.useStyle otherwise.
+		public override bool AppliesToEntity(Item item, bool lateInstantiation) {
+			return lateInstantiation && (item.type == ItemID.Cutlass || item.useStyle == ExampleCustomUseStyle);
+		}
+
+		public override void SetDefaults(Item item) {
+			// Cutlass will now use out custom use style, making it swing up instead of the normal swing.
+			if (item.type == ItemID.Cutlass) {
+				item.useStyle = ExampleCustomUseStyle;
+				//ExampleCustomUseStyleItemSets.HandOffsets[item.type] = -6; // Alternate approach to setting HandOffsets
```
direwolf420, line 109: Same here

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
JavidPack: K.

ExampleMod/Content/Items/Weapons/ExampleCustomUseStyleWeapon.cs
JavidPack: K.

PR #3767, opened 2023-08-27
Rewrite CheckIfPublishedForThisBrowserVersion Logic

Solxanich:
Followup to cleanup logic gaps and simplify code surrounding Checking if the the mod was published for this browser version.

Added commentary on the three cases to consider.
Added non-visible tag 1.4.4-transitive to be used for internal logic
Simplified logic

SHould be backported to 1.4.3-Legacy when approved

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -571,34 +571,31 @@ internal static class ModOrganizer
 		return recommendedTmod.file;
 	}
 
+	/// <summary>
+	/// Case 1:
+	///		tmlVersion = 1.4.4
+	///		modVersion = 2023.1 => 1.4.4 transition versions prior to major breaking changes completion
+	///	Returns false as 2023.1 isn't considered 1.4.4 ready. Supplies modBrowserVersion as 1.4.3 as achievable version
+	///	Case 2:
+	///		tmlVersion = 1.4.3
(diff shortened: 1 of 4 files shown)


Solxanich:
Discord Feedback:

Returning "1.4.4-transitive" from GetBrowserVersionNumber is good, but needs a detailed comment about reasoning. 1.4.4-transitive is a 'dead' browser version, mods built against it won't work on any of the actual LTS/browser versions, and this distinguishes it as such.
a. Another option would be "unsupported" but that's less informative
b. Perhaps 1.4.4-transition"? Lowercase preferred, fewer gotchas that way.
The code in CheckIfPublishedForThisBrowserVersion which returns "1.4.3" instead of "1.4.4-Transitive" 
a. For readability this would be better as if (modBrowserVersion == "1.4.4-Transitive"), makes it clear what case we're catching.
b. Please explain the reasoning for blocking this version in a comment. Best I can think of is "1.4.4-transitive doesn't exist, but if the user goes back to 1.4.3, there will probably be an older version of the mod published for 1.4.3 which they can load". This is non-obvious (took me 4 minutes to reason out)
Perhaps out string modBrowserVersion should be renamed to recommendedModBrowserVersion, to make it clear that it's not the actual version the mod is published against, just one that we're going to display in the UI. 
a. I also would be in favor of moving the entire method to UIModItem since that's the only usage of it, and this special case seems very closely tied to "directing a user to perform an action by presenting a visual indicator" which isn't 'backend' like ModOrganizer is.

PR #3770, opened 2023-08-28
Fix #3692, Also serves as an example for using Obsolete on hooks

### LINKED ISSUE #3692: Rename ModProjectile.Kill hook to OnKill

No

Mod capability as a Modder

### Description

Rename `ModProjectile` `Kill` hook to `OnKill`.

The current name often can be misused as "kill this projectile", while that's not what's the method intentional use.
I often see people accidentally using it instead of `Projectile.Kill()`. In fact even myself almost used wrong `Kill` method.

_No response_

JavidPack:
I made this into a PR so that it could be consulted later as an example of how we Obsolete hooks. 

Currently, there are just notes mentioning when to remove the obsolete hooks `// Remove in 2023_10`, but we could find some way to cause a build error at some later date, I imagine. 

We also want to standardize how long a breaking change like this will live in the source code. Technically we could just leave them in, but in the past we have sought to remove them eventually. 

~~One issue with leaving the obsolete method in the code is that tModPorter won't port the method. This could probably be fixed in tModPorter. The commented out tModPorter code currently won't fix existing mods if enabled.~~

Might need to update <https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#be-aware-of-breaking-changes> as well if we decide on any new approaches.


patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
```diff
@@ -364,9 +364,11 @@ public static bool PreKill(Projectile projectile, int timeLeft)
 		return result;
 	}
 
+	[Obsolete]
 	private static HookList HookKill = AddHook<Action<Projectile, int>>(g => g.Kill);
 
-	public static void Kill(Projectile projectile, int timeLeft)
+	[Obsolete("Renamed to OnKill")]
+	public static void Kill_Obsolete(Projectile projectile, int timeLeft)
```
Chicken-Bones, line 371: We want to call `Kill_Obsolete` from `OnKill` here, keeps things nice and contained. Note that in some cases it may be necessary to interleave the mod and globals calls from the old and new hooks to preserve behaviour.

JavidPack: K, that has been moved.

Chicken-Bones (approved):

Issue #3739, opened 2023-08-19
Vanilla "UseMana" tooltip does not account for ModifyManaCost hooks


stormytuna:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently, if a modded item has its mana cost changed through one of the `ModifyManaCost` hooks, that change is not reflected in its tooltip by default. Though this is easily fixable with a few lines in `ModifyTooltips`, I believe it may be an oversight in tML

The fix I'm proposing, is just slightly changing how the mana cost tooltip is created. Currently, it checks the item has mana and isn't space gun + meteor armour:
```cs
if (item.mana > 0 && (item.type != 127 || !player[myPlayer].spaceGun)) {
    toolTipLine[numLines] = Language.GetTextValue("CommonItemTooltip.UsesMana", (int)((float)item.mana * player[myPlayer].manaCost));
    toolTipNames[numLines] = "UseMana";
    numLines++;
}
```

Vanilla sets a precedent here - items that normally cost mana, but currently cost 0 mana due to some other effect do not have the `UseMana` tooltip displayed. I think we should follow it, doing a similar check:
```cs
int manaCost = player[myPlayer].GetManaCost(item);
if (manaCost > 0) {
    toolTipLine[numLines] = Language.GetTextValue("CommonItemTooltip.UsesMana", manaCost);
    toolTipNames[numLines] = "UseMana";
    numLines++;
}
```

### What does this proposal attempt to solve or improve?

It better conveys the true mana cost of an item to the player

### Which (other) solutions should be considered?

If we don't want the line to be removed when mana cost is zero, we could just inline the `GetManaCost` call
```cs
if (item.mana > 0 && (item.type != 127 || !player[myPlayer].spaceGun)) {
    /*
    toolTipLine[numLines] = Language.GetTextValue("CommonItemTooltip.UsesMana", (int)((float)item.mana * player[myPlayer].manaCost));
    */
    toolTipLine[numLines] = Language.GetTextValue("CommonItemTooltip.UsesMana", player[myPlayer].GetManaCost(item)));
    toolTipNames[numLines] = "UseMana";
    numLines++;
}
```

JavidPack:
Hm, space gun is a well known effect, but with modded effects reducing mana costs in a variety of ways, I'm wondering if not showing the mana cost when reduced to 0 would be interpreted as an error by users rather than a feature. 

[...]

PR #3774, opened 2023-08-30
Fix "UseMana" tooltip

stormytuna:
Implements #3739 

### What is the new feature?

Fixes vanilla "UseMana" tooltips not taking into account the various `ModifyManaCost` hooks.

This also changes Space Gun to say "Uses 0 mana" instead of lacking a tooltip.

See issue

There is more to discuss, we didn't really arrive at a consensus.


JavidPack:
Thanks, looks clean.

JavidPack:
I went ahead and decided it is good to merge as is. It seems a little out of scope to show how mana is adjusted here, since we don't do that in other places.

PR #3783, opened 2023-09-01
ExampleStaff 1.4.4 port.

Vaema:
Ported ExampleStaff and its projectile ("SparklingBall").
I also deleted the old ExampleStaff and SparklingBall in the Old directory.


JavidPack:
Thanks.

I'm not sure why you decided to change SparklingBall, you moved code around and broke the bounce logic. You also undid all the code style changes we made. I reverted these.

PR #3784, opened 2023-09-02
Add additional handling for Corrupt Install Scenarios

Solxanich:


patches/tModLoader/Terraria/release_extras/LaunchUtils/InstallNetFramework.sh
```diff
@@ -30,6 +30,12 @@ if [ -d "$dotnet_dir" ]; then
 			rm -rf "$old_version"
 		fi
 	done
+
+	if [[ ! -f "$LaunchLogs/client.log" && ! -f "$LaunchLogs/server.log" ]]; then
+		echo "Last Run Attempt Failed to Start tModLoader. Deleting dotnet_dir and resetting"
+		rm -rf "$dotnet_dir"
+		mkdir "$dotnet_dir"
+	fi
```
Chicken-Bones: This belongs in `ScriptCaller.sh` then there's no need to `export LaunchLogs`. Just do the `rm -rf` and let the `InstallNetFramework` call do the `mkdir`

JavidPack:
The MarkSteamTerrariaInstallCorrupt method isn't being called, is that intentional, for later?

Solxanich:
@JavidPack  - yes, left for future. 
Intent is for use with anything that may require us to trigger a vanilla update.
Say for example asset issues in vanilla files.
However, I haven't a clue where to inject that, so left the one end in for another to add

PR #3791, opened 2023-09-05
Pre/PostDrawHeldItem hooks

stormytuna:
supercedes #2768 

### What is the new feature?

`PreDrawHeldItem` and `PostDrawHeldItem` hooks on `ModItem` and `GlobalItem` to allow easier custom drawing of held and used items

It's currently a bit annoying to do something like this, requiring either a custom `PlayerDrawLayer` with some duplicated code from vanilla or a held projectile, adding more complexity than really necessary for most simple effects

Not that I can think of, but I am open to suggestions

See `ExampleCustomDrawnSword` 

### ExampleMod updates

Adds `ExampleCustomDrawnSword` to show off a few things that are possible with these new hooks


ghost:
Does this need anything else? Or still being looked at?

stormytuna:
It was rejected in favour of a declarative approach that hasn't been discussed, I just left it open for a maintainer to reject it officially

Chicken-Bones:
> It was rejected in favour of a declarative approach that hasn't been discussed, I just left it open for a maintainer to reject it officially

Can you link the discussion?

Issue #3775, opened 2023-08-30
Hotbar swap buffering is different from Vanilla due to use time changes


jjohnsnaill:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

In vanilla, you can press one of the hotbar keys while already using an item to buffer a swap when you stop using it. Funnily enough, this works because of the longstanding bug where all items are a frame slower.

But because tML fixes that bug, you now have to manually stop using the item for at least a frame for the swap to go through.

### Steps to reproduce

Hold down the use key, then press one of the hotbar keys while the item is in use

### Expected Behavior

Item swaps right when it's done being used

### Actual Behavior

Item doesn't swap until you manually let go of the use key

### Reproduction frequency

100%

JavidPack:
Related to #2478

Issue #2478, opened 2022-06-02
You can't switch to another weapon while using whips with auto swing


Cyrillya:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

It should be able to switch your weapon by holding left while using the whip with auto swing accessories, which is doable in vanilla.
I don't want this feature to disappear though if you consider it as a bug and fix it. Since this has become a special way to use the whip to defeat bosses
For example:
You can attack the enemies using Durendal once, which grants you the Durendal's Blessing buff that increases your melee speed for 3s. Then you can switch to Dark Harvest to gain the  Harvest Time buff that also increases your melee speed. Before these buffs run out, you can switch to whips such as Kaleidoscope to use whips at a insanely fast speed to deal a lot of damage.
It has also became a tip on wiki: [Dark Harvest](https://terraria.wiki.gg/wiki/Dark_Harvest)
![image](https://user-images.githubusercontent.com/35227653/171649934-de4e7400-7ed3-4dab-b5b1-2fca11702fe1.png)

### Steps to reproduce

1. Prepare two whips in the hotbar, here I suppose that they are in the 1st slot and the 2nd slot
2. Equip an accessory like Feral Claws that enables auto swing for whips
3. Select the first slot, holding left
4. Click "2" in your keyboard. In vanilla, it switches to the 2nd slot when you use the whip once more. But in tModLoader, it just keeps using the whip in the 1st slot and only switches to the 2nd slot when you release the left.

Chicken-Bones:
It's definitely a bug. When using feral claws on an item which doesn't have `autoReuse` without it, vanilla gives you one frame where you're 'not using the weapon' allowing you to swap. 

We could bring this back uniformly for _all_ weapons as an intended mechanic. Otherwise you can probably bring it back with a mod.

Chicken-Bones:
After some discussion, we'd like to allow hotbar swap on _all_ weapons (not just ones that get `autoReuse` with feral claws) but it'll take some time to figure out where to do it, and it's far from the top of priority list.

Jofairden:
Related
https://github.com/tModLoader/tModLoader/issues/1770

Cyrillya:
Simply wrapping these checks in `Player.SmartSelectLookup` to `itemAnimation == 1` will help reimplement the feature (for all items)
[...]

PR #3804, opened 2023-09-12
Enable hotbar key item swap and smart select while auto-reusing items

Chicken-Bones:
### What is the bug?
- #3775
- #2478

### How did you fix the bug?
Smart select, and hotbar swap buffering was fixed by moving `SmartSelectLookup` call into `ItemCheck` making it run after `itemAnimation--` decrements to 0, allowing the `selectedItem` to be changed and used immediately while holding down the attack button.

The `SmartSelectLookup` checks could be changed to `ItemAnimationEndingOrEnded` and the original position of the code restored. This would temporarily produce an invalid state where `selectedItem` has changed and is now in its last frame of animation (`itemAnimation == 1`) during movement updating hooks.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -3761,6 +3761,17 @@
  					}
  				}
  				else {
+@@ -19467,8 +_,10 @@
+ 				}
+ 			}
+ 
++			/* Moved into ItemCheck
+ 			if (selectedItem != 58)
+ 				SmartSelectLookup();
++			*/
+ 
+ 			if (stoned != lastStoned) {
+ 				if (whoAmI == Main.myPlayer && stoned) {
 @@ -19566,11 +_,14 @@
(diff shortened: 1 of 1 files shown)


ItsNuked:
Hello. 
I'm facing this issue at the moment, finding it very annoying to swtich weapons mid boss fights since you have to stand there not attacking for a solid 1.5 seconds before switching weapons, or itkeeps using the old one.
If i understood correctly, this is what you have fixed here?
If so, may I be so rude as to ask how one can implement this fix into their game? Im fairly new to github and such, I see that you modified the player.cs.patch file, and I'm wondering how I could use this fix myself? 
Thank you very much for the read, and I'm sorry if this is at all a bother ^^

JavidPack:
The current stable is v2023.11.3.3, at this point I'm pretty sure this should be all fixed in current tmod. If you are still having issues, verify the issue with 0 mods enabled and open a new issue or come to discord for help

PR #3809, opened 2023-09-15
Add parameterless constructor to NPCBestiaryDrawModifiers

Destructor-Ben:
### What is the new feature?
Added a parameterless contructor to `NPCBestiaryDrawModifiers` and obsoleted the old one.

The only reason the one with an unused parameter exists is because vanilla uses an old version of .NET that doesn't support parameterless struct constructors.

Adding a tModPorter rule (don't know if I can do it with this).
Using a partial class (overkill but less patches).

`new NPCID.Sets.NPCBestiaryDrawModifiers();`

### ExampleMod updates

Changed all of the `NPCBestiaryDrawModifiers()` to use the new constructor.

### DIFF
--- a/patches/tModLoader/Terraria/ID/NPCID.cs.patch
+++ b/patches/tModLoader/Terraria/ID/NPCID.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/ID/NPCID.cs
 +++ src/tModLoader/Terraria/ID/NPCID.cs
-@@ -3,29 +_,109 @@
+@@ -3,29 +_,110 @@
  using Microsoft.Xna.Framework;
  using ReLogic.Reflection;
  using Terraria.DataStructures;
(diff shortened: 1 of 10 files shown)


patches/tModLoader/Terraria/ID/NPCID.cs.patch
```diff
@@ -109,9 +109,37 @@
...
  				Position = default(Vector2);
+@@ -42,6 +_,26 @@
+ 				PortraitPositionYOverride = null;
+ 				CustomTexturePath = null;
+ 			}
++
++			/// <summary>
++			/// Creates a new <see cref="NPCBestiaryDrawModifiers"/> with default values.
++			/// </summary>
++			public NPCBestiaryDrawModifiers()
```
Chicken-Bones: Please make one of these constructors call the other

patches/tModLoader/Terraria/ID/NPCID.cs.patch
```diff
@@ -109,9 +109,37 @@
 +			/// Creates a new <see cref="NPCBestiaryDrawModifiers"/> with default values.
 +			/// </summary>
 +			/// <param name="seriouslyWhyCantStructsHaveParameterlessConstructors">Unused.</param>
++			[Obsolete("Use the parameterless constructor instead")]// TODO: make internal at some point
```
Chicken-Bones: Not really any need to make internal at some point

Chicken-Bones (changes requested):
Consider adding tModPorter for it if it's not too much effort?

PR #3810, opened 2023-09-16
Config List Rework (Config Rework Part 1)

Destructor-Ben:
Split off from Config UI Rework: https://github.com/tModLoader/tModLoader/pull/3556

### What is the new feature?
Completely reworked config list UI
- Can now be accessed in the main menu as well as in game
- Has a list for mods and a list for configs
- Selecting a mod will show all it's available configs
- Clicking a config will open it
- Made the config button on `UIModItem` in the mods menu to open the config list with the mod selected
- Added a button to open the mod config list from the mods menu
- Each config has a red or blue dot indicating whether it is client or server side
  - Red is server, blue is client
  - Hovering over the dot will also say if it is a server side or client side config

<details>
<summary><h3>Before</h3></summary>
<br>

Screenshot_20230617_011615

</details>

<details>
<summary><h3>After</h3></summary>
<br>

image

Screenshot_20230726_083406

Screenshot_20230617_120026

</details>

Currently the UI for the configs is ugly and confusing and it has been stated many times on the tML discord server that it should be changed. It is also difficult for users to find configs. In the main menu, the only ways to access configs was through the config button (but that only took you to the first one a mod registered), and the rest through the difficult to notice next and previous buttons. In game, there was the config list, which is how I usually changed configs, but you couldn't change the values that would require a reload, so you had to deal with the main menus config UI. In many cases I have never even realized that a mod has multiple configs because of this.

- It's UI so yes
- A filter for the mods menu that only shows mods that have a config could be implemented like https://github.com/tModLoader/tModLoader/issues/2762 suggested, but it is mostly irrelevant now with the new UI

N/A

### ExampleMod updates
N/A

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
@@ -97,6 +97,7 @@
 		"AskForCommand": "Gib einen Befehl ein: ",
 
 		//Mod Config
+		// "ModConfigs": "Configs",
 		"ModConfigFilterOptions": "Filtereinstellungen",
 		"ModConfigNotification": "Benachrichtigung: ",
 		"ModConfigModConfig": "Modkonfiguration",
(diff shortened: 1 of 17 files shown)


Cyrillya:
Suggestion: Hide the scroll bar when not needed

JavidPack:
Looks good. I don't think the scroll bars need to hide, the game doesn't do that with the resource pack menu, so it's consistent with that.

Chicken-Bones:
> Suggestion: Hide the scroll bar when not needed

Feel free to contribute a scroll-bar hiding upgrade which works across multiple modded UIs perhaps

Issue #2897, opened 2022-09-04
Support for publish a newer version on stable AND discard preview version


Solxanich:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

It often happens that some modders push to preview an older version, than find they want to push a new version to stable.

The problem with that is the versioning for tmodloader & the mods does not facilitate.

A recommended resolution is to have something akin to:
"Upon publishing to Stable, check if there's a pending mod version on Preview.  In that case, pop up a dialog...
Title: Warning!  Are you sure about that?
You have a pending version of this mod in Preview.  Would you like to remove it in favor of this release to Stable?  If not, this release will be OVERWRITTEN by the older Preview version at the end of the month.

Replace (default button)              Keep Both Releases
                                                            (red button)"
                                                            
Combined with having a method to delete any mod copies that are higher than the publishing version, similar to the 9999 check today.

### What does this proposal attempt to solve or improve?

Decrease issues with updating mods as pertains to mod browser

Chicken-Bones:
Closing this as unsupported. Once a preview version is published, modders should maintain it. I see no reasonable way to 'un-publish' it.

We discussed `1.0.0.1 for tML 2023.06` being loaded instead of `1.0.0 for tML 2023.07` when the user is on `2023.07` or higher but ultimately decided that this would lead to confusion and accidental breakage. See the full discussion at https://discord.com/channels/103110554649894912/1153604070267953182

PR #3822, opened 2023-09-19
Improve mod version selection when multiple versions are present

Chicken-Bones:
### What is the new feature?
Changes to the priority of `.tmod` selection in `ModOrganizer` when multiple versions of the same mod exist, as well as more detailed logging about the selection process. A stricter publishing requirement shall ensure that mod version can never decrease on tML rollover.

Old search order:
- Active mod pack 'frozen' mod versions folder
- Local `Mods` folder (which may have been synced from a server)
- Steam workshop folder
  - tML versions in order of newest to oldest

New search order:
- Versions required by the server the user is connecting to (prioritized regardless of location, if present).
- Active mod pack 'frozen' mod versions folder
- Highest _mod_ (not tML) version
- Highest tML version
- Local `Mods` folder (when the hash of the mod differs from an identical version in the workshop folder)
- Steam workshop folder

Requirements (apply in order unless they would exclude all remaining mods):
- Mod must have been built on a tML version in the same 'browser group' as the one loading (1.4.3/1.4.4/etc)
- Mod must not be from a transitional tML version
- Mod must have been built on a tML major.minor version older than or equal to the current one.
- Mod must be playable on preview if the current tML version is preview

Other refactors and cleanups included. Note that currently, the filtering and sorting process will never eliminate all versions of a mod, even if none of them are loadable on the current tML version. In such cases, the ordering process continues as normal and the filter is ignored.

Sample logs for different tML versions. Note that I have a 'frozen' copy of `ExampleMod 9999.0 for tML 9999.0` in the 'active modpack'.
[mod selection tML 9999.0 (dev).log](https://github.com/tModLoader/tModLoader/files/12669666/mod.selection.tML.9999.0.dev.log)
[mod selection tML 2023.06.25.36 (stable).log](https://github.com/tModLoader/tModLoader/files/12669668/mod.selection.tML.2023.06.25.36.stable.log)
[mod selection tML 2022.9.47.87 (1.4.3-legacy).log](https://github.com/tModLoader/tModLoader/files/12669667/mod.selection.tML.2022.9.47.87.1.4.3-legacy.log)

### How does this affect modders?
- Mod version is more important than tML version, this provides clarity and reliability, the newest mod version within the user's tML range will always be the one loaded. 
  - As a result, modders will no longer be able to publish a higher version on stable than on preview. An error UI will recommend that the modder publish a newer preview version as a workaround, and leave the last version number for stable patches in future.
- If a mod is distributed to a tester and placed in the local mods folder, and then a new version is published on the workshop, the workshop version will be loaded instead, and the local mod will be 'hidden'.
- Dev versions of ExampleMod are now version 9999.0 as a consequence, to ensure they always take priority. 
- Mods built against a development tML version will be automatically ignored when on a stable/preview version of tML, despite being in the local mods folder.

### How does this affect users?
- No more dreaded 'pink mod' issue because a copy was downloaded from a server, and now they no longer receive workshop updates.
  - The only way to delete a local copy of a mod when the workshop version is newer will be to go into the mods folder directly.
- On connecting to a server running old mod versions, a reload may be triggered to select them from the local mods folder.
  - If the user reloads again, they will revert to the standard logic, running the newest versions they have available.
- The versions of mods shown in the Mods UI are not the versions which the user has active, they are the versions which would be selected if the user reloads. When a user connects to a server which selects a specific old version of a mod from the workshop folder, and then returns to the mods UI, there is no indication of this mismatch.
  - This could also be an issue if the user goes to load a single-player world after joining a server, and mod versions regress
  - **We should consider** reloading when the user clicks the "Single Player" button if the current mod set is not 'up to date'.
  - **We should consider** some way to show this difference in the UI
  - **We should consider** informing the user when connecting to a server will change their loaded mod list (or download new versions).

Fixes the following issues:
- An older version of a mod in the local mods folder (potentially synced from the server) is loaded instead of a newer version in the steam workshop folder, hiding updates and confusing users.
- Confusing rollover behavior, possibly leading to requests like #2897

Server mods could be synced to a different directory (a decent idea for the future). 

A significant amount of workshop, mod browser, modpack and mod building code currently depends on `ModOrganizer.FindMods`. Some breakage may be incurred due to tight coupling between these systems and load order, and will have to be resolved as issues are identified. Long-term, refactoring is required to reduce unnecessary dependencies and convert queries into more robust forms.


patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -445,4 +449,50 @@ public static void AddSignalTraps()
...
+
+		if(ConfigManager.AnyModNeedsReloadCheckOnly(normalModsToLoad, reloadRequiredExplanationEntries)) {
+			// If reload is required, show message. Back action should leave current ModConfig instances unchanged 
+			needsReload = true;
+		}
+		else if (!needsReload) {
+			// Otherwise, config changes take effect immediately and the player select menu will be shown
+			ConfigManager.LoadAll(); // Makes sure MP configs are cleared.
+			ConfigManager.OnChangedAll();
+		}
```
Chicken-Bones: Better to separate this from the above `if`, since it's more logically distinct. Use an early `return`, change the method name to something which doesn't explicitly mention the reload menu. As is, it implies that nothing will be changed if the reload menu isn't done. Perhaps `PrepareLoadedModsAndConfigsForSingleplayer`

```suggestion

		if (!needsReload) {
			// Otherwise, config changes take effect immediately and the player select menu will be shown
			ConfigManager.LoadAll(); // Makes sure MP configs are cleared.
			ConfigManager.OnChangedAll();
			return;
		}
```

patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -445,4 +449,50 @@ public static void AddSignalTraps()
...
+				continueButtonText,
+				() => {
+					ModLoader.ModLoader.OnSuccessfulLoad += () => { Main.menuMode = 1; };
+					ModLoader.ModLoader.Reload();
+				},
+				"Back",
+				() => {
+					// Do nothing, logic will to back to main menu
+				},
+				reloadRequiredExplanationEntries.OrderBy(x => x.typeOrder).ThenBy(x => x.mod).ToList()
```
Chicken-Bones: Common code here with the other usage of `Show`, consider putting the sorting in `Show`.

patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -445,4 +449,50 @@ public static void AddSignalTraps()
...
+			ConfigManager.OnChangedAll();
+		}
+
+		if (needsReload) {
+			string continueButtonText = "Reload and Continue";
+			Interface.serverModsDifferMessage.Show($"Due to the following mod version or config differences still present from multiplayer, mods will reload if you enter single player. Press \"{continueButtonText}\" to enter single player.",
+				0, // back to main menu
+				null,
+				continueButtonText,
+				() => {
```
Chicken-Bones: prefer named arguments for all of these for readability. Also means you don't need to specify `gotoMenu`

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -263,6 +263,28 @@ internal static bool ModNeedsReload(Mod mod)
 		return false;
 	}
 
+	internal static bool AnyModNeedsReloadCheckOnly(IEnumerable<LocalMod> normalModsToLoad, List<ReloadRequiredExplanation> reloadRequiredExplanationEntries)
+	{
+		bool result = false;
+		foreach (var mod in ModLoader.Mods) {
+			if (Configs.ContainsKey(mod)) {
```
Chicken-Bones: invert if

patches/tModLoader/Terraria/ModLoader/Config/ConfigManager.cs
```diff
@@ -263,6 +263,28 @@ internal static bool ModNeedsReload(Mod mod)
 		return false;
 	}
 
+	internal static bool AnyModNeedsReloadCheckOnly(IEnumerable<LocalMod> normalModsToLoad, List<ReloadRequiredExplanation> reloadRequiredExplanationEntries)
```
Chicken-Bones: Might be better to use an `out List<Mod>` or return an `IEnumerable<Mod>`. There's a lot of similar code here and in the only caller, so it's better to keep similar code together (the creation of explanations for eg)

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -227,12 +229,56 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
+					reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(3, header.name, matching, $"[c/98FB98:Enable]")); 
+				}
+
 				continue;
 			}
 
-			if (downloadModsFromServers && (header.signed || !onlyDownloadSignedMods))
+			if (downloadModsFromServers && (header.signed || !onlyDownloadSignedMods)) {
 				downloadQueue.Enqueue(header);
+				if (modFiles.FirstOrDefault(mod => mod.Name == header.name && mod.Version == header.version) is LocalMod localMod) {
```
Chicken-Bones: Didn't have time to parse this entire if statement, but it's definitely quite a bit of work to parse. Perhaps there's a way to factor or flatten it so the conditions are more clear

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -227,12 +229,56 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
+						else {
+							// We have the mod, but not the correct version.
+							reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(1, header.name, localModMatchingNameOnly, $"[c/00BFFF:Download] v{header.version} from server"));
+						}
+					}
+
+				}
+			}
 			else
 				blockedList.Add(header);
```
Chicken-Bones, line 245: scary, a tiny `else` with no `{}` at the end of a big `if`. Perhaps extracting the code which makes the `ReloadRequiredExplanation` for a mod to download is the way to go.

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -227,12 +229,56 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
+						reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(2, header.name, matching, $"[c/FFFACD:Switch] to v{header.version}\n(Temporarily downgrade from v{clientMod.Version})"));
+					}
+					else {
+						reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(2, header.name, matching, $"[c/FFFACD:Switch] to v{header.version}\n(Temporarily upgrade from v{clientMod.Version})"));
+					}
+				}
+				else {
+					toEnable.Add(matching.Name);
+					reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(3, header.name, matching, $"[c/98FB98:Enable]")); 
+				}
```
Chicken-Bones, line 236: As recommended below, perhaps we can split the functional code (`toEnable.Add(matching.Name)`) from the UI display message creation

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -262,18 +308,63 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
+				modsWithConfigReload.Add(pendingConfig.configname);
+				// We could mention the specific config if useful: \"{loadTimeConfig.DisplayName}\" Config change");
+				Logging.tML.Debug($"{pendingConfig.modname}:{pendingConfig.configname} ({loadTimeConfig.DisplayName}) would require a mod reload to join this server");
+			}
+		}
+
+		// If needsReload is still false, apply configs to real ModConfig instances and join server directly
 		if (!needsReload) {
 			foreach (NetConfig pendingConfig in pendingConfigs)
 				JsonConvert.PopulateObject(pendingConfig.json, ConfigManager.GetConfig(pendingConfig), ConfigManager.serializerSettingsCompact);
```
Chicken-Bones: Is this common with above code?

Chicken-Bones (commented):
Excellent work and comments. I've made some recommendations on how to factor the code for more maintainability. As you've noted, needs localization.

patches/tModLoader/Terraria/ModLoader/ModNet.cs
JavidPack: No, it needs to be separate, one populates a clone for the check, the other populates the active config directly.

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -225,34 +228,34 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
 
-			// overwrite an existing version of the mod if there is one
-			if (localVersions.Length > 0)
-				header.path = localVersions[0].modFile.path;
-
-			if (downloadModsFromServers && (header.signed || !onlyDownloadSignedMods))
+			if (downloadModsFromServers && (header.signed || !onlyDownloadSignedMods)) {
 				downloadQueue.Enqueue(header);
-			else
+				MakeDownloadModExplanationEntries(modFiles, reloadRequiredExplanationEntries, header, clientMod);
```
Chicken-Bones: Should be `reloadRequiredExplanationEntries.Add(MakeDownloadModExplanation(modFiles, header, clientMod))`

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -225,34 +228,34 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
 
 			needsReload = true;
 
-			LocalMod[] localVersions = modFiles.Where(m => m.Name == header.name).ToArray();
-			LocalMod matching = Array.Find(localVersions, mod => header.Matches(mod.modFile));
+			LocalMod matching = modFiles.FirstOrDefault(mod => header.Matches(mod.modFile));
 			if (matching != null) {
-				matching.Enabled = true;
+				MakeEnableOrVersionSwitchExplanationEntries(reloadRequiredExplanationEntries, header, clientMod, matching);
```
Chicken-Bones: As below (factor out the `List.Add`, only need to make one entry and return it)

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -225,34 +228,34 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
-
-		foreach (Mod mod in clientMods)
-			if (mod.Side == ModSide.Both && !syncSet.Contains(mod.Name)) {
-				ModLoader.DisableMod(mod.Name);
-				needsReload = true;
-			}
+		var toDisable = clientMods.Where(m => m.Side == ModSide.Both).Select(m => m.Name).Except(SyncModHeaders.Select(h => h.name));
+		foreach (var name in toDisable) {
+			needsReload = true;
+			reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(4, name, modFiles.Where(mod => mod.Name == name).OrderByDescending(mod => mod.Version).FirstOrDefault(),Language.GetTextValue("tModLoader.ReloadRequiredExplanationDisable", "FFA07A")));
```
Chicken-Bones: Good

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -268,21 +271,115 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
...
+
+			ModConfig currentConfigClone = ConfigManager.GeneratePopulatedClone(ConfigManager.GetConfig(pendingConfig));
+			JsonConvert.PopulateObject(pendingConfig.json, currentConfigClone, ConfigManager.serializerSettingsCompact);
+			Mod mod = ModLoader.GetMod(pendingConfig.modname);
+			ModConfig loadTimeConfig = ConfigManager.GetLoadTimeConfig(mod, pendingConfig.configname);
+
+			if (loadTimeConfig.NeedsReload(currentConfigClone)) {
+				needsReload = true;
+				reloadRequiredExplanationEntries.Add(new ReloadRequiredExplanation(5, mod.Name, modFiles.FirstOrDefault(mod => mod.Name == pendingConfig.modname), Language.GetTextValue("tModLoader.ReloadRequiredExplanationConfigChanged", "DDA0DD")));
+				modsWithConfigReload.Add(pendingConfig.configname);
```
Chicken-Bones: Can we factor this into something like `ConfigManager.AnyModNeedsReloadCheckOnly`? Can be a method here, or in `ConfigManager`. Something like `(List<NetConfig> pendingConfigs, out List<Mod> modsWithChangedConfigs)`. You can still keep the debug log message in the method. If we wanted to report the config(s) we could change it to `out List<ModConfig>` instead at a later time

JavidPack:
# More Info about the multiplayer mod reload user experience

A new UI has been added to inform the user about pending changes that will require a mod reload. This allows users to decide if they want to risk downgrading, if they want to download new mods, or are accidentally on an incorrect version of a mod. This UI should better inform the user about why mods are reloading. While mods reloading should be much less of an issue now that the Mods folder won't override the workshop folder, it is still useful for users who might not be subscribed to a mod and still have an old version.

Players joining a multiplayer server or returning to single player after joining a multiplayer server will be met with a new window showing various reasons mods are reloading, if reloading is needed. The video below shows this new UI in action: 

https://github.com/tModLoader/tModLoader/assets/4522492/8101e260-cebf-4389-b6bc-b79001aaaabf

PR #3827, opened 2023-09-21
Add BuffDefinition Option

EdoanR:
### What is the new feature?
Add a BuffDefinition options to use on config files, it allow users to easily set buffs using a cool interface. Just like NPCDefinition, ItemDefinition, etc.

screenshot

It will improve the quality of life of modders and users by using an interface to select buffs, instead of manually typing the ID of the buff that want to modify in some way.

It's UI, so yes. Maybe cool filters, for example, only show debuffs, etc.

```cs
// Basic example with no buff selected.
public BuffDefinition buffDefinition = new BuffDefinition();
// Example with a default buff selected.
public BuffDefinition buffDefinitionWithDefaultBuff = new BuffDefinition(BuffID.AmmoBox);
// Example with a default (modded) buff selected.
public BuffDefinition buffDefinitionWithDefaultModdedBuff = new BuffDefinition("ExampleMod", nameof(Content.Buffs.ExampleDefenseBuff));
// List of buffs.
public List<BuffDefinition> listOfBuffDefinitions = new List<BuffDefinition>();
// List of buffs with default value.
[JsonDefaultListValue("{\"name\": \"Ironskin\"}")]
public List<BuffDefinition> listOfBuffDefinitionsWithDefaultValue = new List<BuffDefinition>();
```

### ExampleMod updates
Added an example in the data types config showcase.


JavidPack:
Looks great. We'll leave filters for later, in case anyone really want pursue that. It seems you forgot to push the UIModConfig changes to actually hook it up, but I pushed that for you. 

Thanks a bunch.

Issue #3821, opened 2023-09-18
Disabling a mod due to load failure does not disable dependencies


Chicken-Bones:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

An issue with our new 'automatically enable dependencies on load' feature. If a dependency gets disabled due to a load failure (`The mod(s) have been automatically disabled`), the dependent remains enabled, causing the dependency to be re-enabled next loop.

This means they can get stuck in a reload loop. The only way out currently is to skip loading mods, and identify the dependencies to disable as well.

### Steps to reproduce

- Have mod B which depends on mod A. 
- Write an exception during loading in mod A.
- Enable B and load the game.

### Expected Behavior

The dependent mods should also be disabled (if the dependency is a strong one, and not a weak one).

### Actual Behavior

The dependency remains enabled, and re-enables the dependent on each load, leaving the user stuck in a loop.

### Reproduction frequency

100%

JavidPack:
Fixed in #4112

PR #3829, opened 2023-09-22
Unload dependents of a mod which failed to load

MacaylaMarvelous81:
### What is the bug?
#3821 - If a mod which is a hard dependency of other mods is unloaded due to a load error, the dependency mod is unloaded while the dependent mod(s) are not unloaded. This can cause a load loop since the culprit mod will be loaded again when dependencies are automatically enabled.

### How did you fix the bug?
With this pull request, when a mod is unloaded because of a load error, the mod loader will search for mods which depend on it and disable them. It does this by checking if any of the `availableMods` have the dependency mod in the `modReferences`.

I'm not aware of any other way to get the dependents of a mod other than doing the inverse and searching the dependencies of other mods to find them. Another way to get the dependents would likely be more logical for reading the code, and it would be better if not every mod would have to be checked.

I'm also unsure about the message localization. I added a string to the en-US localization so that the message will indicate which dependents were disabled, if any, but I'm unfamiliar with how localization works in tModLoader. Is it OK for me to only add it to the en-US file?

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModLoader.cs
@@ -178,8 +178,16 @@ public static class ModLoader
 
 			Logging.tML.Error(msg, e);
 
-			foreach (var mod in responsibleMods)
+			foreach (var mod in responsibleMods) {
 				DisableMod(mod);
+
+				var dependents = from m in availableMods
+					where m.properties.modReferences.Any(reference => reference.mod.Equals(mod))
+					select m.Name;
+				
+				foreach (var dependent in dependents)
+					DisableMod(dependent);
+			}
 
 			isLoading = false; // disable loading flag, because server will just instantly retry reload
 			DisplayLoadError(msg, e, e.Data.Contains("fatal"), responsibleMods.Count == 0);
--- a/patches/tModLoader/Terraria/ModLoader/ModLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModLoader.cs
@@ -179,9 +179,9 @@ public static class ModLoader
 			foreach (var mod in responsibleMods) {
 				DisableMod(mod);
 
-				var dependents = from m in availableMods
-					where m.properties.modReferences.Any(reference => reference.mod.Equals(mod))
-					select m.Name;
+				var dependents = availableMods
+					.Where(m => m.properties.modReferences.Any(reference => reference.mod.Equals(mod)))
+					.Select(m => m.Name);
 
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -176,11 +176,22 @@ private static void Load(CancellationToken token = default)
 			if (e.Data.Contains("contentType") && e.Data["contentType"] is Type contentType)
 				msg += "\n" + Language.GetTextValue("tModLoader.LoadErrorContentType", contentType.FullName);
 
-			Logging.tML.Error(msg, e);
-
-			foreach (var mod in responsibleMods)
+			foreach (var mod in responsibleMods) {
 				DisableMod(mod);
 
+				var dependents = from m in availableMods
```
Chicken-Bones: We don't use sql linq syntax in tML. Please switch to functional style.

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -176,11 +176,22 @@ private static void Load(CancellationToken token = default)
...
+					where m.properties.modReferences.Any(reference => reference.mod.Equals(mod))
```
Chicken-Bones: `BuildProperties` has the `RefNames` method which will help here. We used explicit named args when passing constants, so you'll want to do `RefNames(includeWeak: false)` which makes it really clear that you're only considering strong references as required deps.

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -176,11 +176,22 @@ private static void Load(CancellationToken token = default)
...
+					select m.Name;
+
+				msg += "\n" + Language.GetTextValue("tModLoader.LoadErrorDependentsDisabled", mod,
```
Chicken-Bones, line 186: Please aggregate all the disabled dependents into one list so the message only gets added once. 

Please also filter to only enabled mods.

Additionally, you will need to write this in a recursive manner, because you may have a series of mod references `B requires C requires A` and if `A` gets disabled, then you have to disable `C` and then `B` as well.

Chicken-Bones (changes requested):

JavidPack:
@MacaylaMarvelous81 is this still supposed to be a draft? You seem to have made the changes addressing the feedback already. I'd be happy to finish this up for you if there is anything still left to implement. 

I noticed, however, that the "Allow edits from maintainers." checkbox is not checked for this PR. You'll need to check this if you want me to finish this up, otherwise I can't commit my edits.

image

PR #3839, opened 2023-09-25
tModLoader.CodeAssist overhaul

BasicallyIAmFox:
### What is the new feature?
Reworked analyzers for tML.
- Renamed tModLoader.CodeAssist to tModCodeAssist to fit with the tML branding
- Moved tModCodeAssist to the tML repository and added it as an for mods analyzer in `tModLoader.targets`
- Ported the old analyzer rules
- Added new rules

Having tModCodeAssist in the tModLoader repository is much better than having it separately and publishing it as a nuget package because we don't have to consider supporting 1.3, 1.4.3, and 1.4.4 at the same time in one codebase. New analyzer rules will also help new modders port vanilla code and can help them simplify their own code.

- Leaving the name as tModLoader.CodeAssist (bad)
- Leaving the analyzer in a separate repository and making tML a submodule of it (bad)
- Leaving the analyzer as a nuget package (bad)

image
image

### ExampleMod updates
N/A

### TODO
* [ ] Magic Number rules
* [ ] Port `rand` simplify rule
* [ ] Implement Array Range rule
* E.g. `NPC.ai[4]` will error now.
* [ ] Implement `Item.value = Item.buyPrice(...)` rule
* E.g. `Item.value = 560` will be suggested to change to `Item.value = Item.buyPrice(silver: 5, copper: 60)`
* [ ] Report usage of `System.Random` (and replace with `Main.rand`, that's what used in 99.9% cases anyway)
* [ ] Implement analyzer that will detect bad usage on `ModContent.GetInstance<T>()`
* E.g. `GetInstance<MyItem>().Foo = 5;` will warn modder
* [ ] `Item.maxStack = 9999` -> `Item.maxStack = Item.CommonMaxStack` rule
* [ ] Add more potential analyzers!


steviegt6:
Points of concern:
- I believe some mods are evil and actually extend the `ai` arrays to allow for additional values,
- there are valid reasons for mods to want to "reset" the research count to 1 (i.e. tweaking another mod).

I understand that analyzers can be configured, and I'm perfectly fine with the former point being a configurable warning (certainly should not be an error though).

The latter should always be a hint (which is how it appears to be based on the screenshot), but I want to establish this as a standard. Changing/resetting research counts is very reasonable mod behavior unlike extending the length of the `ai` array, so I'd like to make sure we leave these suggestions as easily-ignorable hints.

Just my two cents.

steviegt6:
Took a glance at the content of the PR and:
1. might it be better to shove annotations and IdDictionary instances in a TML partial class for IDs,
2. and would it make more sense to do tML annotations through an XML file or something instead of manually patching each field and class?

For 2), it just reduces patch quantity and sizes, but may also make it easier to keep tML up-to-date. Introducing this infrastructure can make it easier to annotate dependent mods or other external dependencies if they aren't annotated.

BasicallyIAmFox:
> Took a glance at the content of the PR and:
> 
>     1. might it be better to shove annotations and IdDictionary instances in a TML partial class for IDs,
> 
>     2. and would it make more sense to do tML annotations through an XML file or something instead of manually patching each field and class?
> 
> 
> For 2), it just reduces patch quantity and sizes, but may also make it easier to keep tML up-to-date. Introducing this infrastructure can make it easier to annotate dependent mods or other external dependencies if they aren't annotated.

1) will do but later. Not major priority atm.
As for 2), it sounds interesting, but I have no clue how to do it, unfortunately.

BasicallyIAmFox:
> Points of concern:
> 
>     * I believe some mods are evil and actually extend the `ai` arrays to allow for additional values,
> 
>     * there are valid reasons for mods to want to "reset" the research count to 1 (i.e. tweaking another mod).
> 
> 
> I understand that analyzers can be configured, and I'm perfectly fine with the former point being a configurable warning (certainly should not be an error though).
> 
> The latter should always be a hint (which is how it appears to be based on the screenshot), but I want to establish this as a standard. Changing/resetting research counts is very reasonable mod behavior unlike extending the length of the `ai` array, so I'd like to make sure we leave these suggestions as easily-ignorable hints.
> 
> Just my two cents.

I haven't seen any mod that modifies research count. Could you point me any mods doing that?

tModCodeAssist/tModCodeAssist.sln
```diff
@@ -0,0 +1,43 @@
...
+VisualStudioVersion = 17.7.33913.275
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "tModCodeAssist", "tModCodeAssist\tModCodeAssist.csproj", "{D9A1DECC-30F4-47C2-BEFF-DAEA6C543130}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "tModCodeAssist.CodeFixes", "tModCodeAssist.CodeFixes\tModCodeAssist.CodeFixes.csproj", "{3377288E-1DE3-4BC0-AB1E-41C119EAB98A}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "tModCodeAssist.Test", "tModCodeAssist.Test\tModCodeAssist.Test.csproj", "{216CBE85-191C-4B1F-A842-036465EEAC88}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "tModCodeAssist.Vsix", "tModCodeAssist.Vsix\tModCodeAssist.Vsix.csproj", "{48A675E6-AC78-4716-9020-6EA3D465CE01}"
+EndProject
```
Mirsario, line 13: Is there a single point in offering a VS extension nowadays? I don't see one.

tModCodeAssist/tModCodeAssist.CodeFixes/tModCodeAssist.CodeFixes.csproj
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
Mirsario, line 4: Most of our `.csproj` and `.targets` files use tabs nowadays.

patches/tModLoader/Terraria/release_extras/tMLMod.targets
```diff
@@ -4,6 +4,7 @@
 		<TerrariaSteamPath>$(MSBuildThisFileDirectory)</TerrariaSteamPath>
 		<tMLSteamPath>$(MSBuildThisFileDirectory)</tMLSteamPath>
 		<tMLLibraryPath>$(tMLSteamPath)Libraries</tMLLibraryPath>
+		<tMLAnalyzerPath>$(tMLSteamPath)Analyzers</tMLAnalyzerPath>
 		<tMLName>tModLoader</tMLName>
```
Mirsario, line 8: No need for a new folder, output analyzers into `Libraries`, the way `tModPorter` and `tModLoader.BuildTasks` get put there:
image

patches/tModLoader/Terraria/ID/AnimationID.TML.cs
```diff
@@ -0,0 +1,11 @@
+﻿using ReLogic.Reflection;
+
+namespace Terraria.ID;
+
+#if TMLCODEASSIST
+[tModCodeAssist.IDType.Sets.AssociatedName(ModLoader.Annotations.IDTypeAttribute.Animation)]
+#endif
+partial class AnimationID
```
Mirsario, line 8: Why are these preprocessor conditionals required?

Mirsario (changes requested):

BasicallyIAmFox:
how do i reopen this lmao

Mirsario:
Recreate the PR and stop rewriting history so much. We squash PRs, so you don't need an overly clean history in PRs.

Mirsario:
.

BasicallyIAmFox:
> Recreate the PR and stop rewriting history so much. We squash PRs, so you don't need an overly clean history in PRs.

wtf you mean "recreate the pr"? i wanted to rewrite the pr from scratch either way

PR #3843, opened 2023-09-26
Fixes and improvements to smart select while using auto reusing

### LINKED ISSUE #3838: Smart cursor deleting items when using it combined with torches
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When spamming your smart cursor hotkey with a small amount of torches (say, 5-6) in your inventory while attempting to mine trees using an axe (I used the Picksaw), items in the first 2 slots will be deleted.

### Log File

Not applicable, no errors in log

### Steps to reproduce

Unsure exactly how to write a steps to reproduce for this, but I'll try:
1. Give yourself around 5 or 6 torches, an axe (preferably a fast one, maybe the picksaw), and a weapon to put in slot 2 using a cheat mod, then disable all mods.
2. Load into a world (preferably a new one) and use the axe item while spamming your smart cursor hotkey (to attempt to place torches).
3. Notice how when the torch stack reaches 0, items will be deleted out of your hotbar.

### Expected Behavior

It should not delete your items.

### Actual Behavior

Items are deleted.

### Reproduction frequency

Every time I try it.

### Additional Information

I recorded a video clip of this happening, and uploaded to YT. It's near the end of the video.

[https://www.youtube.com/watch?v=fuxMfAYiuzY](https://www.youtube.com/watch?v=fuxMfAYiuzY)

Chicken-Bones:
### What is the bug?
- Fixes #3838

Additionally:
- When holding smart select and using a pickaxe, the selected item would switch to a torch after mining the block, which is annoying and undesired.

### How did you fix the bug?
- Added `item = inventory[selectedItem]` when smart select changes the `selectedItem` to fix the 'consume wrong item' bug.
- Whenever the selected item is changed (either via hotbar key or smart select) the cursor release flag is set, allowing the item to be used immediately even if it doesn't have `autoReuse`
  - This allows a glowstick to be thrown immediately on switching to it while holding down the mouse, which is fun
- Prevented smart select from changing the item while the use button is held down, so you can't swap from pickaxe to torch or vice versa while auto-reusing.

- Perhaps the user doesn't want to immediately use the item if they press shift or a hotbar key while holding the mouse button down, but in this case, just release the mouse first?

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -2809,6 +2809,18 @@
  		for (int i = 0; i < maxBuffs; i++) {
  			if (buffType[i] <= 0 || !Main.persistentBuff[buffType[i]]) {
  				buffTime[i] = 0;
+@@ -13526,6 +_,11 @@
+ 			controlTorch = false;
+ 
+ 		if (controlTorch && itemAnimation == 0) {
++
++			// TML feature, prevent changing the smart selected item while the use button is held. This prevents switching from pickaxe to torch after mining a block for eg
++			if (controlUseItem && nonTorch != -1)
++				return;
++
+ 			PlayerInput.smartSelectPointer.SmartSelectLookup_GetTargetTile(this, out var tX, out var tY);
+ 			SmartSelect_GetToolStrategy(tX, tY, out var toolStrategy, out var wetTile);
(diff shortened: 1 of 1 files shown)


JavidPack:
The only issue I found is if you smart select to torch/glowstick and use the last of the stack, you don't keep using the original item when you let go of shift. You have to let go of the use button and then press it again to resume shooting/mining. Not a big deal, but something that might be requested at a later date by users.

Issue #2926, opened 2022-09-15
FNA crashes with bad SDL_VIDEODRIVER environment variable


JavidPack:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Game can't launch for some users with this issue. 

![unknown (11)](https://user-images.githubusercontent.com/4522492/190300166-61e72596-3c20-4061-892e-e3019844fb20.png)

Lines in question:

https://github.com/tModLoader/FNA/blob/24a2c585c08af681718d437f3f793a5ced1b49c5/src/FNAPlatform/SDL2_FNAPlatform.cs#L223-L225

### Log File

[message.txt](https://github.com/tModLoader/tModLoader/files/9571227/message.txt)

Chicken-Bones:
Hmm, most likely `videoDriver` is null. Other lines use `==` instead of `.Equals`. Can someone who is experiencing this issue ping me on Discord?

Chicken-Bones:
Investigating in https://discord.com/channels/103110554649894912/1019436159924969533

Chicken-Bones:
Confirmed, issue is env var, `SDL_VIDEODRIVER=directx`, no longer supported. <https://wiki.libsdl.org/FAQUsingSDL>

Was difficult to debug due to lack of <https://github.com/FNA-XNA/FNA/pull/395>

PR #3844, opened 2023-09-26
Add general purpose SDL driver fallbacks, error reporting and cli args

Chicken-Bones:
### What is the new feature?
- `-videodriver` and `-audiodriver` launch args, which override `SDL_VIDEODRIVER` and `SDL_AUDIODRIVER` env vars respectively.
- Append fallback driver list whenever `SDL_VIDEODRIVER` or `SDL_AUDIODRIVER` is specified.
  - This is a more general solution to #2926
  - Still allows for selecting a default driver, but allows SDL to ignore it and continue on to the normal driver list if it's not present.
- `SDL Error:` if any is appended to `FatalExit` logs
- The _selected_ `SDL Video Driver` and `SDL Audio Driver` are reported in the logs, after `FNA3D` and `FAudio` are initialized respectively.

Sometimes people have bad env vars on their system, particularly for older SDL versions. SDL does not fall-back to the default driver search order internally when an env var is specified, so if the driver in the env var is absent or inoperable, then the game crashes.

Additional launch args and logging make it easier to get users to try alternative drivers when they experience issues.

Only in minor ways, we'll see if any issues crop up.

`-audiodriver wsapi`
`-videodriver x11`

### DIFF
--- a/patches/tModLoader/Terraria/Audio/LegacyAudioSystem.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/LegacyAudioSystem.cs.patch
@@ -31,7 +31,7 @@
  						break;
  				}
  
-@@ -100,10 +_,12 @@
+@@ -100,10 +_,13 @@
  
  	public LegacyAudioSystem()
  	{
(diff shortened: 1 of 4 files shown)


JavidPack:
Looks good to me. I'm not really sure how to test it effectively since my system is launching the game and it's correctly falling back even with a missing driver option.

We'll want to get this into preview and remember to use it in #support-forum as something to tell users to test with. See if we can confirm that similar issues are resolved with this fix without additional efforts.

JavidPack (approved):

Mirsario (approved):

Solxanich (approved):
Arguments added to https://github.com/tModLoader/tModLoader/wiki/Command-Line

PR #3851, opened 2023-09-29
Fix #2020

### LINKED ISSUE #2020: Fullscreen Windowed missing (Settings/Video/Resolution/Borderless Window option missing)
### Version

Latest public-1.4-alpha

### OS

Windows

### Platform

Steam

### Description

The option to use "Fullscreen Windowed"  is missing in the latest Tmodloader (it is available in terraria still)
On discord there was mention that the standard fullscreen option would be borderless windowed now but this does not seem to be the case.

I use 1 monitors, my main monitor is an 3440 x 1440 (WQHD) screen, the second monitor is a regular 1080p resololution.

When in fullscreen mode, if you click in the second monitor, Tmodloader minimizes
When in windowed mode if you click in the second monitor, Tmodloader stays active/in focus. 

Windowed mode has the titlebars visible which is less then ideal. The option "Borderless Window" should fix that, but it is not available in the menu
2021-12-04 14_19_15-Terraria_ May the blocks be with you

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

shuen4:
originally fixed in 5161f2e but broken after 88c4e68

## details
when executing `Terraria.ModLoader.Engine.FNAFixes.Init()`, `Microsoft.Xna.Framework.SDL2_FNAPlatform` is not yet initialized

execution routine:
```
// ...
FNAFixes.Init();
// ...
new Main();
base(); // will initialize FNAPlatform and overwrite previously set SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS value
```

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Engine/FNAFixes.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Engine/FNAFixes.cs
@@ -10,7 +10,7 @@ internal static class FNAFixes
 	{
 		if (OperatingSystem.IsWindows()) {
 			// FNA sets this to "1" on Windows. Terraria does not want this. See #2020
-			SDL.SDL_SetHint(SDL.SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS, "0" );
+			SDL.SDL_SetHintWithPriority(SDL.SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS, "0", SDL.SDL_HintPriority.SDL_HINT_OVERRIDE);
 		}
 
 		ConfigureDrivers();


Chicken-Bones:
Hi, yes, I noticed this recently, but I was considering leaving it in the same place and setting it using `HINT_PRIORITY_OVERRIDE`

shuen4:
changed to use `SDL_SetHintWithPriority` and `SDL_HINT_OVERRIDE` in [17b0802](https://github.com/tModLoader/tModLoader/pull/3851/commits/17b0802f7237474a65e1886ae003b155f7f2ade7)

PR #3853, opened 2023-10-01
translated 2 messages into russian

Doodlinka:
### What is the translation?
"CancellingLoading" and "LoadingCancelled" have been translated into Russian. As far as I know, only unused messages are left untranslated into Russian now.

### How did you translate it?
Edited ru-RU/tModLoader.json to change the values in lines 471 and 472 to the translations, and uncommented them.
"CancellingLoading" -> "Отменение загрузки..."
"LoadingCancelled" -> "Загрузка отменена"

### Are there alternatives to your translation?
Both are literal, and I don't know the context they show up in, so they may be clumsy. "Отменение загрузки..." is a noun, which is consistent with the other translations, but sounds a bit unnatural to me.


GariestGary:
there is no word "Отменение" in russian. "Отмена загрузки..." more suitable in this case

PR #3854, opened 2023-10-01
Steam Store Page zh-Hans Localization

lace-wing:
### What is the new feature?

Simplified Chinese translation for tModLoader's Steam Store Page (needs manual uploads though).

### Why should this be part of tModLoader?

So that Chinese readers can better understand what it is.

PR #3858, opened 2023-10-02
Simplified Chinese translation supplement

ayuyoiM:
### What's in PR?

This PR has a Simplified Chinese translation supplement for tModLoader and some Simplified Chinese translation supplements for Steam

### What did you translate?

See commits

> ~~I know of another PR who also submitted the same content~~


JavidPack:
I'm not quite sure what to do about the overlap between this and #3854, can you two talk and maybe consolidate the best translations into a single PR? I'm not really able to judge which is better myself.

JavidPack:
Looks good. I'll get these uploaded within a day or 2.

PR #3876, opened 2023-10-12
Fixed ExampleTravelingMerchant Shop Not Functioning On Servers

DavidF-Dev:
### What is the bug?
Example Traveling Merchant's shop does not function in multiplayer. An exception occurs in `FillShop` due to `shopItems` being null when attempting to open the shop. `shopItems` is initialised and populated in `OnSpawn` **AND** in `LoadData`, which are only called by the server. This is not synced.

### How did you fix the bug?
I have synced `shopItems` as part of the NPC's extra AI. When the NPC is spawned on the server, it populates `shopItems` and syncs it with the clients.

There may be. Syncing all the items with the NPC's AI may not be the most efficient solution, but it works. If another solution is proposed, it must account for syncing for joining players **AND** when the NPC is first spawned.


Chicken-Bones:
A concerning amount of data for `SyncNPC` which is already one of the most high traffic packets in the game. Terraria doesn't have a way of infrequently sending large amounts of NPC specific data, so this will probably have to be a custom packet and `ModSystem`.

DavidF-Dev:
I agree. 

Seeing as there's only one instance of the Example Traveling Merchant NPC in the world at a time, the `shopItems` list could be stored on a `ModSystem` instead, including saving/loading and syncing. This could piggyback off of `MessageID.WorldData`.  Feels messy. I wonder how vanilla handles this.

I'll try a few approaches tomorrow.

DavidF-Dev:
I have modified the code to only sync the shop items when the NPC is spawned (`OnSpawn`) and when it detects a player has joined (`AI`). It still uses `Send/ReceiveExtraAI`, but it selectively chooses when to send the item list instead of it always being sent.

ExampleMod/Content/NPCs/ExampleTravelingMerchant.cs
```diff
@@ -236,9 +248,37 @@ class ExampleTravelingMerchant : ModNPC
 		}
 
 		public override void LoadData(TagCompound tag) {
-			shopItems = tag.Get<List<Item>>("shopItems");
+  			shopItems.Clear();
+			shopItems.AddRange(tag.Get<List<Item>>("shopItems"));
 		}
 
+  		public override void SendExtraAI(BinaryWriter writer) {
+			if (!netShopItemsUpdate) {
```
Chicken-Bones: This is a good implementation of one of the ideas I had in mind, but it unfortunately has some downsides.
1. The information is not sent to new players immediately
2. The new player detection check can fail if a player leaves and joins in the same frame (unlikely but possible)
3. The update is sent to all players, not just the joining player, and also sent whenever a player leaves.

Given that `TravelingMerchantSystem` already exists, and is required for updating the spawn logic, you could send the items in the `NetSend` hook there.

Surprisingly, tModLoader doesn't offer any hooks in the `OnPlayerJoining` space (response to message 8). Despite vanilla having many systems which send packets with the same name, and an unused `IOnPlayerJoining` interface.

I like the idea of a more NPC centric approach, but I'd like something a bit more directed towards specific use cases. Note also that `SyncNPC` has some poorly implemented 'nearby NPC broadcast culling' via `NPC.netSkip`. This doesn't apply to `townNPC`, `boss` or `netAlways` NPCs, but people might not notice this particular limitation.

DavidF-Dev: Alright. See my two recent commits. The NPC is no longer responsible for saving/loading or syncing `shopItems`. Instead, the ModSystem handles that.

Chicken-Bones:
Thanks for this, I'll fix the merge and get it in. I'll add a comment cautioning that piggy-backing on the `WorldData` packet is probably not the greatest idea for network usage, but it gets sent sufficiently infrequently that it's probably okay.

PR #3879, opened 2023-10-13
Sunflowers, Pumpkins, and Fallen Logs now honor AnchorValidTiles to determine what tiles they can exist on

Lion8cake:
### What is the new feature?
This contains changes to both Sunflowers, Pumpkins and Fallen Logs to prevent them from either not generating on modded tiles or breaking when on modded tiles. The following is how each is changed to allow modders to add support for sunflowers, pumpkins and fallen logs to their own tiles:

For Sunflowers and Pumpkins, they have been updated to react to their AnchorValidTiles instead of a hardcoded list (curtesy of Jopojelly for amazing idea and help with sunflowers) allowing for modders to easily add their tile to the AnchorValidTiles of either Sunflowers or Pumpkins using a GlobalTile. 

For Fallen Logs they have a new TileID.Set called FallenLogPlaceable. A tile with this ID set will allow said tile to have the fallen log tile on top of it without the fallen log breaking. This is good for modders who a biome may have similar to the hallow where when generating grass or hallowed grass will be replaced with a modded tile. This prevents all fallen logs with replaced modded tiles from instantaneously breaking.

This allows modded tiles to have the option to have these three tiles (Sunflowers, Pumpkins and Fallen Logs) to grow and be placed on top of certain modded tiles. This also reduces the amount of IL edits needed for support of the three tiles and all three prior to this require IL editing to edit the hardened lists of what tiles the three tiles can exist upon.

None that I can see currently, there might be alternate routes others might be able to identify.

For Sunflowers and Pumpkins
(curtesy of JopoJelly)
```
namespace ExampleMod.Commom.GlobalTile
{
    public class ExampleGlobalTileSystem : GlobalTile
    {
			TileObjectData tileObjectData = TileObjectData.GetTileData(TileID.Sunflower, 0);
			tileObjectData.AnchorValidTiles = tileObjectData.AnchorValidTiles.Append(ModContent.TileType<ExampleTile>()).ToArray(); //This allows the example tile to have sunflowers placed on top of it

			TileObjectData tileObjectData2 = TileObjectData.GetTileData(TileID.Pumpkins, 0);
			tileObjectData2.AnchorValidTiles = tileObjectData2.AnchorValidTiles.Append(TileID.JungleGrass).ToArray(); //This allows pumpkins to grow on jungle grass.
		}
	}
}
```
For Fallen Logs
```
TileID.Sets.FallenLogPlaceable[Type] = true; //Goes in SetStaticDefaults of whatever tile you want to add this too
```

### DIFF
--- a/ExampleMod/Common/GlobalTiles/SunflowerChanges.cs
+++ b/ExampleMod/Common/GlobalTiles/SunflowerChanges.cs
@@ -1,0 +1,25 @@
+﻿using ExampleMod.Content.Tiles;
+using System.Linq;
+using Terraria.ID;
+using Terraria.ModLoader;
+using Terraria.ObjectData;
+
+namespace ExampleMod.Common.GlobalTiles
+{
+	// This example uses a GlobalTile to affect the properties of an existing tile.
+	// Tweaking existing tiles is doable to some degree.
+	public class SunflowerChanges : GlobalTile
+	{
+		public override void SetStaticDefaults() {
+			// This allows the Sunflower tile to be placed on ExampleBlock
+			TileObjectData tileObjectData = TileObjectData.GetTileData(TileID.Sunflower, 0);
+			tileObjectData.AnchorValidTiles = tileObjectData.AnchorValidTiles.Append(ModContent.TileType<ExampleBlock>()).ToArray();
+		}
+
+		public override void Unload() {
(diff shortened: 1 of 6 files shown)


JavidPack:
A note about world gen. Sunflowers, pumpkins, and fallen logs will all need to be placed manually if desired because the worldgen code that places them only care about specific tiles. Pumpkins growing in-game will are the same. This PR allows them to be placed on the tiles, but won't affect the existing logic for which tiles are candidates for ingame and worldgen placement.

Issue #3887, opened 2023-10-16
Allow Player.Save/LoadPlayer from stream instead of just path


zzp198:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

tModLoader code as a Contributor/Maintainer

### Description

Separate Player.Save/LoadPlayer into two parts.
Data generation, reading and writing.

Allow loading and saving through path or data.

The data of PLR and TPlr will be read and written simultaneously.

### What does this proposal attempt to solve or improve?

Single responsibility, and allowing players to save and load outside of the path.

PR #3888, opened 2023-10-16
solve #3887, Split function for better hooking

zzp198:
Simply dividing a function into two, changing the parameters from only allowing paths to allowing streams


Chicken-Bones (changes requested):
It looks like your editor has made several whitespace changes that are unrelated to the change. Please go through the diffs in github and your git client and revert any lines which aren't code changes

PR #3891, opened 2023-10-17
Show latest news on main menu

JavidPack:
We've found it difficult to keep modders and mod users updated with information. This PR suggests implementing a small notification in the main menu to prompt interested users and modders to click to read the latest tModLoader news.

Posting breaking changes to just Discord doesn't reach our full audience, some mod teams subscribe to our announcement channel within their own Discord servers, but not all do that. Users might notice news on Steam, but might not.

This adds one more way for modders and users to be exposed to the latest tModLoader news. 

The video shows the behavior in action. On reaching the main menu, the steam news api is checked. The news title will be displayed. If the news is newer than previously seen news, it'll display as disco colors to bring attention to it. Once clicked or next game session, the notification will be a less attention grabbing white color:

https://github.com/tModLoader/tModLoader/assets/4522492/a348eddb-dc1a-4f73-a7c1-566631901b74

Unfortunately, we can't easily access localized versions of these news entries using the web api (the current code only cares about the title), maybe that is something we could look into.


Chicken-Bones:
I recommend automatically updating `LatestNewsTimestamp` to the build date version the user is running, perhaps only if stable. The point of the flashy text is to alert people of _upcoming_ changes, not ones they already have.

Otherwise LGTM

JavidPack:
I'm not sure an additional check is needed. Even if a user is automatically updated, they would still be interested in what actually updated. They might have noticed that a mod suddenly is broken, the flashing color would suggest to them somewhere to look for clues. We usually post the announcement when the build goes live, so there never is really any upcoming changes unless a user is launching an old manually installed release.

patches/tModLoader/Terraria/Main.TML.cs
```diff
@@ -447,4 +451,51 @@ public static void AddSignalTraps()
...
+	private static bool newsIsNew = false;
+	private static void HandleNews(Color menuColor)
+	{
+		if (menuMode == 0) {
+			if (!newsChecked) {
+				newsChecked = true;
+				// Download latest news, save to config.json.
+				// https://partner.steamgames.com/doc/webapi/ISteamNews
+				client.GetStringAsync("https://api.steampowered.com/ISteamNews/GetNewsForApp/v2/?appid=1281930&count=1").ContinueWith(response => {
+					JObject o = JObject.Parse(response.Result);
```
Solxanich: Consider error handling in the event computer / steam host is offline to minimize logged text.

Consider error handling for various steam responses: https://partner.steamgames.com/doc/webapi_overview/responses#status_codes

Solxanich (commented):
Hmm. I agree with JavidPack's assessment that adding an additional check of last built version likely won't be of huge benefit. 
This isn't in your face and aggravating (unlike some other games), so I think this is reasonable implementation.

On the code side, some minor comments error handling surrounding the Steam web query

PR #3899, opened 2023-10-25
#3887 Allow Player.Save/LoadPlayer from stream

zzp198:
### What is the new feature?

Refine save/load functions.

Original request: #3887

The current save and load parameters only allow file path or cloud save, which means that data generation and writing are inseparable. Additionally, because the player data is divided into two parts, `.plr` and `.tplr`, it is difficult to use `Hook` to obtain both files simultaneously when intercepting saving.

This change does not break the original interface and allows developers to obtain players directly from binary data without using path, making “memory player” possible.

No, 

If you have a player's data, it can be used without writing files or hooks.

```cs
byte[] plr_data = ... // from plr file or packet
byte[] tplr_data = ... // from tplr file or packet
var streamPlayer = Player.LoadPlayerFromStream(new PlayerFileData, plr_data, tplr_data);
```

```cs
PlayerLoader.PreSavePlayer(player);
byte[] plrData = SavePlayerFile_Vanilla(playerFile);
PlayerLoader.PostSavePlayer(player);
TagCompound tplrData = PlayerIO.SaveData(player);
```

The `SavePlayerFile_Write` function can be hooked to redirect save data to a different storage medium

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -7787,11 +7787,12 @@
  			throw;
  		}
  	}
-@@ -41979,21 +_,70 @@
+@@ -41979,21 +_,71 @@
  		if (string.IsNullOrEmpty(path))
  			return;
  
 +		// TML: Split method into several methods to facilitate separation between saving to disk and generating save data.
++		// TML: Delay vanilla saving until modded data is serialized(in case of exceptions)
 +		// Moved to UpdateLocalPlayerFile
 +		/*
  		if (FileUtilities.Exists(path, isCloudSave))
@@ -8032,13 +8033,12 @@
  		catch {
  		}
  
-@@ -42308,7 +_,8 @@
+@@ -42308,7 +_,7 @@
  			player2.name = player.name;
  		}
  		else {
 -			string[] array = playerPath.Replace('/', Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar);
-+			string[] array = playerFileData.Path.Replace('/', Path.DirectorySeparatorChar)
-+				.Split(Path.DirectorySeparatorChar);
++			string[] array = playerFileData.Path.Replace('/', Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar);
  			player2.name = array[array.Length - 1].Split('.')[0];
  		}
  
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -7792,8 +7792,8 @@
  			return;
  
 +		// TML: Split method into several methods to facilitate separation between saving to disk and generating save data.
-+		// TML: Delay vanilla saving until modded data is serialized(in case of exceptions)
-+		// Moved to UpdateLocalPlayerFile
++		// TML: Delay vanilla saving until modded data is serialized (in case of exceptions)
++		// Moved to SavePlayerFile_Write
 +		/*
  		if (FileUtilities.Exists(path, isCloudSave))
  			FileUtilities.Copy(path, path + ".bak", isCloudSave);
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7964,8 +7974,21 @@
...
++		var plrData = FileUtilities.ReadAllBytes(playerPath, cloudSave);
++		// Read data in advance instead of halfway
++		// What if there is an exception in PlayerIO.TryLoadData? It is not handled the same as before
++		if (!PlayerIO.TryLoadData(playerFileData.Path, playerFileData.IsCloudSave, out var tplrData))
++			tplrData = null;
++
++		return LoadPlayerFromStream(playerFileData, plrData, tplrData);
++	}
++
++	public static PlayerFileData LoadPlayerFromStream(PlayerFileData playerFileData, byte[] plrData,TagCompound tplrData)
```
Chicken-Bones: As you noted, this messes with the current exception handling. I'm surprised you need `LoadPlayerFromStream` and that `Deserialize` isn't sufficient for you. There's not much else in `LoadPlayerFromStream` other than error handling and 3 lines of post-processing.

If you need this method, then the options are:
- Pass `byte[] tplrData` (preferred)
- Pass `Func<PlayerFileData, TagCompound> getTplrData`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7787,56 +7787,66 @@
...
  			FileUtilities.Copy(path, path + ".bak", isCloudSave);
 +		*/
++		PlayerLoader.PreSavePlayer(player);
  
- 		RijndaelManaged rijndaelManaged = new RijndaelManaged();
++		byte[] plrData = GeneratePlayerData(playerFile);
 +
-+		/*
- 		using Stream stream = (isCloudSave ? ((Stream)new MemoryStream(2000)) : ((Stream)new FileStream(path, FileMode.Create)));
-+		*/
```
Chicken-Bones: Please keep the comments around removed vanilla code

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7787,56 +7787,66 @@
...
  		using CryptoStream cryptoStream = new CryptoStream(stream, rijndaelManaged.CreateEncryptor(ENCRYPTION_KEY, ENCRYPTION_KEY), CryptoStreamMode.Write);
  		using BinaryWriter binaryWriter = new BinaryWriter(cryptoStream);
-+
-+		PlayerLoader.PreSavePlayer(player);
-+
  		binaryWriter.Write(279);
  		playerFile.Metadata.Write(binaryWriter);
- 		Serialize(playerFile, player, binaryWriter);
+-		Serialize(playerFile, player, binaryWriter);
++		Serialize(playerFile, playerFile.Player, binaryWriter);
```
Chicken-Bones: Better to copy `var player = playerFile.Player` into the top of the method, to keep all the patches together with the method split/change

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7787,56 +7787,66 @@
...
 +
-+		byte[] bytes = ((MemoryStream)stream).ToArray();
-+
-+		PlayerLoader.PostSavePlayer(player); // For native data only
++		return ((MemoryStream)stream).ToArray();
++	}
 +
-+		//TML: Delay vanilla saving until modded data is serialized (in case of exceptions)
-+		var tag = PlayerIO.SaveData(player);
++	private static void UpdateLocalPlayerFile(PlayerFileData playerFile, byte[] plrData, TagCompound tplrData)
```
Chicken-Bones: Better method would be `SavePlayerFile_Write`

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7787,56 +7787,66 @@
...
- 		using Stream stream = (isCloudSave ? ((Stream)new MemoryStream(2000)) : ((Stream)new FileStream(path, FileMode.Create)));
-+		*/
++		PlayerLoader.PostSavePlayer(player);
++
++		TagCompound tplrData = PlayerIO.SaveData(player);
++
++		UpdateLocalPlayerFile(playerFile, plrData, tplrData);
++	}
++
++	public static byte[] GeneratePlayerData(PlayerFileData playerFile)
```
Chicken-Bones: Consider `SavePlayerFile_Vanilla`

Chicken-Bones (changes requested):
My apologies for taking so long to review this. You've done good work here. Just a few minor things to discuss/adjust

patches/tModLoader/Terraria/Player.cs.patch
zzp198: Serialize and Deserialize are only target to player data, while PLR files will have encrypted headers and metadata.

I separated LoadPlayerFromStream instead of using Deserialize, hoping to use PLR binary as a arg to obtain the player without processing the header (ENCRYPTION_KEY is private and requires reflection).

PR #3900, opened 2023-10-25
Docker and Management Script Bugfixes

turtletowerz:
### What is the bug?
Currently the Docker container doesn't capture Ctrl+C while TML does, meaning that a user could lose data between the last autosave if they try to close the server with SIGINT

### How did you fix the bug?
Switched away from spawning a new process with `/bin/bash` in favor of the in-process replacement `./scriptname.sh`. This release also fixes the management script pulling pre-releases from Github

Not really, besides maybe switching to `exec` in the ENTRYPOINT


patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/manage-tModLoaderServer.sh
```diff
@@ -15,6 +15,20 @@ function popd {
 	command popd > /dev/null || return
 }
 
+function try_make_dir {
+	for dir in "$@"; do
+		if ! [[ -d "$dir" ]]; then
+			mkdir "$dir"
+		fi
+	done
```
based64god: is `mkdir -p` not acceptable for some reason?

turtletowerz: I didn't realize `-p` removes the errors as well, weird behavior but I'll convert all of them to use it. Thanks!

PR #3907, opened 2023-11-01
Creates IDs for dashes from Player.dashType

kurti256:
In similar vain to ItemUseStyleID
no changes to examplemod


JavidPack:
Thanks.

PR #3911, opened 2023-11-02
[Example Mod] - ExampleUpgradeableMinion

TacoConKvass:
Addition of an ExampleUpgradeableMinion.
Only one minion of this type is allowed at the same time per player. Summoning it again raises its damage like the Desert Tiger staff summon, could be extended to change other funcionality dependant on how many times the minion has been summoned, if needed.

No.


Rijam:
Make sure you are following the tModLoader style guide. (https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide) (https://github.com/tModLoader/tModLoader/blob/stable/.github/CONTRIBUTING.md)

I noticed you are using spaces when you should be using tabs and your curly braces should follow this style:
```
if (foo) {
	Bar();
}
```
Edit: You also forgot to commit the localization changes (item name, buff name, projectile name, etc.)

TacoConKvass:
All right, sorry for all of that will get to fixing that

JavidPack:
Can you update your PR to include more info about why this should go in ExampleMod? What will this teach about the API or showcase that the existing examples don't? I'm not seeing anything "upgradeable" about the minion, nor can I tell what that means.

TacoConKvass:
What I mean by upgradeable is that it behaves like Abigail's Flower summon. Instead of summoning more copies, it empowers the one that was already spawned. The example uses the Projectile.ai[0] for keeping track of how many times the item has been used, and it gradually takes up more and more minion slots.
The point of why I wanted to add it here, was because I couldn't find anything on how to do something like this

direwolf420:
The Amulet of Many Minions mod makes upgradeable minions, it calls them Empowered. It works like the tiger staff minion by summoning dummy projectiles which act as counters, whereas the actual damaging minion has no slots. This has the benefit of not requiring any hacks with CanUseItem, OnSpawn, etc. (Tho it does require hooking On_Player.FreeUpPetsAndMinions to prevent despawning of the damaging minion)

https://github.com/westphallm1/tModLoader_Minions/blob/1.4/Projectiles/Minions/MinonBaseClasses/EmpoweredMinion.cs#L14 contains both classes as mentioned above. An implementation is https://github.com/westphallm1/tModLoader_Minions/blob/1.4/Projectiles/Minions/EclipseHerald/EclipseHerald.cs

TacoConKvass:
The Amulet of Many Minions one seems more complex, but thank you for linking it.
If you see this addition as unnecessary, I am fine with that, the idea of adding it to EM was suggested to me by someone else, who as well seemed confused on how to do that. However, I think that it would be nice to have.

EDIT: Also, the OnSpawn stuff could be ommited by changing the damage calculation from using purely `Projectile.ai[0]` to `Projectile.ai[0] + 1f`

JavidPack:
I see, we'll have to look into this more. There might be something that could be done to streamline this type of minion especially since ModProjectiles can hold personalized data. I'd like a solution that is a bit less roundabout than what I'm seeing. Thanks for the PR, it'll be a bit before I have time to look closely at this.

One thing for sure that would be necessary is a visual indicator of the leveling up (the sprite changing a certain level intervals), just like the vanilla examples.

TacoConKvass:
I can add that when I come back home, no problem.
I will also clean up the AI code, because using Pre and PostAI is not really necessary there.

EDIT: Done

Issue #2705, opened 2022-07-10
Console not opening when starting multiplayer server


Vaiskii:
### Version

Latest 1.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When i try to start a multiplayer server from tmodloader and set the option "show server console" on, the console doesn't appear. Also tried to launch from tmodloader files ( file: start-tmodloaderserver ) console works there but crashes after downloading some mods.

### Log File

[client – kopio.log](https://github.com/tModLoader/tModLoader/files/9079893/client.kopio.log)

### Steps to reproduce

1. Start tmodloader
2. Multiplayer
3. Host and play
4. Choose character
5. Choose world
6. Start server ( show server console option on )

### Expected Behavior

Console should open and the server should start.

### Actual Behavior

My console doesn't open and tmodloader gets stuck in the "starting server..." text.
When starting from the tmodloader files, console opens but crashes after loading some mods.

### Reproduction frequency

Very often.

JavidPack:
Note to self: We might need to use the approaches in https://github.com/tModLoader/tModLoader/pull/3940 to restore this, but that fixes something that already worked on Windows only, so something specific to windows is also an issue.

PR #3920, opened 2023-11-06
Improve error logging and reporting

JavidPack:
Users have a hard time investigating errors to fix issues with their mod selections. Currently error messages hint to open up `client.log`, but only if the user is actively modding. Novice modders not using a debugger would see that hint, but would not know how to do that without reading the FAQ here on our wiki. Normal users would only see errors reported if they are thrown during mod loading.

This PR aims to improve error reporting to further facilitate users self-diagnosing issues.

# Repeated Errors
Current behavior is to report an error only once. This behavior hides the severity of some issues. Code has been added to track how many times a specific error has been encountered. Errors will be logged at 1, 10, and 100 repetitions, with 10 and 100 reports escalating to `Error` status and having an accompanying "The following exception has been repeated 100 times:" message. 
image

# ModExceptionDiagnostics window
As the `UIModNetDiagnostics` window suggests the "health" of a mod's networking usage, a `UIModExceptionDiagnostics` window would suggest the "health" of a mod's code. User can press F8 a few times to arrive at this menu, allowing them to quickly appraise the health of a mod. This window will show errors accumulating in real time, making it even better than using the `tail` command or constantly reloading the client.log file. This is especially true for errors that will be omitted from client.log because they are repeats. Users can open the window and repeat a gameplay action that seems to be buggy or laggy. They will be able to immediately see the error count for the buggy mod increasing, suggesting which mod is causing the unexpected behavior.

TODO: This window can be augmented with additional counters for "ignored" exceptions and for displaying the most recent exception or a listing of all exceptions, whatever is deemed useful.

https://github.com/tModLoader/tModLoader/assets/4522492/8ad58bcd-f4e1-43bc-81e0-170b917b6737

# Open up logs folder directly (TODO)
Users don't know hot to get to client.log. We should provide a way to quickly open the logs folder or the specific log file.

One option is to for clicking on the "see client.log for full trace" message to open client.log directly.
image

Another option is adding a 📜 button on the main menu to open the logs folder.
image

# Ignored Exceptions (TODO)
We allow modders to silence specific exceptions via `Logging.IgnoreExceptionSource/IgnoreExceptionContents`. This is useful for handling known exceptions, but it could be misused by modders, preventing effective troubleshooting by users.

At the very least, it would be useful to log these additions: "System.InvalidCastException added to exception ignore list by ExampleMod". It might be useful to provide a "verbose" mode via a command line parameter that would ignore all of these filters, just in case a specific bug is not being reported properly in logs. Another thought is to track these exceptions and report them in the ModExceptionDiagnostics window, explained in another section.

# Console
The console is not visible for normal client gameplay after launching, but it would be visible when doing host and play once #2705 is fixed. It is visible when debugging tModLoader itself as well. The colors can be adjusted.

Colors have been added for info/debug/warn/error: 
image

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2351,7 +2351,7 @@
  		UpdateViewZoomKeys();
  		PlayerInput.SetZoom_Unscaled();
  		UILinkPointNavigator.Update();
-@@ -14667,7 +_,20 @@
+@@ -14667,7 +_,25 @@
  		if (keyState.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.F8) && !drawingPlayerChat && !editSign && !editChest) {
  			if (netRelease) {
  				SoundEngine.PlaySound(12);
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
I would like to see this complimented by several user stories to better understand the situations, and evaluate the effectiveness of the intended solution. Who is the target audience for these improvements? Will these changes cause confusion or problems for players with negligible modding experience?

Clearly good ideas:
- Click on a first chance exception during main menu to see the log
- Logging errors repeatedly every order of magnitude

Probably helpful:
- Console text coloring.
  - Recommend keeping `DEBUG` as white, because we don't have any consistent rule dividing DEBUG and INFO
- Showing silently caught exceptions to regular users
  - Need to direct them to logs and troubleshooting resources.
  - Consider a way to show them in the main menu as well?
  
Probably not useful:
- Logging ignored exception filters, or a way to log all silent exceptions, even ignored ones.
  - This information would be difficult to interpret, and in almost all situations a debugger or profiler would be needed to correctly determine the issue.
- Logs icon on main menu
  - I don't think anyone will notice it. A button in the mods menu could be useful though.
- The `ModExceptionDiagnosticsUI`
  - The main issue here is that it's impossible to correctly attribute an exception to a mod, and misattribution may cause more harm than good. The only thing a user can do with the information in that window is to open the logs, or complain to a mod author. 

It seems like this PR is aimed at silently caught exceptions during gameplay. I think we should start by simply surfacing errors to users, and getting reports with logs. Then we can continue with more guided troubleshooting, like identifying all mods and IL hooks involved in a stack trace.

PR #3922, opened 2023-11-08
Fix issue with bad TileObjectData code causing "leaks" into other TileObjectData instances.

JavidPack:
Over the years there have been mysterious reports of incorrectly behaving tiles. These reports usually involve tiles untouched by the mod's code to be adversely affected.

Most recently, this has been reported with TheDepths mod. Somehow tiles that previously looked correct are now incorrect:

image

These issues are caused by improper ordering of `TileObjectData` methods and properties. Assigning properties after `AddTile` or before `CopyFrom` leaves the `TileObjectData` in an improper state. Annoyingly, these issues are sometimes caused by incorrect code in the ModTile loading immediately before the 1st affected ModTile, leading to bugs that are hard to track down.

In this case, `TileObjectData.newTile.DrawYOffset = -2;` was written after `TileObjectData.addTile(Type);` in a ModTile, causing the next ModTile to exhibit the effects of the error. 

This PR fixes the issue by resetting the affected fields in `CopyFrom`. This PR also throws errors in `CopyFrom` to instruct the modder that their code is not doing what they think it is. 

# Thoughts or Discussion
Alternatively, this issue could be fixed with resetting the affected fields in `CopyFrom` only, but I feel that not throwing errors will hamper  productivity and lead to modders confused about why their code is not affecting the behavior of their tile.

One concern is that the `ModCompile.activelyModding` check will prevent other mods from loading when debugging a working mod, but I think that the preview period should help with that.

### DIFF
--- a/patches/tModLoader/Terraria/ObjectData/TileObjectData.cs.patch
+++ b/patches/tModLoader/Terraria/ObjectData/TileObjectData.cs.patch
@@ -97,6 +97,24 @@
  		get {
  			if (_subTiles == null)
  				return _baseObject.SubTiles;
+@@ -1259,6 +_,17 @@
+ 			_tileObjectDraw = copy._tileObjectDraw;
+ 			_tileObjectStyle = copy._tileObjectStyle;
+ 			_tileObjectCoords = copy._tileObjectCoords;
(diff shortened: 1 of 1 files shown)


JavidPack:
I changed the exception to a log message. I'm not confident the exception wouldn't cause issues for at least someone.

PR #3928, opened 2023-11-14
Tile Definition

Destructor-Ben:
### What is the new feature?
Added a tile definition.
image

Currently there is no built in tile definition, likely because drawing previews for tiles is kinda hard.

Could also draw frame important tiles properly, idk if it's really required though (we need tile styles at some point anyway so code might get refactored).

Tile styles could be handled but would likely do that in conjunction with negative npc ids.

Also, tile ids start from 0 instead of 1, unlike other ids, so there might be an edge case I haven't handled yet (so far it seems to work).

It may also pay to change the display name to the internal name instead, since things like trophies and metal bars have a generic map name, which can be confusing when searched for. Could also search for display name and internal name.

### Sample usage for the new feature`
In a config:
```cs
TileDefinition definition = new TileDefinition(TileID.Dirt);
```

### ExampleMod updates

Added TileDefinition to Showcase A Datatypes.


JavidPack:
Can you add a screenshot of this? Just so we have an idea of how the tiles are being drawn.

JavidPack:
Here is the final look. There are some small gaps in some tiles, likely due to some math error, but it's not too noticeable and good enough for now. 

Tile styles cycle if the TOD has subtiles, so not every style will be shown unfortunately. No way around that. Could be disabled if deemed distracting. 

https://github.com/tModLoader/tModLoader/assets/4522492/b09bebbb-0e70-4629-9c6c-1f0f7a7f7303

JavidPack:
Thanks a bunch for spearheading this. Should be useful.

PR #3939, opened 2023-11-21
Make switching back from Proton less error prone

JavidPack:
There have been many errors in the past attributed to Proton. I was not able to replicate any of them at this time personally, but it is quite common in the Discord support to suggest to Linux users to disable Proton. When disabling Proton, we also have to tell the user to delete the `dotnet` folder. Failure to do so results in the game not launching, suggesting to the user that the fix doesn't work.

This PR ensures that the dotnet folder will be deleted automatically if `dotnet.exe` file is detected on non-Windows or if `dotnet` file is detected on Windows. This change allows a Linux user to seamlessly switch between enabling and disabling Proton. A warning has been added and the wiki updated (https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Usage-FAQ#disable-proton) with information on moving saves from the Proton prefix saves folder to the native saves folder.

Issues with Proton that I was able to confirm today:
- Manually launching, such as manually launching a dedicated server, will use a different saves folder. This is confusing.
- Manually launching will also require the native `dotnet` executable. With this fix it will delete and reinstall dotnet each time, without this fix the game just won't launch. (Users will see a `./LaunchUtils/ScriptCaller.sh: line 95: /run/media/mmcblk0p1/steamapps/common/tModLoader/dotnet/6.0.14/dotnet: No such file or directory` error in Launch.log because `dotnet.exe` exists but `dotnet` doesn't.)
- Saves will be located deep in the file tree in a place unfamiliar to our FAQ and instructions, `/home/[username]/.local/share/Steam/steamapps/compatdata/1281930/pfx/drive_c/users/steamuser/Documents/My Games/Terraria/tModLoader`, making support confusing.
- Other issues have been brought up in the past, but I was unable to confirm.

# Discussion
1. We could force detecting Proton to not launch, but I think some Linux users insist that it is needed for their system. This might be a result of the `dotnet` folder not being deleted during testing, in my opinion, but it might have some merit
2. `WINDOWS_MAJOR` set to `0` and `WINEHOMEDIR` existing were the environment variables I found consistently set on my Linux machine (Steam Deck), but it is possible they might not be 100% reliable. I did not test launching through Wine directly, just through Proton. 
3. Automated porting from Prefix folder. Pretty annoying, but I think the wiki update is sufficient.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
@@ -34,6 +34,10 @@ if [[ "$_uname" == *"_NT"* ]]; then
 	fi
 fi
 
+if [[ "$WINDOWS_MAJOR" == "0" || ! -z "$WINEHOMEDIR" ]]; then
+	echo "Proton has been detected. It is highly recommended to not use it as it causes all manner of issues. Please disable Proton and launch again. See https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Usage-FAQ#disable-proton for information on moving save data to the correct location." 2>&1 | tee -a "$LogFile"
+fi
+
 echo "Verifying .NET...."  2>&1 | tee -a "$LogFile"
 echo "This may take a few moments."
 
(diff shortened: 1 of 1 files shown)


Chicken-Bones (approved):

fishiatee:
Any particular reasons why Proton is discouraged? Direct3D 11's performance is way better than OpenGL so I think using dxvk might be beneficial too.

JavidPack:
> Any particular reasons why Proton is discouraged? Direct3D 11's performance is way better than OpenGL so I think using dxvk might be beneficial too.

I haven't noticed any performance differences, but my device was always running at stable 60fps so I wouldn't know. I'm not aware of anyone verifying this claim with tModLoader, but maybe it is something to think about. I know linux users can use `/gldevice:Vulkan` to use Vulkan directly if they need to, I'm not sure how that affects performance.

As for why proton is discouraged, I mentioned it in the post. It has been the source of many hard to debug bugs in the past and the save file structure leads to incompatibilities with hosting a server.

With this PR it'll be easier to A/B test, so maybe we can learn some new insights later on if there is a merit to Proton.

Solxanich (approved):
Approved, but let's ensure we submit a deficiency summary to Proton official development so they can experiment on why it isn't working, ensuring we emphasize non-critical since we have a valid Linux native build.

Who knows what other games may be affected

PR #3940, opened 2023-11-22
Enable tModPorter on Linux from develop mods menu

### LINKED ISSUE #3933: tModPorter does not appear on non-Windows
### Version

Latest 1.4.4-stable

### OS

Mac

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

So, I have a mod that hasn't updated since 1.4.3 and was attempting to run tModPorter, however when I followed the guide I could not find any tModPorter option. When I went to the discord to trouble shoot, I found that the ! option that was supposed to appear did not appear and was recommended to report a bug.

### Log File

Screenshot 2023-11-16 at 2 44 17 PM
Screenshot 2023-11-16 at 2 44 03 PM

### Steps to reproduce

Boot up tModLoader and attempt to run tModPorter

### Expected Behavior

tModPorter should appear

### Actual Behavior

no tModPorter option

### Reproduction frequency

Every time I boot up tModLoader

### Additional Information

#mod-programming-help thread is where I was discussing it, discussion started 16 minutes ago

JavidPack:
Fixes #3933

tModPorter.sh changed so that it works with `DOTNET_ROLL_FORWARD=Disable` set, as it would be if launched from in-game. `DOTNET_ROOT` is also checked just in case `dotnet` not on `PATH`, such as it would be on a Steam Deck. A console window still does not appear, so a log file was added.

tModPorter has a little fix for an input issue with no console present, and an update to Microsoft.Build.Locator version to take advantage of fixes to locating dotnet. The old code could not locate dotnet in DOTNET_ROOT, for example. 

Additionally, `GetPossibleSystemDotnetPaths` has been updated with `~/.dotnet` as a possible path because that is where the install script will install dotnet to.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
@@ -249,7 +249,6 @@ internal class UIModSourceItem : UIPanel
 
 					string args = $"\"{csprojFile}\"";
 					var tMLPath = Path.GetFileName(Assembly.GetExecutingAssembly().Location);
-					// TODO: We need to find some way of launching Linux scripts in a console window that is shown, not hidden. Probably requires logic to call "gnome-terminal/xterm/konsole -e command" depending on whatever is installed (https://askubuntu.com/a/46630)
 					var porterPath =  Path.Combine(Path.GetDirectoryName(tMLPath), "tModPorter", (Platform.IsWindows ? "tModPorter.bat" : "tModPorter.sh"));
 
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
This is all good work. However, we need to make sure tModPorter always appears in another console window. It may take a couple of minutes to run on some projects, and users need to be informed of progress and any errors.

JavidPack:
Code has been updated to launch a terminal if it detects it not being in a terminal. Works when launched directly, from a terminal, and from in-game.

We'll need someone to test on Mac eventually, not sure what currently happens there, that can be fixed later if needed.

PR #3950, opened 2023-11-27
Added ContainerID and ContainerID2 for chest subtypes

JeseGamerHD:
### What is the new feature?
I got tired of checking the internal id for different chest subtypes so I created an ID. This should also be a bit clearer compared to a number: `ContainerID.FrozenChest` versus using just `11`. Since there are two containers `TileID.Containers` and `TileID.Containers2` I made two classes: `ContainerID` and `ContainerID2`.
To reduce potential confusion with "magical numbers". Also not having to look up the internal id from wiki or from the spritesheets is nice.
The naming could maybe be something other than ContainerID like ChestID for example. Other than that I'm not sure.
Checking if a tile is a Sandstone Chest
`if (chestTile.TileType == TileID.Containers2 && chestTile.TileFrameX == ContainerID2.SandstoneChest * 36) { }`

### ExampleMod updates
Changed `ChestItemWorldGen.cs` to use the new ID and updated the comments.


patches/tModLoader/Terraria/ID/ContainerID.cs
```diff
@@ -0,0 +1,83 @@
...
+	public const int LesionChest = 3;
+	public const int DeadMansChest = 4;
+	public const int SolarChest = 5;
+	public const int VortexChest = 6;
+	public const int NebulaChest = 7;
+	public const int StardustChest = 8;
+	public const int GolfChest = 9;
+	public const int SandstoneChest = 10;
+	public const int BambooChest = 11;
+	public const int DesertChest = 12;
```
JavidPack, line 81: This won't compile. Typo? (DesertChest defined twice.)

JeseGamerHD: Forgot to add Locked in front of it (fixed now). I guess I didn't open the project correctly since the IDE didn't complain. Also, would it be better to multiply the values by 36 (chest width) in advance in the ID classes?

JavidPack: No, these represent the styles, not the coordinates of the top left tile.

PR #3955, opened 2023-12-01
GlobalNPC.SetDefaultsFromNetId

### LINKED ISSUE #3819: `npc.netID` and negative ID variant stats are not set before `NPCLoader.SetDefaults` is run
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

tryin' to check against `netID` in `GlobalNPC.SetDefaults` will not work, as `netID` is set *after* all `GlobalNPC.SetDefaults` hooks are run, preventin' mods from (sensibly) modifyin' stats or addin' needed new data for negative-ID variants of enemies, such as Pinky or the like 50 zombie variants that exist in the game

in the same vein, the intended stats for these variants *also* aren't actually set until after `GlobalNPC.SetDefaults` is run, meanin' that modders can't correctly modify the intended statistics for these enemies in `SetDefaults` like they can with others

### Log File

not needed

### Steps to reproduce

1. make a `GlobalNPC`
2. in the `SetDefaults` hook, check for `netID` before doin' anything, and return if not that `netID`
3. load up a world and watch as your global doesn't apply to the intended variant

ThomasThePencil:
in summation of the followin': `netID` is cringe

### What is the new feature?
a new `GlobalNPC.SetDefaultsFromNetId` hook which allows certain vanilla "variant" enemies that otherwise can't be touched by `SetDefaults` to be modified by mods

`netID` is this funny little vanil- actually, wait a minute. why am I explainin' this again? I opened an issue on it

as described in #3819, NPCs with negative `netID` (a vanilla NPC field which, despite its name, has nothing to do with netcode) can't be modified (correctly) via the `GlobalNPC.SetDefaults` hook

this new hook works around the problem (and technically closes #3819) by addin' a new hook which DOES account for `netID` bein' set, allowin' modders to actually modify these NPC variants in a way that's actually sustainable

multiple! I have a thread open in the actual tML discord on this for people who wish to help find one that works best

(ideally, `netID` wouldn't exist and we would just have positive IDs for all these NPC variants, but that's a completely different can of worms that I'm not in the mood to open)

```csharp
public override void SetDefaultsFromNetId(NPC npc)
{
    // suppose you wanna mess around with Pinky. here's how to do it
    if (npc.netID != NPCID.Pinky)
        return; // do nothing

    // do something. easy as you can get
}
```

### ExampleMod updates
none, but can add one if desired

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
@@ -34,9 +34,10 @@ public abstract class GlobalNPC : GlobalType<NPC, GlobalNPC>
 	public sealed override void SetupContent() => SetStaticDefaults();
 
 	/// <summary>
+	/// Similar to SetDefaults, but only called on NPCs with a negative <see cref="NPC.netID"/>.<br/>
 	/// Vanilla "variant" NPCs which have <see cref="NPC.netID"/> set to a negative value cannot be properly modified through SetDefaults.<br/>
 	/// This method allows you to properly run code you would run in SetDefaults for these NPCs.<br/>
-	/// When running this method, always be sure to check against <see cref="NPC.netID"/> to ensure you're modifying the correct variant.<br/>
+	/// <b><see cref="NPC.type"/> does not support negative numbers! Always be sure to check against <see cref="NPC.netID"/> instead.</b><br/>
 	/// </summary>
 	public virtual void SetDefaultsFromNetId(NPC npc)
 	{
--- a/patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
@@ -34,6 +34,15 @@ public abstract class GlobalNPC : GlobalType<NPC, GlobalNPC>
 	public sealed override void SetupContent() => SetStaticDefaults();
 
 	/// <summary>
+	/// Vanilla "variant" NPCs which have <see cref="NPC.netID"/> set to a negative value cannot be properly modified through SetDefaults.<br/>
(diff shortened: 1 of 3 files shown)


Chicken-Bones:
Call it `SetDefaultsFromNetId`, then it's approved

ThomasThePencil:
changed as requested
awaitin' any further comments, or a merge if there are none

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -33,6 +33,16 @@ protected sealed override void Register()
 
 	public sealed override void SetupContent() => SetStaticDefaults();
 
+	/// <summary>
+	/// Similar to SetDefaults, but only called on NPCs with a negative <see cref="NPC.netID"/>.<br/>
+	/// Vanilla "variant" NPCs which have <see cref="NPC.netID"/> set to a negative value cannot be properly modified through SetDefaults.<br/>
+	/// This method allows you to properly run code you would run in SetDefaults for these NPCs.<br/>
+	/// <b><see cref="NPC.type"/> does not support negative numbers! Always be sure to check against <see cref="NPC.netID"/> instead.</b><br/>
+	/// </summary>
```
Chicken-Bones:
```suggestion
	/// <summary>
	/// Called after SetDefaults for NPCs with a negative <see cref="NPC.netID"/><br/>
	/// This hook is required because <see cref="NPC.SetDefaultsFromNetId"/> only sets <see cref="NPC.netID"/> after SetDefaults<br/>
	/// Remember that <see cref="NPC.type"/> does not support negative numbers and AppliesToEntity cannot distinguish between NPCs with the same type but differet netID<br/>
	/// </summary>
```

PR #3958, opened 2023-12-02
Add docs to ModTile.MineResist and ModTile.MinPick

Destructor-Ben:
Adds a reference to the newly updated wiki which lists links to vanilla wiki and a table containing vanilla MineResist values,


JavidPack:
Thanks.

PR #3965, opened 2023-12-06
Reimplemented custom Terraria.csproj logic using C# MSBuild Tasks, fixed ".Remove(-1)".

Mirsario:
### This PR has the following goals:
- Fix the `The expression """.Remove(-1)" cannot be evaluated.` build-time error, which has haunted TML devs & contributors for quite a while. I was going to work on a completely different PR, but couldn't, due to this hard-to-debug & random-to-resolve issue.
- Maximize readability of our main C# project file.
- Provide a simple way of introducing any other custom logic into our build pipelines for the future.

### This is all achieved by:
- Introducing a `netstandard2.0` project (currently `./tModBuildTasks/tModLoader.BuildTasks.csproj`) containing MSBuild tasks written in C#. The project is automatically built prior to `TML`/`TerrariaNetCore`.
- Rewriting contents of most of our "Targets" as C# Tasks, getting rid of XML tags that invoked .NET methods in the most unreadable way known to mankind. XML, the programming language.
- Prettifying `Terraria.csproj` with organizational & documentational comments, and a sense of space (in TerrariaNetCore & tModLoader steps).

### Checklist:
- [x] `TerrariaNetCore` builds & runs.
- [x] `tModLoader` builds & runs.
- [x] Confirmed to have no impact on the `Libraries` directory's layout.
- [x] Confirmed `TaskHostFactory`'s stability (MSBuild shouldn't keep a file handle on `tModLoader.BuildTasks.dll`).
    - [x] Works on my machine™️.
    - [x] Others' machines & IDEs.

Review one commit at a time, the sum diff isn't pretty.

### DIFF
--- a/patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
+++ b/patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
@@ -1,11 +1,13 @@
 --- src/Terraria/Terraria/Terraria.csproj
 +++ src/TerrariaNetCore/Terraria/Terraria.csproj
-@@ -1,90 +_,116 @@
+@@ -1,90 +_,118 @@
  <Project Sdk="Microsoft.NET.Sdk">
++
  	<Import Project="../Configuration.targets" />
  	<Import Project="../../WorkspaceInfo.targets" />
 +	<Import Project="../../../tModBuildTasks/BuildTasks.targets" />
 +
++	<!-- General -->
  	<PropertyGroup>
 -		<OutputType>WinExe</OutputType>
 +		<OutputType>Exe</OutputType>
@@ -15,29 +17,33 @@
  		<RootNamespace>Terraria</RootNamespace>
 -		<OutputName>Terraria</OutputName>
 +		<AssemblyName>Terraria</AssemblyName>
+-	</PropertyGroup>
+-	<PropertyGroup Condition="$(Configuration.Contains('Server'))">
+-		<OutputType>Exe</OutputType>
+-		<OutputName>$(OutputName)Server</OutputName>
+-	</PropertyGroup>
+-	<PropertyGroup Condition="$(Configuration.Contains('Debug'))">
+-		<OutputName>$(OutputName)Debug</OutputName>
+-	</PropertyGroup>
+-	<PropertyGroup>
+-		<PdbFile>bin\$(OutputName)</PdbFile>
+ 		<ApplicationIcon>Icon.ico</ApplicationIcon>
 +
++		<!-- Avoid overwriting Terraria.exe (if it's not Debug it's release) -->
++		<AssemblyName Condition="$(Configuration.Contains('Debug'))">$(AssemblyName)Debug</AssemblyName>
++		<AssemblyName Condition="!$(Configuration.Contains('Debug'))">$(AssemblyName)Release</AssemblyName>
++		<OutputName>$(AssemblyName)</OutputName>
+ 	</PropertyGroup>
++
++	<!-- Build Configuration -->
++	<PropertyGroup>
++		<LibrariesDirectory>Libraries</LibrariesDirectory>
 +		<GenerateRuntimeConfigDevFile>true</GenerateRuntimeConfigDevFile>
 +		<CopyDebugSymbolFilesFromPackages>true</CopyDebugSymbolFilesFromPackages>
 +		<CopyDocumentationFilesFromPackages>true</CopyDocumentationFilesFromPackages>
- 	</PropertyGroup>
- 	<PropertyGroup Condition="$(Configuration.Contains('Server'))">
--		<OutputType>Exe</OutputType>
--		<OutputName>$(OutputName)Server</OutputName>
-+		<AssemblyName>$(AssemblyName)Server</AssemblyName>
- 	</PropertyGroup>
- 	<PropertyGroup Condition="$(Configuration.Contains('Debug'))">
--		<OutputName>$(OutputName)Debug</OutputName>
-+		<AssemblyName>$(AssemblyName)Debug</AssemblyName>
 +	</PropertyGroup>
-+	<!-- Avoid overwriting Terraria(Server).exe (if it's not Debug it's release) -->
-+	<PropertyGroup Condition="!$(Configuration.Contains('Debug'))">
-+		<AssemblyName>$(AssemblyName)Release</AssemblyName>
- 	</PropertyGroup>
- 	<PropertyGroup>
--		<PdbFile>bin\$(OutputName)</PdbFile>
-+		<OutputName>$(AssemblyName)</OutputName>
- 		<ApplicationIcon>Icon.ico</ApplicationIcon>
- 	</PropertyGroup>
++
++	<!-- References -->
  	<ItemGroup>
 +		<ProjectReference Include="../../../FNA/FNA.Core.csproj" />
  		<ProjectReference Include="../ReLogic/ReLogic.csproj" />
@@ -45,6 +51,7 @@
  			<Link>Libraries/ReLogic/ReLogic.dll</Link>
  			<LogicalName>Terraria.Libraries.ReLogic.ReLogic.dll</LogicalName>
  		</EmbeddedResource>
++
  		<Reference Include="CsvHelper" />
 -		<Reference Include="FNA" Condition="$(DefineConstants.Contains('FNA'))" />
  		<Reference Include="Ionic.Zip.CF" />
@@ -72,7 +79,11 @@
 -			<HintPath Condition="$(DefineConstants.Contains('Mono'))">Libraries/Mono/System.Windows.Forms.dll</HintPath>
 -		</Reference>
 -		<Reference Include="WindowsBase" />
++
++		<PackageReference Include="Steamworks.NET" Version="20.1.0" />
  	</ItemGroup>
++
++	<!-- Embedded Resources -->
  	<ItemGroup>
  		<EmbeddedResource Include="GameContent/Creative/Content/*" />
  		<EmbeddedResource Include="GameContent/Metadata/MaterialData/*" />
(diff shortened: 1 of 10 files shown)


tModBuildTasks/OrganizeReferenceDestinations.cs
```diff
@@ -0,0 +1,82 @@
...
+			}
+			// Direct Managed References
+			else if (referenceSourceTarget == "ResolveAssemblyReference" && assemblyName != null) {
+				destinationSubDirectory = $"{BaseDirectory}/{assemblyName.Name}/{assemblyName.Version}/";
+			}
+			// NuGet Packages
+			else if (!string.IsNullOrEmpty(nugetPackageId)) {
+				// This is used for all NuGet libraries, whether native or managed, whether rid-specific or agnostic.
+				if (!string.IsNullOrEmpty(pathInPackage)) {
+					destinationSubDirectory = $"{Path.GetDirectoryName($"{BaseDirectory}/{nugetPackageId}/{nugetPackageVersion}/{pathInPackage}").Replace('\\', '/')}/";
```
Chicken-Bones: This is unusual. Why is `pathInPackage` not concatenated on the end of the standard path (from the `else` clause below)?
In addition, is there a reason to explicitly use `/` separators rather than `Path.Combine`?

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
```diff
@@ -172,16 +124,59 @@
...
++
++		<Message Text="Copying $(AssemblyName) to '$(_ActualOutputDirectory)'..." Importance="high" />
++		<Copy SourceFiles="@(BinFiles)" DestinationFolder="$(_ActualOutputDirectory)/%(RecursiveDir)" SkipUnchangedFiles="True" />
++
++		<!-- Copy the Libraries folder, purging excessive files -->
++		<SynchronizeDirectories Source="$(TargetDir)Libraries" Destination="$(_ActualOutputDirectory)/Libraries" />
++	</Target>
++
++	<!-- Organize output copies of Project, NuGet, and directly referenced libraries under a single tidy folder. -->
++	<Target Name="OrganizeAllReferences" AfterTargets="ResolveAssemblyReferences" BeforeTargets="GenerateBuildDependencyFile;CopyFilesToOutputDirectory">
```
Chicken-Bones, line 134: Targets should be defined in the order they will be executed to reduce confusion for readers

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
```diff
@@ -66,7 +79,11 @@
 -			<HintPath Condition="$(DefineConstants.Contains('Mono'))">Libraries/Mono/System.Windows.Forms.dll</HintPath>
 -		</Reference>
 -		<Reference Include="WindowsBase" />
++
++		<PackageReference Include="Steamworks.NET" Version="20.1.0" />
  	</ItemGroup>
++
++	<!-- Embedded Resources -->
```
Chicken-Bones, line 86: Please avoid reorganizing the project file too much. We still need to minimize diffs. It's not necessary to add comments for each `ItemGroup` or `PropertyGroup` imo

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
```diff
@@ -172,16 +124,59 @@
...
+-	</Target>
++
++	<!-- Import tasks -->
++	<UsingTask TaskName="SynchronizeDirectories" AssemblyFile="$(BuildTasksAssemblyFile)" TaskFactory="$(BuildTasksTaskFactory)" />
++	<UsingTask TaskName="OrganizeReferenceDestinations" AssemblyFile="$(BuildTasksAssemblyFile)" TaskFactory="$(BuildTasksTaskFactory)" />
++
++	<!-- Installs the build's output into the relevant game folder. -->
++	<Target Name="CopyToActualOutputDirectory" AfterTargets="Build">
++		<PropertyGroup>
++			<_ActualOutputDirectory>$(TerrariaSteamPath)</_ActualOutputDirectory>
```
Chicken-Bones: Is there any reason to define this property? Can we replace it with `TerrariaSteamPath`?

tModBuildTasks/SynchronizeDirectories.cs
```diff
@@ -0,0 +1,95 @@
...
+		else {
+			UnixImplementation();
+		}
+	}
+
+	private void WindowsImplementation()
+	{
+		// Robocopy is always present.
+		ExecuteAndWait(
+			"robocopy", @$"""{Source}"" ""{Destination}"" /MIR",
```
Chicken-Bones: Since we have .NET at our hands, I'd prefer to use the file API and do it ourselves rather than using `robocopy` or `rsync`

Chicken-Bones (changes requested):

tModBuildTasks/SynchronizeDirectories.cs
Mirsario: I've actually written a fully managed implementation for this task prior, but no matter to the presence of metadata & SIMD'ed equality checks together with light multi-threading - it ended up feeling about 15 times slower than robocopy and rsync, so I didn't even bother including that as an rsync-less fallback. .NET doesn't seem very good when it comes to streams, but check if you see anything terrible here:

[SynchronizeDirectoriesBuildTask.zip](https://github.com/tModLoader/tModLoader/files/13635951/SynchronizeDirectoriesBuildTask.zip)

Could skip full integrity checks completely I guess, but I believe that our File.Copy calls will still be slower than robcopy's.

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
Mirsario: Explicitness, and to avoid making 3 edits in TML .csproj, for that's how many times `_ActualOutputDirectory` is referenced. This was present in the original.

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
Mirsario: Unlike in the ancient PR, this is done only on top level, but still makes an insane difference for navigating in the clunky format that is XML. 
image

I don't agree that we should always be choosing patch minimization *in a `.csproj`* over providing the patcher and ourselves more context, making things easier to repair.

tModBuildTasks/OrganizeReferenceDestinations.cs
Mirsario: Goofy if-else, yes, addressed now. For the latter - I was previously dealing with overwriting the `PathInPackage` variable, which used forward slashes, but that turned out unnecessary. I've switched to Path.Combine, but you'll find it taking twice the space now.

tModBuildTasks/SynchronizeDirectories.cs
Chicken-Bones: Neither `rsync` nor `robocopy` compare file contents, only sizes and timestamps by default. The `DirectoryInfo` and `FileInfo` API provides timestamps and file length info, hopefully in a performant manner.

tModBuildTasks/OrganizeReferenceDestinations.cs
```diff
@@ -54,21 +54,16 @@ protected override void Run()
...
 			// NuGet Packages
 			else if (!string.IsNullOrEmpty(nugetPackageId)) {
 				// This is used for all NuGet libraries, whether native or managed, whether rid-specific or agnostic.
-				if (!string.IsNullOrEmpty(pathInPackage)) {
-					destinationSubDirectory = $"{Path.GetDirectoryName($"{BaseDirectory}/{nugetPackageId}/{nugetPackageVersion}/{pathInPackage}").Replace('\\', '/')}/";
-				}
-				else {
-					destinationSubDirectory = $"{BaseDirectory}/{nugetPackageId}/{nugetPackageVersion}/";
-				}
+				destinationSubDirectory = Path.Combine(BaseDirectory, nugetPackageId, nugetPackageVersion, Path.GetDirectoryName(pathInPackage)) + Path.DirectorySeparatorChar;
```
Chicken-Bones: The need to add trailing slashes is a bit annoying, but understandable. Can you double check that it's required?

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
Chicken-Bones: On reflection and further review of the diffs, I believe the changes are acceptable. Although I'd prefer if the `Debug/Release/Server` assembly name diffs were more similar.

tModBuildTasks/OrganizeReferenceDestinations.cs
Mirsario: Removing them broke everything. Replaced the 3 concats with an explicit one.
image

patches/TerrariaNetCore/Terraria/Terraria.csproj.patch
Mirsario: Debug&Release conditions are inserted at `Terraria`, but `Server` comes from `decompiled`. It should take three minutes to update `Terraria.ModLoader.Setup.DecompileTask` to write a csproj closer to this one, minimizing diffs for every step, but that's better off done for 1.4.5/1.5 to not require redecompilations.

tModBuildTasks/SynchronizeDirectories.cs
Mirsario: Pushed the above with integrity checks removed, and some other edits. Seems to average just 10ms now, uh oh.

tModBuildTasks/IOUtils.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		return path.Substring(relativeTo.Length + (endsWithSeparator ? 0 : 1));
+	}
+
+	/// <summary>
+	/// Returns true if both files exist and have the same length. Does not check file contents.
+	/// </summary>
+	public static bool AreFilesSeeminglyTheSame(FileInfo fileA, FileInfo fileB)
+	{
+		return fileA.Exists && fileB.Exists
```
Chicken-Bones, line 30: Gotta check `LastModified`

tModBuildTasks/SynchronizeDirectories.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		destination.Create();
+
+		// Delete files that shouldn't be there.
+		Parallel.ForEach(destination.EnumerateFiles("*", SearchOption.AllDirectories), file => {
+			string relativePath = IOUtils.SubstringToRelativePath(file.FullName, Destination);
+			string sourcePath = Path.Combine(Source, relativePath);
+
+			if (!File.Exists(sourcePath)) {
+				IOActionWrapper(file.Delete);
```
Chicken-Bones: Does this need `FileNotFoundException` checking? Like the `DirectoryNotFoundException` one below? I assume it's a race condition fix.

tModBuildTasks/SynchronizeDirectories.cs
```diff
@@ -0,0 +1,94 @@
...
+				IOActionWrapper(() => {
+					Directory.CreateDirectory(destinationFile.DirectoryName);
+					sourceFile.CopyTo(destinationFile.FullName, overwrite: true);
+				});
+			}
+		});
+	}
+
+	// Runs an action with a fixed amount of retries for the case of IOExceptions occuring from third-party short-lived file locks.
+	private void IOActionWrapper(Action action)
```
Chicken-Bones: Recommend:
- `RunIOActionWithRetries`
- `DelayMs` of `100`, or a sequence of `100, 200, 400, 800, 1600`

Chicken-Bones (changes requested):

tModBuildTasks/SynchronizeDirectories.cs
Mirsario: Would only prevent user-caused races in case of files, but doesn't hurt to have.

Chicken-Bones (approved):

Issue #3991, opened 2023-12-25
Incomplete Shimmer Decrafting Support for Bottle Recipes and Recipe.ConsumeItemCallback


absoluteAquarian:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

Part of the logic within `Terraria.Item::GetShimmered()` handles decrafting the item based on its decrafting recipe, if one exists.

Within this logic, the game checks `recipe.alchemy` in order to conditionally reduce the stack of the ingredients produced by the decrafting process.  
However, since `recipe.alchemy` was replaced by a `Recipe.ConsumeItemCallback` delegate, this field is never `true`.  Thus, the decrafting never reduces the stack for the ingredients in a recipe that uses the Bottle as a crafting station.

Also, for any modded recipes that use `recipe.AddConsumeItemCallback()`, the callbacks are never invoked when decrafting.

### Log File

n/a

### Steps to reproduce

1. Throw an item that uses the Bottle as a crafting station, such as any potion, or any recipe that uses `Recipe.ConsumeItemCallback` callbacks
2. Observe that the full stacks for the ingredients are spawned, even though they shouldn't be

### Expected Behavior

The vanilla behavior should be preserved (recipes using the Bottle have their ingredient stacks reduced).

### Actual Behavior

The ingredients always have their full stacks, meaning they don't get conditionally reduced.

### Reproduction frequency

100%

### Additional Information

The `if (recipe.alchemy)` check — and the subsequent loop in its scope — should be removed in favor of calling `RecipeLoader.ConsumeItem()`.  Alternatively, a call to `RecipeLoader.ConsumeItem()` could be added after the `if (recipe.alchemy)` block like how the logic for `Recipe::Create()` was modified.

absoluteAquarian:
Duplicate of #3722

Issue #3722, opened 2023-08-13
When throwing potions into shimmer it gives all the ingredients back when its supposed to only give 2/3 of the ingredients


Vesite:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When throwing potions into shimmer it gives all the ingredients back (Its supposed to give 2/3 of the ingredients on average so that you can't make infinite potions using the alchemy table). Happens even when I don't have any mods enabled. 1.4.4.9
In vanilla it gives the correct 2/3 Ingredients back.

### Log File

[20:43:19.091] [Main Thread/INFO] [tML]: Starting tModLoader client 1.4.4.9+2023.06.25.31|2023.06|stable|Stable|1cf5af1a24a961fcbe0d1d7fc31dcda160496504|5249959922336091465 built 8/12/2023 12:39 AM
[20:43:19.098] [Main Thread/INFO] [tML]: Log date: 8/13/2023
[20:43:19.098] [Main Thread/INFO] [tML]: Running on Windows (v10.0.19045.0) X64 NetCore 6.0.20
[20:43:19.099] [Main Thread/INFO] [tML]: FrameworkDescription: .NET 6.0.20
[20:43:19.099] [Main Thread/INFO] [tML]: Executable: C:\Program Files (x86)\Steam\steamapps\common\tModLoader\tModLoader.dll
[20:43:19.099] [Main Thread/INFO] [tML]: Working Directory: C:\Program Files (x86)\Steam\steamapps\common\tModLoader
[20:43:19.932] [Main Thread/INFO] [tML]: Saves Are Located At: C:\Users\Daniel\Documents\My Games\Terraria\tModLoader
[20:43:19.934] [Main Thread/INFO] [tML]: Developer mode enabled
[20:43:20.001] [Main Thread/INFO] [tML]: Display DPI: Diagonal DPI is 96. Vertical DPI is 96. Horizontal DPI is 96
[20:43:20.002] [Main Thread/INFO] [tML]: High DPI Display detected: setting FNA to highdpi mode
[20:43:20.004] [Main Thread/DEBUG] [FNA]: Querying linked library versions...
[20:43:20.005] [Main Thread/DEBUG] [FNA]: SDL v2.24.0
[20:43:20.006] [Main Thread/DEBUG] [FNA]: FNA3D v22.8.0
[...]

PR #3993, opened 2023-12-27
Minor Decraft Fixes

Kogsey:
Ref #3991 and #3722 
Doing this because shimmer API is waiting for 1.4.5 and bug needs fix. 

### What is the bug?
1. `Recipe.notDecraftable` is not readable to mods
2. Decrafted bottled stuff + other similar mod effects don't have random 1/3 loss applied
3. Non-stackable items cannot be unstacked when shimmered from stacked items

### How did you fix the bug?
1. New property `Recipe.DecraftDisabled` Wraps `Recipe.notDecraftable` bc `notDecraftable` is vanilla, would be a large patch to replace, and var name DecraftDisabled > notDecraftable 
2. Obsoleted 
3. When shimmering items stack size is limited by max stack not not 9999

1. N/A
2. Previously discussed below.
3. N/A

### Porting notes
- `ConsumeItem` hooks, callbacks and methods have been replaced with `ConsumeIngredient/IngredientQuantity` hooks which have an additional `bool isDecrafting` parameter. 
  - This allows you to apply discounts when shimmering to prevent infinite craft-shimmer exploits.
  - The current classes and hooks are marked `[Obsolete]` but remain functional
- Added read-only property `Recipe.DecraftDisabled`, exposing previously internal state
- The `internal` field `Recipe.alchemy` has been removed, as its functionality is now covered by `Recipe.IngredientQuantityRules.Alchemy`


Chicken-Bones:
Despite the code for consumption being almost entirely reusable, it certainly takes some wrapping your head around. 

I think I favor your proposed alternative for #2. Rather than `ConsumeItemCallback` the decraft version could be called `DecraftItemCallback` or `DecraftIngredientCallback`. 

You'll need to pose the question in #collaborators.

patches/tModLoader/Terraria/Recipe.cs.patch
```diff
@@ -60,7 +60,7 @@
 -	public bool anyFragment;
 +	internal bool anyFragment;
 -	public bool alchemy;
-+	internal bool alchemy;
++	//internal bool alchemy; TML: Recipe.alchemy is unused due to being replaced by consume item hooks
```
Solxanich, line 63: Include in porting notes

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -304,9 +317,21 @@ public Recipe AddOnCraftCallback(OnCraftCallback callback)
 	/// <summary>
 	/// Sets a callback that allows you to determine how many of a certain ingredient is consumed when this recipe is used. Return the number of ingredients that will actually be consumed. By default returns numRequired.
 	/// </summary>
+
+	[Obsolete($"Replaced by {nameof(AddConsumeIngredientCallback)} due to not accounting for shimmer decrafting")]
```
Solxanich, line 321: Include in porting notes

patches/tModLoader/Terraria/Recipe.cs.patch
Solxanich, line 63: Include in porting notes

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -63,6 +71,11 @@ private void AddGroup(int id)
 	/// </summary>
 	public bool Disabled { get; private set; }
 
+	/// <summary>
+	/// Any recipe with this flag won't be decrafted in shimmer.
+	/// </summary>
+	public bool DecraftDisabled => notDecraftable;
```
Solxanich, line 77: Reference as optional in Porting Notes

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -19,32 +19,40 @@ namespace Terraria;
...
-
-			amount = amountUsed;
 		};
 	}
 
 	public readonly Mod Mod;
-	public readonly List<Condition> Conditions = new List<Condition>();
-	public readonly List<Condition> DecraftConditions = new List<Condition>();
+	public readonly List<Condition> Conditions = new();
+	public readonly List<Condition> DecraftConditions = new();
```
Solxanich, line 44: if the code isn't necessary to change, let's try not to change it in future to keep review lite please.

patches/tModLoader/Terraria/Recipe.TML.cs
```diff
@@ -19,32 +19,40 @@ namespace Terraria;
...
+	[Obsolete($"Replaced by {nameof(IngredientQuantityRules)} due to not accounting for shimmer decrafting")]
 	public static class ConsumptionRules
 	{
-		/// <summary> Gives 1/3 chance for every ingredient to not be consumed, if used at an alchemy table. (!) This behavior is already automatically given to all items that can be made at a placed bottle tile. </summary>
-		public static ConsumeItemCallback Alchemy = (Recipe recipe, int type, ref int amount) => {
-			if (!Main.LocalPlayer.alchemyTable) return;
-
-			int amountUsed = 0;
+		[Obsolete($"Replaced by {nameof(IngredientQuantityRules)}.{nameof(IngredientQuantityRules.Alchemy)} due to not accounting for shimmer decrafting")]
+		public static ConsumeItemCallback Alchemy => (Recipe recipe, int type, ref int amount) => IngredientQuantityRules.Alchemy(recipe, type, ref amount, false);
```
Solxanich, line 26: Include in porting notes

patches/tModLoader/Terraria/ModLoader/RecipeLoader.cs
```diff
@@ -196,13 +196,26 @@ public static void OnCraft(Item item, Recipe recipe, Item destinationStack)
 	}
 
 	/// <summary>
-	/// Allows to edit the amount of item the player uses in a recipe.
+	/// Allows to edit the amount of item the player uses in a recipe. Also used to decide the amount a shimmer transformation returns
 	/// </summary>
 	/// <param name="recipe">The recipe used for the craft.</param>
 	/// <param name="type">Type of the ingredient.</param>
 	/// <param name="amount">Modifiable amount of the item consumed.</param>
+	[Obsolete($"Replaced by {nameof(ConsumeIngredient)} due to not accounting for shimmer decrafting")]
```
Solxanich, line 204: Include in porting notes

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -1984,6 +1984,49 @@
...
++				if (item.stack > maxStack)
++					item.stack = maxStack;
++				stackSize -= item.stack;
++				item.shimmerTime = 1f;
++				item.shimmered = true;
++				item.shimmerWet = true;
++				item.wet = true;
++				item.velocity *= 0.1f;
++				item.playerIndexTheItemIsReservedFor = Main.myPlayer;
++				//if (flag) { vanilla uses this to keep different items from stacking not multiple of the same item. Changing here because this fix almost exclusively exists to unstack unstackable items like swords and it feels wrong to leave them stacked.
```
Solxanich: Needs discussion - vanilla deviation?

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -1984,6 +1984,49 @@
...
++				item.velocity.X = 1f * (float)spawnCount;
++				item.velocity.X *= 1f + (float)spawnCount * 0.05f;
++				if (0 % 2 == 0) // Assumed bug in vanilla left in for parity. Vanilla uses if (num14 % 2 == 0) but num14 is only assigned 0 a value. I assume it's supposed to be num15 or 'spawnCount' here.
```
Solxanich: Have to review this with vanilla devs as their bug.

This is likely correct approach to maintain consistency

Solxanich (commented):
Noted a few porting notes / changes related items for future reference.

Main discussion item on the table :
```vanilla uses this to keep different items from stacking not multiple of the same item. Changing here because this fix almost exclusively exists to unstack unstackable items like swords and it feels wrong to leave them stacked.```
Not sure exactly what the change here is and how it varies from vanilla. Could you elaborate or rephrase a different way?

Kogsey:
> Noted a few porting notes / changes related items for future reference.
> 
> Main discussion item on the table : `vanilla uses this to keep different items from stacking not multiple of the same item. Changing here because this fix almost exclusively exists to unstack unstackable items like swords and it feels wrong to leave them stacked.` Not sure exactly what the change here is and how it varies from vanilla. Could you elaborate or rephrase a different way?

Ah yes sorry, that could have been clearer.
So since that entire block is copied from the decrafting section of that function the variable `flag` is usually assigned earlier:
`bool flag = recipe.requiredItem.Count > 1;`
Vanilla uses that block so that the multiple different items that can come from recipe ingredients don't stack, I'm making the assumption that the intent was to stop all items from being in one spot and that weapons should do the same and spread out a little.

The relationship between stack and speed is linear* though so it might not be a bad idea to deviate from vanilla a little more and change it to something that curves down a little harder.
log(x^10) looks pretty good with the added bonus that for x = 1, y = 0
but that problem would apply to the current decrafting system anyway for any mod with higher numbers of recipe ingredients

*technically exponential but the square is negligible for the scale we work with because of the 0.05;

patches/tModLoader/Terraria/Item.cs.patch
Chicken-Bones: Replace `0` with `spawnCount`

patches/tModLoader/Terraria/Item.cs.patch
Chicken-Bones: Comment is incorrect. `flag` is set to `true` when a recipe has more than one ingredient.

The code in this block is meant to try and spread out the recipe ingredients, alternating left and right. However, because of shimmer viscosity, the initial velocities set here barely do anything. Almost all the item spreading out is done by the push away nearby shimmered items code.

So, just remove this entire block.

Chicken-Bones (changes requested):

PR #4005, opened 2024-01-03
Implement Caching for Mod.GetContent() Results

absoluteAquarian:
### What is the new feature?
This pull request implements an internal `Terraria.ModLoader.ContentCache` type that changes how `Mod` adds and gets its `ILoadable`s.

`Mod.GetContent<T>()` has been reported to be [causing lag when used frequently](https://discord.com/channels/103110554649894912/445276626352209920/1191938315620991046).  These changes to `Mod` hope to make this a non-issue.

A simple dictionary in a generic class could be an alternative design, however the code present in this pull request is the only thing that I could come up with.

N/A, internal API change

### ExampleMod updates
N/A, internal API change

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
@@ -61,13 +61,29 @@ internal class ContentCache
 		}
 	}
 
+	private static event Action OnUnload;
+
+	private static readonly Dictionary<string, ContentCache> _sourcesByMod = new();
+
+	internal static ContentCache GetOrCreate(Mod mod) {
+		string key = mod.Name;
+
+		if (!_sourcesByMod.TryGetValue(key, out ContentCache source))
+			_sourcesByMod[key] = source = new ContentCache(mod);
+
+		return source;
+	}
+
+	internal static void Unload() {
+		_sourcesByMod.Clear();
+		Interlocked.Exchange(ref OnUnload, null)?.Invoke();
+	}
+
 	private readonly Mod _mod;
 	private readonly IList<ILoadable> _content = new List<ILoadable>();
 	private event Action OnClear;
 
-	private static event Action OnUnload;
-
-	internal ContentCache(Mod mod) {
+	private ContentCache(Mod mod) {
 		_mod = mod;
 	}
 
@@ -78,10 +94,6 @@ internal class ContentCache
 	internal void Clear() {
 		_content.Clear();
 		Interlocked.Exchange(ref OnClear, null)?.Invoke();
-	}
-
-	internal static void Unload() {
-		Interlocked.Exchange(ref OnUnload, null)?.Invoke();
 	}
 
 	public IEnumerable<ILoadable> GetContent() => _content;
--- a/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
@@ -10,7 +10,7 @@ namespace Terraria.ModLoader;
 /// </summary>
 internal class ContentCache
 {
-	private static Dictionary<Type, IList> _cachedContentForAllMods = new();
+	private static readonly Dictionary<Type, IList> _cachedContentForAllMods = new();
 
 	internal static bool contentLoadingFinished;
 
--- a/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
@@ -67,8 +67,8 @@ internal class ContentCache
 
 	private static event Action OnUnload;
 
-	internal ContentCache(Mod source) {
-		_mod = source;
+	internal ContentCache(Mod mod) {
+		_mod = mod;
 	}
 
 	internal void Add(ILoadable loadable) {
--- a/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ContentCache.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -202,7 +202,8 @@ private static Mod Instantiate(ModLoadContext mod)
 			m.Side = mod.properties.side;
 			m.DisplayName = mod.properties.displayName;
 			m.TModLoaderVersion = mod.properties.buildVersion;
-			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null; 
+			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null;
+			m.content = ContentCache.GetOrCreate(m);
```
Chicken-Bones: No need for this to be a `GetOrCreate`, since `m` is always a new `Mod`. Thus `ContentCache._sourcesByMod` will also be unnecessary.

This should be initialized in the `Mod` constructor.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Container object for <see cref="ILoadable"/> instances added for a <see cref="Mod"/>
+/// </summary>
+internal class ContentCache
+{
+	private class Cache<T> where T : ILoadable {
+		private static Dictionary<string, Cache<T>> _instancesPerMod;
```
Chicken-Bones: While using a static `Cache<T>` dictionary provides better type safety, it makes the code a bit more complicated due to caching and clearing. 

This can be replaced with a `Dictionary<Type, IList> perTypeLists` which you cast to `IReadOnlyList<T>` or `IEnumerable<T>` when looking it up.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,104 @@
...
+		private IReadOnlyList<T> _cache;
+
+		private Cache(ContentCache source) {
+			_source = source;
+		}
+
+		public IEnumerable<T> GetOrCacheContent() {
+			IEnumerable<T> lazyContent = _source._content.OfType<T>();
+
+			if (_source._mod.loading) {
```
Chicken-Bones: This is good.

After applying my above advice, please also add a bail-out for when there's no content at all because the mod hasn't yet been loaded.

Chicken-Bones (changes requested):
Good work on this. I've proposed some adjustments for simplicity.

While this PR addresses the performance incurred by `OfType<T>` at runtime, I expect many users are not as interested in `Mod.GetContent<T>` as they are in `ModContent.GetContent<T>`. 

For `ModContent.GetContent<T>` we should create caches on-demand. Note we need to track whether content has been loaded (See `ResizeArrays`), and this one will need a static clear callback.

See the `ModTypeLookup` static constructor for a reference. Note that we only clear caches for generic types which are defined in tML. Types defined in mods are re-created each load, and the generic caches get cleared by assembly unloading (when it actually works). Registering an unload callback for mod types can actually prevent those types from ever being unloaded, unless you clear the callback as well.

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -202,7 +202,7 @@ private static Mod Instantiate(ModLoadContext mod)
 			m.Side = mod.properties.side;
 			m.DisplayName = mod.properties.displayName;
 			m.TModLoaderVersion = mod.properties.buildVersion;
-			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null;
```
Chicken-Bones: whitespace change, sorry, please revert

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+
+		// Check of the cache already exists
+		// It will already be a ReadOnlyList<T>, so it just needs to be cast back to it
+		nint handle = typeof(T).TypeHandle.Value;
+		if (_cachedContentForAllMods.TryGetValue(handle, out IList cachedContent))
+			return (IReadOnlyList<T>)cachedContent;
+
+		// Construct the cache
+		if (_cachedContentForAllMods.Count == 0)
+			OnUnload += UnloadStaticCache;
```
Chicken-Bones: No need for the `OnUnload` `event` here now. Since you don't have any static fields on generic types.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+			// Content has not fully loaded yet, so we can't rely on the cache.
+			// Return a lazy enumerable instead.
+			return ModLoader.Mods.SelectMany(static m => m.GetContent<T>());
+		}
+
+		// Check of the cache already exists
+		// It will already be a ReadOnlyList<T>, so it just needs to be cast back to it
+		nint handle = typeof(T).TypeHandle.Value;
+		if (_cachedContentForAllMods.TryGetValue(handle, out IList cachedContent))
+			return (IReadOnlyList<T>)cachedContent;
```
Chicken-Bones: Hot path, toss this one first.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+	internal static void Unload() {
+		Interlocked.Exchange(ref OnUnload, null)?.Invoke();
+	}
+
+	private static void UnloadStaticCache() {
+		_cachedContentForAllMods.Clear();
+	}
+
+	public static IEnumerable<T> GetContentForAllMods<T>() where T : ILoadable {
+		if (!hasLoadingStarted || loadingContent) {
```
Chicken-Bones: Do we need 2 flags? Can we just have `contentLoadingFinished`, which we can clear in `Unload`?

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	internal void Add(ILoadable loadable) {
+		_content.Add(loadable);
+	}
+
+	public IEnumerable<ILoadable> GetContent() => _content.AsReadOnly();  // Prevent exposing the list via hard cast
+
+	public IEnumerable<T> GetContent<T>() where T : ILoadable {
+		if (!hasModLoadedYet || _mod.loading) {
```
Chicken-Bones: We can remove the need for `hasModLoadedYet` by returning `Enumerable.Empty<T>` when `_content.IsEmpty`. Mods can only add content during loading, and after loading we can freeze.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+		// Construct the cache
+		if (_cachedContent.Count == 0)
+			OnUnload += UnloadInstanceCache;
+
+		IReadOnlyList<T> content = _content.OfType<T>().ToList().AsReadOnly();
+		_cachedContent[handle] = (IList)content;
+		return content;
+	}
+
+	private void UnloadInstanceCache() {
```
Chicken-Bones: No need to unload, it'll get GC'd with the mod itself.

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,93 @@
...
+/// </summary>
+internal class ContentCache
+{
+	private static event Action OnUnload;
+	private static Dictionary<nint, IList> _cachedContentForAllMods = new();
+
+	internal static bool hasLoadingStarted;
+	internal static bool loadingContent;
+
+	internal static void Unload() {
```
Chicken-Bones, line 17: Just clear the dict and reset the flags.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ContentCache.cs
```diff
@@ -0,0 +1,81 @@
...
+
+		if (!contentLoadingFinished) {
+			// Content has not fully loaded yet, so we can't rely on the cache.
+			// Return a lazy enumerable instead.
+			return ModLoader.Mods.SelectMany(static m => m.GetContent<T>());
+		}
+
+		// Construct the cache
+		IReadOnlyList<T> content = ModLoader.Mods.SelectMany(static m => m.GetContent<T>()).ToList().AsReadOnly();
+		_cachedContentForAllMods[typeof(T)] = (IList)content;
```
Chicken-Bones: Suggest extracting common code
```
		var query = ModLoader.Mods.SelectMany(static m => m.GetContent<T>());
```

PR #4013, opened 2024-01-08
Fix "is_in_docker command not found" error in manage-tModLoaderServer.sh

kairusds:
### What is the bug?
A function not properly typed in the dedicated server management script.

### How did you fix the bug?
By  adding parentheses.

N/A


JavidPack:
All the other functions are written the same way, are you sure this is correct? Are all of those other functions broken? According to <https://linuxize.com/post/bash-functions/> both `function_name ()` and `function function_name` are correct, but I'm not seeing `function function_name ()` as one of the syntax examples.

kairusds:
I checked again and it seems to be an error on my end, my bad.

Issue #3998, opened 2023-12-30
tModLoader won't install with the GOG version on my Mac


djbchi-young:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

I have the GOG version of Terraria so I tried to do the manual installation steps but I keep getting the same error no matter what I try. Am I extracting it somewhere wrong? It always seems to find my Terraria dotnet so I don't think it matters?

### Log File

[TerminalLog.txt](https://github.com/tModLoader/tModLoader/files/13797388/TerminalLog.txt)
[client.log](https://github.com/tModLoader/tModLoader/files/13797390/client.log)

### Steps to reproduce

1. Download latest tModLoader.zip
2. Extract to Terraria.app/Contents
3. run chmod +x on all .sh files
4. run start-tModLoader.sh

TacoConKvass:
If you followed [this guide](https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-players), and you tried to install TMod in the Terraria folder, try installing it [side by side](https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-players#manual-installation) because 

> DO NOT install the tModLoader files directly into the Terraria folder. This option not supported for GOG on Mac.

JavidPack:
I believe for Mac the manual GOG install instructions expect the user places the tModLoader folder in `/Applications/` or `/Applications/Terraria/`. You have it in `/Applications/Terraria/Terraria.app/Contents/`, so you'll need to move it. 

Additionally, you'll need to try running `arch -x86_64 /bin/bash ./start-tModLoader.sh` in the terminal after deleting your dotnet folder (`/Applications/tModLoader/dotnet`). You'll need to be in a terminal navigated to the `/Applications/tModLoader` folder. 

If you are still having issues, the fastest solution is to come to the #support-forum channel in our [Discord chat](https://discord.gg/tmodloader). Be sure to post your logs when asking for support.

JavidPack:
Related to #3986

Chicken-Bones:
Fixed in #4018 and #4017

Issue #3986, opened 2023-12-17
Unable to install on M2 Macbook Pro (GOG)


Marseel-E:
### Version

Latest 1.4.4-stable

### OS

Mac

### Platform

GOG

### This bug affects

Gameplay as a Player

### Description

I followed the manual installation guide, downloaded the latest zip file, ran `chmod +x ...` for all the `.sh` and `.bat` files in the "tModLoader" folder after extraction. then Tried to run `start-tModLoader.sh` by running `./start-tModLoader.sh`, and it didn't work. I believe its a compatability issue, because I'm seeing stuff about SDL2 using x86-64 instead of ARM. But here are the logs maybe I'm missing something.

```
marseel@Marseels-MacBook-Pro tModLoader % ./start-tModLoader.sh
./start-tModLoader.sh: line 6: PPID: readonly variable
marseel@Marseels-MacBook-Pro tModLoader % You are on platform: "Darwin"
Logging to /Users/marseel/Library/Application Support/tModLoader/tModLoader-Logs/Launch.log
Verifying .NET....
This may take a few moments.
Fixing .NET SDL PATH issues
/Users/marseel/Library/Application Support/tModLoader/Libraries/Steamworks.NET -> /Users/marseel/Library/Application Support/tModLoader/Libraries/steamworks.net
Success!
Parsing .NET version requirements from runtimeconfig.json
Success! dotnet_version=6.0.14
Checking for old .NET versions to remove from folder
Cleanup Complete
Checking dotnet install...
Dotnet should be present in "/Users/marseel/Library/Application Support/tModLoader/dotnet/6.0.14"
Attempting Launch...
sed: /proc/1/cmdline: No such file or directory
Launched Using Local Dotnet. Launch command: "/Users/marseel/Library/Application Support/tModLoader/dotnet/6.0.14/dotnet" tModLoader.dll ""
[17:32:58.226] [Main Thread/INFO] [tML]: Starting tModLoader client 1.4.4.9+2023.10.3.0|2023.10|stable|Stable|23aa262bde1ae5440cf2b05b2f676feb787cdc95|5250055878079335727 built 11/30/2023 7:06PM
[17:32:58.234] [Main Thread/INFO] [tML]: Log date: 12/17/2023
[17:32:58.234] [Main Thread/INFO] [tML]: Running on OSX (v14.2.0) Arm64 NetCore 6.0.14
[...]

PR #4017, opened 2024-01-09
Automatically switch to x86 architecture when Rosetta is present

Chicken-Bones:
### What is the new feature?

On M1 Macs, when `./start-tModLoader.sh` is launched via terminal or otherwise outside of steam, the start scripts will automatically detect the `arm64` architecture and restart under `x86_64`

- #3986 
- #3998
- Developers needing custom start scripts

Not that I can think of

`./start-tModLoader.sh` in a terminal where `arch` reports `arm64` on an M1 Mac

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
@@ -8,9 +8,10 @@ cd "$(dirname "$0")"
 
 echo "You are on platform: \"$_uname\" arch: \"$_arch\""
 
-# Detect arm64 launches and the presence of rosetta (stolen from the dotnet official install script)
+# Detect arm64 launches and the presence of Rosetta (oahd running on the system is how the dotnet official install script does it)
 if [ "$_arch" = "arm64" ] && [ "$(/usr/bin/pgrep oahd >/dev/null 2>&1;echo $?)" -eq 0 ]; then
-	echo "arm64 detected, restarting under arch -x86_64"
+	echo "arm64 environment with Rosetta detected, restarting under arch -x86_64"
+	# Note this only changes the environment, so that dotnet install scripts download an x86 version. Launching an x86 process from an arm shell or vice versa does not require using arch, or otherwise intentionally invoking Rosetta.
 	exec arch -x86_64 ./ScriptCaller.sh "$@"
 fi
 
@@ -72,7 +73,7 @@ else
 		echo "A Windows dotnet executable was detected, possibly from a previous Proton launch. Deleting dotnet_dir and resetting"  2>&1 | tee -a "$LogFile"
 		rm -rf "$dotnet_dir"
 		mkdir "$dotnet_dir"
-	elif [[ "$(file "$install_dir/dotnet")" == *"arm64"* ]]; then
+	elif [[ "$_arch" != "arm64" ]] && [[ "$(file "$install_dir/dotnet")" == *"arm64"* ]]; then
 		echo "An arm64 install of dotnet was detected. Deleting dotnet_dir and resetting"  2>&1 | tee -a "$LogFile"
 		rm -rf "$dotnet_dir"
 		mkdir "$dotnet_dir"
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
@@ -6,7 +6,13 @@
 cd "$(dirname "$0")"
 . ./BashUtils.sh
 
-echo "You are on platform: \"$_uname\""
+echo "You are on platform: \"$_uname\" arch: \"$_arch\""
+
+# Detect arm64 launches and the presence of rosetta (stolen from the dotnet official install script)
+if [ "$_arch" = "arm64" ] && [ "$(/usr/bin/pgrep oahd >/dev/null 2>&1;echo $?)" -eq 0 ]; then
+	echo "arm64 detected, restarting under arch -x86_64"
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
```diff
@@ -66,6 +72,10 @@ else
 		echo "A Windows dotnet executable was detected, possibly from a previous Proton launch. Deleting dotnet_dir and resetting"  2>&1 | tee -a "$LogFile"
 		rm -rf "$dotnet_dir"
 		mkdir "$dotnet_dir"
+	elif [[ "$(file "$install_dir/dotnet")" == *"arm64"* ]]; then
+		echo "An arm64 install of dotnet was detected. Deleting dotnet_dir and resetting"  2>&1 | tee -a "$LogFile"
```
Solxanich: Need to limit to Darwin Arm64.
Linux Arm64 is still valid and would loop here

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
```diff
@@ -6,7 +6,13 @@
 cd "$(dirname "$0")"
 . ./BashUtils.sh
 
-echo "You are on platform: \"$_uname\""
+echo "You are on platform: \"$_uname\" arch: \"$_arch\""
+
+# Detect arm64 launches and the presence of rosetta (stolen from the dotnet official install script)
+if [ "$_arch" = "arm64" ] && [ "$(/usr/bin/pgrep oahd >/dev/null 2>&1;echo $?)" -eq 0 ]; then
+	echo "arm64 detected, restarting under arch -x86_64"
+	exec arch -x86_64 ./ScriptCaller.sh "$@"
```
Solxanich: The second calculation is hard to tell at a glance that it limits to Mac's specifically & does not impact Linux Arm64. 
Please add a comment about that the latter 'checks if the Rosetta process is running'.

Not clear at a glance if that check is system-wide or session specific, likely worth a confirmation test of have an app running on rosetta than launch this & verify this does trigger

Solxanich (commented):
The big notice from me is that a refactor to reduce ScriptCaller.sh length meant by in order.
I'd prefer all the platform specific logic be kept within a renamed UnixLinkerFix or a separate .sh to keep readability in the main script manageable & abstract.
But that's not relevant to this PR. 

Of relevance is that as written, Linux Arm64 is not viable.  Ought to limit the _uname to Darwin to avoid conflicts if we intend to restore Arm64 linux for dedicated servers.

No other concerns

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Chicken-Bones: Better to put a `"$_arch" != "arm64"`

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Chicken-Bones: I did add that comment, explicitly, on the line above. `arch` changes it for the process tree of the called command.

Solxanich: Apologies, I'll clarify - 
The `/usr/bin/pgrep oahd >/dev/null 2>&1` checks if it is running, not that it is installed.  Good to clarify 'presence'. Did miss the rosetta portion of that comment and only somehow saw arm64 and part in brackets. Apologies.

My concern is that it is 'running' for a second process already, So the order of ops would go:
1) App 1 is running with Rosetta in background
2) Launch tML
3) `/usr/bin/pgrep oahd >/dev/null 2>&1` returns '1' since 'Rosetta is running', even if not in the current terminal context of first launch
4) And thus it continues to run past that point as if it is non-Rosetta Arm64.

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Solxanich: Further, if '$(file ....) != $_arch' if we are being fastidious to that level?

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Chicken-Bones: `oahd` is a system process which indicates that Rosetta is present on the machine. At least, that's my understanding. The bash script will not be 'running under rosetta' at all. Only when an x86 process is launched is Rosetta 'running' in a sense. Using `arch -x86_64` changes the environment for the process tree so that the dotnet download script can detect the right architecture.

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Chicken-Bones: There's not a guarantee atm that `file` will output something in the exact same form as `uname -m` so I'm only handling the "`arm64` was installed by accident" case

patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
Solxanich: To paraphase back your intended design:
```# Detect arm64 arch and the presence of rosetta via `oahd`, and if present, restart with forced usage of x86 dotnet runtime```

No further concerns with approach, misunderstood as attempting to force rosetta in the terminal itself.

JavidPack (commented):
Logic seems good.

Issue #3807, opened 2023-09-12
[Feature Proposal] Description parameter interpolation


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Very similarly to #3710, interpolation of certain variables like mod version for description and description_workshop. Primary use case is displaying the currently latest mod version in the mod description (i.e. as a link to changelogs on a wiki like `link.here.com/{ModVersion}`) without having to manually update it every time you publish.

### What does this proposal attempt to solve or improve?

Reduce boilerplate work for modders making sure that their versions are all matching across changelog and descriptions.

PR #4036, opened 2024-01-21
Description Parameter Interpolation

Destructor-Ben:
### What is the new feature?
https://github.com/tModLoader/tModLoader/issues/3807

Since this is very hard to test, add the following log function calls and disable the following checks in the shown functions and attempt to publish ExampleMod.

`WorkshopSocialModule.TML.cs`
image
(same method just scroll down)
image

`WorkshopHelper.cs`
`APublisherInstance`
`private void UpdateItem(bool creatingItem = false)`
image

https://github.com/tModLoader/tModLoader/issues/3807

Overhauling the publishing system, which is very messy at the moment and doesn't give much freedom to modders.

In either `description.txt` or `description_workshop.txt`:
```
You can interpolate these values into a mod description:
- ModVersion: {ModVersion}
- ModHomepage: {ModHomepage}
- tMLVersion: {tMLVersion}
- tMLBuildPurpose: {tMLBuildPurpose}
```

### ExampleMod updates

Added test values to the descriptions (I'm not sure if these will be interpolated because of how ExampleMod is published).

### DIFF
--- a/ExampleMod/description.txt
+++ b/ExampleMod/description.txt
@@ -1,3 +1,9 @@
 Learn how to mod with tModLoader by exploring the source code for this mod.
 
 You can find the source code at https://github.com/tModLoader/tModLoader/tree/1.4.4/ExampleMod
+
+You can interpolate these values into a mod description:
+- ModVersion: {ModVersion}
+- ModHomepage: {ModHomepage}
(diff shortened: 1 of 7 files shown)


JavidPack:
Looks great. It's neat that it still exports correctly to the original txt files too.

Issue #360, opened 2018-06-08
Additional suggestions for other language settings


rrryutaro:
**I am using Google translation because I can not speak English.**

### Description
Suggestion.
I want to add language setting.
If you add "Other" to the end of the language menu and select Other, you want to acquire the `Terraria.Localization.Content. *` File on the `Main.SavePath` and read it think.
My main purpose is to use it in Japanese, but I think that it will be possible to customize each json file freely by this method.

### Code example
```
[Terraria.Main.cs]
    protected void DrawMenu(GameTime gameTime)
        :
        array9[10] = Language.GetTextValue("Language.Other");
        num5 = 11;
        :
        array9[10] = Language.GetTextValue("Language.Other");
        array9[11] = Lang.menu[5].Value;
        num5 = 12;

[Terraria.Localization.GameCulture.cs]
    public static readonly GameCulture Other = new GameCulture("ja-JP", 10);

[Terraria.Localization.LanguageManager.cs]
    private void LoadFilesForCulture(GameCulture culture)
    {
        string[] languageFilesForCulture = this.GetLanguageFilesForCulture(culture);
        string[] array = languageFilesForCulture;
        if (array.Length == 0)
        {
            try
            {
                Directory.GetFiles(Main.SavePath, "Terraria.Localization.Content.*")
                    .ToList().ForEach(path => this.LoadLanguageFromFileText(File.ReadAllText(path)));
            }
            catch (Exception) { }
        }
        else
        {
            :

[Terraria.Localization.Content.en-US.Main.json.patch]
"Language": {
    :
  "Other": "Other"
},
```

### About contents of pull request
1. Code modified in my environment
2. Run [Diff tModLoader] at [tModLoader Dev setup]
3. Generated difference patch
Should I pull request this?

JavidPack:
I've been thinking the best approach would just be to implement other languages directly in tModLoader, but I guess a user specified "other" option could also be an approach. 

The main problem isn't this though, have you tried this yet? The main problem with Japanese is that the font files do not have the japanese hiragana and katakana. I also don't think the chinese kanji math the japanese kanji exactly if my memory is correct. Someone using this "Other" functionality would need to make or find fonts: https://forums.terraria.org/index.php?threads/dynamicspritefontgenerator-0-4-generate-fonts-without-xna-game-studio.57127/

Let me know how much testing you have actually done and we can talk about it more.

rrryutaro:
[...]

PR #4039, opened 2024-01-22
Added Turkish Localization

NovaDiablox:


JavidPack:
This is quite impressive, thanks for putting in this effort.

Unfortunately, we currently don't have any way of fully supporting languages other than the languages supported by Terraria. There has been discussions in the past about supporting new languages, but there has not been any work on writing that code: https://github.com/tModLoader/tModLoader/issues/360

For now, this PR won't be merged because it wouldn't do anything. 

There are, however, a few options to get basic Turkish support:

## Resource Pack
What you can do is make a resource pack, specifically a language pack. If you make a resource pack out of your translations, you'll be able to see your changes in-game. (https://forums.terraria.org/index.php?threads/the-ultimate-guide-to-content-creation-and-use-for-the-terraria-workshop.100652/#languagepack and https://terraria.wiki.gg/wiki/Resource_Pack#Creating_a_Resource_Pack) 

You'll likely need to replace one of the existing language options I think. I haven't done this myself but the guide does mention this as an option in the "Limitations of Language Packs" section.

For example, here is a language pack on the Terraria workshop for Turkish: https://steamcommunity.com/sharedfiles/filedetails/?id=2441453094&searchtext=turkish

Here is the code: https://github.com/qabRieL99/TerrariaTurkish

Note that if you wish to publish the resource pack, it would need to be published on the Terraria workshop since the tModLoader workshop is just for mods. 

## Translation Mod
Another option is to put your translations into a mod. This might be better for users because they'll be able to download the translations directly from the tModLoader workshop instead of the Terraria workshop. You'll have to tweak your translation files since mod translations are formatted slightly different. (https://github.com/tModLoader/tModLoader/wiki/Localization#how-localization-works)

Note that users won't see the translations until they enable the mod. 


Feel free to come by [our discord server](https://discord.gg/tmodloader) if you have further questions.

PR #4040, opened 2024-01-23
Config List Fixes

Destructor-Ben:
### What is the bug?
- Selected mod in the config list isn't consistent between closing and reopening config list, opening then closing a config, and using escape and the back button to close the config or config list
- Mod and config names with chat tags at the beginning are sorted to the top because the name starts with a bracket

### How did you fix the bug?
- By handling mod selection differently (instead of deselecting mod on back, it is deselected on open with an option to set a selected mod)
- Cleaning mod and config names of chat tags

No.


JavidPack:
Looks good.

PR #4047, opened 2024-02-02
fix: Sorting of searched mods through download mods Fixes #3792

### LINKED ISSUE #3792: Mods do not sort when searching through download mods in workshop
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When searching mods to download in workshop using a keyword, sorting options do not work.

### Log File

[client.log](https://github.com/tModLoader/tModLoader/files/12528707/client.log)

### Steps to reproduce

1. Go to Download Mods under Workshop
2. Search for mods using keyword, e.g., potion
3. Click through sorting options in the top left of the box

### Expected Behavior

Mods should be sorted as per sorting option (last updated, most popular, most downloaded)

### Actual Behavior

Mods are not sorted as expected - even under the default "sort by recently updated", mods do not appear in chronological order. In fact, the mod list order is the same regardless of the sorting option.

### Reproduction frequency

Constant, has been an issue since Sep 1st update and has not resolved despite re-installation.

### Additional Information

Sorting works fine without a search term - only appears as an issue when a search is applied.

Metratrj:
Added additional sorting when text search is used for the workshop query
Closes #3792 

### What is the bug?

When searching mods to download in workshop using a keyword, sorting options do not work. 

### How did you fix the bug?

I have added additional sorting after recieving the query results. 
The sorting only runs when it got skiped befor in the query.

After my research onto the steamworkshop documentation, it is only possible to filter by text or sort over the ugc query.


Metratrj:
@Solxanich @Chicken-Bones Any feedback? Should something more need to be changed?

JavidPack:
I tested it and it seems to work. I'm not too familiar with the redesigned workshop ui code, so I'm not sure if the approach is correct. I'll let Solxanich take a look.

Before we would just sort the UIModDownloadItem using the UIList sorting capabilities, I'm not sure why that isn't used anymore. It seems that clicking the sort buttons now trigger the workshop items to be queried again each time.

Chicken-Bones:
Your fix only sorts each page, which works when the search text reduces results to only a single page. @Solxanich perhaps it's acceptable in the interim, with a clear note about the issues.

It's a known issue with steam workshop backend that they don't sort and search. Maybe we can forward it to valve?

Solxanich:
Apologies, I've been busy with some IRL stuff.

CB is correct that this solution only works if there is less than `Steamworks.Constant.unPageCount = 50` results. 

Before I even treat this as a notable issue, I need better context on what exactly we are trying to sort by and what the end objective is here.
Is this an acceptable gap or is there an intended use case to search for a name and get `latest updated` result  over `most downloaded`. 

On Steam/Valve's end Text Search is first sorted using 'by relevance', than it is a hardcoded sub-sorting by ~'Most Downloads'~    Edit1: By Most Popular
It is not a configurable item for how the sub-search is performed, this is true.   

More pertinent is `CalculateQuerySort` in `SteamedWraps` under Social.Steam
We could change 'RankedByTextSearch' to only happen when SortMode.DownloadsDescending.
Whether SortMode.Hot also fits better with 'RankedByTextSearch' is up for discussion.

If we do want to try that, someone would first need to go in to Steam Workshop in the Steam client and compare results of those searches and present findings on if it was adequate.

Edit 2: relevant code
```
public static EUGCQuery CalculateQuerySort(QueryParameters qParams)
{
	if (!string.IsNullOrEmpty(qParams.searchGeneric) || !string.IsNullOrEmpty(qParams.searchAuthor))
		return EUGCQuery.k_EUGCQuery_RankedByTextSearch;

	return (qParams.sortingParamater) switch {
		ModBrowserSortMode.DownloadsDescending => EUGCQuery.k_EUGCQuery_RankedByTotalUniqueSubscriptions,
		ModBrowserSortMode.Hot => EUGCQuery.k_EUGCQuery_RankedByPlaytimeTrend,
		ModBrowserSortMode.RecentlyUpdated => EUGCQuery.k_EUGCQuery_RankedByLastUpdatedDate,
		_ => EUGCQuery.k_EUGCQuery_RankedByTextSearch
	};
}
```

Metratrj:
Sorry for my late answer. I was also busy the last days.

I can't get same results when searched with the steam client. The nearest results to a k_EUGCQuery_RankedByTextSearch query I can get are when I sort for relevance. 
terraria relevance
It seems to be that when QueryParameters.searchGeneric variable is filled with the search term, there is no need to sort for k_EUGCQuery_RankedByTextSearch.

The suggestion from @Solxanich modifying the CalculateQuerySort does what we are trying to accomplish.

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
Metratrj, line 1: Remove unnecessary file change

Solxanich:
Approved, thanks for working on this and verifying the solution!

PR #4052, opened 2024-02-05
Separate DisplayName and DisplayNameClean

JavidPack:
image

Some mods use chat tags to augment their displayname. We currently log displayname directly in logs and server console in various places. 

It has been suggested that we show a "clean" displayname in these situations. This seems reasonable. This PR addresses this issue

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
@@ -12,7 +12,7 @@ internal class LocalMod
 
 	public string Name => modFile.Name;
 	public string DisplayName => string.IsNullOrEmpty(properties.displayName) ? Name : properties.displayName;
-	internal readonly string DisplayNameClean; // Suitable for console output, chat tags stripped away.
+	public readonly string DisplayNameClean; // Suitable for console output, chat tags stripped away.
 	public Version tModLoaderVersion => properties.buildVersion;
 
 	public bool Enabled {
--- a/patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
@@ -12,6 +12,7 @@ internal class LocalMod
 
 	public string Name => modFile.Name;
 	public string DisplayName => string.IsNullOrEmpty(properties.displayName) ? Name : properties.displayName;
+	internal readonly string DisplayNameClean; // Suitable for console output, chat tags stripped away.
 	public Version tModLoaderVersion => properties.buildVersion;
 
 	public bool Enabled {
@@ -25,6 +26,7 @@ internal class LocalMod
 	{
 		this.modFile = modFile;
 		this.properties = properties;
+		DisplayNameClean = Utils.CleanChatTags(DisplayName);
 	}
 
 	public LocalMod(TmodFile modFile) : this(modFile, BuildProperties.ReadModFile(modFile))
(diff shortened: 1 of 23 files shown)


patches/tModLoader/Terraria/ModLoader/Mod.cs
```diff
@@ -75,6 +75,10 @@ public partial class Mod
 	/// The display name of this mod in the Mods menu.
 	/// </summary>
 	public string DisplayName { get; internal set; }
+	/// <summary>
+	/// Same as DisplayName, but chat tags are removed. This can be used for more readable logging and console output.
+	/// </summary>
+	public string DisplayNameClean { get; internal set; }
```
Chicken-Bones: Consider a lazy initialized property with ??= on the backing field?

patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
```diff
@@ -12,6 +12,7 @@ internal class LocalMod
 
 	public string Name => modFile.Name;
 	public string DisplayName => string.IsNullOrEmpty(properties.displayName) ? Name : properties.displayName;
+	internal readonly string DisplayNameClean; // Suitable for console output, chat tags stripped away.
```
Chicken-Bones: No need for internal

patches/tModLoader/Terraria/ModLoader/UI/UILoadMods.cs
```diff
@@ -36,25 +36,27 @@ public void SetLoadStage(string stageText, int modCount = -1)
...
 		this.modCount = modCount;
-		if (modCount < 0) SetProgressText(Language.GetTextValue(stageText));
+		if (modCount < 0) SetProgressText(Language.GetTextValue(stageText), Language.GetTextValue(stageText));
 		Progress = 0;
 		SubProgressText = "";
 	}
 
-	private void SetProgressText(string text, string logText = null)
+	// uitext might have chat tags, text is the same without chat tags, logText might be unique
+	private void SetProgressText(string uitext, string text, string logText = null)
```
Chicken-Bones: Probably better to do `text, uiText = null, logText = null`

or just strip the chat tags here.

Chicken-Bones (commented):
Fairly good overall. It's annoying that we need so many `DisplayNameClean` properties everywhere.

Please search for all references to `Mod.DisplayName`

I feel like `DisplayName` should always be clean, and there should be a separate `DisplayNamePretty` which we use selectively in UI components, but that might not be practical overall.

JavidPack:
Updated all usages to use the appropriate version. Mods are using Mod.DisplayName for a variety of purposes, I don't think we should change the behavior of the existing field. I agree that DisplayName/DisplayNamePretty might be better if it was a fresh start.

patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
```diff
@@ -107,7 +107,7 @@ internal static void RemoveAll(Mod mod)
 
 		foreach (var asm in AssemblyManager.GetModAssemblies(mod.Name)) {
 			if (assemblyDetours.TryGetValue(asm, out var list)) {
-				Logging.tML.Debug($"Unloading {list.ilHooks.Count} IL hooks, {list.detours.Count} detours from {asm.GetName().Name} in {mod.DisplayName}");
```
Chicken-Bones, line 110: Logging should use `mod.Name` imo.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -140,7 +140,7 @@ public virtual void AutoStaticDefaults()
 			NPCLoader.bannerToItem[Banner] = BannerItem;
 		}
 		else if (Banner != 0 || BannerItem != 0) {
-			Logging.tML.Warn(Language.GetTextValue("tModLoader.LoadWarningBannerOrBannerItemNotSet", Mod.DisplayName, Name));
+			Logging.tML.Warn(Language.GetTextValue("tModLoader.LoadWarningBannerOrBannerItemNotSet", Mod.DisplayNameClean, Name));
```
Chicken-Bones: And here

PR #4053, opened 2024-02-06
Modded Sand and Sandgun support

blahblahbal:
### What is the new feature?
Easier modded sand Sandgun ammo and general falling sand block support.

Without this change, modded sand sandgun ammo requires a detour and some reflection (due to WorldGen.SpawnFallingBlockProjectile() being private). This change publicizes that method and adds 3 Sets (TileID, ProjectileID, ItemID) that allows for much easier sandgun ammo addition, as well as falling sand block support.

Not that I'm aware of (unless you count using a detour and reflection).

See the ExampleMod updates section for a sample (I've included sand tile and item, falling sand projectile, and sandgun projectile in the updates I've done to ExampleMod).

### ExampleMod updates
I updated ExampleMod to add the previously mentioned files; these files should be pretty self-explanatory.


Rijam:
Make sure you follow the tModLoader style guide https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide (mainly the K&R Style of curly braces).

I'm no authority here, but I think it would be better to put the declarations for the ItemID.Sets and TileID.Sets in `ItemID.TML.cs` and `TileID.TML.cs` instead of the patch file to reduce on patches. Also, you could use the full name IDs instead of the numeric IDs for readability if you want (you already did that for your ItemID.Sets, but you could do it for the TileID.Sets, too).

blahblahbal:
Oh, I missed the bracket style thing. I'll fix and push again. I can also put the Sets in the TML files. Didn't realize that was an option haha. (Will also change the numeric IDs to their appropriate named IDs)

EDIT: Looks like the rest of the sets in TileID.TML.cs are also using their numeric IDs? I'm not totally sure if I should change them now, but I guess it would reduce the possibility for errors. I'll do it anyway.

blahblahbal:
Okay so the diff is confusing me; it _appears_ like the Player.cs.patch file still has the old code for the "if (projToShoot == 42)" section, but I've changed it? Might just be github desktop messing with my mind but the old lines aren't highlighted in red.

(this is in the commit I haven't pushed yet)

EDIT: Seems like it was just github messing with my mind lol, opened Player.cs.patch in an editor and checked the lines and the removed lines do indeed have a - on them.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7441,6 +7441,27 @@
...
+-			else if (item.type == 408) {
+-				projToShoot = 68;
+-				Damage += 5;
+-			}
+-			else if (item.type == 1246) {
+-				projToShoot = 354;
+-				Damage += 5;
+-			}
++			projToShoot = ItemID.Sets.SandgunAmmoToProjectile[item.type];
++			Damage += ItemID.Sets.SandgunAmmoDamage[item.type];
```
steviegt6: It's preferred to use multi-line comments (`/* ... */`) to remove large swathes of code. This minimizes exposure of ReLogic code in patches and makes them more resilient to change.

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2962,6 +2971,60 @@
...
+-			case 333:
+-				projType = 414;
+-				dmg = 0;
+-				break;
+-			case 495:
+-				projType = 812;
+-				break;
+-			default:
+-				projType = 31;
+-				break;
```
steviegt6: See my previous comment about handling large deletions.

patches/tModLoader/Terraria/ID/TileID.TML.cs
```diff
@@ -16,6 +16,28 @@ partial class Sets
 		/// <summary> Allows non-solid tiles to be sloped (solid tiles can always be sloped, regardless of this set). </summary>
 		public static bool[] CanBeSloped = Factory.CreateBoolSet();
 
+		/// <summary>
+		/// The ID of the projectile for this falling block. Defaults to -1.
+		/// </summary>
+		public static int[] FallingBlockProjectileID = Factory.CreateIntSet(-1,
+			Sand, ProjectileID.SandBallFalling,
+			Ebonsand, ProjectileID.EbonsandBallFalling,
+			TileID.Mud, ProjectileID.MudBall,
```
steviegt6: Any reason you qualify `Mud` here (and farther below).

patches/tModLoader/Terraria/ID/ItemID.TML.cs
```diff
@@ -6,6 +6,17 @@ partial class ItemID
...
+		/// <summary>
+		/// The projectile ID of a sandgun ammo item ID. Defaults to 42, as this is the normal sandgun sand projectile.
+		/// This shouldn't be your falling sand projectile - you need to create a second projectile for the sandgun.
+		/// </summary>
+		public static int[] SandgunAmmoToProjectile = Factory.CreateIntSet(42, EbonsandBlock, ProjectileID.EbonsandBallGun, PearlsandBlock, ProjectileID.PearlSandBallGun, CrimsandBlock, ProjectileID.CrimsandBallGun);
+
+		/// <summary>
+		/// The damage a specific sandgun ammo should increase by. Defaults to 0.
+		/// </summary>
+		public static int[] SandgunAmmoDamage = Factory.CreateIntSet(0, EbonsandBlock, 5, PearlsandBlock, 5, CrimsandBlock, 5);
```
steviegt6: Should this be titled `SandgunAmmoBonusDamage` instead? Something to communicate its additive properties.

steviegt6 (changes requested):

patches/tModLoader/Terraria/ID/TileID.TML.cs
blahblahbal: It needs it cause there's a TileID.Sets.Mud as well (probably used for the jungle? I didn't actually check)

patches/tModLoader/Terraria/ID/ItemID.TML.cs
blahblahbal: Yeah that's probably better actually. I'll change it.

patches/tModLoader/Terraria/Player.cs.patch
blahblahbal: Ah, alright. Couldn't find anything about this specifically in the guidelines page (might've missed it though). I'll change it to comment out the code instead of just deleting it.

blahblahbal:
Just pushed my changes. If there's anything else that I missed or need to do please let me know.

patches/tModLoader/Terraria/ID/TileID.TML.cs
Rijam: Mud Blocks were affected by gravity in 1.0.5, then it was reverted in 1.0.6 (same with Ash) https://terraria.wiki.gg/wiki/Mud_Block#History
What tomat is talking about is how you put `TileID.Mud` instead of just `Mud`.

blahblahbal: Yeah, in TileID there's a `TileID.Sets.Mud`, which if I use just `Mud`, it gives an error, since `TileID.Sets.Mud` is a bool array and can't be used as an `int`. The exact error is "Argument 6: cannot convert from 'bool[]' to 'int'".

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7441,6 +7441,27 @@
  		int pickedProjectileId = -1;
  		if (PickAmmo_TryFindingSpecificMatches(sItem.type, item.type, out pickedProjectileId))
  			projToShoot = pickedProjectileId;
+@@ -41015,7 +_,8 @@
+ 			projToShoot = 117;
+ 
+ 		if (projToShoot == 42) {
++			/*
+-			if (item.type == 370) {
++			 if (item.type == 370) {
```
steviegt6: Remove the space.

steviegt6 (changes requested):

patches/tModLoader/Terraria/Player.cs.patch
blahblahbal: Fixed. I'm not sure how I missed that one when I didn't miss the other space in the comment in WorldGen.cs.

steviegt6 (approved):

jjohnsnaill:
Is `SandgunAmmoToProjectile` necessary? `AmmoID.Sets.SpecificLauncherAmmoProjectileMatches` should be sufficient, and if a mod adds its own sand-shooting weapons, it could simply reference the Sandgun's dictionary.

JavidPack:
I've made a lot of changes. The original AI code seemed to be based on an older Terraria version, it had some rare edge cases where things behaved slightly different (regarding actuators I think). It also seemed felt weird to me that the tile and projectile code still required large sections of copied code, so I updated the sets and patches to support things more directly.

Please test my changes and let me know if the set names make sense. Let me know if the new approach seems reasonable or any other feedback.

blahblahbal:
Seems to all work fine from my tests. And the set names make sense to me as well. Didn't realize that the projectile AI code was outdated - I thought I copied it directly from vanilla 1.4.4.9. Unless you meant the old ExampleMod code? Anyway, I don't have any other feedback; your approach seems reasonable enough as well.

Issue #3906, opened 2023-11-01
Fishing Bobbers incompatible with modded fishing rods


JavidPack:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

Fishing bobbers, https://terraria.wiki.gg/wiki/Fishing_Bobbers, added in 1.4.4, replace the bobber projectile spawned by fishing rods. 

This causes issues with custom fishing rods because the line origin and line color are tied to the projectile, not the item. This was how it was initially implemented (https://github.com/tModLoader/tModLoader/pull/1460).

Looking at the DrawProj_FishingLine code, it is clear that the correct approach would be to have the line color and line origin tied to the fishing rod item, not the bobber projectile. All vanilla logic in the method is based off of fishing rod item.

The fix is likely to Obsolete the ModProjectile hooks and move them to the ModItem. Will want to double check with existing mods and see if that has any issues.

### Steps to reproduce

Equip modded fishing rod, bait, throw. Line and color should match code.

![image](https://github.com/tModLoader/tModLoader/assets/4522492/83d1a560-507e-49e1-8782-52595c95e4d8)

Equip Fishing Bobber as well. Now throw. Line color and origin are now wrong.

![image](https://github.com/tModLoader/tModLoader/assets/4522492/8c107668-d0a1-4205-a008-c5ac36d33d99)

JavidPack:
Fixed by #4060

PR #4060, opened 2024-02-10
Fix incorrect fishing line offset with vanilla bobbers on modded rods.

RegularPhoenix:
### What is the bug?
[#3906]
Since the fishing line offset of modded rods was bound to bobbers, vanilla bobber accessories that replaced the original projectile of modded rods were causing incorrect offset.

Vanilla rods:
image

Modded rods:
image

Fixed modded rods:
image

### How did you fix the bug?
Move ModifyFishingLine() hook to the ModItem class. This, alas, has no backward compatibility.

Since vanilla bobber projectiles completely replace modded ones, there probably aren't any.
Additionally, ExampleFishingRod and ExampleBobber should also be updated.

### DIFF
--- a/ExampleMod/Content/Items/Tools/ExampleFishingRod.cs
+++ b/ExampleMod/Content/Items/Tools/ExampleFishingRod.cs
@@ -48,6 +48,15 @@ namespace ExampleMod.Content.Items.Tools
 			return false;
 		}
 
+		public override void ModifyFishingLine(ref Vector2 lineOriginOffset, ref Color lineColor) {
+			// Change these two values in order to change the origin of where the line is being drawn.
+			// This will make it draw 47 pixels right and 31 pixels up from the player's center, while they are looking right and in normal gravity.
+			lineOriginOffset = new Vector2(47, -31);
+
+			// Sets the fishing line's color. Note that this will be overridden by the colored string accessories.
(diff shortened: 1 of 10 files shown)


JavidPack:
Looking go so far. We will want ExampleFishingRod and ExampleBobber fixed in this PR as well. Also, we want existing mod usage to continue working, so make sure the existing modprojectile hook is still being called, but mark it as [Obsolete("", error: true)] so that it must be fixed for the mod to be rebuilt.  

You'll want to move ProjectileLoader.ModifyFishingLine to ItemLoader as well. (You'll be calling an Obsolete ProjectileLoader.ModifyFishingLine from ItemLoader.ModifyFishingLine to preserve existing functionality. )

Once that is done, I'll work on the tModPorter steps, I think we can do it with a HookRemoved entry. If you are adventurous and want to try doing the tModPorter and corresponding tests, you can, but it's a bit finnicky and I can do it when I merge this PR.

RegularPhoenix:
As a side note, since the line color is now depending on the rod, I suppose it will be more difficult to make different line colors on rods that fire multiple bobbers.  At least I could not come up with any way, so I made the ExampleRod use a single color for now.

JavidPack:
Looks good. I made some more changes to add back in the original behavior. Thanks a lot for spearheading this.

PR #4061, opened 2024-02-11
TileID.Sets.PlanterBoxes

Lion8cake:
### What is the new feature?
TileID.Sets.PlanterBoxes is a TileID set for all Planter Boxes. This allows any mod to add ``TileID.Sets.PlanterBoxes[Type] = true;`` and their tile will behave similar to vanilla planter boxes.

These changes include:
Smart Cursor
Herb and Grass growth/placement
Tile Framing
AnchorType.PlanterBox
Grass and Dirt spreading through tiles
Skeletron Prime bombs not falling through

This allows for quite easy implementation of the modder's own planter box for their herb and allows for a bunch of vanilla code to run for their tile.

You need to make 10-12 different IL edits to get a perfectly working planter box. This is not an option for some modders that either don't want to use IL edits or are too new to modding/C# to know how to make an IL edit, this isn't even to mention the amount of IL edits needed or how big some of them would be.
This also allows for modded planter boxes to merge together as currently only modded ones can merge with vanilla planter boxes.

To my knowledge no, I do think this is the best way of adding support for modded planter boxes. Adding planter boxes to the herb anchor in TileObjectData and referencing that for PlantAlch in WorldGen.cs would be a alternate approach to the herb placement issue if we don't need a whole TileID Set for this.

```using Microsoft.Xna.Framework;
using Terraria;
using Terraria.ID;
using Terraria.ModLoader;

namespace ExampleMod.Content.Tiles
{
	public class ExamplePlanterBox : ModTile
	{
		{
			Main.tileFrameImportant[Type] = true;
			Main.tileTable[Type] = true;
			Main.tileSolid[Type] = true;
			Main.tileSolidTop[Type] = true;
			AddMapEntry(new Color(200, 200, 200));
			TileID.Sets.DisableSmartCursor[Type] = true;
			TileID.Sets.IgnoresNearbyHalfbricksWhenDrawn[Type] = true; //All vanilla planter boxes have this
			TileID.Sets.PlanterBoxes[Type] = true; //This does a majority of the work for us,
												   //both allowing vanilla herbs to attach to our planterbox amongst a list of other things
			AdjTiles = new int[] { TileID.PlanterBox };
		}

		public override bool Slope(int i, int j)
		{
			return false; //We dont want our planter box to be slope-able so we return false
		}
	}
}
```

### ExampleMod updates
Added Example Planter Box (Item and Tile), 
This will show new modders how to easily add a planter box alongside their own herb.


steviegt6 (approved):

ExampleMod/Content/Items/Placeable/ExamplePlanterBox.cs
```diff
@@ -0,0 +1,28 @@
...
+using Terraria.GameContent.Creative;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Items.Placeable
+{
+	public class ExamplePlanterBox : ModItem
+	{
+		public override void SetStaticDefaults() {
+			CreativeItemSacrificesCatalog.Instance.SacrificeCountNeededByItemId[Type] = 100;
```
BasicallyIAmFox:
```suggestion
			Item.ResearchUnlockCount = 100;
```

ExampleMod/Content/Items/Placeable/ExamplePlanterBox.cs
```diff
@@ -0,0 +1,28 @@
...
+			Item.height = 12;
+			Item.maxStack = Item.CommonMaxStack;
+			Item.useTurn = true;
+			Item.autoReuse = true;
+			Item.useAnimation = 15;
+			Item.useTime = 10;
+			Item.value = Item.buyPrice(silver: 1);
+			Item.useStyle = ItemUseStyleID.Swing;
+			Item.consumable = true;
+			Item.createTile = ModContent.TileType<Content.Tiles.ExamplePlanterBox>();
```
BasicallyIAmFox:
```suggestion
			Item.DefaultToPlaceableTile(ModContent.TileType<Content.Tiles.ExamplePlanterBox>());
			Item.width = 12;
			Item.height = 12;
			Item.value = Item.buyPrice(silver: 1);
```

ExampleMod/Content/Tiles/ExampleHerb.cs
```diff
@@ -58,7 +58,8 @@ public class ExampleHerb : ModTile
 			};
 			TileObjectData.newTile.AnchorAlternateTiles = new int[] {
 				TileID.ClayPot,
-				TileID.PlanterBox
+				TileID.PlanterBox,
+				ModContent.TileType<ExamplePlanterBox>()
 			};
```
direwolf420, line 63: This is a bit tricky because ideally it should include _all_ modded planter boxes, not just vanillas + the one from the same mod. I'm not sure what the solution here would be except manually change the tileobjectdata somewhere around `TileLoader.PostSetupTileMerge` (when every mod has populated `TileID.Sets.PlanterBoxes`) for every modded tile that has `TileID.PlanterBox` in its `AnchorAlternateTiles`

direwolf420:
do modded planter boxes now allow torches to be placed on them?

ExampleMod/Content/Tiles/ExampleHerb.cs
Tyfyter: It would require redoing everything that uses AnchorAlternateTiles or making a wrapper class that overrides equality comparison with ints, but making AnchorAlternateTiles support sets (or adding a separate field for sets) would solve this issue and potentially future ones.

PR #4076, opened 2024-02-24
feat: Increase maximum supported length of chest names

### LINKED ISSUE #3909: Increase maximum supported length of chest names

No

Gameplay as a Player

### Description

There's a character limit on how long a chest name can be (20), which can be a problem if
1. The chest itself has a long name. This is in the realm of possibilities as one vanilla chest has a length of 19, 1 below the limit. If a mod adds a material with a corresponding chest it will say `VeryLongMaterialName Chest` but gets cut off.
2. Chat tags are used with modded items - Write `[i:2]` which displays dirt block, `[i:ModName/ItemName]` which displays ItemName. You can see immediately that the latter is almost impossible unless the mod and/or the item has a short internal name. This is useful for players that want to have a better visual feedback for item sorting purposes.

Showcase that 2. works with modded items:
[chest name is "[i:ThoriumMod/Blood]", exactly 20](https://cdn.discordapp.com/attachments/445276626352209920/1136990391539879996/Screenshot_6.png?ex=654e19d1&is=653ba4d1&hm=56524e9285310309414ea8b016ecf7fab5f67156ba3ca1c9c8f537fcba9dc199&)

The proposal is to increase the limit.

Make longer chest names and item tags with modded items possible.

Remove the limit entirely

Metratrj:
### What is the new feature?
Increase the MaxNameLength for chest names and implemented the usage of this variable instead of hardcoded (magic)number in the ChestUI and UIVirtualKeyboard.

Feature was requested by @direwolf420 in #3909. 
To quote him:
- > The chest itself has a long name. This is in the realm of possibilities as one vanilla chest has a length of 19, 1 below the limit. If a mod adds a material with a corresponding chest it will say VeryLongMaterialName Chest but gets cut off.

- > Chat tags are used with modded items - Write [i:2] which displays dirt block, [i:ModName/ItemName] which displays ItemName. You can see immediately that the latter is almost impossible unless the mod and/or the item has a short internal name. This is useful for players that want to have a better visual feedback for item sorting purposes.

We could increase the hardcoded values, but it does not make much sense to me because the variable MaxNameLength already exist in the Chest class and take usage should make the thing easier to change for modders too.

Rename chest with longer names ;)

### ExampleMod updates
None

Closes #3909


Metratrj:
@JavidPack, @Solxanich & @direwolf420 what do you think?

patches/tModLoader/Terraria/Chest.cs.patch
```diff
@@ -84,7 +84,8 @@
 +	/// <summary>
 +	/// The maximum length that <see cref="name"/> is allowed to be.
 +	/// </summary>
- 	public const int MaxNameLength = 20;
+-	public const int MaxNameLength = 20;
++	public const int MaxNameLength = 63;
```
direwolf420, line 88: Regarding it not being used as `MaxNameLength` in the source: Since tml uses decompiled vanilla code, `const` only appear in their literal value.
Regarding where it's used, look at what <https://github.com/NotLe0n/BetterChests/blob/1.4.4/src/Edits/ChestNameEdits.cs> edits, you are missing some places

patches/tModLoader/Terraria/UI/ChestUI.cs.patch
```diff
@@ -9,6 +9,15 @@
  using Terraria.UI.Chat;
  using Terraria.UI.Gamepad;
  
+@@ -77,7 +_,7 @@
+ 		if (Main.editChest) {
+ 			text = Main.npcChatText;
+ 			Main.instance.textBlinkerCount++;
+-			if (Main.instance.textBlinkerCount >= 20) {
++			if (Main.instance.textBlinkerCount >= Chest.MaxNameLength) {
```
direwolf420: This is a bad edit, it's a timer for the text blinker when inputting the chest name

direwolf420 (commented):
Review

patches/tModLoader/Terraria/UI/ChestUI.cs.patch
Metratrj: Makes sense... reverted it

patches/tModLoader/Terraria/Chest.cs.patch
Metratrj: Added missing changes but utilize further this const

Solxanich:
Tested, great work. 
Thanks for the contribution!

PR #4079, opened 2024-02-24
feat: journeys end menu persist on restart

### LINKED ISSUE #4015: Make menu themes persist on restart

No

Gameplay as a Player

### Description

Make the "Change Menu Theme:" option that users can change on the main menu persist between game restarts.

It aims to have a customization option persist. It's kind of a useless feature otherwise

_No response_

Metratrj:
### What is the new feature?
When last selected menu was journeys end it also get loaded on startup

To complete the menu saving we already handle + requested in #4015 

Open for suggestions

1. Open Terraria
2. Switch to Journeys End menu theme
3. Close Terraria
4. Open again
5. Voilà 

Tested also with skip loading mods (holding shift while start)
Close #4015

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/MenuLoader.cs
@@ -73,17 +73,11 @@ public static class MenuLoader
 			Main.instance.playOldTile = true; // If the previous menu was the 1.3.5.3 one, automatically reactivate it.
 		}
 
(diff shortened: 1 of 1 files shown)


Metratrj:
@JavidPack are any further changes needed? Maybe invert the actual if clause, but it seemed to better for readability.

JavidPack:
Thanks for doing this. I simplified the logic for the sake of readability. Having the branching logic make it a bit harder to parse.

PR #4086, opened 2024-02-29
Track and report lost player world save data, Fixes #4082

### LINKED ISSUE #4082: Handled modded Player/World save data being lost better.
As a follow-up to #4067 , we want to provide some safety nets for users experiencing save issues. We need users to be informed of the issues and give them an opportunity to restore to an earlier player/world.

After some discussions, CB suggested the following:

We would like a clear way for a user to:
- Notice there's been an issue saving
- Assess the damage (load the player or world and see what data has been lost)
- Restore to the previous fully intact save

I think at a minimum we should:
- Alert the user when saving encounters an issue
- Provide an easy way to navigate to backups from the player/world select menu
- Provide an easy way to identify the last fully intact backup

This would entail adding Main.NewText messages, flagging a save as potentially incomplete, updating the UI to show a save is incomplete, and ensuring that autosaves don't overwrite complete saves. For example, a player might autosave in-game after 10 minutes or after each death. A mod with broken ModPlayer.SaveData code shouldn't cause the player to be unrecoverable. It will be up to the user if they want to restore the last complete save or keep their progress while losing the data from the erroring ModPlayer class.

# Why
Prior to 4067, these sort of errors would just crash the game, so most mods should currently be coded correctly. New mods and updates, however, are more likely to include bugs in SaveData code since the current implementation doesn't crash outright anymore. 

# Additional
Also, SaveWorldHeader and LoadWorldHeader probably need similar try catch.

JavidPack:
Fixes #4082

Player and World related save errors (`ModPlayer.SaveData`, `ModSystem.SaveWorldHeader`, `ModSystem.SaveWorldData`) are tracked and stored in the .tplr and .twld files. These errors are presented to the user after they are encountered, at the player/world select screen, and when joining a world. The errors will persist until the next game session since they are stored in the save files. The error messages should be sufficient for a modder or user to know something has gone wrong with their player or world. Loading an erroring world or player is logged in the logs, hinting that an enabled mod is malfunctioning. A backup file menu in a future follow-up could be used for users to rollback, but for now it is up to the user to report the issue to the modder and roll back if they desire.

## Showcase
In this example, the user has a mod that has saving errors in both a ModPlayer and a ModSystem. 

1. As they leave the game, these errors are presented to the user.
2. The same errors are present in the world and player select screens.
3. The same errors are shown in chat when the player enters the game.

https://github.com/tModLoader/tModLoader/assets/4522492/3d5c002d-c11f-4df2-8089-b9bd04eb3a16

vlc_2024-03-06_11-48-28
vlc_2024-03-06_11-48-41
vlc_2024-03-06_11-48-48
vlc_2024-03-06_11-49-22


patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -519,8 +536,23 @@ internal static List<TagCompound> SaveModData()
 
 	internal static void LoadModData(IList<TagCompound> list)
 	{
+		Main.worldLoadModDataErrors = null;
```
Chicken-Bones: May as well put it in `WorldIO` instead of `Main`

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -519,8 +536,23 @@ internal static List<TagCompound> SaveModData()
...
+				if (tag.TryGet<string>("error", out string errorMessage)) {
+					string message = Language.GetTextValue("tModLoader.WorldCustomDataSaveFail") + "\n" + errorMessage;
+
+					Logging.tML.Warn(message);
+
+					// TODO: Unable to NewText message here. All messages are cleared later when the world finishes loading.
+					Main.worldLoadModDataErrors = Main.worldLoadModDataErrors == null ? errorMessage : Main.worldLoadModDataErrors + "\n" + errorMessage;
+
+					if (Main.dedServ)
+						Console.WriteLine(message);
```
Chicken-Bones: `Utils.LogAndConsoleErrorMessage`? (or `Info`)

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -662,7 +694,25 @@ private static TagCompound SaveModHeaders()
...
+				Logging.tML.Error(message.ToString());
+
+				if (Main.gameMenu && Main.menuMode == 10) {
+					// Save and Quit. Since this isn't on main thread, attempting Interface.errorMessage.Show behaves poorly because the menu will automatically change while the user is reading the message. For now don't do anything.
+				}
+				else if (!Main.gameMenu) {
+					// In-game autosave
+					ChatHelper.BroadcastChatMessage(message, Color.OrangeRed); // Handles SP and Server cases.
+				}
+
```
Chicken-Bones: Can we commonise this code with `PlayerIO`? I think `ChatHelper.BroadcastChatMessage` doesn't actually broadcast if called from client so no need for two different `NewText` calls

patches/tModLoader/Terraria/ModLoader/IO/WorldIO.cs
```diff
@@ -662,7 +694,25 @@ private static TagCompound SaveModHeaders()
...
-			system.SaveWorldHeader(saveData);
+			try {
+				system.SaveWorldHeader(saveData);
+			}
+			catch {
+				var message = NetworkText.FromKey("tModLoader.SaveWorldHeaderExceptionWarning", system.Name, system.Mod.Name);
+				Logging.tML.Error(message.ToString());
+
+				if (Main.gameMenu && Main.menuMode == 10) {
+					// Save and Quit. Since this isn't on main thread, attempting Interface.errorMessage.Show behaves poorly because the menu will automatically change while the user is reading the message. For now don't do anything.
```
Chicken-Bones: Could do an 'on return to menu' thing

patches/tModLoader/Terraria/ModLoader/IO/PlayerIO.cs
```diff
@@ -239,6 +255,13 @@ internal static void LoadModData(Player player, IList<TagCompound> list)
 			string modName = tag.GetString("mod");
 			string modPlayerName = tag.GetString("name");
 
+			if(tag.TryGet<string>("error", out string errorMessage)) {
```
Chicken-Bones: Space between `if` and opening bracket

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPlayer.cs
```diff
@@ -26,4 +27,12 @@ public override void LoadData(TagCompound tag)
 		PlayerIO.LoadModData(Player, tag.GetList<TagCompound>("list"));
 		PlayerIO.LoadResearch(Player, tag.GetList<TagCompound>("unloadedResearch"));
 	}
+
+	public override void OnEnterWorld()
+	{
+		if (!string.IsNullOrWhiteSpace(Main.worldLoadModDataErrors)) {
```
Chicken-Bones: Good

patches/tModLoader/Terraria/GameContent/UI/Elements/UICharacterListItem.cs.patch
```diff
@@ -74,6 +82,15 @@
  
  		((UIImageButton)evt.Target).SetImage(_data.IsCloudSave ? _buttonCloudActiveTexture : _buttonCloudInactiveTexture);
  		if (_data.IsCloudSave)
+@@ -216,6 +_,8 @@
+ 
+ 	private void PlayGame(UIMouseEvent evt, UIElement listeningElement)
+ 	{
++		if(Data.Player.saveErrorMessage != null)
```
Chicken-Bones: Should log the error message itself.

Chicken-Bones (changes requested):

Chicken-Bones (approved):

Chicken-Bones (changes requested):
Overall looks great. Can you post some screenshots in the description?

Before you merge:
- `Player.ModSaveErrors.Clear()` is needed in `PlayerIO.Save`
- Run a search for any `if(` without the space before the opening bracket.

PR #4096, opened 2024-03-06
Fix incorrect date handling & other logging issues

### LINKED ISSUE #4025: Archived log files use incorrect dates
### Version

Latest 1.4.4-stable, dev 1.4.4.

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

The LogArchiver seems to use files' Creation Date in names of its archives, but it just so happens that it doesn't get updated when a TML client is started a new, and thus people get 3-4 digit suffixed archives using the date of pretty much the first time they've installed TML.

image
image-1

### Log File

_No response_

### Steps to reproduce

Theoretical:
1. Get a fresh TML install.
2. Launch-close TML 2-3 times.
3. Change your date to a future day.
4. Launch-close TML 2-3 times again.
5. The new archives do not use the new date.

### Expected Behavior

Archives get the date of the time that session's start or end.

### Actual Behavior

Archives pile up under an arbitrary date.

### Reproduction frequency

_No response_

### Additional Information

_No response_

Mirsario:
Resolves #4025 by doing the following:
- [x] Force-updates creation dates of `(client|server|environment-*).log` files on startup. Log4net does not do that, and envdumps used streamed writes which especially wouldn't have done that.
- [x] Prevents `LogArchiver` from throwing silent exceptions whenever it encounters numbered `.log.old` files, halting its function.
### Additionally:
- [x] Prevents `environment-` files from being created during mod packaging:
image
(They're also deleted if found)

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Logging.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Logging.cs
@@ -55,7 +55,8 @@ public static partial class Logging
 		try {
 			InitLogPaths(logFile);
 			ConfigureAppenders(logFile);
-			ForceUpdateLogCreationDates();
+			// Force-update log4net file's creation date, because log4net does not do that.
+			TryUpdatingFileCreationDate(LogPath);
 		}
 		catch (Exception e) {
 			ErrorReporting.FatalExit("Failed to init logging", e);
@@ -257,7 +258,11 @@ public static partial class Logging
 	{
 		try {
 			string fileName = $"environment-{Path.GetFileName(LogPath)}";
-			using var f = File.OpenWrite(Path.Combine(LogDir, fileName));
+			string filePath = Path.Combine(LogDir, fileName);
+
+			TryUpdatingFileCreationDate(filePath);
+
+			using var f = File.OpenWrite(filePath);
 			using var w = new StreamWriter(f);
 			foreach (var key in Environment.GetEnvironmentVariables().Keys) {
 				w.WriteLine($"{key}={Environment.GetEnvironmentVariable((string)key)}");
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/Logging.cs
```diff
@@ -77,7 +81,8 @@ internal static void LogStartup(bool dedServ)
 			tML.InfoFormat("Parsed Launch Parameters: {0}", string.Join(' ', Program.LaunchParameters.Select(p => ($"{p.Key} {p.Value}").Trim())));
 		}
 
-		DumpEnvVars();
+		if (!Program.LaunchParameters.ContainsKey("-build"))
```
Chicken-Bones: Why do we run `LogStartup` during build at all? What's the point if appenders haven't been configured?

patches/tModLoader/Terraria/ModLoader/Logging.cs
```diff
@@ -47,6 +47,8 @@ public enum LogFile
 
 	internal static void Init(LogFile logFile)
 	{
+		LegacyCleanups();
```
Chicken-Bones, line 50: Maybe move this after the `-build` guard

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Logging.cs
Mirsario: It's in front of the guard because `-build` produced the file being removed there. Not really expensive so shouldn't matter.

patches/tModLoader/Terraria/ModLoader/Logging.cs
Mirsario: Thought that this at least produces stdout, but no. Will short-circuit.

PR #4101, opened 2024-03-10
Minor performance improvement

Golfing7:
### What is the bug?
Not a bug by definition, but this is the closest template to what I could find.
performance

From profiling, the call to `LoaderManager.Get` can use up to 1.65% of the frame time while running.

### How did you fix the bug?
Extracting the condition out of the for-loop condition will reduce the amount of calls to only once per invocation.

As this is minor edit, I don't believe there's anything else that should be done.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
@@ -184,7 +184,7 @@
 +					tileStyle = sandType + 1;
 +					break;
 +				}
-+					
++
 +				if (tile.frameX == 204 || tile.frameX == 202) {
 +					var asset = PlantLoader.GetCactusFruitTexture(sandType);
 +					if (asset != null)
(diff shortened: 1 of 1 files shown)


steviegt6 (approved):
Some issues with removing irrelevant whitespace but I'd rather just get this merged.

JavidPack:
I'm a bit suspect about that profile, I was not able to replicate your findings of such a huge runtime for that line of code. I see that you have Calamity loaded and that it includes IL edits to methods nearby, I wonder if that is somehow messing with how the profiler collects and positions profiling samples.

I don't mind the changes, but there are other places with similar code that would also be useful to fix if this sort of for-loop structure is an issue, so I'd like to confirm that it really is worth the effort. Can you disable mods and get a new profile? Also, could you show a screenshot of the game view so I have an idea about how much water is on screen where you did the profiling.

Golfing7:
Sorry for the delay, I was able to reproduce it in this environment: 
0d32b4541bf51d06c8198f6841ab1f51
It requires the liquid to be drawn behind blocks.
add3ea696c83dc4c18e7fa3325ab3691 (1)
This was taken without any content mods enabled. I only had a few mods for testing enabled.

Chicken-Bones:
Thanks. The `LoaderManager.Get<T>` method should really be optimized to use the same technology as `ModContent.Get` but that can be another PR.

PR #4103, opened 2024-03-11
Prevent config.json corruption and workaround existing corruption.

JavidPack:
Implements various workarounds for config.json corruption issues that prevent version migration from processing correctly.

config.json should now no longer be corrupted due to computer crashes while saving.

config.json file write timestamp is used to make assumptions about the status of porting. A recent edit suggest that a tModLoader version with the porting logic has already been executed.

A final fallback is a prompt for the user to select which version has been launched, allowing the logic to proceed.

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -109,7 +109,7 @@ public static partial class Program
 	/// maxVersionOfSource is used to determine if we even should port the files. Example: 1.4.3-Legacy has maxVersion of 2022.9
 	/// isAtomicLockable could be expressed as CopyToNewlyCreatedDestinationFolderViaTempFolder if that makes more sense to the reader.
 	/// </summary>
-	private static void PortFilesFromXtoY(string superSavePath, string source, string destination, string maxVersionOfSource, bool isCloud, bool isAtomicLockable)
+	private static void PortFilesFromXtoY(string superSavePath, string source, string destination, string maxVersionOfSource, bool isCloud, bool isAtomicLockable, DateTime migrationDay)
 	{
 		string newFolderPath = Path.Combine(superSavePath, destination);
 		string newFolderPathTemp = Path.Combine(superSavePath, destination + "-temp");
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -169,8 +169,28 @@ private static void PortFilesFromXtoY(string superSavePath, string source, strin
...
+				e.HelpLink = "https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Usage-FAQ#configjson-corrupted";
+				ErrorReporting.FatalExit($"Attempt to Port from \"{oldFolderPath}\" to \"{newFolderPath}\" aborted, the \"{sourceFolderConfig}\" file is corrupted.", e);
+			}
+		}
+
+		if (string.IsNullOrEmpty(lastLaunchedTml)) {
+			// If the config.json is missing LastLaunchedTModLoaderVersion entry, we can ask the user. (Most likely the user copied Terraria/config.json over)
+			int result = ErrorReporting.ShowMessageBoxWithChoices(
+				title: "Failed to read config.json configuration file",
+				message: "Your config.json file is incomplete.\n\nPlease select one of the following options and the game will resume loading:\n\nWhat is the highest version of tModLoader that you have launched?",
```
Mirsario: This should be localizable.

patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -169,8 +169,28 @@ private static void PortFilesFromXtoY(string superSavePath, string source, strin
...
+				// If the file was edited recently, we assume that an updated tModLoader edited it. This should ignore modifications made long ago by pre-migration logic tModLoader releases.
+				lastLaunchedTml = BuildInfo.tMLVersion.ToString();
+			}
+			else {
+				e.HelpLink = "https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Usage-FAQ#configjson-corrupted";
+				ErrorReporting.FatalExit($"Attempt to Port from \"{oldFolderPath}\" to \"{newFolderPath}\" aborted, the \"{sourceFolderConfig}\" file is corrupted.", e);
+			}
+		}
+
+		if (string.IsNullOrEmpty(lastLaunchedTml)) {
```
Mirsario: Why have this `else` block and not show the choice box in case the config *is* corrupted? Why not auto-delete it at that point too?

Mirsario (changes requested):

patches/tModLoader/Terraria/Program.TML.cs
Solxanich: All this occurs before localization is loaded - before save paths are finalized.
There was an issue on this last version as well.

patches/tModLoader/Terraria/Program.TML.cs
JavidPack: Added. No need to auto-delete, it'll be replaced during game loading anyway.

patches/tModLoader/Terraria/Program.TML.cs
JavidPack: We can add manual localization if we really need to later, but we should see this issue a lot less for now

PR #4109, opened 2024-03-14
Zip library updated from Ionic.Zip.Reduced v1.9.1.8 to DotNetZip v1.16.0

JavidPack:
Fixes a bug that would cause world saving to hang indefinitely for some users in some specific worlds (https://discord.com/channels/103110554649894912/1216826270818504714 is the support thread where the solution was finally discovered). There were mentions of similar bugs on the bug tracker that were fixed `https://github.com/haf/DotNetZip.Semverd/issues/38` in the zip library we use. This fix is later than our original v1.9.1.8 release. When testing the updated dotnetzip library with the world that wasn't able to save, the issue appears fixed now. 

It is unclear how many are affected by this bug, but the save issue has been an unknown issue that has stumped us for some time now, it just hadn't been replicated by devs until now.

It seems that the library has gone through a rename in the intervening years. Namespaces and classnames are consistent, so source code compatibility is maintained. Binary compatibility is not due to the assembly name change, so additional code has been added to resolve the old assembly name to the new one.

# Porting Notes
Nothing should be required by modders.


JavidPack:
For the record, this is the mod used to verify the mod still loads and works after replacing the dll. It's possible that maybe some other classes or methods have changed that are not included in this test.

```cs
using Ionic.Zip;
using System.Diagnostics;
using System.IO;
using System.Text;
using Terraria;
using Terraria.ModLoader;

namespace ZipTest
{
	public class ZipTest : Mod
	{
		public override void Load() {
			var data = GetFileBytes("My Zips/the zip file.zip");
			using (var zip = ZipFile.Read(new MemoryStream(data)))
				zip.ExtractAll(Path.Combine(ModLoader.ModPath, "ZipTestOutput"));

			string zipFilePath = Path.Combine(ModLoader.ModPath, "ZipTestInput");
			string zipFileOutput = Path.Combine(ModLoader.ModPath, "output.zip");

			using (var zip = new ZipFile(zipFileOutput, Encoding.UTF8)) {
				zip.AddDirectory(zipFilePath, "");
				zip.Save();
			}
		}
	}
}
```

PR #4112, opened 2024-03-14
Unload dependents of a mod which failed to load (Redo)

JavidPack:
Note: Copied over from #3829 because we were unable to update the PR with additional commits. 

In addition, the UI has been updated to reflect dependents and dependencies more consistently

https://github.com/tModLoader/tModLoader/assets/4522492/66213617-a2fb-4b94-a14a-46ba1475d262

### What is the bug?
#3821 - If a mod which is a hard dependency of other mods is unloaded due to a load error, the dependency mod is unloaded while the dependent mod(s) are not unloaded. This can cause a load loop since the culprit mod will be loaded again when dependencies are automatically enabled.

### How did you fix the bug?
With this pull request, when a mod is unloaded because of a load error, the mod loader will search for mods which depend on it and disable them. It does this by checking if any of the `availableMods` have the dependency mod in the `modReferences`.

I'm not aware of any other way to get the dependents of a mod other than doing the inverse and searching the dependencies of other mods to find them. Another way to get the dependents would likely be more logical for reading the code, and it would be better if not every mod would have to be checked.

I'm also unsure about the message localization. I added a string to the en-US localization so that the message will indicate which dependents were disabled, if any, but I'm unfamiliar with how localization works in tModLoader. Is it OK for me to only add it to the en-US file?


JavidPack:
Current text output.

dotnet_2024-03-27_16-28-56    

Contrived multiple mod example, it's actually really hard to get multiple mods at fault in a single exception
dotnet_2024-03-27_17-03-21

patches/tModLoader/Terraria/Localization/Content/zh-Hans/tModLoader.json
```diff
@@ -170,6 +170,8 @@
 		// Mod Errors
 		"LoadError": "该模组加载时发生错误 {0}",
 		"LoadErrorDisabled": "该模组已被自动禁用.",
+		// "LoadErrorDependentsDisabled": "The following mods have also been disabled because of dependencies to {0}:\n    {1}",
```
Chicken-Bones: `because they depend on`

Chicken-Bones (commented):
Honestly. I don't think a new message in the error window is needed. There's enough text that isn't the error there already.

We could just change `the mod(s) have been automatically disabled` to `the mod(s) and any dependencies have been automatically disabled`

It would be good to double check that enabling/disabling in the mods UI also enables and disables any dependencies. Perhaps with a tooltip like:
`Disable mod and 3 dependents`
`Enable mod and 3 dependencies`

Chicken-Bones (approved):

PR #4120, opened 2024-03-17
Fix #4106/#2921, ItemID Sets implementation.

### LINKED ISSUE #2921: [Feature Request] Modify "Journey Mode Duplication menu" item availability via tabs

No

Mod capability as a Modder

### Description

**[Vanilla wiki link for reference](https://terraria.wiki.gg/wiki/Journey_Mode#Duplication)**

It is currently not possible to control in which tab items go in the Journey Mode Duplication menu. Some filters check for fields on the item, some are using private & hardcoded sets, or both (found in `ItemFilters.cs`).
Currently, you have to detour the `FitsFilter` method in the desired filter and adjust the return value as desired.

Use case:
* Adding a custom bug net item to the "Tools" tab

Make the Journey Mode Duplication menu more intuitive to use for the player.

No solution has been concidered so far, just pointing out a gap in the API.

### LINKED ISSUE #4106: Add a hook for manually setting a filter/category type to an item for use in the Duplication Menu

Yes

Gameplay as a Player

### Description

A new hook called (`ModifyItemEntryFilter` or `ModifyItemFilter`) that would allow for modders to manually set their item to any filter/category in the Duplication Menu.

It is currently impossible to manually set an item's filter/category without **detouring** the `Terraria.GameContent.Creative.On_ItemFilters.<Filter/Category>.FitsFilter` method. For say making your modded item go under the "Tools" filter in the Duplication Menu.

image

Could also be a `SetStaticDefaults` or `SetDefaults` thing

JackerAcid:
(Fixes https://github.com/tModLoader/tModLoader/issues/4106, Fixes https://github.com/tModLoader/tModLoader/issues/2921)
### What is the new feature?
Added a new ItemID.Sets called `DuplicationMenuToolsFilter` which allows modders to manually set their item to be filtered under the "Tools" filter in Journey Mode's duplication menu.
image
There is currently no possible way to easily set this filter for your item without detouring the  `Terraria.GameContent.Creative.On_ItemFilters.Tools.FitsFilter` method.
This current implementation **only** allows for the "Tools" filter to be manually set for your item.
If you would like more customization for which filters apply to your item, another approach would probably need to be taken.
```cs
    ItemID.Sets.DuplicationMenuToolsFilter[Item.type] = true;
}
```
### ExampleMod updates
Updated ExampleMod/Content/Items/Tools/ExampleBugNet.cs 2b2796594263dc96760f14b6a69e9ad97024a26c


patches/tModLoader/Terraria/ID/ItemID.cs.patch
```diff
@@ -334,6 +334,12 @@
 +		/// </remarks>
  		public static bool[] IsAPickup = Factory.CreateBoolSet(58, 184, 1734, 1735, 1867, 1868, 3453, 3454, 3455);
 +
++		/// <summary> Indicates that an item is to be filtered under the "Tools" filter in Journey Mode's duplication menu.
++		/// <br/> Useful for manually setting an item to be filtered under the "Tools" filter, if your item does not meet the automatic criteria for the "Tools" filter.
++		/// <br/> See (<see cref="GameContent.Creative.ItemFilters.Tools.FitsFilter(Item)"/>) to check if your item meets the automatic criteria.
++		/// </summary>
++		public static bool[] ToolsFilter = Factory.CreateBoolSet(509, 850, 851, 3612, 3625, 3611, 510, 849, 3620, 1071, 1543, 1072, 1544, 1100, 1545, 50, 3199, 3124, 5358, 5359, 5360, 5361, 5437, 1326, 5335, 3384, 4263, 4819, 4262, 946, 4707, 205, 206, 207, 1128, 3031, 4820, 5302, 5364, 4460, 4608, 4872, 3032, 5303, 5304, 1991, 4821, 3183, 779, 5134, 1299, 4711, 4049, 114);
```
BasicallyIAmFox: Probably would be a good idea to move this to `ItemID.TML.cs`

ghost: > Probably would be a good idea to move this to `ItemID.TML.cs`

Just curious, what is the difference between them?

JavidPack: It's a partial class. Moving things added by tModLoader into the .TML.cs class means that the changes are in a regular .cs file rather than a .patch file. The .patch file is harder to read and track changes in because it is basically a diff within a diff. When Terraria updates, the .patch files require effort to verify due to Terraria source code potentially changing the surrounding code, the .TML.cs files don't.

ghost: Ah ok, makes sense, I will make the necessary changes.

direwolf420:
I can see `ToolsFilter` -> `JourneyModeToolsFilter` because at a glance it's not clear what the set does, you would need to know or look at docs

JackerAcid:
> I can see `ToolsFilter` -> `JourneyModeToolsFilter` because at a glance it's not clear what the set does, you would need to know or look at docs

I considered doing this but thought the name was too long as they were XML summaries on this anyway. Although now thinking about it more, I might change the name to this.

PR #4121, opened 2024-03-18
ExampleMod updated with Asset guidelines

JavidPack:
In preparing [the Asset wiki page](https://github.com/tModLoader/tModLoader/wiki/Assets), it was discovered that ExampleMod didn't always show the recommended approach to dealing with `Asset`s. This PR updates ExampleMod to follow the guidelines in an attempt at showing good code design and efficient Asset usage.

This PR will be updated if needed based on feedback on the new approaches.

### Main.instance.LoadX
ExampleMod uses `Main.instance.LoadX` methods in various places where it is unnecessary. All modded content textures are loaded during mod loading. These have been removed

### Main.dedServ gates
No need to check `Main.dedServ` before calling `Request`. These have been removed.

### Remove code setting Assets to null
There is no need to set `Asset` fields to `null`. These have been removed
```diff
-public override void Unload() {
-	chainTexture = null;
-}
```

### Don't use ImmediateLoad unless needed
`, AssetRequestMode.ImmediateLoad` has been removed from `Request`s where it was not necessary.

### Store Extra Textures in Backing field
Backing fields added to reduce `Request` calls
```diff
+private Asset<Texture2D> rainTexture;
+public override void Load() {
+	rainTexture = Mod.Assets.Request<Texture2D>("Content/Biomes/ExampleRain");
+}
+public override Asset<Texture2D> GetRainTexture() => rainTexture;

-public override Asset<Texture2D> GetRainTexture() {
-	return ModContent.Request<Texture2D>("ExampleMod/Content/Biomes/ExampleRain");
-}
```

### Request assets during load instead of on demand 
To catch typos during mod loading, assets are loaded during loading in most cases.
```diff

+private static Asset<Texture2D> chainTexture;
+private static Asset<Texture2D> chainTextureExtra; // This texture and related code is optional and used for a unique effect

+public override void Load() {
+	chainTexture = ModContent.Request<Texture2D>(ChainTexturePath);
+	chainTextureExtra = ModContent.Request<Texture2D>(ChainTextureExtraPath);
+}

public override bool PreDraw(ref Color lightColor) {
-	Asset<Texture2D> chainTexture = ModContent.Request<Texture2D>(ChainTexturePath);
-	Asset<Texture2D> chainTextureExtra = ModContent.Request<Texture2D>(ChainTextureExtraPath); // This texture and related code is optional and used for a unique effect
	// etc
```


ExampleMod/Content/NPCs/ExampleBoneMerchant.cs
```diff
@@ -24,6 +25,15 @@ namespace ExampleMod.Content.NPCs
...
 		private static Profiles.StackedNPCProfile NPCProfile;
+		private static Asset<Texture2D> shimmerGun;
+
+		public override void Load() {
+			shimmerGun = ModContent.Request<Texture2D>(Texture + "_Shimmer_Gun");
+		}
+
+		public override void Unload() {
+			shimmerGun = null;
+		}
```
Chicken-Bones: Not necessary, I know we don't unload ALCs properly yet, but I don't want to mislead modders into thinking they have to clean up every static field they ever make

ExampleMod/Content/Projectiles/ExampleAdvancedAnimatedProjectile.cs
```diff
@@ -105,18 +106,18 @@ public class ExampleAdvancedAnimatedProjectile : ModProjectile
 				spriteEffects = SpriteEffects.FlipHorizontally;
 
 			// Getting texture of projectile
-			Texture2D texture = (Texture2D)ModContent.Request<Texture2D>(Texture);
+			Texture2D texture = TextureAssets.Projectile[Type].Value;
```
Chicken-Bones, line 109: Hm, it seems that modded entries in `TextureAssets.Projectile/Npc/Item` will always be loaded, where vanilla entries will be loaded on-demand via `Main.LoadProjectile/Item/NPC`.

I'm not sure what we should do about this, but it's a nuance that seems like a gotcha.

ExampleMod/Content/Projectiles/ExampleAdvancedFlailProjectile.cs
```diff
@@ -39,6 +42,16 @@ private enum AIState
 		public ref float CollisionCounter => ref Projectile.localAI[0];
 		public ref float SpinningStateTimer => ref Projectile.localAI[1];
 
+		public override void Load() {
+			chainTexture = ModContent.Request<Texture2D>(ChainTexturePath);
+			chainTextureExtra = ModContent.Request<Texture2D>(ChainTextureExtraPath);
+		}
```
Chicken-Bones, line 48: Sometimes we `Request` on `Load`, sometimes on demand. We should give some advice, or pick one and stick to it

ExampleMod/Content/Tiles/ExampleTorch.cs
```diff
@@ -69,9 +69,7 @@ public class ExampleTorch : ModTile
 			AddMapEntry(new Color(200, 200, 200), Language.GetText("ItemName.Torch"));
 
 			// Assets
-			if (!Main.dedServ) {
-				flameTexture = ModContent.Request<Texture2D>("ExampleMod/Content/Tiles/ExampleTorch_Flame");
-			}
+			flameTexture = Mod.Assets.Request<Texture2D>("Content/Tiles/ExampleTorch_Flame");
```
Chicken-Bones: `TextureAssets.ItemFlame` exists

ExampleMod/Content/Tiles/Plants/ExampleTree.cs
```diff
@@ -41,12 +45,12 @@ public class ExampleTree : ModTree
 
 		// Branch Textures
 		public override Asset<Texture2D> GetBranchTextures() {
-			return ModContent.Request<Texture2D>("ExampleMod/Content/Tiles/Plants/ExampleTree_Branches");
+			return branchesTexture ??= ModContent.Request<Texture2D>("ExampleMod/Content/Tiles/Plants/ExampleTree_Branches");
```
Chicken-Bones: Do we want to show using `Mod.Assets.Request` here? Or are we happy leaving that for 'advanced' modders and sticking to `ModContent` with full paths?

patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -103,6 +103,8 @@ public static bool FileExists(string name)
 	/// <summary>
 	/// Gets the asset with the specified name. Throws an Exception if the asset does not exist.
 	/// <para/>
+	/// Can be safely called on the server despite the server not loading assets, so there is no need to add checks for <see cref="Main.dedServ"/>.
+	/// <para/>
```
Chicken-Bones: Belongs in the inherited `IAssetRepository.Request` docs. Perhaps with a note that on the server assets will always be returned as if loaded via `DoNotLoad`.

Chicken-Bones (changes requested):

ExampleMod/Content/Tiles/Plants/ExampleTree.cs
JavidPack: The plant classes don't actually have a Mod field, it seems.

ExampleMod/Content/Tiles/ExampleTorch.cs
JavidPack: This is a tile, there is a TextureAssets.Flames used for tiles, but they have a unique index and just complicate things. `Main.tileFlame` docs say not to use them.

ExampleMod/Content/Projectiles/ExampleAdvancedFlailProjectile.cs
JavidPack: Most are on load now, I think that is best.

Chicken-Bones (approved):

PR #4126, opened 2024-03-20
Make important arrays in `Main` readonly

BasicallyIAmFox:
### What is the new feature?
Makes important arrays (such as `Main.npc`) `readonly`.

Safety and performance.

No.

```cs
// No longer compiles!
//Main.npc = new NPC[401];
```

### ExampleMod updates
No.


JavidPack:
This would break Modders Toolkit. It temporarily swaps out the npc and item arrays when doing some in-game loot calculations, then swaps them back. I assume I can find another way, but just a thought that this might break mods doing advanced things.

Chicken-Bones:
> This would break Modders Toolkit. It temporarily swaps out the npc and item arrays when doing some in-game loot calculations, then swaps them back. I assume I can find another way, but just a thought that this might break mods doing advanced things.

You can always swap the actual array contents. `Array.Copy` won't take very long.

Chicken-Bones (approved):

Chicken-Bones:
As per discussion in Discord, declined unless it can be proven to have significant performance benefit to real code.

PR #4128, opened 2024-03-22
Improve MessageID Documentation and other netcode methods

JackerAcid:
This pull request is to facilitate improvements to the documentation of `MessageID`.  All contributions and help are welcomed.

### The Problem
The current "documentation" of MessageID consists of simple comments that can't be picked up by editors when typing and glancing over them, they also state that they may have outdated information.
image
It is also not very concise or clear to modders if or what **exact** parameters they should be setting when using one of these MessageID's in the context of `NetMessage.SendData()`

This PR will aim to fix these issues (Targeting `MessageID.cs.patch` and `MessageID.TML.cs`).

### Checklist
- [ ] Convert all simple comments to XML summaries

- [ ] Provide updated info about what each of these MessageIDs mean when possible

- [ ] Provide a clear and easily understandable format for modders to follow if a MessageID needs extra info to be passed into `NetMessage.SendData()` to behave as intended


JackerAcid:
Also would like feedback if this would be an opportunity to move the MessageID's that are in the patch file to the TML file or not.

JavidPack:
I think all the tModLoader added messages are already in the MessageID.TML.cs file.

This is all good, at the very least making them all XML comments is already a big win. We'll want to decide on the final approach via discussion before doing all the work.

Documenting each of the parameters in the `MessageID` is a bit convoluted for the user. The user would have to hover over the `MessageID`, read the mapping of `numberX` to parameter, then apply that to their own `NetMessage.Send` parameters by figuring out which parameter is `numberX`. Rather than (or maybe in addition to) documenting each of the parameters there, I've been considering just making separate methods for each message and documenting the parameters there.

Document on the MessageID itself:
```	
/// <summary>
/// Sends all info about an NPC, position, velocity, and AI
/// NPCLoader.SendExtraAI appends data to this packet
/// Forwarded to other clients
/// <para/><B>NetMessage.SendData parameters:</B>
/// <br/>The <c>number</c> parameter corresponds to the <see cref="Entity.whoAmI"/> of the NPC to be synced.
/// </summary>
public const byte SyncNPC = 23;
```
> image

Stub method approach:
```
/// <summary>
/// Sends all info about an NPC, position, velocity, and AI
/// NPCLoader.SendExtraAI appends data to this packet
/// Forwarded to other clients
/// <br/><paramref name="npcIndex"/> is the <see cref="Entity.whoAmI"/> of the NPC to be synced.
/// <para/> Internally sends the <see cref="MessageID.SyncNPC"/> message.
/// </summary>
public static void SendSyncNPCMessage(int npcIndex, int remoteClient = -1, int ignoreClient = -1) => SendData(MessageID.SyncNPC, remoteClient: remoteClient, ignoreClient: ignoreClient, number: npcIndex);
```

> image

Just an option. This would allow modders to get rid of all the parameters that are unused, and the required arguments will hint to the modder which parameters are needed so they don't accidentally forget any. It would require a lot of work to make sure they are correct. Some of them would probably require separate methods for client and server.

BasicallyIAmFox:
> Stub method approach:
> 
> ```
> /// <summary>
> /// Sends all info about an NPC, position, velocity, and AI
> /// NPCLoader.SendExtraAI appends data to this packet
> /// Forwarded to other clients
> /// <br/><paramref name="npcIndex"/> is the <see cref="Entity.whoAmI"/> of the NPC to be synced.
> /// <para/> Internally sends the <see cref="MessageID.SyncNPC"/> message.
> /// </summary>
> public static void SendSyncNPCMessage(int npcIndex, int remoteClient = -1, int ignoreClient = -1) => SendData(MessageID.SyncNPC, remoteClient: remoteClient, ignoreClient: ignoreClient, number: npcIndex);
> ```
> 
> > image
> 
> Just an option. This would allow modders to get rid of all the parameters that are unused, and the required arguments will hint to the modder which parameters are needed so they don't accidentally forget any. It would require a lot of work to make sure they are correct. Some of them would probably require separate methods for client and server.

I was actually thinking to do similar approach in my Net Packets pr except structs instead of methods, so e.g.:
```cs
new SyncNPCPacket(slot).SendToAllPlayers();
```
(docs on the struct)

JavidPack:
Since this PR was in limbo and there never was community consensus on if we wanted to make the stub methods, I went ahead and just converted all the MessageIDs that are used in ExampleMod and some others to real documentation. I figure messages used in ExampleMod work and are compatible with modded content. Other messages still have the normal comments but since I don't expect modders to use them I didn't do the work yet to convert them.

I also changed ExampleMod to be explicit about the NetMessage.SendData parameters, I think that's a good idea moving forward.

```diff
-	NetMessage.SendData(MessageID.TileManipulation, -1, -1, null, 0, i, j);
+	NetMessage.SendData(MessageID.TileManipulation, number: 0, number2: i, number3: j);
```

If anyone wants anything else commented, just let me know here or on Discord and I'll do it in a separate commit.

PR #4133, opened 2024-03-24
fixed AltGr not working in chat (#3910)

### LINKED ISSUE #3910: Non-english keyboards with Alt-Gr key not supported
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

https://en.wikipedia.org/wiki/AltGr_key

This key is not recognized in the ingame chat and other places where you can type text. This is a big problem when i.e. on german, french, or czech where typing symbols such as `\`, `[]` and `{}` requires pressing that key. Especially `[]` is useful for any form of chat tags https://terraria.wiki.gg/wiki/Chat#Tags.

To clarify: The key works in **vanilla Terraria**, so it's a deficiency, likely caused by the use of FNA.
I however cannot reproduce it anymore, it seems to be bugged in vanilla too: https://forums.terraria.org/index.php?threads/characters-needing-alt-gr-not-entering-in-chat.116332/

### Log File

There's nothing in the logs when using the key.

### Steps to reproduce

Get a keyboard and a layout that has `[]` only accessible by Alt Gr, load the game and try to type them in chat.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

NotLe0n:
fixes #3910

### What is the bug?
It was not possible to write some special characters like `[]{}\` on keyboard layouts that require pressing CTRL+Alt (or AltGr) to write them.

### How did you fix the bug?
The game checks the CTLR key for key combination like CTRL+C, CTRL+V, CTRL+Z, ect. and doesn't write the character. However when trying to write some characters using CTRL+Alt, it still went into the logic for the key combinations.

The fix is to check if the Alt key is being pressed. If it is, then it's not a key combination and the game should write the character.

No, I don't think so

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2443,12 +2443,13 @@
  		wallFrameCounter[144]++;
  		int num3 = 5;
  		int num4 = 10;
-@@ -15840,7 +_,7 @@
+@@ -15840,7 +_,8 @@
  			text = "";
  
  		bool flag = false;
 -		if (inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.LeftControl) || inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.RightControl)) {
-+		if (inputText.IsKeyDown(Keys.LeftAlt) || inputText.IsKeyDown(Keys.RightAlt) && !(inputText.IsKeyDown(Keys.LeftControl) || inputText.IsKeyDown(Keys.RightControl))) {
++		// Old code prevented users from using AltGr modifier key. Needed to type chars such as []{}\| on many languages keyboard layouts. Windows interprets Ctrl + Alt as AltGr.
++		if ((inputText.IsKeyDown(Keys.LeftControl) || inputText.IsKeyDown(Keys.RightControl)) && !(inputText.IsKeyDown(Keys.LeftAlt) || inputText.IsKeyDown(Keys.RightAlt))) {
  			if (inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.Z) && !oldInputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.Z)) {
  				text = "";
  			}
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2443,6 +2443,15 @@
  		wallFrameCounter[144]++;
  		int num3 = 5;
  		int num4 = 10;
+@@ -15840,7 +_,7 @@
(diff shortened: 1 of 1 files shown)


NotLe0n:
@Chicken-Bones It might be worth it to fix it in Terraria too

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -2443,6 +2443,15 @@
  		wallFrameCounter[144]++;
  		int num3 = 5;
  		int num4 = 10;
+@@ -15840,7 +_,7 @@
+ 			text = "";
+ 
+ 		bool flag = false;
+-		if (inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.LeftControl) || inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.RightControl)) {
++		if (Main.inputText.IsKeyDown(Keys.LeftAlt) || Main.inputText.IsKeyDown(Keys.RightAlt) && !(Main.inputText.IsKeyDown(Keys.LeftControl) || Main.inputText.IsKeyDown(Keys.RightControl))) {
```
direwolf420:
```suggestion
-		if (inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.LeftControl) || inputText.IsKeyDown(Microsoft.Xna.Framework.Input.Keys.RightControl)) {
+		if (inputText.IsKeyDown(Keys.LeftAlt) || inputText.IsKeyDown(Keys.RightAlt) && !(inputText.IsKeyDown(Keys.LeftControl) || inputText.IsKeyDown(Keys.RightControl))) {
```

JavidPack:
`
if (inputText.IsKeyDown(Keys.LeftAlt) || inputText.IsKeyDown(Keys.RightAlt) && !(inputText.IsKeyDown(Keys.LeftControl) || inputText.IsKeyDown(Keys.RightControl))) {
`

This logic wasn't right, it broke the ctrl-c/x/v feature unless the user was also pressing RightAlt. I've updated the PR with the correct logic.

I also confirmed the issue on Terraria on Windows.

I've tested the fix with French keyboard on Windows, but I'm not positive how this behaves on Mac or Linux. I assume it works since I think only windows interprets AltGr as Ctrl+Alt, but I do think it would be interesting to confirm if Terraria and tModLoader users on Linux or Mac can actually type these special characters, or if XNA/FNA just doesn't know about AltGr at all and doesn't capture the correct text. There is no Keys.AlrGr enum.

Chicken-Bones (approved):

JavidPack:
Did a little more investigation to see if this logic needed to be applied anywhere else alt/ctrl is used in the game, seems sufficient.

PR #4037, opened 2024-01-21
Port TML from .NET 6 to .NET 8

Mirsario:
.NET 7 and .NET 8 have brought plenty of performance-related improvements and upgrades, as well as some upgrades to C#. However, we've been choosing to stay on .NET 6 for quite a while now, since .NET 7 also happened to drop support of Windows 7, 8.0 and 8.1. Now, a workaround for that has been found, and this PR is here in hopes to bring all the aforementioned goodness without any downsides.

## Checklist
- [x] Port TerrariaNetCore.
- [x] Port tModLoader.
- [x] Set `DOTNET_EnableWriteXorExecute=0` envvar in start scripts if on win7/win8/win8.1.
- [x] Get `tests/TilemapPerformance` working again.
    - [x] Create some benchmarks comparing .NET 6 against .NET 8 in WorldGen.
- [x] Update CI to use .NET 8.
- [x] User feedback on performance is positive.
- [x] Almost all mods are confirmed to be working.
    - Infernum broke due to meddling with Windows-only libraries that got removed. Authors were notified. 
- [x] Update the FNA fork to compile for .NET 8, so that building mods doesn't log warnings.
- [x] Rebuild TerrariaHooks.
- [x] @Solxanich Mitigate problems of .NET updates changing executable paths, causing user-declared process-tied settings such as those of Nvidia Control Panel to reset.
    - ~~Use scripts to maintain a `dotnet/current/` folder through symlinks/hardlinks/junctions?~~
    - Just install directly into `dotnet/`?
- [x] ~~Replace or update `DotNetZip.Reduced`/`Ionic.Zip.Reduced`. It's throwing exceptions.~~ Done by `#4109`.
- [x] Update the pre-packaged runtime on windows (`dotnet-runtime-6.0.14-win-x64.zip`).
- [x] Resolve builds failing at large if-else chains with SDK >=8.0.200. They work in 8.0.100 and in Visual Studio (which ships custom SDK claiming to be 8.0.20X).
    - We've split the `Projectile.SetDefaults`  method on the `TerrariaNetCore` patchstep to avoid this.

PR #4099, opened 2024-03-08
tModCodeAssist rewrite

BasicallyIAmFox:
## DROPPED
See new and updated PR: https://github.com/tModLoader/tModLoader/pull/4727

Reasons being that this PR is simply not up-to-date and last commit were over year ago, so it simply became not worth it and easier to start from scratch. Additionally, scope became too big for one PR.

##

### What is the new feature?
Reworked analyzers for tML.
- Renamed tModLoader.CodeAssist to tModCodeAssist to fit with the tML branding
- Moved tModCodeAssist to the tML repository and added it as an for mods analyzer in `tModLoader.targets`
- Ported the old analyzer rules

### Why should this be part of tModLoader?
Having tModCodeAssist in the tModLoader repository is much better than having it separately and publishing it as a nuget package because we don't have to consider supporting 1.3, 1.4.3, and 1.4.4 at the same time in one codebase. New analyzer rules will also help new modders port vanilla code and can help them simplify their own code.

### Are there alternative designs?
- Leaving the name as tModLoader.CodeAssist (bad)
- Leaving the analyzer in a separate repository and making tML a submodule of it (bad)
- Leaving the analyzer as a nuget package (bad)

### Sample usage for the new feature
image

### ExampleMod updates
N/A

### TODO
* [x] Magic Number rules
* [X] Port `rand` simplify rule
* [ ] Fully support the rest of ID classes (`ItemID`, `NPCID`, `ProjectileID`, etch)

PR #4134, opened 2024-03-24
Refactored mod creation to use a file tree of templates.

Mirsario:
To assist two+ active PRs (#2472 and #4037), this one does the following:
- [x] Moves code related to creating & upgrading mod source folders out of UI and into a new separate `Terraria.ModLoader.Core.SourceManagement` class.
- [x] Replaces textfile-generating hardcode with a filetree of embedded resources that are enumerated and exported to populate mod src folders. The embedded resources use the [Mustache](https://mustache.github.io/mustache.5.html) logicless template format in their names and contents (this concept will be nice to use for mod descriptions too in userland in the future). Optional files, such as the BasicSword, are skipped by making their filenames evaluate to nothing. 
- [x] Removes `TargetFramework` and `PlatformTarget` from default `.csproj` files, putting it into ~~`tModLoader.targets`~~ `tMLMod.targets` instead.
- [x] Makes Csproj upgrades attempt to function via XML edits, if they're not too broken.
- [ ] When this hits Stable, update https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Modding-Guide#mod-skeleton-contents info and screenshots with new `Content.Items` namespace and folder.

### Default mod template changes:
- `launchSettings.json`, uses tabs now.
- Csproj uses tabs now, no longer has an obsolete <?xml> header, is now practically empty besides for the targets import, with every other line just being nice "put this here" markups.
- The item template now uses the `ExampleSword.png`, which is nicer to look at. A tooltip has also been added, to immediately showcase how to make multiline strings in HJSON.
- The two `.cs` files now use file-scoped namespaces. The item template's code had its comments updated, and had magic numbers replaced with ID classes in a few lines.
- `description.txt`:
```diff
-{{ModDisplayName}} is a pretty cool mod, it does...this. Modify this file with a description of your mod.
+Modify this file with a description of your mod.
```
- As too many people don't bother even providing a mod icon, I felt like updating the default mod icon to be slightly more presentable, while communicating its placeholder nature:
 -> 

### Questionable notes:
- I've had to exclude the new template folder from `AssemblyResourcesContentSource`, because of it rather annoyingly crashing the game whenever it encountered a `.cs` + `.png` pair.
- The templates' EmbeddedResource LogicalNames are set to use forward slashes, due to localization files utilizing dots, which slashes get reasonlessly converted to by default. Ideally even the vanilla game would convert to using forward slashes, it's beneficial for its asset system.


Destructor-Ben:
In a future PR could we allow specifying our own template? atm I copy paste a template and hope I remembered to change all file names + contents of files

JavidPack:
Should we move LangVersion to the tMLMod.targets too? I believe a modder should still be able to override it in their mod's csproj, but an issue we have right now is that modders are seeing suggestions in VS for language versions that the ingame complier does not support. I know there is a pr for doing it through dotnet itself, but even then we'd want to make sure that only the highest supported lang version is the default.

Mirsario:
> Should we move LangVersion to the tMLMod.targets too?

I've done that now, with it explicitly set to `10.0` for now. But only overrides with old versions in values will be auto-removed in the upgrade process. `latest`, `preview`, and >=`11.0` will be kept, so I guess it's not really helping what you've described. Oh well, hopefully we merge 2472 soon.

JavidPack:
Will this be able to easily remove `<PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />` once #4099 is merged? I haven't parsed all the code yet, just curious.

Mirsario:
Yes, here's the code to insert even:
```cs
// Remove the analyzer package, since our targets file now handles all that.
RemoveNodes(itemGroups.Elements("PackageReference").Where(
	e => e.Attribute("Include")?.Value == "tModLoader.CodeAssist"
));
```

**EDIT:** This is now a part of the PR, with the PackageReference moved into `tMLMod.targets`.

patches/tModLoader/Terraria/ModLoader/Assets/AssemblyResourcesContentSource.cs
```diff
@@ -3,21 +3,32 @@
...
+
+	public AssemblyResourcesContentSource(Assembly assembly, string rootPath = null, IEnumerable<string> excludedStartingPaths = null)
 	{
 		this.assembly = assembly;
+		excludedStartingPaths ??= Enumerable.Empty<string>();
 
 		IEnumerable<string> resourceNames = assembly.GetManifestResourceNames();
+
+		foreach (string startingPath in excludedStartingPaths ?? Enumerable.Empty<string>()) {
+			resourceNames = resourceNames.Where(p => !p.StartsWith(startingPath));
```
Chicken-Bones, line 29: Is this necessary? You're not using the asset system for your templates anymore

patches/tModLoader/Terraria/ModLoader/Templates/Properties/launchSettings.json
```diff
@@ -0,0 +1,31 @@
...
+			"workingDirectory": "$(TerrariaSteamPath)",
+			"nativeDebugging": true
+		},
+		"TerrariaServer - Native Debug": {
+			"commandName": "Executable",
+			"executablePath": "dotnet",
+			"commandLineArgs": "$(tMLServerPath)",
+			"workingDirectory": "$(TerrariaSteamPath)",
+			"nativeDebugging": true
+		}
```
Chicken-Bones: Why would the average user want these two configurations?

patches/tModLoader/Terraria/ModLoader/Templates/Properties/launchSettings.json
```diff
@@ -0,0 +1,31 @@
+{
+	"profiles": {
+		"Terraria": {
+			"commandName": "Executable",
+			"executablePath": "dotnet",
```
Chicken-Bones: Don't these need to use `$(DotNetName)`?

patches/tModLoader/Terraria/ModLoader/Templates/{{ModName}}.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<!-- Import tModLoader mod properties -->
+	<Import Project="..\tModLoader.targets" />
+
+	<!-- General -->
+	<PropertyGroup>
+		<AssemblyName>{{ModName}}</AssemblyName>
+		<LangVersion>latest</LangVersion>
```
Chicken-Bones: Remove?

patches/tModLoader/Terraria/ModLoader/Templates/{{ModName}}.csproj
```diff
@@ -0,0 +1,17 @@
...
+	</PropertyGroup>
+
+	<!-- References -->
+	<ItemGroup>
+		<PackageReference Include="tModLoader.CodeAssist" Version="0.1.*" />
```
Chicken-Bones: Does this not belong in `tMLMod.targets` too?

patches/tModLoader/Terraria/Terraria.csproj.patch
```diff
@@ -78,7 +79,18 @@
...
-@@ -60,10 +_,11 @@
+ 		<EmbeddedResource Include="IO/Data/*" />
+ 		<EmbeddedResource Include="Localization/Content/**" />
+ 		<EmbeddedResource Include="Microsoft/**" />
++		
++		<!-- Mod Template -->
++		<None Include="ModLoader/Templates/**" />
++		<Compile Remove="ModLoader/Templates/**" />
++		<EmbeddedResource Remove="ModLoader/Templates/**" />
++		<EmbeddedResource Include="ModLoader/Templates/**">
```
Chicken-Bones, line 86: You can use `Update=` instead of `Remove` and then `Include`

patches/tModLoader/Terraria/ModLoader/Templates/{{ModName}}.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Terraria.ModLoader;
+
+namespace {{ModName}};
```
Chicken-Bones: Needs approval from @JavidPack are we doing file scoped namespaces now? We don't in ExampleMod

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+	private static readonly HashSet<string> textExtensions = new() {
+		".txt", ".json", ".hjson", ".toml", ".cs", ".csproj", ".sln"
+	};
+
+
+	/// <summary> Writes mod template files to the provided source-code directory. </summary>
+	public static void WriteModTemplate(string modSrcDirectory, in TemplateParameters templateParameters)
+	{
+		var modLoaderAssembly = typeof(ModLoader).Assembly;
+		object boxedParameters = templateParameters;
```
Chicken-Bones: why use record struct and boxing, rather than regular `record`

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+			try {
+				// Old files can cause some issues.
+				DeleteIfExists(new DirectoryInfo(Path.Combine(modSrcDirectory, "obj")));
+				DeleteIfExists(new DirectoryInfo(Path.Combine(modSrcDirectory, "bin")));
+
+				//TODO: Why do we do this?
+				DeleteIfExists(new FileInfo(Path.Combine(modSrcDirectory, "Properties", "AssemblyInfo.cs")));
+			}
+			catch { }
+		}
```
Chicken-Bones: This needs to be a modification itself. 
I also prefer `modifications.Any()`

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+		result = default;
+		return false;
+	}
+
+	/// <summary> If succesfful - populates the provided list with delegates, executing which will upgrade the provided document. </summary>
+	private static bool TryCollectingCsprojModifications(XDocument document, List<Action> modifications)
+	{
+		// Check if this is even a C# project.
+		if (document!.Root is not { Name.LocalName: "Project", FirstAttribute: { Name.LocalName: "Sdk", Value: "Microsoft.NET.Sdk" } } root) {
+			return false;
```
Chicken-Bones: this is currently the only way to get `FileBroken`. See my above comment about factoring this out, so this method doesn't need a `bool` return, and can just return (or `yield`) the list of modifications instead

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+		// Try to collect required modifications.
+		var modifications = new List<Action>();
+		if (document != null && !TryCollectingCsprojModifications(document, modifications)) {
+			status = UpgradeStatus.FileBroken;
+		}
+
+		// If there are any modifications - the file is upgradeable.
+		if (modifications?.Count > 0) {
+			status = UpgradeStatus.Upgradeable;
+		}
```
Chicken-Bones: I feel like the `status` enum here is not particularly needed or readable. 

```cs
if (File.Exists(csprojPath)
      && TryLoadXmlDocument(csprojPath, LoadOptions.PreserveWhitespace, out var document)
      && IsModernCsproj(document)) {
    
    var modifications = CollectCsprojUpgradeActions(document);
    if (modifications.Count == 0)
        return false;

    action = () => {
        modifications!.ForEach(m => m());
        WriteXmlDocumentToFile(csprojPath, document!);
    }
    return true;
}

action = () => ResetCsprojFile(csprojPath, templateParameters);
return true;
```

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+
+		// If this is not a dry run and there's changes to be made.
+		if (status is not UpgradeStatus.NotRequired) {
+			result = () => {
+				// Apply modifications if any exist.
+				modifications!.ForEach(m => m());
+
+				// Recreate the file from scratch if it's missing or broken, otherwise just save our XML.
+				if (status is UpgradeStatus.FileMissing or UpgradeStatus.FileBroken) {
+					ResetCsprojFile(csprojPath, createBackup: status is not UpgradeStatus.FileMissing, templateParameters);
```
Chicken-Bones: No need for the `createBackup` parameter, just make a backup inside the function if the file exists

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
```diff
@@ -0,0 +1,320 @@
...
+				foreach (var element in nodesToRemove) {
+					// Remove whitespace, which is otherwise kept due to the way we parsed the document.
+					if (element.PreviousNode is XText previous && string.IsNullOrWhiteSpace(previous.Value)) {
+						previous.Remove();
+					}
+
+					element.Remove();
+				}
+			});
+		}
```
Chicken-Bones: Normally I'd factor this as a local function at the top of the method, like:
```cs
void RemoveNodes(IEnumerable<XNode> nodes) {
    modifications.Add(() => ...)
}
```

Chicken-Bones (changes requested):
Overall this is an excellent change. I've provided some factoring feedback.

This PR should _not_ change the templates. It makes it really hard to review. Please split template changes into an additional 1 or 2 PRs (2 if there are changes that have to be done both before and after this PR).

patches/tModLoader/Terraria/ModLoader/Assets/AssemblyResourcesContentSource.cs
Mirsario: Duplicate extension errors are thrown at ContentSource initialization time, not access time. Any duplicate extensions found in EmbeddedResources will crash the game. This behavior seriously needs redesigns, as we still have crashes from some resourcepacks intended for vanilla.

patches/tModLoader/Terraria/ModLoader/Templates/Properties/launchSettings.json
Mirsario: I guess it'll cause more confusion than it's worth, will remove.

patches/tModLoader/Terraria/ModLoader/Core/SourceManagement.cs
Mirsario: A CA rule cries about .Any() being inferior. csharpbad.

PR #4135, opened 2024-03-25
Publicize Eye Helper

### LINKED ISSUE #2350: [1.4] Player blinking hook(s)

No

Mod capability as a Modder

### Description

New hook could help modders to add more eye-related visuals! (ex. half-closed eyes when has some buff/debuff, etch)

It seems to be quite hard right now to choose/do eye frames without some custom drawing and that all, for just something simple like eye blinking when Player has some buff/debuff, wearing some "heavy" armor, etch.

_No response_

BasicallyIAmFox:
Closes #2350 

### What is the new feature?
Publicizes `PlayerEyeHelper` to give modders control over player eyes.

There's no (good) way of interacting with player eyes (excluding reflection which is slow and will box eye helper struct)

.

```cs
Player.eyeHelper.CurrentEyeFrame = PlayerEyeHelper.EyeFrame.EyeOpen;
```

### ExampleMod updates
Added `ExampleBlinkingPlayer` to showcase new possibilities.

### Porting Notes
`Player.eyeHelper.EyeFrameToShow` => `Player.eyeHelper.CurrentEyeFrame` (return type was also changed from `int` to `PlayerEyeHelper.EyeFrame`)

### DIFF
--- a/ExampleMod/Common/Players/ExampleBlinkingPlayer.cs
+++ b/ExampleMod/Common/Players/ExampleBlinkingPlayer.cs
@@ -1,0 +1,31 @@
+﻿using Terraria.GameContent;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	// Showcases modifying eye state and frames under certain condition.
+	public class ExampleBlinkingPlayer : ModPlayer
+	{
+		public override void PostUpdate() {
+			// Set eye state to our custom eye state if player is in water and is in state of normal blinking.
+			if (Player.eyeHelper.CurrentEyeState == PlayerEyeHelper.EyeState.NormalBlinking && Player.wet) {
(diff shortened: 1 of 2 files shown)


Mirsario (approved):

JavidPack:
Looking it over, it all seems right. Can you add porting notes to the description if someone was using Player.eyeHelper.EyeFrameToShow already?

JavidPack:
Looks good, didn't see that EyeFrameToShow had a private setter. Anyone doing this currently would already be using reflection. 

The PR can't really implement new states as is without monomod edits, but we can tackle that if modders bring it up. An implementation of new states would require a way to indicate priority within the SetStateByPlayerInfo logic chain, registering new state ids, and a hook within UpdateEyeFrameToShow.

PR #4137, opened 2024-03-27
Made `EffectParameterCollection` to throw if parameter with specified name does not exist

BasicallyIAmFox:
More helpful/verbose comparing to `NullReferenceException`.

Add null checks in TML.

### Sample usage
If shader is missing a paremeter then it will properly throw instead of `NullReferenceException`


Chicken-Bones:
You can't just push a submodule update. You have to PR to the submodule first.

BasicallyIAmFox:
> You can't just push a submodule update. You have to PR to the submodule first.

Done 😄
see https://github.com/tModLoader/FNA/pull/2

PR #4151, opened 2024-04-04
Make HitInfo and HurtInfo read/init only

Chicken-Bones:
### What is the new feature?

`NPC.HitInfo` and `Player.HurtInfo` are now `readonly` and have `get; init` properties.

> I see people trying to do 
> `hit.Crit = false`, which compiles but doesn't do anything

Make the hook parameters `in` instead. This would be just as breaking, and not carry through to user's own non-hook methods. The dotnet guidelines also recommend making a struct `readonly` if it's going to be used with `in` where possible (though this is a performance detail that isn't too relevant to us).

image

### Porting notes

The `Hit/HurtModifiers.ModifyHit/HurtInfo` backdoors are now a little bit harder to use. Modders will have to create a new `Hit/HurtInfo` (preferably via the `with` syntax) instead of setting fields directly. Eg:

```cs
public override void ModifyHitNPC(NPC target, ref NPC.HitModifiers modifiers) {
	modifiers.ModifyHitInfo += (ref info) => {
		// no longer works
		info.Damage *= 2;

		// instead make a new HitInfo
		info = info with { Damage = info.Damage * 2 };
	};
}
```

### ExampleMod updates
None required

### DIFF
--- a/ExampleMod/Content/Projectiles/ExampleSwingingEnergySwordProjectile.cs
+++ b/ExampleMod/Content/Projectiles/ExampleSwingingEnergySwordProjectile.cs
@@ -160,6 +160,11 @@ namespace ExampleMod.Content.Projectiles
 			Utils.PlotTileLine(Projectile.Center + starting, Projectile.Center + ending, width, DelegateMethods.CutTiles);
 		}
 
+		public override void ModifyHitNPC(NPC target, ref NPC.HitModifiers modifiers) {
+			// Set the target's hit direction to away from the player so the knockback is in the correct direction.
+			modifiers.HitDirectionOverride = (Main.player[Projectile.owner].Center.X < target.Center.X) ? 1 : (-1);
+		}
+
 		public override void OnHitNPC(NPC target, NPC.HitInfo hit, int damageDone) {
 			// Vanilla has several particles that can easily be used anywhere.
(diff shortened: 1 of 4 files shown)


Tyfyter:
This would likely necessitate changes to almost every use of ModifyHitInfo and ModifyHurtInfo, could it be done as well with an analyzer?

Tyfyter:
Actually, wouldn't marking the parameter in `OnHitX` as `in` make more sense?
It's a code breaking change either way, but this way would surely be simpler for tModPorter and definitely show its intent more clearly.

Chicken-Bones:
> This would likely necessitate changes to almost every use of `ModifyHitInfo` and `ModifyHurtInfo`, could it be done as well with an analyzer?

Yep, but those are backdoors which we heavily discourage using. `with` syntax still works fine (`info = info with { Crit = true }`) for eg

> Actually, wouldn't marking the parameter in `OnHitX` as `in` make more sense? It's a code breaking change either way, but this way would surely be simpler for tModPorter and definitely show its intent more clearly.

I discussed this in the "Are there alternative designs?" section, there's not really too much strong argument either way, but this is definitely the 'correct' way to do it.

Tyfyter:
Maybe it's been too long since I was a new modder, but I don't think we can prevent someone from being confused if they see something explicitly marked as read only and conclude that they can write to it if they just pass it to their own method.

PR #4155, opened 2024-04-09
Localized MagicSummonHybrid Tooltip and Fixed Missing Space at the Beginning

Rijam:
### What is the bug?

The tooltip for `DamageClass.MagicSummonHybrid` was not localized and was missing a space at the beginning of the text.

image

### How did you fix the bug?

Added a key in the tModLoader localization file and added a space at the front.

image

The name of the localization key could be different. The actual tooltip text could be reworded, too.

I don't know if the other languages need a commented out key added to their file.


Rijam:
In case anyone was wondering why this damage class exists, it's solely used for the Forbidden armor's Ancient Storm set bonus.

PR #4163, opened 2024-04-13
Updated OneDrive opening logic in Program.TML.cs to open in background w/ no pop-ups

jackashmore:
### What is the new feature?
Changed logic in `Program.TML.cs` to open OneDrive in the background with no pop-ups instead of opening a new File Explorer window that must be manually closed. Passes `/background` as a cmd-line argument using `ProcessStartInfo` from `System.Diagnostics`.

A small QOL change. Based on [this Reddit post](https://www.reddit.com/r/Terraria/comments/15lev8t/tmodloader_opens_onedrive_in_file_explorer_every/?rdt=61396) from 8 months ago, there are still recent replies (as recent as 1 day ago) requesting a fix.

NA.

Upon running tModLoader, OneDrive now opens and appears in the system tray. Tested locally and on a couple of different Windows systems with no issues.

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -142,14 +142,25 @@ public static partial class Program
 			try {
 				var oneDrivePath1 = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Microsoft\\OneDrive\\OneDrive.exe");
 				var oneDrivePath2 = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "Microsoft OneDrive\\OneDrive.exe");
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Program.TML.cs
```diff
@@ -142,14 +142,25 @@ private static void PortFilesFromXtoY(string superSavePath, string source, strin
...
+				}
+				else if (File.Exists(oneDrivePath2)) {
+					oneDriveInfo.FileName = oneDrivePath2;
+					System.Diagnostics.Process.Start(oneDriveInfo);
+				}
 				Thread.Sleep(3000);
 			}
 			catch { }
+
 		}
```
Mirsario:
```suggestion
			catch { }
		}
```

PR #4164, opened 2024-04-14
Fix Logitech RGB and G Hub detection

### LINKED ISSUE #1785: [1.4 MergedTesting] RGB Lighting Integration doesn't work

Anshroid:
### What is the bug?
Logitech RGB would not function on tModLoader, but would in Vanilla. In addition, Logitech G Hub would not detect the game, even when the `dotnet.exe` executable was added in settings (as described in #1785)

### How did you fix the bug?
Added the Logitech LED Illumination SDK wrapper DLL and updated the interop file to use the DLL from the libraries folder. This fix was largely based on #3096. This should also, at least partially, fix #1785.

This fix uses the x64 DLL, whereas the x86 DLL may be preferred. I am not sure.

### Other
DLL from [https://www.logitechg.com/en-gb/innovation/developer-lab.html](https://www.logitechg.com/en-gb/innovation/developer-lab.html) (LED Illumination SDK), at `LED/Lib/LogitechLedEnginesWrapper/x64/`.
I believe `TerrariaNetCore` to be the right place to put this fix, but I am not very familiar with contributing to tModLoader, so my apologies if this is incorrect.


Chicken-Bones:
Why did the original `DLL_NAME` have a space at the end? Are you sure the `.dll` suffix is needed?

Anshroid:
From what I tested, it wouldn't work using the name with the space unless the DLL was placed directly in the same directory as the executable (which is where it is in vanilla). Trying to load the DLL from `libraries` using the spaced name failed.

Anshroid:
The spaced name seems to come from the C# example that Logitech themselves provide with the Illumination SDK.

PR #4167, opened 2024-04-16
Cleanup Old folder, docs.

MrQuba:
Ports *Exclusive Accessory* mentioned in #1403. 
I do not think so


direwolf420:
Most of the logic is obsolete with the 1.4 hook `CanAccessoryBeEquippedWith`. I'm not even sure if the example needed porting when all it showcases is 1 hook.

JavidPack:
The approach used in this does not work correctly with modded accessory slots. I've gone ahead and deleted the changes and added more docs for modders to see instead. 

Also, you had the namespace as `namespace Content.Items.Accessories`, which doesn't even load, did you not test ExampleMod when making this PR? Please make sure to test PRs.

MrQuba:
I'm quite new to this, so thanks for pointing it out, I'll keep that in mind

PR #4171, opened 2024-04-18
Make ExampleMusicBoxTile emit particles

JamzOJamz:
Makes the [ExampleMusicBoxTile](https://github.com/tModLoader/tModLoader/blob/1.4.4/ExampleMod/Content/Tiles/ExampleMusicBoxTile.cs) occasionally emit music note particles when in an open state. The logic is adapted directly from vanilla Terraria.

No, however a comment within the code detailing what purpose it serves could be beneficial.


JavidPack:
Thanks a lot. I added the TileFrameY check in and changed the NextBool(5) back to NextBool(3) to match vanilla. And also the echo coating check.

PR #4174, opened 2024-04-21
correct a spelling error in EnvironmentFix.sh which causes tModLoader to crash when launch with Vulkan on Mac

lhzlhz419:
### What is the bug?

tModLoader crashes when launch with Vulkan on Mac due to a spelling error in EnvironmentFix.sh.

### How did you fix the bug?

Correct the spelling error.

No.


JavidPack:
Good catch. Can I assume you are on a mac and have tested this?

lhzlhz419:
> Good catch. Can I assume you are on a mac and have tested this?

I have tested it on my MacBook Pro 14 m2 pro.
截屏2024-04-21 10 33 38

截屏2024-04-21 10 35 02

PR #4175, opened 2024-04-23
Multiple prefix category support

Tyfyter:
### What is the new feature?
Added support for an item to roll and retain prefixes from multiple categories
Also fixes "any weapon" prefixes potentially being added to the pool twice and accessories which match a weapon category's criteria being unable to roll or retain accessory prefixes

This functionality is desired by some modders and currently requires detouring or IL editing, which is likely to cause incompatibility between mods

None of which I am aware.

### ExampleMod updates
Added "Example Multiple Prefix Category Weapon", which can get both melee and magic prefixes
<sup>The sprite is just blue Example Sword</sup>

# Multiple Prefix Support Test Mod

[MultiplePrefixSupportTestMod.zip](https://github.com/user-attachments/files/16739136/MultiplePrefixSupportTestMod.zip)

Attached is a mod modders can use to test these changes with their mod. To use, load up only this mod and your mod in both stable (v2024.07) and preview (v2024.08). In the `tModLoader` and `tModLoader-preview` save folders there will be a file named `MultiplePrefixSupportTestModResults_{version}.txt`. Open up both of these files in a capable text editor that allow comparing files. I suggest Notepad++'s Compare plugin but there are many options. Compare the files and navigate the "diff", making note of any prefixes that are unexpected for the items that differ. Use this information to apply the porting notes to fix your items to work as desired.

# Porting Notes
- Some existing items in your mod may now count towards multiple prefix categories, especially those using a custom `DamageClass`. Test these items and determine if their prefixes are desired. Adjust `DamageClass.GetPrefixInheritance` or `ModItem.MeleePrefix/WeaponPrefix/RangedPrefix/MagicPrefix` as appropriate.
- The `MultiplePrefixSupportTestMod.tmod` mod contained in the PR description can be used to easily compare prefix results, please use it to verify that the item prefixes are as expected.
- If using `Item.GetPrefixCategory`, use `Item.GetPrefixCategories` instead and adjust your logic


patches/tModLoader/Terraria/ModLoader/PrefixLoader.cs
```diff
@@ -5,6 +5,7 @@
 using Terraria.ID;
 using Terraria.ModLoader.Core;
 using Terraria.Utilities;
+using static Terraria.GameContent.UI.EmoteID;
```
Chicken-Bones: Accidental?

patches/tModLoader/Terraria/Item.TML.cs
```diff
@@ -295,4 +297,29 @@ private void RestoreMeleeSpeedBehaviorOnVanillaItems()
...
+			categories.Add(PrefixCategory.Magic);
+
+		if (PrefixLegacy.ItemSets.SpearsMacesChainsawsDrillsPunchCannon[type] || PrefixLegacy.ItemSets.BoomerangsChakrams[type] || PrefixLegacy.ItemSets.ItemsThatCanHaveLegendary2[type] || ItemLoader.WeaponPrefix(this) || categories.Count != 0)
+			categories.Add(PrefixCategory.AnyWeapon);
+
+		if (IsAPrefixableAccessory())
+			categories.Add(PrefixCategory.Accessory);
+
+		return categories;
+	}
```
Chicken-Bones: Can we bake this into a `static List<PrefixCategory>[]` or `PrefixCategory[][]` by type? Perhaps with some de-duplication?

Chicken-Bones (commented):
I have some minor performance concerns, but aside from that. The entire prefix category system is currently a disaster. Everything about it is hard coded. Mod categories don't exist. Mod prefixes can only belong to one category while vanilla prefixes can belong to multiple...

patches/tModLoader/Terraria/ModLoader/PrefixLoader.cs
Tyfyter: Yes, I I'd guess I must've accidentally autocompleted something that added that and not noticed.

patches/tModLoader/Terraria/Item.TML.cs
Tyfyter: If I'm understanding you correctly and you mean an array mapping item IDs to the collections of prefix categories that item type can get, I should be able to do that pretty quickly;
If you haven't said that I've misunderstood by the next opportunity I have to implement this, it will be done shortly;
I didn't initially consider it because it seemed more practical to avoid spreading changes out across more files than necessary or making maintaining this part of maintaining something else, but I was forgetting that modders regularly find ways to make any performance cost problematic.

Chicken-Bones: Yes, I didn't identify any high traffic call sites in tML, but it seems like the thing we can do fairly easily to avoid allocs. Modders don't tend to be too perf aware. Filling out an array during `PostSetupContent` should suffice

Chicken-Bones: If you want to be particularly zealous you can do the 'compute if not baked' approach like we did for ~~something else~~ `GetContent` recently because items can change before `PostSetupContent`

Tyfyter: Updated to a 'compute if not baked' approach;
I'm not certain if this is the most performant possible implementation, but it should definitely be better for anything that spams GetPrefixCategories than the previous one

patches/tModLoader/Terraria/ModLoader/DamageClass.cs
```diff
@@ -81,6 +81,13 @@ public abstract class DamageClass : ModType, ILocalizedModType
 	/// <param name="damageClass">The <see cref="DamageClass"/> you want to inherit effects from.</param>
 	/// <returns><see langword="false"/> by default - which does not let any other classes' effects trigger on this <see cref="DamageClass"/>.</returns>
 	public virtual bool GetEffectInheritance(DamageClass damageClass) => false;
+	/// <summary> 
+	/// This lets you define the classes that this <see cref="DamageClass"/> will count as (other than itself) for the purpose of prefixes<br/>
+	/// </summary>
+	/// <remarks>Return <see langword="true"/> for each <see cref="DamageClass"/> you want to inherit from</remarks>
+	/// <param name="damageClass">The <see cref="DamageClass"/> you want to inherit prefixes from.</param>
+	/// <returns><see langword="false"/> by default - which does not let any other classes' prefixes roll or remain on this <see cref="DamageClass"/>.</returns>
+	public virtual bool GetPrefixInheritance(DamageClass damageClass) => GetEffectInheritance(damageClass);
```
Chicken-Bones: Documentation needs to say `returns GetEffectInheritance by default` rather than `false`

Chicken-Bones: Should also add the note to use `GetsPrefixesFor` instead of this method when querying. And the same for `CountsAsClass` to `GetEffectInheritance`

Tyfyter: Added and fixed the comments

JavidPack:
Here is the code for comparing prefix results, as mentioned in the discord discussion, in case we need to run the tests again in the future for this or another prefix related PR. Modders can also run this code in their own mod to see how this affects prefixes for items in their own mod by running this code during the preview. To run these tests, build and reload the mod in stable and dev/preview. Do this from VS for dev/preview since there is a `#if DEBUG` preventing a compilation error, or just edit the code to uncomment the code.

```cs
class PrefixChecker : ModSystem
{
	public override void PostAddRecipes() {
		WritePrefixes(false);

		// Comment out for stable build.
#if DEBUG
		WritePrefixes(true);
#endif
	}

	private static void WritePrefixes(bool useGetPrefixCategories) {
		StringBuilder output = new();

		for (int i = 0; i < ItemLoader.ItemCount; i++) {
			Item item = new Item(i);
			//output.AppendLine($"{i}: {item.Name}");
			bool any = false;
			List<int> prefixes = new List<int>();
			for (int j = 0; j < PrefixLoader.PrefixCount; j++) {
				bool can = item.CanApplyPrefix(j);
				if (!any && can) {
					any = true;
					output.AppendLine($"\t{i}: {item.Name}, {item.DamageType.Name}");
					if (item.ModItem is ModItem modItem) {
						output.AppendLine($"\tMeleePrefix: {modItem.MeleePrefix()}, RangedPrefix: {modItem.RangedPrefix()}, MagicPrefix: {modItem.MagicPrefix()}, WeaponPrefix: {modItem.WeaponPrefix()}, IsAPrefixableAccessory: {item.IsAPrefixableAccessory()}");

						/*
						output.AppendLine($"\tItem.GetPrefixCategory: {item.GetPrefixCategory()}");
						if (useGetPrefixCategories)
							output.AppendLine($"\tItem.GetPrefixCategories: {string.Join(", ", item.GetPrefixCategories())}");
						*/

						if (useGetPrefixCategories) {
#if DEBUG
							output.AppendLine($"\tPrefixCategory(s): {string.Join(", ", item.GetPrefixCategories())}");
#endif
						}
						else {
							output.AppendLine($"\tPrefixCategory(s): {item.GetPrefixCategory()}");
						}
					}
				}
				if (can) {
					prefixes.Add(j);
					//output.AppendLine($"{j}: {Lang.prefix[j].Value}");
				}
			}

			if (any) {
				output.AppendLine($"Prefixes: {string.Join(", ", prefixes.Select(x => Lang.prefix[x].Value))}");
				output.AppendLine();
			}
		}
		File.WriteAllText(Path.Combine(Main.SavePath, $"PrefixResults_{BuildInfo.versionTag.Replace("/", "")}{(useGetPrefixCategories ? "_GetPrefixCategories" : "")}.txt"), output.ToString());
	}
}
```

The code will spit out `PrefixResults_v{version}.txt` in each save directory. Use your favorite diff viewer to compare the results.

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
```diff
@@ -47,6 +47,7 @@ public override StatInheritanceData GetModifierInheritance(DamageClass damageCla
 	}
 
 	public override bool GetEffectInheritance(DamageClass damageClass) => damageClass == Melee;
+	public override bool GetPrefixInheritance(DamageClass damageClass) => damageClass == Melee;
```
Chicken-Bones: Unnecessary?

patches/tModLoader/Terraria/Item.cs.patch
```diff
@@ -916,10 +916,11 @@
...
  		while (flag) {
  			flag = false;
+-			if (rolledPrefix == -1 && unifiedRandom.Next(4) == 0)
+-				rolledPrefix = 0;
 +
 +			if (applyPrefixOverride is true) { } else
- 			if (rolledPrefix == -1 && unifiedRandom.Next(4) == 0)
--				rolledPrefix = 0;
++			if (prefixWeWant == -1 && unifiedRandom.Next(4) == 0)
 +				return true; //rolledPrefix = 0;
```
Chicken-Bones: Comment out the vanilla code with `/* and */` to make patch clearer.

Add a comment with `//` explaining why we check `prefixWeWant` instead of `rolledPrefix`

No need for the weird `else` on same line formatting, since we're not trying to keep the existing vanilla lines anymore. You can just combine the `applyPrefixOverride is not true &&` into the condition

patches/tModLoader/Terraria/ModLoader/VanillaDamageClasses.cs
Tyfyter: May have been left over from a design without any default inheritance, unnecessary and removed.

patches/tModLoader/Terraria/Item.cs.patch
Tyfyter: The reason for checking prefixWeWant instead of rolledPrefix is that the unexplained change to returning breaks the part below that forces it try again if no prefix is rolled with prefixWeWant equal to -2;
I don't believe I can sufficiently explain this change without knowing why there was a change that needed to be fixed in the first place, given it seems to originate from [1.4.4.2: Reimplement prefix hooks](https://github.com/tModLoader/tModLoader/commit/2204125d59b16374806b9663f27fbe7a3214632d), was the original change simply intended to improve performance?

PR #4178, opened 2024-04-24
Terraria legacy beta branch support

JavidPack:
Launching tModLoader when the "v1.0.6.1 - Undeluxe Edition" Terraria beta branch is active will prevent tModLoader from running. 

The current error is this:    
image    

This error is a little misleading in this situation, so this PR will instead show the following error:    
image


Chicken-Bones:
Too bad we can't get an image in there. Can we get a link to an image showing the steam betas tab and telling people they need to select none?

PR #4182, opened 2024-04-26
Support Unlisted Publicity Option

Solxanich:
Adds support for the Unlisted publicity option to match Workshop options that we have enabled

image

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/AWorkshopPublishInfoState.cs.patch
@@ -123,7 +123,7 @@
  	}
  
  	protected abstract bool TryFindingTags(out FoundWorkshopEntryInfo info);
-@@ -328,34 +_,39 @@
+@@ -328,41 +_,46 @@
  	private UIElement CreatePublicSettingsRow(float accumulatedHeight, float height, string tagGroup)
  	{
  		CreateStylizedCategoryPanel(height, "Workshop.CategoryTitlePublicity", out var entirePanel, out var innerPanel);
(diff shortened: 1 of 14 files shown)


Rijam:
Slightly modified Bestiary unlocked icon which I think would fit for the unlisted icon:
Unlisted
Or slightly modified Chest Lock item:
Unlisted2

JavidPack:
Finished this up. I wasn't able to use a new texture because `GroupOptionButton` class is hard-coded to use vanilla texture paths, and I didn't want to make a workaround for a class used all over the codebase. I ended up using a different texture though and adjusted the spacing and text scale a bit.

image


I also fixed an issue with `currPublishID`,. The issue was `currPublishID` might be set to something non-0 if you visit the publish menu for an existing mod. After that attempting to publish a new mod would erroneously use that non-0 value in all of the other logic, attempting to overwrite that workshop entry instead of making a new workshop item.


I did run into a common issue where steam complains about "The content folder for the published item is missing?!". I was able to do this by publishing the mod, then deleting it in steam, then attempting to publish it again. The issue is caused by this:
```
if (existingID == 0 && AWorkshopEntry.TryReadingManifest(contentFolderPath + Path.DirectorySeparatorChar + "workshop.json", out var info))
	_publishedFileID = new PublishedFileId_t(info.workshopEntryId);
```
There is a TODO: `//TODO: We ought to delete the TemporaryFolder after successful publishing to prevent future issues if they delete and attempt to re-pub`

I'm not sure about the solution, I figure maybe just don't `TryReadingManifest`, but I'm not sure if that will cause any issues. I'm just making a note of it here so we have something to find in the github search. A workaround is to just delete the `My Games\Terraria\tModLoader\Workshop` folder I think.

PR #4189, opened 2024-05-01
Update Brazilian Portuguese localization, add pt-BR translation for steam page

Pixelnando:


JavidPack:
Thanks a bunch, these should be all live soon.

PR #4194, opened 2024-05-01
Implement error message for 32 bit Windows users rather than black window quickly closing

JavidPack:
32 bit Windows users attempting to launch the game will only see a black command prompt appear for a split second and then close. This can be difficult to diagnose, as they also don't have a Logs folder. If they manually open a command prompt and run `start-tModLoaderServer.bat`, they'll see the actual error, `busybox64.exe is not compatible with the version of Windows you are running`.

This PR adds a check for 32 bit OS and will show a message to the user, clarifying the situation and avoiding confusing troubleshooting.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/busybox-sh.bat
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/busybox-sh.bat
@@ -3,4 +3,8 @@ for /f "tokens=4-5 delims=. " %%i in ('ver') do (
 	set WINDOWS_MINOR=%%j
 )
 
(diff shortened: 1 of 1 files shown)


Chicken-Bones (approved):

JavidPack:
Some notes: someone on preview today was incorrectly being detected as 32 bit. Turns out the top rated stack overflow answer that we adapted does have a logic error.

image

The issue was a string find command finding x86 in an unrelated section. The new code (https://github.com/tModLoader/tModLoader/commit/753d9ccfaab5710e307506fe7782c9ca494f22bf) doesn't require a registry query and seems to be the more intended approach.

PR #4199, opened 2024-05-05
Addition of Example Arrow

harlowo:
Adds in an example arrow item and projectile, due to the lackage of a pre-existing one.
No


JavidPack:
Thanks. I made some changes so that the example has more teaching utility.

PR #4206, opened 2024-05-09
Add error when it is detected that a recently built mod will not be loaded due to version or modpack priority (Fix #4181)

### LINKED ISSUE #4181: Detect and Warn if modder is debugging or building a mod that won't be selected to load due to version number
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

Sometimes a modder will clone a mod git and try to debug, only to be confused as their changes do not take effect. The latest report of this is a modder cloning the main branch that had 0.6 as the build, but steam had a 0.6.1.1 version published. This is either because the author had not pushed some local changes yet or the changes to build.txt were on another branch.

This also happens for modders confused about Version, for example a modders changing v0.200 to v0.3, thinking it is a higher version.

It would be useful to warn of this issue. It might be possible to to detect this when debugging directly. If not, build and reload could warn of this issue. The warning should be full page, as this is definitely a mistake and probably never intentional.

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

JavidPack:
Fix #4181

~Currently only in-game building warning is implemented. It might get complicated implementing the debugging fix and a user experiencing this would likely try building in-game as a sanity check. This issue is also pretty rare so it might not be worth more effort to account for.~

Detects mods built this session and actively debugging mods built just before the game launched to warn the user. The checks should prevent any false positives.

Old messages:

image

image

Updated:

```
"LoadErrorRecentlyBuiltLocalModWithLowerVersionWorkshop": "{0} v{1} was recently built, but was not selected to load because the subscribed workshop copy (v{2}) is newer.\nEither update build.txt and rebuild the mod or temporarily unsubscribe from the mod.",
"LoadErrorRecentlyBuiltLocalModWithLowerVersionModpack": "{0} v{1} was recently built, but was not selected to load because a frozen copy (v{2}) is present in the active modpack.\nThe mod pack will need to be disabled to test this mod.",
```

image


Chicken-Bones:
I think this would be better as a warning on loading mods? It shouldn't prevent the building of mods themselves. 

Perhaps for the wording:
```
Census 0.5.2.1 was recently built, but was not selected to load because the subscribed workshop copy (0.5.2.5) is newer
Census 0.5.2.1 was recently built, but was not selected to load because a frozen copy (version) is present in the active modpack
```

PR #4207, opened 2024-05-09
Fix #3167, Remove ValidModBrowserSignature usage and remove unusable OnlyDownloadSignedModsFromServers config

### LINKED ISSUE #3167: .NET Error when trying to join server on RHEL 9 based servers
### Version

Latest 1.4-stable

### OS

Linux

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When players try to join a hosted world on an RHEL9 based server, they cannot get past the "Server Found" message. 

On the server side, the administrator receives the following message: 
error:03000098:digital envelope routines::invalid digest (see server.log for full trace)

### Log File

[server.log](https://github.com/tModLoader/tModLoader/files/10333779/server.log)

### Steps to reproduce

1. Host RHEL 9 based server (I have set up an Oracle Cloud instance based on Oracle Linux 9 with ARM64 CPUs)
2. Try to join password-protected TmodLoader server hosted on this instance
3. Get stuck on "Server Found" message

### Expected Behavior

Players should connect to the server and load the world successfully.

### Actual Behavior

When a player tries to join a hosted world, they cannot get past the "Server Found" message..

### Reproduction frequency

Everytime when a player tries to join the server and SHA1 signature creation and verification is disabled.

### Additional Information

It seems to have something to do with .NET and strong name signing as mentioned here: https://github.com/dotnet/runtime/issues/65874

Strong name signing by using SHA1 signature verification is disabled by default on RHEL 9 based servers:
https://gitlab.com/redhat/centos-stream/rpms/openssl/-/commit/78fb78d30755ae18fdaef28ef392f4e67c662ff6

A fix for this has already been integrated into .NET:
https://github.com/dotnet/runtime/pull/69668

However, it seems that the TmodLoader server runs with an older .NET runtime, so this solution is not implemented yet.

Workaround is to enable RSA-SHA1 Signing  with `export OPENSSL_ENABLE_SHA1_SIGNATURES=1`.

JavidPack:
We don't use it anymore and it is causing issues in rare Linux environments.

Network api changed.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
@@ -65,6 +65,7 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 
 	internal byte[] Signature { get; private set; } = new byte[256];
 
+	/*
 	private bool? validModBrowserSignature;
 	internal bool ValidModBrowserSignature {
 		get {
@@ -74,6 +75,7 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 			return validModBrowserSignature.Value;
 		}
 	}
+	*/
 
 	internal TmodFile(string path, string name = null, Version version = null)
 	{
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
Better to remove wholesale than comment out. We can always revert if we want to restore it in the future.

PR #4208, opened 2024-05-11
Make Mod Publish Button Dynamic - QoL UI Improvement

Solxanich:
The Publish Button on the Publishing UI has been updated to be 'Update' instead of 'publish' when it is an update.
Also moved around some code for consistency
 
image


Solxanich:
Final UI: 
image
image

PR #4215, opened 2024-05-14
Add RecentlyPublished sort, Time Period sort, and update Hot sort. Better workshop user experience.

JavidPack:
# RecentlyPublished 

There has been some requests for sorting by recently published. The current sort is recently updated, which isn't interesting for someone keeping an eye out for new mods. New mods don't show up on recently updated either until they are updated after their initial publication.

~**Discussion**: It might actually be better to default to recently published, recent updates aren't really interesting for mod users looking for something new and interesting.~

Icon art can also be updated if anyone wants to make something better.

image

# Improve discoverability and Reduce search confusion

Another aspect of this PR is bringing the in-game workshop more in-line with the workshop site itself. 

The main goals are:
1. Improve discoverability
2. Reduce search confusion

To do this, I'm changing the default filter to Hot and adding a "time period" toggle. Hot + 1 week matches what users see when visiting https://steamcommunity.com/app/1281930/workshop/. The current default is "Recently Updated". Another filter called "Recently Published" is also added in the PR. "Recently Published" could be a better default for discoverability, but I feel that Hot+1 Week is a good compromise for New and Rising mods.

Search is confusing because when searching in Hot, it actually searches by relevancy. Now that Hot is the new default filter, users opening up the workshop should be able to quickly search for a mod they heard about and find it, rather than seeing somewhat matching results sorted by "Recently Published". The tooltip will hint at this.

The UI has been updated to dynamically update tooltips and filter visibility to help users understand the results they are seeing. The new icon here will also need to be updated, this is just placeholder art.

https://github.com/tModLoader/tModLoader/assets/4522492/c5bcea34-b549-4abb-8ca4-14d6202891a7


Solxanich:
I agree with the direction. 
Recently updated doesn't ensure new mods get seen.

Recently published fixes that gap and ensures new mods have a fair footing. 

Can we include searching by recently published in the 'SearchByText' early return condition? 
Do we want it to be accurate searching by default?

JavidPack:
PR updated with approach voted on by fellow modders. Should be good to review.

patches/tModLoader/Terraria/Social/Steam/SteamedWraps.cs
```diff
@@ -231,9 +233,11 @@ public static EUGCQuery CalculateQuerySort(QueryParameters qParams)
 			return EUGCQuery.k_EUGCQuery_RankedByTextSearch;
 
 		return (qParams.sortingParamater) switch {
+			ModBrowserSortMode.Hot when qParams.days == 0 => EUGCQuery.k_EUGCQuery_RankedByVote, // Corresponds to "Most Popular" for "All Time" on workshop website
 			ModBrowserSortMode.DownloadsDescending => EUGCQuery.k_EUGCQuery_RankedByTotalUniqueSubscriptions,
-			ModBrowserSortMode.Hot => EUGCQuery.k_EUGCQuery_RankedByPlaytimeTrend,
```
Chicken-Bones, line 235: No longer any filter for `k_EUGCQuery_RankedByPlaytimeTrend`?

JavidPack: `Hot` is now using `k_EUGCQuery_RankedByTrend`, which is votes. Old `Hot`/`k_EUGCQuery_RankedByPlaytimeTrend` was virtually indistinguishable from `k_EUGCQuery_RankedByTotalUniqueSubscriptions`/`DownloadsDescending` , so it didn't really add anything.

Also, Votes (`k_EUGCQuery_RankedByPlaytimeTrend`) for 1 Week and the other time periods shows up-and-coming mods based on the time period. `k_EUGCQuery_RankedByPlaytimeTrend` for different time periods is basically all the same results, so it would too be useless.

Chicken-Bones (approved):

PR #4220, opened 2024-05-17
Clear ReflectionHelper cache

Solaestas:
### What is the bug?

When a mod is not fully unloaded (quite common), `MonoMod.Utils.ReflectionHelper` retains the previously loaded mod assembly. After force reloading the mod, MonoMod incorrectly uses the cached data to generate dynamic methods. This can result in generic methods like `GetInstance<T>` or `GetGlobalPlayer<T>` returning null where T is a type from mod assembly

These often occur when you want to patch another mod or avoiding `EmitDelegate` for performance

Example:

```cs
public class TestSystem : ModSystem
{
	public record JsonData(string Name);

	public override void Load()
	{
		IL_Main.Update += IL_Main_Update;
		JsonConvert.SerializeObject(new JsonData("Prevent fully unloaded"));
	}

	private void IL_Main_Update(ILContext il)
	{
		var cursor = new ILCursor(il);
		var method = ModContent.GetInstance<TestSystem>;
		cursor.Emit(OpCodes.Call, method.Method);
		cursor.EmitDelegate((TestSystem system) =>
		{
			Mod.Logger.Info($"Emit Call return is null: {system is null}");
			system = ModContent.GetInstance<TestSystem>();
			Mod.Logger.Info($"Emit Delegate return is null: {system is null}");
		});
	}
}

```

Expected output

- Before Reload: (false, false)
- After Reload: (true, false)

### How did you fix the bug?

Clear the cache dictionary

Replace emitting calling a generic method info with `EmitDelegate`

Remove generic method calling while patching another mod, which is troublesome

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
+++ b/patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
@@ -3,6 +3,7 @@ using MonoMod.RuntimeDetour;
 using MonoMod.RuntimeDetour.HookGen;
 using MonoMod.Utils;
 using System;
+using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
@@ -125,6 +126,16 @@ public static class MonoModHooks
 		HookEndpointManager.Clear();
 		assemblyDetours.Clear();
 		_hookCache.Clear();
+
+		var type = typeof(ReflectionHelper);
+		FieldInfo[] caches = [
+			type.GetField("AssemblyCache", BindingFlags.NonPublic | BindingFlags.Static),
+			type.GetField("AssembliesCache", BindingFlags.NonPublic | BindingFlags.Static),
+			type.GetField("ResolveReflectionCache", BindingFlags.NonPublic | BindingFlags.Static),
+		];
+		foreach(var cache in caches) {
+			((IDictionary)cache.GetValue(null)).Clear();
+		}
 	}
 
 	#region Obsolete HookEndpointManager method replacement
--- a/patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
+++ b/patches/tModLoader/Terraria/ModLoader/MonoModHooks.cs
@@ -127,13 +127,14 @@ public static class MonoModHooks
 		assemblyDetours.Clear();
 		_hookCache.Clear();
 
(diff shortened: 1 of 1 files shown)


JavidPack:
Works as described.

Chicken-Bones:
While we do need to clear the caches for unloading anyway, clearing them is not a complete solution to the issue presented. I investigated it thoroughly and have added failing test cases to the MonoMod repo to investigate a more complete fix.

In summary, the issue can happen any time a generic method is reused with two different type parameters which have the same full name. 

I'm happy to accept this as a temporary mitigation, and to help with cleanup.

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -183,6 +183,21 @@ private static Assembly hook_ValidateAssemblyNameWithSimpleName(orig_ValidateAss
...
+		FieldInfo[] caches = [
+			type.GetField("AssemblyCache", BindingFlags.NonPublic | BindingFlags.Static),
+			type.GetField("AssembliesCache", BindingFlags.NonPublic | BindingFlags.Static),
+			type.GetField("ResolveReflectionCache", BindingFlags.NonPublic | BindingFlags.Static),
+		];
+		foreach(var cache in caches) {
+			if(cache == null) {
+				continue;
+			}
+			if (cache.GetValue(null) is System.Collections.IDictionary dict) {
```
Chicken-Bones: We don't want to code defensively here, because `MonoMod` is a dependency that we tightly depend on and control, and we don't have a test suite that could catch regressions from updates. 

In cases where external code is involved, defensive programming is warranted, but we would still want to log a warning rather than silently failing.

Suggest:
```cs
((IDictionary)cache.GetValue(null)).Clear();
```

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -183,6 +183,21 @@ private static Assembly hook_ValidateAssemblyNameWithSimpleName(orig_ValidateAss
 			GC.Collect();
 			GC.WaitForPendingFinalizers();
 		}
+
+		var type = typeof(MonoMod.Utils.ReflectionHelper);
```
Chicken-Bones: The right place for this is `MonoModHooks.Clear`

Chicken-Bones (changes requested):

Issue #4083, opened 2024-02-27
Hooks to change Petting, Chat Bubble, Party Hat, and Emote Bubble positions for NPCs.


Rijam:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

The ability to easily change the position and `isPetSmall` for petting Town Pets/Slimes.
The ability to easily change the position and SpriteEffect for the chat bubble, party hat, and emote bubble for NPCs.

*The NPCLoader code shown in the drop downs are just pseudocode. I know the NPC argument probably won't work as is.*

#### Petting

* `ModNPC.SetPettingInfo(ref int playerDistance, ref bool isPetSmall)`
* `GlobalNPC.SetPettingInfo(int npcType, ref int playerDistance, ref bool isPetSmall)`

Hooks to change how far/close the player stands to the Town Pet. The `isPetSmall` bool will make the player's arm either angled up or down.

<details><summary>Vanilla code Player.GetPettingInfo() **Click to Expand**</summary>

```diff
private void GetPettingInfo(int animalNpcIndex, out int targetDirection, out Vector2 playerPositionWhenPetting, out bool isPetSmall)
{
	NPC nPC = Main.npc[animalNpcIndex];
	targetDirection = ((nPC.Center.X > base.Center.X) ? 1 : (-1));
	int num = 36;
	isPetSmall = false;
	switch (nPC.type) {
		case 637: // <- Town Cat
			isPetSmall = true;
			num = 28;
			break;
		case 656: // <- Town Bunny
			isPetSmall = true;
			num = 24;
			break;
		case 670: // <- Nerdy Slime
		case 678: // <- other Town Slimes
		case 679:
		case 680:
		case 681:
		case 683:
			isPetSmall = true;
			num = 26;
			break;
		case 682: // <- Surly Slime
			isPetSmall = true;
			num = 22;
			break;
		case 684: // <- Squire Slime
			isPetSmall = true;
			num = 20;
			break;
	}
	
+	NPCLoader.SetPettingInfo(nPC, ref num, ref isPetSmall);

	playerPositionWhenPetting = nPC.Bottom + new Vector2(-targetDirection * num, 0f);
}
```
	
</details>
  
#### Chat Bubble

* `ModNPC.NPCChatBubblePosition(ref Vector2 position, ref SpriteEffects spriteEffects)`
[...]

PR #4222, opened 2024-05-18
Modify Petting, Chat Bubble, Party Hat, and Emote Bubble positions

Rijam:
### What is the new feature?

Implements #4083

The ability to easily change the position and isPetSmall for petting Town Pets/Slimes.
The ability to easily change the position and SpriteEffect for the chat bubble, party hat, and emote bubble for NPCs.

New sets and hooks:
* Petting:
  * `int NPCID.Sets.PlayerDistanceWhilePetting[]` The distance away the player stands while petting a Town Pet. Default is 36.
  * `bool NPCID.Sets.IsPetSmallForPetting[]` The player's arm will be angled up while petting by default. It will be angled down if true.
* Chat Bubble: (This is the chat bubble that shows up when you hover over a Town NPC)
  * `ModNPC.ChatBubblePosition(ref Vector2 position, ref SpriteEffects spriteEffects)` The position is the final position right before it gets drawn. The SpriteEffects lets you flip the direction of the chat bubble.
  * `GlobalNPC.ChatBubblePosition(NPC npc, ref Vector2 position, ref SpriteEffects spriteEffects)` Same but GlobalNPC.
* Emote Bubble:
  * `ModNPC.EmoteBubblePosition(ref Vector2 position, ref SpriteEffects spriteEffects)` The final position right before it gets drawn.
  * `GlobalNPC.EmoteBubblePosition(NPC npc, ref Vector2 position, ref SpriteEffects spriteEffects)`
* Party Hat:
  * `ModNPC.PartyHatPosition(ref Vector2 position, ref SpriteEffects spriteEffects)` The final position right before it gets drawn.
  * `GlobalNPC.PartyHatPosition(NPC npc, ref Vector2 position, ref SpriteEffects spriteEffects)`
  * New group for `NPCID.Sets.NPCFramingGroup[Type] = 8` which has no offset for the party hat.

I am aware that these features are super niche and will not be utilized by a majority of modders.

*(Copied from the issue)*

Currently, changing the petting settings and emote bubble position require a detour. Changing chat bubble and party hat position requires an IL edit (unless you copy paste the entire thing into a detour or something).

The current only ways to edit the position of the party hat include `NPCID.Sets.HatOffsetY[]` and `NPCID.Sets.NPCFramingGroup[]`. The first one is only a Y offset and the second one only has 8 options available. Since they are sets, you are encuraged to put them in SetStaticDefaults which means they only run once on mod load. [More info about NPCFramingGroup in my Town NPC Guide](https://gist.github.com/Rijam/c67eb86eb3aef9db7ac474ad5ef7dfba#party-hat)

Some specific examples:
I've made a clone of the Town Cat for Town NPC guide. The raised position when the cat is sitting on the chair is achieved with a detour of `Main.NPCAddHeight()`, but as you can see the chat bubble isn't raised up either. You can also see that the party hat isn't on the cat clone's head because its head isn't centered horizontally.
CatAndCatClone
Another example is the Draken from SGAmod (1.3 mod, isn't available for 1.4). His head is not centered so the best we can do is put the party hat on his back. The emote bubbles also come out of his behind instead of his head.
DrakenExample

* ~~The Emote Bubble position gets Floored before it is drawn. Right now the NPCLoader hook is after the Floor(), but it could be before.~~ Done
* The issue has an alternate implementation for the Emote Bubble.
* I was thinking of also adding a new `NPCID.Sets.NPCFramingGroup` that is just a 0 so there aren't any included party hat offsets. I need to investigate if that is necessary, and I don't know if that is a breaking change or not.
* Obviously the set and hook names could be changed. And of course, the xml summaries could be changed.
* This is the first time I've added a new hook, so maybe the NPCLoader stuff could be changed.

A Town Cat clone that has the chat bubble, party hat, and petting settings that exactly match the vanilla Town Cat.
CorrectTownPet

<details><summary>New code for my Tutorial Town Pet **Click to Expand**</summary>

```cs
	// Other stuff omitted.
	
	NPCID.Sets.PlayerDistanceWhilePetting[Type] = 28; // Distance the player stands from the Town Pet to pet.
	NPCID.Sets.IsPetSmallForPetting[Type] = true; // Arm is angled down while petting.
}

public override void ChatBubblePosition(ref Vector2 position, ref SpriteEffects spriteEffects) {
	if (NPC.ai[0] == 5f) { // Sitting in a chair.
		position.Y -= 18f; // Move upwards.
	}
}

public override void EmoteBubblePosition(ref Vector2 position, ref SpriteEffects spriteEffects) {
	// Flip the emote bubble and move it.
	spriteEffects = NPC.spriteDirection == -1 ? SpriteEffects.FlipHorizontally : SpriteEffects.None;
	position.X += NPC.width * NPC.direction;
}

public override void PartyHatPosition(ref Vector2 position, ref SpriteEffects spriteEffects) {
	// Move the party hat forward so it is actually on the cat's head.
	int frame = NPC.frame.Y / NPC.frame.Height;
	int xOffset = 5;
	switch (frame) {
		case 19:
		case 22:
		case 23:
		case 24:
		case 25:
		case 26:
		case 27:
			xOffset -= 2;
			break;
		case 11:
		case 12:
		case 13:
		case 14:
		case 15:
			xOffset += 2;
			break;
	}
	position.X += xOffset * NPC.spriteDirection;
	
	// Move it up to match the location of the head when sitting in a chair.
	if (NPC.ai[0] == 5f) {
		position.Y += -12;
	}
}

// This is not a new hook, but it is how to achieve moving the NPC visually without detouring Main.NPCAddHeight
public override bool PreAI() {
	// Move up visually while sitting in a chair.
	if (NPC.ai[0] == 5f) {
		DrawOffsetY = -10;
	}
	else {
		DrawOffsetY = 0;
	}
	return base.PreAI();
}
```

</details>

Example of the Guide where the Emote Bubble is flipped and moved to the front.
GuideEmoting

### ExampleMod updates
* An example for the Guide with his party hat moved and his emote bubble moved (like the example shown above).
* Example Town Pet: An NPC like the Town Cat/Dog/Bunny who has multiple variations.
  * Example Town Pet License


JavidPack:
Looking good. Flooring makes more sense after all edits from mods, we don't expect modders to know to do that math and I assume that's for avoiding interpolation of the pixels of the bubble drawing.

JavidPack:
Looking great. I made some changes, you might want to update your wiki guide. Also, feel free to link to your wiki guide on our wiki if that is what you want, like right below "Basic NPC Spawning Guide" on the main page.

First issue was you weren't clearing the bools in ModSystem.ClearWorld. 2nd thing is you might want to show off that approach to localizing names. Another thing is the Item.ResearchUnlockCount = 5;.

I also moved the ExampleTownPetLicense logic from OnConsumeItem to UseItem. OnConsumeItem is only called if the item actually gets consumed, and I can imagine someone would make a mod making pet licenses no longer one time use. OnConsumeItem is intended to be for the effects of the item itself reducing in stack size, the effects of "using" the item go in UseItem.

PR #4227, opened 2024-05-20
[1.4.4] Update Polish Localization

TomaszOleszko:
Updated the localization file with Polish translations for all messages.


JavidPack:
Thanks a bunch.

PR #4229, opened 2024-05-23
Run expensive tasks in Mod Sources asynchronously (Fixes #2956)

### LINKED ISSUE #2956: Mod Sources menu slow
### Version

Latest 1.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

Splitting from #2578:

This is what is causing the Mod Sources freeze:
https://github.com/tModLoader/tModLoader/blob/8ef29f1a7ec0aa7d7890d86ec98c81237d28239d/patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs#L159-L161

Initial test did not show any glaring issues, so it is probably just an expensive operation.

CB instructinos: 

> Use Directory.EnumerateFiles(...).Any() as well
just queue on a worker thread, and set some visibility properties

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

oli-x64:
### What is the bug?
#2956 - Lag when entering Mod Sources with a large amount of mods present, due to it enumerating files for each mod on the main thread.

### How did you fix the bug?
Files are now fetched asynchronously using a separate thread per mod source. The UI is then updated on the main thread to prevent issues adding the button after the files are enumerated.


patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
```diff
@@ -383,4 +335,95 @@ internal static void PublishModCommandLine(string modName)
...
+	{
+		langFileTask = Task.Run(
+			() => Directory.GetFiles(_mod, "*.lang", SearchOption.AllDirectories),
+			_modSourcesCts.Token
+		);
+
+		sourceUpgradeTask = Task.Run(
+			() => SourceManagement.SourceUpgradeNeeded(_mod),
+			_modSourcesCts.Token
+		);
```
steviegt6: While we're at it, it would be good to exclude common paths that aren't relevant (`.git/`, `.vs/`, `obj/`, `bin/`). A good option would be ignoring directories that fall under buildIgnore, but I'm not sure whether that's accessible in this context let alone whether it's reasonable to implement.
Either way, the aforementioned directories are reliably ignorable.

steviegt6 (approved):
I like it. Reasonable fixes.
If possible, adding a little loading indicator for as long as it's still processing files could be good too. Shrug.

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -391,7 +391,7 @@ internal void Populate()
 			foreach (string sourcePath in modSources) {
 				var modName = Path.GetFileName(sourcePath);
 				var builtMod = modFiles.Where(m => m.Name == modName).Where(m => m.location == ModLocation.Local).OrderByDescending(m => m.Version).FirstOrDefault();
-				_items.Add(new UIModSourceItem(sourcePath, builtMod));
+				_items.Add(new UIModSourceItem(sourcePath, builtMod, _cts));
```
Chicken-Bones: best practice is to pass the `_cts.Token`, rather than the source.

patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
```diff
@@ -383,4 +335,95 @@ internal static void PublishModCommandLine(string modName)
...
+	private void StartUpgradeTasks()
+	{
+		langFileTask = Task.Run(
+			() => Directory.GetFiles(_mod, "*.lang", SearchOption.AllDirectories),
+			_modSourcesCts.Token
+		);
+
+		sourceUpgradeTask = Task.Run(
+			() => SourceManagement.SourceUpgradeNeeded(_mod),
+			_modSourcesCts.Token
```
Chicken-Bones: Passing a `CancellationToken` to a `Task` is better than nothing, but will only cancel it if it doesn't start before the cancellation is requested. The best practice is to actually check it (`ThrowIfCancellationRequested`) as you're doing work in the long running methods. Unfortunately, `GetFiles` doesn't take one so this will have to do there.

patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
```diff
@@ -175,90 +181,36 @@ protected override void DrawSelf(SpriteBatch spriteBatch)
...
-					}
-
-					upgradeLangFilesButton.Remove();
-				};
+			langFileTask = null;
+		}
 
-				Append(upgradeLangFilesButton);
+		// Display Run tModPorter when .csproj is valid
+		if (sourceUpgradeTask?.IsCompleted ?? false) {
```
Chicken-Bones: The best syntax is probably `if (sourceUpgradeTask is { IsCompleted: true })`

Chicken-Bones (changes requested):

PR #4236, opened 2024-05-28
AllowVanillaClients removal

### LINKED ISSUE #4232: NPC Strike packet ignores AllowVanillaClients flag
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

Looking through the tModLoader protocol modifications to the Terraria protocol, I noticed that the NPC Strike logic doesn't take AllowVanillaClients into account. This would result in incorrect packet being sent to vanilla clients when connecting to a tModLoader server.

Refs:
https://github.com/tModLoader/tModLoader/blob/1.4.4/patches/tModLoader/Terraria/NetMessage.cs.patch#L191
https://github.com/tModLoader/tModLoader/blob/1.4.4/patches/tModLoader/Terraria/MessageBuffer.cs.patch#L327

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

JavidPack:
We retired AllowVanillaClients, but it is still in the codebase. Removing it should help clean up patches and make following the code flow easier. Many labels that skip over now unreachable code are also removed in favor of being commented out.

There is also an incidental fix for syncing `player.hairDye`, it seems that the logic was accidentally flipped a couple years back. It is likely unreported since it would require a lot of modded hair dyes to be added to actually cause sync issues. (This is assuming Write(byte) and Write7BitEncodedInt behave the same for values under 128, which I think is true. There are only 12 vanilla hair dye.)

Brought up by and fixes #4232

### DIFF
--- a/patches/tModLoader/Terraria/MessageBuffer.cs.patch
+++ b/patches/tModLoader/Terraria/MessageBuffer.cs.patch
@@ -137,18 +137,17 @@
  					Item[] array3 = null;
  					Item[] array4 = null;
  					int num261 = 0;
-@@ -411,20 +_,31 @@
+@@ -411,20 +_,29 @@
  					}
  
  					if (flag16) {
 +						player17.trashItem = ItemIO.Receive(reader, readStack: true);
-+						goto SkipVanillaRead;
-+
++						/*
  						player17.trashItem = new Item();
  						player17.trashItem.netDefaults(type16);
  						player17.trashItem.stack = stack7;
  						player17.trashItem.Prefix(num260);
-+						SkipVanillaRead:
++						*/
 +
  						if (num258 == Main.myPlayer && !Main.ServerSideCharacter)
  							clientPlayer.trashItem = player17.trashItem.Clone();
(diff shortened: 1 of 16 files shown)


patches/tModLoader/Terraria/MessageBuffer.cs.patch
```diff
@@ -131,26 +131,23 @@
 -					int stack7 = reader.ReadInt16();
 -					int num260 = reader.ReadByte();
 -					int type16 = reader.ReadInt16();
-+					int stack7 = ModNet.AllowVanillaClients ? reader.ReadInt16() : -1;
-+					int num260 = ModNet.AllowVanillaClients ? reader.ReadByte() : -1;
-+					int type16 = ModNet.AllowVanillaClients ? reader.ReadInt16() : -1;
++					int stack7 = -1;
++					int num260 = -1;
++					int type16 = -1;
```
Chicken-Bones, line 138: Comment out (or leave the original code commented)

patches/tModLoader/Terraria/MessageBuffer.cs.patch
```diff
@@ -373,20 +363,18 @@
 -				int stack5 = reader.ReadInt16();
 -				int prefixWeWant3 = reader.ReadByte();
 -				int type8 = reader.ReadInt16();
-+				int stack5 = ModNet.AllowVanillaClients ? reader.ReadInt16() : -1;
-+				int prefixWeWant3 = ModNet.AllowVanillaClients ? reader.ReadByte() : -1;
-+				int type8 = ModNet.AllowVanillaClients ? reader.ReadInt16() : -1;
++				int stack5 = -1;
++				int prefixWeWant3 = -1;
++				int type8 = -1;
```
Chicken-Bones, line 372: as above

Chicken-Bones (changes requested):

PR #4237, opened 2024-05-28
Correct memory reporting

JavidPack:
**TL;DR;** Memory tracking feature restored, shows full memory usage instead of just physical which was misleading.

**Final Visuals:**

https://github.com/tModLoader/tModLoader/assets/4522492/88df1403-8e97-4af8-95bc-2b7471db04c1

The memory estimates feature has recently been disabled. There was a general thought that it was not that useful anymore. It reportedly caused mods to load slower (`GC.GetTotalMemory(true)`) and wasn't reporting correct info anymore. Some claimed that since 4gb isn't the limit anymore, it wasn't necessary. Much of the support to disable it came from modders who said that modders should be using dottrace or some other memory profiler if they needed to investigate memory issues anyway, so it was useless.

Recently I came across many issue reports from users that convinced me that the feature needs to be brought back and fixed, hence this PR. Profiling is still useful for modders, but users need something.

**Changes:**
1. Memory Bar is now scaled to total memory used by tModLoader, not total installed RAM. This allows each mod's memory usage to be presented proportional to each other, rather than all being a tiny sliver of space for users with lots of installed RAM.
2. Managed memory estimates are less accurate but much faster. `GC.GetTotalMemory(true)` has been replaced with `GC.GetTotalMemory(accurate)`. `accurate` can be set by `-accuratememorytracking` command line argument if it is ever desired by a user for testing but defaults to false. Garbage collection is skipped, but accuracy isn't the goal or really that important. The goal is for users to get a rough understanding of the impact of specific mods on their limited memory.
3. Memory estimates for tModLoader now report total private memory usage ("commit size" or "virtual memory") rather than physical memory used by the process. **This is extremely important.** Mods initializing huge data structures would have their memory usage go under-reported because those allocations would happen in virtual memory (page file) instead of allocating directly into physical memory. In one case a mod was allocating 2.5 GB of extra memory for a feature, effectively preventing host and play for that mod for users with 8GB installed RAM. The `client.log` would report a very moderate RAM usage, but that was misleading.
4. Additional Logging: Log 3 most RAM consuming mods. Log virtual and physical memory as well as installed memory. Log a warning for users exceeding installed RAM and entering page file usage. Log memory usage during OutOfMemory exception. Log in-game memory usage every gigabyte milestone.
5. Memory bar is now directly toggleable.

**Showcase:**
The following videos show before and after, loading Calamity, Thorium, and MemTest (test mod that allocates 2.8 GB of arrays.). Note how the old memory bar is useless and misleading. Also note how the client.log message massively under reports RAM usage. New memory bar correctly shows the correct values and proportions. Note how the new memory bar total matches the commit size reported in task manager as well. 

**Before:**

https://github.com/tModLoader/tModLoader/assets/4522492/30b12218-44af-45a8-87bc-0f4bbfe1e15d

Logs:
```
[tML]: RAM: tModLoader usage: 2.2 GB, All processes usage: 18.9 GB, Available: 13.1 GB, Total Installed: 32.0 GB
```

**After:**

https://github.com/tModLoader/tModLoader/assets/4522492/ef2e67e0-2a43-4038-83a6-c053dfaed7ea

Logs:
```
[tML]: Mods using the most RAM: CalamityMod 3.9 GB, MemTest 2.8 GB, ThoriumMod 326.3 MB
[tML]: RAM physical: tModLoader usage: 2.2 GB, All processes usage: 18.8 GB, Available: 13.2 GB, Total Installed: 32.0 GB
[tML]: RAM virtual: tModLoader usage: 9.3 GB, All processes usage: 26.6 GB
```

**Notes:**
Task manager (tested on Windows 10) by default will only show physical memory used by a process. This will not account for virtual memory that has been requested but not used yet. This is misleading for users trying to track RAM usage of the game, but is what they will most likely see. Eventually the mod will use that memory, causing the memory to be allocate or transfer from page file to RAM. In Task Manager->Performance tab there is a "Committed" entry, that is the total virtual memory (RAM, page file, allocated but not accessed). If Committed caps out, then the computer might crash.

image

In Task Manager->Details the `Commit Size` column can be enabled, that shows the total virtual memory. In testing it was several GB larger than the normal memory metric shown by default.

image

Many users seem to have bad page file settings. Users with bad page file settings running these mods will likely run into OutOfMemory exceptions as gameplay causes more and more of the allocated virtual memory to become allocated, exceeding installed physical RAM. These errors will happen despite client.log saying there is plenty of RAM still available after mod loading.

1. Users with no page file will likely just crash
2. Users with page files with limited capacity will experience game performance issues and then eventually crash. Performance issues would probably be really bad if on a non-SSD.
3. I've been telling users in #support-forum to check their page file settings. Setting them to managed by windows is the best option if they have space on their SSD. https://www.tomshardware.com/news/how-to-manage-virtual-memory-pagefile-windows-10,36929.html

We hope that the new memory reporting will help some mods identify high RAM usage as well. Rather than have users install more RAM or fix page file settings, the best result is for mods to be fixed to not use as much RAM, since many mods are using quite a lot of RAM.


Chicken-Bones (approved):

inikishev:
nice one finally but one question, is the RAM usage different from VRAM usage? does game load all RAM into VRAM or how does it work? Because my VRAM is fully used, not RAM

JavidPack:
> nice one finally but one question, is the RAM usage different from VRAM usage? does game load all RAM into VRAM or how does it work? Because my VRAM is fully used, not RAM

Sorry, missed this. This does not report VRAM usage, that is separate, but related. The graphics api will copy over textures from RAM to VRAM when it needs to draw them, but we don't have any way of tracking that or reporting those values as far as I know.

PR #4247, opened 2024-06-03
Fix #4242

### LINKED ISSUE #4242: Crash with attempting to search mods before mod list can load.
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

Attempting to search mods in your mod list crashes the game if the game hasn't actually loaded the mod list.

### Log File

[2024-06-01-3.zip](https://github.com/user-attachments/files/15522034/2024-06-01-3.zip)

### Steps to reproduce

Get about 200 mods in your mod list (you don't have to enable them and it can be any mods) (this is only to make it take lots of time to load mods, if you have an alternate solution to make it take forever to load it should work however idk how to test that)
Boot up the game
Cancel mod loading and spam any letter/number on your keyboard
Results

### Expected Behavior

The game refuses to let me type into the search bar before mod list loads OR the game can successfully search mods without crashing even if the mod list isn't loaded yet.

### Actual Behavior

The game crashes

### Reproduction frequency

100% assuming I don't mess up the timing parts of the steps to reproduce.

### Additional Information

_No response_

Antirhinnum:
### What is the bug?
#4242

### How did you fix the bug?
Add a null check to the offending line.

There's almost certainly an underlying reason for why this bug only starting to happen now (or at least went unreported this long), but I don't watch tModLoader development closely enough to have an idea what it could be, and `UIModItem` hasn't been changed recently.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
@@ -398,7 +398,7 @@ internal class UIModItem : UIPanel
 
 		// TODO: These should just be UITexts
 		if (_mod.properties.side != ModSide.Server && (_mod.Enabled != _loaded || _configChangesRequireReload)) {
-			drawPos += new Vector2(_uiModStateText.Width.Pixels + left2ndLine, 0f);
+			drawPos += new Vector2((_uiModStateText?.Width.Pixels ?? 0) + left2ndLine, 0f);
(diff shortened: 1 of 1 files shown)


JavidPack:
Thanks for looking into this and making this pull request. I took a look at the underlying issues and made a broader fix in https://github.com/tModLoader/tModLoader/commit/5532e0bb17724393550a3fe05efb429d6677908f. I think fixing just this line would have just caused an error on later lines.

PR #4249, opened 2024-06-03
Add ProjectileID.Sets.FiresFewerFromDaedalusStormbow

direwolf420:
### What is the new feature?
`ProjectileID.Sets.FiresFewerFromDaedalusStormbow` which contains "overpowered" arrows so the [Daedalus Stormbow](https://terraria.wiki.gg/wiki/Daedalus_Stormbow) fires 1 fewer projectile occasionally (instead of 1 more), as per its spec: `Each shot fires three arrows with an additional 1/3 (33.33%) chance to spawn one extra arrow (one fewer arrow while using Holy, Unholy, Hellfire, or Jester's Arrows)`

Requires an IL edit otherwise, helps balancing stronger arrows with the Daedalus Stormbow.

No

See ExampleArrowProjectile

### ExampleMod updates
See ExampleArrowProjectile

### DIFF
--- a/patches/tModLoader/Terraria/ID/ProjectileID.TML.cs
+++ b/patches/tModLoader/Terraria/ID/ProjectileID.TML.cs
@@ -9,6 +9,12 @@ partial class ProjectileID
 		/// </summary>
 		public static bool[] SingleGrappleHook = Factory.CreateBoolSet(false, 13, 315, 230, 231, 232, 233, 234, 235, 331, 753, 865, 935);
 
+		// Values taken from Player.ItemCheck_Shoot type == 3029 check
+		/// <summary>
+		/// Set of Projectile IDs that will fire 1 less from <see cref="ItemID.DaedalusStormbow"/>
+		/// </summary>
+		public static bool[] FiresLessFromDaedalusStormbow = Factory.CreateBoolSet(false, 91, 4, 5, 41);
+
 		/// <summary>
 		/// Used to scale down summon tag damage for fast hitting minions and sentries. 
 		/// </summary>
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/ID/ProjectileID.TML.cs
```diff
@@ -9,6 +9,12 @@ partial class Sets
 		/// </summary>
 		public static bool[] SingleGrappleHook = Factory.CreateBoolSet(false, 13, 315, 230, 231, 232, 233, 234, 235, 331, 753, 865, 935);
 
+		// Values taken from Player.ItemCheck_Shoot type == 3029 check
+		/// <summary>
+		/// Set of Projectile IDs that will fire 1 less from <see cref="ItemID.DaedalusStormbow"/>
```
Chicken-Bones: Documentation error. The code has a 1/3 chance to fire 2 arrows instead of 3, per shot, and regular arrows have a 1/3 chance to fire 4 arrows instead of 3. So the average arrows/shot for this set is 8/3 instead of 10/3 for a relative difference of 20% fewer arrows.

direwolf420: I was aware of the mechanics, but felt like exactly describing the effect would be too verbose and prone to outdated-ness if vanilla changes anything about it. How about something more vague: `Set of Projectile IDs that, when fired from Daedalus Stormbow, will have the amount reduced`. Or would just removing the `1` be enough? Same with the comment in ExampleMod

Chicken-Bones: Yeah, `less` or `20% less` would suffice.

Chicken-Bones: Documentation can always be updated as it becomes inaccurate with 1.4.5 (though at this time, there are no changes to this mechanic)

JavidPack:
I think "fewer" is more correct than "less" here, but I'll have to check.

PR #4256, opened 2024-06-07
Fix documentation references for nuget packages

JavidPack:
When modding, there are a few libraries with documentation that would be useful to have automatically show up. Currently they do not because they are in a different location from the .dll file:

```
tModLoader\Libraries\Newtonsoft.Json\13.0.3\Newtonsoft.Json.xml
tModLoader\Libraries\Newtonsoft.Json\13.0.3\lib\net6.0\Newtonsoft.Json.dll
```

This PR fixes that. I'm not sure why we need that much nesting in the first place, but I seem to remember it causing issues during the initial dotnet core migration.

This PR does not fix `MonoMod.RuntimeDetour.xml`, which was the main motivator for this fix in the first place. For some reason the .xml file is not copied anywhere despite being in the `.nupkg`. Updating to `25.1.1` from `25.1.0` fixes this somehow, but I'm not sure what other changes `25.1.1` might bring.

### DIFF
--- a/tModBuildTasks/OrganizeReferenceDestinations.cs
+++ b/tModBuildTasks/OrganizeReferenceDestinations.cs
@@ -37,6 +37,7 @@ public sealed class OrganizeReferenceDestinations : TaskBase
 			string pathInPackage = item.GetMetadata("PathInPackage");
 			//string fileName = item.GetMetadata("Filename");
 			//string runtimeIdentifier = item.GetMetadata("RuntimeIdentifier");
+			string dllDirectoryInPackage = null;
 
 			// PDBs & XMLs lack some metadata, attempt to get it from the paired .dll.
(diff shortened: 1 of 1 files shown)


JavidPack:
Related to https://github.com/tModLoader/tModLoader/tree/fix/buildtasks_dlls

Mirsario (approved):
I fixed an additional set of packages that had XML files in their /ref/ directories, but the MonoMod XML file is nowhere to be seen in the entire MSBuild log file, so it's unlikely to be our fault that the file isn't there. I doubt that updating the library would introduce any issues, they seem to be following semantic versioning nowadays.
image

Run some sanity checks & merge?

Mirsario:
Confirmed, MM 25.1.0 just didn't have functioning docs all on its own. This is an empty non-TML project:
image
image

JavidPack:
There are a total of 114 xml and pdb files that are moved to reside next to their dll files due to this change. Monomod doc fixes will happen in a separate PR for safety. 

Sample of changes:

image

Issue #4258, opened 2024-06-08
Mod Configs will be lost forever once you join a server with a different configuration.


tieeeeen1994:
### Do you intend to personally contribute/program this feature?

~~No~~ Yes. #4262.

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

> How would I go on about sharing a script that helps a player manage modpacks with mod configs included? It's just a script, but I think this needs attention because once you join a server with a different config, your current configuration will be lost forever and overwritten even if you saved your setup as a mod pack.

I said this in the tML Discord Server. I am a player who hosts and plays multiple tModLoader servers and games respectively. As a player, it does come to my attention that Mod Packs did not actually save configuration, *unless exported as an instance*. Exporting as an Instance, however, has no way to *automagically* update your current configuration the same as that in the instance. Only manually copy pasting the files after exiting tModLoader is the way to do it.

I made a batch script instead to mitigate this problem.

```bat
@echo off
setlocal

set "name=%~1"
set "source=C:\Program Files (x86)\Steam\steamapps\common\tModLoader\%name%\SaveData"

if not exist "%source%" (
    echo Error: Mod Pack %name% does not exist.
    exit /b 1
)

set "destination=%USERPROFILE%\Documents\My Games\Terraria\tModLoader"

xcopy /s /e /y "%source%\ModConfigs" "%destination%\ModConfigs" >nul 2>&1
copy /y "%source%\Mods\enabled.json" "%destination%\Mods\enabled.json" >nul 2>&1

start steam://rungameid/1281930

endlocal
```

Current workflow with the script:
1. Export mod packs as instances.
2. Run script with a command-line argument of the folder name of the mod pack.
3. Enjoy your current configuration.

So back to the issue, I have lost my mod configurations in like... 10 mod packs because I initially did not realize this. I play single player with my desired configuration, then I join a game with my friends, then I realize my singleplayer configs are gone *forever*, hence my workaround above.

My proposal is to also copy mod configs into the mod pack. The folder is already there, but nothing is copied. Only exporting as an instance copies the mod configurations but exporting as an instance serves another purpose. Allow players that when they enable the mod pack list, mod configs are also synced with the mod pack to the current configuration.

### What does this proposal attempt to solve or improve?

This will actually enhance player's experiences when they attempt to create a balance around mod interactions. For example, if Calamity mod is installed, I would disable some QoL features in some mods because of the balance changes in the mode.

[...]

Issue #4228, opened 2024-05-23
Modpack's ModConfigs do not apply when importing locally.


AidenMDA:
### Version

Latest 1.4.4-stable

### OS

Linux

### Platform

Steam

### This bug affects...

Gameplay as a Player

#### Context
I've enabled a lot of mods, each had been configured to my liking.
I've made a new modpack by clicking 'Save Enabled as New Mod Pack' in the 'Mod Packs' sub-menu.

#### Bug
The resulting modpack contains frozen versions of each mod I wanted but it did not copy the configs into the ModConfigs folder.
Even when I manually copy the config files into the modpack's ModConfigs folder, and then import said pack, the configs do not override whatever configs are in my normal ModConfigs folder.
I tried deleting every mod I had installed, including every single config file (leaving the modpack untouched). I then import the pack, mods load but the configs still don't, leaving me with default configurations for every mod.

#### My thoughts
This may be because it's a forgotten feature, from memory it was a thing in pre-Steam tModLoader.
For now I've had to resort to my bootleg scripts to swap configs semi-automatically, but I wanted to use this modpack to play with friends who might not bother, or know how to do a sketchy work-around for something that's meant to be convenient.
If possible, it would be awesome to have this fixed for 1.4.4 as well as 1.4.5 so the current modpack I can easily use the modpack I have set up at the moment.

### Steps to reproduce

1. Load mods.
2. Apply config changes to said mods.
3. From the main menu, go Workshop -> Mod Packs, then click 'Save Enabled as New Mod Pack'.
4. Once you've chosen a name and created the modpack, click 'Open Mod Pack Folder.'
5. Look in the modpack folder, in it, you should see a 'Mods' and 'ModConfigs' folder.
6. See how the 'ModConfigs' folder is empty but the 'Mods' folder is containing the expected mods.
7. Go back until you are in the 'tModLoader' folder, then go into 'ModConfigs'.
8. Move all config files out of it, move them wherever else you like, even in the modpack if you want.
9. Back in-game: Click 'Import Pack (Local)', this should disable all mods you have in your regular mods folder and enable the mod files in the modpack.
10. Modpack has been applied.
11. Check mod configs.

### Expected Behavior

When clicking 'Save Enabled as New Mod Pack', the config files for the enabled mods should also be copied into the pack, alongside the relevant mod files.
When clicking 'Import Pack (Local)', the game should apply whatever config files are stored in the modpack.
[...]

PR #4262, opened 2024-06-10
Enhance Mod pack Capabilities, and fix bugs.

tieeeeen1994:
### What is the bug?

The bug is about mod packs not syncing mod configurations. I've also found some issues when updating the mod pack with enabled. The files aren't really updating, copying and deleting properly, making mod packs very confusing to work with.

#### Related issues

- #4228 - Directly addresses the issue.
- #4258 - Mod configuration loss can be mitigated if mod pack feature can synchronize mod configurations.

### How did you fix the bug?

I uncommented an existing line that copies mod configurations in their proper folder. (If this was intended to be commented temporarily, please explain to me so I have more context.)

I also added code that copies mod config files into the real mod configs folder.

~~I noticed that updating the mod pack with Enabled button inconsistently copies the `.tmod` files into the mod pack folder. Sometimes, it doesn't work. Spamming the button sometimes makes the copying successful. What is the real intention of the button? Should it only copy enabled.json? Or should it also copy `.tmod` files?~~

I also included a fix where Update Mod pack with Enabled does not correctly update the mod pack files. Now, they copy `.tmod` files correctly alongside the mod configs.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/UI/UIModPackItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/UI/UIModPackItem.cs
@@ -366,6 +366,13 @@ internal class UIModPackItem : UIPanel
 			Interface.infoMessage.Show(Language.GetTextValue("tModLoader.ModPackModsMissing", string.Join("\n", modListItem._missing)), Interface.modPacksMenuID);
 		}
 
+		string configsPath = Path.Combine(modListItem._filepath, "ModConfigs");
+		var configFilesFromPack = Directory.EnumerateFiles(configsPath);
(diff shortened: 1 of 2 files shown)


tieeeeen1994:
I hope I can receive inputs from questions raised.

tieeeeen1994:
Merge request marked as ready. Please suggest improvements if there are. I'm not very familiar with tModLoader's code base.

tieeeeen1994:
> That will cause players to lose their default configs. I think the intended design was for ConfigManager.Save/Load to redirect to the ModPack config folder for loading and saving when a modpack is enabled.

PR #4263, opened 2024-06-11
Add PreDrawWings hook & CustomDrawnWings example

ghost:
### What is the new feature?

Adds a `PreDrawWings` hook to `ModItem` that makes it easier to custom draw wings and disable vanilla wing drawing.

This previously required a custom `PlayerDrawLayer`. This should make it easier to draw  wings with a custom amount of frames or different texture sizes. More complex use-cases can still use a custom layer.

Not for this this specifically. Alternatives like introducing more drawing hooks on `EquipTexture` have been suggested, but this is low complexity and in line with other such drawing hooks.

See `ExampleCustomDrawnWings`.

### ExampleMod updates
Added a `ExampleCustomDrawnWings` example.


ghost:
Weird line ending change fixed and multiplayer tested

JavidPack:
Closing in favor of #4264

PR #4264, opened 2024-06-13
ModifyEquipTextureDraw and ModifyItemDraw hooks

JavidPack:
# ModifyEquipTextureDraw  hooks

`EquipTexture`s have always been hard to customize. If a modder wanted to customize how the equip texture of an accessory draws, the current solution is to make a new `PlayerDrawLayer` parented to the vanilla draw layer and do custom drawing there. This requires reimplementing all the math to construct the `DrawData` from all the `PlayerDrawSet` parameters. The autoloaded `EquipTexture` would also likely have to be an empty texture as well to not conflict with the custom layer.

This PR adds hooks, `EquipTexture.ModifyDraw` and `ModItem.ModifyEquipTextureDraw`, allowing each `EquipTexture` (and corresponding `ModItem`) to customize their own appearance, either by tweaking the vanilla `DrawData` or adding new `DrawData`. Allowing a `EquipTexture`/`ModItem` to customize their own appearance helps keep code organized as well.

Some simple examples of what these new hooks can be used to do: Adjust draw location, implement a glow mask, implement a pulse effect, modify draw color.

This was motivated by custom wing animation, something that has been requested frequently. There is a current PR, #4263, that implements this, but this is a more general solution up for consideration.

### ExampleMod updates

The PR showcases how easy or powerful the hook can be. `ExampleCustomDrawWings` has been added and shows implementing a 7 frame wing animation similar to Grox The Great's Wings. 

https://github.com/user-attachments/assets/07498d10-0e3e-4715-bc1b-08ae03628608

### Sample usage
Here is a very simple example showing adjusting the draw position of equip texture (10 pixels to the right, 20 pixels down). This can be useful if the normal drawing bounds for a particular equip texture are not big enough for what you want to draw. Similar code can be done to adjust other fields of `DrawData` like the source rectangle, color, etc.

```cs
public override bool ModifyEquipTextureDraw(ref PlayerDrawSet drawInfo, ref DrawData drawData, EquipTexture equipTexture, string methodName) {
	drawData.position += new Microsoft.Xna.Framework.Vector2(10, 20) * drawInfo.drawPlayer.Directions;
	return true;
}
```

# ModifyItemDraw hooks
`ModifyItemDraw` hooks have also been added (`ModItem.ModifyItemDraw`, `GlobalItem.PreModifyItemDraw`, and `GlobalItem.PostModifyItemDraw`). These allow modifying how a held item (and in-use) is drawn. 

This has been a highly requested feature, especially since `(Pre|Post)DrawInInventory/(Pre|Post)DrawInWorld` have existed but there was no way to modify the held item rendering without disabling the vanilla `HeldItem` `PlayerDrawLayer` and supplying your own `PlayerDrawLayer`. This is highly inconvenient due to the logistics and the tedious code repetition needed to implement.

Now, modders can directly modify the `DrawData` created in the vanilla `HeldItem` `PlayerDrawLayer`. 

### ExampleMod updates

`CustomItemDrawingShowcase` has been updated to demonstrate the existing visual effects examples now while being held.

https://github.com/user-attachments/assets/63ad37d1-0f87-466d-a8a0-147bd8429144

### Sample usage

Here is a simple example. This sword uses `ModifyItemDraw` to draw a chaotic pulse effect. The passed in `drawData` is cloned and modified using the `with` keyword, then added to `drawInfo.DrawDataCache`, resulting it the cloned `DrawData` being drawn before the normal item drawing.

```cs
public override bool ModifyItemDraw(ref PlayerDrawSet drawInfo, ref DrawData drawData, ref DrawData? coloredDrawData, ref DrawData? glowmaskDrawData) {
	for (int i = 0; i < 3; i++) {
		var color = Color.Lerp(Color.Aqua, Color.Red, (float)i / 2);
		drawInfo.DrawDataCache.Add(drawData with { position = drawData.position + Main.rand.NextVector2Circular(6, 6), color = color });
	}
	return true;
}
```
This video shows this code in action during the first few seconds. There is a shadow effect showcased as well, which is also something that is now easily possible using these new hooks.

https://github.com/user-attachments/assets/956575ea-5150-43b2-a1b6-dd5646ffdaab

<details><summary>Shadow effect example code</summary>

```cs
public override bool ModifyItemDraw(ref PlayerDrawSet drawInfo, ref DrawData drawData, ref DrawData? coloredDrawData, ref DrawData? glowmaskDrawData) {
	Player player = drawInfo.drawPlayer;
	int totalShadows = Math.Min(player.availableAdvancedShadowsCount, 6);
	int skip = 1;
	for (int i = totalShadows - totalShadows % skip; i > 0; i -= skip) {
		EntityShadowInfo advancedShadow = player.GetAdvancedShadow(i);
		float shadow = Utils.Remap((float)i / totalShadows, 0, 1, .6f, .1f, clamped: true);
		var adjust = advancedShadow.Position- player.position;

		drawInfo.DrawDataCache.Add(drawData with { position = drawData.position + adjust, color = drawData.color * shadow });
	}

	return true;
}
```    

</details>


direwolf420:
In favor of this PR.

An additional common thing modders want: glowmasks. `Body` supports glowmasks (with some extra code in ModPlayer.ModifyDrawInfo) but the rest doesn't, an easy way to have the regular draw code run and then add a glowmask ontop would be nice to have.

patches/tModLoader/Terraria/DataStructures/PlayerDrawLayers.cs.patch
```diff
@@ -65,7 +65,36 @@
...
++			drawinfo.DrawDataCache.Add(item);
+ 		if (drawinfo.drawPlayer.wings == 43 && drawinfo.shadow == 0f) {
+ 			Vector2 vector13 = vector12;
+ 			Vector2 origin = new Vector2(TextureAssets.Wings[drawinfo.drawPlayer.wings].Width() / 2, TextureAssets.Wings[drawinfo.drawPlayer.wings].Height() / num13 / 2);
+@@ -752,7 +_,7 @@
+ 			for (int l = 0; l < 2; l++) {
+ 				item = new DrawData(position: vector13 + new Vector2((float)Main.rand.Next(-10, 10) * 0.125f, (float)Main.rand.Next(-10, 10) * 0.125f), texture: TextureAssets.GlowMask[272].Value, sourceRect: value5, color: new Color(230, 230, 230, 60), rotation: drawinfo.drawPlayer.bodyRotation, origin: origin, scale: 1f, effect: drawinfo.playerEffect);
+ 				item.shader = drawinfo.cWings;
+-				drawinfo.DrawDataCache.Add(item);
++				// drawinfo.DrawDataCache.Add(item);
```
Chicken-Bones: hm?

PR #4268, opened 2024-06-23
Fix Dockerfile permissions

Reboot-Codes:
Fixing the default Dockerfile.

### What is the bug?

Attempting to run the default dockerfile in `tModLoader/tModLoader/Terraria/release_extras/DedicatedServerUtils/Dockerfile` is not possible due to an attempt to write into `/home/tml/server/tModLoader-Logs/Natives.log` as noticed by @reniivali in the following screenshot:

image

This causes the server to error and exit immediately.

This problem is most likely caused by setting the home directory path in the `adduser` command with `-h` explicitly (see [the alpine linux wiki](https://wiki.alpinelinux.org/wiki/Setting_up_a_new_user) and the [`adduser`](https://linux.die.net/man/8/adduser) manpage; specifically in the manpage (when setting the homedir path directly): "The directory HOME_DIR does not have to exist but will not be created if it is missing.").

By default, this means that `/home/tml` is owned by `root:root`, with `755`, `750`, or `700` permissions. `tml` cannot write into, or perhaps even read `/home/tml` and/or its contents.

### How did you fix the bug?

The order of operations within the Dockerfile was changed from first switching to and setting the `tml` user's environment variables, then using said variables to create `/home/tml`; to creating `/home/tml` first, and setting the ownership of `/home/tml` to `tml:tml` and setting the permission octets to `755` (`rwxr-x--x`), then proceeding with environment variable configuration. The remaining portions of the file are unchanged.

Yes, it would be possible to just omit `-h /home/tml` entirely, and let the alpine system defaults take over home directory creation, but I assume that it was specified directly to ensure reproducibility. My fix keeps most of the existing code, but will ensure `tml` is able to write to `/home/tml` and its subdirectories if any were created when `/home/tml` was first made (however, there shouldn't be any as it's created by `WORKDIR`, not `adduser`).

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/Dockerfile
+++ b/patches/tModLoader/Terraria/release_extras/DedicatedServerUtils/Dockerfile
@@ -20,10 +20,14 @@ ARG GID
 RUN addgroup -g $GID tml \
  && adduser tml -u $UID -G tml -h /home/tml -D
 
+# Fix permissions for tml's home directory due to the creation of `/home/tml/server/tModLoader-Logs/Natives.log`
+WORKDIR /home/tml
+RUN chown -R tml:tml /home/tml
+RUN chmod 755 /home/tml
+
 USER tml
 ENV USER tml
 ENV HOME /home/tml
-WORKDIR $HOME
 
 # Update SteamCMD and verify latest version
 RUN steamcmd +quit


ppebb:
Can you post reproduction steps for encountering the bug this covers? This issue hasn't shown up since the lines you point out were committed two years ago so I'd like to know what caused you to hit it.

Also would you be able to test with just changing -h to --home? That seems to be the correct switch and I'd prefer to use that behavior instead.

Reboot-Codes:
To be quite frank, this was a fix to an issue presented by a friend of ours. I don't have the exact `docker` command or `docker-compose.yml` to present for reproducibility. Currently waiting on the user mentioned in the pr comment for an example.

I do think this is an issue despite not having a reproducible case, as, usage of home directory specification here is mildly redundant and will lead to issues relating to: assuming all users in the container are able to write to dockerfile-created directories, or permission inheritance from the host FS being configured properly, if inherited at all (e.g. via a mount).

However, ensuring this issue is reproducible and not a user-misconfiguration is good, and I'll be happy to run testing myself if required.

ppebb:
> ... assuming all users in the container are able to write to dockerfile-created directories, or permission inheritance from the host FS being configured properly, if inherited at all (e.g. via a mount).

This is a good point that I did not think of.

I'd still like to know the reproduction steps first, but I'll go ahead and test this (sometime soon...) and if it works I think this is fine to be merged.

turtletowerz:
> To be quite frank, this was a fix to an issue presented by a friend of ours. I don't have the exact `docker` command or `docker-compose.yml` to present for reproducibility. Currently waiting on the user mentioned in the pr comment for an example.

If they have a Discord then joining the TML server and specifically [this thread](https://discord.com/channels/103110554649894912/1118434145786417182) would be the best way to collaborate on this. If you prefer to keep it on Github for clarity that is also fine.

> I do think this is an issue despite not having a reproducible case, as, usage of home directory specification here is mildly redundant and will lead to issues relating to: assuming all users in the container are able to write to dockerfile-created directories, or permission inheritance from the host FS being configured properly, if inherited at all (e.g. via a mount).

I agree that the initial command is wrong but since I had no issues in testing I never thought to look into it further. If we can confirm that this is indeed an issue that is reasonably reproducible then it may be best to introduce a breaking change to fix it.

> However, ensuring this issue is reproducible and not a user-misconfiguration is good, and I'll be happy to run testing myself if required.

Again I had not seen this issue in my initial testing and even within the past month I can say I was unable to reproduce this. Could you share what UID and GID are set for the container as well as what it is set to both inside and outside the container? To debug within the container you can uncomment the `entrypoint` line in the compose file. The UID and GID of the `tml` user in the container should be set to whatever you provide. The fact that the logs file has no permissions makes it seem like a permissions issue with the mounted volume.


Taking a quick look inside the container, I see `tml` as the owner of the `/home/tml` directory and all files in it, so i cannot reproduce this issue.
image

PR #4271, opened 2024-06-24
`X` and `Y` properties for `Tile`

BasicallyIAmFox:
### What is the new feature?
Gives `Tile` struct `X` and `Y` properties.

Allow to trace back the X/Y coordinate for a `Tile` without knowing the initial coordinate, could be useful for stuff where we have the tile "instance" but not the coordinate associated to it and need to coordinate to find what's around it for example.

1. Enabling `TILE_X_Y`.
3. Making `X` and `Y` methods instead of properties.
2. Giving 1 property (or method) to get x and y. (`XY`)

```cs
// oh no! we're only given a Tile instance!
// whatever would we do if this PR didn't pop up to save the day?!
// ... i would answer but i'd like to remain safe.
public static void Foo(Tile tile) {
  int x = tile.X;
  int y = tile.Y;
  
  // rest of code
}
```


Chicken-Bones:
This was not enabled because it reduces performance. In the vast majority of code, `x` and `y` are present as integers in the surrounding context anyway. The `readonly` are also less ergonomic than separate variables. You can't just use `x++`

It may be worth considering after re-running the benchmarks with .net8, but the motivation for this change is not strong
> **could** be useful

direwolf420:
> could be useful

The main use I see is when you are being given only a `Tile` object but need its coordinates. This happens regularly when detouring vanilla methods, often they don't pass in the coordinates, so as a modder you are forced to detour or IL edit somewhere higher up the call chain to cache the coordinates where possible

Chicken-Bones:
> > could be useful
> 
> The main use I see is when you are being given only a `Tile` object but need its coordinates. This happens regularly when detouring vanilla methods, often they don't pass in the coordinates, so as a modder you are forced to detour or IL edit somewhere higher up the call chain to cache the coordinates where possible

Yeah, I'm sympathetic to the use case, but it is quite limited in scope. If it was zero-cost I would approve it.

BasicallyIAmFox:
> > > could be useful
> > 
> > 
> > The main use I see is when you are being given only a `Tile` object but need its coordinates. This happens regularly when detouring vanilla methods, often they don't pass in the coordinates, so as a modder you are forced to detour or IL edit somewhere higher up the call chain to cache the coordinates where possible
> 
> Yeah, I'm sympathetic to the use case, but it is quite limited in scope. If it was zero-cost I would approve it.

we wish but we can't due to the tile id layout

PR #4275, opened 2024-06-28
Fix several TileObjectData issues and document TileObjectData class

JavidPack:
Document TileObjectData class and much more. Modders have long been confused about how alternate placements work. The documentation strives to standardize terminology, specifically regarding tile style, alternate placements, and placement style. The [Basic Tile](https://github.com/tModLoader/tModLoader/wiki/Basic-Tile) wiki page has also been updated.

In an effort to document TileObjectData, the following fixes were applied to issues discovered:

Fix issue in `TileLoader.CheckModTile` checking incorrect alternate `TileObjectData`, causing tiles with alternate anchors to immediately be killed unless an additional alternate is registered. Update `ExampleSign` with fix.

https://github.com/tModLoader/tModLoader/assets/4522492/4ad3f4b4-a5ef-4bfc-b8aa-ad2e79806e76

Fix `ExampleDoorOpen`, it was using the vanilla open door `TileObjectData` which incorrectly uses `StyleLineSkip` and alternate placements. Consequently, it did not obey the correct anchors when opened to the left:

https://github.com/tModLoader/tModLoader/assets/4522492/79b8e5da-89bb-4d6a-8c6a-de496862f4ae

Added `TileObjectDataShowcase.cs` to showcase advanced usage of `TileObjectData` including left and right placements, random placements, animation, toggle states, custom anchors, and multiple styles. It shows how each of these affect the layout of the expected tile spritesheet. ([Tile Wiki page](https://github.com/tModLoader/tModLoader/wiki/Basic-Tile#stylemultiplier) updated with annotated tile spritesheet as well)

https://github.com/tModLoader/tModLoader/assets/4522492/e5bb3e43-78d4-4049-bc65-1b1b625a98c4

Fix some bottom anchor checks incorrectly being applied to tiles that did not have the specific `AnchorType`s being checked in their bottom anchor.

Fixed `TileObjectData.GetTileStyle`/`GetTileInfo` incorrectly ignoring `StyleMultiplier` when used with `StyleLineSkip`, which will allow tiles also using `StyleWrapLimit` to function correctly.

Fix random placement cache not working with alternates with different origins (mistakenly cycles new random placement style each update)

https://github.com/tModLoader/tModLoader/assets/4522492/c34db67b-4657-489c-9d74-94a92ae39279

Also fix `ExampleLamp` flame flip regression.

image

image

# Porting Notes

- Modders should fix their open door tiles if they are using `TileObjectData.newTile.CopyFrom(TileObjectData.GetTileData(TileID.OpenDoor, 0));` or if their doors break when opening to the left.
- Modders should double check that tiles using placement alternates (`TileObjectData.newAlternate`/`TileObjectData.addAlternate()`) work as expected. The fixes should not cause issues, but it is possible that incorrect workarounds to issues fixed in this PR are now broken.
- Modders should use `StyleLineSkip` and `StyleMultiplier` where correct. This should reduce the amount of workarounds such as using `RegisterItemDrop` or `GetItemDrops` to fix item drops that are incorrectly being calculated.


ExampleMod/Content/Tiles/Furniture/ExampleDoorOpen.cs
```diff
@@ -35,9 +37,8 @@ public class ExampleDoorOpen : ModTile
 			AddMapEntry(new Color(200, 200, 200), Language.GetText("MapObject.Door"));
 
 			// Placement
-			// In addition to copying from the TileObjectData.Something templates, modders can copy from specific tile types. CopyFrom won't copy subtile data, so style specific properties won't be copied, such as how Obsidian doors are immune to lava.
-			TileObjectData.newTile.CopyFrom(TileObjectData.GetTileData(TileID.OpenDoor, 0));
-			/* This is what is copied from the OpenDoor tile
+			// The TileID.OpenDoor TileObjectData is incorrect, as they sometimes are, so we will not be copying from it in this case
```
Chicken-Bones: Incorrect how? This implies a bug in vanilla. If there's no vanilla bug, it may be better to explain why it doesn't apply to our tile/spritesheet.

patches/tModLoader/Terraria/TileObject.cs.patch
```diff
@@ -28,6 +36,33 @@
  						flag4 = true;
  
  					if (flag4 || flag2 || flag3) {
+@@ -373,7 +_,7 @@
+ 					Tile tileSafely = Framing.GetTileSafely(num8 + num29, num9 + num28);
+ 					bool flag6 = false;
+ 					if (tileSafely.nactive()) {
+-						if (Main.tileSolid[tileSafely.type] && !Main.tileSolidTop[tileSafely.type] && !Main.tileNoAttach[tileSafely.type] && (tileData2.FlattenAnchors || tileSafely.blockType() == 0))
++						if ((anchorBottom.type & AnchorType.SolidTile) == AnchorType.SolidTile && Main.tileSolid[tileSafely.type] && !Main.tileSolidTop[tileSafely.type] && !Main.tileNoAttach[tileSafely.type] && (tileData2.FlattenAnchors || tileSafely.blockType() == 0)) // AnchorTop
```
Chicken-Bones, line 44: Do these need backporting?

patches/tModLoader/Terraria/TileObject.cs.patch
```diff
@@ -38,3 +73,17 @@
...
+@@ -601,8 +_,13 @@
+ 				Point16 objectStart = TileObjectPreviewData.randomCache.ObjectStart;
+ 				int num36 = coordinates.X + objectStart.X;
+ 				int num37 = coordinates.Y + objectStart.Y;
++				int num38 = x - tileObjectData.Origin.X;
++				int num39 = y - tileObjectData.Origin.Y;
++				/* Fix random cache not working with alternates with different origins (mistakenly cycles new random each update), it should use the alternates TOD, not the tile styles TOD 
+ 				int num38 = x - tileData.Origin.X;
+ 				int num39 = y - tileData.Origin.Y;
++				*/
```
Chicken-Bones, line 85: Vanilla backport?

Chicken-Bones (commented):
Looks good. Didn't have time to read all the comments in detail.

patches/tModLoader/Terraria/TileObject.cs.patch
JavidPack: This could be backported. It doesn't technically need to since all vanilla bottom anchors are solid tile anchors from what I remember, so it doesn't affect anything.

patches/tModLoader/Terraria/TileObject.cs.patch
JavidPack: This also can be backported. It doesn't affect vanilla at all currently since vanilla never uses a TileObjectData with RandomStyleRange and alternates with different origins, but I don't see any reason not to.

ExampleMod/Content/Tiles/Furniture/ExampleDoorOpen.cs
JavidPack: I updated this to be a bit clearer.

These aren't "technically" vanilla bugs because vanilla doesn't use anchors after placement and you can't place open doors. The layout of DoorOpen tile is also incorrect because the alternates are supposed to be in-line with styles. This was likely changed in an effort to fix doors into the 2048 pixel limit in vanilla, but the fix does not fit with the default TileObjectData behaviors we use in tModLoader. This too, however, doesn't count as a bug because the vanilla framing checks are all hardcoded as well.

I don't think it helps to explain too deeply why this and ExampleSign aren't using the vanilla TOD, as the whole situation can be confusing to a modder, but I think the new comments help explain it.

PR #4287, opened 2024-07-09
Fix world corruption when changing frameImportant for modded tiles

Chicken-Bones:
When a modded tile is `frameImportant` then we save the `frameX` and `frameY`. We also save the value of `frameImportant` in an entry for that tile in the file. 

A bug in the tile loading code was discarding the saved value of `frameImportant` and using the current value for the mod. This causes an over/underread error, offsetting or corrupting the remainder of modded tile data.

The fix is to use the saved entry for deserialization, and just update the `loadedType` field to the current runtime `Type` of the tile


ryanpolasky:
+1 to this issue, would love to see this merged 🫶

ryanpolasky (approved):
Look good to me, good job on catching this issue so quickly!

PR #4288, opened 2024-07-09
Cleanup some code in TileIO

Chicken-Bones:
This should contain no functional changes, just a follow up to keep the code maintainable after #4287 

Note that we can't test `canPurgeOldData` yet, as there is no UI/config/debug setting to trigger it.


JavidPack:
I did some testing to ensure these changes didn't break anything, this setup of tiles properly went from loaded to unloaded back to loaded.

dotnet_2024-07-30_11-53-16
dotnet_2024-07-30_11-55-49

PR #4291, opened 2024-07-10
Implement workshop tag filtering in Mod Browser, Fix #245

### LINKED ISSUE #245: Suggestion: Mod categories
I noticed that there are a lot of mods on tmodloader, and it s hard to find the type of mod one is looking for just by searching in a bar, and scrolling throught all the mods scrolls way too fast, so it would be nice if perhaps mods could be part of categories to help people find what they are looking for.

some categories i would do are: Content (thorium, calamity, tremor), Utility (Recipe browser, Boss checklist, Cheat Sheet), Small additions (Morepotions, NPC vending machines), RPG/ Levels (learn by doing, experience and classes). 

there would be a thing at the top next to the search bar that let you choose the categories you want to search in.
this would make searching for mods easier, and if the scroll bar scrolled at a reasonable pace when viewing a lot of mods then that would make it even better.

JavidPack:
Workshop tag filtering has been implemented into the in-game mod browser. Users can select tags to filter by. Categories and languages are listed in 2 columns. Users can only select a single language at a time and can select multiple categories. 

Discussion: It is possible to auto-select the users current language, but I decided against that since I don't think that behavior is necessarily what all non-English speakers would prefer. 

https://github.com/tModLoader/tModLoader/assets/4522492/aa3dbf36-a5bc-4bcc-916b-901a3be15b88

There might be a more straightforward approach to implementing a dropdown menu UI, but the approach in this code matches the bestiary behavior.

The columns adjust their width for the launch language:

image

Edit: Updated Clear Tags button color
image


patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.Elements.cs
```diff
@@ -189,16 +192,37 @@ public override void OnInitialize()
...
+		TagFilterToggle.OnLeftClick += OpenOrCloseTagFilterDropdown;
+		TagFilterToggle.OnLeftClick += (a, b) => RefreshTagFilterState(); // Undo the automatic state cycle rather than modify existing public UIElement class.
+		TagFilterToggle.OnRightClick += (a, b) => RefreshTagFilterState();
+
 		SearchFilterToggle.SetCurrentState(SearchFilter.Name);
 		TimePeriodToggle.SetCurrentState(ModBrowserTimePeriod.OneWeek);
 		UpdateFilterToggle.SetCurrentState(UpdateFilter.All);
 		ModSideFilterToggle.SetCurrentState(ModSideFilter.All);
 		SortModeFilterToggle.SetCurrentState(ModBrowserSortMode.Hot);
+		ResetTagFilters();
```
Solxanich: This code is duplicated in `Reset()` 
Any chance we can refactor while doing this to eliminate that?

Solxanich (approved):
Code looks good, 
No issues encountered when testing it out. Easy to use, design makes sense.

Just noticed a minor code duplication which could clean up while in here.

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.Elements.cs
JavidPack: Done

Issue #4305, opened 2024-07-18
Add extra hooks in WorldGen.TileFrame() for ease of custom framing without needing to entirely copy huge chunks of vanilla code


IbanPlay:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

- Add a **ModTile.PostTileFrame()** hook that runs after the vanilla framing code has decided on Tile.TileFrameX/TileFrameY to let modders apply simple edits (Like for example, offsetting the frame's position for larger patterns). It would need to be called both for FrameImportant tiles and non FrameImportant ones

- Add a **ModTile.ModifyFrameMerge()** that runs after the adjacent tile's merge type has been calculated, but before the frame gets calculated, exposing the adjacent tile's merge type as references so modders can modify them as needed

### What does this proposal attempt to solve or improve?

Currently the only way to have custom tile framing is by using the **ModTile.TileFrame()** hook, which is overkill in a lot of cases. It preventing the ENTIRE vanilla tile framing code means that even for simple changes, modders are forced to copy huge chunks of the vanilla tile framing code

- A common use for this would be tiles with large patterns like stone slabs. The current most common way to do this with modded tiles is to override **ModTile.AnimateIndividualTile()** to add an offset to the tile frame on the sheet which runs every single frame. This means that patterns with more complex rules can cause hits in performance due to needing to recalculate the frame offset on every draw loop
![image](https://github.com/user-attachments/assets/8cfc23eb-0884-491b-be29-624aeddb5ac7)
> Example code for a tile that uses AnimateIndividualTile to create a pattern that spans 2x2 tiles

- Another pain point is wanting to modify the connections of a tile with more control. As an example, I'd like to be able to make it so a tile only connects to another tile from the bottom, but considers it as an empty space if its on top, and have that modification happen before vanilla code that decides the frame runs. Another big thing having control over the connections would bring is the ability to easily make tiles that merge with tiles that aren't dirt by using a combination of **TileID.Sets.ChecksForMerge** and **WorldGen.TileMergeAttempt**. Currently, once again, making tiles that merge with tiles that aren't dirt is impossible unless the entire TileFrame code is copied

![placeholder](https://github.com/user-attachments/assets/9eb601b2-30da-4b43-989c-50a4a55822eb)
[...]

Issue #3878, opened 2023-10-12
TerrariaHooks IL Hooks can get garbage collected


Mirsario:
### Version

All - latest 1.4.4-stable, 1.4.4-preview, dev branch too.

### Requestors / This bug affects

- Gameplay as a Player
- Mod capability as a Modder
- tModLoader code as a Contributor/Maintainer

### Description

Initially reported by @jjohnsnaill months ago, but persistently reproduced only now. 

It seems like injections done via `IL_X` classes can get collected on the next serious garbage collection, even though they're supposed to be put into a `ConcurrentDictionary` inside `MonoMod.RuntimeDetour.HookGen.HookEndpointManager`.

Our MonoMod is up to date, and regenerating TerrariaHooks hasn't changed anything significant.

### Reproduction

Setup:
1. Have a Release (optimized) TML build.
2. Download 
[IssueReproduction_InjectionGC.zip](https://github.com/tModLoader/tModLoader/files/12884470/IssueReproduction_InjectionGC.zip)
3. Install the packaged mod and world, or compile the src.
4. Close the game.

Actual:
1. Start the game having the mod already enabled. Do not reload mods.
2. Go straight into the world titled `1`.
3. The world has an ongoing Frost Legion invasion. If the mod's injection hasn't yet been unloaded it makes it so that only 3 of its enemies (snowmen) are able to spawn at once.
4. Run the `/gc` command, which forces a full garbage collection. This breaks the injection, making a ton of enemies appear like in vanilla.

### Expected Behavior

IL injections persist until mods are reloaded.

### Actual Behavior

IL injections may or may not get unloaded during a serious garbage collection.

### Reproduction frequency

100% if done correctly and with swiftness.

jjohnsnaill:
Also occurs with injections done via `new ILHook`.

Mirsario:
After fighting MonoMod's seemingly undocumented logging for quite a bit, I found that this is due to an internal SimpleNativeDetour object getting collected. Not an ILHook. 

![image](https://github.com/tModLoader/tModLoader/assets/6795251/67d97993-7f38-42b2-a1b5-297796209666)
![image](https://github.com/tModLoader/tModLoader/assets/6795251/84f5bf7d-6e5b-4bd6-9b1f-992ec655001d)

Mirsario:
[...]

PR #4306, opened 2024-07-18
Added PostTileFrame and ModifyFrameMerge hooks for ModTile

IbanPlay:
### What is the new feature?
Implementation of https://github.com/tModLoader/tModLoader/issues/4305
Adds `ModTile.PostTileFrame()` to let modders modify the tile frame after the vanilla calculations have taken place, and `ModTile.ModifyFrameMerge()`, a hook that runs after the merge connections have been calculated but before the frame has been decided to let modders change the merging of the tile selectively

`WorldGen.TileFrame()` is such a large method that it cannot be IL Edited safely due to https://github.com/tModLoader/tModLoader/issues/3878, making it impossible for modders to implement these easy hooks by themselves, and `ModTile.TileFrame()` is very restrictive, running before any vanilla code and only giving the modders the option to override the entire vanilla framing code and redo it from scratch. These 2 hooks can allow modders a much greater control over the framing of their modded tiles without needing to entirely copy the vanilla code to add 5 lines of edits on top.

None that I can think of that would let modders have a similar degree of freedom.

Making tiles with larger than 1x1 patterns like stone slabs without needing to recalculate the frame offset every frame in `AnimateIndividualTile()`, easily making tiles that have custom merge frames with other tiles than dirt and having selective control over tile merging in general, without needing to entirely override `TileFrame()` and copy huge chunks of it for minor changes

### ExampleMod updates
Added `ExampleCustomFramingTile` that showcases how the new hooks can be used to make a tile that has custom merge frames with snow and a 1x2 tile pattern.


Solxanich:
Just giving this a try here - doesn't look like the framing is working for merging with Snow? 
image

IbanPlay:
My bad, I was focusing on the merging with snow on the side of the tile and I forgot to set tilemerge for snow blocks so they try to merge with the tile as well (In the screenshot you can see the exampletile correctly use the custom merge frames, the issue was just that the snow tiles didn't try to merge back

PR #4325, opened 2024-07-30
Finish implementing deleting unloaded tiles feature, add VanillaFallbackOnModDeletion examples

JavidPack:
When a user decides to disable a mod, that mod will leave unloaded tiles in the world. If the user wishes to permanently remove a mod, then the tiles are a nuisance. The code exists to do this, replacing those tiles with `VanillaFallbackOnModDeletion`, but the code currently will not run because it is not hooked up to any UI for the user to click.

This PR also will add proper `VanillaFallbackOnModDeletion` usage and also update the documentation to warn about how to correctly use `VanillaFallbackOnModDeletion`.

TODO:
- [x] Do we also need to remove items from chests? Modded storages? Inventories? It seems like this could get complicated quickly and isn't quite as difficult for the user to manage themselves.
  - We can add additional parameters to the chat command in later PRs to allow users to pick and choose what they want to clear. For now we assume that users can easily clear out any unloaded item in chests by trashing items. 
- [x] ~~How should this effect be triggered? It's a rare desire, so we don't want users to accidentally press it or be confused by it. If we can't decide on a UI, a hidden keypress like holding shift can trigger this and we can just mention this to anyone who asks.~~ 
  - We've landed on a chat command.

# Example
In this example several tiles are replaced with suitable alternatives via `VanillaFallbackOnModDeletion` usage. ExampleTorch (bottom middle, near the player) shows the default behavior of turning back to dirt. ExampleWallAdvanced shows the default wall behavior, turning to air. ExampleChair, Chest, Block, Sign, Platform, Ore, Bar all revert to a vanilla tile in this example. On the left is TileObjectDataShowcase, which uses different FrameX and FrameY values from a normal 2x2 tile. This shows how a modded might make a mistake if they use `VanillaFallbackOnModDeletion` on a `tileframeimportant` tile that does not exactly match the layout of the vanilla tile they intend to fallback to.

**Before:**

**Unloaded:**

**Deleted and replaced:**

# Usage
The user will need to load the world in single player, then open up the chat and type and run the command. Here are some typical commands and options:

`/purgeunloaded -w -t` - This command will do a "dry run" of purging all unloaded walls and tiles from the world. Since it is a dry run, no tiles will actually be affected. The user can review the output to make sure nothing unexpected is being deleted. (for example, maybe they forgot ContentModA wasn't enabled currently.)

`/purgeunloaded -w -t -p` - This command will do the same as above but will actually remove the walls and tiles.


Chicken-Bones:
Removing items from storages would be a good option to provide, but isn't needed in the initial implementation. Items are much easier for the player to manage themselves.

Solxanich:
As far as modded chest to regular chest goes, 
If my memory serves the contents are preserved if it goes from a chest to a different chest. It's why unloaded chest works.

An alternative question is if the VanillaFallback type needs to default that modded chests fallback to wooden chests?

JavidPack:
We could do wooden chests specifically, or try to have all tiles default to style 0, but that might be hard to implement. The fallback feature is a really rare situation and I don't want every mod with content thinking they need to spend multiple lines declaring a fallback tile and tile style for all modtile. 

I think keeping it simple is fine, if someone ends up with a locked golden chest, I think that might be better than a selection of different chest style all turning into the same wooden chest. That said, most mods don't even use tile styles so most will be wooden chests anyway.

JavidPack:
New logging:

```
[18:00:19.209] [.NET TP Worker/INFO] [Terraria]: Loading World: 1313, IsCloud=True, Width: 4200, Height: 1200, Evil: 1, GameMode: 1
[18:00:19.266] [.NET TP Worker/INFO] [StatusText]: Loading world data
[18:00:21.553] [.NET TP Worker/INFO] [tML]: Unloaded tiles and walls have been purged from this world using the Ctrl-Alt-Shift shortcut while selecting the world. If this was unintentional please restore to a backup.
[18:00:23.098] [.NET TP Worker/INFO] [tML]: Purged tiles: ExampleMod/BasicTileEntityTile, ExampleMod/EnemyBanner, ExampleMod/Example1x1RubbleNatural, ExampleMod/ExampleAnimatedTile, ExampleMod/ExampleBar, ExampleMod/ExampleBlock, ExampleMod/ExampleCampfire, ExampleMod/ExampleCritterCage, ExampleMod/ExampleCustomFramingTile, ExampleMod/ExampleLamp, ExampleMod/ExampleMusicBoxTile, ExampleMod/ExampleOre, ExampleMod/ExampleSand, ExampleMod/ExampleStatue, ExampleMod/ExampleTorch, ExampleMod/ExampleTrap, ExampleMod/ExampleChandelier, ExampleMod/ExampleChest, ExampleMod/ExampleDoorClosed, ExampleMod/ExampleDoorOpen, ExampleMod/ExampleDresser, ExampleMod/ExamplePlatform, ExampleMod/ExampleTable, ExampleMod/ExampleWideBannerTile, ExampleMod/MinionBossRelic
[18:00:24.108] [.NET TP Worker/INFO] [tML]: Purged walls: ExampleMod/CrimsonFossilWall, ExampleMod/ExampleWall, ExampleMod/ExampleWallAdvanced
```

JavidPack:
I couldn't come up with any UI change I'd want to make to the UIWorldSelect that would look good or be sensible, so I figured we might as well merge the feature and just let people in support ever know if they need it. We can do a UI later.

Another option is a chat command in-game, but I don't think we should fill up the /help output more than we need to.

direwolf420:
Not a big fan of ctrl-alt-shift combination. These are all common "modifier" keys used in alot of programs. By themselves they shouldn't cause an action this severe, it should be coupled with some other key, maybe U for "unloaded".

Regarding no ingame way to view this feature, it should at the very least be documented on the wiki in a visible place (common issues faq?)

JavidPack:
Yeah, we decided it'll be moving it to a chat command instead.

JavidPack:
Here is the new command in action.

https://github.com/user-attachments/assets/dadb9c5a-5283-4cac-ac19-2f61f835fb38

patches/tModLoader/Terraria/ModLoader/Default/PurgeUnloadedCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+		Dictionary<string, int> purgedTiles = [];
+		Dictionary<string, int> purgedWalls = [];
+		var unloadedTileTypes = TileIO.Tiles.unloadedTypes.ToHashSet();
+		var unloadedWallType = ModContent.WallType<UnloadedWall>();
+		List<Point16> coordinates = [];
+
+		for (int i = 0; i < Main.maxTilesX; i++) {
+			for (int j = 0; j < Main.maxTilesY; j++) {
+				Tile t = Main.tile[i, j];
+				if (purgeTiles && t.HasTile && t.TileType >= TileID.Count && unloadedTileTypes.Contains(t.TileType)) {
```
Chicken-Bones: This will be faster if you use an array than a `HashSet`, No need for the `t.TileType >=` check then either

patches/tModLoader/Terraria/ModLoader/Default/PurgeUnloadedCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+			caller.Reply("At least one of '-w' or '-t' must be set");
+			caller.Reply(Usage);
+			return;
+		}
+
+		Dictionary<string, int> purgedTiles = [];
+		Dictionary<string, int> purgedWalls = [];
+		var unloadedTileTypes = TileIO.Tiles.unloadedTypes.ToHashSet();
+		var unloadedWallType = ModContent.WallType<UnloadedWall>();
+		List<Point16> coordinates = [];
```
Chicken-Bones: This _could_ OOM. You can call `Main.sectionManager.SetAllSectionsLoaded()` instead to reset the section framed cache to the freshly loaded state, then sections will re-frame when they come in view as normal.

patches/tModLoader/Terraria/ModLoader/IO/TileIO_Basic.cs
```diff
@@ -260,7 +260,7 @@ internal static TagCompound SaveBasics()
 		return tag;
 	}
 
-	internal static bool canPurgeOldData => false; //for deleting unloaded mod data in a save; should point to UI flag; temp false
+	internal static bool canPurgeOldData => false; //for deleting unloaded mod data in a save during loading; should point to UI flag; temp false; The /purgeunloaded chat command is the current approach for this
```
Chicken-Bones: Lets remove this :)

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
```diff
@@ -23,7 +23,11 @@ public abstract class ModBlockType : ModTexturedType, ILocalizedModType
 	/// <para/> Defaults to 0, which is <see cref="DustID.Dirt"/>. To prevent spawning any hit dust, set this to -1 instead. </summary>
 	public int DustType { get; set; }
 
-	/// <summary> The vanilla ID of what should replace the instance when a user unloads and subsequently deletes data from your mod in their save file. Defaults to <see cref="TileID.Dirt"/> (0). </summary>
+	/// <summary>
+	/// The vanilla ID of what should replace the instance when a user unloads and subsequently deletes data from your mod in their save file.
+	/// <br/><br/> <see cref="Main.tileFrameImportant"/> tiles should try to match the the layout (FrameX and FrameY values) of the fallback tile so that the resulting tiles aren't broken.
```
direwolf420: `the the`, ~~and it's `TileFrameX/Y` respectively~~

I'm thinking if the default being 0 could cause issues for exactly the reason outlined above. Maybe TML should handle  fallbacks differently based on `Main.tileFrameImportant`? A tileFrameImportant tile rn would fall back to 0 which would most certainly break stuff (if the original tile's frames are out of bounds of the 0 type sheet)

JavidPack: So, if a frameimportant tile falls back to a non-frameimportant tile, the resulting tiles will all be framed during the process so they'll fix themselves. Even though they had invalid frameXY values for a Dirt tile, during framing those will be fixed.

PR #4326, opened 2024-07-31
Fix #4283

### LINKED ISSUE #4283: Can't publish update to non-public mod, it creates a new workshop entry each time.
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

As of 2024.05, attempting to publish an update to a private/unlisted mod no longer updates the existing workshop entry. This is a regression caused by https://github.com/tModLoader/tModLoader/pull/4208.

We are looking into a fix. (https://discord.com/channels/103110554649894912/711551818194485259/1257506081228197908)

If you need to test a mod privately with friends, remember that you don't need to publish a mod to do that, just host a server and the mod will be synced automatically to connecting clients. 

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

Solxanich:
Adds a check against published mods by the user to determine if they have a non-public mod that general queries don't see. 

This fixes #4283, however:
This fix only works for the owner of the mod. A co-publisher/maintainer working via friends-only won't be able to see it. 

That said, that is highly unlikely and could be solved with some other more unstable tertiary attempts if absolutely needed. 

Also refactors more of the code to scale slight better and make code paths easier to identify

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -74,12 +74,19 @@ public partial class WorkshopHelper
 	{
 		item = null;
 
-		var query = new QueryHelper.AQueryInstance(new QueryParameters() { searchModSlugs = new string[] { modSlug } });
-		if (!query.TrySearchByInternalName(out var items))
+		var query = new QueryHelper.AQueryInstance(new QueryParameters() { queryType = QueryType.SearchDirect });
+		if (!query.TrySearchByInternalName(modSlug, out item))
 			return false;
 
-		item = items[0];
-		return item != null; // TODO, return value is ambiguous between a connection error and the mod not existing on workshop, currently both are logged as an error and the item is skipped
+		if (item == null) {
+			// search of all mods that were published by this user - Does not work with co-published/non-owner
+			// this should catch private / friends-only visibility
+			query.queryParameters.queryType = QueryType.SearchUserPublishedOnly;
+			if (!query.TrySearchByInternalName(modSlug, out item))
+				return false;
+		}
+
+		return true;
 	}
 
 	// Should this be in SteamedWraps or here?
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -74,12 +74,19 @@ internal static bool TryGetModDownloadItem(string modSlug, out ModDownloadItem i
...
-		return item != null; // TODO, return value is ambiguous between a connection error and the mod not existing on workshop, currently both are logged as an error and the item is skipped
+		if (item == null) {
+			// search of all mods that were published by this user - Does not work with co-published/non-owner
+			// this should catch private / friends-only visibility
+			query.queryParameters.queryType = QueryType.SearchUserPublishedOnly;
+			if (!query.TrySearchByInternalName(modSlug, out item))
+				return false;
+		}
+
+		return true;
```
Chicken-Bones, line 89: `TrySearchByInternalName` only returns `false` when the query fails. Previously, `TryGetModDownloadItem` would return `false` if `item == null`. Now it returns `true`. Perhaps the final line should still be `return item != null`?

In general, I think this approach of using `TryGet` with `false` indicating error, and requiring users to do an additional null check is quite error prone. In previous cleanups we've tried to keep nulls out of the system, and use exceptions or more detailed error codes.

It's okay to return `false` on both error, and item not existing, since this just classifies the item not existing as an error.

Chicken-Bones: In this case, the only prior usage of `TryGetModDownloadItem` is this missing deps downloader in `Interface.cs`. It needs a `null` check now.

If we had nullable assertions enabled on the codebase, this would be more okay. Unfortunately, an `Option` type in the system C# libraries is only planned for a future version of the language, but it might be worth using one of them to explicitly encode these presence or absences in the type system. Perhaps `Optional<T>` from `Microsoft.CodeAnalysis` which we already have available. (`null` is replaced by `default(Optional<T>)`)

I note that `TryGroupSearchByInternalName` also has to handle these `null` items, and `TryGetPublishIdByInternalName` converts them to `"0"` entries.

The two functions which call `TryGetPublishIdByInternalName` both end up discarding these `"0"` entries anyway, but perhaps there are times when `ModDownloadItem.PublishId.m_ModPubId` could also be `"0"`?

Solxanich: || I think this approach of using TryGet with false indicating error is error-prone
Suggesting separate PR to clean up the code around the null/false. 
Proposing that if 'Steam API unavailable', that it throw an exception instead of return false. 
Change 'false' to actually indicate the 'item not found' . 

|| perhaps there are times when ModDownloadItem.PublishId.m_ModPubId could also be "0"
0 is the default for type. The actual steamworks api will return a zero in several methods that fail.

PR #4328, opened 2024-07-31
[EM] Fix wrong order in ExampleChest unlock check

direwolf420:
### What is the bug?
ExampleChest, when clicked on, would consume the key even before unlocking the chest, so if any custom conditions a modder may have in `CanUnlockChest` (such as "plantera defeated") are not fullfilled, key is consumed but chest is not opened.

### How did you fix the bug?
Swap the consume and unlock lines.

No, this now mirrors vanilla logic.


JavidPack:
Your initial commit allowed unlocking the chest even if the key was not in inventory.

direwolf420:
My bad, oversight on my part

PR #4340, opened 2024-08-09
Add equipment loadouts support for ModAccessorySlot

### LINKED ISSUE #3285: [1.4.4]Make extra ModAccessorySlot follow current Equipment Loadout's color automatically

No

Mod capability as a Modder

### Description

tmodloader already provided a basic class called ModAccessorySlot for modders to create extra accessory slot, that's good;but the 1.4.4 update added Equipment Loadouts: a feature that allows the player to conveniently and quickly switch between three sets of armorand accessories, and those three sets have different background color,for example the second is blue and the third is pink, however, extra accessory slots added by mod are still green; this brings a sense of visual conflict.

image
image

_No response_

alueck:
### What is the new feature?
Equipment loadouts support for `ModAccessorySlot`
Also fixes #3285 

Because modded accessory slots should support the vanilla equipment loadout switching

The current implementation attempts to replicate the vanilla logic as closely as possible to facilitate adapting the code to future Terraria updates as easily as possible. 

https://github.com/tModLoader/tModLoader/blob/1.4.4/ExampleMod/Common/Players/ExampleAccessorySlot.cs shows the current examples of `ModAccessorySlot`. The `ExampleModWingSlot` example has been updated to support toggleable `HasEquipmentLoadoutSupport` in this PR: https://github.com/tModLoader/tModLoader/pull/4340/files#diff-f4ad9f01db3f16a5d9ef3f2e6aec684593990b7a861dc15fdee66056fe3ec2a6

Aside from setting `HasEquipmentLoadoutSupport` to an appropriate value, modders should not need to do anything different.  Action is only required if the modder wants the slot to not support loadouts. Modded accessories are still accessed in the same manner as before.

### Porting Notes
- `ModAccessorySlot` now supports equipment loadouts
  - By defaults loadout support will be true for any `ModAccessorySlot`. A modder can set the `ModAccessorySlot.HasEquipmentLoadoutSupport` property to `false` to disable it. The property value can also be toggled by a server-side config value, but a reload is required. When set to false, the `ModAccessorySlot` is "shared".
    - Changing the setting from `true` to `false` will cause the extra items stored to be spawned on the player when they enter the world.
    - Changing the value from `false` to `true` will result in the currently selected loadout holding the items.
    - `ModAccessorySlot` without custom backgrounds that support loadouts will be tinted with loadout colors as normal. If using a custom slot background, however, they will not be affected. A new hook, `ModAccessorySlot.BackgroundDrawColor`, can be used to further customize the color.
- `Player.DropItem` is now obsolete, use `Player.TryDroppingSingleItem` instead.
- `(Mod|Global)CanAccessoryBeEquippedWith/CanEquipAccessory` and `ModAccessorySlot.CanAcceptItem` are now potentially called on the server and remote clients instead of just the local client. This is to facilitate correctly syncing shared accessory slot conflicts. This means that there might be some syncing issues that need to be addressed.

### Discussion
- Support for additional loadouts is not included in this PR.
- `ModAccessorySlotPlayer.DropItems` has been changed to drop items even from disabled slots to match vanilla disabled slot logic. This change might affect some mod using slots for odd purposes.
- ~~I think `ModAccessorySlot.HasEquipmentLoadoutSupport` defaulting to `true` makes more sense. There was a brief discussion on this but I think we should revisit it.~~ We have gone with `true` now.
- This PR aligns drawing of modded slots with loadout support with vanilla drawing of accessory slots. The default background texture changed to `InventoryBack13`, which is basically white and is colored with a loadout and slot type specific color. Loadout colors are applied to `HasEquipmentLoadoutSupport` slots without a custom background texture.
  - Slots that do not support loadouts will continue drawing with the old textures. For players not using loadouts, they will be the same color, but players using loadouts will notice that "shared" slots stay the original green color to indicate that they are unchanged.

New background texture:
Inventory_Back13
Old textures:
Inventory_Back3
Inventory_Back8
Inventory_Back12
Shared and non-shared ModAccessorySlot, and vanilla slots in loadout 1:
image
The same in loadout 3:
image

### TODOs

- [x] Test multiplayer (tested locally with two tModLoader instances; synchronization worked fine)
- [x] Show info text if an item cannot be equipped in a shared slot, because of conflicts with any of the items equipped in any equipment loadout.
- [x] ~~Attempting to use a player that was saved while this PR was active prevents it from loading (OOB in ModAccessorySlotPlayer.LoadData). This will be an issue since we usually support loading players on older tModLoader versions, which users will be doing if anything goes wrong and they need to downgrade.~~ Fixed. But users will still lose accessories stored in ModAccessorySlot for the non-active loadouts if downgrading. This is fine.
- [x] Hook for customizing ModAccessorySlot draw color.

# Showcase
This video demonstrates the new behavior. There are several `ModAccessorySlot` active, 3 of them are shared. As the user switches loadouts, the shared slots stay green (like loadout 1) indicating that they are shared.

When the user switches to loadout 3, which has a functional wing accessory already, the shared wing accessory goes dark indicating that it is currently disabled. A chat message and hover message communicate to the player the issue. The player then moves the conflicting wing accessory and the slot becomes enabled again.

https://github.com/user-attachments/assets/74a8d906-7272-4a6a-9c41-7be9717e9aec

image


direwolf420:
Over on the tmodloader discord there is a discussion ongoing about implementation details and use cases: https://discord.com/channels/103110554649894912/1148455439491682425 if you are interested

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -325,55 +317,55 @@ public override void OnEquipmentLoadoutSwitched(int oldLoadoutIndex, int loadout
...
+			return false;
+
+		int functional = slotType % SlotCount;
+		int vanity = functional + SlotCount;
+
+		// Handle conflicts that arise from supporting shared slots.
+		bool functionalIssue = Loader.IsAccessoryInConflict(Player, exAccessorySlot[functional], functional, Terraria.UI.ItemSlot.Context.ModdedAccessorySlot);
+		bool vanityIssue = Loader.IsAccessoryInConflict(Player, exAccessorySlot[vanity], vanity, Terraria.UI.ItemSlot.Context.ModdedVanityAccessorySlot);
+
+		return functionalIssue || vanityIssue;
```
Chicken-Bones: It would be good to disable the vanity and functional slots independently. I'm not sure how feasible that is.

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -503,9 +497,20 @@ public bool ModdedIsItemSlotUnlockedAndUsable(int index, Player player, bool ign
...
+	/// </summary>
+	public bool IsAccessoryInConflict(Player player, Item checkItem, int slot, int context)
+	{
+		if (checkItem.IsAir)
+			return false;
+		bool canGoInSlot = CanPlayerAcceptItem(player, slot, checkItem, context);
+		if (!canGoInSlot)
+			return true;
+		Item[] itemCollection = player.armor.Concat(ModSlotPlayer(player).exAccessorySlot).ToArray();
+		itemCollection[slot + player.armor.Length] = new Item();
```
Chicken-Bones: Might be worth caching this `Item` instance

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
JavidPack: I was able to implement this approach.

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -3,40 +3,65 @@
...
 	internal Item[] exAccessorySlot;
+	/// <summary> <see cref="ModAccessorySlot"/> corollary to <see cref="Player.dye"/>. </summary>
 	internal Item[] exDyesAccessory;
+	/// <summary> <see cref="ModAccessorySlot"/> corollary to <see cref="Player.hideVisibleAccessory"/>. </summary>
 	internal bool[] exHideAccessory;
-	internal Dictionary<string, int> slots = new Dictionary<string, int>();
+	/// <summary> All slots that can potentially show for this user. First the loaded ModAccessorySlots, followed by any unloaded slots. </summary>
+	private readonly Dictionary<string, (int SlotType, bool HasLoadoutSupport)> slots = [];
+	/// <summary> Which slots are shared, indexed by ModAccessorySlot.Type </summary>
+	private readonly HashSet<int> sharedLoadoutSlotTypes = [];
```
Chicken-Bones: Should still be an array.

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -3,40 +3,65 @@
...
 	// Arrays for modded accessory slot save/load/usage. Used in DefaultPlayer.
+	/// <summary> <see cref="ModAccessorySlot"/> corollary to the accessory and vanity slots of <see cref="Player.armor"/>. </summary>
 	internal Item[] exAccessorySlot;
+	/// <summary> <see cref="ModAccessorySlot"/> corollary to <see cref="Player.dye"/>. </summary>
 	internal Item[] exDyesAccessory;
+	/// <summary> <see cref="ModAccessorySlot"/> corollary to <see cref="Player.hideVisibleAccessory"/>. </summary>
 	internal bool[] exHideAccessory;
-	internal Dictionary<string, int> slots = new Dictionary<string, int>();
+	/// <summary> All slots that can potentially show for this user. First the loaded ModAccessorySlots, followed by any unloaded slots. </summary>
+	private readonly Dictionary<string, (int SlotType, bool HasLoadoutSupport)> slots = [];
```
Chicken-Bones: Looks like this `bool HasLoadoutSupport` is unused now.

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -86,6 +126,26 @@ public override void LoadData(TagCompound tag)
...
+	public override void OnEnterWorld()
+	{
+		// Need to enter world with disabled slots.
+		DetectConflictsWithSharedSlots();
+
+		if (extraItems.Count == 0)
+			return;
+
+		foreach (var items in extraItems) {
+			if(!items.Accessory.IsAir)
```
Chicken-Bones: Might be worth running an `(if|for|foreach)\(` to find missing spaces periodically

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -193,16 +285,242 @@ public override void SendClientChanges(ModPlayer clientPlayer)
...
+		/// <returns>
+		/// A collection of <see cref="SlotInfo"/> objects for slots, which are not added to the loadout,
+		/// because <see cref="ModAccessorySlot.HasEquipmentLoadoutSupport"/> changed since the last save.
+		/// </returns>
+		public IReadOnlyList<SlotInfo> LoadData(
+			TagCompound tag,
+			List<string> order,
+			Dictionary<string, (int SlotType, bool HasLoadoutSupport)> slots)
+		{
+			List<SlotInfo> result = [];
```
Chicken-Bones: This should be `List<Item> itemsToDrop`. Modify the return type and `extraItems` accordingly

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -193,16 +285,242 @@ public override void SendClientChanges(ModPlayer clientPlayer)
...
+			IList<Item> items;
+			IList<Item> dyes;
+			IList<bool> visible;
```
Chicken-Bones: No need for forward declaration, just `var` below

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -193,16 +285,242 @@ public override void SendClientChanges(ModPlayer clientPlayer)
...
+				(int type, bool hasLoadoutSupport) = slots[order[i]];
+
+				Item dye = dyes.ElementAtOrDefault(i) ?? new Item();
+				Item accessory = items.ElementAtOrDefault(i) ?? new Item();
+				Item vanityItem = items.ElementAtOrDefault(i + order.Count) ?? new Item();
+				bool isHidden = visible.ElementAtOrDefault(i);
+
+				// If this slot has any acc/van/dye item, but doesn't have loadout support, the items shouldn't be in the loadout at all.
+				// We can try to move them to the inventory or try to put them in empty acc slots, but it might get complicated with slots and restrictions. Also the user might not want it there. They will be returned to the player OnEnterWorld.
+				bool hasLoadoutSupportSettingForSlotChanged = !hasLoadoutSupport && (!dye.IsAir || !accessory.IsAir || !vanityItem.IsAir);
```
Chicken-Bones: Can simplify this with above ^. Just `if (!hasLoadoutSupport)` add each non-air item to `itemsToDrop`

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -193,16 +285,242 @@ public override void SendClientChanges(ModPlayer clientPlayer)
...
+			dyes = tag.GetList<TagCompound>("dyes").Select(ItemIO.Load).ToList();
+			visible = tag.GetList<bool>("hidden").ToList();
+
+			for (int i = 0; i < order.Count; i++) {
+				(int type, bool hasLoadoutSupport) = slots[order[i]];
+
+				Item dye = dyes.ElementAtOrDefault(i) ?? new Item();
+				Item accessory = items.ElementAtOrDefault(i) ?? new Item();
+				Item vanityItem = items.ElementAtOrDefault(i + order.Count) ?? new Item();
+				bool isHidden = visible.ElementAtOrDefault(i);
```
Chicken-Bones: Why are we using `ElementAtOrDefault`? Is there some way the data could be missing? Scary

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
```diff
@@ -193,16 +285,242 @@ public override void SendClientChanges(ModPlayer clientPlayer)
...
+			bool[] hideVisibleAccessory = modAccessorySlotPlayer.exHideAccessory;
+			for (int k = 0; k < hideVisibleAccessory.Length; k++) {
+				if (modAccessorySlotPlayer.IsSharedSlot(k))
+					continue;
+				Utils.Swap(ref hideVisibleAccessory[k], ref ExHideAccessory[k]);
+			}
+		}
+	}
+
+	internal sealed record SlotInfo
```
Chicken-Bones: Correspondingly, this class will be unused

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
JavidPack: Done.

patches/tModLoader/Terraria/ModLoader/Default/ModAccessorySlotPlayer.cs
JavidPack: Done.

PR #4343, opened 2024-08-11
Fix item variants not being shown in NPC shops (#4280)

### LINKED ISSUE #4280: Magic Dagger Prices in Don't Dig Up World
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

In Don't dig up worlds the Stats of the Magic Dagger don't change accordingly.
I have no mods installed and did a fresh install of tModLoader.

### Log File

Test World

### Steps to reproduce

1. Have a Journey mode character and Journey mode don't dig up world ready. (Unless you want to waste a couple of hours)
2. Activate 10x Enemy slider and stay in the Underground Layer for the Skeleton Merchant to spawn. I advise a Hunter potion and Lifeform analyzer
3. Check if the moon phase is in third quarter. If it isn't change the ingame timer to 24x
4. Check the Price on the Magic Dagger

### Expected Behavior

Magic Dagger should have it's nerfed stats.

### Actual Behavior

Magic Dagger has it's normal stats.

### Reproduction frequency

I reproduced it 7 times and every time it came out the same.

### Additional Information

If you spawn in the Magic Dagger from the Journey menu, it has it's adjusted stats.

alueck:
### What is the bug?
NPC shops only list the base items, not item variants if any are active. Fixes #4280 

### How did you fix the bug?
I edited the NPC shop entry class to always return a matching item variant if an active variant exists.
Since item variants are always registered with conditions which must be met, this is probably feasible. On the other hand this behavior might be a bit unexpected if an item is added to the shop using the `Add(Item item, params Condition[] condition)` method instead of `Add(int item, params Condition[] condition)`.

The NPC shop database is build on game startup, so selecting the item variant cannot be done at this stage. The only alternative I see, is the `NPCShop.FillShop` method where the items are currently read from the shop entries.

### DIFF
--- a/patches/tModLoader/Terraria/Item.cs.patch
+++ b/patches/tModLoader/Terraria/Item.cs.patch
@@ -484,7 +484,7 @@
  	public sbyte handOnSlot = -1;
  	public sbyte handOffSlot = -1;
  	public sbyte backSlot = -1;
-@@ -121,49 +_,247 @@
+@@ -121,49 +_,248 @@
  	public sbyte faceSlot = -1;
  	public sbyte balloonSlot = -1;
  	public sbyte beardSlot = -1;
@@ -580,6 +580,7 @@
 +	/// <summary>
 +	/// The number of copper coins this item is worth (aka, cost to buy from a merchant). Setting it to <c>10462</c> would mean the item cost 1 gold, 4 silver, and 62 copper. The sell price of an item is one fifth of its value. Value also influences reforge costs with the goblin tinkerer.
 +	/// <br/> For convenience, you can also use the <see cref="Item.buyPrice(int, int, int, int)"/> method for setting values: <c>Item.value = Item.buyPrice(0, 1, 4, 62);</c> You can also use the <see cref="Item.sellPrice(int, int, int, int)"/> method if you would rather think about an item's value the other way. Both <c>Item.buyPrice(0, 0, 10, 55)</c> and <c>Item.sellPrice(0, 0, 2, 11)</c> would set the value to <c>1055</c>.
++	/// <para/> If custom purchase prices disconnected from the item value are needed, <see cref="shopCustomPrice"/> should be set for the item instance added to a shop entry.
 +	/// </summary>
  	public int value;
  	public bool buy;
@@ -778,7 +779,7 @@
  	public bool PaintOrCoating {
  		get {
  			if (paint <= 0)
-@@ -196,11 +_,19 @@
+@@ -196,12 +_,24 @@
  
  	public bool FitsAccessoryVanitySlot => true;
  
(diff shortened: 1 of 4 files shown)


JavidPack:
Thanks for looking into this and getting a fix started. 

I believe, however, that the correct approach to this is to do the item swap in `Player.RefreshItems` and call `Item.Refresh` on the items directly. I haven't tested yet, but if you could look into that approach that would be great.

Note: I'll need to check shopcustomprice as well, not sure what the correct solution is.

alueck:
@JavidPack  `Player.RefreshItems` only handles items the player already owns. The item tooltip in the shop window would still display incorrect values.

JavidPack:
> @JavidPack `Player.RefreshItems` only handles items the player already owns. The item tooltip in the shop window would still display incorrect values.

Yeah, I know, I'm saying we need to add code there to handle shop entries as well.

alueck:
Ah, sorry, misunderstood your comment. I did a quick test, it works. But two things:
- Shop entries use items by reference from `ContentSamples.ItemsByType`. By calling `Refresh` we actually modify the item in `ContentSamples`. Can't tell if this might actually be desired behavior.
- To me, it seems a bit weird that the `Player` instance updates items of NPC shops. I would not see this as a concern to be handled by the player instance.

In general, I agree it should be done in a similar way to `Player.RefreshItems` though. So that it would also work for mods, which use their own implementation of an `AbstractNPCShop.Entry`.

Edit: After looking a bit more into it, I did not find a good alternative for `Player.RefreshItems`. Doing the NPC shop refresh there will at least also work in multiplayer.

JavidPack:
I see, `public NPCShop Add(int item, params Condition[] condition) => Add(ContentSamples.ItemsByType[item], condition);` is probably incorrect, might need to be `public NPCShop Add(int item, params Condition[] condition) => Add(new Item(item), condition);` to work correctly.

I'll double check with others that this makes sense when I review this. Looks good in the meantime, I'll get to this in a bit.

alueck:
`shopCustomPrice` is still a problem. If there is an item variant active for an item in a world the player loads into, the custom price will always be lost. We could just temporarily save the custom price, before refreshing the item and restore it afterwards. This could be kind of unexpected though and it also does not make much sense to set one plain custom price for all variants of an item.

If we don't restore the custom price, we run into another issue.
**Given**: You register all the variants with the appropriate conditions and custom prices for a shop yourself. 
**When**: You load into a world, where only the weak variant condition is met.
**Then**: The custom price will work, because the item variant did not change and the item is therefore not refreshed. All other shop entries containing the other item variants will be refreshed to represent the weak variant. 

**When**: You then load into a world where only the condition for the strong variant is met.
**Then**: The custom price for the strong variant will be lost, because of the previous refresh which happened when entering world that met the weak variant condition.

JavidPack:
Thanks for your insight. I went with the approach of leaving it up to the modder to support custom prices for these items by adding multiple shop entries for each item. This matches the pattern shown by npc loot and chest loot for similar situations.

This also makes sense because a modnpc selling one of these items would want to anticipate that and NOT sell the item on those secret seeds in the first place. 

Note: This PR keeps `ContentSamples.ItemsByType` consistent once again. These item instance won't change based on world. This behavior can be changed if needed but I think this approach makes the most sense. Prior to this commit half the ItemVariant items were being changed, so not changing all of them sounds more correct.

PR #4345, opened 2024-08-14
Store Mod Hash Data in Developer Metadata

Solxanich:
### What is the new feature?
To support investigations in to mods and verifying  authenticity, we would like to store ongoing hash data of .tmods published to tML Workshop. 

This PR implements the storage of Mod Hashes on to the Developer Metadata field. 
This PR also refactors the related ModDownloadItem processing method.

NOTE: Although Workshop can store if a mod is currently banned, Workshop also deletes banned mods after 30 days. So, treat all non-Workshop matchable mods as potentially banned mods.

Add the opportunity to validate authenticity of a tmod file.

TBD

### TODO:
- [x] Test in-game publishing and all newly added methods.
- [x] Test failure modes
- [x] Test publishing a brand new mod 
- [x] Test publishing an update to a mod that wasn't involved in PR development
- [x] Build a tool to backfill data in to mods already on Workshop

### Deferred Github Actions Scope
- [ ] Add this feature to SteamCMD Publish Preparer && build.yml --> Deferred
- [ ] Test SteamCMD Publish Preparer --> Deferred
- [ ] Test build.yml (example Mod) --> Deferred


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -34,6 +34,14 @@ public class ModPublisherInstance : UGCBased.APublisherInstance
 		protected override void PrepareContentForUpdate() { }
 	}
 
+	internal enum WorkshopSearchReturnState
+	{
+		SearchSuccessfulNoMatch,
+		SearchFailed,
+		SearchSuccessMatchUseable,
+		SearchSuccessMatchUnusable
+	}
```
Chicken-Bones, line 43:
```suggestion
	internal enum SearchResult
	{
		NotFound,
		SearchFailed,
		Success,
		RetrievalFailed,
	}
```

patches/tModLoader/Terraria/Social/Steam/SteamedWraps.cs
```diff
@@ -146,6 +147,7 @@ private static void ModifyQueryHandle(ref UGCQueryHandle_t qHandle, QueryParamet
 		if (SteamClient) {
 			SteamUGC.SetAllowCachedResponse(qHandle, 0); // Anything other than 0 may cause Access Denied errors.
 
+			SteamUGC.SetReturnMetadata(qHandle, true);
```
Chicken-Bones: Do we want this always on? Will it affect our retrievable performance? Perhaps we should have a `QueryParameters` flag for it

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -264,6 +273,63 @@ private static string CalculateDescriptionAndChangeNotes(bool isCi, NameValueCol
...
+	//TODO: Add a Spec for what this is supposed to be doing
+	private static string CalculateDeveloperMetadata(List<ModVersionHash> versionHashes)
+	{
+		string devMetadata = JsonConvert.SerializeObject(new DeveloperMetadata() { hashes = versionHashes.Select(h => h.ToString()).ToList() });
+		while (devMetadata.Length > Steamworks.Constants.k_cchDeveloperMetadataMax) {
+			// If we had any reserved metadata space, we would process that first.
+
+			// We don't have any reserved metadata space, so we can set availableChars to DeveloperMetadataMax
+			StripOldHashes(versionHashes, Steamworks.Constants.k_cchDeveloperMetadataMax);
+		}
```
Chicken-Bones: Need to put the `JsonConvert` inside the loop. Unfortunately this doesn't fit `do-while` either, so you'll have to use 
```cs
while (true) {
    string devMetadata = JsonConvert
    if (devMetadata.Length <= Steamworks.Constants.k_cchDeveloperMetadataMax)
        break;
    
    StripOldHashes
}

File: patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs [hunk 4]
```diff
@@ -264,6 +273,63 @@ private static string CalculateDescriptionAndChangeNotes(bool isCi, NameValueCol
...
+		string devMetadata = JsonConvert.SerializeObject(new DeveloperMetadata() { hashes = versionHashes.Select(h => h.ToString()).ToList() });
+		List<ModVersionHash> current = versionHashes;
+
+		int index = 100;
+		while (devMetadata.Length > availableCharacters) {
+			current = OptimizeHashesForCurrentBrowserVersion(versionHashes, index-- - reserved.Count());
+			current.AddRange(reserved);
+
+			devMetadata = JsonConvert.SerializeObject(new DeveloperMetadata() { hashes = current.Select(h => h.ToString()).ToList() });
+		}
```
Chicken-Bones: No while loop here. Keep the code simple. We should just drop the oldest entry each time. Rather than 'reserving' items, we were just going to sort them so that the ones we most want to keep are at the top of the list.

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -264,6 +273,63 @@ private static string CalculateDescriptionAndChangeNotes(bool isCi, NameValueCol
...
+			.Take(amountToKeep).ToList();
+	}
+
+	private static List<ModVersionHash> OptimizeHashesForCurrentBrowserVersion(List<ModVersionHash> versionHashes, int amountToKeep)
+	{
+		return versionHashes.Where(e => SocialBrowserModule.GetBrowserVersionNumber(e.tmlVersion) == SocialBrowserModule.GetBrowserVersionNumber(BuildInfo.tMLVersion))
+			.OrderByDescending(t => t.tmlVersion.MajorMinor())
+			.ThenByDescending(v => v.modVersion)
+			.Take(amountToKeep).ToList();
+	}
```
Chicken-Bones: This is a bit more complicated than what we discussed in Discord. I'm also not sure the ordering logic is correct. The list should contain the newest from each tML major version, followed by all the remaining versions from newest to oldest.

I thought the plan was to ignore browser versions, since we're not adding the workshop logic to those releases. When 1.4.5 comes, we'll still have plenty of months of overlap with 1.4.4 versions.

Chicken-Bones (changes requested):
Good cleanup

patches/tModLoader/Terraria/Social/Base/SocialBrowserModule.cs
```diff
@@ -16,6 +21,46 @@ public struct ModPubId_t
...
+
+	public ModVersionHash(string serializedVersionHash)
+	{
+		var match = new Regex(pattern).Match(serializedVersionHash);
+
+		if (!match.Success)
+			throw new SocialBrowserException($"Malformed Hash Data Detected {serializedVersionHash}");
+
+		tmlVersion = new Version(match.Groups[1].Value);
+		modVersion = new Version(match.Groups[2].Value);
```
Chicken-Bones: I thought we had agreed on not including versions in the hash list

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -459,41 +466,57 @@ internal void WaitForQueryResult(SteamAPICall_t query)
...
 				string ownerId = pDetails.m_ulSteamIDOwner.ToString();
 
 				DateTime lastUpdate = Utils.UnixTimeStampToDateTime((long)pDetails.m_rtimeUpdated);
 				string displayname = pDetails.m_rgchTitle;
+				bool banned = pDetails.m_bBanned;
 
-				// Item Tagged data
+				if (banned) {
+					throw new SocialBrowserException($"Item {id}: '{displayname}' is banned. Skipping...");
+				}
```
Chicken-Bones: This is fine for now, but the final download mods from servers feature will probably need to be extra serious about preventing the download of banned mods. Maybe we'll make an exception subclass

SteamWebAPIPublisherTool/ModLoaderSimple/EntryReadStream.cs
```diff
@@ -0,0 +1,84 @@
+using System;
+using System.IO;
+
+namespace ModLoaderSimple;
+
+internal sealed class EntryReadStream : Stream
```
Chicken-Bones: Any reason you can't reference tML as a library?

Chicken-Bones: Use https://github.com/tModLoader/tModLoader/pull/4731 and `[InternalsVisibleTo]` if required

SteamWebAPIPublisherTool/BatchMetadataSetRunner.cs
```diff
@@ -0,0 +1,82 @@
...
+	{
+		var devMetadataKvp = IterateWorkshopFilesForDevMetadata(actualWorkshopItemsFolder);
+
+		foreach (var item in devMetadataKvp) {
+			SteamWebApi.SteamWebWrapper.SetDeveloperMetadata(item.publishedId, item.metadata);
+			Console.WriteLine($"Metadata for Workshop Item {item.publishedId} has been updated");
+		}
+
+		// Free up disk drive space by cleaning out workshop items folder when complete
+		Directory.Delete(actualWorkshopItemsFolder, deleteModsWhenComplete);
```
Chicken-Bones: According to docs, this will throw if the directory is non empty and `deleteModsWhenComplete: false`

patches/tModLoader/Terraria/Social/Base/SocialBrowserModule.cs
```diff
@@ -16,13 +21,52 @@ public struct ModPubId_t
...
+	public override string ToString() => $"{hash}";
+
+	public ModVersionHash(string hash)
+	{
+		this.hash = hash;
+	}
+
+	public ModVersionHash(TmodFile modFile)
+	{
+		hash = Encoding.UTF8.GetString(modFile.Hash);
```
Chicken-Bones: I thought we were going to use `Convert.ToBase64String`. `TmodFile.Hash` is a full `byte[20]` not text encoded.

patches/tModLoader/Terraria/Social/Base/SocialBrowserModule.cs
```diff
@@ -16,13 +21,52 @@ public struct ModPubId_t
 	public string m_ModPubId;
 }
 
+public struct ModVersionHash
+{
+	private string hash; // 28+2 chars, SHA1. +2 is for string type
```
Chicken-Bones: comment accuracy? What's 'string type' here (I forget)

SteamWebAPIPublisherTool/SteamWebAPIPublisherTool.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
```
Chicken-Bones: I think this project belongs subfoldered in `extras`

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadItem.cs
```diff
@@ -37,7 +41,7 @@ public class ModDownloadItem
 
 	public bool IsInstalled => Installed != null;
 
-	public ModDownloadItem(string displayName, string name, Version version, string author, string modReferences, ModSide modSide, string modIconUrl, string publishId, int downloads, int hot, DateTime timeStamp, Version modloaderversion, string homepage, string ownerId, string[] referencesById)
+	public ModDownloadItem(string displayName, string name, Version version, string author, string modReferences, ModSide modSide, string modIconUrl, string publishId, int downloads, int hot, DateTime timeStamp, Version modloaderversion, string homepage, string ownerId, string[] referencesById, bool banned, string devMetadata)
```
Chicken-Bones: Remove `Banned` argument, should always be false for this constructor

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadItem.cs
```diff
@@ -56,6 +60,15 @@ public ModDownloadItem(string displayName, string name, Version version, string
 		TimeStamp = timeStamp;
 		Version = version;
 		ModloaderVersion = modloaderversion;
+		Banned = banned;
+		DevMetadata = devMetadata;
+	}
+
+	internal ModDownloadItem(string displayName, string publishId, bool banned = true)
```
Chicken-Bones: Remove default value for `banned` arg. Ideally we'd make this `ctor` and have a clearly named factory method, but `new ModDownloadItem(e.displayName, e.modPubId, banned: true)` works perfectly fine as you have it

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadItem.cs
```diff
@@ -94,6 +107,15 @@ public override int GetHashCode()
 		return GetComparable().GetHashCode();
 	}
 
+	public List<ModVersionHash> GetModVersionHashes()
+	{
+		var devMetadata = JsonConvert.DeserializeObject<DeveloperMetadata>(DevMetadata);
```
Chicken-Bones: is it worth doing some future proofing by having the `DevMetadata` be a json object at root, and having the `hashes` list be a field? If we are concerned about the limit we could also just have the hashes be a single string with a separator like `|`.

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -25,7 +26,8 @@ namespace Terraria.Social.Steam;
 public partial class WorkshopHelper
 {
 	internal static string[] MetadataKeys = new string[8] { "name", "author", "modside", "homepage", "modloaderversion", "version", "modreferences", "versionsummary" };
-	private static readonly Regex MetadataInDescriptionFallbackRegex = new Regex(@"\[quote=GithubActions\(Don't Modify\)\]Version Summary: (.*) \[/quote\]", RegexOptions.Compiled);
+	//private static readonly Regex MetadataInDescriptionFallbackRegex = new Regex(@"\[quote=GithubActions\(Don't Modify\)\]Version Summary: (.*) \[/quote\]", RegexOptions.Compiled);
+	private static readonly Regex MetadataInDescriptionFallbackRegex = new Regex(@"Version Summary: ", RegexOptions.Compiled);
```
Chicken-Bones: why is this changing?

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -314,8 +321,12 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
 
 						for (int j = 0; j < _queryReturnCount; j++) {
 							var itemsIndex = j + i * Constants.kNumUGCResultsPerPage;
-							var item = GenerateModDownloadItemFromQuery((uint)j);
-							if (item is null) {
+							var match = TryGenerateModDownloadItem((uint)j, out var item);
+							if (match != WorkshopSearchReturnState.Success) {
+								if (match == WorkshopSearchReturnState.Banned) {
+									throw new BannedModException($"The mod {item.DisplayName} with ID {item.PublishId} is Banned!", item.DisplayName, item.PublishId.ToString());
```
Chicken-Bones: The `message` here should probably be generated inside the `BannedModException` class, and should probably be localized. Also no need for braces for single line guard `if`

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -314,8 +321,12 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
...
+								}
+
```
Chicken-Bones, line 315: Is throwing an exception going to cause problems if the query accidentally matches a banned mod? This method previously didn't throw. Is this one of the cases where a mod should actually be skipped if banned, or is it a case where an item should be returned with `Banned: true`? Needs caller analysis.

In addition, the comment at the top of this method says
```
			/// Note that the List size is 1 to 1 with the provided array.
			/// If the Mod is not found, the space is filled with a null.
```

Which appears to be outdated, probably from when we added the `out List<string> missingMods`

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -369,8 +380,8 @@ private IEnumerable<ModDownloadItem> ProcessPageResult()
 			{
 				// Appx. 10 ms per page of 50 items
 				for (uint i = 0; i < _queryReturnCount; i++) {
-					var mod = GenerateModDownloadItemFromQuery(i);
-					if (mod == null)
+					var state = TryGenerateModDownloadItem(i, out var mod);
+					if (state != WorkshopSearchReturnState.Success)
```
Chicken-Bones: skipps banned items, lgtm :)

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -411,15 +425,20 @@ internal bool TrySearchByInternalName(string slug, out ModDownloadItem item)
...
 					if (_queryReturnCount == 0) {
 						Logging.tML.Info($"No Mod on Workshop with internal name: {slug}");
-						return true;
+						return WorkshopSearchReturnState.RetrievalFailed;
+					}
+
+					var match = TryGenerateModDownloadItem(0, out item);
+
+					if (match == WorkshopSearchReturnState.Banned) {
+						throw new BannedModException($"The mod {item.DisplayName} with ID {item.PublishId} is Banned!", item.DisplayName, item.PublishId.ToString());
```
Chicken-Bones: Similarly, to above, is it okay to throw here now?

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -459,41 +478,61 @@ internal void WaitForQueryResult(SteamAPICall_t query)
...
 
 				string ownerId = pDetails.m_ulSteamIDOwner.ToString();
 
 				DateTime lastUpdate = Utils.UnixTimeStampToDateTime((long)pDetails.m_rtimeUpdated);
 				string displayname = pDetails.m_rgchTitle;
+				bool banned = pDetails.m_bBanned;
 
-				// Item Tagged data
+				if (banned) {
+					throw new BannedModException($"Item {id}: '{displayname}' is banned. Skipping...", displayname, id.ToString());
```
Chicken-Bones: No need to say `Skipping...`, in fact it's not really this method's job to say what's going to happen with the exception when it's handled

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -123,7 +129,7 @@ private bool _PublishMod(TmodFile modFile, NameValueCollection buildData, Worksh
 		}
 
 		// Check for Beta
-		if (BuildInfo.IsDev) {
+		if (BuildInfo.IsDev && modFile.Name != "ToBeDeleted") {
```
Chicken-Bones, line 134: leftover test code?

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -19,28 +20,33 @@ public partial class WorkshopSocialModule
...
+		if (state == WorkshopHelper.WorkshopSearchReturnState.SearchFailed || state == WorkshopHelper.WorkshopSearchReturnState.RetrievalFailed) {
 			IssueReporter.ReportInstantUploadProblem("tModLoader.NoWorkshopAccess");
 			return false;
 		}
 
 		currPublishID = 0;
+		
 
-		if (mod == null) {
+		if (state == WorkshopHelper.WorkshopSearchReturnState.RetrievalFailed) {
```
Chicken-Bones, line 39: the condition above already guards this one

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -189,6 +202,54 @@ private bool _PublishMod(TmodFile modFile, NameValueCollection buildData, Worksh
...
+	private static void CalculateModHashes(string workshopPath, ModDownloadItem modDownloadItemAsFound, ref DeveloperMetadata devMetadata)
+	{
+		// Get the hashes from the existing modDownloadItem as found on the workshop
+		var prevHashes = modDownloadItemAsFound.GetModVersionHashes();
+
+		// Get the new hashes
+		var currentHashes = new List<ModVersionHash>();
+		foreach (var tModPath in Directory.EnumerateFiles(workshopPath, "*.tmod*", SearchOption.AllDirectories)) {
+			var tModFile = new TmodFile(tModPath);
+			tModFile.Open(); // Needed for Hash data to be populated
```
Chicken-Bones: `using var _ = tModFile.Open();` since it returns an `IDisposable` handle

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -189,6 +202,54 @@ private bool _PublishMod(TmodFile modFile, NameValueCollection buildData, Worksh
...
+		devMetadata.modVersionHashes = totalHash.Select(h => h.ToString()).ToList();
+	}
+
+	// This methods trims contents of developer metadata based on the preferred order of discarding information.
+	// It is primarily written with the intent of 'in case' we need to store other information in this Workshop text field
+	private static void TrimDevMetadata(ref DeveloperMetadata devMetadata)
+	{
+		const int MaxMetadataLength = Steamworks.Constants.k_cchDeveloperMetadataMax;
+
+		var overflowLength = JsonConvert.SerializeObject(devMetadata).Length - MaxMetadataLength;
```
Chicken-Bones: We should probably use serializer params that remove any unnecessary whitespace.

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -189,6 +202,54 @@ private bool _PublishMod(TmodFile modFile, NameValueCollection buildData, Worksh
...
+		if (overflowLength <= 0)
+			return;
+
+		int charsSaved = 0;
+
+		// Check if we can reduce the number of ModHashes
+		var minNumberOfHashes = 2 * SocialBrowserModule.keepRequirements.Select(a => a.keepCount).Sum();
```
Chicken-Bones: Surely this whole function could be simpler if we just ran a while loop and dropped the oldest hash and reserialized. There's only ever going to be a small handful of hashes to drop, and serialization takes a negligible amount of time. 

We can check that the number of hashes we kept at the end is at least the number in the current upload for your exception below

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,124 @@
...
+			modVersionHashes = new List<ModVersionHash>();
+			return;
+		}
+
+		modVersionHashes = devMetadata.modVersionHashes;
+	}
+
+	internal string GetSerialize()
+	{
+		return System.Text.Json.JsonSerializer.Serialize(this);
```
Chicken-Bones: Lets stick to Newtonsoft.Json, since that's what we use elsewhere in tML.

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,124 @@
...
+	public override string ToString() => hash;
+
+	public ModVersionHash(string encodedHash)
+	{
+		hash = encodedHash;
+	}
+
+	public ModVersionHash(TmodFile modFile)
+	{
+		hash = Encoding.UTF32.GetString(modFile.Hash);
```
Chicken-Bones: As mentioned on Discord, we want to make sure we use `Convert.ToBase64`

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,124 @@
...
+using Newtonsoft.Json;
+using Terraria.ModLoader.Core;
+using Terraria.ModLoader.UI.ModBrowser;
+using Terraria.Social.Steam;
+
+namespace Terraria.Social.Base;
+
+public struct ModVersionHash
+{
+	private string hash;
```
Chicken-Bones, line 15: We want to make sure that json serializes this as a string, and not as an object with a `hash` field `{ hash: "..." }`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -206,11 +206,18 @@ internal static void ModLoaderMenus(Main main, int selectedMenu, string[] button
...
+						if (state == WorkshopHelper.WorkshopSearchReturnState.SearchFailed)
+							break;
+
+						if (state != WorkshopHelper.WorkshopSearchReturnState.Success) {
 							Logging.tML.Error($"Could not find required mod dependency on Workshop: {slug}");
 							continue;
 						}
 
+						if (item.Banned)
+							throw new BannedModException($"The missing dependency {item.DisplayName} is Banned on Workshop.", item.DisplayName, item.PublishId.ToString());
```
Chicken-Bones: Tested, throwing at this point causes the game to hard crash, probably sufficient to just log error and skip

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -224,11 +231,18 @@ await UIModBrowser.DownloadMods(
 					// Revisit this code at a later date. Its not apparent how well the interaction of both dependencies and removed mods will play out in terms of UX
 					HashSet<ModPubId_t> removedDownloads = new();
 					foreach (var slug in removedMods) {
-						if (!WorkshopHelper.TryGetModDownloadItem(slug, out var item) || item == null) {
+						var state = WorkshopHelper.QueryHelper.AQueryInstance.TryGetModDownloadItem(slug, out var item);
+						if (state == WorkshopHelper.WorkshopSearchReturnState.SearchFailed)
+							break;
+
+						if (state != WorkshopHelper.WorkshopSearchReturnState.Success) {
 							Logging.tML.Error($"Could not find removed mod on Workshop: {slug}");
```
Chicken-Bones: might as well pass the `state` through to the log message. Lil more info

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+				modVersionHashes = new List<ModVersionHash>();
+				return;
+			}
+
+			modVersionHashes = devMetadata.modVersionHashes;
+		}
+		catch (Exception) {
+			modVersionHashes = new List<ModVersionHash>();
+		}
+	}
```
Chicken-Bones: Better practice is to move the error handling to the `Deserialize` factory method. You can put `string.IsNullOrEmpty` check in there, as well as return a 'default' metadata on error

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+public class DeveloperMetadata
+{
+	public List<ModVersionHash> modVersionHashes { get; set; } = new List<ModVersionHash>();
+
+	[JsonConstructor]
+	private DeveloperMetadata() { }
+
+	internal DeveloperMetadata(string workshopPath, bool useWebApi, ModDownloadItem modDownloadItemAsFound = null)
```
Chicken-Bones: Similar to my below comment, this would be better as a factory method. It 'does too much work' to be considered a constructor. At a start, I'd consider making `CalculateModHashes` return a `DeveloperMetadata`

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	internal static DeveloperMetadata DeSerialize(string serializedDevMetadata)
+	{
+		return JsonConvert.DeserializeObject<DeveloperMetadata>(serializedDevMetadata, new ModVersionHash.VersionHashConverter());
+	}
+
+	// PR 4345 - We combine the hash data that is currently on workshop with the hash data from the updated publishing folder to ensure that when mods are updated it is backwards compatible
+	// It is backwards compatible while Steam spends up to an hour rolling out workshop item updates
+	private void CalculateModHashes(string workshopPath, bool useWebApi, ModDownloadItem modDownloadItemAsFound = null)
```
Chicken-Bones: In addition, you have an invariant where `useWebApi` is `true` if and only if `modDownloadItemAsFound == null`, so as a start, remove the `useWebApi` and document that if no `modDownloadItemAsFound ` is provided, the item will be obtained from the workshop.

Better yet, this function seems to be taking on too much responsibility, and in the principle of making parameters to functions as narrow and minimal as possible, instead of passing in an entire `ModDownloadItem` we should just pass in a `DeveloperMetadata` and leave the job of getting the metadata from workshop or mod download item to the call sites (of which there is only one of each)

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+			currentHashes.Add(new ModVersionHash(tModFile));
+		}
+
+		// Combine the hashes
+		modVersionHashes = currentHashes.Concat(prevHashes.Except(currentHashes).ToList()).ToList();
+	}
+
+	// This methods trims contents of developer metadata based on the preferred order of discarding information.
+	// It is primarily written with the intent of 'in case' we need to store other information in this Workshop text field
+	private void TrimDevMetadata()
```
Chicken-Bones: seems my previous comment got lost. Instead of this complicated logic, I recommend just running `GetSerialize()` in a `while` loop and dropping the last hash on length overflow. This could run at the end of `CalculateModHashes`

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+			modVersionHashes = new List<ModVersionHash>();
+		}
+	}
+
+	internal string GetSerialize()
+	{
+		return JsonConvert.SerializeObject(this, Formatting.None, new ModVersionHash.VersionHashConverter());
+	}
+
+	internal static DeveloperMetadata DeSerialize(string serializedDevMetadata)
```
Chicken-Bones: `Deserialize` is one word, as seen in the json call below

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -0,0 +1,146 @@
...
+			}
+
+			modVersionHashes = devMetadata.modVersionHashes;
+		}
+		catch (Exception) {
+			modVersionHashes = new List<ModVersionHash>();
+		}
+	}
+
+	internal string GetSerialize()
```
Chicken-Bones: just `Serialize` is good

patches/tModLoader/Terraria/Social/Steam/WorkshopBrowserModule.cs
```diff
@@ -92,6 +92,8 @@ public bool DoesItemNeedUpdate(ModPubId_t modId, LocalMod installed, System.Vers
 	public void DownloadItem(ModDownloadItem item, IDownloadProgress uiProgress)
 	{
 		item.UpdateInstallState();
+		if (item.Banned)
+			throw new BannedModException($"Attempted to Download a Banned Mod {item.DisplayName} with ID {item.PublishId}. Aborting...", item.DisplayName, item.PublishId.ToString());
```
Chicken-Bones, line 96: Good place for this guard. Tested that the exception shows a UI. Found that `PublishId` was missing a `ToString` in the process

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -19,28 +19,34 @@ public partial class WorkshopSocialModule
 	public override List<string> GetListOfMods() => _downloader.ModPaths;
 	private ulong currPublishID = 0;
 
+	private ModDownloadItem modDownloadItemAsFound;
```
Chicken-Bones: Scary! Please store `DevMetadata` on `FoundWorkshopEntryInfo` or add it as another `out` param in `TryGetInfoForMod`. Then pass it through to `PublishMod` as a field in `WorkshopItemPublishSettings` (such as `ExistingMetadata`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Core/LocalMod.cs
```diff
@@ -40,4 +40,11 @@ public LocalMod(ModLocation location, TmodFile modFile, BuildProperties properti
 	public LocalMod(ModLocation location, TmodFile modFile) : this(location, modFile, BuildProperties.ReadModFile(modFile))
 	{
 	}
+
+	internal static LocalMod OpenModFile(string path)
```
Chicken-Bones: call it `FromWorkshopModFile` and we're good

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -727,6 +727,19 @@ internal static void CleanupOldPublish(string repo)
 		return information;
 	}
 
+	internal static List<ModVersionHash> GetModHashesFromFolder(string folderPath)
```
Chicken-Bones: I think this was better in `WorkshopSocialModule`, I don't see this function being relevant for the general operation of mod loading and launching

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -44,84 +44,37 @@ public class DeveloperMetadata
...
 	{
-		// Try-Catch is for error correction for any bizarre dev metadata put on items previously
+		if (string.IsNullOrWhiteSpace(serializedDevMetadata))
+			return new();
+
 		try {
-			var devMetadata = DeSerialize(serializedDevMetadata);
+			var devMetadata = JsonConvert.DeserializeObject<DeveloperMetadata>(serializedDevMetadata, new ModVersionHash.VersionHashConverter());
 
 			if (devMetadata == null || devMetadata.modVersionHashes == null) {
```
Chicken-Bones: do we really need this guard? I'm not sure how it would be possible to hit this condition.

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -44,84 +44,37 @@ public class DeveloperMetadata
...
-		List<ModVersionHash> prevHashes = new List<ModVersionHash>();
-
-		// Get the old / existing hashes on Workshop
-		if (useWebApi) {
-			var itemDetails = SteamWebWrapper.GetItemMetadata(Path.GetFileNameWithoutExtension(workshopPath));
-
-			prevHashes = string.IsNullOrEmpty(itemDetails.Metadata) ?
-				new List<ModVersionHash>() :
-				new DeveloperMetadata(itemDetails.Metadata).modVersionHashes;
+			return default;
```
Chicken-Bones: Shouldn't we return `new()` like above?

patches/tModLoader/Terraria/Social/Steam/WorkshopBrowserModule.cs
```diff
@@ -160,5 +160,31 @@ public List<ModDownloadItem> DirectQueryItems(QueryParameters queryParams, out L
...
+			return DeveloperMetadata.Deserialize(itemDetails.Metadata);
+		}
+
+		// Mod should Exist, check Mod Browser
+		var items = DirectQueryItems(new QueryParameters() { searchModIds = [modId], queryType = QueryType.SearchDirect, returnDevMetadata = true }, out _);
+
+		if (!items.Any())
+			return new();
+
+		return items.First().DevMetadata;
```
Chicken-Bones: Can do `items.FirstOrDefault()?.DevMetadata ?? new()` if you feel like being fancy

patches/tModLoader/Terraria/Social/Steam/WorkshopSocialModule.TML.cs
```diff
@@ -247,11 +250,32 @@ internal static HashSet<string> DetermineSupportedVersionsFromWorkshop(string re
...
+		}
+
+		var pubId = new ModPubId_t() { m_ModPubId = publishId.ToString() };
+		var oldDevMetadata = WorkshopBrowserModule.Instance.GetDeveloperMetadataFromModBrowser(pubId);
+
+		newDeveloperMetadata.modVersionHashes = currentHashes.Concat(oldDevMetadata.modVersionHashes.Except(currentHashes).ToList()).ToList();
+
+		newDeveloperMetadata.TrimDevMetadataForPublish();
+
+		return newDeveloperMetadata;
```
Chicken-Bones: This code will need maintenance if we ever add anything new to developer metadata that isn't recalculated each time in this function. I think it would be better if we just did something like

```cs
var developerMetadata = WorkshopBrowserModule.Instance.GetDeveloperMetadataFromModBrowser(pubId);
developerMetadata.modVersionHashes = currentHashes.Concat(developerMetadata.modVersionHashes.Except(currentHashes).ToList()).ToList();
developerMetadata .TrimDevMetadataForPublish();
return developerMetadata;
```

I think `GetDeveloperMetadataFromModBrowser` will return an empty metadata for 0 publishid, but if not, we can `: new()` it

Chicken-Bones (changes requested):
Looking pretty good

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
Solxanich: Yeah, that's an artifact of System.Text.Json which return null instead of throwing; removed.

patches/tModLoader/Terraria/Social/Base/DeveloperMetadata.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.Intrinsics.Arm;
```
Chicken-Bones: Remove this line. Approved

Issue #3250, opened 2023-02-12
Modded prefixes should persist through mod unload


ScalarVector1:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Gameplay as a Player

### Description

I am proposing a placeholder prefix, similar to the placeholder items and tiles in TModLoader so that modded prefixes can persist on items through reloading, similar to how modded items or tiles do. 

I can see 3 ways in which this is presented to the end user:

1) The item keeps the name of the modded prefix, but the prefix tooltip is changed to something along the lines of 'no effect, re-enable [Mod that the prefix came from] to use this prefix'

2) The item's prefix is displayed as "Unloaded prefix", and has no tooltip

3) The item appears to the player as if it has no prefix, but will re-gain its previous prefix if the mod adding it is loaded again

### What does this proposal attempt to solve or improve?

Currently, if a player reforges an item and gets a prefix from a mod, then proceeds to unload the mod that gave the prefix but not the item, and load their player, the prefix will disappear. This can be irritating for players who invested resources into obtaining a specific prefix only to have it deleted by the game.

Chicken-Bones:
Given that every `Item` has an `UnloadedGlobalItem` component (regardless of if the item has any unloaded data on it).
1. Make an `UnloadedPrefix : ModPrefix` in tML, localize it.
2. Add an `unloadedPrefix` field to `UnloadedGlobalItem`, add to `SaveData`
3. Only save the `unloadedPrefix`, if `Item.prefix` is `UnloadedPrefix` (so the data gets ignored if the item changes prefix).
4. Adjust `ItemIO.Load` to set the `unloadedPrefix` field when the `ModContent.TryFind` call fails on `"modPrefix..."`
5. Move the above code into an internal `ItemIO.LoadModdedPrefix` method.
6. Adjust `UnloadedGlobalItem.LoadData` to call `ItemIO.LoadModdedPrefix` if there's an `unloadedPrefix`.
7. (Optional) use `GlobalItem.ModifyTooltips` to show details about the unloaded prefix

direwolf420:
I would prefer if the prefix would not actually have a name, and only displayed some text (in green or red like the usual prefix colors for stats) that it's unloaded and that its actual prefix is "modname/prefixname". (For the purposes of `Lang.prefix` use by mods it should still have one, it just shouldn't be displayed in the relevant gameplay locations imo)

[...]

PR #4349, opened 2024-08-16
Preserve prefixes for unloaded items and unloaded prefixes

Tyfyter:
The benefits of this addition are listed in #3250, the details of its implementation are roughly as described in [this comment](https://github.com/tModLoader/tModLoader/issues/3250#issuecomment-1428491828_) by @Chicken-Bones

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
@@ -10,9 +10,9 @@ public class UnloadedGlobalItem : GlobalItem
 	internal IList<TagCompound> data = new List<TagCompound>();
 
 	[CloneByReference]
-	internal string modPrefixMod = null;
+	public string ModPrefixMod { get; internal set; } = null;
 	[CloneByReference]
-	internal string modPrefixName = null;
+	public string ModPrefixName { get; internal set; } = null;
 
 	public override bool InstancePerEntity => true;
 
--- a/patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
@@ -9,9 +9,9 @@ public class UnloadedGlobalItem : GlobalItem
 	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
 	internal IList<TagCompound> data = new List<TagCompound>();
 
-	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
+	[CloneByReference]
 	internal string modPrefixMod = null;
-	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
+	[CloneByReference]
 	internal string modPrefixName = null;
 
 	public override bool InstancePerEntity => true;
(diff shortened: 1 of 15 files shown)


Tyfyter:
Concerns that a displayed name would cause "item names [to] become pretty silly" were not taken into account as the expected course of action is to remove the prefix, either by re-enabling the unloaded mod or reforging the item, and other unloaded content looks similarly out of place.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
```diff
@@ -9,6 +9,11 @@ public class UnloadedGlobalItem : GlobalItem
 	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
 	internal IList<TagCompound> data = new List<TagCompound>();
 
+	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
+	internal string modPrefixMod = null;
+	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
+	internal string modPrefixName = null;
```
Chicken-Bones: No need for comment. `string` is immutable.

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPrefix.cs
```diff
@@ -0,0 +1,16 @@
...
+public sealed class UnloadedPrefix : ModPrefix
+{
+	public override LocalizedText DisplayName => Language.GetText("Mods.ModLoader.Unloaded");
+	public override IEnumerable<TooltipLine> GetTooltipLines(Item item)
+	{
+		UnloadedGlobalItem unloadedGlobalItem = item.GetGlobalItem<UnloadedGlobalItem>();
+		yield return new TooltipLine("UnloadedPrefix", $"{unloadedGlobalItem.modPrefixMod}/{unloadedGlobalItem.modPrefixName}") {
+			IsModifier = true,
+			OverrideColor = new(215, 123, 186)
+		};
```
Chicken-Bones: Please add screenshots of some unloaded prefixed item tooltips to the PR description.
Also include one of an unloaded item with an unloaded prefix

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Default/UnloadedPrefix.cs
Tyfyter: At this time, if unloaded items retain prefixes they still do not support actually having them — which is necessary for them to be displayed — an image of a loaded item with an unloaded prefix has been attached.

JavidPack:
Code was added to preserve prefixes on UnloadedItem. Previously even vanilla prefixes would be lost on unloaded items.

Chicken-Bones (approved):

patches/tModLoader/Terraria/ModLoader/Default/UnloadedGlobalItem.cs
```diff
@@ -9,6 +9,11 @@ public class UnloadedGlobalItem : GlobalItem
 	[CloneByReference] // safe to share between clones, because it cannot be changed after creation/load
 	internal IList<TagCompound> data = new List<TagCompound>();
 
+	[CloneByReference]
+	internal string modPrefixMod = null;
+	[CloneByReference]
+	internal string modPrefixName = null;
```
direwolf420:
```suggestion
	[field: CloneByReference]
	public string ModPrefixMod { get; internal set; } = null;
	[field: CloneByReference]
	public string ModPrefixName { get; internal set; } = null;
```
To follow the same design as UnloadedItem.ModName/ItemName. There are use cases for this to be accessible by other mods (including one of my mods). Obviously the case change should be applied to all places where it's used.

JavidPack: Done.

PR #4352, opened 2024-08-20
More BitsByte helper methods for reading and writing

JavidPack:
It has often been brought up that using `BitsByte` or `BinaryWriter` to send and receive a number of booleans, as is common in modding, could be more modder friendly. The most recent request for this happened here: https://discord.com/channels/103110554649894912/534215632795729922/1273844915310432287

This PR adds methods to streamline working with `BitsByte` to make sending and receiving bools easier to implement and easier to parse. 

# Issue
## Reading
```cs
BitsByte flags = reader.ReadByte();
flags.Retrieve(
	ref unlockedExamplePersonSpawn,
	ref savedExamplePerson
);

BitsByte flags2 = reader.ReadByte();
// etc
```
The current approach shown in ExampleMod stores a byte as a BitsByte. This implicit conversion can be a bit confusing if a modder doesn't understand implicit conversions. It also necessitates reusing local variable `flags` or making `flags2` for subsequent data. Reusing local variables is something many modders struggle with, as we've seen with recipes.

```cs
((BitsByte)reader.ReadByte()).Retrieve(
	ref unlockedExamplePersonSpawn,
	ref savedExamplePerson
);
```
This avoids the intermediary local variable, but is ugly and the syntax is error prone.

## Writing
```cs
BitsByte flags = new BitsByte();
flags[0] = boughtExampleTownPet;
writer.Write(flags);
```
This is the typical approach, this is a bit verbose.

```cs
writer.Write(new BitsByte(
	unlockedExamplePersonSpawn
));
```
This is a newer approach used most recently in the TownNPCRespawnSystem PR, but can potentially be confusing with the syntax.

# Solutions
## Reading
**Option Read 1 (ReadBitsByte)**
```cs
reader.ReadBitsByte().Retrieve(
	ref unlockedExamplePersonSpawn,
	ref savedExamplePerson
);
```
A `ReadBitsByte` extension method simplifies situations where the implicit conversion would be ugly. This might be useful if modders have other methods using `BitsByte` directly that they want to pass into, but the 2nd option below is better for all other cases.

**Option Read 2 (ReadFlags), Recommended**
```cs
reader.ReadFlags(
	out unlockedExamplePersonSpawn,
	out savedExamplePerson
);
```
A `ReadFlags` extension method is even cleaner and easy to understand. This is the approach that is being used in ExampleMod and learning resources.

**Option Read 3 (Deconstruct)**

```cs
(unlockedExamplePersonSpawn, savedExamplePerson) = reader.ReadBitsByte();
// or...
(unlockedExamplePersonSpawn, savedExamplePerson) = (BitsByte)reader.ReadByte();
```
Using a deconstruction method approach results in clean code and is easy to comprehend at a glance if the modder is familiar with the syntax. The only issue here is that this is more modern syntax, that might be new to existing programmers. This approach does not work well for reading a single bool, however, so it will not be used in learning materials.

## Writing
For writing adding a `BinaryWriter.WriteFlags/Write` extension method works and feels extremely natural. There are no proposed alternatives for this.
```cs
writer.WriteFlags(
	unlockedExamplePersonSpawn,
	savedExamplePerson
);
```


JavidPack:
Issue: Deconstructing a single bool is a bit confusing, since it doesn't work as expected if following the pattern:
image
```
(unlockedExamplePersonSpawn, _) = reader.ReadBitsByte();
or
unlockedExamplePersonSpawn = reader.ReadBool();
```
this works, but consistent syntax is a good idea and this potential pitfall is frustrating.

Chicken-Bones:
I recommend:
- `out` instead of `ref`
- `Retrieve` -> `ToFlags`
- `ReadBits` -> `ReadFlags`
- Add a corresponding `WriteFlags`

If the single bool deconstruct works without brackets, I'm happy for deconstruction to be an option.
I think the approach we should teach in EM is `Read/WriteFlags`

ExampleMod/Common/Systems/TownNPCRespawnSystem.cs
```diff
@@ -33,13 +33,11 @@ public override void LoadWorldData(TagCompound tag) {
...
 
 		public override void NetSend(BinaryWriter writer) {
-			writer.Write(
-				unlockedExamplePersonSpawn
-			);
+			writer.WriteFlags(unlockedExamplePersonSpawn);
 		}
 
 		public override void NetReceive(BinaryReader reader) {
-			(unlockedExamplePersonSpawn, savedExamplePerson) = reader.ReadBitsByte();
```
Chicken-Bones: Can you check ExampleMod for any other usages of `ReadBoolean`? We may as well teach `ReadFlags` as a complete replacement, because it has the same performance as `ReadBoolean`, but extends bettter to multiple bools

patches/tModLoader/Terraria/CommonDocs.xml
```diff
@@ -27,4 +27,19 @@
 		<para/> Lighting is tracked at Tile Coordinate granularity (<see href="https://github.com/tModLoader/tModLoader/wiki/Coordinates">Coordinates wiki page</see>). Typically the center of an entity, converted to tile coordinates, is used to query the lighting color at the entity's location. That color is then used to draw the entity.
 		<para/> Lighting values in-between tile coordinates can be interpolated using <see cref="GetSubLight(Vector2)"/>, but usually the lighting values at the tile coordinates closest to the center of an entity are sufficient. There are many more GetColorX methods for more advanced situations, if needed. If drawing a really large sprite, one might consider splitting up the drawing to allow sections of the sprite to be drawn at different light values.
 	</LightingGetColor>
+	<BitArrayUsage>
+		<para/> Useful for sending/receiving large amounts of flags/bools. If the data length is unknown to the receiver, length would need to be sent as well.
```
Chicken-Bones: `s/amounts/numbers` or `s/large amounts/a large number`

patches/tModLoader/Terraria/CommonDocs.xml
```diff
@@ -27,4 +27,19 @@
...
+		This example sends a bool array of length <c>ItemLoader.ItemCount</c>, which will be consistent, so length is not sent.
+		<code>
+// Field
+bool[] itemsFound = new bool[ItemLoader.ItemCount];
+		
+// NetSend
+Utils.SendBitArray(new (itemsFound), writer);
```
Chicken-Bones, line 38: If we're going to promote these, we should make overloads which accept `bool[]` as the input and output. Including ones that don't alloc a new `bool[]` or `BitArray`.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/CommonDocs.xml
JavidPack: Deferred for now, as suggested.

PR #4353, opened 2024-08-21
Setup CLI

alueck:
### What is the new feature?
A platform independent `Setup.CLI` project. I created a new `Setup.Core` project which contains all the core features of the setup tool. `Setup.CLI` and `Setup.GUI` are built on top of this. This PR also resolves issues with decompilation on Linux/MacOS systems thanks to @Chicken-Bones. The CI build now also uses the setup CLI.

So people can also contribute to tModLoader if working with Linux or MacOS systems.

Execute `./setup-cli.bat` or `sh ./setup-cli.sh` from the root folder of the repository.


setup.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Building Setup.CLI.csproj"
+dotnet build setup/Setup.CLI/Setup.CLI.csproj -c Release --output "setup/bin/Release/net8.0"
+
+if command -v ln &> /dev/null
+then
+    ln -sf ./setup/bin/Release/net8.0/Setup.CLI ./Setup.CLI
+    echo "Created symbolic link for CLI. You can now use ./Setup.CLI"
+else
+    echo "Successfully built CLI. You can now use ./setup/bin/Release/net8.0/Setup.CLI"
+fi
```
Chicken-Bones: The `setup.sh` command shouldn't make a symlink. Instead, it should pass all the args to `setup.sh` through to the built `Setup.CLI` executable. 

We want to make sure that people have up to date submodules and setup project whenever they run it. Having a permanent link to a potentially stale `Setup.CLI` circumvents this.

setup/.gitignore
```diff
@@ -0,0 +1 @@
+appsettings.json
```
Chicken-Bones: I like that you have combined the settings for both GUI and CLI, I was going to suggest this.
I don't like that it's just called `appsettings.json` in the root. Perhaps `setup.user` or `setup.usersettings`

I'd also like to include a migration of the old settings to the new format when the UI project is launched.

setup.sh
Chicken-Bones: It would also be good to add a `setup_cli.bat` for windows users

setup/setup.Designer.csproj
```diff
@@ -1,16 +0,0 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
```
Chicken-Bones: Removing this means we can't use the VS UI designer to adjust the settings form

setup/Setup.CLI/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+CoconaApp app = builder.Build();
+
+app.AddCommands<SetupCommands>();
+app.AddCommands<DecompileCommand>();
+app.AddSubCommand("patch", x => x.AddCommands<PatchSubCommands>()).WithDescription("Applies patches.");
+app.AddSubCommand("diff", x => x.AddCommands<DiffSubCommands>()).WithDescription("Diffs the source to recalculate patches.");
+app.AddCommands<RegenSourceCommand>();
+app.AddCommands<OtherCommands>();
+
+await app.RunAsync();
```
Chicken-Bones: The app looks really good. Some things I noticed:
- The help text for the commands is too long. It gets line wrapped on the default console with, and the line wrapping makes it hard to read because the rest of the line isn't indented.
- Typing `Setup.CLI patch` still shows `Usage: Setup.CLI [command]` rather than `Usage: Setup.CLI patch [command]`
- The help text includes the line `Setup.CLI` after the `Usage:` and before `Commands:` which just looks a bit ugly and unhelpful, can we remove it?

Can we give it a name like `setup_cli` perhaps? Since it'll be invoked via `setup_cli.bat` or `setup_cli.sh` anyway.

setup/Setup.CLI/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "Setup.CLI": {
+      "commandName": "Project",
+      "workingDirectory": "$(ProjectDir)/../.."
+				}
```
Chicken-Bones: Formatter

setup/Setup.CLI/LiveConsoleProgress.cs
```diff
@@ -0,0 +1,41 @@
...
+
+		if (!string.IsNullOrEmpty(value.Status)) {
+			table.Rows.Update(0, 0, new Text(value.Status));
+			update = true;
+		}
+
+		if (update) {
+			ctx.Refresh();
+		}
+	}
```
Chicken-Bones: Fantastic work. It looks really good. Even has `Ctrl+C` support.

It would be good if the `setup` task could output some lines saying that each phase has been completed. The final console state also ends up looking a bit silly. 

```
Information Ready for Setup
Any changes in /src will be lost.

Progress: 2297/2297
Old Src Removed











Done
```

setup/Setup.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Terraria.ModLoader.Setup.Core;
+
+[ExcludeFromCodeCoverage]
+public static class ServiceCollectionExtensions
```
Chicken-Bones: I understand that the Cocona library you selected for the CLI relies on dependency injection, and thus you migrated the GUI project to use it as well.

This may require further discussion on Discord, but I'm not 'all in' on dependency injection frameworks like these. I'd like to keep their usage minimal, and stick to using regular C# to construct objects where possible.

Can we replace all the factory delegates with regular `new XTask(services.Get<>(), ...)` calls? Even ignoring the expression building that enables it, the need to make and store `.Factory` delegates is a fair bit of additional boilerplate and indirection

setup/Setup.CLI/Commands/SetupCommands.cs
```diff
@@ -0,0 +1,77 @@
...
+	private SetupOperation GetSetupOperation(SetupParameters parameters, bool skipDecompileIfExists = false, bool disableParallelDecompilation = false)
+	{
+		if (skipDecompileIfExists && Directory.Exists("src/decompiled")) {
+			Console.WriteLine("src/decompiled found. Skipping decompilation...");
+			return regenSourceTaskFactory(parameters.NoPrompts);
+		}
+
+		return setupTaskFactory(
+			DecompileTaskParameters.CreateDefault(noPrompts: parameters.NoPrompts, maxDegreeOfParallelism: disableParallelDecompilation ? 1 : null),
+			parameters.NoPrompts);
```
Chicken-Bones: Is there a way we can configure `NoPrompts` at a global level? Passing it through the constructors is a bit of a pain.

setup/Setup.Core/TerrariaExecutableSetter.cs
```diff
@@ -0,0 +1,110 @@
...
+
+	public async Task FindAndSetTerrariaDirectoryIfNecessary(bool noPrompts = false, CancellationToken cancellationToken = default)
+	{
+		if (Directory.Exists(programSettings.TerrariaSteamDir)) {
+			return;
+		}
+
+		if (noPrompts) {
+			Console.WriteLine($"Critical failure. Terraria steam directory '{this.programSettings.TerrariaSteamDir}' does not exist.");
+			Environment.Exit(1);
```
Chicken-Bones: Prefer to throw exceptions and let the top level handler print it out.

setup/Setup.CLI/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,76 @@
...
+				: new Text("Done", new Style(foreground: Color.Green, decoration: Decoration.Bold));
+
+			AnsiConsole.Write(text);
+		}
+		catch (OperationCanceledException) { }
+		catch (Exception exception) when (exception is not OperationCanceledException) {
+			AnsiConsole.WriteException(exception, ExceptionFormats.NoStackTrace);
+
+			SetupOperation.CreateDirectory(ProgramSettings.LogsDir);
+			await File.WriteAllTextAsync(errorLogFile, exception.ToString(), cancellationToken);
```
Chicken-Bones: Need to print a message which directs people to log file path. (We should do this in the GUI message box too)

setup/Setup.Core/PatchTask.cs
```diff
@@ -0,0 +1,198 @@
...
+	public override void FinishedPrompt()
+	{
+		if (fuzzy > 0) {
+			return;
+		}
+
+		this.userPrompt.Prompt(
+			"Patch Results",
+			$"Patches applied with {failures} failures and {warnings} warnings.\nSee /logs/patch.log for details",
+			Failed() ? PromptSeverity.Error : PromptSeverity.Warning);
```
Chicken-Bones: Perhaps this should be called `Inform` rather than `Prompt`? Just so it's clear that it still runs even when `NoPrompts` is set

setup/Setup.CLI/Commands/PatchSubCommands.cs
```diff
@@ -0,0 +1,62 @@
...
+
+		return patchTaskFactory(parameters);
+	}
+}
+
+public record PatchParameters : BaseParameters
+{
+	[Option('m', Description = "Set the patch mode. Default is EXACT. If you change the patch mode, it will be saved and can be omitted for further calls.")]
+	[HasDefaultValue]
+	public Patcher.Mode? PatchMode { get; set; }
```
Chicken-Bones: I know the default is for this flag to be sticky in GUI, but it's likely that CLI users will not read this explanation in the help text, and that having an argument to a command be 'sticky' will contradict their intuitions. 

If we do want a way to apply this setting permanently via CLI, I recommend either a separate command, or a separate argument (eg `--set-patch-mode`)

setup/Setup.CLI/Commands/SetupCommands.cs
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		if (!string.IsNullOrWhiteSpace(parameters.TMLDevSteamDir)) {
+			this.ProgramSettings.TMLDevSteamDir = parameters.TMLDevSteamDir;
+			if (!Directory.Exists(parameters.TMLDevSteamDir)) {
+				Directory.CreateDirectory(parameters.TMLDevSteamDir);
+			}
+		}
+
+		ProgramSettings.Save();
```
Chicken-Bones: Is it wise to save settings when using `setupAuto`? I know CI pipelines normally clean the entire workspace and do a fresh checkout each time, but for reliable testing I think that `setupAuto` shouldn't read from or write to the settings file at all. It should rely entirely on arguments

Chicken-Bones (changes requested):
Excellent work, it's a large task so it's great to see you've really taken it on fully

setup.sh
alueck: Actually had the same thought. Done. Does not feel as snappy though, because `dotnet build` always runs first.

setup/Setup.CLI/Program.cs
alueck: - I shortened all descriptions
- Seems like a bug in Cocona's help generation. This would work if I did not use the class based style. Will take a look.
- I actually prefer to see the `Setup.CLI` (now `setup-cli`) after `Usage:`. Before `Commands` you would normally see `patch`, so the same problem like before.

setup/Setup.CLI/LiveConsoleProgress.cs
alueck: Yeah, that's a bit of a problem with the `Live` control. It also worked a bit weird in a case where an error message was printed through `IUserPrompt`. The message basically started on the first line of the table that is used for the live display and overlapped it. I can see if I find any way to do something about this.
Still better than printing thousands of lines of text, I think 😄

setup/Setup.Core/PatchTask.cs
alueck: Renamed. I struggled a bit with the naming there.

setup/Setup.CLI/Commands/PatchSubCommands.cs
alueck: Changed. The save calls were more or less based on a previous state of the application where I always had to save the configuration.

setup/Setup.CLI/Commands/BaseCommand.cs
alueck: Added to both

setup/Setup.Core/TerrariaExecutableSetter.cs
alueck: Done

setup/Setup.CLI/Commands/SetupCommands.cs
alueck: I put it into `ProgramSettings`. It kind of fits there, but will not be persisted.

setup/.gitignore
alueck: File is named `user.settings` now and settings are migrated.

setup/Setup.CLI/Program.cs
alueck: I switched the CLI framework from Cocona to Spectre.Console.CLI. Output now looks a bit different.

setup/setup.Designer.csproj
alueck: Re-added as Setup.GUI.Designer.csproj

setup/Setup.CLI/LiveConsoleProgress.cs
alueck: I fully refactored the progress handling. Works alot better now. Maybe give it a try and see if I missed anything.

setup/GUI/MainForm.resx
```diff
@@ -60,6 +120,9 @@
   <metadata name="toolTipButtons.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
     <value>17, 17</value>
   </metadata>
+  <metadata name="toolTipButtons.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+    <value>17, 17</value>
+  </metadata>
```
Chicken-Bones: Duplicated property somehow? Does this file need changing at all?

setup/GUI/MainForm.cs
```diff
@@ -1,48 +1,61 @@
...
+			taskButtons[buttonRegenSource] = () => new RegenSourceTask(serviceProvider);
 
-			taskButtons[buttonSetup] = () =>
-				new SetupTask(this, new[] { buttonDecompile, buttonRegenSource }
-					.Select(b => taskButtons[b]()).ToArray());
+			taskButtons[buttonSetup] = () => new SetupTask(DecompileTaskParameters.CreateDefault(), serviceProvider);
 
-			SetPatchMode(Settings.Default.PatchMode);
-			formatDecompiledOutputToolStripMenuItem.Checked = Settings.Default.FormatAfterDecompiling;
+			SetPatchMode(this.programSettings.PatchMode);
```
Chicken-Bones, line 59: General request to remove redundant `this.` references across the codebase

setup/Core/ProgramSetting.cs
```diff
@@ -1,22 +1,39 @@
...
+			this.programSettings = programSettings;
 		}
 
-		public void Set(T value) {
-			Settings.Default[key] = value;
-			Settings.Default.Save();
+		public void Set(T value)
+		{
+			setter(programSettings, value);
+			programSettings.Save();
```
Chicken-Bones: Might be nice to only save on change.

setup/GUI/MainForm.cs
```diff
@@ -122,29 +110,30 @@ private void buttonTask_Click(object sender, EventArgs e)
...
 			cancelSource = new CancellationTokenSource();
 			foreach (var b in taskButtons.Keys) b.Enabled = false;
 			buttonCancel.Enabled = true;
 
-			new Thread(() => RunTaskThread(task)).Start();
+			_ = RunTaskThread(task);
 		}
 
-		private void RunTaskThread(SetupOperation task)
+		private async Task RunTaskThread(SetupOperation task)
```
Chicken-Bones: No point returning `Task` if you're not going to await it and it can't throw. `async void` is fine

setup/GUI/MainForm.cs
```diff
@@ -55,64 +68,39 @@ public MainForm()
...
-
 		private void buttonCancel_Click(object sender, EventArgs e)
 		{
 			cancelSource.Cancel();
 		}
 
 		private void menuItemTerraria_Click(object sender, EventArgs e)
 		{
-			SelectAndSetTerrariaDirectoryDialog();
+			terrariaExecutableSetter.SelectAndSetTerrariaDirectory();
```
Chicken-Bones: Make method `async` and `await` this one to silence the code warning

setup/Core/UpdateLocalizationFilesTask.cs
```diff
@@ -1,38 +1,33 @@
...
+		Process? p = Process.Start(new ProcessStartInfo {
 			FileName = "python",
 			Arguments = Path.GetFileName(UpdateLocalizationFilesPath),
-			WorkingDirectory = new FileInfo(UpdateLocalizationFilesPath).Directory.FullName,
+			WorkingDirectory = new FileInfo(UpdateLocalizationFilesPath).Directory!.FullName,
 		});
-		p.WaitForExit();
-		// MessageBox.Show("Success. Make sure you diff tModLoader after this");
+
+		if (p != null) {
```
Chicken-Bones: Should throw if null

setup/Core/UpdateLocalizationFilesTask.cs
```diff
@@ -1,38 +1,33 @@
...
 
-		Process p = Process.Start(new ProcessStartInfo {
+		Process? p = Process.Start(new ProcessStartInfo {
 			FileName = "python",
 			Arguments = Path.GetFileName(UpdateLocalizationFilesPath),
-			WorkingDirectory = new FileInfo(UpdateLocalizationFilesPath).Directory.FullName,
+			WorkingDirectory = new FileInfo(UpdateLocalizationFilesPath).Directory!.FullName,
 		});
-		p.WaitForExit();
-		// MessageBox.Show("Success. Make sure you diff tModLoader after this");
```
Chicken-Bones, line 36: We need to keep this info message/popup. `userPrompt.Inform`?

setup/Core/UpdateLocalizationFilesTask.cs
```diff
@@ -1,38 +1,33 @@
...
+public sealed class UpdateLocalizationFilesTask : SetupOperation
 {
-	private const string UpdateLocalizationFilesPath = $"solutions/UpdateLocalizationFiles.py";
+	private const string UpdateLocalizationFilesPath = "solutions/UpdateLocalizationFiles.py";
 
-	public UpdateLocalizationFilesTask(ITaskInterface taskInterface) : base(taskInterface)
+	public override async Task Run(IProgress progress, CancellationToken cancellationToken = default)
 	{
-	}
+		using var taskProgress = progress.StartTask("Updating localization files...");
```
Chicken-Bones, line 22: This task bugs out the console if run without `--plain-progress`

setup/Core/UpdateLocalizationFilesTask.cs
Chicken-Bones: I see the console output already has `Make sure to run Diff.` but that comes from python. Maybe we should remove the note from the python script. The message here is better.

setup/CLI/Commands/DiffSubCommands.cs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+public sealed class DiffTerrariaCommand(TaskRunner taskRunner, ProgramSettings programSettings)
+	: DiffBaseCommand(taskRunner, programSettings)
+{
+	protected override DiffTaskParameters GetDiffTaskParameters(ProgramSettings programSettings) =>
+		DiffTaskParameters.ForTerraria(programSettings);
+
+	protected override void ResetTimestamp(ProgramSettings programSettings) =>
+		programSettings.TerrariaDiffCutoff = null;
```
Chicken-Bones: I believe you can generify this in the base class via `DiffTaskParameters.Cutoff`

setup/Core/RoslynTask.cs
```diff
@@ -1,86 +1,76 @@
...
-				// todo proper error log
-				workspace.WorkspaceFailed += (o, e) => Console.Error.WriteLine(e.Diagnostic.Message);
+		protected string? ProjectPath { get; private set; }
+		protected abstract string Status { get; }
+		protected virtual int MaxDegreeOfParallelism => 0;
 
-				Console.WriteLine($"Loading project '{projectPath}'");
+		public override async ValueTask<bool> ConfigurationPrompt(CancellationToken cancellationToken = default)
+		{
+			ProjectPath = await projectSelectionPrompt.Prompt(ProjectPath, cancellationToken).ConfigureAwait(false);
```
Chicken-Bones: Would be nice if this could be a CLI arg instead of needing the prompt

setup/CLI/CSharpProjectSelectionPrompt.cs
```diff
@@ -0,0 +1,29 @@
+using Spectre.Console;
+using Terraria.ModLoader.Setup.Core.Abstractions;
+using Terraria.ModLoader.Setup.Core.Utilities;
+
+namespace Terraria.ModLoader.Setup.CLI;
+
+public class CSharpProjectSelectionPrompt : ICSharpProjectSelectionPrompt
```
Chicken-Bones: Maybe we don't need this at all if we have cli args for it?

setup/Core/TerrariaExecutableSetter.cs
```diff
@@ -0,0 +1,109 @@
...
+
+	private async Task FindTerrariaDirectory(CancellationToken cancellationToken = default)
+	{
+		if (!SteamUtils.TryFindTerrariaDirectory(out string? terrariaFolderPath)) {
+			const string messageText = "Unable to automatically find Terraria's installation path. Please select it manually.";
+
+			bool continueSelection = userPrompt.Prompt("Error", messageText, PromptOptions.OKCancel, PromptSeverity.Error);
+			if (continueSelection) {
+				terrariaFolderPath = await this.TrySelectTerrariaDirectory(cancellationToken).ConfigureAwait(false);
+			}
```
Chicken-Bones: This should just be informative. The double prompt feels bad in CLI.

 The user can cancel inside the `TrySelectTerrariaDirectory` prompting process

Chicken-Bones: There should also be a way to supply the terraria steam dir from a CLI arg.

setup/CLI/UserPrompt.cs
```diff
@@ -0,0 +1,49 @@
...
+	public bool Prompt(
+		string caption,
+		string text,
+		PromptOptions options,
+		PromptSeverity severity = PromptSeverity.Information)
+	{
+		this.Inform(caption, text, severity);
+
+		string answer = AnsiConsole.Prompt(
+			new SelectionPrompt<string>().AddChoices(Cancel, options == PromptOptions.OKCancel ? OK : Retry));
```
Chicken-Bones: I think the default selected option for `OK/Cancel` prompts should be `OK`

setup/Core/ProgramSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+		new() { WriteIndented = true, TypeInfoResolver = new DefaultJsonTypeInfoResolver() };
+
+	private readonly string jsonFilePath;
+
+	internal ProgramSettings(string jsonFilePath)
+	{
+		this.jsonFilePath = jsonFilePath;
+	}
+
+	public string LogsDir { get; set; } = Path.Combine("setup", "logs");
```
Chicken-Bones: No need for this to be a setting. Can just be a `static readonly` or something

setup.bat
```diff
@@ -22,23 +22,23 @@ if NOT ["%errorlevel%"]==["0"] (
...
+	dotnet build setup/CLI/Setup.CLI.csproj -c Release --output "setup/bin/Release/net8.0" -p:WarningLevel=0 -v q
 
 	if NOT ["%errorlevel%"]==["0"] (
 		pause
 		exit /b %errorlevel%
 	)
 
-	"setup/bin/Debug/net6.0-windows/setup-auto.exe" %2
+	"setup/bin/Release/net8.0/setup-cli.exe" "setup-auto" %2 "steam_build" "--plain-progress"
 ) Else (
```
Chicken-Bones: Please remove the `auto` code from `setup.bat` and instead adjust the `build.yml` to pass the right args to `setup-cli.bat`

setup/Core/SetupOperation.cs
```diff
@@ -1,194 +1,207 @@
...
+				}
+
+				await Parallel.ForEachAsync(
+					items,
+					new ParallelOptions {
+						MaxDegreeOfParallelism = maxDegreeOfParallelism > 0 ? maxDegreeOfParallelism.Value : Environment.ProcessorCount,
+						CancellationToken = cancellationToken,
+					},
+					async (item, ct) => {
+						ct.ThrowIfCancellationRequested();
```
Chicken-Bones: Do you need this when using `ForEachAsync` with a `CancellationToken`?

setup/Core/Ref.cs
```diff
@@ -1,10 +1,11 @@
-﻿namespace Terraria.ModLoader.Setup
+﻿namespace Terraria.ModLoader.Setup.Core
 {
 	public class Ref<T>
```
Chicken-Bones: May as well make this an inner class of the `SetupOperation`. Alternatively we could probably just make the `List<Ref<string>>` in `ExecuteParallel` into a `List<Func<string>>` and take advantage of closures to simplify things greatly.

setup/Core/Utilities/PathUtils.cs
```diff
@@ -0,0 +1,21 @@
+namespace Terraria.ModLoader.Setup.Core.Utilities
+{
+	public static class PathUtils
+	{
+		public static string GetCrossPlatformFullPath(string path)
```
Chicken-Bones, line 5: Can you explain more about this one to me?

setup/Core/Utilities/PathUtils.cs
```diff
@@ -0,0 +1,21 @@
...
+
+			string result = path;
+			if (path.StartsWith('~')) {
+				result = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), path[2..]);
+			}
+
+			return Path.GetFullPath(result);
+		}
+
+		public static string SetCrossPlatformDirectorySeparators(string path) => path.Replace('\\', '/');
```
Chicken-Bones: Perhaps a better name would be `WithUnixSeparators`, since it's called statically under `PathUtils`

Chicken-Bones: It might also be worth making `PathUtils.Combine` or `UnixCombine` to save a bit of verbosity on teh consumer side where we use `Path.Combine` a lot.

setup/Core/RoslynTask.cs
```diff
@@ -1,86 +1,68 @@
...
-			//if (!RequiresCodeAnalysis)
-			//	return new AdhocWorkspace();
+		public override async Task Run(IProgress progress, CancellationToken cancellationToken = default)
+		{
+			using var taskProgress = GetTaskProgress(progress);
+			using var workspace = CreateWorkspace(taskProgress);
+			// todo proper error log
+			workspace.WorkspaceFailed += (o, e) => Console.Error.WriteLine(e.Diagnostic.Message);
+
+			Console.WriteLine($"Loading project '{ProjectPath}'");
```
Chicken-Bones: Report/Inform instead of `Console`?

setup/Core/PatchTask.cs
```diff
@@ -1,175 +1,196 @@
...
+			}
 
-			taskInterface.SetStatus("Deleting Old Src's Empty Directories");
-
-			DeleteEmptyDirs(patchedDir);
+			taskProgress.ReportStatus("Deleting Old Src's Empty Directories", overwrite: false);
+			DeleteEmptyDirs(parameters.PatchedDir);
 
-			taskInterface.SetStatus("Old Src Removed");
+			taskProgress.ReportStatus("Old Src Removed", overwrite: false);
```
Chicken-Bones: I appreciate the use of named args for boolean constants, but I think it's fine to let the default value do its work here. Other calls in this same file don't pass `overwrite: false`

setup/Core/PatchTask.cs
```diff
@@ -1,175 +1,196 @@
...
-		public override void FinishedDialog() {
-			if (fuzzy > 0)
+		public override void FinishedPrompt()
+		{
+			if (fuzzy > 0 || (failures == 0 && warnings == 0)) {
 				return;
+			}
 
-			MessageBox.Show(
+			this.userPrompt.Inform(
```
Chicken-Bones: Unnecessary `this.` please fix usages in solution

setup/Core/SetupOperation.cs
```diff
@@ -23,45 +18,66 @@ public WorkItem(string status, Worker worker) {
...
 			}
 
-			public WorkItem(string status, Action action) : this(status, _ => action()) { }
+			public WorkItem(string status, Func<CancellationToken, ValueTask> action) : this(status, (_, ct) => action(ct)) { }
+
+			public WorkItem(string status, Action action) : this(status, (_, _) => { action(); return ValueTask.CompletedTask; }) { }
 		}
 
-		protected void ExecuteParallel(List<WorkItem> items, bool resetProgress = true, int maxDegree = 0) {
+		private int currentProgress;
```
Chicken-Bones: This can be a local variable

setup/Core/PatchTask.cs
```diff
@@ -1,175 +1,196 @@
...
 		}
 
-		private FilePatcher Patch(string patchPath)
+		private async Task<FilePatcher> Patch(string patchPath, CancellationToken cancellationToken = default)
 		{
-			var patcher = FilePatcher.FromPatchFile(patchPath);
-			patcher.Patch(mode);
+			FilePatcher? patcher = FilePatcher.FromPatchFile(patchPath);
+			patcher.Patch(this.programSettings.PatchMode);
 			results.Add(patcher);
```
Chicken-Bones: Consider moving outside this method. Allows us to make `results` local to `Run`

setup/Core/PatchTask.cs
```diff
@@ -1,175 +1,196 @@
...
+			log.AppendLine(
+				$"{patcher.patchFile.basePath},\texact: {exact},\toffset: {offset},\tfuzzy: {fuzzy},\tfailed: {failures}");
 
-			foreach (var res in patcher.results)
+			foreach (Patcher.Result? res in patcher.results) {
 				log.AppendLine(res.Summary());
+			}
 
-			Log(log.ToString());
+			if (logFile != null) {
```
Chicken-Bones: assert rather than checking

setup/Core/SetupOperation.cs
```diff
@@ -151,44 +156,33 @@ private static bool DeleteEmptyDirsRecursion(string dir) {
...
+		///     Will prevent successive tasks from executing and cause FinishedDialog to be called
 		/// </summary>
 		/// <returns></returns>
 		public virtual bool Failed() => false;
 
 		/// <summary>
-		/// Will cause FinishedDialog to be called if warnings are not supressed
-		/// </summary>
-		/// <returns></returns>
-		public virtual bool Warnings() => false;
```
Chicken-Bones, line 187: auto setup (for CI) isn't allowed to complete with warnings. Please account for this.

setup/Core/CompositeTask.cs
```diff
@@ -1,40 +1,52 @@
...
 
-		public override void Run() {
+		public override async Task Run(IProgress progress, CancellationToken cancellationToken = default)
+		{
 			foreach (var task in tasks) {
-				task.Run();
+				await task.Run(progress, cancellationToken).ConfigureAwait(false);
 				if (task.Failed()) {
 					failed = task;
-					return;
```
Chicken-Bones, line 35: `return;` removed? Changes semantics. Maybe it never mattered in practice.

See my other notes about replacing the `Failed` checks with exceptions like normal sane people anyway

setup/Core/SetupOperation.cs
```diff
@@ -151,44 +156,33 @@ private static bool DeleteEmptyDirsRecursion(string dir) {
...
-
-		/// <summary>
-		/// Called to display a finished dialog if Failures() || warnings are not supressed and Warnings()
+		///     Called to display a finished dialog if Failures() || warnings are not supressed and Warnings()
 		/// </summary>
-		public virtual void FinishedDialog() {}
+		public virtual void FinishedPrompt() { }
```
Chicken-Bones, line 169: I wonder if we can get rid of `FinishedPrompt` and `Failed` entirely, and just use exceptions and `Inform` in the `Run` methods. 

I can't remember why I had `FinishedPrompt` split out originally. Probably overzealous implementation of `CompositeTask`. I was much greener 9 years ago.

setup/Core/PatchTask.cs
```diff
@@ -1,175 +1,196 @@
...
+			string removedFileList = Path.Combine(parameters.PatchDir, DiffTask.RemovedFileList);
+			HashSet<string> noCopy = File.Exists(removedFileList)
+				? [
+					..(await File.ReadAllLinesAsync(removedFileList, cancellationToken).ConfigureAwait(false)).Select(PathUtils.SetCrossPlatformDirectorySeparators),
+				]
+				: [];
 
 			var items = new List<WorkItem>();
-			var newFiles = new HashSet<string>();
+			var newFiles = new ConcurrentDictionary<string, object?>();
```
Chicken-Bones, line 63: Good spotting.

setup/Core/HookGenTask.cs
```diff
@@ -1,90 +1,94 @@
...
+			foreach (TypeDefinition type in mm.Module.Types) {
+				if (!type.FullName.StartsWith("Terraria") || type.FullName.StartsWith("Terraria.ModLoader")) {
 					continue;
+				}
 
-				gen.GenerateFor(type, out var hookType, out var hookILType);
-				if (hookType == null || hookILType == null || hookType.IsNested)
+				gen.GenerateFor(type, out TypeDefinition hookType, out TypeDefinition hookILType);
+				if (hookType == null || hookILType == null || hookType.IsNested) {
 					continue;
```
Chicken-Bones: Probably just an overzealous autoformatter, but we do allow single line statements in our code style, and prefer `var`. Update the `.editorconfig` if you need to, but otherwise, try and avoid reformatting code that you aren't changing the semantics of as it makes review and `git blame` harder.

I made some diff reduction commits to help me review earlier

We can do a reformat commit after this PR is merged, with file scoped namespaces and proper brace style and that jazz. Refactor commits are easy to skip when blaming when listed as such.

setup/Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+		this IServiceCollection serviceCollection,
+		IConfiguration configuration,
+		string appSettingsPath)
+	{
+		ProgramSettings programSettings = new(appSettingsPath);
+		configuration.Bind(programSettings);
+
+		return serviceCollection
+			.AddTransient<TerrariaExecutableSetter>()
+			.AddTransient<TargetsFilesUpdater>()
```
Chicken-Bones: Is the reasoning for `AddTransient` just, prefer `Transient` where possible? Shorter lived objects?

.gitignore
```diff
@@ -36,4 +36,4 @@ solutions/**/*.tar.gz
 solutions/tModLoader v*
 solutions/Release/
 **/msbuild.binlog
-/steam_build
+/steam_build
```
Chicken-Bones: accidental removal of newline?

Chicken-Bones (changes requested):

setup/Core/UpdateLocalizationFilesTask.cs
alueck: Yeah, if anything writes to stdout not using `ITaskProgress` while the live display is rendered, it will bug out. There is not much I can do about that. Maybe we should just disable live progress for this task.

setup/CLI/Commands/DiffSubCommands.cs
alueck: Yes, thanks. Didn't notice that.

setup/Core/SetupOperation.cs
alueck: Nope, removed.

setup/Core/HookGenTask.cs
alueck: Rider doesn't seem to care about the severity of the code styles in the `.editorconfig`. `var style` and braces are configured as `silent` or `suggestion` but were still applied. I might have to change a setting somewhere.

setup/Core/SetupOperation.cs
alueck: This is actually required because of what I mentioned here: https://github.com/tModLoader/tModLoader/pull/4353#discussion_r1734974285

setup/Core/UpdateLocalizationFilesTask.cs
alueck: Updated.

setup/Core/Utilities/PathUtils.cs
alueck: .NET does not correctly resolve the Home directory shortcut `~` by itself. So I had to manually write code to account for this. Otherwise paths input by the user might not work although the path exists. If you use `Path.GetFullPath("~")` you get the current working directory combined with `~` at the end.

setup/Core/Utilities/PathUtils.cs
alueck: I thought about a `Combine` method. But `Path.Combine` does a bit more than just joining the paths. I added a `UnixJoin` method.

setup/Core/RoslynTask.cs
alueck: Yes, thanks.

setup/Core/PatchTask.cs
alueck: Yeah, don't exaclty know why I added it there.

setup/Core/Extensions/ServiceCollectionExtensions.cs
alueck: I used transient when I added the DI framework because many of the existing classes were stateful and just added all the other classes as transient without later reviewing it again. Changed it to `Singleton` because all of these are stateless.

setup/GUI/MainForm.resx
alueck: I reset the file. Visual Studio just kept changing it.

setup/Core/RoslynTask.cs
alueck: It is now.

setup/CLI/CSharpProjectSelectionPrompt.cs
alueck: Yes, deleted.

setup/Core/Ref.cs
alueck: I refactored the List to be `List<WorkItem` and deleted the `Ref` class.

.gitignore
alueck: Well, I even checked out the .gitignore file from 1.4.4 and pushed it, to get rid of this. It still shows up though.

alueck: Nevermind it was just Github unwilling to update the page correctly...

setup/Core/SetupOperation.cs
Chicken-Bones: Perhaps a --strict CLI arg

setup/CLI/LiveConsoleProgress.cs
```diff
@@ -45,8 +45,8 @@ private sealed class TaskProgress : ITaskProgress
 		private readonly string description;
 		private readonly Table table;
 
-		private readonly int row;
-		private readonly int detailsRow;
+		private readonly int HeaderRow;
+		private readonly int StatusRow;
```
alueck: @Chicken-Bones These field names don't conform to the naming of any other readonly fields in the solution. Should be lowercase.

setup/Core/SetupOperation.cs
alueck: The CLI exits with status code 1 now if the task failed or the strict option is passed to `TaskRunner` and the task has warnings. I did not add an additional CLI arg for this, but just passed `strict: true` to `TaskRunner` in the `SetupAutoCommand`.

setup/CLI/LiveConsoleProgress.cs
alueck: Fixed it.

setup.bat
alueck: Done

setup/CLI/PlainConsoleProgress.cs
```diff
@@ -0,0 +1,42 @@
...
+		public WorkItemProgress(string status)
+		{
+			WriteStatus(status);
+		}
+
+		public void Dispose() { }
+
+		public void ReportStatus(string status) => WriteStatus(status);
+
+		private void WriteStatus(string status) => Console.WriteLine($"  {status}");
```
Chicken-Bones: If you make `WorkItemProgress` an inner class of `TaskProgress`, then you can access `Indent`. That or just remove the `Indent` method from `TaskProgress` and pull `WriteStatus` out to `PlainConsoleProgress`, make it `static` and use it in both classes

.github/workflows/build.yml
```diff
@@ -213,17 +213,36 @@ jobs:
...
+            sed -i "s#<\!-- TML stable version define placeholder -->#<DefineConstants>\$(DefineConstants);$tmlVersionDefine</DefineConstants>#g" ./patches/tModLoader/Terraria/release_extras/tMLMod.targets
+          fi
+
+          if [ ! -d "./src/decompiled" ]; then
+            ./setup-cli.bat decompile --key $TERRARIA_OWNERSHIP_KEY --terraria-steam-dir ./$TERRARIA_VERSION/Windows --tml-dev-steam-dir steam_build --plain-progress --max-parallelism 1 -f --strict
+
+            err=$?
+            if [ $err -ne 0 ]; then
+              exit $err
+            fi
```
Chicken-Bones: Github actions runs the shell with `-e` so there's no need to handle exit codes yourself.

setup/CLI/Commands/SetupAutoCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+	[CommandArgument(0, "<TERRARIA_STEAM_DIR>")]
+	[Description("Path to the Terraria steam directory.")]
+	public string TerrariaSteamDir { get; set; } = string.Empty;
+
+	[CommandArgument(0, "<TML_DEV_STEAM_DIR>")]
+	[Description("Path to the TML dev steam directory.")]
+	public string TMLDevSteamDir { get; set; } = string.Empty;
+}
+
+public sealed class SetupAutoCommand : CancellableAsyncCommand<SetupAutoCommandSettings>
```
Chicken-Bones: Can now be removed

setup/Core/SetupOperation.cs
```diff
@@ -1,96 +1,69 @@
...
-			if (basePath.Last() != Path.DirectorySeparatorChar)
-				basePath += Path.DirectorySeparatorChar;
+			public Worker Worker { get; }
+		}
 
-			if (path+Path.DirectorySeparatorChar == basePath) return "";
+		protected async Task ExecuteParallel(
+			List<WorkItem> items,
+			ITaskProgress progress,
+			bool resetProgress = true,
```
Chicken-Bones: There are no calls with `resetProgress: false`, so we can remove this parameter.

setup/CLI/LiveConsoleProgress.cs
```diff
@@ -0,0 +1,138 @@
...
+				});
+
+			lock (currentWorkItems) {
+				currentWorkItems.Add(progress);
+			}
+
+			UpdateStatusFromWorkItems();
+
+			return progress;
+		}
```
Chicken-Bones: We could probably encapsulate this common code a bit more. The `UpdateStatusFromWorkItems` delegate is the only difference between the `MainForm` and `LiveConsoleProgress` impls

Chicken-Bones (changes requested):

setup/CLI/LiveConsoleProgress.cs
alueck: I added a base class which is now used by `MainForm` and `LiveConsoleProgress`.

setup/CLI/LiveConsoleProgress.cs
```diff
@@ -0,0 +1,102 @@
...
+using Timer = System.Timers.Timer;
+
+namespace Terraria.ModLoader.Setup.CLI;
+
+public sealed class LiveConsoleProgress : IProgress, IDisposable
+{
+	private readonly LiveDisplayContext context;
+	private readonly Table table;
+	private readonly Timer timer;
+	private readonly ConcurrentDictionary<TaskProgress, object?> taskProgresses = [];
```
Chicken-Bones: Since the way we do tables for live console progress doesn't support two simultaneous progress updating tasks, it's probably better if the code doesn't pretend as such. It should be fine for this to just be `TaskProgress?` and the `onCompleted` just sets `taskProgress = null`. We can include an assertion that `taskProgress == null` in the constructor if we want to...

Alternatively, if we move the `timer` and `Refresh` methods into the `TaskProgress` class, we can `Dispose` it all together, and there's no need to track a reference here. (Similar to how the `MainForm` impl works)

setup/Core/TerrariaExecutableSetter.cs
```diff
@@ -0,0 +1,109 @@
...
+
+			terrariaFolderPath = await PromptForTerrariaDirectory(cancellationToken);
+		}
+
+		SetTerrariaDirectory(terrariaFolderPath, tmlDevSteamDirectoryOverride);
+	}
+
+	private void SetTerrariaDirectory(string terrariaSteamDirectory, string? tmlDevSteamDirectoryOverride)
+	{
+		workspaceInfo.UpdatePaths(terrariaSteamDirectory, tmlDevSteamDirectoryOverride ?? workspaceInfo.TMLDevSteamDirectory);
```
Chicken-Bones: This should probably overwrite the tML steam directory each time the Terraria directory is changed. I believe that was the old behavior

alueck: Updated to work that way.

Chicken-Bones (approved):

PR #4361, opened 2024-08-27
Custom error message for building in VS while tModLoader has tmod file open

JavidPack:
A common confusion for new modders is attempting to build or debug their mod through Visual Studio while tModLoader is already open. They are unaware that they need to disable the mod or close tModLoader first because the `.tmod` file is in-use and unable to be replaced.

The error is confusing and unhelpful:
image

The actual error is shown in the Output panel:
image

This PR adds custom error messages that should hopefully instruct the modder on how to fix the issue:     
`Please close tModLoader or disable the mod in-game to build mods directly`:    
image

In case the error isn't caused by the usual `IOException`, a different message directing the modder to the Output panel is shown:
image

## Alternate solution
We could also change the `ModCompile.BuildModCommandLine` logic to save the file to a custom filename to avoid the issue, such as `ModName (1).tmod`. Assuming `build.txt` has a newer version, it will load the newer .tmod when the debugger launches or the user reloads mods in an already open tModLoader client. This might lead to more issues than it fixes, but maybe some solution avoiding the issue is better in the long run.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
@@ -171,18 +171,13 @@ $@"<Project ToolsVersion=""14.0"" xmlns=""http://schemas.microsoft.com/developer
 			new ModCompile(new ConsoleBuildStatus()).Build(modFolder);
 		}
 		catch (BuildException e) {
-			ErrorReporting.LogStandardDiagnosticError(e.Message, ErrorReporting.TMLErrorCode.TML002);
+			ErrorReporting.LogStandardDiagnosticError(e.Message, e.errorCode);
 			if (e.InnerException != null)
 				Console.Error.WriteLine(e.InnerException);
 			Environment.Exit(1);
 		}
 		catch (Exception e) {
-			if (e is IOException iOException) {
-				ErrorReporting.LogStandardDiagnosticError("Please close tModLoader or disable the mod in-game to build mods directly.", ErrorReporting.TMLErrorCode.TML003);
-			}
-			else {
-				ErrorReporting.LogStandardDiagnosticError(e.Message, ErrorReporting.TMLErrorCode.TML001);
-			}
+			ErrorReporting.LogStandardDiagnosticError(e.Message, ErrorReporting.TMLErrorCode.TML001);
 			Environment.Exit(1);
 		}
 
@@ -229,7 +224,13 @@ $@"<Project ToolsVersion=""14.0"" xmlns=""http://schemas.microsoft.com/developer
 				loadedMod.Close();
 			}
 
-			mod.modFile.Save();
+			try {
+				mod.modFile.Save();
+			}
+			catch (IOException e) {
+				throw new BuildException("Please close tModLoader or disable the mod in-game to build mods directly.", e, ErrorReporting.TMLErrorCode.TML003);
+			}
+
 			ModLoader.EnableMod(mod.Name);
 			// TODO: This should probably enable dependencies recursively as well. They will load properly, but right now the UI does not show them as loaded.
 			LocalizationLoader.HandleModBuilt(mod.Name);
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
@@ -16,6 +16,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Terraria.Localization;
+using Terraria.ModLoader.Engine;
 using Terraria.ModLoader.Exceptions;
 
 namespace Terraria.ModLoader.Core;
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/release_extras/tMLMod.targets
```diff
@@ -39,7 +39,12 @@
 	</ItemGroup>
 
 	<Target Name="BuildMod" AfterTargets="Build">
-		<Exec Command="dotnet $(tMLServerPath) -build $(ProjectDir) -eac $(TargetPath) -define &quot;$(DefineConstants)&quot; -unsafe $(AllowUnsafeBlocks) $(ExtraBuildModFlags)" WorkingDirectory="$(tMLSteamPath)"/>
+		<Exec Command="dotnet $(tMLServerPath) -build $(ProjectDir) -eac $(TargetPath) -define &quot;$(DefineConstants)&quot; -unsafe $(AllowUnsafeBlocks) $(ExtraBuildModFlags)" WorkingDirectory="$(tMLSteamPath)" ConsoleToMSBuild="true" ContinueOnError="ErrorAndContinue">
+			<Output TaskParameter="ConsoleOutput" PropertyName="ModBuildOutput"/>
```
Chicken-Bones: I don't see a use of `ModBuildOutput`. Do output and errors still show in the normal build window?

patches/tModLoader/Terraria/release_extras/tMLMod.targets
```diff
@@ -39,7 +39,12 @@
...
+			<Output TaskParameter="ExitCode" PropertyName="ModBuildExitCode"/>
+		</Exec>
+		<Error Text="Please close tModLoader or disable the mod in-game to build mods directly" Condition=" '$(ModBuildExitCode)' == '1001' "/>
+		<Error Text="Build failed, see output panel for details" Condition=" '$(ModBuildExitCode)' == '1' "/>
```
Chicken-Bones: Consider using `!= 0` instead of `== 1`. If it causes both error messages to show, then consider using the `<Choose><When></When><Otherwise></Otherwise></Choose>` tags

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/release_extras/tMLMod.targets
JavidPack: Right, I was testing earlier to output the console output to the error instead and it was more confusing. I'll remove that. We can do string operations on the output to display a portion of the output one day if we ever need to.

patches/tModLoader/Terraria/release_extras/tMLMod.targets
JavidPack: That works. Error Tasks seem to work in order, stopping at the first one that fails, so it only shows 1.

JavidPack:
PR has been updated to use the "MSBuild and Visual Studio format for diagnostic messages" (https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=vs-2022) approach.

Error codes TML001, TML002, and TML003 added. `LogStandardDiagnosticError` method added. Can be easily expanded to new situations later if we ever need to.

devenv_2024-08-29_12-59-50

devenv_2024-08-29_13-00-04

devenv_2024-08-29_13-17-25

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
```diff
@@ -170,13 +171,18 @@ internal static void BuildModCommandLine(string modFolder)
...
+			ErrorReporting.LogStandardDiagnosticError(e.Message, ErrorReporting.TMLErrorCode.TML002);
 			if (e.InnerException != null)
 				Console.Error.WriteLine(e.InnerException);
 			Environment.Exit(1);
 		}
 		catch (Exception e) {
-			Console.Error.WriteLine(e);
+			if (e is IOException iOException) {
+				ErrorReporting.LogStandardDiagnosticError("Please close tModLoader or disable the mod in-game to build mods directly.", ErrorReporting.TMLErrorCode.TML001);
+			}
```
Chicken-Bones: I imagine this exception is thrown during `TmodFile.Save`. It would be better to report it there. Perhaps we can add an optional error code to `BuildException`

patches/tModLoader/Terraria/ModLoader/Engine/ErrorReporting.cs
```diff
@@ -123,4 +123,19 @@ internal static int ShowMessageBoxWithChoices(string title, string message, stri
 		}
 		return buttonID;
 	}
+
+	/// <summary> Various error codes to show in Visual Studio. Mainly used to cross reference with source code. Subject to change if more granular error codes are needed. </summary>
+	internal enum TMLErrorCode
+	{
+		TML001, // tMod file in use
+		TML002, // Any BuildException
+		TML003, // Other exception
```
Chicken-Bones: Might make sense to reverse these so we can add more specific ones in a more logical order

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
JavidPack: Well, the error message is specifically relating to building the mod directly not in-game, so I figure having it here in `BuildModCommandLine` makes sense. I can put it in `TmodFile.Save` though, it's just a bit ugly try-catching the `using (fileStream = File.Create(path))` line since I'd have to wrap the whole scope of `fileStream`.

Chicken-Bones: Perhaps wrap the `Save` call and throw a custom exception?

Chicken-Bones (approved):

PR #4366, opened 2024-09-03
The Long-Overdue Revival of Content Tags

ThomasThePencil:
### what is this addition?
**content tags!** content tags are designed to be an easy way for tML and modders alike to mark items, NPCs, projectiles, and whatever else have you as a specific sort of thing!

### why do I care?
good question! this system is meant to intermingle with and, in a _very-long-term_ sense, eventually supercede the current ID sets system, which, while definitely very useful (as any modder worth their salt can attest), has proven time and time again to have its shortcomings --- while also simplifyin' a lot of common checks people make for items AND allowin' easier communication between mods on what should and shouldn't be a given thing

### progress checklist
- [X] basic sketch w/ item tags
- [ ] `ItemLoader.ModifyTagList(bool late)` hookset
- [ ] extend to NPCs, projectiles, maybe tiles, potentially other things as well
- [ ] add a ***ton*** of common tags
- [ ] `ModifyTagList(bool late)` hook
- [ ] **item-specific:** items inherit any tags granted by prefixes, figure out how to set this up for vanilla prefixes and add an `InheritedTags` field/property/hook? to `ModPrefix`
- [ ] make new examples for how to manually set a tag and how to check against a tag effectively, includin' when you should use tags and when you should use normal fields
- [ ] gather feedback for 6 to 9 weeks. or a few days, dependin' on how smoothly the rest goes
- [ ] get merged
- [ ] celebrate

### are there other ways to do it?
several, but I picked the string-based design due to ease of use (it is criminally easy to add a string to a hash set) and havin' the highest likelihood to not spontaneously combust with future changes

### how to check an item's tags to do things
```
if (item.Tags.Contains("a_tag"))
{
    // do things
}
```

### ExampleMod updates
to be determined! should be fairly easy to come up with something worth makin' an example for

### DIFF
--- a/patches/tModLoader/Terraria/ID/ItemID.cs.patch
+++ b/patches/tModLoader/Terraria/ID/ItemID.cs.patch
@@ -1,6 +1,8 @@
 --- src/TerrariaNetCore/Terraria/ID/ItemID.cs
 +++ src/tModLoader/Terraria/ID/ItemID.cs
-@@ -2,21 +_,57 @@
+@@ -1,22 +_,59 @@
++using System;
+ using System.Collections.Generic;
  using Microsoft.Xna.Framework;
  using ReLogic.Reflection;
  using Terraria.DataStructures;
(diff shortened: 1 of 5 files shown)


Mirsario:
I.. unfortunately have nothing good to say about this design. Usage of strings & hashsets past load-stage does not scale in the slightest when it comes to tags, and 1.3 Terraria Overhaul had serious framerate drops occur for this reason, prompting me to optimize these approaches out of it.

As there was a [predecessor PR](<https://github.com/tModLoader/tModLoader/pull/1083>), it would've been a good idea to start designing the new one by discussing the past one's failures:
- The old PR had over-use of object-oriented-defined "tag groups" (or tag storages), making them a hassle bigger than the game's "overtypization" that caused it. The new PR does not even have a standardized API for handling tag storages for/and custom types of content. This means that if someone were to want to tag cloud variants, they would have to create their own lists \*somewhere\*, which would make that data inaccessible by other mods.
- The old PR had non-fixed performance costs on every read still, however it at least used lists of integers instead of strings, thus still having lower costs than the new PR. Tags aren't difficult to achieve `O(1)` access speeds for, their presence can even be stored in bitmasks.
- It has proven to be rather uncomfortable to populate tags in that design. 

Over the years, I have, time from time, worked on alternative tag designs right in Overhaul's codebase. You can find the most recent experiment here:
https://github.com/Mirsario/TerrariaOverhaul/compare/dev...misc/content_tags_rewrite
Highlights of that design are:
- `O(1)-ish` access.
- Simple cache definition, as tag handles are content-type agnostic, and are constructed by implicitly casting strings:
  `private static readonly ContentSet explosiveSet = "Explosive";`
- Far more ergonomic handling of tag storages, declared in style of
  `ContentSets.RegisterStorage<ItemID>(static () => ItemLoader.ItemCount)`.
  With more integration with `XyzID` classes and `IdDictionary Search` statics - I suppose that even this step could theoretically be omitted.
- Programmatic population of tags used a builder pattern. I did not like it in the context of declaring footstep information for tiles, but it may be appreciated in content pieces' own files.
- Alternative style of tag population used HJSON files, with present support for populating vanilla XyzID sets, as well as sugar such as adding tags to multiple tag storages all at once:
  `Bullet: { Includes.ItemID|ProjectileID: [ "CrystalBullet" ] }`.
  Good for Overhaul's and similar mods' use-cases.
- Could be improved to use bitmasks instead of boolean arrays, to fit eight times more memory in caches.
- Could be cleaner.

If the community seriously wants to have tags be changeable on per-instance basis - that could still be achieved in a performant-enough manner with modifications to the above design, however content instances whose tags don't differ from their template shouldn't allocate any additional memory IMO.

JavidPack:
Closing since #4381 should cover most of the desired functionality.

PR #4369, opened 2024-09-06
Add hook for armor set bonus activation

Destructor-Ben:
### What is the new feature?
Added a hook that is called when the player wants to activate an armor set bonus by double tapping either down or up based on the users settings (e.g. the Vortex set bonus, which is toggling stealth mode).

Without a hook, a detour is needed for modders to easily accomplish activating a set bonus for their modded armour that uses a double tap to activate.

Yes, such as putting the hook in different places/combining it into a more universal hook that handles double presses in all directions (which would simplify ExampleShield and dashes).

I went with this implementation because it's simpler and I also had an issue where the double tap timer arrays weren't working as expected and I couldn't be bothered figuring out why.

```cs
public class ArmorSetSystem : ModSystem
{
    public override void ArmorSetBonusActivated()
    {
        if (TheSetIsActive)
            DoTheSetBonus();
    }
}
```

### ExampleMod updates

I didn't update, but I could if needed.

### Porting Notes

If you previously used a detour to activate a set bonus or had your own hotkey, then you can switch over to this hook if you like.

As [Ozzatron stated on discord](https://discord.com/channels/103110554649894912/445276626352209920/1281385529346232371)
You can keep a dedicated hotkey because some people like it more than the double tap.

### DIFF
--- a/ExampleMod/Common/Players/ExampleArmorSetBonusPlayer.cs
+++ b/ExampleMod/Common/Players/ExampleArmorSetBonusPlayer.cs
@@ -1,0 +1,49 @@
+﻿using Terraria;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	// This ModPlayer facilitates a set bonus effect. This example shows how either ArmorSetBonusActivated or ArmorSetBonusHeld can be used depending on how you want the player to interact with the set bonus effect. 
+	public class ExampleArmorSetBonusPlayer : ModPlayer
+	{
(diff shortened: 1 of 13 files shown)


JavidPack:
The PR has been updated with ArmorSetBonusHeld, ExampleArmorSetBonusPlayer, LearningExampleKeybind, and docs.

ArmorSetBonusHeld is for hold down effects. ExampleArmorSetBonusPlayer showcases an armor set now and also showcases shadow styles. LearningExampleKeybind showcases hold down and double tap behaviors for arbitrary keybinds.

PR #4381, opened 2024-09-16
Custom and Named ID Sets

JavidPack:
This PR makes the creation of **custom ID sets** (like `ItemID.Sets.IsDrill`) easier, and allows registering them with a name to be shared with other mods. This allows mods to easily implement content id-specific data. Custom ID sets are designed to behave like the existing ID sets modders are familiar with and work seamlessly with other mods, regardless of load order or mod dependencies. 

When a custom ID set is registered with a name, it becomes a **named ID set**. Mods can use named ID sets to collaborate in an "ad-hoc" manner, accessing shared data that is not already present in tModLoader but also isn't exclusively defined by a single dependency mod. The mods do not depend on each other, but do require a shared consensus on the meaning of the data. tModLoader will continue to add official ID sets for helping mods implement existing behaviors, but this named ID set feature is intended for new ideas and features shared between several mods. 

**It is important to note** that named sets are shared even when the mod in the name is not present (eg, multiple mods can use `ExampleMod/FlamingWeapon` without `ExampleMod` being present)

This PR also adds a highly requested hook after mod content has all loaded but before mod content is setup. This is `ModSystem.ResizeArrays`. 

This PR was motivated by discussions of proposed "Content Tags" (https://discord.com/channels/103110554649894912/1280557208408555613). This PR should cover the main use-case brought up in those discussions in a manner familiar to modders, but "content tags" are still being discussed and it remains to be seen how that will be implemented. Additionally, other PRs are also being prepared that relate to this PR.

# Custom/Named ID Set
To create a named ID set, we suggest creating a static field in a static class. This class is also annotated with `ReinitializeDuringResizeArrays`, explained below. The static field should be initialized by calling the `CreateNamedSet` method with a `key`, then optionally chaining to the `Description` method, and finally chaining to the corresponding `RegisterXSet` method to provide the default value and initial data. 

In this example, a named ID set is initialized in the `CustomItemSets` class. The key is `"ExampleMod/FlamingWeapon"` (see below), the default value is `false`, and `FieryGreatsword` and `ExampleSword` are initialized to being `true`. This array is then updated in `ExampleFlail.SetStaticDefaults` in the same manner as a vanilla ID set (`ItemID.Sets.ToolTipDamageMultiplier` in this example). The `CustomSetsModPlayer` class then uses the set to implement custom behavior for every item in that set. (This code is a simplified version of the [CustomItemSets.cs](https://github.com/tModLoader/tModLoader/blob/preview/ExampleMod/Content/Items/CustomItemSets.cs) included in this PR as part of ExampleMod.)

```cs
[ReinitializeDuringResizeArrays]
public static class CustomItemSets
{
	public static bool[] FlamingWeapon = ItemID.Sets.Factory.CreateNamedSet("FlamingWeapon")
		.Description("Causes \"Hahahah, burn!\" to randomly show in chat when used")
		.RegisterBoolSet(false, ItemID.FieryGreatsword, ModContent.ItemType<ExampleSword>());
	// Other sets...
}

public class ExampleFlail : ModItem
{
		ItemID.Sets.ToolTipDamageMultiplier[Type] = 2f;
		CustomItemSets.FlamingWeapon[Type] = true;
	}

	// Other code
}

public class CustomSetsModPlayer : ModPlayer
{
	public override void OnHitAnything(float x, float y, Entity victim) {
		if (CustomItemSets.FlamingWeapon[Player.HeldItem.type] && Main.rand.NextBool(5)) {
			CombatText.NewText(Player.getRect(), Color.Red, "Hahahah, burn!");
		}
	}
}
```

These examples are `bool[]`, but other types work as well. Custom/Named ID sets are not limited to vanilla content or primitive types. 

<details><summary>Expand for custom Type example:</summary><blockquote>

Here is an example using a Tuple (`ValueTuple<string, int>`), allowing each entry to contain a pair of data rather than making 2 separate sets. The same approach would be used for a custom `Type` except the `Type` would be initialized as normal. 

```cs
// ExampleMod:
public static (string description, int cost)[] TupleTest = HairID.Sets.Factory.CreateNamedSet("HaircutInfo")
    .Description("Contains a description of a hair style and a custom price")
    .RegisterCustomSet<(string, int)>(default,
        6, ("Long Braid", Item.buyPrice(gold: 3)),
        9, ("Ponytail", Item.buyPrice(platinum: 1))
    );
// OtherMod:
TupleTest = HairID.Sets.Factory.CreateNamedSet("ExampleMod/HaircutInfo")
    .Description("Affects NPC happiness")
    .RegisterCustomSet<(string, int)>(default,
        3, ("Guile", Item.buyPrice(gold: 55))
    );
// Test code:
if(CustomItemSets.TupleTest[Player.hair] != default) {
    var tuple = CustomItemSets.TupleTest[Player.hair];
    Main.NewText($"Players hair: {tuple.description}, {tuple.cost}");
}
// /customsets output:
HairID ValueTuple`2(String, Int32)[] ExampleMod/HaircutInfo
    Default Value: (, 0)
    Used by: ExampleMod, CustomSetTest1
    Descriptions:
        ExampleMod: Contains a description of a hair style and a custom price
        CustomSetTest1: Affects NPC happiness
    Contents:
        3, (Guile, 550000)
        6, (Long Braid, 30000)
        9, (Ponytail, 1000000)
```

</blockquote></details>

It is also possible to make your own `SetFactory` instance for a custom content type (such as a `ModType`) in your mod. See the `Custom SetFactory for custom ModType` section below.

# Sharing Named ID Sets
In the above example, the named ID set was registered with `"FlamingWeapon"` as its key. The actual key for this set is "ExampleMod/FlamingWeapon" since the mod name is automatically prepended if there is no `/` in the provided key. Named ID sets with the same `key` and `Type` are automatically merged and will share the same reference. For example, if `YourMod` also registers a `bool[]` by calling `CreateNamedSet` with the "ExampleMod/FlamingWeapon" key, both mods will get a reference to the same array and any additional edits made by `YourMod` will be seen by `ExampleMod` as well.

```cs
public static bool[] FlamingWeapon = ItemID.Sets.Factory.CreateNamedSet("ExampleMod/FlamingWeapon").RegisterBoolSet(false, ItemID.HelFire);
```

If both mods are enabled, `ExampleMod/FlamingWeapon` will have true values for `FieryGreatsword`, `ExampleSword`, and `HelFire`. This allows each mod to implement whatever behaviours they want for items tagged as "ExampleMod/FlamingWeapon" without needing to collaborate directly. Mods using named ID sets need to come to an agreement on the meaning of the named ID sets as well as their `Type` and default value. 

## Set keys
The `CreateNamedSet` method has 3 overloads, each will results in the set key as follows:
```
CreateNamedSet(Mod mod, string key) -> "{mod.Name}/{key}"
CreateNamedSet(string modName, string key) -> "{modName}/{key}"
CreateNamedSet(string fullKey) -> "{key}" if key contains a "/". Otherwise, the final key will be derived automatically from the currently loading mod: "{loadingMod.Name}/{key}"
```

By requiring each key have a `"namespace/"` (normally a mod name) we prevent undesired conflicts. There is no global namespace, but you're free to make up whatever namespaces you want.

## /customsets chat command
The descriptions of named ID sets can help other modders know what sets mean. In-game a user can type `/customsets` to output the data contained in custom sets to "CustomSets.txt" in the logs directory. It also supports outputting values as well, `/customsets -v`. See `/customsets -h` to see more options.

Sample Output:
```
ItemID Boolean[] ExampleMod/FlamingWeapon
	Default Value: False
	Used by: ExampleMod
	Description: Causes "Hahahah, burn!" to randomly show in chat when used
	Contents:
		FieryGreatsword
		HelFire
		FireWhip
		ExampleMod/ExampleFlail
		ExampleMod/ExampleSword
```

## Merging Sets
We expect modders to collaborate on set names with equivalent meaning. For example, if `ModA` pioneers a set called `SpicyFood`, then we expect `ModB` to use `ModA/SpicyFood` as the set key so that the data is shared. There will be situations, however, where multiple mods arrive at their own set names independently and are unable to change their set name to collaborate, such as if they have other dependent mods using each of their sets already. In this case, the `SetFactory.MergeSets` method can be used to ensure that the specified sets should share the same key. All mods will end up accessing the same array reference despite using separate keys.

```cs
SetFactory.MergeSets(ItemID.Sets.Factory, typeof(bool), "ModA/SpicyFood", "ModB/HotFood");
```

When this happens, the set will be reported as `{ModA/SpicyFood, ModB/HotFood}` in logs and the `/customsets` command. 

# Collaboration
Named ID sets are only useful if mods collaborate on their meaning. For example, does a set called "Hot" mean the item is a spicy food or a weapon that has a fire tag? Should it hurt the user to wield it, or does it do extra damage to water enemies? The only way to know is to collaborate with other mods and agree on the intention and meaning of the name. Each mod can provide their own entries, effects and behaviors for the set, but there should generally be a consensus as to their meaning among the modding community. Usually the meaning of a set comes from the mod that pioneered the set name, especially if the set key starts with their mod name ("ExampleMod/Fiery"). Use a description to guide other mod authors on what your set does and when to add things to it. 

## Wiki Page
To facilitate collaboration, the tModLoader wiki has been updated with a page listing of keys and how various mods are using various named ID set names. Once definite plans have been made or your mod has been updated, feel free to update the wiki page: https://github.com/tModLoader/tModLoader/wiki/Named-ID-Sets

# ReinitializeDuringResizeArraysAttribute
During mod loading, a static field initializer might run too early to register a custom set. If called before all content is loaded, the resulting array would have the incorrect length. You can now add `[ReinitializeDuringResizeArrays]` to your class to have its static constructor called again after all content has been registered. Just be aware that other static fields will be reinitialized as well, so you may want to split some fields into a separate class. Alternatively, you can use `ModSystem.ResizeArrays` to initialize your static fields with custom ID sets. 

# ModSystem.ResizeArrays
Called after all mods register content and before all mods start setting up content. Use this to initialize data structures dependent on the count of various content IDs. tModLoader uses this space to initialize data structures to the right size and prepare for content to be configured by the rest of loading (such as `SetStaticDefaults`). Prior to this PR a modder would need  to use resizable data structures such as dictionaries, sets or lazy initialization.

# ExampleMod
`CustomItemSets.cs` shows off custom ID sets usage.

# Porting Notes
This PR does not contain breaking changes, but any mod interested in custom ID sets is encouraged to use the approach shown. We have worked hard to ensure that mod load order complications are automatically handled transparently. 

If you were using a `Dictionary` or `HashSet` previously to handled these situations, you might find that using `[ReinitializeDuringResizeArrays]` can simplify some logic as well. Those data structures can still be used, this PR just adds support for fixed length arrays typical to the `Sets` classes.

There are some situations that were incorrect before but now throw errors during loading due to this PR. The most common issue is calling `EquipLoader.AddEquipTexture` for an equip type that has already been autoloaded using `[AutoloadEquip]`. To fix, get rid of one or the other.

# Other Mod
This PR was tested using this mod. Data is shared as expected and can be verified using the `/customsets` chat commands. Consult the code below to see how another mod interacts with the custom ID sets used by ExampleMod.

<details><summary>Expand for Source Code:</summary><blockquote>

```cs
using System.Linq;
using Terraria;
using Terraria.ID;
using Terraria.ModLoader;

namespace CustomSetTest1
{
	public class CustomSetTest1 : Mod
	{
	}

	public class MySets : ModSystem
	{
		public static bool[] FireWeapons;

		public override void ResizeArrays() {
			// This example uses ResizeArrays instead of the field initializer, this also works.

			FireWeapons = ItemID.Sets.Factory.CreateNamedSet("ExampleMod/FlamingWeapon")
				.Description("Mod specific information here.")
				.RegisterBoolSet(false);
			FireWeapons[ItemID.Flamethrower] = true;
			FireWeapons[ItemID.Flamelash] = true;
		}
	}

	public class MyFireWeapon : ModItem
	{
			MySets.FireWeapons[Type] = true;
		}
	}
}
```

</blockquote></details>

# Custom `SetFactory` for custom `ModType`
While `ExampleMod` doesn't currently contain a custom `ModType` suitable for a custom `SetFactory` example (until next month, see below), we did do some basic testing and added code to ensure that a custom `SetFactory` for a custom ID (likely from a `ModType`, but could be for any custom content tracked by your mod) will work. This is intended for new content types, not existing types.

<details><summary>Expand for more info:</summary><blockquote>

Some guidelines for implementing a custom `SetFactory`:

1. Initialize the `SetFactory` in the same class as the custom sets (the class with `[ReinitializeDuringResizeArrays]`) and make sure it is placed on lines above the custom set registrations.
2. Other mods referencing the `SetFactory` directly should be sorted after the base mod in load order. This is automatic with a mod reference, but if not using a reference somehow, like through reflection, initialize the sets in `ResizeArrays` rather than the `[ReinitializeDuringResizeArrays]` class.
3. As always, the sets are not fully initialized until after `ResizeArrays`, so leave any further set manipulation for `ModType.SetStaticDefaults` methods.

Here is a simplified Example.  It isn't an actual working example but shows how a `SetFactory` might be used for a custom ID. If you have issues implementing a `SetFactory` for a custom ID please reach out, there may be corner cases we have not anticipated yet that we can fix if the issue is reported to us:

```cs
// Base Mod
public class BookID
{
	[ReinitializeDuringResizeArrays]
	public static class Sets
	{
		public static SetFactory Factory = new SetFactory(BookID.BookCount, "MyMod/BookID", Search);

		public static bool[] PictureBooks = Factory.CreateNamedSet("PictureBooks")
			.Description("Can be placed on tables.")
			.RegisterBoolSet(false, 1);
	}

	internal static int BookCount { get; set; } = 0;

	public static IdDictionary Search = IdDictionary.Create<BookID, int>();
}

public class BookLoaderSystem : ModSystem
{
	public override void Load() {
		// Simulate loading 4 instances of a custom ModType
		RegisterBook("TestBook1");
		RegisterBook("TestBook2");
		RegisterBook("TestBook3");
		RegisterBook("TestBook4");
	}

	public void RegisterBook(string name) {
		BookID.Search.Add(name, BookID.BookCount);
		BookID.BookCount++;
	}
}

// Other Mod accessing the custom SetFactory
[ReinitializeDuringResizeArrays]
public class OtherModSets
{
	public static bool[] StudyBooks = BookID.Sets.Factory.CreateNamedSet("StudyBooks").RegisterBoolSet(false, 4, 5); // new set

	public static bool[] PictureBooks = BookID.Sets.Factory.CreateNamedSet("MyMod/PictureBooks").RegisterBoolSet(false, 4); // set from base mod
}

public class OtherModSystem : ModSystem
{
	public override void Load() {
		// Simulate loading 2 instances of a custom ModType
		ModContent.GetInstance<BookLoaderSystem>().RegisterBook("OtherModBook1"); // 4
		ModContent.GetInstance<BookLoaderSystem>().RegisterBook("OtherModBook2"); // 5
	}
}
```

</blockquote></details>

## ExampleMod Example of Custom `SetFactory` for custom `ModType`

In the month after this PR goes live in stable, a custom `ModType` example will be available in ExampleMod. The [CustomModType/VictoryPoseID.cs](https://github.com/tModLoader/tModLoader/pull/4611/files#diff-eb50a2783ebf12746f7c0c37569e036a87a443ed9f369c45e1e450e0c54e5841) class contained in the [Custom ModType Example pull request](https://github.com/tModLoader/tModLoader/pull/4611) shows an example of implementing a custom `SetFactory` for a custom `ModType`.

# Additional Changes
`public static IdDictionary Search = IdDictionary.Create<IDCLASSHERE, int>();` has been added to most of the ID classes that did not have a `Search` previously.


Chicken-Bones:
Custom sets for item and projectile ids definitely need to be separated. One option would be to put the `CreateCustomSet` methods on `SetFactory`, this way they're not tied to `ItemID` and `ProjectileID` but rather to `ItemID.Sets`, and modders can create their own `SetFactory` instances without needing an id type.

We're encouraging people to only call `SetFactory.Create...` methods during or after `ResizeArrays`, we can probably enforce this with an exception.

We should create a `SetFactory(Func<int> getCount)` constructor, to reduce ordering dependencies with mods that have custom content. Otherwise, if `MyMod` adds `MyStuffLoader` and only updates `MyStuffLoader.SetFactory` in `ResizeArrays` once all the `MyStuff` loadables have finished and the final count is known, then all other mods which want to reference that `SetFactory` in  `ResizeArrays` would have to be ordered after `MyMod`.

Chicken-Bones:
I'm not opposed to `DataInstance` but the values ought to be strongly typed. We already have `ContentInstance` and `ModTypeLookup`, so `DataInstance` would be purely for more democratic data sharing, where weak refs aren't required, and only types from common dependencies are used. 

`ModTypeLookup` could be implemented in terms of `DataInstance` by using `fullName` and `(modName, contentName)` as keys. Custom sets could use `(Factory, fullName)` as keys

Chicken-Bones:
We may also want to a `static ulong LookupCount` which is reset each `Update` and incremented every time an attempt is made to retrieve a value from `DataInstance`, then we can output a performance warning if there are a lot (say, more than 100) requests per frame after `Load` has finished, encouraging people to cache their lookups.

patches/tModLoader/Terraria/ModLoader/SystemLoader.cs
```diff
@@ -37,9 +38,22 @@ internal static void Unload()
 		SystemsByMod.Clear();
 	}
 
-	internal static void ResizeArrays()
+	internal static void ResizeArrays(bool unloading)
 	{
 		RebuildHooks();
+
+		if (!unloading) {
```
Chicken-Bones: Invert `if`

patches/tModLoader/Terraria/ModLoader/Mod.cs
```diff
@@ -107,6 +108,8 @@ public partial class Mod
 	/// <inheritdoc cref="Terraria.ModLoader.PreJITFilter"/>
 	public PreJITFilter PreJITFilter { get; protected set; } = new PreJITFilter();
 
+	internal HashSet<Type> ReinitializeDuringResizeArraysTypes = new();
```
Chicken-Bones: No need to cache this, it will only be used once.

patches/tModLoader/Terraria/ModLoader/ModSystem.cs
```diff
@@ -408,4 +408,13 @@ public virtual void ModifyLightingBrightness(ref float scale) { }
...
 	/// </summary>
 	public virtual void TileCountsAvailable(ReadOnlySpan<int> tileCounts) { }
+
+	/// <summary>
+	/// Called after all mods register content and before all mods start setting up content. Use this to initialize data structures dependent on the count of various content IDs, usually through SetFactory instances directly.
+	/// <para/> For example, <c>ItemID.Sets.Factory.CreateBoolSet(false, ItemID.PoisonDart, ItemID.PoisonedKnife)</c> will create an array with length equal to the total number of items in the game with the specified item types set to true.
+	/// <para/> <see cref="SetHandler.RegisterCustomSet{T}(string, T, T[])"/> can be used to expose a custom set for other mods to use by name without a dependency.
+	/// </summary>
+	public virtual void ResizeArrays() {
+		// TODO: bool unloading? Or maybe just only call this hook when loading in the first place?
```
Chicken-Bones: `TODO`  can be removed, only call during `Load` is fine. We have `Unload` if they need it

patches/tModLoader/Terraria/ModLoader/DataInstance.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System.Collections.Concurrent;
+using Terraria.ModLoader.Core;
+
+namespace Terraria.ModLoader;
+
+public static class DataInstance
```
Chicken-Bones: Lets split this to a separate PR

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,159 @@
...
+			throw new Exception($"Previously registered set for {key} has a default value of {existingMetadata.defaultValue} but {newMetadata.defaultValue} was supplied. Custom data set will not be registered");
+			// TODO: We could just allow this and output a warning that the data will not be shared. Keep both somehow?
+		}
+		/*if (newmetadata.length != existingmetadata.length)
+		{
+			throw new exception($"previously registered set for {key} is has length {existingmetadata.length} but supplied set has length {newmetadata.length}. custom data set will not be registered");
+		}*/
+
+		// Custom sets are registered as (SetFactory, key) in DataInstance.
+		T[] value = DataInstance<T[]>.GetOrAdd((this, key), input);
```
Chicken-Bones: Both this and the `setMetadataMapping` perform basically the same function with similar performance costs. `DataInstance` is compile time type safe, but more complicated. There's no reason to have both. Store the `T[]` in the `SetMetadata` here (`setMetadataMapping`) or in `DataInstance` (probably here, for simplicity)

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,159 @@
...
+using ReLogic.Utilities;
+using Terraria.ModLoader;
+using Terraria.ModLoader.Core;
+
+namespace Terraria.ID;
+
+public partial class SetFactory
+{
+	// Additional code to support named custom sets for ad-hoc collaboration
+	private record SetMetadata(Type type, object defaultValue); // Note: Stores the array Type, not Type[]
```
Chicken-Bones: `type` no longer required?

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,159 @@
...
+		// If it already exists, merge the data
+		if (value != input) {
+			if (value.Length != input.Length) {
+				throw new Exception("Input set and existing set are of different lengths.");
+			}
+
+			// To merge, we find entries in the input that aren't defaultValue and assign them to the result.
+			// Existing changes should persist as long as mods agree on the defaultValue passed in and used in CreateXSet
+			for (int i = 0; i < input.Length; i++) {
+				if (!input[i].Equals(defaultValue)) {
```
Chicken-Bones: Standard advice is to use `EqualityComparer<T>.Default.Equals(...)` when doing generic comparisons that may include value types. Minor performance difference, and also saves you from the `NullReferenceException` you have here if `input[i] is null`

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,159 @@
...
+	{
+		var set = CreateIntSet(defaultState, inputs);
+		RegisterNamedCustomSet(key, defaultState, ref set);
+		return set;
+	}
+	public int[] CreateNamedIntSet(string key, params int[] types)
+	{
+		var set = CreateIntSet(types);
+		RegisterNamedCustomSet(key, -1, ref set);
+		return set;
```
Chicken-Bones: You could make this single line by replacing the `ref` with a regular parameter, and having `RegisterNamedCustomSet` return the merged result

ExampleMod/Content/Items/CustomItemSets.cs
```diff
@@ -0,0 +1,63 @@
...
+			// For example, we could put "CustomItemSets.FlamingWeapon = ItemID.Sets.Factory.CreateNamedBoolSet(FlamingWeaponCustomSetKey, false, ItemID.FieryGreatsword);" here instead of in the CustomItemSets class field initializers.
+		}
+
+		public override void SetStaticDefaults() {
+			// We can further edit the data sets here. These changes will still be consistent between all mods accessing this set since the object reference is shared.
+			CustomItemSets.FlamingWeapon[ItemID.FireWhip] = true;
+			CustomItemSets.FlamingWeapon[ItemID.HelFire] = true;
+
+			// For example, we can add user-defined items to CustomItemSets.FlamingWeapon
+			// Note that this example would need more code to work correctly for config values changed in-game, this is just showing a potential use-case.
```
Chicken-Bones: config values changed ingame for something like this should have a `NeedsReload`. It's likely that custom sets will be used in `AppliesToEntity` methods which are frozen after loading completes.

Chicken-Bones (changes requested):

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: (Will be in a separate PR since there are more issues with ItemDefinition/ItemID.Search)

ExampleMod/Content/Items/Accessories/WaspNest.cs
```diff
@@ -64,10 +68,21 @@ public override void UpdateAccessory(Player player, bool hideVisual) {
 			waspNestPlayer.strongBeesUpgrade = true;
 			waspNestPlayer.strongBeesItem = Item;
 		}
+	}
 
+	// We can safely use ReinitializeDuringResizeArrays on this class to properly initialize CantEquipWith_HiveBackpack because their are no other static fields that we wouldn't want to reset, but if we did have that then we would want to move CantEquipWith_HiveBackpack to it's own class or move it to CustomItemSets.
```
Chicken-Bones: `their` > `there`

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,141 @@
...
+
+	// This is private to prevent potential modder mistake.
+	private T[] RegisterNamedCustomSet<T>(string key, T defaultValue, T[] input)
+	{
+		RegisterNamedCustomSet(key, defaultValue, ref input);
+		return input;
+	}
+
+	/// <summary>
+	/// Registers a "custom data set" for collaboration, meaning an array of values of length equal to the count of the content the set corresponds to. This is typically done through the Terraria.ID.XID.Sets.Factory.CreateXSet methods, but this method can be used for manually initialized arrays.
```
Chicken-Bones: `Terraria.ID.XID.Sets.Factory.CreateXSet` `CreateNamedXSet`

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,141 @@
...
+	/// <para/> The <paramref name="key"/> is a special identifier that all mods intending to collaborate on this custom data set will use. Mod makers should collaborate on the key and its associated meaning. Use the <see cref="RegisterNamedCustomSet{T}(string, string, T, ref T[])"/> or <see cref="RegisterNamedCustomSet{T}(Mod, string, T, ref T[])"/> overload instead to create a custom data set for a mod-specific idea to give it a unique key that will not conflict with a key used by any other mod.
+	/// <para/> More information on properly using these methods can be found in the <see href="https://github.com/tModLoader/tModLoader/pull/4381">Custom Data Sets pull request</see>. Custom data set keys used by other mods are listed in the <see href="https://github.com/tModLoader/tModLoader/wiki/Custom-Data-Sets">Custom Data Sets wiki page</see>, please read and consult this page to collaborate on custom data set key meanings.
+	/// <para/> Throws an exception if the data length or default value does not match the data registered using the same key by any mod loaded before this mod.
+	/// </summary>
```
Chicken-Bones, line 232: It's a bit weird to have so much detailed documentation on a method which isn't the main one people will use. I don't think it's necessary to go into extreme detail here. The information in the PR description and on the `CreateNamedSet` methods should suffice

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,141 @@
...
+	public void RegisterNamedCustomSet<T>(string key, T defaultValue, ref T[] input)
+	{
+		// TODO: Return bool to represent already exists or merged?
```
Chicken-Bones: No real use for return value. Remove TODO imo

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		// Note: Intended to be load order independent as long as all parties agree on default value. Any deviation will throw exception.
+		SetMetadata newMetadata = new SetMetadata(defaultValue, input);
+		(string key, Type) dictionaryKey = (key, typeof(T));
+		SetMetadata existingMetadata = setMetadataMapping.GetOrAdd(dictionaryKey, newMetadata);
+
+		if (!EqualityComparer<object>.Default.Equals(newMetadata.defaultValue, existingMetadata.defaultValue)) { // Primitive might be boxed, so != doesn't work.
+			throw new Exception($"Previously registered set for {key} has a default value of {existingMetadata.defaultValue} but {newMetadata.defaultValue} was supplied. Custom data set will not be registered");
+			// TODO: We could just allow this and output a warning that the data will not be shared. Keep both somehow?
```
Chicken-Bones: Lets not

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		// TODO: Once DataInstance feature is merged, Custom sets can be registered as (SetFactory, key) in DataInstance.
```
Chicken-Bones: Remove TODO imo, I think this impl is good

patches/tModLoader/Terraria/ModLoader/Config/EntityDefinition.cs
```diff
@@ -12,6 +12,7 @@ namespace Terraria.ModLoader.Config;
 
 /// <summary>
 /// Classes implementing EntityDefinition serve to function as a way to save and load the identities of various Terraria objects. Only the identity is preserved, no other data such as stack size, damage, etc. These classes are well suited for ModConfig, but can be saved and loaded in a TagCompound as well.
+/// <para/> An EntityDefinition can potentially refer to content removed from a mod or content from a mod not currently enabled. Use <see cref="IsUnloaded"/> to check if this is the case, especially before using <see cref="Type"/> as an index into an array. It is usually desirable to preserve all unloaded EntityDefinition entries so user data is not lost, so structure your code accordingly.
```
Chicken-Bones, line 15: I assume these doc updates were from the config example. I won't make you remove it, just doesn't really belong in this PR

patches/tModLoader/Terraria/ModLoader/ReinitializeDuringResizeArraysAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Classes annotated with this attribute will have their static constructor called again during the ResizeArrays stage of mod loading.
+/// <para/> This is intended for classes containing data sets created through <see cref="Terraria.ID.SetFactory"/>, similar to the design of vanilla classes such as <see cref="ID.ItemID.Sets"/>. This attribute removes the need to manually initialize these data sets in <see cref="ModSystem.ResizeArrays"/> and helps avoid mod ordering issues that would complicate the implementation logic.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+public sealed class ReinitializeDuringResizeArraysAttribute : Attribute
```
Chicken-Bones, line 12: Should we note that it runs before `ModSystem.ResizeArrays` for the current mod?

patches/tModLoader/Terraria/ModLoader/SystemLoader.cs
```diff
@@ -37,9 +40,31 @@ internal static void Unload()
...
 	{
 		RebuildHooks();
+
+		if (unloading) {
+			return;
+		}
+		foreach (var mod in ModLoader.Mods) {
+			var reinitializeTypes = AssemblyManager.GetLoadableTypes(mod.Code)
+				.Where(t => t.GetAttribute<ReinitializeDuringResizeArraysAttribute>() != null)
+				//.Where(t => !t.IsAbstract && !t.ContainsGenericParameters) // static is IsAbstract and IsSealed
```
Chicken-Bones: No need to keep these commented out. But we might want to note that `[ReinitializeOnResizeArrays]` can't work on generic classes

patches/tModLoader/Terraria/ModLoader/SystemLoader.cs
```diff
@@ -37,9 +40,31 @@ internal static void Unload()
...
+				//.Where(t => t.IsAssignableTo(typeof(ILoadable)))
+				//.Where(t => t.GetConstructor(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic, Type.EmptyTypes) != null) // has default constructor
+				.Where(t => AutoloadAttribute.GetValue(t).NeedsAutoloading)
```
Chicken-Bones: This one is tricky. We want it to work with `Autoload(false)` but not with `Autoload(Side = ...)` on the wrong side. I think it's best to remove this to reduce footguns. 

The `Side =` and `ModOrganizer.LoadSide` checks should really be applied to `GetLoadableTypes` but no-one has complained yet...

patches/tModLoader/Terraria/ModLoader/SystemLoader.cs
```diff
@@ -37,9 +40,31 @@ internal static void Unload()
...
+				.OrderBy(type => type.FullName, StringComparer.InvariantCulture);
+
+			foreach (var typesToReinitialize in reinitializeTypes) {
+				// Uninitialized static ctor will be initialized twice here for some reason.
```
Chicken-Bones: haha, not a problem though

Chicken-Bones (changes requested):

Chicken-Bones (approved):

Mirsario (changes requested):
The ability to specify default values on shared sets as a concept can only lead to a load-time crash or a silent failure, both of which would be non-discoverable at development & playtesting time unless the developer has all 7000+ mods downloaded and loaded for debugging. It is a can of worms, so the parameter shouldn't be offered to the users at all, lest we end up with tons of mod incompatibilities that aren't possible to resolve, as collaboration with inactive people will be required. Boolean sets thus should all default to `false`, integer sets often need to point to zero content IDs, so `-1` should be an agreeable default for those?

In more detail, message in Discord: https://discord.com/channels/103110554649894912/445276626352209920/1311660844182011914

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -16,14 +18,23 @@ namespace Terraria.ID;
 public partial class SetFactory
 {
 	// Additional code to support named custom sets for ad-hoc collaboration
+
+	// Contains all SetFactory instances relating to named ID sets.
+	internal static HashSet<SetFactory> setFactories = new HashSet<SetFactory>();
```
Chicken-Bones: `ProperCase` please

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -16,14 +18,23 @@ namespace Terraria.ID;
...
+
 	private record SetMetadata(object defaultValue, object array);
+	private record SetNameTypePair(string setName, Type type);
 
-	private ConcurrentDictionary<(string, Type), SetMetadata> setMetadataMapping = new ConcurrentDictionary<(string, Type), SetMetadata>();
+	private ConcurrentDictionary<SetNameTypePair, SetMetadata> setMetadataMapping = new ConcurrentDictionary<SetNameTypePair, SetMetadata>();
+	private ConcurrentDictionary<SetNameTypePair, HashSet<string>> involvedMods = new ConcurrentDictionary<SetNameTypePair, HashSet<string>>();
+	private ConcurrentDictionary<SetNameTypePair, Dictionary<string, string>> setDescriptions = new ConcurrentDictionary<SetNameTypePair, Dictionary<string, string>>();
```
Chicken-Bones: Seems like these should be merged, with one class containing all the values here

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -85,6 +96,11 @@ private T[] RegisterNamedCustomSet<T>(string key, T defaultValue, T[] input)
 	/// </summary>
 	public void RegisterNamedCustomSet<T>(string key, T defaultValue, ref T[] input)
 	{
+		// The "global namespace" aspect to this feature is disabled until we get more modder feedback. In the meantime modders are free to collaborate "globally" by using "Terraria" as the mod name if they wish.
+		if (!key.Contains("/")) {
+			throw new Exception($"Custom sets must be registerd with a key that contains `/`. Usually in the form of 'ModName/SetName'. The key '{key}' is not valid, either adjust the key or use an overload that requires a mod or mod name to be passed in.");
```
Chicken-Bones: typo, `registered`

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -119,8 +135,16 @@ public void RegisterNamedCustomSet<T>(string key, T defaultValue, ref T[] input)
 					value[i] = input[i];
 				}
 			}
+
+			if(ModCompile.activelyModding)
```
Chicken-Bones: ` ` after `if`

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
```diff
@@ -119,8 +135,16 @@ public void RegisterNamedCustomSet<T>(string key, T defaultValue, ref T[] input)
...
+				Logging.tML.Info($"Custom Set '{key}' (Type: {typeof(T).Name}) is merging with additional data");
```
Chicken-Bones: We want to log the mods which are using the set, perhaps only when there's 2 or more of them?

patches/tModLoader/Terraria/ModLoader/Default/CustomSetsCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+		var sb = new StringBuilder();
+		foreach (var factory in SetFactory.setFactories) {
+			string metadata = factory.CustomMetadataInfo();
+			sb.Append(metadata);
+		}
+
+		string outputText = sb.ToString();
+		caller.Reply(outputText);
+
+		string outputPath = Path.Combine(Logging.LogDir, "CustomSetData.txt");
```
Chicken-Bones: `CustomSets.txt`?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
JavidPack: Currently this is buggy due to the static initializer running twice issue. I could make a check for "any changes", which would solve it but would leave mods that are just accessing the named set unreported. That might be good though.

ExampleMod/Content/Items/CustomItemSets.cs
```diff
@@ -25,7 +25,7 @@ public static class CustomItemSets
 		// To create a named ID set for items, we use the ItemID.Sets.Factory.CreateNamedXSet method and provide a string key.
 		// We can also pass in any initial data, in this case we are indicating that FieryGreatsword and ExampleSword should be true in this set. Note that it is also possible to set these set values in SetStaticDefaults instead, which is more typical. See ExampleFlail.cs for an example.
 		// This method also exposes the set for other mods to access via this key. The key and default value must be consistent with other mods.
-		public static bool[] FlamingWeapon = ItemID.Sets.Factory.CreateNamedBoolSet(nameof(ExampleMod), FlamingWeaponCustomSetKey, false, ItemID.FieryGreatsword, ModContent.ItemType<ExampleSword>());
+		public static bool[] FlamingWeapon = ItemID.Sets.Factory.CreateNamedBoolSetWithInfo(FlamingWeaponCustomSetKey, false, "Causes \"Hahahah, burn!\" to randomly show in chat when used", ItemID.FieryGreatsword, ModContent.ItemType<ExampleSword>());
```
Chicken-Bones: description should definitely go before the default value, and we should use multiple lines to split up such a big call

ExampleMod/Content/Items/CustomItemSets.cs
```diff
@@ -0,0 +1,79 @@
...
+	// "Named ID sets" are automatically merged when other mods declare the same ID set, facilitating "ad-hoc" collaboration that doesn't require mod dependencies.
+	// See https://github.com/tModLoader/tModLoader/pull/4381 for more information about custom and named ID sets.
+
+	// This [ReinitializeDuringResizeArrays] attribute will cause this class's static constructor to be called during the ResizeArrays step of mod loading. This is essential for any class with field initializers calling SetFactory methods.
+	// This will allow the arrays to have the correct lengths after all content has been loaded into the game. This reinitialization happens before ModSystem.ResizeArrays, avoiding potential issues from mod load order.
+	[ReinitializeDuringResizeArrays]
+	public static class CustomItemSets
+	{
+		// Named ID set example. This will behave the same as any other ItemID.Sets array.
+		public const string FlamingWeaponCustomSetKey = "FlamingWeapon"; // This string MUST be consistent between mods working together.
```
Chicken-Bones: Comment here is not entirely correct anymore, since we're not doing global namespace. The documentation here and below now needs to indicate that other mods would need to use `ExampleMod/FlamingWeapon` in order for them to merge.

ExampleMod/Content/Items/Accessories/WaspNest.cs
```diff
@@ -64,10 +68,21 @@ public override void UpdateAccessory(Player player, bool hideVisual) {
...
 			waspNestPlayer.strongBeesItem = Item;
 		}
+	}
 
+	// We can safely use ReinitializeDuringResizeArrays on this class to properly initialize CantEquipWith_HiveBackpack because there are no other static fields that we wouldn't want to reset, but if we did have that then we would want to move CantEquipWith_HiveBackpack to it's own class or move it to CustomItemSets.
+	[ReinitializeDuringResizeArrays]
+	public class WaspNestGlobalItem : GlobalItem
+	{
+		// This is a custom item set, facilitating mod collaboration. See CustomItemSets.cs for more information.
+		public static bool[] CantEquipWith_HiveBackpack = ItemID.Sets.Factory.CreateNamedBoolSetWithInfo(nameof(ExampleMod), nameof(CantEquipWith_HiveBackpack), false, "Accessories that can't be equipped alongside the Hive Backpack item", ItemID.HiveBackpack);
```
Chicken-Bones: As you pointed out, this example is kinda contrived, perhaps it's best to implement accessory conflict sets in tML. Two accessories which are both hive pack upgrades are incompatible themselves, not just incompatible with hive pack, and having a lot of globals like this wouldn't really scale well

ExampleMod/Content/Items/CustomItemSets.cs
```diff
@@ -0,0 +1,79 @@
...
+		// This method also exposes the set for other mods to access via this key. The key and default value must be consistent with other mods.
+		public static bool[] FlamingWeapon = ItemID.Sets.Factory.CreateNamedBoolSetWithInfo(FlamingWeaponCustomSetKey, false, "Causes \"Hahahah, burn!\" to randomly show in chat when used", ItemID.FieryGreatsword, ModContent.ItemType<ExampleSword>());
+		// Note that by using the ReinitializeDuringResizeArrays approach, ModContent.ItemType<ExampleSword>() is a valid input since modded content IDs will be assigned and retrievable during the reinitialization. Without ReinitializeDuringResizeArrays the code will incorrectly use 0 as the value of ModContent.ItemType<ExampleSword>() because modded IDs haven't been assigned yet.
+
+		// If a named ID set is specific to content in your mod, consider using the CreateNamedBoolSet(string modName, string key, ...) overload instead to avoid any potential conflicts with another mod that happens to use that same key for unrelated purposes.
```
Chicken-Bones: This documentation is not accurate anymore either

ExampleMod/Content/Items/CustomItemSets.cs
```diff
@@ -0,0 +1,79 @@
...
+				CombatText.NewText(Player.getRect(), Color.Red, "Hahahah, burn!");
+			}
+		}
+	}
+
+	// This command helps visualize the custom ID set data to verify its contents.
+	// Type /customsets in chat to view the contents of these sets
+	public class CustomSetsCommand : ModCommand
+	{
+		public override string Command => "customsets";
```
Chicken-Bones: conflict with the command we just added in tML? Probably not needed anymore? Maybe we want a way to dump all non-default values for a set too, but getting names is hard? Maybe not really necessary

patches/tModLoader/Terraria/ID/ArmorIDs.cs.patch
```diff
@@ -17,7 +17,7 @@
 +		public partial class Sets
  		{
 -			public static SetFactory Factory = new SetFactory(Count);
-+			public static SetFactory Factory = new SetFactory(EquipLoader.nextEquip[EquipType.Head]);
++			public static SetFactory Factory = new SetFactory(EquipLoader.nextEquip[EquipType.Head], $"{nameof(ArmorIDs)}+{nameof(Head)}");
```
Chicken-Bones: Perhaps best to use `.` instead of `+`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ID/ArmorIDs.cs.patch
JavidPack: These ones were `+` since they were an inner class. The `ChildSafety.Dust` and `ChildSafety.Gore` I put a period since they were a separate `SetFactory` within a class. I thought that `+` would make sense since that is how inner class's Types are named, but I don't mind either way, I was just trying to ensure no unintentional name conflicts.

patches/tModLoader/Terraria/ID/SetFactory.TML.cs
JavidPack: Made some more fixes here, only logs when actual changes happen.

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: Comments updated.

ExampleMod/Content/Items/Accessories/WaspNest.cs
JavidPack: Removed

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: Updated

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: I've removed this. The built-in command can now output non-default values. The only issue is we don't have a mapping to the `IdDictionary` instances like `ItemID.Search`, so it's not pretty printing the values with the content names, just numbers.

Chicken-Bones: I suppose we can add a name function to the constructors

patches/tModLoader/Terraria/ID/ArmorIDs.cs.patch
Chicken-Bones: Well people access inner classes via `.` as well, so it's probably simplest.

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: Added the IdDictionary to the ones that had one.

patches/tModLoader/Terraria/ID/ArmorIDs.cs.patch
JavidPack: Sure, done.

ExampleMod/Content/Items/CustomItemSets.cs
JavidPack: done.

PR #4384, opened 2024-09-21
Add ModConfig.Open

Destructor-Ben:
### What is the new feature?
Added `ModConfig.Open()` to allow modders to open their configs from places other than the config list.

Some modders may want to refer to their config in parts of the game such as UI or messages when the world loads, but they would have to instruct the user to find the config.

This allows modders to directly let users access configs.

[Discord Discussion](https://discord.com/channels/103110554649894912/445276626352209920/1285336534039531695)

No.

```cs
var config = ModContent.GetInstance<TheModConfig>();

// Basic opening
config.Open();

// Open without playing the menu open sound and with a custom action when the config is closed.
config.Open(playSound: false, onClose: () => {
        Main.NewText("Thanks for doing that, the current settings are now: " + config.SomeConfigEntry);
    }, scrollToOption: nameof(TheModConfig.SomeConfigEntry), centerScrolledOption: true, playSound: false
);
```

### ExampleMod updates

`ExampleModalUI` has been added (type `/modal` to test in-game). It has a button that opens the Mod Config menu for `ModConfigShowcaseDataTypes` and focuses on the `itemDefinitionExample` member. When closed, it re-opens `ExampleModalUI` to show the new config values.

`ExampleModalUI` also shows off using `IngameFancyUI` to streamline UI code for "modal" UI situations.

### Porting Notes

If you were previously opening a config in custom ways from places like UI, you can now use `ModConfig.Open()`.


JavidPack:
Here is `ExampleModalUI` showing off `ModConfig.Open` usage. Notice how the `ItemDefinition Example` entry is flashing and centered in the listing to bring attention to it. Also notice how `ExampleModalUI` hides all other gameplay UI similar to the settings and bestiary menus.

https://github.com/user-attachments/assets/1b81f2d3-af90-4207-baa1-b9ec6581c10f

PR #4397, opened 2024-10-02
Fix MouseText misalignment

TigerChenzzz:
### What is the bug?
When the mouse text contains tags (like [i:GoldBar]), it miscalculated the size of the string.
For example, if you write `Main.instance.MouseText("[i:MusicBoxOWUndergroundCorruption][i:MusicBoxOWUndergroundCorruption][i:MusicBoxOWUndergroundCorruption]"); ` in a draw method, and when you mouse is close to the right edge, you would find the text (or three music box) far away from the left of the mouse.
### How did you fix the bug?
Main.MouseTextInner uses `FontAssets.MouseText.Value.MeasureString` to measure the string size, but uses `ChatManager.DrawColorCodedStringWithShadow` to draw. I change the measure method to `ChatManager.GetStringSize` to match the draw method.
None.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -2726,7 +2726,7 @@
  			X += 34;
  
  		new Microsoft.Xna.Framework.Color(mouseTextColor, mouseTextColor, mouseTextColor, mouseTextColor);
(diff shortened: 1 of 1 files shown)


JavidPack:
Thanks, tested, works.

Issue #4396, opened 2024-10-01
Player.breatheMax does not Reset. Add it to Player.ResetEffects()


BugraPearls:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Player.breatheMax does not Reset its value in Player.ResetEffects(), and its coded to be 200. 

I propose adding it to Player.ResetEffects(), setting it to 0, and in Update() giving it the 200 value back.

(Out of scope of this issue but some documentation/description regarding how breath works is also good, otherwise requires decent vanilla code reading)

### What does this proposal attempt to solve or improve?

This proposal should resolve any possible conflicts, and bring convenience to the Modder, since a Modder will not have to set their own ResetEffects() code for this spesifically if they want to modify it in any way, and if there is another mod that does something with this value, they won't have to hard-lock the Reset value to be 200 inside their mod, instead, the game itself can do that and Modders can simply += the Player.breatheMax value.

direwolf420:
Can confirm this is the case. Thorium has a few things increasing maximum breath, and it resets the cap back to 200 in ModPlayer.PostUpdateEquips manually.
This change _may_ be causing things to not work for mods expecting it to not reset, would need to be tested

JavidPack:
Sounds good. I also agree that documenting breath and others would be good. I don't imagine resetting it in tModLoader would cause issues for mods.

Feel free to let us know if you run into any issues setting up the project. We can help here or on discord.

BugraPearls:
I'd like to contribute regarding this, both with documenting and the Resetting of breatheMax, however, truth be told I have very little experience with GitHub and haven't done this before, so what should I do from here?

I think I have to create a Pull Request and do changes, and u guys will accept it but I am a bit confused, I really don't know what to do there.

JavidPack:
So, since tModLoader is built on patches to the source code, things are a little different from a normal GitHub project. We have the full instructions shown in https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-contributors, give them a glance and see if you are up to it or not.

The process is a bit involved if you are not familiar with Git and Github, unfortunately. If you don't feel up to doing those steps, and since this change is fairly simple, I can make the changes for you instead and I'll mark you as the co-author if you'd like. You'll just need to type out the exact changes here and the documentation text you want added.

[...]

PR #4399, opened 2024-10-03
Added Player.breathEffectiveness to facilitate "extends underwater breathing"

JavidPack:
Modders currently adjust `Player.breath` to fake an effect similar to diving helmet, as mentioned in #4396. https://github.com/tModLoader/tModLoader/commit/0d55bc751d5b2b7c5fabb5ea5e3a028a2215283a fixed some issues with doing that for current modders, but complete support would allow for better control.

Currently, adjusting `Player.breath` can linearly lengthen underwater breathing time, but the effect isn't numerically equivalent to vanilla approaches. Vanilla approaches are multiplicative, which currently isn't possible to do correctly due to mod execution order. Also, adjusting `Player.breath` past 400 results in a 2nd row of bubbles that keeps extending, which is not ideal.

This PR adds `Player.breathEffectiveness` as a `StatModifier`, allowing for base/additive/multiplicative adjustments to breathing time without adjusting `Player.breath`, which would be incompatible with multiple mods doing conflicting effects.

As a result, `Player.breath` now represents a breath capacity, while `Player.breathEffectiveness` represents a breath efficiency. Combined this should allow for, modders to individually adjust both. For example, a race with low breath capacity, an accessory doubling breathing efficiency, and an accessory increasing breathing capacity should all work together correctly no matter the load order.

For most modders, they'll want to switch from adjusting `Player.breath` to adjusting `Player.breathEffectiveness` instead for compatibility and avoiding the strange bubble indicator issue.

Also, `Player.accDivingHelm` applies a 6x multiplicative effect and Breathing Reed applies a 100% additive effect. We recommend modders stick to additive effects for the most part unless they are making an upgrade to diving helmets, otherwise the multiplicative effects will scale very quickly.

Sample usages:
```
public override void UpdateAccessory(Player player, bool hideVisual) {
	player.breathMax += 200; // double breath capacity.
	player.breathEffectiveness += .5f; // Increase breath efficiency by 50%, additively.
```

### DIFF
--- a/patches/tModLoader/Terraria/Player.TML.cs
+++ b/patches/tModLoader/Terraria/Player.TML.cs
@@ -85,6 +85,8 @@ public partial class Player : IEntityWithInstances<ModPlayer>
 	/// </summary>
 	public const int ManaCrystalMax = 9;
 
+	public StatModifier breathEffectiveness = StatModifier.Default;
+
 	public RefReadOnlyArray<ModPlayer> ModPlayers => modPlayers;
 
 	RefReadOnlyArray<ModPlayer> IEntityWithInstances<ModPlayer>.Instances => modPlayers;
--- a/patches/tModLoader/Terraria/Player.TML.cs
+++ b/patches/tModLoader/Terraria/Player.TML.cs
@@ -86,7 +86,7 @@ public partial class Player : IEntityWithInstances<ModPlayer>
 	public const int ManaCrystalMax = 9;
 
 	/// <summary>
-	/// How effectively the player can hold their breath underwater. Controls how long it takes for <see cref="breath"/> to decrease. Breathing Rod doubles this value and Diving Gear multiplies it by 6.
+	/// How effectively the player can hold their breath underwater. Controls how long it takes for <see cref="breath"/> to decrease. Breathing Reed adds 1 (100%) to this value and Diving Gear multiplies it by 6.
 	/// <para/> Modded effects should add to this instead of multiplying to avoid values getting unreasonable large. Adding 1.5f for example will increase breath time by 150%.
 	/// <para/> Applied in the calculation of <see cref="breathCDMax"/>.
 	/// </summary>
--- a/patches/tModLoader/Terraria/Player.TML.cs
+++ b/patches/tModLoader/Terraria/Player.TML.cs
@@ -88,6 +88,7 @@ public partial class Player : IEntityWithInstances<ModPlayer>
 	/// <summary>
 	/// How effectively the player can hold their breath underwater. Controls how long it takes for <see cref="breath"/> to decrease. Breathing Rod doubles this value and Diving Gear multiplies it by 6.
 	/// <para/> Modded effects should add to this instead of multiplying to avoid values getting unreasonable large. Adding 1.5f for example will increase breath time by 150%.
+	/// <para/> Applied in the calculation of <see cref="breathCDMax"/>.
 	/// </summary>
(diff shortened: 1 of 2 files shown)


TacoConKvass:
Can we get this anytime soon?

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -85,6 +85,8 @@ public bool MeleeEnchantActive {
 	/// </summary>
 	public const int ManaCrystalMax = 9;
 
+	public StatModifier breathEffectiveness = StatModifier.Default;
+
```
Chicken-Bones, line 94: Needs docs. Explain that an effectiveness of 2 means breath lasts twice as long. Recommend modders use additive? 

Is there something without the Base/Flat?

JavidPack: Vanilla is all multiplicative, but that can get out of hand pretty quickly I imagine. I'm not seeing a class like StatModifier without base/flat. We have AddableFloat and MultipliableFloat classes.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1107,10 +1107,19 @@
 +	/// <summary> How long it takes for <see cref="breathCD"/> to cause <see cref="breath"/> to decrease by 1. Defaults to 7 and is increased by factors such as holding a Breathing Reed, which doubles this, or having a Diving Gear, which multiplies this by 6, to make the process of losing breath value slower. </summary>
```
Chicken-Bones: Change doc to simply say, `How long it takes for <see cref="breathCD"/> to cause <see cref="breath"/> to decrease by 1. Defaults to 7 before <see cref="breathEffectiveness"/> is applied`

patches/tModLoader/Terraria/Player.TML.cs
Chicken-Bones: Okay, I'm happy with `StatModifier`. Please mention that this is applied in the calculation of `breathCDMax`

direwolf420:
I'm curious, what's preventing this from being merged?

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1107,10 +1107,19 @@
...
  	public int breathCDMax {
  		get {
-+			// TODO: hook or StatModifier breathEffectiveness
- 			int num = 7;
+-			int num = 7;
++			int num = 1;
  			if (inventory[selectedItem].type == 186 && itemAnimation == 0)
  				num *= 2;
```
Chicken-Bones: Should we make breathing reed additive?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Player.cs.patch
JavidPack: Done

patches/tModLoader/Terraria/Player.TML.cs
```diff
@@ -85,6 +85,13 @@ public bool MeleeEnchantActive {
 	/// </summary>
 	public const int ManaCrystalMax = 9;
 
+	/// <summary>
+	/// How effectively the player can hold their breath underwater. Controls how long it takes for <see cref="breath"/> to decrease. Breathing Rod doubles this value and Diving Gear multiplies it by 6.
```
Chicken-Bones: `Breathing Reed adds 1 (100%)`

PR #4403, opened 2024-10-07
Allow Player.Hurt to be cancelled

Xetaxheb:
### What is the new feature?
This is a convenience feature, making it easier for mods to make a player situationally ignore certain damage instances.

Adds `HurtModifiers.Cancel()` and `HurtInfo.Cancelled`. Cancelling a hurt has the same effect as returning `true` from `FreeDodge`, though even undodgeable hurts can be cancelled. Unlike `FreeDodge`, immune frames are not applied, so the player can still be hit by another attack this tick, or the same attack next hit.

This is already possible, or should've been if not for a bug, failing to call `FreeDodge when `HurtInfo.Dodgeable` was overridden to `true`. This change just makes it easier. This is preferable to otherwise requested '0 damage' hits.

- Should we apply i-frames? Perhaps an argument to `Cancel` could indicate i-frame duration?
- Some mods may want to trigger on-hit effects, and just negate the damage portion.

```cs
	public virtual void ModifyHurt(ref Player.HurtModifiers modifiers)
	{
		if (condition)
			modifiers.Cancel();
	}
```

### ExampleMod updates
None

### Porting Notes
None. Modders do not need to check if a `HurtInfo` is `Cancelled`, tML will not be passing cancelled hurts to hooks.

### DIFF
--- a/patches/tModLoader/Terraria/Player.TML.Hurt.cs
+++ b/patches/tModLoader/Terraria/Player.TML.Hurt.cs
@@ -212,11 +212,11 @@ public partial class Player
 		/// <summary>
 		/// The amount of damage received by the player. How much life the player will lose. <br/>
 		/// Is NOT capped at the player's current life.<br/>
-		/// Cannot be set to less than 1.
+		/// Cannot be set to less than 0.
 		/// </summary>
 		public int Damage {
 			readonly get => _damage;
-			set => _damage = Math.Max(value, 1);
+			set => _damage = Math.Max(value, 0);
 		}
 
 		/// <summary>
--- a/patches/tModLoader/Terraria/Player.TML.Hurt.cs
+++ b/patches/tModLoader/Terraria/Player.TML.Hurt.cs
@@ -107,6 +107,13 @@ public partial class Player
 		/// </summary>
 		public MultipliableFloat KnockbackImmunityEffectiveness = new();
 
+		private bool _cancelled = default;
+		/// <summary>
+		/// Cancels the Hurt. Further hooks like <see cref="ModPlayer.FreeDodge"/> and <see cref="ModPlayer.OnHurt(HurtInfo)"/> will not be called. <br/>
+		/// Does not automatically apply immune frames, so the player can get hit again next frame. 
+		/// </summary>
+		public void Cancel() => _cancelled = true;
+
 		private bool _dustDisabled = default;
 		/// <summary>
 		/// Prevents dust from spawning
@@ -161,6 +168,7 @@ public partial class Player
 				SourceDamage = (int)SourceDamage.ApplyTo(damage),
 				Damage = (int)GetDamage(damage, defense, defenseEffectiveness),
 				Knockback = GetKnockback(knockback, knockbackImmune),
+				Cancelled = _cancelled,
 				DustDisabled = _dustDisabled,
 				SoundDisabled = _soundDisabled,
 			};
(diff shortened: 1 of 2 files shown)


Xetaxheb:
https://github.com/tModLoader/tModLoader/blob/f773d097be298d2db22d4afec5e6f06b7ed17f05/patches/tModLoader/Terraria/Player.TML.Hurt.cs#L219

https://github.com/tModLoader/tModLoader/blob/f773d097be298d2db22d4afec5e6f06b7ed17f05/patches/tModLoader/Terraria/Player.TML.Hurt.cs#L141

Chicken-Bones:
Every attack does at least 1 damage is a rule in Terraria. There was an extended discord discussion about the reasoning for this limitation, including things that could break and the benefits of mods being able to rely on it. If you want to ignore damage entirely, you can heal it back `OnHit`, and optionally suppress the combat text. You can read the discussion before and after this message https://discord.com/channels/103110554649894912/1067667472687312987/1083540587627679896

If you still want the feature, please tell us why you want it. What is your mod trying to achieve by setting `Damage` to `0`? When do you do it? There may be other ways to enable your end goals

Xetaxheb:
I understand where you're coming from but I feel my point still matters of: It's already enforced in HurtModifiers.GetDamage.
If one is overriding that then it is clearly deliberate.

My specific use case is to have exactly what was happening with 0 damage hits happen when one has enough defense to overcome it. That had numerous gameplay uses I saw from people including especially 1 hp challenges, where they could for example eliminate slimes as a threat by stacking enough defense. (my mod has multiple defense "thresholds" and was not as straightforward as "instead of 1 damage you take 0")

I can vaguely see several other use-cases including ones similar to what calamity's dev was bringing up; but at the end of the day i'm just not seeing a good reason to be **_clamping it on L219 in a getter/setter_** (to 1, whereas 0 to prevent negative/healing makes sense) specifically let alone in a way that's a hassle to override. It's already covered in the GetDamage, as it should be for the desired "keep damage hits at 1", so it doubling up on that isn't beneficial.

On top of everything else the only places it's even used in terraria/tml are calling GetDamage anyway https://github.com/tModLoader/tModLoader/blob/2569fc2ce3c862697034cd91e455c769f9f55517/patches/tModLoader/Terraria/Player.TML.Hurt.cs#L162 and receiving data (which shouldn't be 0 unless deliberate) https://github.com/tModLoader/tModLoader/blob/2569fc2ce3c862697034cd91e455c769f9f55517/patches/tModLoader/Terraria/MessageBuffer.cs.patch#L707

Hurt() already returns 0 when immune so there's really nowhere except interfering in my or others use-cases that this being clamped to 1 even does anything.

Overriding onhit and suppressing text I suppose is one way to go about it but that seems much more conflict prone and less reliable than what was previously working fine. I would probably override Hurt() to cancel the hit instead for better compatibility. There are a number of alternative solutions but what I was doing touches the least things and has the most compatibility to do exactly what is advertised.

Chicken-Bones:
Thankyou for explaining more about your use case. The case for players is a bit better than the case for NPCs. Players already have a relationship where `FreeDodge` can cancel the attack, including `OnHurt` effects, after the `HurtInfo` has been fully calculated. The only issue with using `FreeDodge` for your use case is that you have no control over the `if (dodgeable)` line in `Player.Hurt`. I think this is actually a bug, and should be changed to `if (info.Dodgeable)` since `HurtInfo.Dodgeable` is freely modifiable. 

Since your mod already has a way to calculate whether a given hurt should be 'ignored' or not, and you're not proposing a change to the general rule that attacks can only do a minimum of 1 damage this might be an ideal solution

Chicken-Bones:
After further discussion in Discord ([message](https://discord.com/channels/103110554649894912/1067667472687312987/1299630001112743936)) we have agreed to add `HurtInfo.Cancelled` alongside `HurtModifiers.Cancel()`.

Xetaxheb:
Looks great, per discord discussions 👍

Issue #1356, opened 2021-03-12
lime modpack color


tomiscool850:
its not really needed but a cool suggestion if you enable more mods before putting them on the modpack the color of the text then will turn into a lime color/light green, kinda like how when your missing a mod it turns red and when you disable a mod it turns yellow. I don't know I just feel like it fits

Hotmud9:
The problem with that is that some mods, like MrPlague's Race Mod, have their own custom colors.

Cerbrus:
Assuming you're talking about the `1 total, 1 enabled,....`  line at the top right, this is already the case. That line turns green when the mods are enabled, yellow when they're disabled:

![image](https://github.com/tModLoader/tModLoader/assets/2911988/7fb7671c-f352-42c3-88a8-6eb791394d13)
![image](https://github.com/tModLoader/tModLoader/assets/2911988/ae5bded7-2040-4c6f-b58e-69e7722f4f1e)
![image](https://github.com/tModLoader/tModLoader/assets/2911988/dc27dc71-d8c7-44dc-924c-c4c1e3c2dcb1)

(@JavidPack I think this issue is resolved?)

steviegt6:
They're asking for an indicator when you have extra mods enabled that aren't part of the modpack.

TechnicalityCreations:
I don't mind doing it

TechnicalityCreations:
<img width="386" alt="image" src="https://github.com/user-attachments/assets/276e1b0f-3124-4feb-95ad-54fa093bd48c">
Done

PR #4404, opened 2024-10-08
Lime Green Mod Pack statuses for mod packs with additional mods enabled

TechnicalityCreations:
image

### What is the new feature?
As described in #1356, the text in the status of a mod pack turns lime green if the mod pack is enabled and other mods are, too.
Because, if you are playing with a specific mod pack, you might want to know if you have accidentally enabled additional mods.
Having a text based indicator
### ExampleMod updates

None
### Porting Notes

None


JavidPack:
Thanks.

Issue #3597, opened 2023-07-04
Minecart Upgrade Kit change to make it more compatible with modded carts


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

The [Minecart Upgrade Kit](https://terraria.wiki.gg/wiki/Minecart_Upgrade_Kit) sets fixed values for movement stats on carts. This is fine in vanilla because every cart has the exact same movement stats. In modded it proves problematic because changes to those default stats will not reflect with the Kit.

I propose a solution that will instead scale the default values via a factor, or specify the override values directly (both 
customizable as fields in `MountData`). This also allows for balancing.

### What does this proposal attempt to solve or improve?
Currently, if you want to have your changes scale properly, you need to use `ModPlayer.PostUpdateRunSpeeds`, check for the mount, check for the kit being active, and then adjust the relevant movement fields on the player.

### Which (other) solutions should be considered?

See Description

TechnicalitiesDevelopment:
I’ll do it

TechnicalityCreations:
I wouldn’t mind implementing it. I would suggest that we take the latter option and just add a parameter to the `Mount.MountData` type, as different base speeds exist for different types of vanilla minecarts, making it hard to create a factor.

JavidPack:
Good work, there are some things that will have to be changed to prevent breaking existing mods, but we'll fix those when we finalize the design.

I'm not sure if the approach presented here is flexible enough for what modders want, so I've made a discussion thread on our Discord in the development forum (https://discord.com/channels/103110554649894912/1294026615910699160). If you want to participate in that discussion as it happens, ping me on Discord so I can add you to the thread, I didn't see a user with your name when I checked.

TechnicalityCreations:
I don’t use discord. Feel free to just tell me what happens

JavidPack:
Fixed in #4407

PR #4407, opened 2024-10-10
New MountData params to allow Minecarts to have custom stats after Minecart Upgrade

TechnicalityCreations:
### What is the new feature?
The `MountData` class now contains 5 new parameters: `MinecartUpgradeRunSpeed`, `MinecartUpgradeDashSpeed`, `MinecartUpgradeJumpSpeed`, `MinecartUpgradeAcceleration` and `MinecartUpgradeJumpHeight`. These replace the previous SuperCart versions of these parameters in the `Mount` class, which were static and the same for all minecarts. This means that modded minecarts with custom stats can scale up these stats when the Minecart Upgrade Kit is used. See #3597.
Because this makes modders' lives easier when creating minecarts with custom stats. It allows them to keep any bonuses when upgrading minecarts as well as adding a way to customize existing minecarts.
Yes, but they are overly complex for little benefit.
See the description of #3597 to see how it would be possible.
From an updated `ExampleMinecartMount`:
`MountData.MinecartUpgradeRunSpeed = 100f;
MountData.MinecartUpgradeDashSpeed = 100f;
MountData.MinecartUpgradeAcceleration = 50f;`
### ExampleMod updates

The file: `ExampleMinecartMount.cs` has been updated to show off the new parameters.
### Porting Notes

None

### DIFF
--- a/ExampleMod/Content/Mounts/ExampleMinecartMount.cs
+++ b/ExampleMod/Content/Mounts/ExampleMinecartMount.cs
@@ -25,8 +25,10 @@ namespace ExampleMod.Content.Mounts
 			MountData.delegations.MinecartDust = DelegateMethods.Minecart.SparksMeow;
 			MountData.delegations.MinecartLandingSound = DelegateMethods.Minecart.LandingSoundFart;
 			MountData.delegations.MinecartBumperSound = DelegateMethods.Minecart.BumperSoundFart;
-
-			// Important to note is that runSpeed, dashSpeed, and acceleration will get overridden when the player has used the Minecart Upgrade Kit. Keep that in mind when changing the values yourself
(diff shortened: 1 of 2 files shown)


JavidPack:
I noticed the branch got deleted, was that intentional?

TechnicalityCreations:
Oops, I’m an idiot. I accidentally created another branch but then deleted this by mistake. Branch restored

TechnicalityCreations:
Thanks for telling me

JavidPack:
Looks good. 

One change I made was marking the existing public fields as obsolete. We need to do that to prevent mods from suddenly breaking when the game updates, since removing a field breaks binary compatibility.

PR #4419, opened 2024-10-15
Added the shorthand for NPC.Target

AlienBlox:
I added the shorthand for NPC.Target for quick uses


JavidPack:
Adding shortcuts like this to the ModX classes will just add confusion to modders for the fields that aren't duplicated and will quickly get out of hand if we try to decide on which fields are the most common fields to add getters for, so we won't be making this change to tModLoader.

Issue #3643, opened 2023-07-20
Increase requirements for Workshop publishing


Mirsario:
<p align="center">
<img src="https://github.com/tModLoader/tModLoader/assets/6795251/91da936d-568e-4f44-b30c-a457ea132354" />
</p>

Since the Mod Browser has been moved to Steam Workshop, it has been nearly flooded with various low effort submissions, and even high effort ones tend to look lame due to only 80px icons being suggested.
This issue lists suggestions that could help mitigate these problem.

### Requirements
- [x] The mod initializes correctly.
	This is already handled by the fact that the `Publish` button only shows up for mods that are loaded.
	Might be worth keeping this in mind for various "Github CI" ordeals of the future.
- [x] **description.txt**
	Require a non-default ("X is a pretty cool mod") description.txt with length of at least ~20 characters.
	If the required character count is increased further - it should be noted that we'll probably need to count different languages' characters differently, as some languages (Chinese, Japanese, perhaps Korean) are known to be far more "compact".

	- [x] Additionally, we should change the auto-generated description to mention that it must be changed.
- [ ] **Assembly-checking**
	Require the published mods to contain anything that differs them from what building "Create mod" outputs would produce.
	This may be a topic of its own, but it's helpful that the mod assemblies must be loaded and running before publishing.
- [x] **icon.png**
	Disallow using the empty frame icon added to new mods by default.
- [x] **icon_workshop.png**
	Require a special workshop icon with the minimal size of `480x480` px. `512px` is prefered, but `480` is what you get if you want `40/80px` to scale up well.
	Alternatively, should we just upscale `icon.png` to 480x480?

### Not to be required
- **icon_small.png**
	A few mods would benefit from other mods always having a small icon to be used in GUIs, but they could just use a fallback texture if they don't.

- **description_workshop.txt**
	Not a serious necessity if we make mod info open workshop pages, thus making `description.txt` already mostly be used for bb-code.

GlasiaVD23:
Hello, 

first: I am not sure if I am allowed to write this here.

I used the DragonLens mod together with my own mod (not published) where I made a prefix that tries to boost max minion count to test said prefix. When DragonLens tried to load the icon_small which I didn't have, it errored out. Took me some days to find this issue until I stumbled upon this thread. 

[...]

PR #4424, opened 2024-10-19
Increase Workshop Publication Requirements

TechnicalityCreations:
### What is the new feature?
It increases the checks made before publishing on the workshop to make it impossible to publish mods without changing their icon or description. It also checks for a description with fewer than 50 letters/digits to publish and scales up the icon.png file when publishing. If the workshop icon is from another folder it will be copied over to the mod sources folder as well.
See #3643
Heavy moderation of the workshop
### Porting Notes
If they have a mod with the default description or icon then they have to change that prior to updating.


Mirsario (approved):
- `icon_workshop.png` is no longer required but will be automatically created from the chosen preview image, upscaling it to at least `480x480` if needed. Users are notified of this in GUI: 
image
- Descriptions must be present and have at least not 20 but **50** characters, counting only letters and digits.

TechnicalityCreations:
Thank you, I will edit the description

Issue #3787, opened 2023-09-03
Modded Banners Not Placeable on Platforms


Solxanich:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

Modded Banners Not Placeable on Platforms

In 1.4.4, Terraria added the ability to place banners on platforms.
Our banner code doesn't appear to have caught up to match.

direwolf420:
To elaborate, this is not a bug in any way. ExampleBanner does not exist so modders can only copy 1.3's version and port that, which lacks some critical anchor data for altenates.

The code to change/add would be:
All the tileobjectdata stuff to just:
```cs
TileObjectData.newTile.CopyFrom(TileObjectData.GetTileData(TileID.Banners, 0));
TileObjectData.addTile(Type);
```
`TileID.Banners` needs to be `TileID.HangingLanterns` for ExampleLamp, with `TileObjectData.newTile.StyleLineSkip = 2; //If you need it so tiles drop the correct items for multiple frame states per style` added.

Then, inside the ModTile
```cs
public override void SetDrawPositions(int i, int j, ref int width, ref int offsetY, ref int height, ref short tileFrameX, ref short tileFrameY)
{
    Tile tile = Main.tile[i, j];
    TileObjectData data = TileObjectData.GetTileData(tile);
    int topLeftX = i - tile.TileFrameX / 18 % data.Width;
    int topLeftY = j - tile.TileFrameY / 18 % data.Height;
    if (WorldGen.IsBelowANonHammeredPlatform(topLeftX, topLeftY))
    {
        offsetY -= 8;
    }
}
```

JavidPack:
Related to #3442

JavidPack:
Fixed in #4442

Issue #3460, opened 2023-05-07
ExampleMod : ExampleChandeliers Should Be Provide


Omnielement:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

I think ExampleMod needs to provide an ExampleChandeliers just like Chandeliers in 1.4 Terraria.
*It will tilt in the wind but keep silence in front of the wall.
*It will swing violently when players swept through it swiftly.
*It also needs to suit with colorful paints.
*It may vary in physical structure, like sandstone chandelier, copper chandelier, crystal chandelier, and stardust chandelier.

I have tried to make modChandeleier by myself, but it can't suit with paints.(Referring to dfgasdfg/Everglow/Sources/Modules/Myth/TheFirefly/Tiles/Furnitures/GlowWoodChandelier)

### What does this proposal attempt to solve or improve?

Making a mod chandelier is really hard.This can also standardize modders making chandeliers. 

### Which (other) solutions should be considered?

Mod vines, swing like 1.4Terraria.
According to the vines behaved physical-badly  in vanilla, this should be take more consider.
(When the vine too long, you can see an unforgettable sight)

Omnielement:
![Chandeliers](https://user-images.githubusercontent.com/103014909/236686373-59ce414c-d502-4338-823f-c433745f498d.gif)

Omnielement:
![Chandeliers](https://user-images.githubusercontent.com/103014909/236686495-dbc35ee3-23c0-473a-8d18-b926386d8f21.gif)
This gif shows my chandeliers.

JavidPack:
I'll revisit this when implementing #3442

JavidPack:
So, it's not a chandelier, but in https://github.com/tModLoader/tModLoader/pull/4429 I updated `ExampleAnimatedTile`/"Red Firefly in a Bottle" to support swaying. It is a lantern, 1x2, but the code should work all the same for a chandelier tile with no modifications. ExampleLamp has the code for toggling state to turn off the lights.

Closing this now, but we wouldn't reject an ExampleChandelier if anyone wanted to submit one sometime.

Issue #3442, opened 2023-04-29
Port Banners


tomiscool850:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

so currently there are no example banners for example mod and i feel that this is something that should be in example mod already, so then we can easily make banners for custom enemys

i suggest you do a banner for the following enemys
-example worm
-example zombie Theif
-party zombie
-example Custom AI Slime NPC

if you ever port octopus or seal npcs from 1.3 example mod then you can add banners for that to

### What does this proposal attempt to solve or improve?

this would solve the banner problems for modded enemys

JavidPack:
This has been on my todo list for a while, don't worry, it'll happen.

JavidPack:
Fixed in https://github.com/tModLoader/tModLoader/pull/4442

PR #4429, opened 2024-10-24
Multi-Tile Wind Sway Capabilities

### LINKED ISSUE #2975: [Feature Request] A way to make modded tiles sway in wind/on player movement

No

Mod capability as a Modder

### Description

The code responsible for drawing tiles that "sway" (be it from wind or when a player walks by) is currently super hardcoded to only support vanilla tiles (either by vanilla type checks, or the variables related to sway being private). A way to either automatically handle this through sets, and/or (optionally) through hooks, on a tml level would benefit modders greatly.

Make it much easier to implement swaying tiles, such as lanterns, vines, banners, etc.

No solution proposed directly here

MutantWafflez:
### What is the new feature?
Resolves #2975. Also related: #3442, #3787 and #3460.

Allows modders to "register" their multi-tiles to sway in the wind and with player interaction. 

Vines and reverse vines should now be easier to implement as documentation has been added for these and some methods have been made public (`CrawlToTopOfVineAndAddSpecialPoint`/`CrawlToBottomOfReverseVineAndAddSpecialPoint`). There are no examples yet, however.

`TileObjectData.IsTopLeft` has been added to facilitate detecting the top left tile of a multitile, which is necessary for using the special tile drawing options.

Before this PR, modders could only use `TileID.Sets.WindSwayBasic`, which only functioned properly on single, 1x1 tiles. Using this set with any multi-tiles would not trigger the proper drawing for wind sway with multi-tiles.

Instead of using a set, some hooks could likely be designed for more fine-tuned control.

See ExampleMod's new `ExampleWideBannerTile`. Shows how one would create a 2x3 banner with the proper wind sway, similar to the War Banner from vanilla.

`ExampleAnimatedTile`/"Red Firefly in a Bottle" has also been updated. It is a lantern tile and now sways.

### ExampleMod updates
See above (`ExampleWideBannerTile` and `ExampleAnimatedTile`).


patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -653,3 +696,22 @@
...
++		for (int i = 0; i < sizeX; i++) {
++			if (WorldGen.IsBelowANonHammeredPlatform(topLeftX + i, topLeftY)) {
++				continue;
++			}
++
++			isBelowNonHammeredPlatform = false;
++			break;
++		}
+-		if ((type != 465 && (uint)(type - 591) > 1u) ? (sizeX == 1 && WorldGen.IsBelowANonHammeredPlatform(topLeftX, topLeftY)) : (WorldGen.IsBelowANonHammeredPlatform(topLeftX, topLeftY) && WorldGen.IsBelowANonHammeredPlatform(topLeftX + 1, topLeftY))) {
++		//if ((type != 465 && (uint)(type - 591) > 1u) ? (sizeX == 1 && WorldGen.IsBelowANonHammeredPlatform(topLeftX, topLeftY)) : (WorldGen.IsBelowANonHammeredPlatform(topLeftX, topLeftY) && WorldGen.IsBelowANonHammeredPlatform(topLeftX + 1, topLeftY))) {
```
Chicken-Bones: Use block comment around this line of code

patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -14,8 +14,20 @@
 -public class TileDrawing
 +public partial class TileDrawing
  {
- 	private enum TileCounterType
+-	private enum TileCounterType
++	public enum TileCounterType
```
Chicken-Bones, line 18: Does this need to be `public`?

JavidPack:
Thanks a ton. I'll be finishing up ExampleBanner soon with these changes. (https://github.com/tModLoader/tModLoader/pull/4442)

PR #4430, opened 2024-10-24
Bestiary Categorization Control

MutantWafflez:
### What is the new feature?
Allows modders to "categorize" a custom bestiary info element, giving more control over where the element will be drawn in a given Bestiary entry.

Previously, any additions to a bestiary entry with a custom element would place that element at the very bottom of the entry, due to how vanilla "categorizes" various parts of an entry. This unfortunately is hard-coded to specific classes with a private hard-coded `enum`, so in order to "categorize" your element, you'd have to IL edit/detour.

Probably, but the usage of an interface is in line with how vanilla has structured the bestiary, so I think this design fits just fine.

See ExampleMod's new `ExampleBestiaryGlobalNPC`, which shows this categorization in action.

### ExampleMod updates
See above ^ (`ExampleBestiaryGlobalNPC`).


JavidPack:
Very nice, thanks.

PR #4437, opened 2024-10-29
Add ExampleSandRod to ExampleMod

Silencersn:
Add ExampleSandRod which is a sand version of Dirt Rod.
It cound be used for move sand blocks including mod sands.
Also showcase the conversion from mouse position to tile position when the screen is reversed.

Probably not.

### Other matters
Why isn't the Ash Block in the TileID.Sets.FallingBlockProjectile


JavidPack:
Looks good. I made a change to CanUseItem to keep the multiplayer behavior consistent with dirt rod. 

I noticed that the projectiles don't seem to spawn on other clients, I'm not sure why that is, might be something to do with how these projectile sync.

Ash isn't in FallingBlockProjectile since the logic for that block falling in special circumstances is custom, not default.

PR #4439, opened 2024-10-31
Bash launch script fixes

emi-YFcz5XDj:
### What is the bug?
1. `start-tModLoader-FamilyShare.sh` doesn't cd to script directory, and will fail when started from elsewhere.

2. `start-tModLoader.sh` doesn't exit explicitly when cd fails, which is inconsistent with `start-tModLoaderServer.sh`.

3. `start-tModLoaderServer.sh` exits immediately when cd fails, which causes certain terminals to close immediately.

### How did you fix the bug?
All launch scripts now execute this instead:
```
cd "$(dirname "$0")" ||
read -n 1 -s -r -p "Can't cd to script directory. Press any button to exit..." && exit 1
```

There are, but they are likely more intrusive changes.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/start-tModLoader.sh
+++ b/patches/tModLoader/Terraria/release_extras/start-tModLoader.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-cd "$(dirname "$0")"
+cd "$(dirname "$0")" ||
+read -n 1 -s -r -p "Can't cd to script directory. Press any button to exit..." && exit 1
 
 chmod a+x ./LaunchUtils/ScriptCaller.sh
 ./LaunchUtils/ScriptCaller.sh "$@" &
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/release_extras/start-tModLoader.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-cd "$(dirname "$0")"
```
Solxanich, line 2: The primary concern with directly exit-ing is that it does not tell the user why it failed nor us that the issue was with the CD execution.

As an alternative, I'd generally prefer instead of directly exitting that we write to the terminal and then request input from user to acknowledge that it failed (or give a chance for a screenshot). 

Especially important since Mac users are also involved with this same script and not confident their terminal would stay visible on exit

Solxanich (changes requested):

Chicken-Bones:
This seems fine, but are we aware of any situations where `cd` actually can fail? I don't want to maintain code for no reason

emi-YFcz5XDj:
> This seems fine, but are we aware of any situations where `cd` actually can fail? I don't want to maintain code for no reason

Mostly custom setups like symlinks, which can potentially be broken.

Worst case scenario the scripts will grant execute permission and run the wrong `ScriptCaller.sh`.

This might be too far fetched though, I'm not really sure.

PR #4441, opened 2024-11-03
Update vulnerable packages

tygyh:
### What is the bug?
There are several outdated plugins which are open to exploitation through DoS and ReDoS attacks.

### How did you fix the bug?
I updated the plugins to their latest versions.

Writing the fixes yourself.

### DIFF
--- a/tModPorter/tModPorter/tModPorter.csproj
+++ b/tModPorter/tModPorter/tModPorter.csproj
@@ -10,9 +10,9 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Locator" Version="1.7.8" />
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.9.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.12.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.FlowAnalysis.Utilities" Version="2.9.6" />
-		<PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.9.2" />
+		<PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.12.0" />
 		<PackageReference Include="UTF.Unknown" Version="2.5.0" />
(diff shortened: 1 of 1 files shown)


Solxanich:
Just to ensure we understand the issue correctly, would you please add a link regarding why those particular system packages are outdated?

tygyh:
> please add a link regarding why those particular system packages are outdated

Here are the articles for the vulnerabilities fixed

System.Text.Json:
[Inefficient Algorithmic Complexity](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMTEXTJSON-8168848)

System.Net.Http:
[Denial of Service (DoS)](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60045)
[Improper Certificate Validation](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60046)
[Privilege Escalation](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60047)
[Authentication Bypass](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60048)
[Information Exposure](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-72439)

System.Text.RegularExpressions:
[Regular Expression Denial of Service (ReDoS)](https://security.snyk.io/vuln/SNYK-DOTNET-SYSTEMTEXTREGULAREXPRESSIONS-174708)

Chicken-Bones:
While I do like minimising warnings and keeping packages up to date, none of these vulnerabilities are relevant to the capacity in which we use the packages. For example, `System.Net.Http` is just a transitive dependency of one of the test framework packages, and our tests don't execute any networking. 

I'm uncomfortable with adding explicit versions of transitive dependencies to our `.csproj` files, because then we forget why they're there. At the very least, the new lines you've added need to be annotated with a `<!--- Temporary transitive dependency version bump due to security advisory #4441--->`

Ideally, we would just do version bumps on the dependent packages instead.

tygyh:
> Ideally, we would just do version bumps on the dependent packages instead.

In that case I think we should wait a little bit since version 4.12 of Workspaces and MSBuild will be released soon.

tygyh:
> In that case I think we should wait a little bit since version 4.12 of Workspaces and MSBuild will be released soon.

4.12 was released on December 4. I've changed the PR to just update Workspaces and MSBuild. We'll see if this fixes the underlying problems

PR #4457, opened 2024-11-28
Modernize language usage

tygyh:
Use modern language features to increase readability.

### DIFF
--- a/tModPorter/tModPorter/Config.Terraria.cs
+++ b/tModPorter/tModPorter/Config.Terraria.cs
@@ -267,9 +267,6 @@ public static partial class Config
 	private static void AddTextureRenames() {
 		var getValue = AccessMember("Value");
 
-		void RenameTextureAsset(string from, string to) =>	RenameStaticField("Terraria.Main", from, to, newType: "Terraria.GameContent.TextureAssets").FollowBy(getValue);
-		void RenameFontAsset(string from, string to) =>		RenameStaticField("Terraria.Main", from, to, newType: "Terraria.GameContent.FontAssets").FollowBy(getValue);
-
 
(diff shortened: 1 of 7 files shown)


direwolf420:
My two cents about this

Specifically about edits to ExampleMod:
* It has been established before that ExampleMod, which is often the first programming related thing new modders see, should not use new syntax, and should stick to old/classic syntax, given that many (good) tutorials for modding and C# are older aswell so they will get confused. Things like == and != being replaced by python-style "and/not" etc., LINQ, or switch expressions. That's also why for example the "namespace block" still exists in ExampleMod while it's gone for tml files.

In general:
* Based on the [style guide](https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide), refactoring other people's code just for the sake of it is not encouraged in tml. Quote: *Please be mindful of past developers and don't reformat code just for the sake of it. Reformatting PRs will only be considered when there is an active need for change in the associated piece of code*. The `Array.Empty` change for example improves performance slightly but is so minor that it also falls into this category. 
* Some changes violate the `// Braces required on for loops` convention based on the style guide.
* What changes were made in FNA?

tygyh:
> What changes were made in FNA?

Each commit with changes to the FNA files has the changes explained in the commit message. Should I have left FNA alone?

Chicken-Bones:
Thankyou for your interest. As per the style guide, we don't accept reformatting PRs. Some of the changes you have made are good and quite informative, I'll keep them in mind when I'm writing in the future. Others are matters of personal preference (like `Where` in `foreach` vs a `continue` guard clause, or putting local methods at the bottom of a method). 

Please also note that reformatting FNA will make it harder to merge upstream changes. It's important to keep the diffs on our fork minimal.

PR #4459, opened 2024-11-30
Add checks to prevent path traversals

tygyh:
### What is the bug?
The input `path` in `zipdir` is not checked for [path traversals](https://cwe.mitre.org/data/definitions/23.html) which could allow someone to access arbitrary files.

### How did you fix the bug?
Add checks which raise errors if a malicious path is detected.

None which I know of.


JavidPack:
This file is not used since 1.3, and was only used by a developer zipping up a release on their own system, there is no risk of someone maliciously adjusting the inputs. It looks like you didn't adjust the `zipdir` method call sites either, but since we aren't even touching this file anymore there is no need to fix it.

PR #4472, opened 2024-12-09
Add new ItemID Set to simplify item swap

Silencersn:
### What is the new feature?
Add `ItemID.Sets.RightClickItemSwap` and `ItemID.Sets.UseUnlockSoundStyleAfterItemSwap` to simplify the item swap operation when right clicking in inventory.
`ItemID.Sets.RightClickItemSwap` is used for determine which item should it become after right clicking.
`ItemID.Sets.UseUnlockSoundStyleAfterItemSwap` is used for determine which sound to play after right clicking.

It feels a little difficult to create a kind of multi-state item like Shellphone.
What is also difficult is that I want to make a new kind Shellphone to travel to other positions.
But I could not insert it to vanilla swap process easily。

Add hooks to ModItem and GlobalItem.

```c#
// Insert my shellphone to vanilla swap process
// SetStaticDefaults
ItemID.Sets.RightClickItemSwap[Type] = ItemID.ShellphoneOcean;
ItemID.Sets.RightClickItemSwap[ItemID.Shellphone] = Type;
ItemID.Sets.UseUnlockSoundStyleAfterItemSwap[Type] = true;
```

### ExampleMod updates
`ExampleMod.Content.Items.Tools.ExampleShellphone`: a simple custom shellphone
`ExampleMod.Content.Items.Weapons.ExampleWhip` and `ExampleMod.Content.Items.Weapons.ExampleWhipAdvanced`: make them can be swapped for each other


Chicken-Bones:
Thankyou for your PR! Your examples and implementation are good and clean.

I'm not sure about the value of this compared to the maintenance cost. Currently, you can implement item swap behaviour using `ModItem.RightClick`, `CanRightClick` and `ConsumeItem`. The only think you can't do currently is change the right click sound, which seems like something we ought to change regardless. 

@JavidPack @direwolf420 thoughts?

direwolf420:
> Thankyou for your PR! Your examples and implementation are good and clean.
> 
> I'm not sure about the value of this compared to the maintenance cost. Currently, you can implement item swap behaviour using `ModItem.RightClick`, `CanRightClick` and `ConsumeItem`. The only think you can't do currently is change the right click sound, which seems like something we ought to change regardless.
> 
> @JavidPack @direwolf420 thoughts?

Thorium Mod currently has such an item, particularly a weapon, being able to be switched between three classes. It's implemented like this:

```cs
public abstract int ConvertToItemType { get; }
public override bool CanRightClick() => true;
public override void RightClick(Player player) => Item.SetDefaults(ConvertToItemType);
public override bool ConsumeItem(Player player) => false; /*Prevents the item stack change after right clicking*/
```

One downside of this implementation is that it loses the prefix between swaps. It may be seen as a blessing in this case since  some prefixes are locked to certain classes (i.e. mana based ones for mana reduction), and from my memory, trying to apply such a prefix to an item that doesn't use any mana will lock the game up. Currently, the prefix gets automatically cleared, which allows you to get rid of a bad prefix.

So for an implementation within tml, a way to either (safely) preserve or reset the prefix would be needed in case of weapons. What if this is done to swap from an accessory, to a weapon, to a stackable item (which you can then add stacks to), and then back?

direwolf420:
I looked into this PR and found that it would suit my needs perfectly. A small issue I found is that if you for example want to do some extra actions when right clicked you need to override `CanRightClick`, which then would run some boilerplate code within `ItemLoader.RightClick` including playing the regular sound ONTOP of the unlock sound.

Also typo in your comment in ExampleMod: Unlook instead of Unlock.

Something this PR doesn't address is the fact that you can also just regularly right-click the item as you hold it to swap the modes. The code for that is in `Player.ItemCheck_ManageRightClickFeatures` and follows the same logic as the one in `ItemSlot.TryItemSwap` except it has an extra `self.releaseUseTile = false;`

Silencersn:
> I looked into this PR and found that it would suit my needs perfectly. A small issue I found is that if you for example want to do some extra actions when right clicked you need to override `CanRightClick`, which then would run some boilerplate code within `ItemLoader.RightClick` including playing the regular sound ONTOP of the unlock sound.
> 
> Also typo in your comment in ExampleMod: Unlook instead of Unlock.
> 
> Something this PR doesn't address is the fact that you can also just regularly right-click the item as you hold it to swap the modes. The code for that is in `Player.ItemCheck_ManageRightClickFeatures` and follows the same logic as the one in `ItemSlot.TryItemSwap` except it has an extra `self.releaseUseTile = false;`

I've fixed the typo in ExampleMod and synced the logic to `Player.ItemCheck_ManageRightClickFeatures`.

> I looked into this PR and found that it would suit my needs perfectly. A small issue I found is that if you for example want to do some extra actions when right clicked you need to override `CanRightClick`, which then would run some boilerplate code within `ItemLoader.RightClick` including playing the regular sound ONTOP of the unlock sound.

For this issue, I'm not sure how to fix it properly, so I've left it unchanged.

Also, I'm not very familiar with the current 1.4.5 branch or git — I'm not sure if my commits break anything, sorry if they do.

PR #4475, opened 2024-12-13
Moved TileLoader.DrawEffects to a better position in DrawSingleTile

Lion8cake:
### What is the bug?
TileLoader.DrawEffects is being called before the check for if tile is rendered and the setting of drawdata.tintcolor and finalcolor.
This means that even if the tile isn't rendered on screen (for example if the tile is hidden with ecto coating) the draw effects are still being called for the tile.
This also in addition means that modded tiles (and global tiles) are unable to edit the drawdata.tintColor and drawdata.finalcolor of tiles. Such as editing the color at which rainbow tiles are drawn with or players being able to make their own tile with a color similar to rainbow bricks. (A mod I currently help develop has to use glowmasks to make their tile be colored differently due to finalcolor being inaccessible)

### How did you fix the bug?
Moved the position of TileLoader.DrawEffects to be after the setting of drawdata.finalcolor

Moving the setting of drawdata.tintcolor and drawdata.finalcolor to before the flag check if the tile is rendered as some mods may use DrawEffect for their tiles to spawn dusts or gores. Although this should be done in either pre or postdraw. This also has effects at messing with possible IL edits for DrawSingleTile as some edits may want to edit drawdata after the tile rendering check and use the setting of finalcolor as a hook point.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
@@ -185,14 +185,14 @@
  				if (drawData.tileLight.R < sightColor.R)
  					drawData.tileLight.R = sightColor.R;
  
-@@ -715,6 +_,7 @@
+@@ -720,6 +_,7 @@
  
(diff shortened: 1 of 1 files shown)


JavidPack:
I moved the visible flag check to after the DrawEffects call again. ExampleMod has always used DrawEffects for dust and gore spawns, and some tiles are expected to spawn those even while invisible, so I don't want to silently break that behavior in existing mods. 

#4551 will add a more appropriate hook for spawning dust and gore with the right logic already built in, but for now I expect mods to be using DrawEffects and don't want to break them.

PR #4492, opened 2024-12-30
feat: postpone hash verification until read error or mod load

steviegt6:
### What is the new feature?

`.tmod` file loading as been adjusted to only verify data blob lengths rather than full hash verification. Instead, hash verification has been postponed to mod loading to minimize unnecessary hashing. Hashes will also be taken if the initial `TmodFile::Read` invocation fails.

It significantly speeds up start-up times.

A better hashing algorithm and streaming in the background are possible but unnecessary (the former) and terrible (the latter).

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
@@ -451,8 +451,9 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 
 	internal bool VerifyHash()
 	{
+		IDisposable handle = null;
 		if (fileStream == null)
-			return false;
+			handle = Open();
 
 		if (hashStartPos == 0)
 			return false;
@@ -464,6 +465,7 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 		fileStream.Position = hashStartPos;
 		var result = Hash.SequenceEqual(sha1.ComputeHash(fileStream));
 		fileStream.Position = oldPos;
+		handle?.Dispose();
 		return result;
 	}
 
--- a/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
@@ -1,5 +1,4 @@
 using Ionic.Zlib;
-using Microsoft.Build.Tasks;
 using System;
 using System.Collections;
 using System.Collections.Generic;
--- a/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
@@ -1,7 +1,9 @@
 using Ionic.Zlib;
+using Microsoft.Build.Tasks;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
@@ -45,6 +47,8 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 
 	private static string Sanitize(string path) => path.Replace('\\', '/');
 
+	private static readonly SHA1 sha1 = SHA1.Create();
+
 	public readonly string path;
 
 	private FileStream fileStream;
@@ -65,6 +69,9 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 
 	internal byte[] Signature { get; private set; } = new byte[256];
 
+	// Starting position of the hashable part of the stream.
+	private long hashStartPos;
+
 	internal TmodFile(string path, string name = null, Version version = null)
 	{
 		this.path = path;
@@ -76,11 +83,17 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 
 	public byte[] GetBytes(FileEntry entry)
 	{
-		if (entry.cachedBytes != null && !entry.IsCompressed)
-			return entry.cachedBytes;
+		try {
+			if (entry.cachedBytes != null && !entry.IsCompressed)
+				return entry.cachedBytes;
 
-		using (var stream = GetStream(entry))
-			return stream.ReadBytes(entry.Length);
+			using (var stream = GetStream(entry))
+				return stream.ReadBytes(entry.Length);
+		}
+		catch (Exception e) {
+			VerifyHashOrThrow(e);
+			throw;
+		}
 	}
 
 	public List<string> GetFileNames() => files.Keys.ToList();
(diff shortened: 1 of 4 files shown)


Mirsario (approved):

steviegt6:
As a side note: I'm personally of the opinion we should remove hashing entirely. Data corruption of this nature is rare and most mods are now sourced from the Steam workshop, which should verify file integrity itself anyway. Many other mod loaders (such as various Minecraft ones) don't hash and face no issues.

It really seems to exist as a remnant of early tModLoader (the entire `.tmod` file format does, for that matter) with little consideration for its continued existence in the current day. For compatibility reasons, we could continue to supply a hash but just skip checking entirely.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -65,6 +68,11 @@ internal FileEntry(string name, int offset, int length, int compressedLength, by
 
 	internal byte[] Signature { get; private set; } = new byte[256];
 
+	// Starting position of the hashable part of the stream.
+	private long hashStartPos;
+	private bool hashHasBeenTaken;
+	private bool hashMatches;
```
Chicken-Bones: `bool? hashVerified`

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -45,6 +46,8 @@ internal FileEntry(string name, int offset, int length, int compressedLength, by
 
 	private static string Sanitize(string path) => path.Replace('\\', '/');
 
+	private static readonly SHA1 sha1 = SHA1.Create();
```
Chicken-Bones: Does this actually make a difference to performance? We're not multithreading right now, but it's a reduction in encapsulation and potential for future mistakes.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -436,4 +448,40 @@ private void Upgrade()
 		Open();
 		// Read contract fulfilled
 	}
+
+	internal bool VerifyHash()
+	{
+		IDisposable handle = null;
+		if (fileStream == null)
+			handle = Open();
```
Chicken-Bones: No need for the `if` check here. You need to use `using`. The guard statements should also run before trying to get a file handle.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -436,4 +448,40 @@ private void Upgrade()
...
+		if (hashStartPos == 0)
+			return false;
+
+		if (hashHasBeenTaken)
+			return hashMatches;
+
+		var oldPos = fileStream.Position;
+		fileStream.Position = hashStartPos;
+		var result = Hash.SequenceEqual(sha1.ComputeHash(fileStream));
+		fileStream.Position = oldPos;
```
Chicken-Bones: No need to put `Position` back, but you should assert that `sharedEntryReadStream == null` just like in `GetStream`. Or just open a new `FileStream`, I don't think there's any real benefit to reusing the existing one for this.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -436,4 +448,40 @@ private void Upgrade()
 		Open();
 		// Read contract fulfilled
 	}
+
+	internal bool VerifyHash()
```
Chicken-Bones: Should probably be something like `VerifyHash() => hashVerified ??= _VerifyHash()` where `_VerifyHash` has the impl and is `private`.

Chicken-Bones (changes requested):
Overall seems sensible, just some implementation feedback

Chicken-Bones:
> As a side note: I'm personally of the opinion we should remove hashing entirely. Data corruption of this nature is rare and most mods are now sourced from the Steam workshop, which should verify file integrity itself anyway. Many other mod loaders (such as various Minecraft ones) don't hash and face no issues.
> 
> It really seems to exist as a remnant of early tModLoader (the entire `.tmod` file format does, for that matter) with little consideration for its continued existence in the current day. For compatibility reasons, we could continue to supply a hash but just skip checking entirely.

Other mod loaders do hash during the data retrieval process. While I agree that data corruption is rare, and the mod would most likely fail to load if it occurred, I don't entirely trust steam our our own server-client mod syncing. In the past hashing was more important because of the signing and trust chain, but I'd like to keep it if the performance impact can be acceptably mitigated.

hakusaro (approved):

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
steviegt6: You're right: when I was glancing at the implementation of Open I misunderstood the `fileStream != null` check.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
steviegt6: It does not and I didn't explicitly intend to leave that in there. The multithreading issues are a good catch.

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -436,4 +445,31 @@ private void Upgrade()
...
+
+		using var fs = File.OpenRead(path);
+		fs.Position = hashStartPos;
+		return Hash.SequenceEqual(SHA1.Create().ComputeHash(fs));
+	}
+
+	private void VerifyHashOrThrow(Exception exception)
+	{
+		if (fileStream == null)
+			throw new IOException($"File not open: {path}", exception);
```
Chicken-Bones: No need for this anymore

patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
```diff
@@ -436,4 +445,31 @@ private void Upgrade()
...
+
+		// It shouldn't be possible to get this far with an unknown hash starting
+		// position:
+		// - erroneous user code would fail due to no file stream or similar,
+		// - saving, etc. does not use this.
+		Debug.Assert(hashStartPos != 0);
+
+		if (!VerifyHash())
+			throw new Exception(Language.GetTextValue("tModLoader.LoadErrorHashMismatchCorrupted"), exception);
```
Chicken-Bones: May as well inline this, since the method has no other uses. Then you probably don't need the assert.

Chicken-Bones (changes requested):
Looking good! Just minor cleanups

Chicken-Bones:
[2024.12.2.1](https://github.com/tModLoader/tModLoader/releases/tag/v2024.12.2.1) now on preview

PR #4499, opened 2025-01-13
Fix shimmer lag by passing `quiet: true` to `AddBuff`

drunderscore:
### What is the bug?
Clients and servers will constantly (and redundantly) sync the `Shimmer` buff. This wastes bandwidth, and will eventually cause significant lag to certain clients.

See commit message for details.

### How did you fix the bug?
Passing `quiet: true` to `AddBuff`

No.


JavidPack:
This should work for 1.4.4. Thanks.

PR #4500, opened 2025-01-13
Fix inverted netmode check for `PrimeLaser` NPC

drunderscore:
### What is the bug?
The Prime Laser NPC AI will spawn projectiles on the client, which is only meant to happen on the server.

### How did you fix the bug?
Correct the `Main.netMode` check.

No.


JavidPack:
Thanks for finding this. I've also reported this to the Terraria devs for fixing there as well.

drunderscore:
Thanks, do appreciate that.

PR #4501, opened 2025-01-13
Fix `PlayerDisconnect` hook being called with bogus player

drunderscore:
### What is the bug?
The `ModPlayer.PlayerDisconnect` hook can be invoked with a bogus/useless player. This is because the player object is re-instantiated by a call to `RemoteClient.Reset` before the player disconnection proceeds.

See commit message for details.

### How did you fix the bug?
Re-ordering the calls to `RemoteClient.Reset` and `NetMessage.SyncDisconnectedPlayer` resolves this, ensuring the disconnection hook is invoked _before_ the player is re-instantiated.

No.

I don't _believe_ this fix has any other side-effects (pertaining to the Terraria code), but there's always the possibility I've missed something.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModPlayer.cs
@@ -1102,6 +1102,8 @@ public abstract class ModPlayer : ModType<Player, ModPlayer>, IIndexed
 
 	/// <summary>
 	/// Called when a player disconnects.
+	/// <para/> Called on the server when this player disconnects.
+	/// <para/> Called on other clients when this player disconnects.
 	/// </summary>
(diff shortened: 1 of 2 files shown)


JavidPack:
K, I stepped through things and the logic seems to be preserved, but now the ModPlayer data is still valid when PlayerDisconnect  happens. (Tested with logging `exampleLifeFruits` on the server.) Note that this fix only changes server behavior, it seems when PlayerDisconnect is called on clients it has the expected ModPlayer instance already.

drunderscore:
Awesome, thank you!
> Note that this fix only changes server behavior, it seems when PlayerDisconnect is called on clients it has the expected ModPlayer instance already.

Totally correct, I definitely should have pointed this out when talking about the path it takes, or at least in the PR description :^)

Issue #4413, opened 2024-10-13
[Proposal] Shake hook for vanilla trees


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently it's not possible without much manual work to make items drop from vanilla trees via [Tree Shaking](https://terraria.wiki.gg/wiki/Trees/Shaking). The options are either:
1. Detour `WorldGen.ShakeTree` and replicate the preamble code handling the conditions (which also involves reflection as those fields are private), then add your own drop
2. IL edit around `PlantLoader.ShakeTree`

Detouring `PlantLoader.ShakeTree` is not possible as the method is too short and gets inlined.

The proposal is a "global" hook that gets incorporated into `PlantLoader.ShakeTree`, allowing mods to add shake loot to vanilla trees, and mods to add loot to other mods' trees.

### What does this proposal attempt to solve or improve?

Make it easier

### Which (other) solutions should be considered?

Problem is placing the hook in a suitable location as there is no `GlobalTree` class

**Optional:** `NPCLoader.blockLoot` support so one can remove drops (effectively "replacing" them by adding your own)

JavidPack:
I guess GlobalTile could work if we need it to. Probably won't be an issue with making a hook as described.

Whalios:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

PR #4503, opened 2025-01-14
Simple Vanilla Tree Shaking Hook

Rijam:
### What is the new feature?

Implements #4413 : Tree shaking hook for vanilla trees. New hooks, `GlobalTile.PreShakeTree(int x, int y, TreeTypes treeType)` and `GlobalTile.ShakeTree(int x, int y, TreeTypes treeType)`, are added which run in `WorldGen.ShakeTree` before all of the vanilla drop logic. `PreShakeTree` is for bonus drops and blocking items while `ShakeTree` is for drops that follow the existing pattern of a primary item drop. 

Enum `TreeTypes` is also updated to include `Custom`. `ModTree`s no longer default to Forest trees and `ModPalmTree`s no longer default to Palm trees.

As stated in the issue, making things happen when shaking a vanilla tree requires Detouring `WorldGen.ShakeTree` or IL edit `PlantLoader.ShakeTree`. The detour isn't that straight forward because it requires you to copy a lot of the checks that `ShakeTree` does and requires you to reflect a few fields which are private.

* The hook could be incorporated into `PlantLoader` instead of `GlobalTile` as suggested in the issue.
* The names of everything, of course.
* The vanilla tree shaking system doesn't support Gem trees (and Bamboo and Cactus), which some people may want to add drops to. That is beyond the scope of this PR, though.

Showcase in `ExampleMod/Common/GlobalTiles/ShakeTrees.cs`

### ExampleMod updates

`GlobalShakeTrees` class showscases some uses for the new hook. It shows spawning items, projectiles, and NPCs.

### Short Summary

* New `GlobalTile.ShakeTree` hook to add drops when shaking trees.
* New `GlobalTile.PreShakeTree` hook to add bonus drops when shaking trees or prevent specific items from dropping.

### Porting Notes

* `ModTree` and `ModPalmTree` `CountsAsTreeType` now defaults to `TreeTypes.Custom` instead of `TreeTypes.Forest` and `TreeTypes.Palm` respectively. Adjust this back if that behavior was desired.

### DIFF
--- a/ExampleMod/Common/GlobalTiles/ShakeTrees.cs
+++ b/ExampleMod/Common/GlobalTiles/ShakeTrees.cs
@@ -1,0 +1,48 @@
+using Terraria;
+using Terraria.ID;
+using Terraria.Enums;
+using Terraria.ModLoader;
+using Terraria.DataStructures;
+using ExampleMod.Content.Items;
+using ExampleMod.Content.Projectiles;
+
+namespace ExampleMod.Common.GlobalTiles
+{
+	public class GlobalShakeTrees : GlobalTile {
+		// With this hook, we can make things happen when a tree is shook.
+		// See ExampleSourceDependentItemTweaks for another example for tree shaking.
+		public override void ShakeTree(int x, int y, TreeTypes treeType) {
+
+			// Normal forest trees have a 33% chance to drop an Example Item.
+			if (treeType == TreeTypes.Forest && WorldGen.genRand.NextBool(3)) {
+				Item.NewItem(WorldGen.GetItemSource_FromTreeShake(x, y), x * 16, y * 16, 16, 16, ModContent.ItemType<ExampleItem>());
+			}
+
+			// Glowing Mushroom trees have 10% chance to drop between 3 and 10 Mushroom Torches.
+			if (treeType == TreeTypes.Mushroom && WorldGen.genRand.NextBool(10)) {
+				Item.NewItem(WorldGen.GetItemSource_FromTreeShake(x, y), x * 16, y * 16, 16, 16, ItemID.MushroomTorch, WorldGen.genRand.Next(3, 11));
+			}
+
+			// Snow (Boreal) trees have a 5% chance to spawn an Example Paper Airplane projectile that flies left or right.
+			// The owner of the projectile is set to Main.myPlayer which means the server owns the projectile in multiplayer.
+			if (treeType == TreeTypes.Snow && WorldGen.genRand.NextBool(20)) {
+				Projectile.NewProjectile(new EntitySource_ShakeTree(x, y), x * 16, y * 16, Main.rand.Next(-16, 16), 0f, ModContent.ProjectileType<ExamplePaperAirplaneProjectile>(), Damage: 4, KnockBack: 1f, Owner: Main.myPlayer);
+			}
+
+			// Jungle (Mahogany) trees have a 14% chance to spawn a Giant Flying Fox if located on the surface and in Hardmode.
+			// y == 0 is the top of the world, so y < Main.worldSurface is the area from the surface height to the top of the world.
+			if (treeType == TreeTypes.Jungle && WorldGen.genRand.NextBool(7) && y < Main.worldSurface && Main.hardMode) {
+				NPC.NewNPC(WorldGen.GetNPCSource_ShakeTree(x, y), x * 16, y * 16, NPCID.GiantFlyingFox);
+			}
+
+			// In this example, there is 20% chance for any tree to shoot a hostile arrow downwards on No Traps and Get Fixed Boi worlds.
+			if (WorldGen.genRand.NextBool(5) && Main.noTrapsWorld) {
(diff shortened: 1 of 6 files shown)


patches/tModLoader/Terraria/ModLoader/ModPlants.cs
```diff
@@ -276,7 +276,7 @@ public abstract class ModPalmTree : ITree
 	/// <summary>
 	/// Used mostly for vanilla tree shake loot tables
 	/// </summary>
-	public virtual TreeTypes CountsAsTreeType => TreeTypes.Palm;
+	public virtual TreeTypes CountsAsTreeType => TreeTypes.ModPalmTree;
```
Chicken-Bones: This will need porting notes. Minor breakage for people relying on the default behavior. Also need to update the doc string for this member to mention the new hook.

patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
```diff
@@ -316,4 +317,15 @@ public virtual bool CanReplace(int i, int j, int type, int tileTypeBeingPlaced)
...
 	{
 	}
+
+	/// <summary>
+	/// This hook runs when any tree is shook. This allows you to add drops when you shake trees.
+	/// </summary>
+	/// <param name="x">The x tile coordinate of the tree.</param>
+	/// <param name="y">The y tile coordinate of the top of the tree.</param>
+	/// <param name="treeType">The type of tree that is being shaken. Modded trees will be <see cref="TreeTypes.ModTree"/> or <see cref="TreeTypes.ModPalmTree"/></param>
+	public virtual void ShakeTree(int x, int y, TreeTypes treeType)
```
Chicken-Bones: Should we supply the `ITree` instance as a parameter for modded trees? Otherwise, an example that uses the hook to target `ExampleTree` specifically is needed.

ExampleMod/Common/GlobalTiles/ShakeTrees.cs
```diff
@@ -0,0 +1,52 @@
...
+				NPC.NewNPC(WorldGen.GetNPCSource_ShakeTree(x, y), x * 16, y * 16, NPCID.GiantFlyingFox);
+			}
+
+			// In this example, there is 20% chance for any tree to shoot a hostile arrow downwards on No Traps and Get Fixed Boi worlds.
+			if (WorldGen.genRand.NextBool(5) && Main.noTrapsWorld) {
+				Projectile.NewProjectile(new EntitySource_ShakeTree(x, y), x * 16, y * 16, Main.rand.Next(-100, 101) * 0.002f, 8f, ProjectileID.WoodenArrowHostile, Damage: 10, KnockBack: 0f, Owner: Main.myPlayer);
+			}
+
+			// Modded trees will be tree type ModTree or ModPalmTree.
+			// In this example, there is a 50% chance for ANY modded tree or modded palm tree to drop a Party Zombie at night.
```
Chicken-Bones: This isn't exactly correct, since mod trees can change their `CountsAsTreeType`. Maybe there should just be a single new `TreeTypes` member of `Custom`.

Docs will need to indicate that `TreeTypes.None` will prevent the global `ShakeTree` hook, while `TreeTypes.Custom` will permit it...

Chicken-Bones (changes requested):

ExampleMod/Common/GlobalTiles/ShakeTrees.cs
```diff
@@ -41,11 +42,21 @@ public override void ShakeTree(int x, int y, TreeTypes treeType) {
...
-			if ((treeType == TreeTypes.ModTree || treeType == TreeTypes.ModPalmTree) && WorldGen.genRand.NextBool(2) && !Main.dayTime) {
+			// Modded trees will be tree type Custom by default.
+			// In this example, there is a 50% chance for a modded tree to drop a Party Zombie at night.
+			if (treeType == TreeTypes.Custom && WorldGen.genRand.NextBool(2) && !Main.dayTime) {
 				NPC.NewNPC(WorldGen.GetNPCSource_ShakeTree(x, y), x * 16, y * 16, ModContent.NPCType<PartyZombie>());
 			}
+
+			// In this example, we want to target Example Trees specifically. We don't want other modded trees such as Example Palm Tree.
+			if (treeType == TreeTypes.Custom) {
+				WorldGen.GetTreeBottom(x, y, out int newX, out int newY); // Finds the block that the tree is planted on.
```
Chicken-Bones: `baseX, baseY`

patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
```diff
@@ -319,11 +319,12 @@ public virtual void PostSetupTileMerge()
 	}
 
 	/// <summary>
-	/// This hook runs when any tree is shook. This allows you to add drops when you shake trees.
+	/// <br>This hook runs when any tree is shook. This allows you to add drops when you shake trees.</br>
+	/// <br><see cref="TreeTypes.None"/> will not allow the tree to be shaken.</br>
```
Chicken-Bones: `shook` -> `shaken`
Maybe remove the `TreeTypes.None` line since the tense doesn't quite make sense. The documentation on the `CountsAsTreeType` will suffice

Chicken-Bones (approved):
Couple doc suggestions, but I think it's good

PR #4514, opened 2025-01-22
ModCommand IsCaseSensitive property

xNarnia:
When creating a ModCommand, the input passed into the Action method is lower-cased, making case-sensitive operations impossible. A new property, `ModCommand.IsCaseSensitive` has been added to disable this behaviour.

Command names themselves remain case insensitive.

### What is the bug?
Text that is passed into the `CommandLoader.HandleCommand` method are lower-cased internally by Terraria.

### How did you fix the bug?
Change the argument passed into `HandleCommand` from `text` to `text2` for server console commands
`text` performs a `ToLower()`. `text2` is a copy of text before it lower-cased.

Rewrote `CommandLoader.HandleCommand` to lowercase the arguments only if `!mc.IsCaseSensitive`

We could leave casing to the implementors of each `ModCommand.Action`, but this would be a breaking change and force unnecessary work on the majority of command implementors.

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -1634,7 +1634,7 @@
 +						Console.WriteLine(entry.Item1 + new string('\t', num2 - entry.Item1.Length / 8) + entry.Item2);
 +					}
  				}
-+				else if (CommandLoader.HandleCommand(text, commandCaller)) { }
++				else if (CommandLoader.HandleCommand(text2, commandCaller)) { }
  				else if (text == Language.GetTextValue("CLI.Settle_Command")) {
  					if (!Liquid.panicMode)
  						Liquid.StartPanic();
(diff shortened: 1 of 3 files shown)


xNarnia:
I wanted to add this. When using commands, the name is case-insensitive. 
So /myCommand and /MYCOMMAND both still work. This is consistent in-game and on the command line console.
The input comes back case-sensitive as expected with this change.

Chicken-Bones:
Your patch only affects server console commands. What about chat commands?

Since this is a breaking change, and I don't imagine it to be useful for the majority of users, I'd prefer to enable it with a property on the command

xNarnia:
It affects specifically ModCommands loaded from the tModLoader CommandLoader.
I tested in-game and observed that both server and client ModCommands changed the input to the Action method, making it case-sensitive.

The command itself stays case-insensitive when called. Just the input becomes case-sensitive. Thats because ModCommand "TryGet" from the Command dictionary is case-insensitive.

Should I add a new optional bool for "isCaseSensitive" that accepts text2 and ToLower()'s text2 input based on that bool parameter? I can default it to false so it keeps the original behavior. This could prevent a lot of side-effects!

xNarnia:
I'll prepare an implementation for the optional bool!

xNarnia:
New implementation tested and ready. Thanks Chicken Bones helping me with the code!

PR #4515, opened 2025-01-25
ModAchievement implementation and Example Mod examples

bonsall2004:
### What is the new feature?

Mod achievements and a slight modification to the achievement menu to allow for resetting achievements.

There is a relatively decent demand for mods to be able to come along with their own achievements to help with a sense of progression. I've noticed many people in the tModLoader discord server talking about achievements. I've read through the discussion about it and it seems like many people would like to see it implemented.

It could be implemented with an API mod or some other way but it wouldn't interface directly with the native Terraria achievements like this pull request's content does.

Modders can add achievements for killing specific bosses, entering particular biomes or getting quest items.

An achievement consists of 1 or more conditions. The majority of achievements only have a single condition. The simplest usage comes from using existing conditions, such as `AddNPCKilledCondition`, `AddItemCraftCondition`, `AddItemPickupCondition`, or `AddTileDestroyedCondition`. These conditions will automatically be satisfied and require no additional code to be added to other files. Conditions are added in `SetStaticDefaults`.

An achievement most likely also needs a position within the existing achievements indicating where in the progression it would likely fall. See the newly added [Achievement Identifiers section of the Vanilla Content IDs wiki page](https://github.com/tModLoader/tModLoader/wiki/Vanilla-Content-IDs#achievement-identifiers) to look up the internal names of vanilla achievements. The positioning is declared with the `GetDefaultPosition` hook.

An achievement also needs a texture file with the icons in it. The texture should be 130x64 pixels and includes 2 64x64 icons. The icon on the left is the achievement icon when it has been achieved, and the icon on the right is when it's locked. The black borders indicate where the achievement border will be drawn over it, so the visible area is more like 60x60 for each icon. https://github.com/tModLoader/tModLoader/blob/stable/ExampleMod/Content/Achievements/Template.png can be used as a template.

Here is a simple example:
```cs
public class MinionBossKilled : ModAchievement
{
		Achievement.SetCategory(AchievementCategory.Slayer);

		AddNPCKilledCondition(ModContent.NPCType<MinionBossBody>());
	}

	public override Position GetDefaultPosition() => new After("EYE_ON_YOU");
}
```
Corresponding MinionBossKilled.png:
    

Please see the examples in https://github.com/tModLoader/tModLoader/blob/stable/ExampleMod/Content/Achievements/ for more information. Many more concepts are explained in the actual examples, this is just a very simple example.

### ExampleMod updates

**MinionBossKilled**: Showcases the most straightforward use-case, defeating a boss. Also showcases ordering a `ModAchivement` after a specific vanilla achievement.

**ManyExampleWormsKilled**: A more advanced example showcasing a custom achievement condition and conditional hiding of the achievement details in the achievements menu. This example will be shown as a progress bar automatically.

**AdvancedExampleAchievement**: Advanced example showcasing all vanilla achievement conditions, sharing a achievement icon texture, a custom `IAchievementTracker`, ordering a `ModAchivement` after a specific modded achievement, and when to use a custom condition key. It also talks about multiplayer considerations of custom achievement conditions. 

### Showcase

This video shows the new changes to the achievements menu and the ExampleMod achievements. It shows how modded achievements can be ordered, satisfied/unlocked, and hidden until some conditions. It also shows how the achievements menu now has a search box, a button to reset achievements, and a new toggle to show only modded achievements.

https://github.com/user-attachments/assets/5f74678e-4dcc-486c-a9cd-ba1f7e087988

### Porting Notes

If you have manually manipulated the achievements previously, please use this new system.

### Additional Notes

Achievements are per-user, not per player or per-world. Don't use them to reward a player with critical items necessary for progression since they might have already unlocked the achievement in a previous playthrough.


patches/tModLoader/Terraria/ModLoader/ModContent.cs
```diff
@@ -166,6 +166,9 @@ public static bool RequestIfExists<T>(string name, out Asset<T> asset, AssetRequ
 	/// <inheritdoc cref="ItemLoader.GetItem"/>
 	public static ModItem GetModItem(int type) => ItemLoader.GetItem(type);
 
+	/// <inheritdoc cref="AchievementLoader.GetAchievement"/>
+	public static ModAchievement GetModAchievement<T>() where T : ModAchievement => AchievementLoader.GetAchievement(GetInstance<T>().Type);
```
Chicken-Bones: No need for this. `ModContent.GetInstance<T>()` should be returning the correct instance. You'll note we only have specific `GetModX` methods when the lookup is by `int type` or `string texture`

ExampleMod/Content/Achievements/Conditions/ExampleAchievementCondition.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Terraria.Achievements;
+
+namespace ExampleMod.Content.Achievements.Conditions;
+
+public class ExampleAchievementCondition : AchievementCondition
```
Chicken-Bones: What's the point of making a custom `AchievementCondition`? I don't see this used in any of the example achievements. 

Can we automate the string naming by using a default constructor that sets the name to something like
```cs
static string GetNameFromModType(Type type) {
    if (!AssemblyManager.GetAssemblyOwner(GetType().Assembly, out modName))
        throw new ArgumentException($"{type} is not from a mod");
    
    return $"{modName}/{type.Name}";
}
```

ExampleMod/Content/Achievements/ExampleWormKilled.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace ExampleMod.Content.Achievements;
+
+public class ExampleWormKilled : ModAchievement
+{
+	public override void SetStaticDefaults() {
+		Achievement.Hidden = false;
+
+		// The achievement can be completed like this:
+		// ModContent.GetModAchievement<ExampleWormKilled>().GetCondition("EXAMPLEMOD_KILL_BOOLEAN_CONDITION").Complete();
+		Achievement.AddCondition(CustomFlagCondition.Create("EXAMPLEMOD_KILL_BOOLEAN_CONDITION"));
```
Chicken-Bones: Conditions should probably use the `FullName` of the `ModAchievement` as a prefix, or if there's only one condition, the `FullName` itself

Chicken-Bones: It's also not advisable to have string constants floating throughout the codebase. Better would be to make a field.

```cs
public readonly AchievementCondition Condition = CustomFlagCondition.Create(FullName);
```

Then other code can complete it via `ModContent.GetInstance<ExampleWormKilled>().Condition.Complete();`

ExampleMod/Content/Achievements/ExampleWormKilled.cs
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	public override void OnCompleted(Achievement achievement) {
+		Main.LocalPlayer.QuickSpawnItem(Main.LocalPlayer.GetSource_FromThis(), ModContent.ItemType<ExampleBar>(), 5);
+		base.OnCompleted(achievement);
```
Chicken-Bones: We don't include `base` calls in our hook examples when the base hook methods are empty.

ExampleMod/Content/Achievements/ManyExampleWormsKilled.cs
```diff
@@ -0,0 +1,24 @@
...
+
+public class ManyExampleWormsKilled : ModAchievement
+{
+	public override void SetStaticDefaults() {
+		Achievement.Hidden = true;
+
+		// Int conditions will automatically complete once you've incremented it enough.
+		// ((CustomIntCondition)ModContent.GetModAchievement<ManyExampleWormsKilled>().GetCondition("EXAMPLEMOD_KILL_WORMS")).Value++;
+		Achievement.AddCondition(CustomIntCondition.Create("EXAMPLEMOD_KILL_WORMS", 5));
+		Achievement.UseTrackerFromCondition("EXAMPLEMOD_KILL_WORMS");
```
Chicken-Bones: As with the advice above, recommend making the `Condition` a field, and then using `UseTracker(Condition.GetAchievementTracker())`

ExampleMod/Content/NPCs/ExampleWorm.cs
```diff
@@ -49,6 +52,13 @@ public override void SetBestiary(BestiaryDatabase database, BestiaryEntry bestia
 			});
 		}
 
+		public override void OnKill() {
+			ModContent.GetModAchievement<ExampleWormKilled>().GetCondition("EXAMPLEMOD_KILL_BOOLEAN_CONDITION").Complete();
+			((CustomIntCondition)ModContent.GetModAchievement<ManyExampleWormsKilled>().GetCondition("EXAMPLEMOD_KILL_WORMS")).Value++;
```
Chicken-Bones: As with the advice above, the cast would not be needed if the `CustomIntCondition Condition` field could be accessed directly.

ExampleMod/Content/NPCs/ExampleWorm.cs
```diff
@@ -49,6 +52,13 @@ public override void SetBestiary(BestiaryDatabase database, BestiaryEntry bestia
...
+
+			base.OnKill();
```
Chicken-Bones: As above, no need to call `base`

ExampleMod/ExampleMod.csproj
```diff
@@ -5,8 +5,6 @@
     <AssemblyName>ExampleMod</AssemblyName>
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
-  <ItemGroup>
-  </ItemGroup>
```
Chicken-Bones, line 9: A helpful cleanup, but please refrain from changes outside the scope of the PR in the future.

ExampleMod/Localization/ru-RU.hjson
```diff
@@ -9,6 +9,17 @@
...
+		    ExampleWormKilled: {
+		        Name: Example Worm Killed (Russian)
+		        Description: Congrats... you murdered a worm. I bet you feel so proud of yourself - Example Achievement
+		    }
+		    
+            ManyExampleWormsKilled: {
+                Name: MORE WORMS?? (Russian)
+                Description: Okay... we get it, now please stop murdering these poor worms. - Example Achievement
+            }
+		}
```
Chicken-Bones: Please omit this, localizers will handle it

patches/tModLoader/Terraria/Achievements/Achievement.cs.patch
```diff
@@ -0,0 +1,53 @@
...
++	public LocalizedText FriendlyName;
+-	public readonly LocalizedText Description;
++	public LocalizedText Description;
+ 	public readonly int Id = _totalAchievements++;
+ 	private AchievementCategory _category;
+ 	private IAchievementTracker _tracker;
+ 	[JsonProperty("Conditions")]
+ 	private Dictionary<string, AchievementCondition> _conditions = new Dictionary<string, AchievementCondition>();
+ 	private int _completedCount;
++	public int Type;           // TML: Added for keeping track of modded achievements during loading, Native achievements don't get assigned a type as they don't need one.
```
Chicken-Bones: Do modded achievements actually need a `Type`? Is there ever any reason to use `AchievementLoader.GetAchievement(int type)`?

Recommend removing

patches/tModLoader/Terraria/Achievements/Achievement.cs.patch
```diff
@@ -0,0 +1,53 @@
...
++	public bool Modded = false; // TML: Added to keep vanilla achievements working.
```
Chicken-Bones: Could all uses of this be replaced with `ModAchievement != null`

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// </summary>
+    public bool AchievementHidden = false;
+
+    protected override sealed Achievement CreateTemplateEntity()
+    {
+        if (string.IsNullOrWhiteSpace(Name))
+        {
+            throw new InvalidOperationException("Achievement name cannot be null or empty during template creation.");
+        }
+        return new Achievement($"{Mod.Name.ToUpper()}_{Name.ToUpper()}", true);
```
Chicken-Bones: Could we add a constuctor to `Achievement` that takes a `ModAchievement` instead? That way we could keep many of the fields in `Achievement` `readonly`

patches/tModLoader/Terraria/ModLoader/AchievementLoader.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections.Generic;
+
+namespace Terraria.ModLoader;
+
+public static class AchievementLoader
```
Chicken-Bones: Do we need `AchievementLoader` at all?

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+	    }
+
+	    if (Description == null)
+	    {
+		    throw new InvalidOperationException($"Description for achievement '{Name}' could not be found.");
+	    }
+
+	    Achievement.FriendlyName = FriendlyName;
+	    Achievement.Description = Description;
+	    Achievement.ModAchievement = this;
```
Chicken-Bones: These can all be set in `CreateTemplateEntity`, perhaps as part of the proposed `Achievement(ModAchievement)` constructor

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+    public virtual LocalizedText Description => Mod.GetLocalization($"{LocalizationCategory}.{Name}.Description");
+
+    protected override sealed void Register()
+    {
+	    ModTypeLookup<ModAchievement>.Register(this);
+
+	    if (string.IsNullOrWhiteSpace(Name))
+	    {
+		    throw new InvalidOperationException("Achievement name cannot be null or empty.");
+	    }
```
Chicken-Bones: Code style, no need for braces, keep all the guard statements compact

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+
+	    if (FriendlyName == null)
+	    {
+		    throw new InvalidOperationException($"FriendlyName for achievement '{Name}' could not be found.");
```
Chicken-Bones: Recommend `ArgumentNullException(nameof(FriendlyName))`, the type will be part of the stack trace

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    public override void Load()
+    {
+    }
+
+    public override void Unload()
+    {
+        Main.Achievements.Unregister(Achievement);
+        Achievement.OnCompleted -= OnCompleted;
+        Texture = null;
```
Chicken-Bones: No need to null out fields like this on unload, garbage collection will take care of it

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    public Achievement Achievement => Entity;
+
+    /// <summary>
+    /// Achievement type, this is used for keeping track of modded achievements with ModContent.GetModAchievement(int type);
+    /// </summary>
+    public int Type => Achievement.Type;
+
+    private string TextureName => (GetType().Namespace + "." + Name).Replace('.', '/');//GetType().FullName.Replace('.', '/');
+    private Asset<Texture2D> Texture;
```
Chicken-Bones: Make into a public property with `{ get; private set; }`

patches/tModLoader/Terraria/Achievements/AchievementManager.cs.patch
```diff
@@ -10,3 +18,67 @@
...
++		return new Dictionary<string, object>();
++	}
+ 	public void Load()
+ 	{
+ 		Load(_savePath, _isCloudSave);
+@@ -94,6 +_,7 @@
+ 				using CryptoStream stream2 = new CryptoStream(stream, new RijndaelManaged().CreateDecryptor(_cryptoKey, _cryptoKey), CryptoStreamMode.Read);
+ 				using BsonReader reader = new BsonReader(stream2);
+ 				dictionary = JsonSerializer.Create(_serializerSettings).Deserialize<Dictionary<string, StoredAchievement>>(reader);
++				Logging.tML.Info(dictionary);
```
Chicken-Bones: Debug logging left in

patches/tModLoader/Terraria/GameContent/Achievements/CustomIntCondition.cs.patch
```diff
@@ -0,0 +1,10 @@
+--- src/TerrariaNetCore/Terraria/GameContent/Achievements/CustomIntCondition.cs
++++ src/tModLoader/Terraria/GameContent/Achievements/CustomIntCondition.cs
+@@ -37,7 +_,6 @@
+ 		_value = 0;
+ 		base.Clear();
+ 	}
+-
```
Chicken-Bones: Accidental patch, please revert

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
++		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, ModName, ModNameLocation, value, 0f, Vector2.Zero, baseScale, num3);
+ 		vector3.X -= 17f;
+ 		Vector2 position = vector2 + Vector2.UnitY * 27f + vector;
+ 		DrawPanelBottom(spriteBatch, position, num3, color);
+@@ -126,14 +_,15 @@
+ 			if (!base.IsMouseHovering)
+ 				color3 = Color.Lerp(color3, Color.Black, 0.25f);
+ 
+-			DrawProgressBar(spriteBatch, progress, vector4 - Vector2.UnitX * num5 * 0.7f, num5, color3, color2, color2.MultiplyRGBA(new Color(new Vector4(1f, 1f, 1f, 0.5f))));
++			DrawProgressBar(spriteBatch, progress, vector4 - (Vector2.UnitX * num5 * 0.7f), num5, color3, color2, color2.MultiplyRGBA(new Color(new Vector4(1f, 1f, 1f, 0.5f))));
```
Chicken-Bones: Accidental reformat? (added parens)

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
+-			vector4.X -= num5 * 1.4f + stringSize2.X;
++			vector4.X -= num5 * 1.4f + stringSize2.X + _offsetProgressBarX - 4f;
+ 			ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, text2, vector4, value, 0f, new Vector2(0f, 0f), baseScale3, 90f);
+ 		}
+ 	}
+ 
+ 	private void UpdateIconFrame()
+ 	{
++		if (_achievement.Modded) return;
```
Chicken-Bones: Doesn't this mean modded achievements which get unlocked while the UI element is on screen won't update?

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
++
++		if (achievement.Modded) {
++			_achievementIcon =
++				new UIImageFramed(modAchievement.GetTexture(), new Rectangle((_achievement.IsCompleted ? 0 : 65), 0, 64, 64));
++		}
++		else
++		{
+-		_achievementIcon = new UIImageFramed(Main.Assets.Request<Texture2D>("Images/UI/Achievements"), _iconFrame);
++			_achievementIcon = new UIImageFramed(Main.Assets.Request<Texture2D>("Images/UI/Achievements"), _iconFrame);
++		}
```
Chicken-Bones: `_achievementIcon = new UIImageFramed(modAchievement?.Texture ?? Main.Assets.Request<Texture2D>("Images/UI/Achievements"), _iconFrame);`

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
+ 		float num3 = innerDimensions.Width - dimensions.Width + 1f - (float)(num * 2);
+ 		Vector2 baseScale = new Vector2(0.85f);
+ 		Vector2 baseScale2 = new Vector2(0.92f);
++
++		string descriptionText = _achievement.Description.Value;
++		string nameText = _achievement.FriendlyName.Value;
++		if (_achievement.Hidden && _locked) {
++			nameText = "???";
++			descriptionText = "???";
++		}
```
Chicken-Bones: Here's a fun trick, you can use shadowing and value tuples to reduce patch size and make it more maintainable.
```cs
		var _achievement = (this._achievement.Description, this._achievement.FriendlyName, this._achievement.Hidden, this._achievement.Category);

		if (_achievement.Hidden && _locked) {
			_achievement.Description = _achievement.FriendlyName = new LocalizedText(null, "???");
		}
```

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
+ 
+ 		float num4 = 38f + (float)(_large ? 20 : 0);
+ 		if (stringSize.Y > num4)
+@@ -104,7 +_,11 @@
+ 		spriteBatch.Draw(_categoryTexture.Value, vector3, _categoryTexture.Frame(4, 2, (int)category), base.IsMouseHovering ? Color.White : Color.Silver, 0f, Vector2.Zero, 0.5f, SpriteEffects.None, 0f);
+ 		vector3.X += 4f;
+ 		vector3.X += 17f;
+-		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, _achievement.FriendlyName.Value, vector3, value, 0f, Vector2.Zero, baseScale, num3);
++		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, nameText, vector3, value, 0f, Vector2.Zero, baseScale, num3);
++		string ModName = _achievement.Modded ? _achievement.ModAchievement.Mod.DisplayNameClean : "Terraria";
```
Chicken-Bones: Add a blank line before inserting new tML code, for some visual separation

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
+ using Terraria.Localization;
++using Terraria.ModLoader;
+ using Terraria.UI;
+ using Terraria.UI.Chat;
+ 
+@@ -19,6 +_,7 @@
+ 	private const int _iconsPerRow = 8;
+ 	private int _iconIndex;
+ 	private Rectangle _iconFrame;
++	private float _offsetProgressBarX = 0;
```
Chicken-Bones: No need for `= 0`

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,123 @@
...
+ 	{
++		if (_achievement.Modded) return;
+ 		if (!_locked)
+ 			_iconFrame = _iconFrameUnlocked;
+ 		else
+@@ -202,6 +_,7 @@
+ 
+ 		if (BackColor == Color.Transparent)
+ 			FillingColor = new Color(255, 255, 255);
++		spot.X -= _offsetProgressBarX - 16;
```
Chicken-Bones: Recommend dropping in a new line and doing `// Added by TML` for little surprise things like this where it's not instantly obvious. Also code convention is a blank line after reducing indents

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.cs.patch
```diff
@@ -0,0 +1,224 @@
...
++		uIPanel.Append(uISearchBar);
++
+ 		UIScrollbar uIScrollbar = new UIScrollbar();
+ 		uIScrollbar.SetView(100f, 1000f);
+ 		uIScrollbar.Height.Set(-50f, 1f);
+@@ -94,7 +_,6 @@
+ 			_categoryButtons.Add(uIToggleImage);
+ 			uIElement2.Append(uIToggleImage);
+ 		}
+-
```
Chicken-Bones: Accidental blank line patch

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.cs.patch
```diff
@@ -0,0 +1,224 @@
...
+ 		uITextPanel.BackgroundColor = new Color(73, 94, 171);
+ 		uIElement.Append(uITextPanel);
++
+ 		UITextPanel<LocalizedText> uITextPanel2 = new UITextPanel<LocalizedText>(Language.GetText("UI.Back"), 0.7f, large: true);
+-		uITextPanel2.Width.Set(-10f, 0.5f);
++		uITextPanel2.Width.Set(-10f, 0.8f);
+ 		uITextPanel2.Height.Set(50f, 0f);
+ 		uITextPanel2.VAlign = 1f;
+-		uITextPanel2.HAlign = 0.5f;
++		uITextPanel2.HAlign = 0f;
```
Chicken-Bones: Perhaps instead of editing the values here, you can override them down where you make the `resetAchievementsButton`

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.cs.patch
```diff
@@ -0,0 +1,224 @@
...
++
++	Asset<Texture2D> blockingBackground;
++	Asset<Texture2D> blockingBorder;
++
++	private UIPanel achievementResetAreYouSure;
+ 
+ 	public void InitializePage()
+ 	{
++		blockingBackground = ModLoader.ModLoader.ManifestAssets.Request<Texture2D>($"Terraria.ModLoader.UI.InnerPanelNotRounded");
++		blockingBorder = ModLoader.ModLoader.ManifestAssets.Request<Texture2D>($"Terraria.ModLoader.UI.PanelBorderNotRounded");
```
Chicken-Bones: blank line between new tML code and vanilla code please

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.cs.patch
```diff
@@ -0,0 +1,224 @@
...
+ 		}
+-
+ 		uIPanel.Append(uIElement2);
+ 	}
+ 
+@@ -152,6 +_,102 @@
+ 		IngameFancyUI.Close();
+ 	}
+ 
++	private void ResetAchievements(UIMouseEvent evt, UIElement listeningElement)
```
Chicken-Bones: Recommend moving all new methods to a `UIAchievementsMenu.TML.cs` partial

patches/tModLoader/Terraria/Initializers/AchievementInitializer.cs.patch
```diff
@@ -0,0 +1,10 @@
+--- src/TerrariaNetCore/Terraria/Initializers/AchievementInitializer.cs
++++ src/tModLoader/Terraria/Initializers/AchievementInitializer.cs
+@@ -606,7 +_,6 @@
+ 			Main.Achievements.RegisterAchievementCategory("GLORIOUS_GOLDEN_POLE", category);
+ 			Main.Achievements.RegisterAchievementCategory("TEMPLE_RAIDER", category);
+ 			Main.Achievements.RegisterAchievementCategory("LIKE_A_BOSS", category);
+-			Main.Achievements.Load();
```
Chicken-Bones: Comment out instead of removing. ideally using block comments to avoid changing indent

patches/tModLoader/Terraria/ModLoader/AccessorySlotLoader.cs
```diff
@@ -10,6 +10,7 @@
 using Terraria.Localization;
 using Terraria.ModLoader.Default;
 using Terraria.GameContent.UI.Elements;
+using Terraria.Graphics.Renderers;
```
Chicken-Bones: Accidental patch

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,116 @@
...
+    public string LocalizationCategory => "Achievements";
+
+    public override sealed bool IsCloneable => false;
+
+    /// <summary>
+    /// Gets the localized friendly name of the achievement.
+    /// </summary>
+    public virtual LocalizedText FriendlyName => Mod.GetLocalization($"{LocalizationCategory}.{Name}.Name");
+
+    public AchievementCondition GetCondition(string condition) => Achievement.GetCondition(condition);
```
Chicken-Bones: If we follow the convention of having conditions be public fields on the the `ModAchivement` subclasses, we don't need this method.

patches/tModLoader/Terraria/ModLoader/UI/UILoadMods.cs
```diff
@@ -23,6 +24,7 @@ public override void OnActivate()
 		};
 		gotoMenu = 888; // ModLoader will redirect to the mods menu if there are no errors during cancel
 		ModLoader.BeginLoad(_cts.Token);
+
```
Chicken-Bones: Accidental patches

patches/tModLoader/Terraria/ModLoader/UI/UICommon.cs
```diff
@@ -174,8 +174,11 @@ Asset<Texture2D> LoadEmbeddedTexture(string name)
 		DividerTexture = Main.Assets.Request<Texture2D>("Images/UI/Divider");
 		InnerPanelTexture = Main.Assets.Request<Texture2D>("Images/UI/InnerPanelBackground");
 
+
 		InfoDisplayPageArrowTexture = LoadEmbeddedTexture("UI.InfoDisplayPageArrow");
 		tModLoaderTitleLinkButtonsTexture = LoadEmbeddedTexture("UI.tModLoaderTitleLinkButtons");
 		CopyCodeButtonTexture = LoadEmbeddedTexture("UI.CopyCodeButton");
+
+
```
Chicken-Bones: Accidental patches

patches/tModLoader/Terraria/UI/AchievementAdvisor.cs.patch
```diff
@@ -0,0 +1,34 @@
...
+ 
+@@ -146,7 +_,18 @@
+ 
+ 		Rectangle frame = card.frame;
+ 		frame.X += 528;
++		if (card.achievement.Modded) {
++			if (card.achievement.ModAchievement.GetTexture() == null) {
++				throw new Exception($"{card.achievement.Name}.png was not found, add it in the same directory as your source file.");
++			}
++		}
```
Chicken-Bones: This should be checked during `SetupContent` after which we can assume that `Texture != null`

patches/tModLoader/Terraria/UI/AchievementAdvisor.cs.patch
```diff
@@ -0,0 +1,34 @@
...
++
++		if (!card.achievement.Modded) {
```
Chicken-Bones: Use similar trick to reduce patch size, shadow `_achivementsTexture` locally

Chicken-Bones (changes requested):

ExampleMod/Content/Achievements/ExampleWormKilled.cs
bonsall2004: After a little thought, there is no need for prefixing with the mod name as the conditions are instantiated per achievement and could not interfere with other mods. As for the string names for the conditions themselves, to keep the functionality of allowing multiple conditions per achievement it's difficult to integrate another system on top of Terraria's native achievement condition system. This could be achieved by limiting conditions per achievement but to allow for complex achievements with multiple steps, I can't think of a better way of allowing it without breaking things.

patches/tModLoader/Terraria/ModLoader/AchievementLoader.cs
bonsall2004: it keeps track of all the modded achievements for each mod allowing GetAchievement to work, I can't think of another way of doing that in such an easy way for modders to access their achievements.

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
bonsall2004: Yes but it was set elsewhere as a test that i'd forgotten about, I've moved it into the correct place now.

patches/tModLoader/Terraria/ModLoader/AchievementLoader.cs
Chicken-Bones: Modders will access their achievements via the class instance. For eg `ModContent.GetInstance<ExampleWormKilled>()`

ExampleMod/Content/Achievements/ExampleWormKilled.cs
```diff
@@ -0,0 +1,31 @@
...
+		Achievement.Hidden = false;
+
+		// There are 4 AchievementCategorys: Slayer, Collector, Explorer, and Challenger.
+		// Slayer is the default.
+		// If you want to change the achievement's category, you can do this:
+		// Achievement.SetCategory(AchievementCategory.Collector);
+
+		// The achievement can be completed like this:
+		// ModContent.GetInstance<ExampleWormKilled>().Condition.Complete();
+		Achievement.AddCondition(Condition);
```
Chicken-Bones: Probably the best approach is to do `Condition = AddCondition()` here, and remove `readonly` from the field. (It can be `{ get; private set; }` instead if we care

ExampleMod/Content/Achievements/ManyExampleWormsKilled.cs
```diff
@@ -0,0 +1,29 @@
...
+
+		// There are 4 AchievementCategorys: Slayer, Collector, Explorer, and Challenger.
+		// Slayer is the default.
+		// If you want to change the achievement's category, you can do this:
+		// Achievement.SetCategory(AchievementCategory.Collector);
+
+		// Int conditions will automatically complete once you've incremented it enough.
+		// ModContent.GetInstance<ManyExampleWormsKilled>().IntCondition.Value++;
+		Achievement.AddCondition(IntCondition);
+		Achievement.UseTracker(IntCondition.GetAchievementTracker());
```
Chicken-Bones: I feel like `UseTracker` should be automatically set to the last trackable condition.

ExampleMod/Content/Achievements/ManyExampleWormsKilled.cs
```diff
@@ -0,0 +1,29 @@
...
+using Terraria;
+using Terraria.Achievements;
+using Terraria.GameContent.Achievements;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Achievements;
+
+public class ManyExampleWormsKilled : ModAchievement
+{
+	public readonly CustomIntCondition IntCondition = new("Condition", 5);
```
Chicken-Bones: Just call this field `Condition`

patches/tModLoader/Terraria/Achievements/Achievement.cs.patch
```diff
@@ -0,0 +1,71 @@
...
+ 				value.Load(condition.Value);
+@@ -66,7 +_,11 @@
+ 
+ 	public void AddCondition(AchievementCondition condition)
+ 	{
++		if (_conditions.ContainsKey(condition.Name))
++			throw new Exception($"Cannot add two conditions with the same name: {condition.Name}");
++
+ 		_conditions[condition.Name] = condition;
++
```
Chicken-Bones: accidental newline added

patches/tModLoader/Terraria/Achievements/AchievementManager.cs.patch
```diff
@@ -10,3 +18,59 @@
...
++
++				foreach (var kvp in _achievements) {
++					existingData[kvp.Key] = kvp.Value;
++				}
++
+ 				using MemoryStream memoryStream = new MemoryStream();
+ 				using CryptoStream cryptoStream = new CryptoStream(memoryStream, new RijndaelManaged().CreateEncryptor(_cryptoKey, _cryptoKey), CryptoStreamMode.Write);
+ 				using BsonWriter bsonWriter = new BsonWriter(cryptoStream);
+-				JsonSerializer.Create(_serializerSettings).Serialize(bsonWriter, _achievements);
++				JsonSerializer.Create(_serializerSettings).Serialize(bsonWriter, existingData);
```
Chicken-Bones: Comment out vanilla code instead of removing, as per https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#comment-out-vanilla-code-rather-than-removing-it

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
++			if (_achievement.ModAchievement != null) {
++				_iconFrame = new Rectangle(65, 0, 64, 64);
++			}
++		}
+-
++			
+-		if (_achievementIcon != null)
++		if (_achievementIcon != null) 
+-			_achievementIcon.SetFrame(_iconFrame);
++			_achievementIcon.SetFrame(_iconFrame);
```
Chicken-Bones: No need for this complicated interleaved patch. Can just have the `if (_achievement.ModAchievement != null)` code run first and `return;`

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
++		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, ModName, ModNameLocation, value, 0f, Vector2.Zero, baseScale, num3);
+ 		vector3.X -= 17f;
+ 		Vector2 position = vector2 + Vector2.UnitY * 27f + vector;
+ 		DrawPanelBottom(spriteBatch, position, num3, color);
+@@ -127,20 +_,28 @@
+ 				color3 = Color.Lerp(color3, Color.Black, 0.25f);
+ 
+ 			DrawProgressBar(spriteBatch, progress, vector4 - Vector2.UnitX * num5 * 0.7f, num5, color3, color2, color2.MultiplyRGBA(new Color(new Vector4(1f, 1f, 1f, 0.5f))));
+-			vector4.X -= num5 * 1.4f + stringSize2.X;
++			vector4.X -= num5 * 1.4f + stringSize2.X + _offsetProgressBarX - 4f;
```
Chicken-Bones: Comment original code instead of removing. Also why the `-4`?

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
++		}
++
+ 		string text = FontAssets.ItemStack.Value.CreateWrappedText(_achievement.Description.Value, (num3 - 20f) * (1f / baseScale2.X), Language.ActiveCulture.CultureInfo);
+ 		Vector2 stringSize = ChatManager.GetStringSize(FontAssets.ItemStack.Value, text, baseScale2, num3);
+ 		if (!_large)
+@@ -104,7 +_,12 @@
+ 		spriteBatch.Draw(_categoryTexture.Value, vector3, _categoryTexture.Frame(4, 2, (int)category), base.IsMouseHovering ? Color.White : Color.Silver, 0f, Vector2.Zero, 0.5f, SpriteEffects.None, 0f);
+ 		vector3.X += 4f;
+ 		vector3.X += 17f;
++
```
Chicken-Bones: Accidental newline?

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
+ 		_iconFrameLocked.X += 528;
+ 		_iconFrame = _iconFrameLocked;
+ 		UpdateIconFrame();
++
+-		_achievementIcon = new UIImageFramed(Main.Assets.Request<Texture2D>("Images/UI/Achievements"), _iconFrame);
++		_achievementIcon = new UIImageFramed(modAchievement?.Texture ?? Main.Assets.Request<Texture2D>("Images/UI/Achievements"), _iconFrame);
++
+ 		_achievementIcon.Left.Set(num2, 0f);
+ 		_achievementIcon.Top.Set(num3, 0f);
++
```
Chicken-Bones: Accidental newline

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
++		_locked = !this._achievement.IsCompleted;
+ 		UpdateIconFrame();
+ 		CalculatedStyle innerDimensions = GetInnerDimensions();
+ 		CalculatedStyle dimensions = _achievementIconBorders.GetDimensions();
+@@ -82,6 +_,13 @@
+ 		float num3 = innerDimensions.Width - dimensions.Width + 1f - (float)(num * 2);
+ 		Vector2 baseScale = new Vector2(0.85f);
+ 		Vector2 baseScale2 = new Vector2(0.92f);
++
++		var _achievement = (this._achievement.Description, this._achievement.FriendlyName, this._achievement.Hidden, this._achievement.Category);
```
Chicken-Bones: Could this be moved higher to remove the `_locked` patch?

patches/tModLoader/Terraria/GameContent/UI/Elements/UIAchievementListItem.cs.patch
```diff
@@ -0,0 +1,135 @@
...
+@@ -104,7 +_,12 @@
+ 		spriteBatch.Draw(_categoryTexture.Value, vector3, _categoryTexture.Frame(4, 2, (int)category), base.IsMouseHovering ? Color.White : Color.Silver, 0f, Vector2.Zero, 0.5f, SpriteEffects.None, 0f);
+ 		vector3.X += 4f;
+ 		vector3.X += 17f;
++
+ 		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, _achievement.FriendlyName.Value, vector3, value, 0f, Vector2.Zero, baseScale, num3);
++		string ModName = this._achievement.ModAchievement != null ? this._achievement.ModAchievement.Mod.DisplayNameClean : "Terraria";
++		_offsetProgressBarX = ((ModName.Length + 1) * 10);
++		Vector2 ModNameLocation = new Vector2(vector3.X + num3 - _offsetProgressBarX, vector3.Y);
++		ChatManager.DrawColorCodedStringWithShadow(spriteBatch, FontAssets.ItemStack.Value, ModName, ModNameLocation, value, 0f, Vector2.Zero, baseScale, num3);
```
Chicken-Bones: Toss some whitespace around code added by tML

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.TML.cs
```diff
@@ -0,0 +1,122 @@
...
+	private void FilterSearch(object sender, EventArgs e)
+	{
+		_achievementsList.Clear();
+		string searchText = uISearchBar.Text?.ToLowerInvariant() ?? string.Empty; // Get the search text, ensuring it's lowercase
+
+		foreach (UIAchievementListItem achievementElement in _achievementElements) {
+			string friendlyName = achievementElement.GetAchievement().FriendlyName.Value.ToLowerInvariant(); // Convert to lowercase for case-insensitive comparison
+			string modName = achievementElement.GetAchievement().ModAchievement != null ? achievementElement.GetAchievement().ModAchievement.Mod.DisplayName.ToLowerInvariant() : string.Empty; // Convert to lowercase for case-insensitive comparison
+
+			if (friendlyName.Contains(searchText) || modName.Contains(searchText)) {
```
Chicken-Bones: You can use `StringComparison.CurrentCultureIgnoreCase` (current culture is probably better than invariant, right @JavidPack) 

https://learn.microsoft.com/en-us/dotnet/api/system.string.contains?view=net-9.0#system-string-contains(system-char-system-stringcomparison)

patches/tModLoader/Terraria/GameContent/UI/States/UIAchievementsMenu.cs.patch
```diff
@@ -0,0 +1,114 @@
...
+ 		uITextPanel.Top.Set(-33f, 0f);
+ 		uITextPanel.SetPadding(13f);
+ 		uITextPanel.BackgroundColor = new Color(73, 94, 171);
+ 		uIElement.Append(uITextPanel);
++
+ 		UITextPanel<LocalizedText> uITextPanel2 = new UITextPanel<LocalizedText>(Language.GetText("UI.Back"), 0.7f, large: true);
+-		uITextPanel2.Width.Set(-10f, 0.5f);
+ 		uITextPanel2.Height.Set(50f, 0f);
+ 		uITextPanel2.VAlign = 1f;
+-		uITextPanel2.HAlign = 0.5f;
```
Chicken-Bones: Comment out (block comment) instead of removing please

patches/tModLoader/Terraria/Initializers/AchievementInitializer.cs.patch
Chicken-Bones: As per style guide given above, the block comments start and end should be on separate lines. The indentation of the vanilla code should not change

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,102 @@
...
+    public bool AchievementHidden = false;
+
+    protected override sealed Achievement CreateTemplateEntity()
+    {
+        if (string.IsNullOrWhiteSpace(Name))
+        {
+            throw new InvalidOperationException("Achievement name cannot be null or empty during template creation.");
+        }
+
+		return new Achievement($"{Mod.Name.ToUpper()}_{Name.ToUpper()}", this);
```
Chicken-Bones: Is there a reason we need to follow the vanilla convention rather than the tML convention here? I'd rather we use `FullName` if possible.

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// <summary>
+    /// Gets the localized friendly name of the achievement.
+    /// </summary>
+    public virtual LocalizedText FriendlyName => Mod.GetLocalization($"{LocalizationCategory}.{Name}.Name");
+
+    /// <summary>
+    /// Gets the localized description of the achievement.
+    /// </summary>
+    public virtual LocalizedText Description => Mod.GetLocalization($"{LocalizationCategory}.{Name}.Description");
```
Chicken-Bones: As per other `ModType` examples, should use `this.GetLocalization(nameof(Description));`

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,102 @@
...
+
+	protected override sealed void Register()
+    {
+	    ModTypeLookup<ModAchievement>.Register(this);
```
Chicken-Bones: Move below guard statements

patches/tModLoader/Terraria/ModLoader/ModAchievement.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    public override void Load()
+    {
+    }
+
+    public override void Unload()
+    {
+        Main.Achievements.Unregister(Achievement);
+        Achievement.OnCompleted -= OnCompleted;
+        base.Unload();
```
Chicken-Bones: No need to call `base.Unload`, it's empty

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -132,6 +132,8 @@ private static void Load(CancellationToken token = default)
 
 			ModContent.Load(token);
 
+			Main.Achievements.Load(); // Reload the achievements with each mod change to make sure any new mod's achievements are loaded with their completion value set.
+
```
Chicken-Bones: This should be in `ModContent.Load` or `ModContent.ResizeArrays`

patches/tModLoader/Terraria/UI/AchievementAdvisor.cs.patch
```diff
@@ -0,0 +1,30 @@
...
+ 		Rectangle frame = card.frame;
+ 		frame.X += 528;
+-		spriteBatch.Draw(_achievementsTexture.Value, position, frame, color, 0f, Vector2.Zero, scale, SpriteEffects.None, 0f);
++
++		if (card.achievement.ModAchievement != null) {
++			position = Vector2.Zero;
++			frame = new Rectangle(64, 0, 64, 64);
++		}
++		
++		spriteBatch.Draw(card.achievement.ModAchievement?.Texture.Value ?? _achievementsTexture.Value, position, frame, color, 0f, Vector2.Zero, scale, SpriteEffects.None, 0f);
```
Chicken-Bones: Use local scoping trick here too?

patches/tModLoader/Terraria/UI/InGamePopups.cs.patch
```diff
@@ -0,0 +1,13 @@
+--- src/TerrariaNetCore/Terraria/UI/InGamePopups.cs
++++ src/tModLoader/Terraria/UI/InGamePopups.cs
+@@ -57,7 +_,9 @@
+ 			_title = achievement.FriendlyName.Value;
+ 			int num = (_iconIndex = Main.Achievements.GetIconIndex(achievement.Name));
+ 			_achievementIconFrame = new Rectangle(num % 8 * 66, num / 8 * 66, 64, 64);
+-			_achievementTexture = Main.Assets.Request<Texture2D>("Images/UI/Achievements", AssetRequestMode.AsyncLoad);
++			if (achievement.ModAchievement != null)
++				_achievementIconFrame = new Rectangle(0, 0, 64, 64);
++			_achievementTexture = achievement.ModAchievement?.Texture ?? Main.Assets.Request<Texture2D>("Images/UI/Achievements", AssetRequestMode.AsyncLoad);
```
Chicken-Bones: Can just replace `_achievementTexture` here in the lines below rather than null coalescing

patches/tModLoader/Terraria/ModLoader/Mod.cs
```diff
@@ -79,7 +79,7 @@ public partial class Mod
 	/// The display name of this mod in the Mods menu.
 	/// </summary>
 	public string DisplayName { get; internal set; }
-	
+
```
Chicken-Bones: auto formatters be damned

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/Achievements/ItemPickupCondition.cs.patch
```diff
@@ -0,0 +1,48 @@
...
+ {
+ 	private const string Identifier = "ITEM_PICKUP";
+@@ -10,8 +_,10 @@
+ 	private static bool _isListenerHooked;
+ 	private short[] _itemIds;
+ 
++	private static string IDToIdentifier(short itemId) => ModLoader.ItemLoader.GetItem(itemId)?.FullName ?? itemId.ToString();
++
+-	private ItemPickupCondition(short itemId)
++	public ItemPickupCondition(short itemId)
```
Chicken-Bones, line 20: There's no need to make these `public` because there are static factory `Create` methods. Arguably though, the factory methods are noise/bloat/non idiomatic, so if that's why you chose to expose the constructors instead that's fine

JavidPack: `ItemPickupCondition.Create` returns a `AchievementCondition`, not a `ItemPickupCondition`. Same with the other classes. I guess it is technically safe to change their return values (although they are public, but I don't expect anyone to mind that a non-supported method changed return type in an update), but then I'd just have another method in the chain. `ModAchiemvent.AddNPCKilledCondition->AddCondition->NPCKilledCondition.Create->NPCKilledCondition.ctor` instead of `ModAchiemvent.AddNPCKilledCondition->AddCondition->NPCKilledCondition.ctor`.

We should definitely revisit this more, it is quite messy. I have an advanced example I've been working on not yet pushed.

ExampleMod/Content/NPCs/ExampleCritter.cs
```diff
@@ -152,6 +152,21 @@ public override void HitEffect(NPC.HitInfo hit) {
...
+		public override void OnHitByItem(Player player, Item item, NPC.HitInfo hit, int damageDone) {
+			ModContent.GetInstance<AdvancedExampleAchievement>().TotalDamageCondition.Value += damageDone;
+			if (item.type == ItemID.IronPickaxe) {
+				ModContent.GetInstance<AdvancedExampleAchievement>().IronPickaxeCondition.Complete();
+			}
+		}
+
+		public override void OnHitByProjectile(Projectile projectile, NPC.HitInfo hit, int damageDone) {
+			// OnHitByProjectile can be called on servers, so we need to check this to make sure to only access the achievement on a client to avoid null exceptions.
+			if (!Main.dedServ) {
```
Chicken-Bones: We generally prefer `Main.netMode != NetmodeID.Server`. Do we want an `owner` check to ignore traps and other player's projectiles?

Chicken-Bones (approved):

ExampleMod/Content/NPCs/ExampleCritter.cs
JavidPack: Made those changes. Docs say the method is only called on the owner anyway, so checking .owner is not needed.

Chicken-Bones (approved):

PR #4520, opened 2025-01-27
Fix various ProjectileLoader/NPCLoader.ReceiveExtraAI exception messages, optimize message size, and reorder modded and global bytes in message

### LINKED ISSUE #3275: [1.4.3.6 and 1.4.4] Projectile/NPCLoader.WriteExtraAI returns valid data with no mods, causing excess data being sent
### Version

Latest 1.4-stable and 1.4.4

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description
https://github.com/tModLoader/tModLoader/blob/a09f58addd99e38225c6f76d0b59d8f0cff4ec25/patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs#L210
https://github.com/tModLoader/tModLoader/blob/a09f58addd99e38225c6f76d0b59d8f0cff4ec25/patches/tModLoader/Terraria/ModLoader/IO/BitWriter.cs#L24-L25
`bitWriter.Flush(modWriter);` causes the underlying stream to progress one byte even if no data was written by any mods, due to the `BitWriter` calling `Write7BitEncodedInt` unconditionally. This causes the method to return an array with the contents of a single byte with the value 0. This in turn always sends this empty data (**one byte for the length, one byte for the data**):

Showcase for Projectile netcode (less complex for the NPC equivalent despite a comment mentioning that it uses the same approach!)
```cs
//In NetMessage.cs
byte[] extraAI = !ModNet.AllowVanillaClients ? ProjectileLoader.WriteExtraAI(projectile) : null;
bool hasExtraAI = extraAI?.Length > 0;
//...
if (hasExtraAI) {
	ProjectileLoader.SendExtraAI(writer, extraAI);
}
```
https://github.com/tModLoader/tModLoader/blob/a09f58addd99e38225c6f76d0b59d8f0cff4ec25/patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs#L185-L191

### Log File

_No response_

### Steps to reproduce

Disable all mods, place breakpoint on the method call in `NetMessage`, Debug.

### Expected Behavior
No excess data should be sent if no mods have written any data.

### Actual Behavior
Excess data is being sent even when no data was written by any mods.

### Reproduction frequency

_No response_

### Additional Information

The current implementation of BitWriter.Flush does not clear the underlying data (`i`, `bytes`, and `cur`), should it?
https://github.com/tModLoader/tModLoader/blob/a09f58addd99e38225c6f76d0b59d8f0cff4ec25/patches/tModLoader/Terraria/ModLoader/IO/BitWriter.cs#L24-L34

JavidPack:
This PR fixes various issues with Projectile and NPC network syncing.

## Fix ReceiveExtraAI exception reported byte length messages
The current exception messages are misleading. For example, when sending 100 bits and 3 bytes, but reading 100 bits and 1 byte, the error should say that 2 of 3 bytes were unread, but the way it is currently doesn't take into account the `BitReader` length overhead or the `BitReader` bytes.
Old
`Read underflow 2 of 17 bytes in ReceiveExtraAI`
Fixed
`Read underflow 2 of 3 bytes in ReceiveExtraAI`

(There are 17 bytes total, but only 3 are pertaining to the error. 1 byte for BitReader length, 13 bytes to read 100 bits, and 3 bytes of actual data.)

Similarly, if just sending bytes and no bits, the `BitReader` length still requires a byte and is still not taken into account. Here is sending 10 bytes but reading 5:

Old
`Read underflow 5 of 11 bytes in ReceiveExtraAI`
Fixed
`Read underflow 5 of 10 bytes in ReceiveExtraAI`

## Fix "may be caused by one of these GlobalProjectile" hint
For some reason, the `ProjectileLoader.ReceiveExtraAI` code did not report `GlobalProjectile` classes as potential sources of the sync exception as `NPCLoader.ReceiveExtraAI` does. The code has been updated to match and `GlobalProjectile` classes will be mentioned in exception logs as expected now.

In addition, the last GlobalProjectile/NPC accessed during the ReceiveExtraAI enumeration is tracked so that the listing of the potential cause of an exception is limited to just the Global classes that were actually called. 

The exception message previously was logged twice, so the duplicate logging of the exception was removed.

### Fix #3275, WriteExtraAI can now properly send 0 bytes of modded data
Due to `BitWriter.Flush` writing the bit length, code intended to skip sending modded Projectile/NPC data did not work. Data was always sent even if the ModProjectile/NPC and all GlobalProjectile/NPC did not write any bytes or bits. 

Code has been added to check if no real data is being sent and `hasExtraAI` bools now work correctly to limit sending this extra data. `SyncNPC` message code did not originally have a `hasExtraAI` check, likely due to some legacy consideration such as all available bits in the existing message being used already. There are unused bits now so this has been implemented.

### Reorder WriteExtraAI to make underflow exceptions more reliable
Finally, the order of the bytes and bits has been changed to simplify the code and make the exception messages more reliable. The old order was `ModProjectile Bytes, GlobalProjectile BitWriter, GlobalProjectile Bytes, new order is `GlobalProjectile BitWriter, ModProjectile Bytes, GlobalProjectile Bytes`. (both for Projectile and NPC.) 

The old order meant that an error in reading `ModProjectile` bytes would make the error reporting for `BitReader.MaxBits` inaccurate. It would potentially mislead a modder.

# Porting Notes
No effort by modders should be required.

### DIFF
--- a/patches/tModLoader/Terraria/MessageBuffer.cs.patch
+++ b/patches/tModLoader/Terraria/MessageBuffer.cs.patch
@@ -269,7 +269,7 @@
  					item4.position = position3;
  					item4.velocity = velocity3;
  					item4.active = true;
-@@ -1419,11 +_,18 @@
+@@ -1419,11 +_,16 @@
  				if (num158 == 668)
  					NPC.deerclopsBoss = num155;
  
@@ -278,10 +278,8 @@
  					nPC5.releaseOwner = reader.ReadByte();
  
 +				// Extra AI is read into a temporary buffer for parity with ProjectileLoader code, and to avoid exceptions causing underreads.
-+				byte[] extraAI = bitsByte12[3] ? NPCLoader.ReadExtraAI(reader) : null;
-+				if (extraAI != null) {
-+					NPCLoader.ReceiveExtraAI(nPC5, extraAI);
-+				}
++				if (bitsByte12[3])
++					NPCLoader.ReceiveExtraAI(nPC5, NPCLoader.ReadExtraAI(reader));
 +
  				break;
  			}
--- a/patches/tModLoader/Terraria/MessageBuffer.cs.patch
+++ b/patches/tModLoader/Terraria/MessageBuffer.cs.patch
@@ -269,7 +269,7 @@
  					item4.position = position3;
  					item4.velocity = velocity3;
  					item4.active = true;
-@@ -1419,11 +_,15 @@
+@@ -1419,11 +_,18 @@
  				if (num158 == 668)
  					NPC.deerclopsBoss = num155;
  
@@ -278,7 +278,10 @@
  					nPC5.releaseOwner = reader.ReadByte();
  
 +				// Extra AI is read into a temporary buffer for parity with ProjectileLoader code, and to avoid exceptions causing underreads.
-+				NPCLoader.ReceiveExtraAI(nPC5, NPCLoader.ReadExtraAI(reader));
++				byte[] extraAI = bitsByte12[3] ? NPCLoader.ReadExtraAI(reader) : null;
++				if (extraAI != null) {
++					NPCLoader.ReceiveExtraAI(nPC5, extraAI);
++				}
 +
  				break;
  			}
--- a/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
@@ -354,18 +354,19 @@ public static class NPCLoader
 
 	public static byte[] WriteExtraAI(NPC npc)
 	{
+		// Data is ordered as follows: GlobalProjectile BitWriter, ModProjectile Bytes, GlobalProjectile Bytes
 		using var stream = new MemoryStream();
 		using var modWriter = new BinaryWriter(stream);
 
-		npc.ModNPC?.SendExtraAI(modWriter);
-
 		using var bufferedStream = new MemoryStream();
-		using var globalWriter = new BinaryWriter(bufferedStream);
+		using var binaryWriter = new BinaryWriter(bufferedStream);
 
 		BitWriter bitWriter = new BitWriter();
 
+		npc.ModNPC?.SendExtraAI(binaryWriter);
+
 		foreach (var g in HookSendExtraAI.Enumerate(npc)) {
-			g.SendExtraAI(npc, bitWriter, globalWriter);
+			g.SendExtraAI(npc, bitWriter, binaryWriter);
 		}
 	
 		bitWriter.Flush(modWriter);
@@ -394,9 +395,10 @@ public static class NPCLoader
 
 		GlobalNPC lastGlobalNPC = null;
 		try {
-			npc.ModNPC?.ReceiveExtraAI(modReader);
-
 			BitReader bitReader = new BitReader(modReader);
+
+			var bitReaderEnd = stream.Position;
+			npc.ModNPC?.ReceiveExtraAI(modReader);
 
 			foreach (var g in HookReceiveExtraAI.Enumerate(npc)) {
 				lastGlobalNPC = g;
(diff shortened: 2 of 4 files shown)


patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -370,7 +370,9 @@ public static byte[] WriteExtraAI(NPC npc)
 		bitWriter.Flush(modWriter);
 		modWriter.Write(bufferedStream.ToArray());
 
-		return stream.ToArray();
+		byte[] bytes = stream.ToArray();
+		// If the only byte is the bitWriter.Flush length byte, no extra data.
+		return bytes.Length == 1 && bytes[0] == 0 ? null : bytes;
```
Chicken-Bones: Kinda feels like the `bytes[0] == 0` should be an assert when `bytes.Length == 1`

patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
```diff
@@ -268,11 +269,13 @@ public static void ReceiveExtraAI(Projectile projectile, byte[] extraAI)
 		catch (IOException e) {
 			string message = $"Error in ReceiveExtraAI for Projectile {projectile.ModProjectile?.FullName ?? projectile.Name}";
 			if (anyGlobals) {
-				message += ", may be caused by one of these GlobalNPCs:";
+				message += ", may be caused by one of these GlobalProjectiles:";
```
Chicken-Bones, line 273: We can potentially make this message 50% better if we store the last `GlobalProjectile` which had its hook called, and break the loop below after encountering it.

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -401,7 +401,15 @@ public static void ReceiveExtraAI(NPC npc, byte[] extraAI)
 			}
 
 			if (stream.Position < stream.Length) {
-				throw new IOException($"Read underflow {stream.Length - stream.Position} of {stream.Length} bytes in ReceiveExtraAI, more info below");
+				// Calculate original length of Read7BitEncodedInt
```
Chicken-Bones: If an under/over-read occurred in `ModNPC.ReceiveExtraAI` the contents of the `BitReader` could be complete garbage.

patches/tModLoader/Terraria/ModLoader/ProjectileLoader.cs
JavidPack: That would add overhead to the normal code path, wouldn't it? This is an exception, I feel like this code looping over all is better than making a local variable always run.

JavidPack: Oh wait, I see what you mean, since we are culling out GlobalProjectiles that couldn't be the issue.

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -401,19 +408,29 @@ public static void ReceiveExtraAI(NPC npc, byte[] extraAI)
...
+			if (lastGlobalNPC != null) {
 				message += ", may be caused by one of these GlobalNPCs:";
 				foreach (var g in HookReceiveExtraAI.Enumerate(npc)) {
 					message += $"\n\t{g.FullName}";
+					if (lastGlobalNPC == g)
+						break;
 				}
 			}
 
-			Logging.tML.Error(message, e);
```
Chicken-Bones, line 416: Please include the original exception. It may be a legitimate exception in a mod's netcode and we don't rethrow. (I made a mistake in my recommendation earlier, seeing it was unused in `ProjectileLoader`)

JavidPack: The exception will already be logged before this message, so I figured we can omit it.

image

Chicken-Bones: Hmm, I feel like relying on the `SilentlyCaughtException` is bad practice, but maybe I'm wrong

patches/tModLoader/Terraria/MessageBuffer.cs.patch
```diff
@@ -278,7 +278,10 @@
  					nPC5.releaseOwner = reader.ReadByte();
  
 +				// Extra AI is read into a temporary buffer for parity with ProjectileLoader code, and to avoid exceptions causing underreads.
-+				NPCLoader.ReceiveExtraAI(nPC5, NPCLoader.ReadExtraAI(reader));
++				byte[] extraAI = bitsByte12[3] ? NPCLoader.ReadExtraAI(reader) : null;
++				if (extraAI != null) {
```
Chicken-Bones:
```cs
if (bitsByte12[3])
    NPCLoader.ReceiveExtraAI(nPC5, NPCLoader.ReadExtraAI(reader));

direwolf420:
Would be very nice if this got merged before the february release so that players can enjoy the optimizations in march

PR #4524, opened 2025-02-03
description_workshop generates by default when creating mods

flye-name:
### What is the new feature?
Added a `description_workshop.txt` with an example description to the Templates folder. 
Upon creating a new mod, tModLoader would automatically add said file to the mod's folder.

A lot of modders (myself included up until a few days ago) are completely unaware of the existence of the `description_workshop.txt` file. Having it automatically generate would help these modders realize that they can separate the description for the workshop page and that of the in-game page.

This would also prevent cases where updating a mod that lacks a `description_workshop.txt` would completely reset its workshop page to use the `description.txt` even if the prior version had a different description written through the steam interface (something that has happened to me)

I don't think so? Maybe allowing modders to write base descriptions right from the mod creation menu but that's an entirely different thing.

### ExampleMod
ExampleMod already contains a `description_workshop.txt`, which further proves that this should be a thing by default as well.

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/de-DE/tModLoader.json
@@ -75,6 +75,8 @@
 		"BetaModCantPublishError": "Du kannst keinen Mod der in einer Beta-Version von tModLoader erstellt wurde veröffentlichen",
 		// "ModDescriptionMissing": "The file {0} doesn't exist. Please create a valid description for your mod.",
 		// "ModDescriptionLengthTooShort": "Mod description is too short, please create a valid description for your mod by modifying {0}.",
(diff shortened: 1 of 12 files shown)


Silencersn:
A question about `description_workshop.txt`.

Is there any way to let `description_workshop.txt` match different languages?
(there is the same issue with the DisplayName)

For example:
`description_workshop.en-US.txt` for English
`description_workshop.zh-Hans.txt` for Chinese
(just like Localization filename)

Now, I always edit them in Steam.

JavidPack:
> A question about `description_workshop.txt`.
> 
> Is there any way to let `description_workshop.txt` match different languages? (there is the same issue with the DisplayName)
> 
> For example: `description_workshop.en-US.txt` for English `description_workshop.zh-Hans.txt` for Chinese (just like Localization filename)
> 
> Now, I always edit them in Steam.

That has been discussed before, and is tracked on https://github.com/tModLoader/tModLoader/issues/1728. So there is currently no way but it could be implemented. Be sure to 👍 the issue if this is something you want prioritized.

Issue #4065, opened 2024-02-14
Prevent hammering for tiles similar to PreventsTileRemovalIfOnTopOfIt


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

If a modder wants to make a tile like a chest or demon altar that prevents the blocks it's attached to (below) to not be messed with, they can use the two vanilla sets `TileID.Sets.PreventsTileRemovalIfOnTopOfIt` and `PreventsTileReplaceIfOnTopOfIt`. This however is insufficient because the tiles can be hammered, thus causing the object to break since the anchors are now invalid.

A current workaround is to use GlobalTile.CanKillTile (and CanExplode for completeness sake), which covers all of the use cases for the set, including hammering, **which @JavidPack on [discord](https://discord.com/channels/103110554649894912/1074860279948656843/1074989305405120574) pointed out may be unintended**. Currently, `CanKillTile` IS called in vanillas `CanPoundTile` which prevents hammering. There's also another separate `CanPoundTile` method used in a different context which doesn't call `CanKillTile`.

(For demon altar, this is handled manually in `CanPoundTile`, in extension to the sets.)

Another workaround would be a detour or IL edit of the hammering methods.

### What does this proposal attempt to solve or improve?

Avoid an edge case the modder has to solve with workarounds.
This would help some existing mods that add altar-like tiles to remove the workarounds.

### Which (other) solutions should be considered?
I guess this issue can be extended/split off into "A new hook for CanPoundTile".

_No response_

PR #4528, opened 2025-02-04
Basic tile entity and more

JavidPack:
One of the most common issues modders bring up in the chat is trouble getting a `ModTileEntity` to work. The current examples in `ExampleMod` are `ModPylon`s that aren't very useful for learning from since they obfuscate away some of the basic structure. There is a `TEScoreBoard.cs` in the Old folder that was never ported, but it's a bit too complex to serve as a basic template. In addition, the <https://github.com/tModLoader/tModLoader/wiki/Basic-Tile-Entity> page has some instruction but not enough to serve as an actual example modders can use as a starting point.

This PR adds `BasicTileEntity`, a full Tile Entity example. The intention was to make a complete working `ModTileEntity` example that could be used as a template for modders to use. The example is a "water barrel" tile, it collects rainwater and will give the player a free "water bucket" item when full when the player mines or right clicks on it. This interaction method is what the weapon rack and item display tiles use. The example teaches proper networking for interacting with a tile entity and other commonly requested features like custom mouse over/map text while being simple enough to be used as a template and reworked by a modder for other functionality.

In addition to the new example, there are various related bug fixes, helper methods, documentation, and new features.

# TODO
- [x] On May 1st, 2025, these changes will go live on the Stable branch. The https://github.com/tModLoader/tModLoader/wiki/Basic-Tile-Entity wiki page will need to be updated, some of the information is wrong and some of the boilerplate code can be removed.
- `TileObjectData.TopLeft` can be applied in a few more places in ExampleMod in a follow-up PR. A lot of ExampleMod tiles use hardcoded numbers for calculating the top left tile, this makes them difficult to adapt and prone to bugs when modders try to use them for their own tiles. Using `TileObjectData.TopLeft` will always work and lets the modder focus on the more important unique code of the examples.

# Additional Changes
- `Point16` now has a `Deconstruct` method. This can be used with `TileObjectData.TopLeft` to simplify initializing top left tile coordinate ints.
- `TEModdedPylon.NetPlaceEntityAttempt` now calls `OnNetPlace` as intended.
- `TileObjectData.TopLeft` added, returns the top left coordinates of a multitile.
- `Generic_Hook_AfterPlacement` added with a default implementation of the tile placement `Hook_AfterPlacement` method. It should work for the `HookPostPlaceMyPlayer` of any tile entity. `Generic_HookPostPlaceMyPlayer` helper property also added that returns a `PlacementHook` directly.
- `TileID.Sets.AvoidedByMeteorLanding` had a typo, `FallenLog` is now protected not `SillyBalloonGreen`
- `TileID.Sets.PreventsTileHammeringIfOnTopOfIt` added. Partially addresses #4065 

# Porting Notes
- `TEModdedPylon.OnNetPlace` was not being called correctly, this has been fixed. If you had workarounds in place they are no longer needed.
- Modders can use `TileObjectData.TopLeft` to greatly simplify logic in their mods. Get rid of `GetTopLeftTileInMultitile` if you copied it from the Basic Tile Entity wiki page.
- Modders can use `TileEntity.TryGet` to simplify logic. Get rid of `TryGetTileEntityAs` if you copied it from the Basic Tile Entity wiki page.
- Add `TileID.Sets.PreventsTileHammeringIfOnTopOfIt` to chest-like tiles that shouldn't break, similar to `PreventsTileRemovalIfOnTopOfIt` usage, if desired. Any workarounds used previously for this issue should be able to be removed.
- If you want to simplify some `ModTileEntity` code, you can use `TileObjectData.newTile.HookPostPlaceMyPlayer = ModContent.GetInstance<MyTileEntity>().Generic_HookPostPlaceMyPlayer;` to replace the `TileObjectData.newTile.HookPostPlaceMyPlayer = new PlacementHook(ModContent.GetInstance<MyTileEntity>().Hook_AfterPlacement, -1, 0, true);` line and delete the `Hook_AfterPlacement` method override. This is only if your tile entity resides on the top left corner of the multitile.

# Video
Here is the tile entity in action:

https://github.com/user-attachments/assets/04847def-bcc8-409e-b9da-42bbc0e8f9db

# Discussion (outdated, all resolved.)
1. ~~`ModTileEntity.Hook_AfterPlacement` is empty, but a generic implementation that should work for any tile is shown in BasicTileEntity.cs. We could update `ModTileEntity.Hook_AfterPlacement` directly with this default implementation, but that could potentially cause issues for existing mods.~~
    * Rather than adjusting `ModTileEntity.Hook_AfterPlacement` directly, I added an optional `Generic_Hook_AfterPlacement` with the default implementation. Maybe it's still better to replace `Hook_AfterPlacement` since no one would be using it without overriding it anyway.... `Generic_HookPostPlaceMyPlayer` helper property also added to simplify code for normal tile entities.
2. ~~https://github.com/tModLoader/tModLoader/wiki/Basic-Tile-Entity shows a `TryGetTileEntityAs` method, do we want to adapt this directly into the ModTileEntity class? Right now `BasicTileEntityTile.TryGetBasicTileEntity` is used in the example, but it might be good to have something all mods can use without making their own helper method. We just need to decide on the best name for the method and where it should be.~~
    * `TileEntity.TryGet` added. I think placing it in TileEntity makes sense, TileObjectData is for the anchor data and other utility classes would be a bit inconvenient to use. Tested and it's working with all existing vanilla TileEntity as well since they all follow the top left corner precedent.

3. ~~BasicTileEntity.cs currently holds all the `ModTileEntity`, `ModTile`, and `ModItem` classes for the example. We have done this in the past for similarly coupled examples, but I'm not sure if it is the best approach for this example. It might be worth moving them all to their own files in a TileEntities/BasicTileEntity folder instead, or just moving the ModItem and ModTile to their normal namespaces.~~
    * Yeah, this is fine.


ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+				Player player = Main.LocalPlayer;
+				player.noThrow = 2;
+				player.cursorItemIconEnabled = true;
+				player.cursorItemIconID = -1;
+				player.cursorItemIconText = StatusText.Format(tileEntity.WaterFillPercentage);
+			}
+		}
+
+		public override bool RightClick(int i, int j) {
+			if(TryGetBasicTileEntity(i, j, out BasicTileEntity tileEntity)) {
```
Chicken-Bones: Space between `if` and `(`

Chicken-Bones: I think tile entities are a sufficiently advanced place to be promoting the best practice of using guard statements

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+				Main.NewText(StatusText.Format(tileEntity.WaterFillPercentage));
+			}
+			
+			return true;
+		}
+
+		// This helper method retrieves the BasicTileEntity instance from any tile coordinate of the placed tile.
+		public static bool TryGetBasicTileEntity(int i, int j, out BasicTileEntity entity) {
+			Point16 topLeft = TileObjectData.TopLeft(i, j);
+			int index = ModContent.GetInstance<BasicTileEntity>().Find(topLeft.X, topLeft.Y);
```
Chicken-Bones: May as well just use `TileEntity.ByPosition` and `is` directly

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+				player.cursorItemIconEnabled = true;
+				player.cursorItemIconID = -1;
+				player.cursorItemIconText = StatusText.Format(tileEntity.WaterFillPercentage);
+			}
+		}
+
+		public override bool RightClick(int i, int j) {
+			if(TryGetBasicTileEntity(i, j, out BasicTileEntity tileEntity)) {
+				if(tileEntity.WaterFillPercentage == 100) {
+					Main.LocalPlayer.QuickSpawnItem(Main.LocalPlayer.GetSource_TileInteraction(i, j), ItemID.WaterBucket);
```
Chicken-Bones: Should the item be spawned by the server instead? If multiple players try and get water at the same time (within latency) they'll dupe it. Not the best example to set. What do item frames do?

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+			// This code shows detecting those situations, but the basic lesson is don't assume a TileEntity will always be present.
+			/*
+			Point16 topLeft = TileObjectData.TopLeft(i, j);
+			int index = ModContent.GetInstance<BasicTileEntity>().Find(topLeft.X, topLeft.Y);
+			if (index == -1) {
+				if (!Main.sectionManager.TileLoaded(topLeft.X, topLeft.Y)) {
+					return $"{name}: World section not loaded yet";
+				}
+				return $"{name}: No TileEntity found at coordinate";
+			}
```
Chicken-Bones: Is there a reason you wouldn't chain this into the `if` above? No need to requery the `Find` right? `MapHoverText` is only called for our tiles, and `TryGetBasicTileEntity` can only fail if the section isn't loaded, or the TE is broken

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+		// This water barrel will fill up in 1 minute of rain, or 3600 game updates.
+		private const int MaxFill = 3600;
+		private const int SyncInterval = 3600 / 10;
+
+		// We use a property to consolidate logic clamping water level value and triggering a network sync.
+		private int waterFillLevel;
+		public int WaterFillLevel {
+			get { return waterFillLevel; }
+			set {
+				int fillLevelValue = Math.Clamp(value, 0, MaxFill);
```
Chicken-Bones: call this `newFillLevel`

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,239 @@
...
+	// - ModItem - places the tile.
+
+	// BasicTileEntity is essentially a water barrel that collects water while it is raining. 
+	public class BasicTileEntity : ModTileEntity
+	{
+		private bool syncNeeded; // Used to track when a network sync is needed
+
+		// This water barrel will fill up in 1 minute of rain, or 3600 game updates.
+		private const int MaxFill = 3600;
+		private const int SyncInterval = 3600 / 10;
```
Chicken-Bones: Reference `MaxFill` here. Or consider doing `SyncIntervals = 10` and `waterFillLevel * MaxFill / SyncIntervals`

Chicken-Bones (changes requested):

ExampleMod/Content/TileEntities/BasicTileEntity.cs
JavidPack: Tile has been updated to follow vanilla conventions, which is to do the item dropping through KillTile. This also allows the player to "mine" the tile to obtain the dropped item now. This also prevents any potential desync duplication.

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -189,36 +174,55 @@ public override void MouseOver(int i, int j) {
...
-			int index = ModContent.GetInstance<BasicTileEntity>().Find(topLeft.X, topLeft.Y);
-			if (index == -1) {
-				entity = null;
-				return false;
+		public override void KillTile(int i, int j, ref bool fail, ref bool effectOnly, ref bool noItem) {
+			// Some tile entities that hold items, like Weapon Rack and Item Frame, release their item when hit with a pickaxe. When doing this they will not be destroyed on the first hit no matter how strong the pickaxe is.
+			// Others like Hat Rack and Mannequin do not release the item and prevent the tile from being mined at all.
+
+			// This example releases the item.
+			if (TileEntity.TryGet(i, j, out BasicTileEntity tileEntity) && tileEntity.WaterFillPercentage == 100) {
```
Chicken-Bones, line 205: Does the extra `WaterFillPercentage == 100` check work here in the case of two clients both sending the `TileManipulation` packet? Because the second packet will arrive after the tile has been emptied, and `fail = true` will not get set

JavidPack: So, I think it would work if 2 clients both right click, since that specifically sends `NetMessage.SendData(MessageID.TileManipulation, number: 0, number2: i, number3: j, number4: 1f);` and the `number4: 1f` means call `WorldGen.KillTile(x, y, fail: true);`

With 2 separate players mining the tile, and if the 2nd player has a strong pickaxe to 1-hit mine it, the `PickTile` code looks like it'll send `TileManipulation` without the `fail` flag, so the server will run `KillTile` as normal and it will potentially mine the tile. I'm not positive if that will lead to desync or not, but I think the tile would not show as mined on the 2nd players screen.

I'm not sure there is any way around that. From what I'm seeing I think the same would be true of 2 players mining Weapon Rack and Item Frame as well, since KillTile doesn't communicate back to the calling code which parameters to set in the `TileManipulation` netmessage.

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,243 @@
...
+		private int waterFillLevel;
+		public int WaterFillLevel {
+			get { return waterFillLevel; }
+			set {
+				int newFillLevel = Math.Clamp(value, 0, MaxFill);
+				if (waterFillLevel / SyncInterval != newFillLevel / SyncInterval) {
+					// To reduce network spam while raining, we only sync this tile entity at 10% fill intervals.
+					// This may or may not be the correct approach for other ModTileEntity, depending on how accurate of
+					// data the client will need access to.
+					syncNeeded = true;
```
Chicken-Bones: I don't think we should teach delayed sync here. I think we should just send the packet immediately, either here or where we set the value. Delayed syncs can cause the order of operations to change between server and client (not here, but in more complicated code that could be adapted from this example).

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,243 @@
...
+				player.cursorItemIconText = StatusText.Format(tileEntity.WaterFillPercentage);
+			}
+		}
+
+		public override bool RightClick(int i, int j) {
+			if (!TileEntity.TryGet(i, j, out BasicTileEntity tileEntity)) {
+				return true;
+			}
+			if (tileEntity.WaterFillPercentage == 100) {
+				/* While it is tempting to spawn the item like this, this will not work correctly in multiplayer. See below for the correct approach.
```
Chicken-Bones: `this will not work correctly` -> `this risks duplicating the item`

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,243 @@
...
+
+				// This example follows the convention of similar vanilla tile entitites that drop items: When right clicked, the tile will be "mined" and the item is released by the KillTile code running on the server, which will automatically spawn the item and adjust the tile entity data on the server preventing various potential network syncing issues.
+				WorldGen.KillTile(i, j, fail: true);
+				if (Main.netMode == NetmodeID.MultiplayerClient) {
+					// This is sending the KillTile manipulation for the (i, j) coordinates. number4 being 1f means KillTile should fail. This is what will cause KillTile to run on the server, ultimately spawning the Water Bucket item and setting WaterFillLevel back to 0. 
+					NetMessage.SendData(MessageID.TileManipulation, number: 0, number2: i, number3: j, number4: 1f);
+				}
+				// You can use a different approach for the item dropping, if you prefer, but care must be taken to ensure that multiple users interacting with the tile do not result in duplicate item spawns due to network latency or desync. This means using a custom ModPacket to inform the server to do something rather than running the code on the client.
```
Chicken-Bones: Simplify this text. Perhaps:

```
// If you don't want to use TileManipulation and KillTile, you may send a custom packet instead. Whatever you do, running the code on the server ensures that multiple users interacting with the tile do not result in duplicate item spawns due to network latency or desync.
```

ExampleMod/Content/TileEntities/BasicTileEntity.cs
```diff
@@ -0,0 +1,243 @@
...
+				Main.NewText(StatusText.Format(tileEntity.WaterFillPercentage));
+			}
+
+			return true;
+		}
+
+		public override void KillTile(int i, int j, ref bool fail, ref bool effectOnly, ref bool noItem) {
+			// Some tile entities that hold items, like Weapon Rack and Item Frame, release their item when hit with a pickaxe. When doing this they will not be destroyed on the first hit no matter how strong the pickaxe is.
+			// Others like Hat Rack and Mannequin do not release the item and prevent the tile from being mined at all.
+
```
Chicken-Bones, line 203: A line here noting that the `TIleManipulation` packet sent from `RightClick` will have `fail = true` so we just run this code to re-check if there's an item to drop.

Chicken-Bones (changes requested):

ExampleMod/Content/TileEntities/BasicTileEntity.cs
JavidPack: Delay has been removed.

Chicken-Bones (approved):

JavidPack:
https://github.com/tModLoader/tModLoader/wiki/Basic-Tile-Entity updated.

PR #4539, opened 2025-02-15
PreHoverInteract hook for NPCs

BasicallyIAmFox:
### What is the new feature?
Adds new `NPC` hook called `PreHoverInteract`

There is currently no way to override or piggyback off of right clicking on or hovering over an NPC. There are chat hooks, but recent Terraria NPC like Old Shaking Chest are interacted with outside of the normal chat range and logic, this new hooks supports that approach for modded content.

We could add a `NPCID.Sets.IsInteractible` like we did recently for `ProjectileID` to allow smart interact targeting an `NPC`, but `CanChat` works just fine for this functionality. We could also add more granular controls to which vanilla logic is skipped, but that would need more input from modders for a feature they haven't even experienced yet.

Vanilla tiles and projectiles draw an interaction outline when smart interact targets them, but NPC don't. They draw the chat icon over the npc head usually, or just the mouse cursor icon in the case of Old Shaking Chest. We could show an interaction outline example with the `ExampleDrawBehindNPC` example, but it doesn't match the vanilla experience.

```cs
public override bool PreHoverInteract(NPC npc, bool mouseIntersects) {
  if (!WikithisConfig.Config.CanWikiNPCs)
    return true;

  if (mouseIntersects && WikithisSystem.WikiKeybind.JustPressed && GetWiki<NPCWiki>().Entries.TryGetValue((short)npc.netID, out var entry)) {
    entry.OpenWikiPage(false);
  }
  return true;
}
```

### ExampleMod updates
`ExampleDrawBehindNPC` is now right-clickable alike Old Shaking Chest NPC. This example is also the 1st instance of an `NPC.SimpleStrikeNPC` example in ExampleMod, which is something we added but never showcased.

P.S. most recent changes were done by jopo; i simply updates the description to match changes


patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -92,6 +92,10 @@ public virtual void ModifyHoverBoundingBox(NPC npc, ref Rectangle boundingBox)
 	{
 	}
 
+	public virtual void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects, bool smartCursorInteraction)
```
Chicken-Bones: 1. Needs docs. Docs should include the conditions for the hook being called (`ShowNameOnHover` and mouse hovered or selected via smart interact )
2. `smartCursorInteraction` isn't quite the right name for this variable, but I've suggested removing it below anyway

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -4961,15 +4961,19 @@
...
-@@ -34825,6 +_,8 @@
+@@ -34825,8 +_,12 @@
  			if (nPC.type >= 87 && nPC.type <= 92)
  				value = new Microsoft.Xna.Framework.Rectangle((int)((double)nPC.position.X + (double)nPC.width * 0.5 - 32.0), (int)((double)nPC.position.Y + (double)nPC.height * 0.5 - 32.0), 64, 64);
  
 +			NPCLoader.ModifyHoverBoundingBox(nPC, ref value);
 +
  			bool flag = mouseRectangle.Intersects(value);
  			bool flag2 = flag || (SmartInteractShowingGenuine && SmartInteractNPC == i);
++			NPCLoader.OnHoverBoundingBox(nPC, value, flag, flag2);
```
Chicken-Bones: I'd rather this hook only be called `if (flag2)`, you can still pass `flag` as `mouseOver`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Main.cs.patch
BasicallyIAmFox: wdym '`if (flag2)`'? inside the if statement below or its own if stamaent
the only thing im concerned about is that if statment mbelow ignores certain npc types and ai styles

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -4969,7 +4969,7 @@
 +
  			bool flag = mouseRectangle.Intersects(value);
  			bool flag2 = flag || (SmartInteractShowingGenuine && SmartInteractNPC == i);
-+			NPCLoader.OnHoverBoundingBox(nPC, value, flag, flag2);
```
Chicken-Bones: Hook should only be called `if (flag2)`

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -85,14 +85,22 @@ public virtual void ModifyTypeName(NPC npc, ref string typeName)
 
 	/// <summary>
 	/// Allows you to modify the bounding box for hovering over the given NPC (affects things like whether or not its name is displayed).
+	/// <br/>This hook is not called if <see cref="NPC.ShowNameOnHover"/> is set to <see langword="false">.
```
Chicken-Bones: `This hook is only called when <see cref="NPC.ShowNameOnHover"/> returns <see langword="true">`

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -85,14 +85,22 @@ public virtual void ModifyTypeName(NPC npc, ref string typeName)
...
 
-	public virtual void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects, bool smartCursorInteraction)
+	/// <summary>
+	/// Gets called for hovering over the given NPC (things like whether or not its name is displayed).
+	/// <br/>This hook is not called if <see cref="NPC.ShowNameOnHover"/> is set to <see langword="false">.
+	/// </summary>
+	/// <param name="npc">The NPC in question.</param>
+	/// <param name="boundingBox">The bounding box used for determining whether or not the NPC counts as being hovered over.</param>
+	/// <param name="mouseIntersects">Whether or not mouse is actually hovering over the NPC's bounding box.</param>
+	public virtual void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects)
```
Chicken-Bones: Is the `boundingBox` param actually useful?

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
```diff
@@ -85,14 +85,22 @@ public virtual void ModifyTypeName(NPC npc, ref string typeName)
...
 	}
 
-	public virtual void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects, bool smartCursorInteraction)
+	/// <summary>
+	/// Gets called for hovering over the given NPC (things like whether or not its name is displayed).
+	/// <br/>This hook is not called if <see cref="NPC.ShowNameOnHover"/> is set to <see langword="false">.
+	/// </summary>
+	/// <param name="npc">The NPC in question.</param>
+	/// <param name="boundingBox">The bounding box used for determining whether or not the NPC counts as being hovered over.</param>
+	/// <param name="mouseIntersects">Whether or not mouse is actually hovering over the NPC's bounding box.</param>
```
Chicken-Bones:
```
Called when the the NPC is selected via smart interact, or when the mouse is hovering over it.
<br/>Use to ...
<br/>
<br/>Only works on NPCs where <see cref="NPC.ShowNameOnHover"/> returns <see langword="true">
```

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1146,15 +1146,15 @@ public static void ModifyHoverBoundingBox(NPC npc, ref Rectangle boundingBox)
...
 
-	private delegate void DelegateOnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects, bool smartCursorInteraction);
+	private delegate void DelegateOnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects);
 	private static HookList HookOnHoverBoundingBox = AddHook<DelegateOnHoverBoundingBox>(g => g.OnHoverBoundingBox);
-	public static void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects, bool smartCursorInteraction)
+	public static void OnHoverBoundingBox(NPC npc, Rectangle boundingBox, bool mouseIntersects)
 	{
 		if (npc.ModNPC != null)
-			npc.ModNPC.OnHoverBoundingBox(boundingBox, mouseIntersects, smartCursorInteraction);
+			npc.ModNPC.OnHoverBoundingBox(boundingBox, mouseIntersects);
```
Chicken-Bones: simpler syntax with `?`

`npc.ModNPC?.OnHoverBoundingBox(...)`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
BasicallyIAmFox: wouldn'tit be better to have "only works on npcs where" in remarks

patches/tModLoader/Terraria/ModLoader/GlobalNPC.cs
BasicallyIAmFox: it might be but can't think of any ues cases tbh

Chicken-Bones: Remove it and we're good

JavidPack:
Thanks. This PR ended up being more relevant than I first thought.

BasicallyIAmFox:
guess im legally obligated to update my mod now

PR #4546, opened 2025-02-17
Server-side Upload Fixes

xNarnia:
### What is the bug?
Server-side mods currently do not publish from the in-game mod browser or command line.

### How did you fix the bug?
- Added pre-publish CLI to test `server-side` mods in `Main.dedServ` environment
- Allow publish tags and publicity to command line with `args`
  - Made sure to prevent duping of tags and prevent illegal tags
- CLI exit codes return to client for logic to move the publish menu
- Patches tested with every mod-side use-case + `server-side` mod reference use cases
 
There are these alternatives.
- Use the command line to publish the `server-side` mods. No client publishing for `server-side` mods.
- Client-side `server-side` publish without pre-publish checking in `Main.dedServ` environment.
- Client-side `server-side` publishing only. Keep pre-publish check, remove command line publishing. Removes a lot of invasive code changes. (Very open to this alternative)

### Considerations
There is [Line#173](https://github.com/tModLoader/tModLoader/compare/1.4.4...xNarnia:tModLoader:1.4.4?template=new_feature.md#diff-7b9c571be613901c926ea85768271b4e062b2d9c851e76648fbc1e1fd2e32069R173) for a localization key. I wanted to have this reviewed first before implementation a localization key to make sure it was okay.
```
if (bp.side != ModSide.Server && Main.dedServ)
{
	// After review, will add a localization key for this error
	// Message: "You can only publish server-side mods using the command line."
	Utils.ShowFancyErrorMessage(Language.GetTextValue("tModLoader.LoadError", modFile.Name), Interface.modSourcesID);
}
```

### Notes
For CLI for publishing, example command of 
`-server -steam -publish ToBeDeleted -publishtags Library&tModLoader.TagsContent -publicity 2`
will result in a public visibility mod with the Content tag. The Library tag is not properly keyed so it won't be set

### TODO
- [ ] Test on Linux or Mac
  - [x] command line publish
  - [ ] Client check for testservermodloading (make sure console appears)
  - [ ] Show server console option in host and play (make sure it was broken on stable, fixed in this PR
 - [x] Update https://github.com/tModLoader/tModLoader/wiki/Command-Line


JavidPack:
Note: I did more testing, Linux command line publishing and client testing features should work, but the console will not show unless it was visible to begin with (this is true for the new testing feature and show console for host and play). It seems this is a fundamental difference between how consoles work on Linux and Windows. I believe to fix this we would have to make a separate .sh file and launch that, but we can leave that for later if it becomes important. For now, everything works as before, it's just less clear on linux that the server side mod is being tested in the client.

PR #4548, opened 2025-02-21
[Feature] Enable developer features for local mods built from outside of "Mod Sources" folder

### LINKED ISSUE #4547: [Request] allow tModLoader to discover additional mod source directories outside of `ModSources`

Yes

Mod capability as a Modder

### Description

tModLoader should be able to determine the path to the mod source directory of built mods for additional feature support without requiring hard-coded paths.

After discussion on the Discord, it's been determined the ideal would be an additional `Info` property defining the directory the mod was built from.

Currently, tModLoader is functionally limited to only supporting mods in the `ModSources` directory. This limits features like localization file hot reloading, formatting, and generation from working without the mod source being present there and also prevents easily publishing mods (the highest priority) or even building in-game (which is a low priority, personally).

Symlinks work, but require admin privileges on Windows.

steviegt6:
### What is the new feature?

*Resolves #4547.*

Adds a `modSource` property to a mod's build properties when built that allows tModLoader to then discover the mod's source directory.

Allows tModLoader to treat mods not in `ModSources` with the same level of development support as mods in `ModSources`.

See #4547 for discussion.

# How to use
1. Make a folder where you want to store custom mod source folders, something like `CustomModSources`, for example. This will be similar to the `ModSources` folder in the default saves folder.
2. Copy the `tModLoader.targets` file from the original `ModSources` folder to your new folder.
3. Move each mod's source code folder to this folder.
4. For each mod, you'll need to build the mod directly at least once. This is usually done in [Visual Studio](https://github.com/tModLoader/tModLoader/wiki/Why-Use-an-IDE#build), but you can also just open the folder and type in `dotnet build` into the address bar and press `Enter`. 
5. After this, the mod should appear in the Develop Mods menu in-game if you need to build in-game. The blue folder icon indicates that the mod was built from a custom location, hovering over it will show that path.

____
    
____
    
___

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
@@ -58,7 +58,7 @@ public class WorkshopPublishInfoStateForMods : AWorkshopPublishInfoState<TmodFil
 
 		if (Main.MenuUI.CurrentState?.GetType() != typeof(UIReportsPage)) {
 			// Copy the used preview image to the mod's source directory if it's not a resize and if one isn't there already.
-			string iconWorkshopPath = Path.Combine(ModCompile.ModSourcePath, _dataObject.Name, "icon_workshop.png");
+			string iconWorkshopPath = Path.Combine(ModCompile.GetModSource(_dataObject), "icon_workshop.png");
 			if (_previewImagePath != iconWorkshopPath && !resizedPreviewImage && !File.Exists(iconWorkshopPath)) {
 				try {
 					File.Copy(_previewImagePath, iconWorkshopPath, overwrite: true);
--- a/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
@@ -59,7 +59,7 @@ public class WorkshopPublishInfoStateForMods : AWorkshopPublishInfoState<TmodFil
 
 		if (Main.MenuUI.CurrentState?.GetType() != typeof(UIReportsPage)) {
 			// Copy the used preview image to the mod's source directory if it's not a resize and if one isn't there already.
-			string iconWorkshopPath = Path.Combine(BuildProperties.ReadModFile(_dataObject).modSource, "icon_workshop.png");
+			string iconWorkshopPath = Path.Combine(_buildData["sourcesfolder"], "icon_workshop.png");
 			if (_previewImagePath != iconWorkshopPath && !resizedPreviewImage && !File.Exists(iconWorkshopPath)) {
 				try {
 					File.Copy(_previewImagePath, iconWorkshopPath, overwrite: true);
--- a/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
@@ -9,6 +9,7 @@ using Terraria.Audio;
 using Terraria.GameContent.UI.Elements;
 using Terraria.ID;
 using Terraria.Localization;
+using Terraria.ModLoader;
 using Terraria.ModLoader.Core;
 using Terraria.Social;
 using Terraria.Social.Base;
@@ -58,7 +59,7 @@ public class WorkshopPublishInfoStateForMods : AWorkshopPublishInfoState<TmodFil
 
 		if (Main.MenuUI.CurrentState?.GetType() != typeof(UIReportsPage)) {
 			// Copy the used preview image to the mod's source directory if it's not a resize and if one isn't there already.
-			string iconWorkshopPath = Path.Combine(ModCompile.GetModSource(_dataObject), "icon_workshop.png");
+			string iconWorkshopPath = Path.Combine(BuildProperties.ReadModFile(_dataObject).modSource, "icon_workshop.png");
 			if (_previewImagePath != iconWorkshopPath && !resizedPreviewImage && !File.Exists(iconWorkshopPath)) {
 				try {
 					File.Copy(_previewImagePath, iconWorkshopPath, overwrite: true);
--- a/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
@@ -210,9 +210,7 @@ public static class AssemblyManager
 			m.DisplayName = mod.properties.displayName;
 			m.TModLoaderVersion = mod.properties.buildVersion;
 			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null;
-
-			if (Directory.Exists(mod.properties.modSource))
-				m.SourceFolder = mod.properties.modSource;
+			m.SourceFolder = Directory.Exists(mod.properties.modSource) ? mod.properties.modSource : "";
 			return m;
 		}
 		catch (Exception e) {
--- a/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
@@ -209,7 +209,8 @@ public static class AssemblyManager
 			m.Side = mod.properties.side;
 			m.DisplayName = mod.properties.displayName;
 			m.TModLoaderVersion = mod.properties.buildVersion;
-			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null; 
+			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null;
+			m.SourceFolder = mod.properties.modSource;
 			return m;
 		}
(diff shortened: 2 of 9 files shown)


patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -412,9 +412,8 @@ private static bool DoesDotnetWork(string path)
 	internal void Populate()
 	{
 		Task.Run(() => {
-			var modSources = ModCompile.FindModSources();
+			var modSources = ModCompile.FindModSources(refindAllMods: true, out var modFiles);
```
Chicken-Bones: Lets keep it simple. 
```cs
// call before finding mod sources, in case someone has built a new mod externally after opening the game but before re-opening the sources menu
var modFiles = ModOrganizer.FindAllMods(); 
var modSources = ModCompile.FindModSources();
```

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
```diff
@@ -54,20 +54,26 @@ public BuildingMod(TmodFile modFile, BuildProperties properties, string path) :
 
 	public static readonly string ModSourcePath = Path.Combine(Program.SavePathShared, "ModSources");
 
-	internal static string[] FindModSources()
+	internal static string[] FindModSources(bool refindAllMods, out LocalMod[] builtMods)
```
Chicken-Bones: The extra arguments are just kinda painful. With my other comment we don't really need them.

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
```diff
@@ -54,20 +54,26 @@ public BuildingMod(TmodFile modFile, BuildProperties properties, string path) :
...
+
+		return modSources.ToArray();
 	}
 
 	// Silence exception reporting in the chat unless actively modding.
 	public static bool activelyModding;
 	internal static DateTime recentlyBuiltModCheckTimeCutoff = DateTime.Now - TimeSpan.FromSeconds(60);
 
-	public static bool DeveloperMode => Debugger.IsAttached || Directory.Exists(ModSourcePath) && FindModSources().Length > 0;
+	public static bool DeveloperMode => Debugger.IsAttached || Directory.Exists(ModSourcePath) && FindModSources(refindAllMods: false, out _).Length > 0;
```
Chicken-Bones: Might be good to use a cached result for the `FindModSources`. Will the `Directory.Exists(ModSourcePath)` potentially be invalidated now?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
steviegt6: Yes, it would be potentially invalidated: `ModSourcePath` is created upon resolving mod sources (in `FindModSources`), so it should be fine to omit instead.

Chicken-Bones: That guard probably exists explicitly to prevent that from happening then

steviegt6: I've updated it to skip the check but not create/search the directory in `FindModSources` if it doesn't exist.

patches/tModLoader/Terraria/ModLoader/UI/UIModSources.cs
```diff
@@ -412,9 +412,9 @@ private static bool DoesDotnetWork(string path)
 	internal void Populate()
 	{
 		Task.Run(() => {
+			var modFiles = ModOrganizer.FindAllMods();
```
Chicken-Bones, line 417: Add comment on why the order of these two calls matters

patches/tModLoader/Terraria/ModLoader/Core/ModCompile.cs
```diff
@@ -56,18 +56,38 @@ public BuildingMod(TmodFile modFile, BuildProperties properties, string path) :
...
+	}
+
+	internal static string GetModSource(TmodFile tmod)
+	{
+		var buildInfo = BuildProperties.ReadModFile(tmod);
+		return string.IsNullOrEmpty(buildInfo.modSource) ? Path.Combine(ModSourcePath, tmod.Name) : buildInfo.modSource;
 	}
 
+	internal static string GetModSource(LocalMod mod) => string.IsNullOrEmpty(mod.properties.modSource) ? Path.Combine(ModSourcePath, mod.Name) : mod.properties.modSource;
+
```
Chicken-Bones: Do we need this at all? It would only affect modders trying to publish a mod they had built on a prior tML update. A rebuild of the mod would ensure it has source path metadata

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/LocalizationLoader.cs
```diff
@@ -323,7 +327,7 @@ private static void UpdateLocalizationFilesForMod(Mod mod, string outputPath = n
 		};
 
 		// TODO: Maybe optimize to only recently built?
-		string sourceFolder = outputPath ?? Path.Combine(ModCompile.ModSourcePath, mod.Name);
+		string sourceFolder = outputPath ?? BuildProperties.ReadModFile(mod.File).modSource;
```
Chicken-Bones: Bleh, really second guessing keeping the `BuildProperties` on `LocalMod` rather than `Mod`, but anyway. Please copy `modSource` to a new property in `Mod`, call it `SourceFolder` perhaps, and assign it in `AssemblyManager.Instantiate`

Chicken-Bones: Maybe copy it only `if (Directory.Exists(modSource))` just so people who are looking through `Mod` instances in their debugger don't get nosy

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -209,7 +209,8 @@ private static Mod Instantiate(ModLoadContext mod)
 			m.Side = mod.properties.side;
 			m.DisplayName = mod.properties.displayName;
 			m.TModLoaderVersion = mod.properties.buildVersion;
-			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null; 
+			m.TranslationForMods = mod.properties.translationMod ? mod.properties.RefNames(true).ToList() : null;
+			m.SourceFolder = mod.properties.modSource;
```
Chicken-Bones: `Directory.Exists(mod.properties.modSource)) ? mod.properties.modSource : ""`

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -170,7 +170,7 @@ internal static void PublishMod(LocalMod mod, string iconPath)
 		if (bp.buildVersion != modFile.TModLoaderVersion)
 			throw new WebException(Language.GetTextValue("tModLoader.OutdatedModCantPublishError"));
 
-		var changeLogFile = Path.Combine(ModCompile.ModSourcePath, modFile.Name, "changelog.txt");
+		var changeLogFile = Path.Combine(BuildProperties.ReadModFile(modFile).modSource, "changelog.txt");
```
Chicken-Bones: Change this and the other usages over to use the new property

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/UI/States/WorkshopPublishInfoStateForMods.TML.cs
```diff
@@ -58,7 +59,7 @@ protected override void GoToPublishConfirmation()
 
 		if (Main.MenuUI.CurrentState?.GetType() != typeof(UIReportsPage)) {
 			// Copy the used preview image to the mod's source directory if it's not a resize and if one isn't there already.
-			string iconWorkshopPath = Path.Combine(ModCompile.ModSourcePath, _dataObject.Name, "icon_workshop.png");
+			string iconWorkshopPath = Path.Combine(BuildProperties.ReadModFile(_dataObject).modSource, "icon_workshop.png");
```
Chicken-Bones: _buildData["sourcesfolder"]

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
Chicken-Bones: You have `bp` just above here.

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -187,7 +187,7 @@ internal static void PublishMod(LocalMod mod, string iconPath)
 			{ "homepage", bp.homepage },
 			{ "description", bp.description },
 			{ "iconpath", iconPath },
-			{ "sourcesfolder", Path.Combine(ModCompile.ModSourcePath, modFile.Name) },
+			{ "sourcesfolder", BuildProperties.ReadModFile(modFile).modSource },
```
Chicken-Bones: `bp.modSource`

Chicken-Bones (changes requested):

JavidPack:
I've noticed that ExampleMod shows up twice during normal development. Other mods also will show up twice and it is impossible to tell which mod is the custom path. See #4576 for more information on the fix I propose

Issue #4542, opened 2025-02-15
[Feature Request] Add Support for Configurable Rigidity in Swaying Multitiles


Arcri:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently, multitiles that sway in wind cannot have their rigidity configured. All swaying tiles use the same wind physics calculations, resulting in floppy movement. This makes it difficult to create stiff-but-swaying objects like chandeliers that simply pivot that should move more rigidly than organic objects like vines or plants.

An example of the floppyness on chandeliers, where the crystal-blue one is a modded tile, and the copper chandelier is the vanilla reference: 

![Image](https://github.com/user-attachments/assets/27f19069-03ea-4460-b194-655f4c700d31)

### What does this proposal attempt to solve or improve?

This feature would solve cases where multitiles like chandeliers appear floppy and organic. Adding configurable rigidity would allow for more realistic movement across different material types.

JavidPack:
This has been resolved in [#4551](https://github.com/tModLoader/tModLoader/pull/4551)

PR #4551, opened 2025-02-24
ExampleChandelier and several new related ModTile hooks

JavidPack:
#4542 brought up that the recently added support for swaying tiles was insufficient to properly replicate tiles like Chandeliers. This PR adds `ExampleChandelier` and updates many other ExampleMod tiles to more accurately reflect the behavior (physics and flame visuals) of their vanilla counterparts.

# New Hooks
## AdjustMultiTileVineParameters
Tiles drawn using `TileID.Sets.MultiTileSway` default to behaving physically like the War Table Banner, or cloth. This looks very incorrect for other tiles. Chandeliers, suspended pots, and other "rigid" tiles set specific parameters to behave as they do. This hook allows modded tiles to set those same parameters. These parameters control the deformation, rigidity, selective movement, etc.

https://github.com/user-attachments/assets/777eee12-0cac-4ade-973f-300448867925

## GetTileFlameData
Flames in ExampleMod tile examples are usually drawn using `PostDraw`, but that does not work for tiles drawn to the screen rather than the tile render targets. While it is possible to use `TileCounterType.CustomNonSolid` and `SpecialDraw` to do this for tiles drawing to the screen, that would require recalculating all the tile sway physics offsets to position correctly. Instead, a hook in `TileDrawing.GetTileFlameData` was added to allow modders to use the same approaches as vanilla code. This new flame drawing code is concise and data driven, so it is more convenient. `TileDrawing.GetTileFlameData` is currently only called for `DrawMultiTileVinesInWind` tiles.

## EmitParticles
Due to legacy code, tiles emitting custom dust/gore has happened in `ModTile.DrawEffects` in ExampleMod. This approach requires a check for `if (Main.gamePaused || !Main.instance.IsActive || Lighting.UpdateEveryFrame && !Main.rand.NextBool(4))` to work in the same manner as vanilla tiles. This boilerplate code is unintuitive. Also, `DrawEffects` isn't called for tiles that return false in `PreDraw`, so it is ill suited for tiles drawing to the screen as swaying tiles do. The new `EmitParticles` method is called reliably and can replace `DrawEffects` particle code without the need for the boilerplate code. It also passes in a value indicating if the tile is visible or hidden using echo paint, which further reduces repeated code.

# Animation.NewTemporaryAnimation support
The `Animation.NewTemporaryAnimation` method drives temporary animation used by MushroomStatue, Mini Volcano, Large Volcano, FakeContainers, and FakeContainers2. (Chests are handled separately as they are an older system.) Up until now, this system was essentially off limits for modders because there was no way to use a custom frame sequence or frame rate. Support has been added, mods can call `Animation.RegisterTemporaryAnimation` to register an animation sequence and then use the returned value later in `Animation.NewTemporaryAnimation` to start an animation. This type of animation is per-tile, allowing modded tiles to animate an individual tile on-demand without implementing their own animation system that tracks tile coordinates and frame counts. This animation is also synced automatically to other clients.

https://github.com/user-attachments/assets/7f5eaa13-7dce-4ce9-aa37-1f3dfad277ea

# ExampleChandelier
`ExampleChandelier` showcases these new hooks. It is essentially a clone of several unique vanilla chandeliers meant to showoff the various `AdjustMultiTileVineParameters` parameters. It also shows off `Animation.NewTemporaryAnimation` to make a unique tile animation effect, similar to the [Mushroom statue](https://terraria.wiki.gg/wiki/Statues#Mushroom_Statue) or mini/large volcano tile animations.

# Torch God support
Modded torches now count towards torch god spawning and will extinguish and spawn torch god attacks. There is currently no control over the dust choice of the torch god attacks, however, since that is a bit complicated and uses systems tModLoader hasn't touched.

dotnet_2025-02-26_16-35-04

# The Constant rain extinguishing support
Modded torches now properly extinguish in rain in The Constant worlds rather than breaking. 

TODO: A `ToggleLight` hook could be added to support all light sources such as `ExampleLamp` and `ExampleCampfire`, but that hook isn't added in this PR. Please make a separate issue so we can gauge interest if you have need of this hook. The end result is that something like `ExampleCampfire.RightClick` and `HitWire` would call `ToggleLight`. This is fine, I'm just worried about changing how things are done for light tiles specifically, it could be confusing as to why some toggleable tiles use `HitWire` directly while others use `HitWire` to call code in `ToggleLight`.

image

# Other ExampleMod Changes
- `ExampleAnimatedTile` now has a proper flame texture. It also has correct `AdjustMultiTileVineParameters` implementation to match the vanilla firefly in a bottle tile.
- `ExampleCampfire` now uses `EmitParticles` instead of `DrawEffects` for spawning dust.
- `ExampleLamp` now uses `EmitParticles` instead of `DrawEffects` for spawning dust.
- `ExampleMusicBoxTile` now uses `EmitParticles` instead of `DrawEffects` for spawning dust.
- `ExampleTorch` now spawns dust when placed. This was missing.

# Porting Notes
- Revisit `TileDrawing.TileCounterType.MultiTileVine` tiles and see if any of the new features would help make them better.
- Consider moving dust and gore code from `DrawEffects` to `EmitParticles`. Doing so would allow simplifying code greatly by removing the boilerplate code.
- Update torch tiles with dust spawning if you had used `ExampleTorch` as a guide.


patches/tModLoader/removed_files.list
```diff
@@ -1,14 +0,0 @@
...
-ReLogic/Content/Sources/XnaContentSource.cs
-Terraria/DataStructures/AEntitySource_OnHit.cs
-Terraria/DataStructures/EntitySource_ByItemSourceId.cs
-Terraria/DataStructures/EntitySource_ByProjectileSourceId.cs
-Terraria/DataStructures/EntitySource_OnHit_ByItemSourceID.cs
-Terraria/DataStructures/EntitySource_OnHit_ByProjectileSourceID.cs
-Terraria/Libraries/Common/Ionic.Zip.CF.dll
-Terraria/Libraries/FNA/FNA.dll
-Terraria/Libraries/Mono/Steamworks.NET.dll
-Terraria/Libraries/Mono/System.Windows.Forms.dll
```
Chicken-Bones: oops?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2798,6 +2798,14 @@
  		int num = i - 1;
  		int num2 = i + 2;
  		int num3 = j - 1;
+@@ -52948,6 +_,7 @@
+ 		if (!tile.active())
+ 			return false;
+ 
++		// TODO: TryToggleLight handles the feature where rain extinguishes light sources in The Constant worlds. It is currently unsupported for modded tiles. Modded lights with WaterDeath will be destroyed instead of extinguished.
```
Chicken-Bones: Do any of the example tiles have wiring support at all? Separate PR (low priority) if not?

patches/tModLoader/Terraria/Animation.cs.patch
```diff
@@ -0,0 +1,10 @@
+--- src/TerrariaNetCore/Terraria/Animation.cs
++++ src/tModLoader/Terraria/Animation.cs
+@@ -162,6 +_,7 @@
+ 		}
+ 	}
+ 
++	// TODO: Need an example of this. Would require SetDefaults type as well maybe, or alternate approach. Network. Could just use tileType, but that won't allow multiple animations per tile.
```
Chicken-Bones: Auto assigned type is fine I suppose. Something like

```cs
static int myAnimType;

public void SetStaticDefaults() {
    myAnimType = Animation.RegisterTemporaryAnimation(frameRate: 5, frames: [1, 2, 2, 2, 5, 3]);
}
```

patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -415,6 +415,27 @@ public virtual void DrawEffects(int i, int j, SpriteBatch spriteBatch, ref TileD
...
+	/// </summary>
+	/// <param name="i"></param>
+	/// <param name="j"></param>
+	/// <param name="tileCache"></param>
+	/// <param name="tileFrameX"></param>
+	/// <param name="tileFrameY"></param>
+	/// <param name="tileLight"></param>
+	/// <param name="visible"></param>
+	/// <returns></returns>
+	public virtual bool DrawTilesEmitParticles(int i, int j, Tile tileCache, short tileFrameX, short tileFrameY, Color tileLight, bool visible)
```
Chicken-Bones: Add docs to `DrawEffects` to tell modders to use this method instead to emit particles (and why)

patches/tModLoader/Terraria/WorldGen.cs.patch
JavidPack: Yeah, we have Example toilet, campfire, statue, lamp, and trap that all do things with wiring. The issue is terraria added this new code path for light toggling specifically that was never integrated into tmod. It's never come up in discord I don't think, I didn't even realize it was a feature until I looked closely at the wiki page for "The Constant" seed.

It's definitely not something highly requested, I just noticed it and made a note if it ever is asked about later.

Chicken-Bones: Yes, but toggling torches with wiring was a thing long before the constant seed afaik.

JavidPack: Yes, toggling works fine, it's just that it is done through the normal HitWire. The Constant adds code to force a light source off, not just toggle it or calling hitwire on the tile. It's a new feature to have an API that tells the ModTile a specific state is being forced. 

Looking at the code again, we actually handle torch HitWire code in tModLoader not in the ModTile. We can easily replace `case 4` with `case int _ when TileID.Sets.Torch[type]:` in `TryToggleLight` to support it for torches automatically. The issue is other light sources that don't have sets or guaranteed tile layouts, like ExampleLamp and ExampleCampfire. These would need a new hook.

JavidPack: Looks like Torch God also doesn't support unlighting modded torches.

Chicken-Bones: Good to know torches have guaranteed sprite layout so wiring support is in tML. We should add torch god and constant support then since it seems easy enough

ExampleMod/Content/Tiles/ExampleCampfire.cs
```diff
@@ -130,27 +130,23 @@ public override void AnimateIndividualTile(int type, int i, int j, ref int frame
...
-					dust.velocity *= 0.2f;
-					dust.velocity.Y -= 0.5f + Main.rand.Next(10) * 0.1f;
-					dust.fadeIn = 0.5f + Main.rand.Next(10) * 0.1f;
-				}
+		public override void EmitParticles(int i, int j, Tile tileCache, short tileFrameX, short tileFrameY, Color tileLight, bool visible) {
+			// Unlike a typical tile, campfire tiles intentionally still spawn dust even when the tile is invisible. This means we do NOT check visible as other examples do.
+
+			Tile tile = Main.tile[i, j];
+			// Only emit dust from the top tiles, and only if toggled on. This logic limits dust spawning under different conditions.
+			if (tile.TileFrameY == 0 && Main.rand.NextBool(3) && ((Main.drawToScreen && Main.rand.NextBool(4)) || !Main.drawToScreen)) {
```
Chicken-Bones: This condition could be simplified to `!Main.drawToScreen || Main.rand.NextBool(4)` but... shouldn't we take care of making sure that particles are generated at the same rate in all lighting modes by only calling `EmitParticles` every 4 frames when `drawToScreen` is true?

patches/tModLoader/Terraria/ModLoader/ModTile.cs
```diff
@@ -417,23 +417,18 @@ public virtual void DrawEffects(int i, int j, SpriteBatch spriteBatch, ref TileD
...
-	/// <param name="j"></param>
-	/// <param name="tileCache"></param>
-	/// <param name="tileFrameX"></param>
-	/// <param name="tileFrameY"></param>
-	/// <param name="tileLight"></param>
-	/// <param name="visible"></param>
-	/// <returns></returns>
-	public virtual bool DrawTilesEmitParticles(int i, int j, Tile tileCache, short tileFrameX, short tileFrameY, Color tileLight, bool visible)
+	/// <para/> Note that this is called even if the tile is invisible due to echo coating, so check <paramref name="visible"/> if dust should only be spawned if the tile is visible. Tiles that still spawn particle effects while invisible can be useful to builders. Some tiles that spawn dust even when invisible include BubbleMachine, FogMachine, BrazierSuspended, Campfire, Chimney, SillyBalloonMachine, LeafBlock, and PoopBlock.
+	/// <para/> The <paramref name="tileFrameX"/> and <paramref name="tileFrameY"/> values differ from the Tile frame values in that they incorporate the changes from <see cref="SetDrawPositions"/> and likely should be used instead of <see cref="Tile.TileFrameX"/> and Y directly.
```
Chicken-Bones: `likely should` -> `should normally`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ModTile.cs
Chicken-Bones: @JavidPack reminder for this one

ExampleMod/Content/Tiles/ExampleCampfire.cs
JavidPack: Vanilla bug, logged. Fixed in ExampleCampfire now.

PR #4553, opened 2025-02-25
Localize custom player death messages (PlayerDeathReason.ByCustomReason)

### LINKED ISSUE #4224: [Feature Request] Localized PlayerDeathReason

No

Mod capability as a Modder

### Description

`PlayerDeathReason.ByCustomReason` only supports literal values meaning that the message will be displayed as is for all players on the server regardless of their locale. This proposal adds the following:
An new `ByCustomReason` with `params object[] substitutions`, which will then get stored in a new property.
```cs
//Added by TML.
public object[] Substitutions { get; set; } = System.Array.Empty<object>();

// Added by TML.
public static PlayerDeathReason ByCustomReason(string key, params object[] substitutions)
{
	return new PlayerDeathReason {
		_sourceCustomReason = key,
		Substitutions = substitutions
	};
}
```

`GetDeathText` will have to be changed accordingly:
```cs
if (_sourceCustomReason != null) {
	if (Language.Exists(_sourceCustomReason))
		return NetworkText.FromKey(_sourceCustomReason, Substitutions);

	return NetworkText.FromLiteral(_sourceCustomReason);
}
```

`WriteSelfTo` and `FromReader` need to accomodate the new property. Serializing it will likely require publicizing `NetworkText.ConvertSubstitutionsToNetworkText`

Make sure the text is localized for other clients to their locale.

The new approach should be backwards compatible with no degradation for mods compiled before this change, but new mods should default to this new behavior (without substitutions), so it's likely theres some additional tweaks to be made to the methods for this to be accomplished.

_No response_

JavidPack:
Fixes #4224 

Currently, `PlayerDeathReason.ByCustomReason` takes a `string` and that `string` is synced across the network to other clients. Those clients will see the custom death reason in the language of the player who died, not in their chosen language.

This PR adds additional `PlayerDeathReason.ByCustomReason` overloads that support localization and marks the existing overload as `Obsolete`.

This image shows the fixed behavior. One player is on Chinese the other English. Each sees the message in their own language:
image

Without the fix each player would either see the message in English or in Chinese, depending on which player died.

# Discussion
1. ~~We could get rid of the `ByCustomReason(LocalizedText reason)` overload if using `LocalizedText.WithFormatArgs` like this is not advised. The docs for `LocalizedText.WithFormatArgs` suggest that it is not a correct approach.~~

2. ~~Should `ByCustomReason(string reasonInEnglish)` be marked as `Obsolete` to encourage usage of the new overloads?~~

# Porting Notes
- If already using `LocalizedText`, change `localizedText.Format(substitutions)` to `localizedText.ToNetworkText(substitutions)`.
```diff
-player.KillMe(PlayerDeathReason.ByCustomReason(DeathMessage.Format(player.name)), 9999, 0);
+player.KillMe(PlayerDeathReason.ByCustomReason(DeathMessage.ToNetworkText(player.name)), 9999, 0);
```
- If still using `string` directly, consider supporting custom death message localization by following the example shown in `ExampleOnBuyItem.SetStaticDefaults` and `OnCreated`.

### DIFF
--- a/patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
+++ b/patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
@@ -1,6 +1,7 @@
 --- src/TerrariaNetCore/Terraria/DataStructures/PlayerDeathReason.cs
 +++ src/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs
-@@ -1,5 +_,7 @@
+@@ -1,5 +_,8 @@
++using System;
  using System.IO;
  using Terraria.Localization;
 +using Terraria.ModLoader;
@@ -24,13 +25,13 @@
 +	private int _sourceItemType => _sourceItem?.type ?? 0;
 +	private int _sourceItemPrefix => _sourceItem?.prefix ?? 0;
  	private string _sourceCustomReason;
-+	private NetworkText _sourceCustomReasonLocalized;
++	private NetworkText _customReason;
  
 +	/*
  	public int? SourceProjectileType {
  		get {
  			if (_sourceProjectileLocalIndex == -1)
-@@ -22,7 +_,25 @@
+@@ -22,7 +_,26 @@
  			return _sourceProjectileType;
  		}
  	}
@@ -43,8 +44,9 @@
 +	public ref int SourceOtherIndex => ref _sourceOtherIndex;
 +	public ref int SourceProjectileType => ref _sourceProjectileType;
 +	public ref Item SourceItem => ref _sourceItem;
++	[Obsolete("CustomReason should be used instead")]
 +	public ref string SourceCustomReason => ref _sourceCustomReason;
-+	public ref NetworkText SourceCustomReasonLocalized => ref _sourceCustomReasonLocalized;
++	public ref NetworkText CustomReason => ref _customReason;
 +
 +	/// <summary>
 +	/// Only safe for use when the local player is the one taking damage! <br/>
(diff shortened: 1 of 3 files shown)


patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -32,6 +32,8 @@
 -		}
 -	}
 -
```
Chicken-Bones: Not your patch from this change, but while you're here, change to block comment instead of removal?

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -63,10 +66,40 @@
...
-@@ -72,18 +_,26 @@
+@@ -65,6 +_,10 @@
  		};
  	}
  
++	/// <summary>
++	/// A custom death message.
++	/// <para/> This overload does not support localization, all players will see the same text regardless of their language settings. Use <see cref="ByCustomReason(LocalizedText)"/> or <see cref="ByCustomReason(NetworkText)"/> instead for localization support.
++	/// </summary>
+ 	public static PlayerDeathReason ByCustomReason(string reasonInEnglish)
```
Chicken-Bones, line 82: Yes, `Obsolete`

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -63,10 +66,40 @@
...
+ 	{
+ 		return new PlayerDeathReason {
+@@ -72,18 +_,45 @@
+ 		};
+ 	}
+ 
++	/// <inheritdoc cref="ByCustomReason(NetworkText)"/>
++	public static PlayerDeathReason ByCustomReason(LocalizedText reason)
```
Chicken-Bones: Yes, remove

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -63,10 +66,40 @@
...
++	{
++		return new PlayerDeathReason {
++			_sourceCustomReasonLocalized = reason.ToNetworkText()
++		};
++	}
++
++	/// <summary>
++	/// A localized custom death message. Each client will see the death message in their own language, unlike with <see cref="ByCustomReason(string)"/>.
++	/// <para/> Text substitutions can be provided by using <see cref="LocalizedText.ToNetworkText(object[])"/> or <see cref="LocalizedText.WithFormatArgs(object[])"/>
```
Chicken-Bones: Don't mention `WithFormatArgs` here. Standard ways of creating `NetworkText` are all good

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -40,6 +42,7 @@
 +	public ref int SourceProjectileType => ref _sourceProjectileType;
 +	public ref Item SourceItem => ref _sourceItem;
 +	public ref string SourceCustomReason => ref _sourceCustomReason;
```
Chicken-Bones, line 48: Also obsolete this one.

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
```diff
@@ -40,6 +42,7 @@
...
++	public ref NetworkText SourceCustomReasonLocalized => ref _sourceCustomReasonLocalized;
```
Chicken-Bones: unfortunate name clash with previous member. Perhaps we should call it `CustomReason`

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
Chicken-Bones: Ideally, we could convert it to a wrapper property which does `NetworkText.FromLiteral` and `.ToString()` on `set/get` but... up to you whether we want to introduce runtime breakage.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/DataStructures/PlayerDeathReason.cs.patch
JavidPack: I think I'll avoid breakage, the obsolete should be all that is needed for modders to migrate to the new approach if that is what they are working with.

Chicken-Bones (approved):

PR #4564, opened 2025-03-02
Added MusicID.Sets.VolumeMult[]

flye-name:
### What is the new feature?
Continuation of https://github.com/tModLoader/tModLoader/pull/4530
Adds `MusicID.Sets.VolumeMult[]` which modifies the volume of specified music tracks.
Clamped between 0 and 2 as a reasonable value, but if a track has `SkipsVolumeRemap[]` set to false, `VolumeMult[]` gets clamped between 0 and 1 because:
bild

Quite a few people reported that the recent `SkipsVolumeRemap[]` addition made certain tracks too loud, so this feature gives modders the ability to selectively adjust the volume of music.

This also consequently allows modders to adjust the volume of music in real-time without tampering with the volume slider or `Main.musicFade`.

None that I can think of.  

The intended usage of adjusting the overall volume of a track:
```cs
public class MusicSystem : ModSystem {
    {
        MusicID.Sets.VolumeMult[MusicLoader.GetMusicSlot(Mod, "Assets/Music/MySong")] = 0.8f;
    }
}
```

But also the ability to adjust the volume of certain tracks in real time for something like a cutscene:
```cs
public override void SomeMethodThatUpdates()
{
    MusicID.Sets.VolumeMult[MusicLoader.GetMusicSlot(Mod, "Assets/Music/MySong")] = MathHelper.SmoothStep(1, 0, t);
    //It's important to reset this to its original value afterwards though, otherwise it will stay as 0 until the game reloads
}
```

### DIFF
--- a/patches/tModLoader/Terraria/Audio/LegacyAudioSystem.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/LegacyAudioSystem.cs.patch
@@ -1,11 +1,12 @@
 --- src/TerrariaNetCore/Terraria/Audio/LegacyAudioSystem.cs
 +++ src/tModLoader/Terraria/Audio/LegacyAudioSystem.cs
-@@ -2,8 +_,11 @@
+@@ -2,8 +_,12 @@
  using System.Collections;
  using System.Collections.Generic;
  using System.IO;
 +using System.Linq;
  using Microsoft.Xna.Framework.Audio;
++using Microsoft.Xna.Framework;
  using ReLogic.Content.Sources;
 +using Terraria.ID;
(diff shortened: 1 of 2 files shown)


Mirsario:
Closing as not planned, per discussion here:
https://discord.com/channels/103110554649894912/445276626352209920/1345730223458619433

> The original PR added a new possibility, to make music louder than was possible prior, without the need to introduce potential for clipping in the audio file. The ability to use files tweaked in volume for 1.3 I saw as more of a bonus on top of that. A volume multiplier doesn't add any new possibilities, if there's a problem with volume people should just normalize their files & not use the set.

Issue #2738, opened 2022-07-20
make WorldGen.Convert more modder-friendly


ThomasThePencil:
### Do you intend to personally contribute/program this feature?
not impossible, but don't count on it

### I would like to see this change made to improve my experience with
mod capability as a modder

### Description

the proposal here is to extend `WorldGen.Convert` in some fashion to allow for additional conversion types. the most concrete idea I have is a hook in `(Mod/Global)Tile` to allow modders to define conversion rules for block types individually, like so
``` csharp
public override void ConvertTile(Tile tile, int x, int y)
{
	if (tile.TileType != TileID.StoneBlock)
		return; // do nothing if not a stone block

	if (Main.rand.NextBool(5))
	{
		Main.tile[k, l].type = TIleID.Grass;
		SquareTileFrame(k, l);
		NetMessage.SendTileSquare(-1, k, l);
	}
	else
	{
		Main.tile[k, l].type = TIleID.Dirt;
		SquareTileFrame(k, l);
		NetMessage.SendTileSquare(-1, k, l);
	}
}
```

### What does this proposal attempt to solve or improve?

while lookin' at `WorldGen.Convert` the other day for #2670, I realized that the method in question --- which is used to handle all vanilla tile conversions --- is...rigid *at best*. it provides little-to-no room for modders to add their own conversion rules, be it for their own biomes or for new tiles they're addin' to existin' biomes (e.g. evil mud). considerin' that countless mods add new biomes and a number of them make an effort to add (particularly, Clentaminator-centric) conversion support for those biomes, it seems within the realm of reason to add better tools that can facilitate such implementations

### Which (other) solutions should be considered?

I've got nothing for alternatives at the moment, save for the idea of a struct (a `List` of which would be contained within any given `ModBiome`) to define conversion rules (and even then, that gets a little iffy)
as mentioned, however, if anyone has other ideas to bring to the table, they're welcome to be presented at any time

PR #4566, opened 2025-03-02
Added Convert() hook for ModBlockTypes, Tile/WallLoader.RegisterConversion(), and other biome conversion related features

IbanPlay:
### What is the new feature?
1. Adds `ModTile/ModWall.Convert` hook to allow modded tiles to be converted via clentaminator, purification powder, hardmode worldgen or infection spreading.
2. Adds `Tile/WallLoader.RegisterConversion` methods to provide or block 'global' conversions for tiles.
3. Adds `WorldGen.ConvertTile/Wall()` and `SpreadInfectionToNearbyTile` helper methods to cut down on repeated code
4. Adds `BiomeConversionID.PurificationPowder`
5. Adds `ModBiomeConversion` to register custom conversions.

#2738 has been an open issue for 2 and a half years. Right now, despite biome conversion being an important game mechanic, there is simply no easy way for modders to make modded tiles have biome variants that properly get converted, or to add biome converted versions of vanilla tiles that lack them. The conversion code is spread between `WorldGen.Convert()`, `Projectile.AI()` for purification powder, and `WorldGen.GERunner()` for the hardmode biome strip conversion.

It is now possible to add corruption/crimson/hallow variants of vanilla tiles that lack them, for example evil mud. Modders can also make new modded tiles with converted variants, like adding new plants to the forest that properly get converted into corruption/crimson/hallow. This also allows for more complex conversion rules than vanilla's Conversion ID sets, like making stone turn into "deep ebonstone" instead of ebonstone in the cavern layer.

An earlier design involved adding `Convert` to `GlobalBlockType`, but there were performance concerns about calling a large number of hooks on tiles that don't need any conversion. `RegisterConversion` takes both a tile type and conversion type in order to minimise the amount of code that needs to run. Performance testing indicates that the time taken to do hardmode worldgen is unchanged, and clentaminator does not produce lag.

`ModBiomeConversion` doesn't contain any hooks (other than `PostSetupContent`), and could just be an id, but it provides a central location to register custom conversions for modded biomes/solutions, and lets modders use the `ModContent.GetInstance` patterns they're familiar with. Hooks or properties may be added in the future to support queries about the conversion type.

```cs
public class CorruptFossilTile : ModTile
{
		TileLoader.RegisterConversion(TileID.DesertFossil, BiomeConversionID.Corruption, ConvertToCorruptFossil);
	}

	// convert a regular desert fossil into this tile (only registered to BiomeConversionID.Corruption)
	public bool ConvertToCorruptFossil(int i, int j, int type, int conversionType) {
		WorldGen.ConvertTile(i, j, Type);
		return false;
	}

	// convert this tile into a different type of fossil based on the conversionType
	public override void Convert(int i, int j, int conversionType) {
		switch (conversionType) {
			case BiomeConversionID.Purity:
			case BiomeConversionID.Sand:
			case BiomeConversionID.PurificationPowder:
				WorldGen.ConvertTile(i, j, TileID.DesertFossil);
				return;
			case BiomeConversionID.Hallow:
				WorldGen.ConvertTile(i, j, ModContent.TileType<HallowedFossilTile>());
				return;
			case BiomeConversionID.Crimson:
				WorldGen.ConvertTile(i, j, ModContent.TileType<CrimsonFossilTile>());
				return;
		}
	}
}
```

### ExampleMod Changes
- `ExampleSolution` has been updated with Terraformer support, and demonstrates making a new conversion type for a custom biome including converting trees.
- `ExampleHellSolution` implements a hell biome conversion using vanilla tiles
- `Hallowed/Corrupt/CrimsonFossilTile` in `ExampleVanillaConversionTiles.cs` shows how to add new biome variants of existing tiles that respond to the normal vanilla conversions, along with infection spreading
- `Hallowed/Corrupt/CrimsonFossilWall` in `ExampleVanillaConversionWalls.cs` showcases the same for walls


patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1495,6 +1495,103 @@
  		if (aiStyle == 1) {
  			AI_001();
  		}
+@@ -17290,53 +_,55 @@
+ 							continue;
+ 
+ 						if (type == 10) {
++							if (TileLoader.Convert(num96, num97, 0, true)) {
```
Chicken-Bones: Use a `goto` or put the condition here in the `if (type == 10)` check to avoid changing the indentation of the entire block

patches/tModLoader/Terraria/ModLoader/GlobalBlockType.cs
```diff
@@ -130,4 +130,21 @@ public virtual void PlaceInWorld(int i, int j, int type, Item item)
...
+	/// Allows you to change what happens when this tile/wall is converted into another biome. Return false to stop the game from converting the tile normally. Returns true by default.
+	/// <para/> Purification powder doesn't convert hallowed tiles back into purity, so hallowed tiles should check <paramref name="fromPurificationPowder"/> to mimic vanilla behavior.
+	/// <para/> You can use <see cref="WorldGen.ConvertTile(int, int, int, bool)"/> or <see cref="WorldGen.ConvertWall(int, int, ushort)"/> to automatically handle tile framing and multiplayer syncing.
+	/// </summary>
+	/// <param name="i"></param>
+	/// <param name="j"></param>
+	/// <param name="type"></param>
+	/// <param name="conversionType">The <see cref="ID.BiomeConversionID"/> of the conversion</param>
+	/// <param name="fromPurificationPowder">If this conversion was caused by purification powder.<br/>
+	/// Walls are not affected by purification powder, so this hook isn't called on ModWalls.</param>
```
Chicken-Bones: `hook isn't called` -> `will always be false`

Chicken-Bones (changes requested):
Please also include some Example Mod content. Both a tile/wall that gets converted, and something which uses the `Convert` methods

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -0,0 +1,255 @@
...
+			return false;
+		}
+
+		//This code is called when the game attempts to convert our hallowed tile into a new biome
+		public override bool Convert(int i, int j, int conversionType) {
+			//Purification powder doesn't convert hallow tiles back into pure versions
+			if (conversionType == BiomeConversionID.PurificationPowder)
+				return false;
+			//Yellow (desert) solution also converts evil/hallowed tiles back into purity, so don't forget that check!
+			else if(conversionType == BiomeConversionID.Purity || conversionType == BiomeConversionID.Sand) {
```
Chicken-Bones: `if(` -> `if (` here and other places in your PR

Chicken-Bones: No need to use `else` statements when using `return` but consider using a `switch` statement

patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
```diff
@@ -170,4 +170,18 @@ public virtual void PlaceInWorld(int i, int j, Item item)
...
+	/// <summary>
+	/// Allows you to change what happens when this tile/wall is converted into another biome. Return false to stop the game from converting the tile normally. Returns true by default.
+	/// <para/> Purification powder uses a separate conversionType, as it doesn't convert hallowed tiles back to purity tiles. Be sure to check for <see cref="BiomeConversionID.PurificationPowder"/> as well as <see cref="BiomeConversionID.Purity"/> when handling corruption/crimson tiles.
+	/// <para/> You can use <see cref="WorldGen.ConvertTile(int, int, int, bool)"/> or <see cref="WorldGen.ConvertWall(int, int, ushort)"/> to automatically handle tile framing and multiplayer syncing.
+	/// </summary>
+	/// <param name="i"></param>
+	/// <param name="j"></param>
+	/// <param name="conversionType">The <see cref="ID.BiomeConversionID"/> of the conversion</param>
+	/// <returns></returns>
+	public virtual bool Convert(int i, int j, int conversionType)
```
Chicken-Bones: `from converting the tile normally`. Modded tiles don't normally have any conversions defined though?

I'm not really sure under what cases you would want to have a specific return value. Needs more thought as to whether one is needed at all.

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -0,0 +1,255 @@
...
+			//else if (conversionType == BiomeConversionID.Snow) {
+			//	WorldGen.ConvertTile(i, j, TileID.Slush);
+			//}
+
+			return true;
+		}
+
+		public override void RandomUpdate(int i, int j) {
+			//Account for journey mode disabling infections and plantera slowing down biome spread rate
+			if (Main.hardMode && (NPC.downedPlantBoss && WorldGen.genRand.NextBool(2)) || !WorldGen.AllowedToSpreadInfections)
```
Chicken-Bones: Might be worth splitting this into two separate guard statements to make it clearer. Also I figure there's no need for the `hardMode` check, since you can't defeat plantera otherwise

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -0,0 +1,255 @@
...
+			}
+			else if (conversionType == BiomeConversionID.Hallow) {
+				WorldGen.ConvertTile(i, j, ModContent.TileType<HallowedFossilTile>());
+				return false;
+			}
+
+			return true;
+		}
+
+		public override void RandomUpdate(int i, int j) {
```
Chicken-Bones, line 159: Might be worth making a helper method in tML for this kind of thing? That or a common subclass. @JavidPack to weigh in.

patches/tModLoader/Terraria/ID/BiomeConversionID.cs.patch
```diff
@@ -0,0 +1,9 @@
+--- src/TerrariaNetCore/Terraria/ID/BiomeConversionID.cs
++++ src/tModLoader/Terraria/ID/BiomeConversionID.cs
+@@ -10,4 +_,6 @@
+ 	public const int Sand = 5;
+ 	public const int Snow = 6;
+ 	public const int Dirt = 7;
++	public const int PurificationPowder = 8;
```
Chicken-Bones: Add a `// Added by TML`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -190,9 +192,11 @@ internal static void ResizeArrays(bool unloading = false)
 		Array.Resize(ref GameContent.Metadata.TileMaterials.MaterialsByTileId, nextTile);
 		Array.Resize(ref HouseUtils.BlacklistedTiles, nextTile);
 		Array.Resize(ref HouseUtils.BeelistedTiles, nextTile);
+		Array.Resize(ref tileConversionDelegates, nextTile);
 
 		for (int i = 0; i < nextTile; i++) { //oh dear
 			Array.Resize(ref Main.tileMerge[i], nextTile);
+			Array.Resize(ref tileConversionDelegates[i], BiomeConversionID.Count);
```
Chicken-Bones: I wouldn't bother pre-populating this, may as well save a bit of memory.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -204,6 +208,7 @@ internal static void ResizeArrays(bool unloading = false)
 			TileObjectData._data.Add(null);
 		}
 
+
```
Chicken-Bones, line 212: Accidental patch

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -686,6 +691,36 @@ public static void ModifyLight(int i, int j, int type, ref float r, ref float g,
 		}
 	}
 
+
+	/// <summary>
+	/// Registers a tile type as having custom biome conversion code for this specific <see cref="BiomeConversionID"/>. For modded tiles, you can directly use <see cref="Convert(int, int, int, bool)"/>
+	/// </summary>
+	public static void RegisterConversion(int tileType, int conversionType, ConvertTile conversionDelegate)
+	{
+		if (tileConversionDelegates[tileType][conversionType] == null)
```
Chicken-Bones:
```cs
var conversions = tileConversionDelegates[tileType] ??= new[BiomeConversionID.Count];
var list = conversions[conversionType] ??= new();
```

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -686,6 +691,36 @@ public static void ModifyLight(int i, int j, int type, ref float r, ref float g,
...
+
+	public static bool Convert(int i, int j, int conversionType)
+	{
+		int type = Main.tile[i, j].type;
+		ModTile modTile = GetTile(type);
+		if (modTile != null && !modTile.Convert(i, j, conversionType)) {
+			return false;
+		}
+
+		if (tileConversionDelegates[type][conversionType] != null) {
```
Chicken-Bones: retrieve to local `var list`

patches/tModLoader/Terraria/ModLoader/WallLoader.cs
```diff
@@ -7,6 +7,8 @@
 using Terraria.GameContent;
 using Terraria.ID;
 using Terraria.ModLoader.Core;
+using static Terraria.ModLoader.TileLoader;
+using static Terraria.WorldBuilding.Actions;
```
Chicken-Bones: Avoid if not necessary please.

patches/tModLoader/removed_files.list
```diff
@@ -1,14 +0,0 @@
...
-ReLogic/Content/Sources/XnaContentSource.cs
-Terraria/DataStructures/AEntitySource_OnHit.cs
-Terraria/DataStructures/EntitySource_ByItemSourceId.cs
-Terraria/DataStructures/EntitySource_ByProjectileSourceId.cs
-Terraria/DataStructures/EntitySource_OnHit_ByItemSourceID.cs
-Terraria/DataStructures/EntitySource_OnHit_ByProjectileSourceID.cs
-Terraria/Libraries/Common/Ionic.Zip.CF.dll
-Terraria/Libraries/FNA/FNA.dll
-Terraria/Libraries/Mono/Steamworks.NET.dll
-Terraria/Libraries/Mono/System.Windows.Forms.dll
```
Chicken-Bones: accidentally deleted this file somehow?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
IbanPlay: That's a good point. You could make a case for some niche scenario where a modder puts a modded tile in TileID.Sets.Conversion for crossmod conversion convenience, while also wanting to run custom code for some specific conversions (Depth checks or genrand calls or whatever). It is a pretty far fetched example though, so I don't really know how much actual practicality it has, yeah

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
IbanPlay: The hardmode check was intended to be an OR check, and using !hardmode, since spreading tiles don't spread until hardmode (besides grass), but then get reduced in speed post plant. Agreed on separating this part from the journey mode toggle check though

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -38,66 +41,72 @@ public bool ConvertToHallow(int i, int j, int type, int conversionType) {
...
 			//Check a random nearby tile and try to convert it into hallow
 			int testX = i + WorldGen.genRand.Next(-3, 4);
 			int testY = j + WorldGen.genRand.Next(-3, 4);
 			if (!WorldGen.InWorld(testX, testY, 10))
-				continue;
+				return;
 
 			//Chlorophyte prevents and repels biome spread, so we should account for that
 			if (WorldGen.nearbyChlorophyte(testX, testY)) {
 				WorldGen.ChlorophyteDefense(testX, testY);
```
Chicken-Bones: Recommend adding a `return` after `ChlorophyteDefense`. Also please add docs to `ChlorophyteDefense`

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -38,66 +41,72 @@ public bool ConvertToHallow(int i, int j, int type, int conversionType) {
...
 			}
-			else {
-				//Sunflowers prevent spread
-				if (WorldGen.CountNearBlocksTypes(testX, testY, 2, 1, TileID.Sunflower) > 0)
-					continue;
-
+			//Sunflowers prevent spread
+			else if (WorldGen.CountNearBlocksTypes(testX, testY, 2, 1, TileID.Sunflower) <= 0) {
```
Chicken-Bones: Recommend converting to guard statement, makes it clearer.

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
```diff
@@ -38,66 +41,72 @@ public bool ConvertToHallow(int i, int j, int type, int conversionType) {
...
 				WorldGen.Convert(testX, testY, BiomeConversionID.Hallow, 1);
 			}
 		}
+
+		public override void ModifyFrameMerge(int i, int j, ref int up, ref int down, ref int left, ref int right, ref int upLeft, ref int upRight, ref int downLeft, ref int downRight) {
+			//We use this method to set the merge values of the adjacent tiles to -2 if the tile nearby is a pearlsandstone block
+			//-2 is what terraria uses to designate the tiles that will merge with ours using the custom frames
+			WorldGen.TileMergeAttempt(-2, TileID.HallowSandstone, ref up, ref down, ref left, ref right, ref upLeft, ref upRight, ref downLeft, ref downRight);
```
Chicken-Bones, line 72: @JavidPack is this correct?

patches/tModLoader/Terraria/ModLoader/ModBlockType.cs
Chicken-Bones: Yep, lets change it to a `void` return, and move it after the global hooks. That way the mod tile `Convert` implementation is the 'vanilla' implementation for this tile, and globals can cancel it just like they can for vanilla tiles.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -711,10 +710,14 @@ public static bool Convert(int i, int j, int conversionType)
 			return false;
 		}
 
-		if (tileConversionDelegates[type][conversionType] != null) {
-			foreach (var hook in CollectionsMarshal.AsSpan(tileConversionDelegates[type][conversionType])) {
-				if (!hook(i, j, type, conversionType)) {
-					return false;
+		var conversions = tileConversionDelegates[type];
```
Chicken-Bones: Turns out you can do `var list = tileConversionDelegates[type]?[conversionType]`

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1495,6 +1495,24 @@
  		if (aiStyle == 1) {
  			AI_001();
  		}
+@@ -17290,6 +_,8 @@
+ 							continue;
+ 
+ 						if (type == 10) {
++							if (TileLoader.Convert(num96, num97, BiomeConversionID.PurificationPowder))
```
Chicken-Bones: Is this correctly placed? What about the `if (type == 11 || type == 463)` below?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1933,68 +1933,261 @@
...
++
++				bool convertWall = WallLoader.Convert(k, l, conversionType);
++				bool convertTile = TileLoader.Convert(k, l, conversionType);
++
  				switch (conversionType) {
  					case 4:
 -						if (type <= TileID.Count && wall <= WallID.Count) {
 +						if (type <= TileLoader.TileCount && wall <= WallLoader.WallCount) {
++							if (!convertWall)
++								goto SkipWallCrimsonConversion;
```
Chicken-Bones, line 1952: Add a blank line after an indented block

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1933,68 +1933,261 @@
...
  							if (WallID.Sets.Conversion.Grass[wall] && wall != 81) {
  								tile.wall = 81;
  								SquareWallFrame(k, l);
-@@ -39444,7 +_,7 @@
+@@ -39395,6 +_,9 @@
+ 								SquareWallFrame(k, l);
+ 								NetMessage.SendTileSquare(-1, k, l);
+ 							}
++							SkipWallCrimsonConversion:
```
Chicken-Bones, line 1961: Add a blank line before this label

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1933,68 +1933,261 @@
...
++
++	/// <summary>
++	/// Converts the single tile at the given coordinate into a specified new tile type<br/>
++	/// Automatically handles tile framing and multiplayer syncing.
++	/// </summary>
++	/// <param name="i">The X coordinate of the target tile.</param>
++	/// <param name="j">The Y coordinate of the target tile.</param>
++	/// <param name="newType">The new type to convert the tile into</param>
++	/// <param name="tryBreakTrees">Should the conversion try to break trees above the converted tile if the new type is invalid for the tree</param>
++	public static void ConvertTile(int i, int j, int newType, bool tryBreakTrees = false)
```
Chicken-Bones: Is this method recommended for use with custom clentaminator likes, or only in `RandomUpdate`?

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2690,6 +2883,17 @@
  		return false;
  	}
  
+@@ -50401,7 +_,9 @@
+ 		if (!Main.remixWorld)
+ 			return;
+ 
++		if (TileLoader.Convert(x, y, 0)) {
```
Chicken-Bones: Whole method is an `else if` chain so you could convert this into a guard statement for cleaner code and to avoid patching the next line

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2711,6 +2915,41 @@
  			int num = genRand.Next(4);
  			int num2 = 0;
  			int num3 = 0;
+@@ -50715,7 +_,11 @@
+ 					if (CountNearBlocksTypes(num11, num12, 2, 1, 27) > 0)
+ 						continue;
+ 
++					if (!TileLoader.Convert(num11, num12, 1)) {
```
Chicken-Bones: As above, guard statement with `continue`

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2711,6 +2915,41 @@
...
++						if (genRand.Next(2) == 0)
++							flag2 = true;
```
Chicken-Bones, line 2931: Hmm, the return value from `TileLoader.Convert` doesn't actually tell you if anything was converted. Perhaps we could check if `tile.type != preHookType`

patches/tModLoader/removed_files.list
Chicken-Bones: Please revert newline change, not sure why it's happening

Chicken-Bones (changes requested):
Good work on the patches!

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
IbanPlay: I wrote the PR that added modifyframemerge, the comment here is actually just ripped from examplecustomframingtile, but with sandstone instead of snow

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2915,38 +2952,69 @@
...
  			int num3 = 0;
-@@ -50715,7 +_,11 @@
+@@ -50715,6 +_,21 @@
  					if (CountNearBlocksTypes(num11, num12, 2, 1, 27) > 0)
  						continue;
  
++					int preHookType = Main.tile[num11, num12].type;
 +					if (!TileLoader.Convert(num11, num12, 1)) {
 +						if (genRand.Next(2) == 0)
 +							flag2 = true;
```
Chicken-Bones: I think this should only be when the type changes. If the conversion just gets blocked, no need to try again elsewhere.

ExampleMod/Content/Tiles/ExampleVanillaConversionTiles.cs
JavidPack: We're talking about RandomUpdate code right? It's a tough question, maybe a shared static method would be easier for some modders to understand rather than inheritance of a base class. The logic seems like something we can add directly to tmod, and maybe we do want to be able to add a hook in it for allowing additional biome spread limiters that will be honored by modded tiles spreading as well.

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+	public class ExampleSolution : ModItem
+	{
+		public override void SetStaticDefaults() {
+			Item.ResearchUnlockCount = 99;
+			ItemID.Sets.SortingPriorityTerraforming[Type] = 101; //One past dirt soulution
+		}
+
+		public override void SetDefaults() {
+			Item.DefaultToSolution(ModContent.ProjectileType<ExampleSolutionProjectile>());
+			return;
```
Chicken-Bones: No need for return

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+			itemGroup = ContentSamples.CreativeHelper.ItemGroup.Solutions;
+		}
+	}
+
+	public class ExampleSolutionProjectile : ModProjectile {
+
+		public ref float Progress => ref Projectile.ai[0];
+		// Solutions shot by the terraformer get an increase in conversion area size, indicated by the second AI parameter being set to 1
+		public bool ShotFromTerraformer => Projectile.ai[1] == 1f;
+		public static int ConversionType;
```
Chicken-Bones: Newline above this field

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+				sprayDust.noGravity = true;
+				sprayDust.scale *= 1.75f * dustScale;
+			}
+
+			Progress++;
+			Projectile.rotation += 0.3f * Projectile.direction;
+		}
+	}
+
+	public class ExampleSolutionConversion : ModBiomeConversion {
```
Chicken-Bones: braces on newline for classes please https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+		public override void SetStaticDefaults() {
+
+			//Go over every tile and add a conversion to it for our conversion type if they're part of the list of usual conversion tiles
```
Chicken-Bones: ` ` after `//`

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		public bool HellifyTile(int i, int j, int type, int conversionType) {
+
+			//Turn tiles into hell-appropriate versions
+			if (type == TileID.ClayBlock || TileID.Sets.Conversion.Stone[type] || TileID.Sets.Conversion.Dirt[type] || TileID.Sets.Conversion.Sand[type]) {
+				WorldGen.ConvertTile(i, j, TileID.Ash, true);
+				return false;
+			}
+			if (TileID.Sets.Conversion.Grass[type] || TileID.Sets.Conversion.GolfGrass[type]) {
```
Chicken-Bones: Blank line after closing braces, here and a few other places

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+					// Necessary because the "IsATreeTrunk" ID set doesn't care about the tile's frame and returns true even if the tile isnt the tree's "trunk"
+					int treeFrameX = Main.tile[treeCenterX, treeTop].TileFrameX / 22;
+					int treeFrameY = Main.tile[treeCenterX, treeTop].TileFrameY / 22;
+					bool isTreeTrunk = (treeFrameX != 1 && treeFrameX != 2) || treeFrameY < 6;
+					
+					// If the tile above wasn't a tree trunk, check the adjacent two tiles to find it
+					if (!isTreeTrunk) {
+						for (int x = treeCenterX - 1; x < treeCenterX + 2; x += 2) {
+
+							if (!Main.tile[x, treeTop].HasTile || !TileID.Sets.IsATreeTrunk[Main.tile[x, treeTop].TileType])
```
Chicken-Bones: Please extract repeated `tile[x, y]` accesses for the same coordinate to a variable

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+				}
+
+				WorldGen.ConvertTile(i, j, TileID.AshPlants);
+				return false;
+			}
+			if (type == TileID.Plants2) {
+				WorldGen.KillTile(i, j);
+				if (Main.netMode != 0)
+					NetMessage.SendTileSquare(-1, i, j);
+				return false;
```
Chicken-Bones: blank line after indented statements

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -686,6 +691,37 @@ public static void ModifyLight(int i, int j, int type, ref float r, ref float g,
...
 
+	/// <summary>
+	/// Registers a tile type as having custom biome conversion code for this specific <see cref="BiomeConversionID"/>. For modded tiles, you can directly use <see cref="Convert"/>
+	/// </summary>
+	/// <param name="tileType">The tile type that has is affected by this custom conversion.</param>
+	/// <param name="conversionType">The conversion type for which the tile should use custom conversion code.</param>
+	/// <param name="conversionDelegate">Code to run when the tile attempts to get converted. Return false to signal that your custom conversion took place and that vanilla code shouldn't be ran.</param>
+	public static void RegisterConversion(int tileType, int conversionType, ConvertTile conversionDelegate)
+	{
+		var conversions = tileConversionDelegates[tileType] ??= new List<ConvertTile>[BiomeConversionLoader.BiomeConversionCount];
```
Chicken-Bones, line 708: Perhaps we should make sure that `RegisterConversion` can only be called after `ResizeArrays`, otherwise `BiomeConversionCount` won't be set in stone, and a mod calling `RegisterConversion` later for the same tile may throw (and not be at fault)

The easiest way is probably just to set `tileConversionDelegates = null` by default and in `Unload` and fully reallocate it in `ResizeArrays`

patches/tModLoader/Terraria/WorldGen.TML.cs
```diff
@@ -32,4 +33,78 @@ public static void DetourPass(PassLegacy pass, GenPassDetour hookDelegate)
...
+	/// <param name="i">The X coordinate of the target tile.</param>
+	/// <param name="j">The Y coordinate of the target tile.</param>
+	/// <param name="newType">The new type to convert the tile into</param>
+	/// <param name="tryBreakTrees">Should the conversion try to break trees above the converted tile if the new type is invalid for the tree</param>
+	public static void ConvertTile(int i, int j, int newType, bool tryBreakTrees = false)
+	{
+		Tile tile = Main.tile[i, j];
+		if (tile.type == (ushort)newType)
+			return;
+		if (tryBreakTrees)
```
Chicken-Bones, line 51: newlines after indented statements, as per code style

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -2711,6 +2909,81 @@
...
++					int preHookType = Main.tile[num11, num12].type;
++					if (!TileLoader.Convert(num11, num12, 1)) {
++						if (preHookType == Main.tile[num11, num12].type)
++							continue;
++
++						if (genRand.Next(2) == 0)
++							flag2 = true;
++
++						continue;
++					}
```
Chicken-Bones: `bool convertTile = TileLoader.Convert(...)`

later, after the `preHookType`

`if (!convertTile) continue;`

Also, did you want to add the `HasTile` guards here too?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/BiomeConversionLoader.cs
```diff
@@ -0,0 +1,42 @@
...
+		//Nothing for now
+	}
+
+	internal static void Unload()
+	{
+		conversions.Clear();
+		BiomeConversionCount = BiomeConversionID.Count;
+	}
+
+	internal static int ReserveID() => BiomeConversionCount++;
```
Chicken-Bones: No need for separate method, inline please

patches/tModLoader/Terraria/ModLoader/BiomeConversionLoader.cs
```diff
@@ -0,0 +1,42 @@
...
+
+	public static int Register(ModBiomeConversion conversion)
+	{
+		int type = ReserveID();
+		ModTypeLookup<ModBiomeConversion>.Register(conversion);
```
Chicken-Bones: Prefer to put this in the `ModType.Register` method as most but not all mod types do, sorry for the inconsistency

patches/tModLoader/Terraria/ModLoader/EquipLoader.cs
```diff
@@ -267,10 +268,10 @@ internal static int GetPlayerEquip(Player player, EquipType type)
 	public static int AddEquipTexture(Mod mod, string texture, EquipType type, ModItem item = null, string name = null, EquipTexture equipTexture = null)
 	{
 		if (!mod.loading)
-			throw new Exception("AddEquipTexture can only be called from Mod.Load or Mod.Autoload");
+			throw new Exception(Language.GetTextValue("tModLoader.LoadErrorLoadOnlyMethod", "AddEquipTexture"));
```
Chicken-Bones: Thanks for the localization. While we don't normally bother localizing error messages, as we assume developers can read English, it doesn't hurt here. Please apply to the other relevant places (solution search for `can only be called from Mod.Load`)

Chicken-Bones (changes requested):

ExampleMod/Content/Items/ExampleSolution.cs
```diff
@@ -0,0 +1,225 @@
...
+					WallID.Sets.Conversion.Grass[i] ||
+					WallID.Sets.Conversion.Stone[i])
+					WallLoader.RegisterConversion(i, Type, HellifyWall);
+			}
+		}
+
+		public bool HellifyTile(int i, int j, int type, int conversionType) {
+
+			//Turn tiles into hell-appropriate versions
+			if (type == TileID.ClayBlock || TileID.Sets.Conversion.Stone[type] || TileID.Sets.Conversion.Dirt[type] || TileID.Sets.Conversion.Sand[type]) {
```
Chicken-Bones: Part of the advantage of the `RegisterConversion` methods is that we can register separate methods for separate types, so there should be barely any need for any type checks in this method if we split it up!

Call the methods, `ConvertToAsh`, `HellifyGrass`, `ConvertPlants`

PR #4570, opened 2025-03-06
Made critter cage lids account for multi-column spritesheets

IbanPlay:
### What is the bug?
Modded critter cages that use `TileID.Sets.CritterCageLidStyle[]` and  more than one column (Either because they have a lot of frames, or because it was exported with animations split into columns) have the top lid break because the frame coordinates aren't wrapping properly.

### How did you fix the bug?
Made the X frame of the critter cage lid get wrapped around by the tile's width (account for 6x3 vs 3x2 critter cages)

Can't think of anything more straightforward than this. A more general system for critter cages of any width could maybe get made, but at this point just use custom drawing code.


JavidPack:
Tested, works. Thanks.

Issue #2449, opened 2022-05-29
Modify the size of the box behind item tooltips


Cyrillya:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Modders will be able to modify the offset size of the tooltip box in ModifyTooltips.
Or maybe add a Vector2 called "OffsetTextSize" or something to TooltipLine and like this in Main.MouseText_DrawItemTooltip():
![image](https://user-images.githubusercontent.com/35227653/170856084-69aa9280-f8ce-4123-bc18-3fa4a687d272.png)
Unfortunally due to some reason I can't make a pr to contribute to tModLoader :(

### What does this proposal attempt to solve or improve?

If you directly modify the position of the tooltip in PreDrawTooltipLine, this will happen:
![image](https://user-images.githubusercontent.com/35227653/170855778-dcb6d491-d9b4-44f3-84de-5c3206b93f1a.png)
Some of the text is out of the tooltip box.
It'll be better if modders can modify the size of the box.

JavidPack:
I'm pretty sure there was a hook to use for this on 1.3. We'll have to look into it.

Chicken-Bones:
It'd be nice if it could be auto-calculated properly yeah.

Jofairden:
Like jopo said, I remember there was something made for this in 1.3
If it's gone, we must go back and see what it was.
In general though, UI in this game is finicky. May as well just look into it without going back in the old code

steviegt6:
Can we start discussion on this and #3546 again? I work on a mod that modifies the scaling of various tooltip lines while drawing and it creates jarring amounts of empty space and I would like to resolve this.

I'm not too concerned about the design of the API -- I don't mind duplicating logic to determine the size of the tooltip box -- I'd just like a solution that doesn't force me to modify the body of the method responsible for rendering the box.

PR #4571, opened 2025-03-06
Tooltip Background Drawing Fix

### LINKED ISSUE #3546: Tooltip backgrounds not affected by PreDrawTooltip

Yes

Mod capability as a Modder

### Description

Currently returning false in preDrawTooltips does not cancel the vanilla behavior of drawing a box the size of the tooltip text. This makes it difficult to create custom tooltips using this hook if the user has these backgrounds enabled in the settings. See attached image.

image

The ability for modders to have control over everything about tooltip drawing, not just text.

Perhaps API to modify the dimensions of the box instead of cancel it altogether may be useful aswell, but that is likely a seperate feature proposal and would require greater scope.

MutantWafflez:
### What is the bug?
Resolves #3546 and #2449

### How did you fix the bug?

Moved the `ItemLoader.PreDrawTooltip` call in `Main` when drawing tooltips to be called before the background box is drawn. If `PreDrawTooltip` returns `false`, the box will also no longer be drawn, instead of drawing regardless before.

In addition, `PreDrawTooltip` has been updated with new parameters. `boxDrawn` communicates that only the item hover box should not be drawn. This should be used instead of returning false if only drawing a custom hover box. `textAreaSize` informs how big the text area is expected to be. To accommodate tooltip lines reserving additional space for icons, resource bars, etc., `TooltipLine.OffsetTextSize` has been added and contributes towards the `textAreaSize` calculation, allowing custom hover boxes to know the intended tooltip content size.

`ExampleDrawTooltips.cs` has been ported and updated. It showcases the tooltip drawing hooks. Without the fixes in this PR the icon line could potentially go over the edges of the box (if Tooltip0 happened to be shorter) and the vanilla box would still be drawn as well.

image

image

Finally, `PreDrawTooltipLine`'s `yOffset` parameter is properly reset to 0 now during iteration. Previously mods were required to reset this value on subsequent tooltip lines, this would be very error prone and is incompatible with multiple mods trying to do custom drawing. This is technically a behavioral change but I doubt the structure of any existing code using the parameter correctly would cause this change to have any ill effects. 

Not practically, no.

# Porting Notes

- If you use `(Mod|Global)Item.PreDrawTooltip` in your mod, run tModPorter to update the method signatures.
- Set and check `boxDrawn` as appropriate to prevent multiple boxes being drawn from multiple mods. Use `textAreaSize` in your code if useful.
- Use `TooltipLine.OffsetTextSize` in `ModifyTooltipLines` to "reserve" extra space for a specific tooltip line if custom drawing will be used to draw additional things on the line.


patches/tModLoader/Terraria/ModLoader/TooltipLine.cs
```diff
@@ -42,6 +42,11 @@ public class TooltipLine
 	/// </summary>
 	public Color? OverrideColor;
 
+	/// <summary>
+	/// Additional space allocated to this tooltip to accommodate additional drawing, such as an image or icon.
+	/// </summary>
+	public Vector2 OffsetTextSize;
```
Chicken-Bones, line 48: Should probably call this `AdditionalTextSize`

PR #4574, opened 2025-03-08
Add check for broken TML methods when verifying mod assemblies

### LINKED ISSUE #3820: Mod assembly validation assumes that every unimplemented method comes from another mod
### Description

TML's own breaking changes of abstract methods' signatures results in the following message being shown to modders upon a load error:

image
image

It'd be best if we check the base class or interface of the unimplemented method and show a separate message in case said method is owned by the TML/game assembly, suggesting using tModPorter instead of the `[ExtendsFromMod]` attribute.

CornHusker89:
### What is the bug?
fixes https://github.com/tModLoader/tModLoader/issues/3820

### How did you fix the bug?
Added a new variable to keep track of where the validation fails, then iterated through that type's methods to find a method which meets the following conditions:
- Method is abstract
- Method has a declaring type
- The type which declares the method is NOT the type that failed validation
- Assembly which first declared the method is the TML assembly

When a method meets these conditions, the validation will throw an error prompting the user to use tModPorter.

- The original caught error (which triggers the code in the pr) has no easily accessible data in it but contains the failed method and type in the message, which could be regex'ed out instead of searching for the method. 
- More minor, but this pr compares assemblies using `.ToString()`, which feels hacky. An alternative is to use `.FullName`

### Other Notes
- There is a very long if statement, could be compacted in accordance with formatting guidelines.
- Somehow reducing the nesting inside the `catch`

The mod I used to test this is linked. It reproduces this bug by having a class which inherits `BuilderToggle` without implementing the abstract method `DisplayValue()`.
[TestMod.zip](https://github.com/user-attachments/files/19146315/TestMod.zip)

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
@@ -334,16 +334,32 @@ public static class AssemblyManager
 
 	private static IDictionary<Assembly, Type[]> GetLoadableTypes(ModLoadContext mod, MetadataLoadContext mlc)
 	{
+		Type loadingType = null;
 		try {
 			return mod.Assemblies.ToDictionary(a => a, asm =>
 				mlc.LoadFromAssemblyName(asm.GetName()).GetTypes()
 					.Where(mType => IsLoadable(mod, mType))
-					.Select(mType => asm.GetType(mType.FullName, throwOnError: true, ignoreCase: false))
+					.Select(mType => {
+						loadingType = mType;
+						return asm.GetType(mType.FullName, throwOnError: true, ignoreCase: false);
+					})
 					.ToArray());
 		}
 		catch (Exception e) {
+			if (loadingType != null) {
+				foreach (MethodInfo method in loadingType?.GetMethods()) {
+					// check if it is an abstract method which is not overriden, and originates from tModLoader's assembly
+					if (method.IsAbstract && method.DeclaringType != loadingType &&	method.DeclaringType != null && Assembly.GetAssembly(method.DeclaringType)?.ToString() == Assembly.GetExecutingAssembly().ToString()) {
+						throw new Exceptions.GetLoadableTypesException(
+							"This mod seems to contain a class which inherits from TML but do not implement required methods. Use tModPorter to update required methods." + "\n\n" + $"The method \"{method.Name}\" in class \"{loadingType.FullName}\" caused this error.\n\n" +	e.Message,
+							e
+						);
+					}
+				}
+			}
+
 			throw new Exceptions.GetLoadableTypesException(
-				"This mod seems to inherit from classes in another mod. Use the [ExtendsFromMod] attribute to allow this mod to load when that mod is not enabled." + "\n\n" + (e.Data["type"] is Type type ? $"The \"{type.FullName}\" class caused this error.\n\n" : "") + e.Message,
+				"This mod seems to inherit from classes in another mod. Use the [ExtendsFromMod] attribute to allow this mod to load when that mod is not enabled." + "\n\n" + $"The \"{loadingType}\" class caused this error.\n\n" + e.Message,
 				e
 			);
 		}
--- a/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
@@ -359,7 +359,7 @@ public static class AssemblyManager
 			}
 
 			throw new Exceptions.GetLoadableTypesException(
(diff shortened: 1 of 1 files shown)


JavidPack:
Tested, seems to work. I made some changes for readability. The other custom exception has a help link that is misleading so I changed it to a normal Exception. It's interesting that comparing the assemblies directly doesn't work, but they are different objects at this point.

The test mod doesn't work anymore due to an update on it's dependency mod, but to replicate you can just make an InfoDisplay class in any mod, build it, then change DisplayValue to abstract and debug tmod.

CornHusker89:
Sorry about my testing mod not working! There was a file left over from some other testing I was doing earlier. For anyone else who'd like it, an updated version is here [TestMod.zip](https://github.com/user-attachments/files/19202474/TestMod.zip)

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -334,16 +334,32 @@ internal static bool FirstModInStackTrace(StackTrace stack, out string modName)
...
 	{
+		Type loadingType = null;
 		try {
 			return mod.Assemblies.ToDictionary(a => a, asm =>
 				mlc.LoadFromAssemblyName(asm.GetName()).GetTypes()
 					.Where(mType => IsLoadable(mod, mType))
-					.Select(mType => asm.GetType(mType.FullName, throwOnError: true, ignoreCase: false))
+					.Select(mType => {
+						loadingType = mType;
+						return asm.GetType(mType.FullName, throwOnError: true, ignoreCase: false);
```
Chicken-Bones: Instead of storing `loadingType` outside the `Select` clause, can you extract the `asm.GetType` call to a separate method `GetType(asm, mType)` which contains the `try/catch`? You can then annotate the thrown exception in the same way that `IsLoadable` does.

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -334,16 +334,32 @@ internal static bool FirstModInStackTrace(StackTrace stack, out string modName)
...
+					})
 					.ToArray());
 		}
 		catch (Exception e) {
+			if (loadingType != null) {
+				foreach (MethodInfo method in loadingType?.GetMethods()) {
+					// Check if it is an abstract method which is not overridden and originates from tModLoader's assembly
+					if (method.IsAbstract && method.DeclaringType != loadingType &&	method.DeclaringType != null && Assembly.GetAssembly(method.DeclaringType)?.FullName == Assembly.GetExecutingAssembly().FullName) {
```
Chicken-Bones: No need to restrict this check to tModLoader. May as well just report the name of the method's declaring type, though the tModPorter recommendtation would only be if the declaring type's assembly is us.

You should also be able to just use `method.DeclaringType.Assembly == Assembly.GetExecutingAssembly()`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
CornHusker89: Intrestingly, `method.DeclaringType.Assembly == Assembly.GetExecutingAssembly()` always evaluates to false. It's probably something to do with how it's 2 different objects representing the same assembly, like JavidPack said.

JavidPack: Yeah, one is `ReflectionOnly`, the other isn't. Guess it's not fully realized at the point of exception.

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -390,6 +393,24 @@ private static bool IsLoadable(ModLoadContext mod, Type type)
...
 
+	/// <summary>
+	/// Gets and validates the <see cref="Type"/> from the given <see cref="Assembly"/>.
+	/// </summary>
+	/// <param name="assembly">Assembly to load type from</param>
+	/// <param name="type">Target type to get</param>
+	/// <returns></returns>
+	#nullable enable
+	private static Type? GetType(Assembly assembly, Type type) {
+	#nullable disable
```
CornHusker89: Double check that the `nullable enable`/`disable` here is proper. The indentation especially feels off.

CornHusker89: I moved the `#nullable disable` to make a lot more sense, so I'll close this for now. It can be reopened later if needed.

Chicken-Bones:
Thankyou!

PR #4576, opened 2025-03-10
Add icon for custom mod source path. Fix UIModSourceItem/builtMod duplication

JavidPack:
This is a followup to #4548

I've noticed that ExampleMod shows up twice during normal development:

image

The 2 entries in `FindModSources` point to different folder paths. One is the normal path and the other a symbolic link path to the tModLoader project git folder setup according to the [contribution guide](https://github.com/tModLoader/tModLoader/wiki/tModLoader-guide-for-contributors). 

A similar issue happens for other mods built from a separate folder:

image

There is no way to distinguish them, and they both say they were built recently even though only one was built recently. The situation resolves itself (the extra disappears) if the mod in the ModSources folder is rebuilt in-game, but since you don't know which one is being built it could be an issue if you were trying to test a new feature in mod code contained in the non-mod sources folder copy, like a fork someone sent you to test.

# Fix
This PR does 2 things. 

First, it attempts to prevent both entries from saying they were built recently by detecting which source folder is attributed to the built local .tmod file by comparing `builtMod.properties.modSource`, if it exists. Old .tmod won't have this until they are built on a more recent tModLoader release.

Second, for `UIModSourceItem` corresponding to a path outside the normal `ModCompile.ModSourcePath`, the open folder button is customized. The icon is changed and the hover text now shows the custom path. This allows the modder to build that mod source folder if they intend to do that.

dotnet_2025-03-10_17-42-53

These fixes currently do not prevent the duplicate entries for ExampleMod specifically because of the symlink, although now only 1 entry shows the "built x minutes ago" information. The symlink is not needed anymore, actually, so the wiki has been updated as well. Users experimenting with ExampleMod through downloading this git just need to build in Visual Studio and it will always work without the symlink.


steviegt6:
Is there a point in keeping the symlinks in the first place? I had duplicates during development and opted to remove them because they were made redundant.

And specifically regarding mods other than ExampleMod, if you're depending on symlinks for resolve targets correctly, you probably shouldn't; are you opening your projects directly from the symlink path?

Chicken-Bones:
> Is there a point in keeping the symlinks in the first place? I had duplicates during development and opted to remove them because they were made redundant. If you're depending on symlinks for resolve targets correctly, you probably shouldn't; are you opening your projects directly from the symlink path?

I raised this point to jopo independently and we have decided to remove the symlink instructions for developers building from the tML git

steviegt6:
Supporting duplicate paths is OK and something I considered addressing in my initial merge request. I'm not confident it's necessary, though.

More work for little reward: it may be nicer to check for symlinks and resolve their actual path to avoid cases like these, at least?

Chicken-Bones (approved):

Issue #4072, opened 2024-02-23
Utility method for restarting SpriteBatch


BasicallyIAmFox:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

New utility method, allowing to restart spritebatch in a more friendly way.

```
using (spriteBatch.UseBegin(
            // All of those are optional parameters
            sortMode: ...,
            blendState: ...,
            samplerState: ...,
            depthStencilState: ...,
            effect: ...,
            matrix: ...
          ))
{
  Draw();
}
```

### What does this proposal attempt to solve or improve?

Currently, it can be tedious to find correct parameters in whatever case (e.g. NPC or Projectile drawing) since there's no list or anything which in the end resorts to decompiling Terraria to find correct parameters.

### Which (other) solutions should be considered?

Adding `BeginX` (e.g. `BeginNPC`, `BeginProjectile`) utility methods which would begin spritebatch with correct parameters, but that'd be more effort.

Chicken-Bones:
You've mixed two different proposals. 
1. An `IDisposable` type for `Begin/End`
2. A way to call `Begin` with the same arguments as the previous `Begin` call (and some optional overrides) (perhaps something like `BeginAgain`.

No implementation details have been provided. Is 2 possible without modifying `SpriteBatch`?

BasicallyIAmFox:
> You've mixed two different proposals.
> 
>     1. An `IDisposable` type for `Begin/End`
> 
>     2. A way to call `Begin` with the same arguments as the previous `Begin` call (and some optional overrides) (perhaps something like `BeginAgain`.
> 
> 
> No implementation details have been provided. Is 2 possible without modifying `SpriteBatch`?

Yeah? If by modifying you mean without touching `FNA.Core`, yes. Always can resort to reflection. If TML was on .NET 8+ then could use `UnsafeAccessor`

PR #4578, opened 2025-03-14
Fix ugly text rendering in various places

harperstoeckle:
`Main.GUIBarsDrawInner` sets the sampler state to `PointClamp`, which ends up leaking into text rendering in a few places. This includes
- Mouse text in `IngameFancyUI`, including the bestiary.
- Mouse text in the options menu.
- Camera mode.

I believe this issue also affects vanilla.

Here are a few screenshots showing the difference before and after this patch:
| **Before** | **After** |
| --- | --- |
| image | image |
| image | image |
| image | image |

I opted to reset the sampler state at the end of `GUIBarsDrawInner`, but it could also probably be handled by setting it at the top of the affected functions (like `MouseTextInner`).

As far as I can tell, the parameters I give to `spriteBatch.Begin` are the same as the ones set prior to every call to `GUIBarsDrawInner`, but I'm not 100% sure. I think optimally, there would be some way to save and restore the sprite batch state, a bit like the proposal in #4072.

### DIFF
--- a/patches/tModLoader/Terraria/Graphics/Capture/CaptureInterface.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Capture/CaptureInterface.cs.patch
@@ -5,7 +5,7 @@
  	private void DrawButtons(SpriteBatch sb)
  	{
 +		sb.End();
-+		sb.Begin(SpriteSortMode.Deferred, BlendState.AlphaBlend, SamplerState.PointClamp, DepthStencilState.None, RasterizerState.CullCounterClockwise, null, Main.UIScaleMatrix);
++		sb.Begin(SpriteSortMode.Deferred, BlendState.AlphaBlend, SamplerState.PointClamp, DepthStencilState.None, RasterizerState.CullCounterClockwise, null, Main.UIScaleMatrix); // #4578 vanilla fix
 +
  		new Vector2(Main.mouseX, Main.mouseY);
  		int num = 9;
@@ -16,7 +16,7 @@
  			Main.instance.MouseText(text, 0, 0);
 +
 +		sb.End();
-+		sb.Begin(SpriteSortMode.Deferred, null, null, null, null, null, Main.UIScaleMatrix);
++		sb.Begin(SpriteSortMode.Deferred, null, null, null, null, null, Main.UIScaleMatrix); // #4578 vanilla fix
  	}
  
  	private static bool GetMapCoords(int PinX, int PinY, int Goal, out Point result)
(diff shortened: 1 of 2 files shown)


JavidPack:
Note: This bug can only be replicated at non-100% UI zoom levels.

I've noticed other issues such as hover text and the cursor itself:

NVIDIA_Share_2025-03-17_14-47-39
NVIDIA_Share_2025-03-17_14-47-48
NVIDIA_Share_2025-03-17_14-48-50
NVIDIA_Share_2025-03-17_14-48-56
NVIDIA_Share_2025-03-17_14-49-53
NVIDIA_Share_2025-03-17_14-50-21

One other thing this PR changes as a result of the fix is the look of the Camera Mode buttons:

https://github.com/user-attachments/assets/546c4a11-e4de-45f5-b8d3-425c8e34ce31

487666123:
"Personal opinion, the original pixel rendering mode was actually more comfortable."

harperstoeckle:
Yeah I just added a commit that should fix the camera buttons to match how they are drawn in vanilla. The text rendering, however, is a major issue, because it makes it almost impossible to read text in `IngameFancyUI`s with the UI zoomed out. I'm currently working on a mod that uses `IngameFancyUI`, and it's nigh unusable at low zoom levels, even though the normal inventory is perfectly fine.

JavidPack:
> Yeah I just added a commit that should fix the camera buttons to match how they are drawn in vanilla. The text rendering, however, is a major issue, because it makes it almost impossible to read text in `IngameFancyUI`s with the UI zoomed out. I'm currently working on a mod that uses `IngameFancyUI`, and it's nigh unusable at low zoom levels, even though the normal inventory is perfectly fine.

Are you saying with the fix it is bad? Or are you saying the fix will make it better?

harperstoeckle:
> > Yeah I just added a commit that should fix the camera buttons to match how they are drawn in vanilla. The text rendering, however, is a major issue, because it makes it almost impossible to read text in `IngameFancyUI`s with the UI zoomed out. I'm currently working on a mod that uses `IngameFancyUI`, and it's nigh unusable at low zoom levels, even though the normal inventory is perfectly fine.
> 
> Are you saying with the fix it is bad? Or are you saying the fix will make it better?

In its original form, this PR fixed the text rendering issue, but also changed how the camera buttons were being rendered, as you pointed out. I recently added a new commit that updates the camera buttons to look the same as they do in vanilla. I.e., in its current form, this PR fixes the text rendering problem without changing how the camera buttons look.

@487666123 said that they preferred the original pixel rendering, and I wasn't sure whether they were talking about the camera buttons or the text; In my reply, I was trying to respond to both. I agree that the camera buttons look better with the vanilla rendering, so the first sentence was me explaining that I had added the new commit. The rest of my reply was me explaining why I thought the text rendering changes introduced in this PR are necessary, in case they were saying that they preferred the original vanilla text rendering.

PR #4585, opened 2025-03-25
Addition to ExampleSoul.cs

Thanatos-Dev:
- Makes the Example Soul item render at full brightness when floating in the world, matching the vanilla souls and fragments

Before:
Before
After:
After


JavidPack:
Thanks.

Issue #4583, opened 2025-03-22
Expand map layer control


Jadams505:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

I propose the following additions (in order of preference):

1.  Be able to configure the ordering of layers (similar to builder toggles)
2. Add missing vanilla layers (NPC/Boss/Player heads, death markers, golf balls, etc)
3. Add missing draw methods to `MapOverlayDrawContext` at least one that accounts for `SpriteEffects`

### What does this proposal attempt to solve or improve?

1. Currently map layers are added after all vanilla layers. This means that modded layers are drawn on top of vanilla layers. This is particularly an issue with the pings layer since in vanilla all pings are above all other icons, but with the current design they can be hidden behind modded layers.
2. Currently only 3 vanilla layers are available to "Hide" (Pylons, Pings, Spawn), but those layers actually make up a very small portion of the things that are drawn on the map. If the other icons were wrapped in layers modders would be able to toggle their visibility. This could be useful if a mod wanted to implement some sort of fog of war mechanic and hide icons that were far away.
3.  Town NPCs heads are drawn on the map using SpriteEffects depending on their direction. While they don't use `MapOverlayDrawContext` having a overload which considers SpriteEffects would be convenient for making mod layers with similar functionality.

### Which (other) solutions should be considered?

1. A less flexible solution would be to insert modded layers before the pings layers. This would more accurately reflect vanilla.
2. 
3. One alternative would be to expose some of the state in `MapOverlayDrawContext` to allow modders to do whatever they wanted with the draw information provided.

JavidPack:
These all seem useful. 

Feature 3. Adding new overloads to allow for SpriteEffects shouldn't be too hard. I do think some people mentioned other shortcomings, but I don't remember those right now, we might want to ask around. The `MapOverlayDrawContext.Draw` methods don't actually have to be used if the modder needs to do advanced things. They only serve to simplify the code, a modder can do whatever they want with normal drawing methods, they just need to do the `Main.MouseScreen` hover check themselves.

Feature 2. The remaining hover layers pre-date the map layer system. I'm worried they may be more effort to implement and maintain as a map layer than they are worth. We might be able to request that they be made into map layers in 1.4.5, since it might be easier to do that change at the source. I don't mind if you look into it, but if it will require a lot of work to keep updated for future Terraria releases we might not want to do that.

[...]

PR #4587, opened 2025-03-27
ModMapLayer positioning

Jadams505:
### What is the new feature?
Allows modders to position their `ModMapLayers` between vanilla layers. (Feature 1 of #4583)
It adds the following methods to `ModMapLayer`:

- `public virtual Position GetDefaultPosition()`
- `public virtual IEnumerable<Position> GetModdedConstraints()`

Currently map layers are added after all vanilla layers. This means that modded layers are drawn on top of vanilla layers. By being able to position modded layers it will be easier to deal with cases where icons overlap.
The positioning approach used is the same as was used for `ExtraJumps` (#3552), but there are also the designs of `PlayerDrawLayers`, and `BuilderToggles`.
Indicates that the new layer will be ordered before the vanilla pings layer.
```c#
public override Position GetDefaultPosition() {
	return new Before(IMapLayer.Pings);
}
```
Specifies that this layer will show up in ordering after `ExampleMapLayer`. (assuming they are in the same vanilla slot)
```c#
public override IEnumerable<Position> GetModdedConstraints() {
	yield return new After(ModContent.GetInstance<ExampleMapLayer>());
}
```
### ExampleMod updates
Added `ExampleItemMapLayer` which makes any `ExampleItems` that are on the ground in the world show up on the map. It showcases how to use `GetModdedConstraints()` and orders itself before `ExampleMapLayer`.

### Porting Notes
Shouldn't be any breaking changes. Use the new methods if you want.

### DIFF
--- a/ExampleMod/Common/ExampleMapLayer.cs
+++ b/ExampleMod/Common/ExampleMapLayer.cs
@@ -16,11 +16,9 @@ namespace ExampleMod.Common
 	// ModMapLayers are used to draw icons and other things over the map. Pylons and spawn/bed icons are examples of vanilla map layers. This example adds an icon over the dungeon.
 	public class ExampleMapLayer : ModMapLayer
 	{
-		public override Position GetDefaultPosition() {
-			// Here you can define where to put this layer in the vanilla map layer order.
-			// In this case we go before the pings layer since all other vanilla map layers are ordered before pings,
-			return new Before(IMapLayer.Pings);
-		}
+		// Here you can define where to put this layer in the vanilla map layer order.
+		// In this case we go before the pings layer since all other vanilla map layers are ordered before pings,
+		public override Position GetDefaultPosition() => new Before(IMapLayer.Pings);
 
 		// In the Draw method, we draw everything. Consulting vanilla examples in the source code is a good resource for properly using this Draw method.
 		public override void Draw(ref MapOverlayDrawContext context, ref string text) {
@@ -58,9 +56,7 @@ namespace ExampleMod.Common
 	// This example makes any ExampleItems that are on the ground in the world show up on the map.
 	public class ExampleItemMapLayer : ModMapLayer
 	{
-		public override Position GetDefaultPosition() {
-			return new Before(IMapLayer.Pings);
-		}
+		public override Position GetDefaultPosition() => new Before(IMapLayer.Pings);
 
 		public override IEnumerable<Position> GetModdedConstraints() {
 			// By default, modded map layers are positioned between two vanilla layers (via After and Before) and are ordered in load order.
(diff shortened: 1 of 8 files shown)


Chicken-Bones (approved):

ExampleMod/Common/ExampleMapLayer.cs
```diff
@@ -7,12 +10,19 @@
 using Terraria.Map;
 using Terraria.ModLoader;
 using Terraria.UI;
+using static Terraria.Map.IMapLayer;
```
Chicken-Bones: We don't normally support `using static` in `ExampleMod` because it can be a bit annoying to teach. I suppose it's needed here because `Position` is defined in an interface?

Chicken-Bones (approved):

ExampleMod/Common/ExampleMapLayer.cs
Jadams505: > I suppose it's needed here because `Position` is defined in an interface?

That is correct I opted to be explicit with `IMapLayer.Pings`, so that you could use autocomplete to choose the right layer, but I can remove using static and be explicit with `IMapLayer.Position` if it is confusing for new modders.

Actually I was thinking technically the positioning logic is never ran on vanilla layers which only implement `IMapLayer`, so technically we could just add the positioning hooks to `ModMapLayer` and leave `IMapLayer` unchanged. This would move `Position` into `ModMapLayer` and thus eliminate the need for `using static`. Thoughts?

Chicken-Bones: That is the best solution, please do so

Chicken-Bones (approved):

patches/tModLoader/Terraria/ModLoader/ModMapLayer.cs
```diff
@@ -9,6 +10,25 @@ public abstract class ModMapLayer : ModType, IMapLayer
 {
 	public bool Visible { get; set; } = true;
 
+	/// <summary>
+	/// Returns the map layer's default position in regard to vanilla's map layer ordering. Make use of e.g. <see cref="Before"/>/<see cref="After"/>, and provide a map layer.<br/><br/>
+	///
+	/// <b>NOTE:</b> The position must specify a vanilla <see cref="IMapLayer"/> otherwise an exception will be thrown.<br/>
+	/// By default, this hook positions this map layer after all vanilla layers.
+	/// </summary>
+	public virtual Position GetDefaultPosition() => new Before(null);
```
direwolf420: I find the syntax a bit strange. docs say `after all vanilla layers.`, but the code says `new Before(null)`, which implies it's before all vanilla layers. Is passing null and how it affects position documented to reflect this unintuitive relationship?

patches/tModLoader/Terraria/ModLoader/ModMapLayer.cs
```diff
@@ -9,6 +10,25 @@ public abstract class ModMapLayer : ModType, IMapLayer
...
+
+	/// <summary>
+	/// Modded layers are placed between vanilla layers via <see cref="GetDefaultPosition"/> and, by default, are sorted in load order.<br/>
+	/// This hook allow you to sort this map layer before/after other modded layers that were placed between the same two vanilla layers.<br/>
```
direwolf420: allow -> allows

patches/tModLoader/Terraria/ModLoader/ModMapLayer.cs
Jadams505: `Before(null)` means this layer should be before nothing, which means it is after everything. I don't love it either, but that's how `ExtraJump` does it and I figured they should be consistent.
I can add special positions `BeforeFirstVanillaLayer` and `AfterLastVanillaLayer` like `PlayerDrawLayers` to make it more clear.

JavidPack:
Thanks a ton, this was well done.

PR #4593, opened 2025-03-29
Add networking related docs

stormytuna:
Currently only adds docs to ModPlayer virtual methods. I'm planning to document most of the more commonly used classes (Global/Mod Item/NPC/Projectile, ModSystem), but that will take me some time, so I will do them in chunks over the next however long this takes. I've also fixed one place that wasn't displaying properly on Rider due to the content being _within_ a `<br>` element, and added clarification to `ModPlayer` `HideDrawLayers`.

If you wish to help, please leave a comment so I know I can skip whatever section you'd like to help with.

This snippet will be helpful, simply override a method and call `Print` within it, you should see "Local!!" printed if it's ran on the local client, `Server!!` if the local client's player has this information processed, and "Client!!" printed in the other tML instance if that client processes the local client's information.
```csharp
	public class Test : ModPlayer
	{
		private void Print() {
			if (Main.gameMenu || Player.whoAmI != 0) {
				return;
			}
			
			if (Main.myPlayer == 0) {
				Main.NewText("Local!!");
			} else if (Main.myPlayer == 1) {
				Main.NewText("Client!!");
			} else {
				ChatHelper.BroadcastChatMessage(NetworkText.FromLiteral("Server!!"), Color.White, 1);
			}
		}
	}
```

I'll also note things as I go through, here are the notes for `ModPlayer`:
```txt
ModPlayer GetHealLife, GetHealMana, ModifyManaCost, ModifyWeaponDamage called locally for tooltips, but everywhere for item use
ModPlayer ModifyWeaponDamage, CanConsumeAmmo, OnConsumeAmmo called locally for some projectiles
ModPlayer ModifyItemScale has many usages I didnt look at in detail
ModPlayer CanUseItem also called locally constantly
Didn't bother with pvp related hooks
Also added clarification to HideDrawLayers as it's not immediately obvious how to use it
Fixed where content was within <br> elements as it wasnt showing on rider
```


stormytuna:
Notes on `ModItem` and `GlobalItem`
- Clarified that `ModifyTooltips` is ran on a clone of the item, I've seen this trip a few people up in the help channel and it seems reasonable to document
- Updated `GlobalItem::RightClick` to not recommend using it for goodie bag items, assumedly this is a remnant from 1.3, I just copied over the docs from `ModItem::RightClick` and tweaked the wording to make sense

JavidPack:
Let's get whatever is ready for now in before the end of the month. Is there anything remaining in the currently added docs that needs changing before I review and merge?

stormytuna:
I'm almost finished with Mod/GlobalNPC, I'll try get that finished today or tomorrow
I'll also do a cursory once over all of them again to make sure wording is fine

stormytuna:
Did a quick look through, nothing egregious but I may have missed something. Do you want me to fix conflicts or leave it?

JavidPack:
I'll do that.

JavidPack:
There are a few that are noted as "Called on the server only.", but that's not true. They are called in single player or on the server. Are we just ignoring single player for these docs since pretty much all of them are called in single player mode? As it is, it seems a bit confusing if someone were to see that in the docs.

Also, many of the ModProjectile drawing methods say "Called on the local client only.", but that's not true either, they are called on all clients. (but not the server)

Many of the ModSystem methods mention "Called on the local client and server.", but they are called on all clients actually.

stormytuna:
I was writing them with the assumption that they're always called in singleplayer yeah, if you think clarifying that it's also called in singleplayer is better than I'm fine with that

I can fix the ModProjectile methods, it was just an oversight, I'll go back through all drawing related hooks and make sure they're correct

And I'll fix ModSystem ones too, those were a decision I changed my mind on later, I'll also use the wording you updated some with - "Called on the server and *remote* clients"

JavidPack:
Looks great, merging.

I've also added this section to the wiki with some more info we can point modders to if they are interested in more information:  https://github.com/tModLoader/tModLoader/wiki/Basic-Netcode#hooks-in-multiplayer

PR #4598, opened 2025-04-02
Tile Placement Preview Draw Hooks

### LINKED ISSUE #4597: [API Addition] Add tile hooks in TileObject.DrawPreview for better control over the appearance of placement previews

Yes

Mod capability as a Modder

### Description

My proposal consists of two new hooks being added to `ModTile` and `GlobalTile`, named `PreDrawPreview` and `PostDrawPreview`.
These hooks would allow modders to easily define dynamic drawing behavior for a tile's placement preview.

Currently, a cross-mod compatible system for drawing dynamic placement previews doesn't exist, and current solutions would be difficult to keep compatible and non-breaking.

### Example Usage:

In [Terramon](https://github.com/JamzOJamz/Terramon), we wanted to communicate a change in data held by an item via the preview of the tile that the item placed. We were able to achieve this with an IL edit:

https://github.com/user-attachments/assets/9c19ac8f-6cf7-4873-8ff2-b856c733d9ee

---

For tiles that have custom drawing, or custom logic when placed, it might be useful to use these hooks to communicate that extra information about the tile before placement.

---

This is my first time contributing, I'm looking forward to some feedback on the draft PR that will come with this issue.

For modders, the only alternatives are either a detour that completely rewrites `TileObject.DrawPreview`, or an [IL edit.](https://github.com/JamzOJamz/Terramon/blob/feature/banners/Content/Tiles/CustomPreviewTile.cs#L22)

queueAngel:
Fixes #4597 

### What is the new feature?
Pre-draw and post-draw hooks specifically for placement previews in `ModTile` and `GlobalTile`. These are named `PreDrawPlacementPreview` and `PostDrawPlacementPreview`.

Currently the placement preview for a tile is done by drawing the associated tile sprite. Due to custom tile rendering, sometimes the actual tile sprite is limited and does not communicate the final visuals of the tile. It is also possible that the final placed tile will differ from the placement preview due to `ModTile.PlaceInWorld` logic.

For example, vanilla boss relics and even `MinionBossRelic` from ExampleMod all consist of the tile sprite (pedestal) and a bobbing relic that is drawn manually. When placing, the user can't see the final result:

Preview:

image

Actual result:

image

This functionality cannot be moved to be part of another hook (such as `PreDraw`) without significant changes to the method signature and existing calls.
A library mod could be made, but I see no reason to not just have this feature be available natively. I talked about the problems with other potential implementations in the corresponding issue.

The new hooks are called once for each section. This is to support drawing each section in the red or white tint depending on if the location is valid or invalid. It is possible to only call the hooks once for each multitile, but we want to make drawing with the correct color the default approach rather than a hack the modder would have to implement themselves.

Modders can use `PreDrawPlacementPreview` to skip the vanilla drawing altogether or use it to adjust some of the parameter values used to draw the preview. The following example shows a wide variety of potential ways to use the new hooks:

Consider the following code. It adjusts the draw color for valid tiles to be rainbow colored. It randomizes the draw position while the player has the Tipsy debuff. It uses `PlaceInWorld` and `PreDrawPlacementPreview` together to allow the player to place the tile in the off state while holding the down button. And finally, it animates the preview, normal previews do not animate.

```cs
public override bool PreDrawPlacementPreview(int i, int j, SpriteBatch spriteBatch, ref Rectangle frame, ref Vector2 position, ref Color color, bool validPlacement, ref SpriteEffects spriteEffects) {
	if(validPlacement && Main.LocalPlayer.direction == 1) {
		color = Main.DiscoColor;
	}
	if (Main.LocalPlayer.tipsy) {
		position += Main.rand.NextVector2Circular(4, 4);
	}
	if (Main.LocalPlayer.controlDown) {
		frame.Y += 288;
	}
	else {
		frame.Y += 36 * Main.tileFrame[Type];
	}
	return true;
}

public override void PlaceInWorld(int i, int j, Item item) {
	if (Main.LocalPlayer.controlDown) {
		ToggleTile(i, j);
	}
}
```

This video shows all those in action:

https://github.com/user-attachments/assets/d60493ed-353c-4bfe-b1a9-b955f95f9955

`PostDrawPlacementPreview` is similar, it runs after vanilla drawing and as such can be used to draw over the normal preview image.

### ExampleMod updates
[MinionBossRelic.cs](https://github.com/tModLoader/tModLoader/pull/4598/files#diff-295f2ae176a5c595d029de278e78edc024d6ed76508f9db7575774671bcbee95) has been updated to use these hooks, as shown below. Remember that the actual relic texture is drawn separately from the tile sprite, which is just the pedestal portion.

In this video, a vanilla relic tile is placed on the left of the player, showing the placement preview limitations without the new hooks. On the right `MinionBossRelic` is placed. Note how the preview shows the relic texture and even how the texture is drawn in red when the placement is invalid. It is entirely possible to animate the relic in the preview, if a modder wishes to do that.

https://github.com/user-attachments/assets/adeae68f-7930-473b-b8fb-98ad5693c66e


patches/tModLoader/Terraria/TileObject.cs.patch
```diff
@@ -87,3 +95,17 @@
  				if (num36 != num38 || num37 != num39)
  					flag9 = true;
  			}
+@@ -743,7 +_,12 @@
+ 					if (op.Type == 114 && j == 1)
+ 						num12 += 2;
+ 
+-					sb.Draw(sourceRectangle: new Rectangle(x, num8, coordinateWidth, num12), texture: value, position: new Vector2(num9 * 16 - (int)(position.X + (float)(coordinateWidth - 16) / 2f) + drawXOffset, num10 * 16 - (int)position.Y + num5), color: color, rotation: 0f, origin: Vector2.Zero, scale: 1f, effects: spriteEffects, layerDepth: 0f);
```
Chicken-Bones: Please put this line a block comment instead of removing it. https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#comment-out-vanilla-code-rather-than-removing-it

JavidPack:
This is marked as a draft, is there anything else that needs to be done @Angeltrisi ? I'm looking into it, going to try to make an ExampleMod example and finish it up.

Also, why is `frame` a `Rectangle?` instead of `Rectangle`?

queueAngel:
@JavidPack I wasn't thinking too much about the design of the hook at the time so I guess I was unsure of it? I thought that perhaps for convenience, a bool for checking if the current drawn section is colored red (blocked) would be good.
Otherwise, that bool could be inferred by the user by checking collision again, or checking the color directly, like for example `if (color) == Color.Red * 0.35f`. Similarly, tile position can be inferred from the given position parameter.

About the rectangle thing, it was an oversight, I've changed it now.

PR #4604, opened 2025-04-04
Throw exception when player draw layers attempt to position relative to other movable layers

Chicken-Bones:
### What is the bug?

A lack of validation in `PlayerDrawLayer.Between(after, before)` meant that if the layers `after` or `before` were not positioned with `Between` themselves, then those layers would be added into the draw order multiple times, affecting rendering and potentially creating conflicts.

Specifically, the layer `HeldItem` has `Multiple` positions depending on `PlayerDrawSet.weaponDrawOrder` so a constraint like `Between(HeldItem, ArmOverItem)` is not valid.

It's reported that two mods had `Between(ArmOverItem, HeldItem)` and `Between(HeldItem, ArmOverItem)` neither of which are valid, but which didn't throw an exception until the mods were loaded together.

### How did you fix the bug?

Added validation of the constraints with `throw new ArgumentException($"{layer.FullName} cannot be positioned after {after.FullName} because it does not have a fixed position. Consider using AfterParent or referring to a different layer (or null)")`

We could ignore these constraints and print a warning to the logs instead. Perhaps we should do so for a transition period.


JavidPack:
I tried to replicate this by changing `ExamplePlayerDrawLayer` to `public override Position GetDefaultPosition() => new Between(PlayerDrawLayers.ArmOverItem, PlayerDrawLayers.HeldItem);`, but it did not throw an error. I believe this is what is not supposed to be allowed.

I changed `if (b.Layer2 is { } before && !positions.ContainsKey(before))` to `if (b.Layer2 is { } before && (!positions.ContainsKey(before) || positions[before] is Multiple))` and it did throw an error, but I'm not positive if that is the correct approach since I'm not familiar with what `position` and `betweens` are supposed to contain at this stage.

Also, I think the error message could be a little clearer, it's not 100% clear what "it" in "it does not have a fixed position" refers to. It might be best to repeat the layer we are referring to.

image

PR #4606, opened 2025-04-06
UpdateVisibleAccessory and UpdateItemDye hooks for items, CustomVisualEquipType example showcase

### LINKED ISSUE #4055: UpdateVisibleAccessory item hook

No, unless other contributors are too busy

Mod capability as a Modder

### Description

A hook in `Player.UpdateVisibleAccessory` and `Player.UpdateVisibleAccessories` (for items that change the player's appearance even when hidden) that items can call to set flags that are then checked by PlayerDrawLayers and other drawcode, just like various vanilla vanity items do.

Currently, items cannot set flags on players drawn on the main menu. `UpdateAccessory` and `UpdateVanity` are not called since `UpdateEquips` is only called during gameplay.

`Player.ApplyEquipVanity` could be called on items, but this would deviate from vanilla behaviour and cause the werewolf and merfolk transformations to be displayed on the main menu (unless this is changed in 1.4.5?).

flye-name:
### What is the new feature?
Resolves https://github.com/tModLoader/tModLoader/issues/4055
Adds `UpdateVisibleAccessory` in `ModItem` and `GlobalItem`
Adds `UpdateItemDye` in `ModItem` and `GlobalItem`
Add the CustomVisualEquipType folder, containing a full implementation of a custom visual `EquipType` (earrings).

As the original issue states:
> Currently, items cannot set flags on players drawn on the main menu. UpdateAccessory and UpdateVanity are not called since UpdateEquips is only called during gameplay.

Modders should use `UpdateVisibleAccessory` and `UpdateItemDye` to set custom draw flags, such as with a custom EquipType or similar visual.

As the original issue states:
> Player.ApplyEquipVanity could be called on items, but this would deviate from vanilla behavior and cause the werewolf and merfolk transformations to be displayed on the main menu (unless this is changed in 1.4.5?).

In an accessory:
```cs
public override void UpdateVisibleAccessory(Player player, bool hideVisual) {
    if (!hideVisual) // hideVisual could be used for some accessories that would change the shape of the player + add a visual, so hiding the accessory would keep the shape change and hide the accessory itself
        player.GetModPlayer<MyModPlayer>().drawsThing = true;
}
```

In a `PlayerDrawLayer`:
```cs
public override bool GetDefaultVisibility(PlayerDrawSet drawInfo) {
    return drawInfo.drawPlayer.GetModPlayer<MyModPlayer>().drawsThing;
}
```


Solxanich (changes requested):
This should realistically be one single hook in 
Player.UpdateVisibleAccessory().

A nice to have is if pass whether or not it is vanity to modders as a bool in the hook as opposed to a second hook instead. 

Discussion needed on whether a ModPlayer.UpdateVisibleAccessory taking the same parameters as the Player.UpdateVisibleAccessory or continue with ModItem/GlobalItem hook.

Solxanich (approved):

jjohnsnaill:
Could `itemSlot` be passed into this hook? That would allow modders to more conveniently check for dyes and other things involving the slot.

JavidPack:
I've taken the opportunity to make an example equipment slot to further test the hooks and see what else is lacking.

Some thoughts:
1. While we could pass in the `slot` and `modded` to `ResetVisibleAccessories`, we already have an `UpdateDyes` hook in `ModPlayer`, maybe we should have one in `Global`/`ModItem` as well instead of passing those into `ResetVisibleAccessories`.  (for consistency with vanilla and existing hooks). Having to check `hideVisual` 99% of the time in a hook called `UpdateVisibleAccessory` seems weird as well. Current implementation manually iterates over all vanilla and modded accessory slots in ModPlayer.UpdateDyes to update the shaders.
2. There is no good place to implement something similar to `AutoloadEquip`. That code happens in `ModItem.Register`, but `GlobalItem` won't work for that. The current implementation is in `GlobalItem.SetDefaults`, but that will be called multiple times so it requires additional checks. 
3. Current implementation doesn't use `EquipTexture`, so it can't delegate hooks to specific textures if visible.

# Video

Here is the new equipment slot in action, matching vanilla behavior for visibility, ordering, dyes, etc.

https://github.com/user-attachments/assets/2983dc52-d50e-4244-8ad7-d18228ccf000

ExampleMod/Common/CustomEquipmentSlot/EarringsGlobalItem.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace ExampleMod.Common.CustomEquipmentSlot
+{
+	/// <summary>
+	/// EarringsGlobalItem manages registering and assigning earring equipment slot IDs.
+	/// </summary>
+	public class EarringsGlobalItem : GlobalItem
+	{
+		internal static int nextEarringSlot = 1; // 0 is empty slot.
+		internal static readonly Dictionary<int, int> earringItemToSlot = new();
+		internal static readonly Dictionary<int, Asset<Texture2D>> earringSlotToTexture = new(); // or array?
```
Chicken-Bones: Definitely just `List<>` here, and use `.Count` instead of `nextEarringSlot`

ExampleMod/Common/CustomEquipmentSlot/EarringsGlobalItem.cs
```diff
@@ -0,0 +1,63 @@
...
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.CustomEquipmentSlot
+{
+	/// <summary>
+	/// EarringsGlobalItem manages registering and assigning earring equipment slot IDs.
+	/// </summary>
+	public class EarringsGlobalItem : GlobalItem
```
Chicken-Bones: This is half `GlobalItem` half 'loader'. That's not a problem per se, but the example is really hard to read.

ExampleMod/Common/CustomEquipmentSlot/EarringsGlobalItem.cs
```diff
@@ -0,0 +1,63 @@
...
+	{
+		internal static int nextEarringSlot = 1; // 0 is empty slot.
+		internal static readonly Dictionary<int, int> earringItemToSlot = new();
+		internal static readonly Dictionary<int, Asset<Texture2D>> earringSlotToTexture = new(); // or array?
+
+		public override bool InstancePerEntity => true;
+
+		public int earringSlot = -1;
```
Chicken-Bones: Having this be a `GlobalItem` does allow people to re-assign the `earringSlot` property for the item dynamically, but I think this example would be better and simpler with just a static `Dictionary [Item.type] = Asset<Texture2D>`, then `earringSlot` could just be an item id.

ExampleMod/Common/CustomEquipmentSlot/EarringsGlobalItem.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		public override void SetStaticDefaults() {
+			// Add an earring equipment slot to a vanilla item.
+			AddEarringTexture(ItemID.AnglerEarring, "ExampleMod/Common/CustomEquipmentSlot/AnglerEarring_Earrings");
+		}
+
+		public override void SetDefaults(Item entity) {
+			if (entity.ModItem is ModItem modItem && modItem.GetType().GetAttribute<AutoloadEquip_EarringAttribute>() != null) {
```
Chicken-Bones: This is fine from a performance standpoint, but it's a bit out of line with other uses of the 'autoload' terminology. Runtime registration would fail if someone was trying to make `EarringsID.Sets`, or net sync the earring id.

ExampleMod/Common/CustomEquipmentSlot/EarringsPlayer.cs
```diff
@@ -0,0 +1,74 @@
...
+	/// EarringsPlayer stores which earring slot will be drawn.
+	/// </summary>
+	public class EarringsPlayer : ModPlayer
+	{
+		// The earring equipment slot ID.
+		public int earring = -1; // Actually, should this be ear?
+		public int earringShader;
+
+		public override void ResetEffects() {
+			// TODO: Do we need a ResetVisibleAccessories? Player.Update and UpdateDead don't have the usual ResetEffects->ResetVisibleAccessories call order so there might be some edge case.
```
Chicken-Bones: Depends if there's any way to change the visible accessories in those states.

ExampleMod/Common/CustomEquipmentSlot/EarringsPlayer.cs
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		private void UpdateModdedSlotDyes(bool socialSlots) {
+			var loader = LoaderManager.Get<AccessorySlotLoader>();
+			var ModAccessorySlotPlayer = Player.GetModPlayer<ModAccessorySlotPlayer>();
+
+			for (int i = 0; i < ModAccessorySlotPlayer.SlotCount; i++) {
+				if (loader.ModdedIsItemSlotUnlockedAndUsable(i, Player)) {
+					var slot = loader.Get(i, Player);
+					UpdateItemDye(!socialSlots, slot.HideVisuals, socialSlots ? slot.VanityItem : slot.FunctionalItem, slot.DyeItem);
```
Chicken-Bones: yeah, ideally this would be a hook on `ModItem`

Chicken-Bones (changes requested):

ExampleMod/Common/CustomEquipmentSlot/EarringsPlayer.cs
JavidPack: Made a hook and replaced this code.

ExampleMod/Common/CustomEquipmentSlot/EarringsGlobalItem.cs
```diff
@@ -0,0 +1,35 @@
+using Terraria;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.CustomEquipmentSlot
+{
+	// Note: To fully understand this example, please start by reading https://github.com/tModLoader/tModLoader/tree/stable/ExampleMod/Common/CustomEquipmentSlot/README.md
```
Chicken-Bones: I think that this should not be called an 'equipment slot' perhaps `CustomAccessoryVisual`

Chicken-Bones (changes requested):

PR #4611, opened 2025-04-08
Custom ModType Example

JavidPack:
Custom `ModType`s have been available for modders to use for a while, but ExampleMod has been lacking an example implementation. This PR adds a custom `ModType` called `ModVictoryPose` showing off how modders can utilize `ModType` to easily implement new content types.

A `ModVictoryPose` is a short sequence that will trigger when a player defeats a boss. A banner texture will be displayed, a victory cheer message will be displayed, and the player will be affected by the provided hooks to affect their pose and other effects.

# ModType
The main concept taught is implementing a `ModType`.  There are many benefits to using a ModType class for making new content. The main benefit is that loading the content is handled using the same logic as existing content types, so other modders will be able to use them as expected as if they were provided by tModLoader itself. Other modders will simply need to inherit from the base class, no need for `Mod.Call` to interface with the mod providing the `ModType`. This also includes loading a primary texture, automatic localization registration, being able to use `ModContent.Find` and `ModContent.GetInstance`, support for custom ID sets, and support for manually loading multiple instances of a class.

## Loading Content
Just like with other content types added by tModLoader, `ModType`s will autoload automatically. Mods implementing a custom `ModType` simply need to inherit from that class as usual. Manually loading is also supported, as shown in `VictoryPoseAdditionalLoader` and `NonAutoloadVictoryPose`. Supporting these two approaches allows modders to use custom ModType using the experience they already know.

## Primary Texture
The example inherits from `ModTexturedType` instead of `ModType`. `ModTexturedType` inherits from `ModType` and adds a `Texture` property for the content's texture path.

How the primary texture is used is up to the modder, but this example displays the texture behind the player, celebrating the victory. 

In `SetupContent`, modders should make sure to call `ModContent.Request<Texture2D>(Texture);` to ensure that the texture exists. This will prevent the mod from loading if the texture is missing, which is usually the intended behavior.

## Automatic Localization Registration
Automatically populating the localization files is another benefit to `ModType`. Modders can call `ILocalizedModType.GetLocalization` (if implementing the `ILocalizedModType` interface) or `Language.GetOrRegister` during the `SetupContent` method to register a localization key for the content. In this example, the line `_ = VictoryCheer;` ensures that the localization entries are registered for the content.

## Content Retrieval
The `ModTypeLookup<ModVictoryPose>.Register(this);` line in `Register` ensures that `ModContent.Find` and `ModContent.GetInstance` are supported.

## Custom ID Sets
The `VictoryPoseID` class contains custom ID sets for this content. The sets behave exactly the same as vanilla sets, allowing familiar programming patterns. (Other mods can add their own custom ID sets as well if they want.)

The `VictoryPoseID.Search.Add(FullName, Type);` call makes it so the `/customsets` chat command can show content names.

# Discussions
1. ~~File layout. I'm thinking at the very least the actual `ModVictoryPose` content should be in their own classes, but having all the code together does help guide the modder in learning step by step as topics are explored.~~ Split into multiple files, also added a Readme.md that should be useful.
2. ~~Do we need to implement multiplayer syncing, or leave that up to the modder if relevant for their `ModType`.~~ Implemented.
3. Organization: Currently `VictoryPoseLoader` and `VictoryPoseID` both exist to mirror tModLoader implementations, but I guess technically they can be the same class. The `Sets` inner class can also go if we want. 
4. The current example doesn't include `VictoryPoseID.DefaultContent` const ints, which we could implement, but that would mean the load order in the example would need to be forced and I don't think there is much benefit for having const ints for "default content" as we see with vanilla content ids.
5. `ModType.LocalizationCategory` can potentially conflict with other mods. The BossChecklist mod uses keys like `Mods.ModName.NPCs.NPCName.BossChecklistIntegration.SpawnInfo` to avoid any potential issue, but it might be overkill to suggest that localization categories need to include the source mod name.
6. `ModPlayer.OnHitNPC` is being used as a `OnKillNPC` method since one doesn't exist. An `OnKillNPC` method would be great to add, it has been requested, but we need it to handle all sorts of damage sources so it might get complicated to implement a working hook.
7. The current texture usage is a bit contrived, maybe we can do something else or make better victory banner images.
8. `ModVictoryPose.Clone`, do we need to go into more detail about proper non-singleton usage of `ModType`?

# Showcase
Here is `ModVictoryPose` in action. Shown is `HandsUpWithFireworksVictoryPose`. This implementation poses for 3 seconds, prevents the 1st autoswing to strike a pose, holds both hands up, plays thunder sounds, and spawns fireworks. The VictoryCheer and texture are also spawned by default for `ModVictoryPose`

https://github.com/user-attachments/assets/f69f3e4b-251e-4c82-acd6-245d9a7d3cdc

# Other Mod
One of the main benefits of using `ModType` is allowing other mods to easily add additional content. This example shows a mod adding their own `ModVictoryPose`. The `ModVictoryPose` will automatically be loaded just by nature of the class existing. This Mod will need a mod reference to ExampleMod to load correctly.

```cs
using ExampleMod.Content.CustomModType;
using Terraria;
using Terraria.Audio;
using Terraria.ID;
using Terraria.ModLoader;

namespace MoreVictoryPoses
{
	public class MoreVictoryPoses : Mod
	{
	}

	public class MyVictoryPose1 : ModVictoryPose
	{
			PoseTime = 120;

			VictoryPoseID.Sets.NonBoss[Type] = true;
		}
		
		public override void OnStartPose(Player player) {
			Main.blockMouse = true;
		}
		
		public override void Update(Player player) {
			if (player.itemAnimation == 0) {
				player.bodyFrame.Y = player.bodyFrame.Height * 5; // 2 Hands up / falling
			}

			float elapsedPoseTime = ElapsedPoseTime(player);
			if (elapsedPoseTime == 30 || elapsedPoseTime == 60) {
				SoundEngine.PlaySound(SoundID.Thunder);
			}
		}
	}
}
```

# Additional Teachings
## IParticle
The victory banner texture spawned is currently implemented as a `IParticle`. I'm not sure yet exactly what benefits an `IParticle` have over `Dust`, but modders have been asking about it. I think `IParticle` is just a more customizable but harder to implement entity. It also is guaranteed to spawn where Dust potentially won't spawn due to graphics settings and current dust counts.


ExampleMod/Content/CustomModType/CustomModType.cs
```diff
@@ -0,0 +1,353 @@
...
+			ChatHelper.DisplayMessage(activeVictoryPose.VictoryCheer.ToNetworkText(), Color.White, (byte)Main.myPlayer);
+
+			Asset<Texture2D> texture = ModContent.Request<Texture2D>(activeVictoryPose.Texture);
+			Rectangle? frame = activeVictoryPose.GetTextureFrame(texture);
+			Main.ParticleSystem_World_BehindPlayers.Add(new PoseIconParticle(texture, frame, new Vector2(Main.rand.NextFloat(-2, 2), -5f), Player.Center - new Vector2(0, 80), Main.rand.NextFloat(-0.03f, 0.03f)) {
+				AccelerationPerFrame = new Vector2(0f, 0.16350001f),
+				ScaleOffsetPerFrame = 1f / 60f,
+			});
+
+			// TODO: A real implementation would want to sync the pose to other clients with a ModPacket
```
Chicken-Bones: Seems easy enough. Only need one packet for pose start, can call `OnStartPose` on all clients, good example of documenting sidedness of hooks ig.

Only thing to consider is that the packet may arrive while an existing pose is still running if a remote client is running slow. Just need to end the running pose immediately and call `OnEndPose` in that case

ExampleMod/Content/CustomModType/CustomModType.cs
```diff
@@ -0,0 +1,353 @@
...
+	public class VictoryPoseLoader : ILoadable
+	{
+		internal static readonly List<ModVictoryPose> VictoryPoses = [];
+
+		internal static int VictoryPoseCount { get; set; } = 0;
+
+		internal static int Add(ModVictoryPose victoryPose) {
+			VictoryPoses.Add(victoryPose);
+			return VictoryPoseCount++;
+			// or return VictoryPoses.Count - 1;
```
Chicken-Bones: Do this, or `int type = VictoryPoses.Count; VictoryPoses.Add(victoryPose); return type;`

ExampleMod/Content/CustomModType/CustomModType.cs
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+		public void Load(Mod mod) {
+		}
+
+		public void Unload() {
+		}
+	}
+
+	public class VictoryPoseAdditionalLoader : ILoadable
```
Chicken-Bones: I think the name of this class is potentially confusing, also I recommend making it an inner class of the `NonAutoloadVictoryPose`

ExampleMod/Content/CustomModType/CustomModType.cs
```diff
@@ -0,0 +1,353 @@
...
+			activeVictoryPose.Update(Player);
+			activeVictoryPose.ElapsedPoseTime++;
+			if (activeVictoryPose.ElapsedPoseTime >= activeVictoryPose.PoseTime) {
+				activeVictoryPose.OnEndPose(Player);
+				activeVictoryPose = null;
+			}
+		}
+	}
+
+	public class VictoryPoseLoader : ILoadable
```
Chicken-Bones: I think it best to split all classes here to their own `.cs` files, as long as they remain contained in the one folder

ExampleMod/Content/CustomModType/CustomModType.cs
```diff
@@ -0,0 +1,353 @@
...
+
+		public override void Update(Player player) {
+			if (player.itemAnimation == 0) {
+				player.bodyFrame.Y = player.bodyFrame.Height; // 1 Hand up
+			}
+		}
+	}
+
+	// A slightly tweaked CreativeSacrificeParticle. IParticle are similar to Dust, but can be customized completely. 
+	public class PoseIconParticle : IParticle
```
Chicken-Bones: ...except this one, it could be an inner class, or just a companion class to `VictoryPosePlayer`

Chicken-Bones (approved):

JavidPack:
More changes. The most recent concern is `ModVictoryPose.Clone`, which I added since there can technically be multiple instances of a specific `ModVictoryPose` active and `ModVictoryPose.ElapsedPoseTime` is tracked on the `ModVictoryPose`. 

I'm not sure if the current comments and implementation are sufficient for typical usage

ExampleMod/Content/CustomModType/ModVictoryPose.cs
```diff
@@ -0,0 +1,83 @@
...
+
+		/// <summary>
+		/// Use to control the frame of the texture to display.
+		/// </summary>
+		public virtual Rectangle? GetTextureFrame(Asset<Texture2D> texture) {
+			return null;
+		}
+
+		/// <inheritdoc cref="VictoryPosePlayer.ElapsedPoseTime"/>
+		public static float ElapsedPoseTime(Player player) => player.GetModPlayer<VictoryPosePlayer>().ElapsedPoseTime;
```
Chicken-Bones, line 79: Should be in the same place as `GetActivePose` whether you put them both here or in the loader I don't mind.

Chicken-Bones (approved):

direwolf420:
Can you include setting a value in the custom named set in the "Other Mod" section?

JavidPack:
> Can you include setting a value in the custom named set in the "Other Mod" section?

Added

direwolf420:
A future PR could add some code for I/O-ing a ModType via TagCompound, particularly with using FullName, and keeping track of unloaded ones

PR #4612, opened 2025-04-08
(Mod|Global)Wall/ModPlayer.CanBeTeleportedTo()

### LINKED ISSUE #4596: [Feature Request] Add hook to conditionally blacklist tiles from being teleported to

No

Mod capability as a Modder

### Description

The **Teleportation Potion** picks a random location in the world and teleports the player to that. There are some implemented vanilla restrictions, both general and conditional. The conditional ones are interesting for modders that add special structures or biomes to the world that are progression-locked and shouldn't ever be entered until a certain point. This extends to other means of getting into those places such as the **Rod of Discord/Harmony**.

The vanilla conditions for Teleportation Potion include Dungeon before Skeletron, and Lihzard Temple before Plantera. The RoD/RoH only checks for the latter.

Those checks are within `Player.CheckForGoodTeleportationSpot` and `Player.ItemCheck_UseTeleportRod`.

The hook would likely find its place in the `GlobalTile` class, and cover both cases.

Help modders avoid workarounds like IL editing or detouring + repeating the method until it doesn't hit the blacklisted tile.

separate hooks for each way of teleportation.
The hook can obviously also be used for non-conditional rules such as "never teleport into water or honey".

flye-name:
### What is the new feature?
Resolves https://github.com/tModLoader/tModLoader/issues/4596
Adds `CanBeTeleportedTo()` to `ModWall`, `GlobalWall` and `ModPlayer` to give modders the ability to blacklist their tiles from things like teleportation potions or the rod of discord.

As the original issue states: 
> Help modders avoid workarounds like IL editing or detouring + repeating the method until it doesn't hit the blacklisted tile.

As the original issue states:
> separate hooks for each way of teleportation.
The hook can obviously also be used for non-conditional rules such as "never teleport into water or honey".

In a `ModWall` within a structure the modder doesn't want players to reach early:
```cs
public override bool CanBeTeleportedTo(int i, int j, Player player, string context) => false;
```

In a `GlobalWall` to blacklist certain vanilla tiles:
```cs
public override bool CanBeTeleportedTo(int i, int j, int type, Player player, string context)
{
    if (type == WallID.SpiderUnsafe)
        return false;
    return base.CanBeTeleportedTo(i, j, type);
}
```

### DIFF
--- a/ExampleMod/Content/Walls/ExampleWallAdvanced.cs
+++ b/ExampleMod/Content/Walls/ExampleWallAdvanced.cs
@@ -1,3 +1,4 @@
+using ExampleMod.Common.Systems;
 using Microsoft.Xna.Framework;
 using Terraria;
 using Terraria.ID;
@@ -51,6 +52,11 @@ namespace ExampleMod.Content.Walls
 				}
 			}
 			return base.WallFrame(i, j, randomizeFrame, ref style, ref frameNumber);
+		}
+
+		public override bool CanBeTeleportedTo(int i, int j, Player player, string context) {
+			// Limit teleportation similar to how teleporting to a location with Dungeon walls or Lihzahrd walls is limited by defeating some bosses.
+			return DownedBossSystem.downedMinionBoss;
 		}
 	}
 }
--- a/patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
+++ b/patches/tModLoader/Terraria/ModLoader/CombinedHooks.cs
@@ -322,4 +322,9 @@ public static class CombinedHooks
 	{
 		return ItemLoader.OnPickup(item, player) && PlayerLoader.OnPickup(player, item);
 	}
+
+	public static bool CanBeTeleportedTo(Player player, int i, int j, string context)
+	{
+		return PlayerLoader.CanBeTeleportedTo(player, i, j, context) && WallLoader.CanBeTeleportedTo(i, j, Main.tile[i, j].WallType, player, context);
+	}
 }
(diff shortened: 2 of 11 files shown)


pilchardheck:
it would be so awesome
it would be so cool

direwolf420:
Regarding the patches in CheckForGoodTeleportationSpot:
Why do you currently call `TileLoader.CanBeTeleportedTo` twice, once before and once after the "raycast downward to find standable ground" step?
Also, wouldn't the best place for the hook call be just once, mainly after the one you currently have last, so before `Vector2 vector2 = Vector2.UnitX * 16f;` and all the `Collision.TileCollision` calls. That makes it filter out all the vanilla conditions first, causing less hook calls overall.

For the same reason, I'd shift the `ItemCheck_UseTeleportRod` patch down to just before the `Teleport` call aswell.

Also lastly, the docs should mention explicitely the code path for the more "general" patch (which is only used in vanilla by teleportation potion for the context of dungeon and temple), that being `Player.CheckForGoodTeleportationSpot`. It should mention that modders using this method will also call the hook automatically (most likely to their benefit).

flye-name:
> Why do you currently call `TileLoader.CanBeTeleportedTo` twice

Just didn't notice that I forgot to remove the previous one. Oops.

I'll fix that and move the checks down later today.

JavidPack:
I've removed the ModTile/GlobalTile hook. The coordinates almost never actually refer to an active tile anyway (unless non-solid or actuated) and would be misleading since the coordinates are at the top left of the player. A modder might assume it should work when really it would only check that specific tile. I added a ModPlayer hook for modders to do any sort of check they need, they can check every tile in the space if they want.

PR #4618, opened 2025-04-11
Uncapped sound effect pitch

### LINKED ISSUE #4301: Uncap sound effect pitch

No

Mod capability as a Modder

### Description

Remove the -1/1 clamp from `SoundStyle.Pitch` and `SoundEffectInstance.Pitch`.

FNA appears to support values lower than -1 and higher than 1 without issue, so this is now just an unnecessary restriction to modders.

_No response_

flye-name:
Resolves https://github.com/tModLoader/tModLoader/issues/4301

This also fixes the fact that `SoundEffectInstance` has a typo where `FAUDIO_DEFAULT_FREQ_RATIO` gets passed to the max frequency param in `FAudio_CreateSourceVoice`, instead of `FAUDIO_MAX_FREQ_RATIO`. I've PR'd a fix to this in the main FNA repo as well.

### DIFF
--- a/patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
+++ b/patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
@@ -25,9 +25,6 @@ public enum SoundLimitBehavior
 /// </summary>
 public record struct SoundStyle
 {
-	private const float MinPitchValue = -1f;
-	private const float MaxPitchValue = 1f;
-
 	private static readonly UnifiedRandom Random = new();
 
 	private int[]? variants;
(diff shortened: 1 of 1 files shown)


Mirsario (approved):
This may change behavior of already written code, but adding a second property will prove to be more of a hassle for users than to just announce the change well when it goes into preview.

Mirsario (changes requested):
Changes within FNA require tests and measurements for by how much the memory usage was increased in FAudio.

Mirsario (approved):
Memory usage differences seem very insignificant and unnoticeable in testing.

PR #4619, opened 2025-04-12
Long type support for ModConfig

### LINKED ISSUE #4613: Add support for the Long data type into the config

No

Mod capability as a Modder

### Description

Add support for long values in ModConfig and Config.Range to allow for larger numbers to be configured if necessary.

The config currently apparently doesn't properly support long values for config options, as well as long values for the Range(), adding support for those is necessary for a system I am working on and would be nice to have for everyone even if it doesn't come up that often.

_No response_

flye-name:
### What is the new feature?
Resolves https://github.com/tModLoader/tModLoader/issues/4613
Support for the `long` and `ulong` data types within `ModConfig`s

As stated by the original issue:
> The config currently apparently doesn't properly support long values for config options, as well as long values for the Range(), adding support for those is necessary for a system I am working on and would be nice to have for everyone even if it doesn't come up that often.

https://github.com/user-attachments/assets/baddc7f9-418a-4b65-ba67-c61e2541d087


direwolf420:
Can you show a screenshot of how a very large `long` value appears in the config? Both as a manual input one, and a slider

flye-name:
Sliders are unsupported

I've updated the description to include a video

JavidPack:
Thanks a bunch.

PR #4627, opened 2025-04-17
Catch sourceUpgradeTask exception and add button to view exception

JavidPack:
https://github.com/tModLoader/tModLoader/commit/6497b9d8c8b564695b9a5b584522184f5677ecf4 added a check for "dllReferences contains duplicate of modReferences" in `ReadBuildFile`. Since `ReadBuildFile` runs on the Mod Sources menu to check if the mod source needs an upgrade, this exception can be thrown when visiting the menu rather than just when building. When the exception happens in this manner, it is currently uncaught and crashes to desktop.

This change is on stable currently (although other similar issues can cause the same exception), so we will need to cherrypick this to stable and preview.

This fix catches the exception and adds a button the modder can click to view the error.

https://github.com/user-attachments/assets/eecc587e-7653-46be-a310-3d01342a5072

### DIFF
--- a/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
+++ b/patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
@@ -516,7 +516,7 @@
 		"DownloadNetSDK": "Download .NET SDK",
 		"DevModsInSandbox": "tModLoader cannot develop mods while running in a sandbox (such as Flatpak). Please make sure to install Steam directly on your system.",
 		"MSSourceIssue": "Mod Source Issue",
-		"MSSourceIssueMessage": "The following error was found with this mod's source: {0}",
+		"MSSourceIssueMessage": "The following error was found with the {0} mod's source code: {1}",
 
 		// Create Mod
 		"CreateModName": "ModName (no spaces)",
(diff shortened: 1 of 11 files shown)


patches/tModLoader/Terraria/Localization/Content/en-US/tModLoader.json
```diff
@@ -515,6 +515,8 @@
 		"MSNetSDKNotFound": "Making mods for tModLoader requires the .NET {0} SDK to be installed, but it has not been detected. Click the \"Download .NET {0} SDK\" button below for installation instructions. After installing, you may need to restart tModLoader.",
 		"DownloadNetSDK": "Download .NET SDK",
 		"DevModsInSandbox": "tModLoader cannot develop mods while running in a sandbox (such as Flatpak). Please make sure to install Steam directly on your system.",
+		"MSSourceIssue": "Mod Source Issue",
+		"MSSourceIssueMessage": "The following error was found with this mod's source: {0}",
```
Chicken-Bones: Message should probably say which mod it is, just to make it super user friendly.

patches/tModLoader/Terraria/ModLoader/UI/UIModSourceItem.cs
```diff
@@ -205,17 +205,23 @@ protected override void DrawSelf(SpriteBatch spriteBatch)
 
 		// Display Run tModPorter when .csproj is valid
 		if (sourceUpgradeTask is { IsCompleted: true }) {
-			bool result = sourceUpgradeTask.Result;
+			try {
+				bool result = sourceUpgradeTask.Result;
```
Chicken-Bones: Recommended to change this to `.GetAwaiter().GetResult()` to avoid wrapping the exception in an `AggregateException` (yeah, .NET didn't do great here)

Chicken-Bones (changes requested):

PR #4629, opened 2025-04-18
ModNPC.DeathMessage, ModNPC.ModifyDeathMessage(), and NPCID.Sets.IsTownChild

### LINKED ISSUE #4617: [Feature Request] IsATownChild set for NPCs

Yes

Mod capability as a Modder

### Description

A new boolean in NPCID.Sets that flags a town NPC as a child, the only difference is that any "X was slain..." message gets changed to "X has left..."

Currently altering a town NPC's death message to match that of the Angler or Princess requires IL edits (or setting IsTownPet to true but that brings other unwanted changes). 

Adding a TownNPCDeathMessage property / method to allow full control of what the message says.

This would allow for more flexibility but I'm not sure if it's worth the added effort.

flye-name:
### What is the new feature?
Resolves https://github.com/tModLoader/tModLoader/issues/4617
- Adds `ModNPC.DeathMessage` which modifies the message broadcasted when a town NPC or boss dies. This also affects what the town NPC's tombstone says in hardcore mode. This allows custom messages similar to The Twins, which were previously not possible. Color can also be modified with `ModNPC.ModifyDeathMessage()`.
- Adds `NPCID.Sets.IsTownChild` which prevents tombstones from being spawned in hardcore and sets the death message to "X has left!" unless specified otherwise by the method above. This is what the Angler and Princess use.

Both these additions open up more flexibility when making town NPCs, allowing modders to prevent tombstones from being spawned and to assign whatever message they'd want their NPC to broadcast when dying.

Unsure

A basic example is shown in the Example Mod change.

An interesting application would be to have different messages for the same NPC depending on certain conditions.

It's also possible to return false in `ModifyDeathMessage()` to skip the message altogether if needed.

### ExampleMod updates
Added a `DeathMessage` and `ModifyDeathMessage()` example to ExamplePerson

# Porting Notes
- `ModNPC.BossLoot(name, potionType)` became obsoleted and replaced with `ModNPC.BossLoot(potionType)`. Both will still be called for now but you should consider changing to `DeathMessage` ahead of time. 
- If you had a workaround for paired or group bosses, you can get rid of it and use `DeathMessage` instead to prevent duplicate boss defeat messages.


flye-name:
I've changed TownNPCDeathMessage to simply be DeathMessage and affect the "X has been defeated!" message for bosses.

direwolf420:
Why does the Global hook default to false while the Mod hook defaults to true? The docs both say `Returns true by default`

JavidPack:
Fixed that.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -404,6 +404,7 @@ public virtual void ModifyNPCLoot(NPCLoot npcLoot)
 
 	/// <summary>
 	/// Allows you to customize what happens when this boss dies, such as which name is displayed in the defeat message and what type of potion it drops.
+	/// <para/> Note that <see cref="DeathMessage(ref NetworkText, ref Color)"/> gives more control over the defeat message and supports localization and should be used instead.
```
Chicken-Bones: This version of `BossLoot` should be obsoleted (but still called for now) and replaced with a version that only has the `potionType` parameter.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -809,6 +810,18 @@ public virtual void OnGoToStatue(bool toKingStatue)
...
+	/// <para/> By default the text will be "NPC has been defeated!" for bosses and "NPC was slain..." (or "NPC has left!" if <see cref="NPCID.Sets.IsTownChild"/> or <see cref="NPCID.Sets.IsTownPet"/> are set to <see langword="true"/>) for town NPCs. Town NPC usually use <see cref="NPC.GetFullNetName"/> to retrieve the NPC name in the form of "GivenName the TypeName". For modded NPC, if a localization key for the ModNPC named "DeathMessage" exists, it will automatically be used and populated with <see cref="NPC.GetFullNetName"/> (or <see cref="NPC.GetTypeNetName"/> if a boss) even without using this hook.
+	/// <para/> The keys "Announcement.HasBeenDefeated_Plural" and "Announcement.HasBeenDefeated_Single" will be useful if creating a boss defeat message with a custom boss name. For example <c>customText = NetworkText.FromKey("Announcement.HasBeenDefeated_Plural", NetworkText.FromKey("Mods.MyMod.NPCs.MyBoss.TheTriplets"));</c> could be used for a death message for a boss with multiple enemies similar to The Twins. 
+	/// <para/> This won't have any effect if the given NPC isn't a town NPC or a boss.
```
Chicken-Bones: There's good info here, but too much info. The documentation should tell you when you need to use this hook. A lot of the info here looks like it could/should be applied in the `DeathMessage` localization instead.

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1323,6 +1323,24 @@ public static void BuffTownNPC(ref float damageMult, ref int defense)
...
 		}
 	}
+
+	private delegate bool DelegateDeathMessage(NPC npc, ref NetworkText custom, ref Color color);
+	private static HookList HookDeathMessage = AddHook<DelegateDeathMessage>(g => g.DeathMessage);
+
+	public static bool DeathMessage(NPC npc, ref NetworkText customText, ref Color color)
+	{
+		if(npc.ModNPC != null && Language.Exists(npc.ModNPC.GetLocalizationKey("DeathMessage"))) {
+			customText = NetworkText.FromKey(npc.ModNPC.GetLocalizationKey("DeathMessage"), npc.boss ? npc.GetTypeNetName() : npc.GetFullNetName());
```
Chicken-Bones: why the `npc.boss` check here? `npc.GetFullNetName` falls back to `npc.GetTypeNetName`

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -1323,6 +1323,24 @@ public static void BuffTownNPC(ref float damageMult, ref int defense)
 			g.BuffTownNPC(ref damageMult, ref defense);
 		}
 	}
+
+	private delegate bool DelegateDeathMessage(NPC npc, ref NetworkText custom, ref Color color);
+	private static HookList HookDeathMessage = AddHook<DelegateDeathMessage>(g => g.DeathMessage);
+
+	public static bool DeathMessage(NPC npc, ref NetworkText customText, ref Color color)
+	{
+		if(npc.ModNPC != null && Language.Exists(npc.ModNPC.GetLocalizationKey("DeathMessage"))) {
```
Chicken-Bones: `DeathMessage` should be a `virtual LocalizedText` property on `ModNPC` just like `DisplayName` is. We can let it return `null` by default. 

After further discussion, jopo and I decided that it's better to get the modder to override the property when they want the localization, rather than have magic localization entries.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
flye-name: What about customizing the message color?

Chicken-Bones: We'll keep this hook for advanced usages, rename it to `ModifyDeathMessage`

flye-name:
Done the changes

Chicken-Bones (approved):

ExampleMod/Localization/zh-Hans.hjson
```diff
@@ -108,6 +108,7 @@ Mods: {
 				// DisplayName: Minion Boss
 				# This is part of ModIntegrationsSystem.cs: Information for the player so he knows how to encounter the boss
 				// BossChecklistIntegration.SpawnInfo: Use a [i:ExampleMod/MinionBossSummonItem]
+				BossFightName: The Minion Boss and his minions in chinese
```
direwolf420: Is this temporary?

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -145,6 +146,7 @@ public sealed override void SetupContent()
 		AutoStaticDefaults();
 		SetStaticDefaults();
 		NPCID.Search.Add(FullName, Type);
+//		_ = DeathMessage;
```
direwolf420: Is this leftover from testing?

JavidPack: Yep, missed that, thanks.

ExampleMod/Localization/zh-Hans.hjson
JavidPack: missed this too.

PR #4632, opened 2025-04-21
Fix #4624

### LINKED ISSUE #4624: SetFactory for DamageClass

Yes

Mod capability as a Modder

### Description

It has been requested that we add a SetFactory for DamageClass, complete with Search and DamageClassID classes

_No response_

_No response_

JavidPack:
Fixes #4624 by adding `DamageClass.Search` and `DamageClass.Sets.Factory`.

Modders should be able to use it to set custom DamageClass metadata.

The implementation does cause all the vanilla DamageClass instances to be reinitialized since they are also static fields of the class, I'm not sure if there is any potential issue with that.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
@@ -28,7 +28,7 @@ public static class DamageClassLoader
 	static DamageClassLoader()
 	{
 		RegisterDefaultClasses();
-		ResizeArrays(fromStaticCtor: true);
+		RebuildEffectInheritanceCache();
 	}
 
 	internal static int Add(DamageClass damageClass)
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/ModLoader/DamageClassLoader.cs
```diff
@@ -26,7 +28,7 @@ public static class DamageClassLoader
 	static DamageClassLoader()
 	{
 		RegisterDefaultClasses();
-		ResizeArrays();
+		ResizeArrays(fromStaticCtor: true);
```
Chicken-Bones: yuck, lets just call `RebuildEffectInheritanceCache()` here instead

direwolf420:
I think this lacks `Search.Add(FullName, Type);` in `SetupContent`

Chicken-Bones (approved):

PR #4640, opened 2025-04-30
Fix incorrect tile update check in Worldgen.TileFrame()

### LINKED ISSUE #4639: Modifying tile frame in 2 axis at once in PostTileFrame() causes unnecessary extra TileFrame() calls that can lead to crashes
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

`PostTileFrame()` was added as a hook to allow modders to modify a tile's frame after vanilla framing code has executed
Unfortunately, the vanilla check to see if a tile's frame has been updated does not account for changes made in `PostTileFrame()`, which means that the game erroneously thinks a tile's frame has been updated (comparing the original frame that was affected by `PostTileFrame()` and the newly calculated vanilla frame), leading to `TileFrame()` being called on adjacent tiles, causing potential endless back & forth reframing loops.
This issue only appears if TileFrameX and TileFrameY are modified at once, as the adjacent tile reframing only happens if both the X and Y coordinates don't match the original ones, so tiles with simple frame edits that only change the frame in 1 dimension are spared from this issue, but any tile with weirder logic or patterns that are 2x2 in size or bigger will be prone to this issue

### Steps to reproduce

1. Make a ModTile that uses `PostTileFrame()` with an offset every other X and Y coordinate 
```
Tile t = Main.tile[i, j];
if (i % 2 == 0) 
	t.TileFrameX += 288;
if (j % 2 == 0) 
	t.TileFrameY += 270;
```
3. Spawn dust in `PostTileFrame()` to easily visualize which tiles are being framed
4. Build a big square out of your modtile
5. Hit any tile that's got no X/Y frame offset 

### Expected Behavior

Dust should appear in a square ring around the hit tile, without propagating the tile framing to any tile that's not directly adjacent to the hit tile

### Actual Behavior

Dust appears on the 8 adjacent tiles around the center one, but when the tiles with both X and Y framing offset get framed (In the diagonal corners from the original one), those will also frame the 4 tiles adjacent to them as well

IbanPlay:
### What is the bug?
As mentionned in #4639, `WorldGen.TileFrame()`'s doesn't account for potential framing modifications that may have happened in `TileLoader.PostTileFrame()` when checking for if the tile's frame has been modified. This leads to it reframing adjacent tiles, and depending on the tile configuration & `PostTileFrame()` logic, can lead to the game freezing and crashing

### How did you fix the bug?
Made `TileFrame()` check the actual tile's frame when comparing them agaisnt the original frame coordinates, instead of using the coordinates that were calculated by the vanilla code ( Which lacks any of the potential `PostTileFrame()` modifications)

This is the most straightforward fix for the issue


patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -3611,6 +3611,15 @@
  								if (TileID.Sets.IsVine[num]) {
  									up = ((tile2 == null) ? num : ((!tile2.nactive()) ? (-1) : ((!tile2.bottomSlope()) ? tile2.type : (-1))));
  									if (num != up) {
+@@ -63276,7 +_,7 @@
+ 								if (!noTileActions && tile.active() && TileID.Sets.Falling[num])
+ 									SpawnFallingBlockProjectile(i, j, tile, tile2, tile3, num);
+ 
+-								if ((rectangle.X != frameX && rectangle.Y != frameY && frameX >= 0 && frameY >= 0) || flag14) {
++								if ((tile.frameX != frameX && tile.frameY != frameY && frameX >= 0 && frameY >= 0) || flag14) {
```
Chicken-Bones, line 3619: For future reference, these kind of changes best come with a comment (ideally one that includes the PR number)

PR #4649, opened 2025-05-11
Update ExampleBar.cs

TerrorPenguin:
Basically just fixes them being able to be supported below by another hammered bar tile, uses the same method vanilla does for the bar tiles.

Needs a comment to explain what it's doing, I omitted it because I'm unsure what other placement scenarios this may affect.


JavidPack:
Thanks, I didn't know bars could be hammered.

TerrorPenguin:
I'd like to note that bars prior to this already broke if the tile below them was hammered, my change as far as I found only prevents them being supported by other _hammered bars_, so I wasn't really sure what to say for the comment, in case it affected other placements.

The comment changes on the slope fields seem nice.

PR #4654, opened 2025-05-17
Fix escape not working while inputting text on platforms other than Windows

harperstoeckle:
Note: This is a vanilla bug.

`WinImm32Ime` receives input characters via `WM_CHAR` messages, which are posted even for control characters (including escape). On the other hand, `FnaIme` receives characters through FNA's `TextInputEXT::TextInput` event, which does *not* include escape.

As a result, `GetInputText` sets `inputTextEscape` when escape is pressed on Windows, but not on other platforms. In the bestiary, for example, pressing escape does not deselect the search bar.

This PR just explicitly checks in `GetInputText` whether the escape key has been pressed. This could also probably be solved by modifying `FnaIme` to detect escape explicitly, but it's not obvious to me the best way to do that.


Chicken-Bones:
I've reported this in the vanilla bug tracker and will update this PR when we've made a decision on the best fix there

PR #4671, opened 2025-06-02
Additional sound fixes, VariantsAreIndependent support

JavidPack:
**Note: This is a PR to another PR, not to the 1.4.4 branch**

This PR addresses a remaining TODO of #2566, "A lot of sounds need `MaxInstances` of 0.". The big ones are Coins, CoinPickup, and ZombieMoan, these were very obvious sound glitches. 

Many item sounds such as FallingStar/Star Cannon, meowmere, gem staffs, and flamethrower cut off the playing sound when triggered again. These have been fixed.

This PR also fixes many ambient sounds that were not marked as ambient.

This PR also found several PitchVariance values which were wrong and fixed them.

This PR also fixes the `PitchRange` getter, it was mistakenly doubling the value. The getter is unused in vanilla code, but it is good to have it corrected.

`Bird15` was removed and `Bird` will no longer play the `Critter` (Zombie_15) sound as a variant option.

`PlayerHit` no longer has `PitchVariance`. I could not find where that came from in the vanilla source code.

`ShimmerWeak1` and `ShimmerWeak2` now share an identifier to prevent them from playing at the same time, as intended.

The big change of this PR is the addition of `SoundStyle.VariantsAreIndependent` to indicate if the variants of a SoundStyle should be considered as separate sounds rather than the same sound. For example, Clown does not use this, only 1 random sound can be playing at a time, but the vanilla logic for Toucan says that each of the random sound variants can be playing at once, as if they were separate sounds.

To implement `VariantsAreIndependent`, the random choice decision had to be moved earlier in the logic to precede the `IsTheSameAs` check. 

`IsTheSameAs` has been overhauled. Previously 2 sounds with different `Identifier`s but the same `SoundPath` would both evaluate to true ("the same"). This logic is incorrect, an `Identifier` should indicate that 2 SoundStyles are separate even if their `SoundPath` matches. This bug would mistakenly prevent sounds using the `SoundPath` of "{Prefix}Zombie_" (but different `Identifier`s) from playing at the same time. Also, the random choice is also taken into account during `IsTheSameAs` if `VariantsAreIndependent` is also true. 2 `VariantsAreIndependent` `SoundStyle` with different random choice shouldn't be seen as "the same", allowing them both to coexist.

Comments have been added for some remaining issues to address, including pausing the game interrupting the MenuOpen sound, Duck having unique pitches per variant, and the Pixie sound not properly updating position and volume as the pixie moves around the game world.

# Discussions

1. These changes technically change some existing behavior for mods, such as `IsTheSameAs` mistakenly seeing 2 sounds with unique `Identifier` as the same, but I think the fixes shouldn't be an issue since the current behavior is buggy and it is rare for a mod to reuse a `SoundPath` like that.

2. It's possible that `VariantsAreIndependent` should default to true. It is a change to existing behavior for existing mods, however. I don't know how many mods are using the variants option, but it might affect some mods. Only `Clown` and `QueenSlime` sounds have variants but aren't unlimited (`MaxInstances = 0`) and only allow a single variant at a time (`VariantsAreIndependent = true`)

### DIFF
--- a/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/Audio/SoundPlayer.cs
 +++ src/tModLoader/Terraria/Audio/SoundPlayer.cs
-@@ -2,24 +_,84 @@
+@@ -2,24 +_,86 @@
  using Microsoft.Xna.Framework;
  using ReLogic.Utilities;
  
@@ -48,8 +48,10 @@
 +	// The part that HAS to run on the main thread.
 +	private SlotId Play_Inner(in SoundStyle style, Vector2? position, SoundUpdateCallback? updateCallback)
 +	{
++		SoundStyle chosenStyle = style.WithChosenRandom();
++
 +		// Handle the MaxInstances & RestartIfPlaying properties
-+		int maxInstances = style.MaxInstances;
++		int maxInstances = chosenStyle.MaxInstances;
 +
 +		if (maxInstances > 0) {
 +			int instanceCount = 0;
@@ -57,11 +59,11 @@
 +			foreach (var pair in _trackedSounds) {
 +				var activeSound = pair.Value;
 +
-+				if (!activeSound.IsPlaying || !style.IsTheSameAs(activeSound.Style) || ++instanceCount < maxInstances) {
++				if (!activeSound.IsPlaying || !chosenStyle.IsTheSameAs(activeSound.Style) || ++instanceCount < maxInstances) {
 +					continue;
 +				}
 +
-+				switch (style.SoundLimitBehavior) {
++				switch (chosenStyle.SoundLimitBehavior) {
 +					case SoundLimitBehavior.ReplaceOldest: //TODO: Make this actually true to its name -- replace the *oldest* sound.
 +						activeSound.Sound?.Stop(true);
 +						break;
@@ -71,10 +73,10 @@
 +			}
 +		}
 +
-+		SoundStyle styleCopy = style;
++		SoundStyle styleCopy = chosenStyle;
 +
 +		// Handle 'UsesMusicPitch'.. This property is a weird solution for keeping vanilla's old instruments' behavior alive, and is currently internal.
-+		if (style.UsesMusicPitch) {
++		if (chosenStyle.UsesMusicPitch) {
 +			styleCopy.Pitch += Main.musicPitch;
 +		}
  
--- a/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
+++ b/patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
@@ -1,6 +1,6 @@
 --- src/TerrariaNetCore/Terraria/Audio/SoundPlayer.cs
 +++ src/tModLoader/Terraria/Audio/SoundPlayer.cs
-@@ -2,24 +_,86 @@
+@@ -2,24 +_,84 @@
  using Microsoft.Xna.Framework;
  using ReLogic.Utilities;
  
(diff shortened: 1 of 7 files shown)


patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -211,8 +222,11 @@ public SoundStyle(string soundPath, ReadOnlySpan<(int variant, float weight)> we
 	// To be optimized, improved.
 	public bool IsTheSameAs(SoundStyle style)
 	{
-		if (Identifier != null && Identifier == style.Identifier)
-			return true;
+		if (VariantsAreIndependent && RandomChoice != -1 && RandomChoice != style.RandomChoice)
```
Chicken-Bones: This check isn't symmetrical. If `a.RandomChoice == -1` but `b.RandomChoice == 1` then `a.IsTheSameAs(b)` may return `true`, but `b.IsTheSameAs(a)` will be `false`

You probably want to remove the `RandomChoice != -1` check. If you want to check if a randomized style is derived from a given base style, then that should be a different method

patches/tModLoader/Terraria/Audio/SoundPlayer.cs.patch
```diff
@@ -48,20 +48,22 @@
 +	// The part that HAS to run on the main thread.
 +	private SlotId Play_Inner(in SoundStyle style, Vector2? position, SoundUpdateCallback? updateCallback)
 +	{
++		SoundStyle chosenStyle = style.WithChosenRandom();
```
Chicken-Bones: Probably want to remove the randomization from `GetRandomSound` call in `ActiveSound.Play` since you're doing the randomization here

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -254,8 +268,20 @@ public SoundStyle WithVolumeScale(float scale)
 	public SoundStyle WithPitchOffset(float offset)
 		=> this with { Pitch = Pitch + offset };
 
+	public SoundStyle WithChosenRandom(int random = -1)
```
Chicken-Bones: Possibly acceptable, but honestly it would be better to rename some methods to move the randomization concept deeper.
- `WithChosenRandom` -> `WithSelectedVariant`
- `GetRandomSound` -> `GetSoundEffect`

Chicken-Bones (changes requested):
I don't think `VariantsAreIndependent` should default to `true` even if that is the most common choice in vanilla, because of the principle of least surprise.

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
JavidPack: GetRandomSound is public, so I don't think we should rename that.

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -254,8 +286,20 @@ public SoundStyle WithVolumeScale(float scale)
...
+		return this with { RandomChoice = random == -1 ? GetRandomVariantIndex() : random };
+	}
+
 	private int GetRandomVariantIndex()
 	{
+		if (this.random != -1) {
+			// Use existing random decision if already made.
+			return this.random;
+		}
+
```
Chicken-Bones: This shouldn't be needed, since it's only called in `WithSelectedVariant`

patches/tModLoader/Terraria/Audio/SoundStyle.TML.cs
```diff
@@ -254,8 +286,20 @@ public SoundStyle WithVolumeScale(float scale)
 	public SoundStyle WithPitchOffset(float offset)
 		=> this with { Pitch = Pitch + offset };
 
+	internal SoundStyle WithSelectedVariant(int random = -1)
+	{
+		if (variants == null || variants.Length == 0)
+			return this;
+		return this with { RandomChoice = random == -1 ? GetRandomVariantIndex() : random };
```
Chicken-Bones: Shall we call `RandomChoice` `Variant`? It can also be an auto-property

Issue #4672, opened 2025-06-03
TmodFile.GetStream Conditional branching order


NanTingPer:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

When I use `TmodFile.GetStream()`, even if `bool newFileStream = true`, I will first perform the judgment `fileStream == null` to determine whether the order should be shifted?
https://github.com/tModLoader/tModLoader/blob/6aec675acbc9d5c4624913d836c2c648d4cc9676/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs#L101-L110
```cs
[ModuleInitializer]
public static void LoadLocalizationJson()
{
    var fileEntry = ThisTModFile.FirstOrDefault(f => f.Name == fileName);
    if (fileEntry is null)
        return "";

   //Erro
    var fileStream = ThisTModFile.GetStream(fileEntry, true);
}
```

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/3a84e434-1a13-451c-bd01-566559d20cf3" />

NanTingPer:
Sorry, this situation doesn't occur under normal circumstances. I forgot that I obtained the modFile using reflection.

Chicken-Bones:
I think that you are trying to access the contents of the tModFile too early. If you really need to do this, you should call `handle = TModFile.Open()` yourself, and make sure to call `handle.Dispose()` when you're done.

Note that `Mod.Close` is `virtual`, so you may want to dispose all your handles there before calling `base.Close`, otherwise it will throw the exception
```cs
		if (File != null && File.IsOpen)
			throw new IOException($"TModFile has open handles: {File.path}");
```

NanTingPer:
thank

PR #4673, opened 2025-06-03
TmodFile.GetStream Conditional branching order

NanTingPer:
### What is the bug?
#4672 

### How did you fix the bug?
Adjust the branch sequence

I used reflection and conducted the test in the same operation mode as within this branch, and it ran normally
```cs
Type type = typeof(ModLoader).Assembly.GetType("Terraria.ModLoader.Core.EntryReadStream")!;
var ersCtor = type.GetConstructor([
    typeof(TmodFile),
    typeof(TmodFile.FileEntry),
    typeof(Stream),
    typeof(bool)
]);

var BIN = BindingFlags.Instance | BindingFlags.NonPublic;
var indep = (IList)typeof(TmodFile).GetField("independentEntryReadStreams", BIN)!.GetValue(ThisTModFile)!;
var fileStream = (Stream)ersCtor!.Invoke([ThisTModFile, fileEntry, File.OpenRead(ThisTModFile.path), false]);
//var fileStream = ThisTModFile.GetStream(fileEntry, true);
indep.Add(fileStream);

string fileContent = GetText(fileStream, fileEntry.IsCompressed) ?? "";
fileStream.Close();

```

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/TmodFile.cs
@@ -98,15 +98,15 @@ public class TmodFile : IEnumerable<TmodFile.FileEntry>
 		if (entry.cachedBytes != null) {
 			stream = entry.cachedBytes.ToMemoryStream();
 		}
-		else if (fileStream == null) {
-			throw new IOException($"File not open: {path}");
-		}
-		else if (newFileStream) {
+  		else if (newFileStream) {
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Your change would break `TModFile.IsOpen` and the contract, contents can only be accessed while the file is 'open'.

You can see more of this in `TModFile.Close` where we have:
```cs
if (independentEntryReadStreams.Count != 0)
	throw new IOException($"Shared entry read streams not closed: {string.Join(", ", independentEntryReadStreams.Select(e => e.Name))}");
```

Issue #4679, opened 2025-06-08
Boss Bar


zlzhaidou:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Add a method to calculate the boss max health and health now

### What does this proposal attempt to solve or improve?

Now the boss bar style example won't calculate the EOW 's health and ML
And the vanilla method to calculate health won't works in even mod
So could you add a method to do it🥺

PR #4682, opened 2025-06-08
Added GetBossHealth method to NPCUtils

mt1nr:
Calculates the total current and maximum health of a boss, including all its segments if applicable.

### What is the new feature?
A utility method that calculates the total current and maximum health of a boss, including all its segments if applicable.

An issue has been opened: [#4679](https://github.com/tModLoader/tModLoader/issues/4679) requesting such a method.

Modders could implement their own health summation logic per boss, but this leads to duplicated code and inconsistent implementations.

```
NPCUtils.GetBossHealth(npc, out int current, out int max);
Main.NewText($"Boss: {npc.TypeName} | HP: {current}/{max} ({(int)((float)current / max * 100)}%)", 255, 255, 0);
```

### DIFF
--- a/patches/tModLoader/Terraria/Utilities/NPCUtils.cs.patch
+++ b/patches/tModLoader/Terraria/Utilities/NPCUtils.cs.patch
@@ -1,0 +1,38 @@
+--- src/TerrariaNetCore/Terraria/Utilities/NPCUtils.cs
++++ src/tModLoader/Terraria/Utilities/NPCUtils.cs
+@@ -273,4 +_,35 @@
+ 				searcher.FaceTarget();
+ 		}
+ 	}
++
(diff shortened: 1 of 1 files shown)


direwolf420:
The issue you linked to mentions EOW (Eater of Worlds), for which this approach will not work, because of its unique splitting behavior.

mt1nr:
I have developed a new approach for the EOW and other segmented bosses from other mods. I am currently testing it but would this be a solution?

emyhrberg:
This would be really useful. I had to do a lot of manual work to make my DPS mod work, especially for worm segments like EoW, for example: https://github.com/emyhrberg/DPSPanel/blob/master/DamageCalculation/Classes/EoWFight.cs 
I could imagine other DPS mods like DPSExtreme could make very good use of this utility method too.

PR #4684, opened 2025-06-09
Fixes #3633: Negative NPCIDs now compatible with NPCDefinition

### LINKED ISSUE #3633: Use of NPCDefinition in ModConfig does not support negative NPC types
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Mod capability as a Modder

### Description

When using NPCDefinition in ModConfig, if the ID of the NPC is negative, the slot will show an item rather than an NPC and will also show it as unloaded. However, the correct name of the NPC will appear when mousing over the slot. Furthermore you cannot select the negative ids when opening the list to change the NPC in the slot since the list starts at ID 0.

Capture

### Log File

no log needed

### Steps to reproduce

1. Create a new NPCDefinition with any negative NPC type like Pinky.
2. Save the definition to the config. 
3. Open the mod's config through the menu and check the slot.

### Expected Behavior

Negative NPC IDs should work just fine

### Actual Behavior

Negative NPC IDs will show an item rather than an NPC and will also show as unloaded

### Reproduction frequency

100%

### Additional Information

I think having a netID field in NPCDefinition would be beneficial for npcs like slimes to accurately represent which slime type they are (jungle slime, yellow slime, blue slime, red slime)

Vinixss:
Fixes #3633

### What is the bug?
Negative NPCIDs, like those of Pinky, would not register as loaded under `NPCDefinition` when checking a mods config.

### How did you fix the bug?
I expanded the size of for loop that creates the options list in NPCDefinitionElements. I changed the declaration of type in DrawSelf() so that it takes in the netID of the NPC. Finally, for NPCDefinition I overloaded IsUnloaded so that it no longer views any ID less than zero to be unloaded

Most likely. I have little experience with C# and this is my first time contributing so there may be a few issues. One part I am unhappy with is how I overloaded IsUnloaded by using the NPCID.Search.TryGetId() method in place of Type < 0;

# Porting Notes

- Due to some negative values now being valid for `NPCDefinition.Type`, the `NPCDefinition.Type` for an "unloaded" `NPCDefinition` has changed from `-1` to `-66`. If you have code checking `NPCDefinition.Type == -1`, it would be better to check `NPCDefinition.IsUnloaded` instead.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Config/EntityDefinition.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Config/EntityDefinition.cs
@@ -147,10 +147,12 @@ public class NPCDefinition : EntityDefinition
 	public static readonly Func<TagCompound, NPCDefinition> DESERIALIZER = Load;
 
 	public override bool IsUnloaded
-		=> !(Mod == "Terraria" && Name == "None" || Mod == "" && Name == "") && !(NPCID.Search.TryGetId(Mod != "Terraria" ? $"{Mod}/{Name}" : Name, out int id));
+		=> Type <= NPCID.NegativeIDCount && !(Mod == "Terraria" && Name == "None" || Mod == "" && Name == "");
 
-	// TODO: doesn't handle negative I think?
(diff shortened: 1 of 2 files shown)


JavidPack:
Thanks, you did great.

PR #4686, opened 2025-06-14
Add hook for when a tile is being replaced via block swap

### LINKED ISSUE #4674: Block Swap success only runs GetItemDrops (and hooks run on failure), not KillTile or an alternative hook
### Version

Latest 1.4.4-stable

### Steps to reproduce

1. Make a tile that overrides KillTile
2. Break it with block swap

### Expected Behavior

KillTile — or a hook specifically for block swap — runs.

### Actual Behavior

KillTile does not run, and there isn't a hook specifically for block swap.

efcawesome:
### What is the bug?
As per #4674, there is no hook to act upon a tile being replaced by the block swap feature. Fixes #4674 

### How did you fix the bug?

A `ReplaceTile` hook has been added.

Other options were explored, but would lead to inconsistent behavior. A new hook was needed.

#### Side Note: 
This is my first contribution, so I don't really know what I am doing, and I feel like I am creating a solution that might change some intended functionality --- please tell me if I did anything wrong and what I can fix.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
+++ b/patches/tModLoader/Terraria/ModLoader/GlobalTile.cs
@@ -344,6 +344,11 @@ public abstract class GlobalTile : GlobalBlockType
 		return true;
 	}
 
+	/// <inheritdoc cref="ModTile.ReplaceTile(int, int, int, int)"/>
+	public virtual void ReplaceTile(int i, int j, int type, int targetType, int targetStyle)
+	{
+	}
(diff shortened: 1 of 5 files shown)


JavidPack:
Thanks for getting this started. I changed the PR to do a new hook rather than attempt to reuse the existing hook because we don't want mods unexpectedly behaving differently. It's possible KillTile code in existing mods might do something incorrect if a tile were to be placed at the same location.

PR #4706, opened 2025-07-09
Better portable storage support (Port of #1978)

JavidPack:
**Note: I was unable to get Git to open the original branch to resolve conflicts in the original PR. This PR is a port of #1978. The initial work was all done by @pbone64 .**

### What is the new feature?
Added better portable storage support, by introducing a tracker (similar to the void lens and piggy banks) for Safes and Defenders Forges, adding a set that lets you count a projectile as a portable storage projectile, and added the `BankID` class to easily reference special bank ids.

1.4.3 greatly improved the portable storage system, but it still wasn't at the point where mods could easily add their own without IL editing. This saves mods from multiple annoying IL edits.

Yes, probably. The way sounds are handled could be changed to deterministic, however I can't think of a good way to do that right now

Here is an example of a portable safe projectile:
```
public class Awa : ModProjectile
{
    {
        ProjectileID.Sets.CountAsBank[Projectile.type] = BankID.Safe;
    }

    public bool init = false;

    public override void AI()
    {
        if (!init)
        {
            Player localPlayer = Main.player[Projectile.owner];

            localPlayer.chest = BankID.Safe;
            for (int i = 0; i < 40; i++)
            {
                ItemSlot.SetGlow(i, -1f, chest: true);
            }

            localPlayer.safeProjTracker.Set(Projectile);
            localPlayer.chestX = (int)Projectile.Center.X;
            localPlayer.chestY = (int)Projectile.Center.Y;
            localPlayer.SetTalkNPC(-1);
            Main.SetNPCShopIndex(0);
            Main.playerInventory = true;
            SoundEngine.PlaySound(SoundID.MenuOpen);
            Recipe.FindRecipes();

            init = true;
        }
    }

    public override void PlayBankCloseSound()
    {
        SoundEngine.PlaySound(SoundID.MenuClose);
    }
}
```
### ExampleMod updates
I did not update ExampleMod.

Side note: For some reason, an old commit was included in this. The old commit doesn't actually change anything, and I'm not sure why GitHub included it in this PR.

### DIFF
--- a/ExampleMod/Content/Tiles/Furniture/ExampleDresser.cs
+++ b/ExampleMod/Content/Tiles/Furniture/ExampleDresser.cs
@@ -98,8 +98,7 @@ namespace ExampleMod.Content.Tiles.Furniture
 					}
 				}
 				else {
-					player.piggyBankProjTracker.Clear();
-					player.voidLensChest.Clear();
+					player.ClearPortableBankProjectileTrackers();
 					int chestIndex = Chest.FindChest(left, top);
 					if (chestIndex != -1) {
 						Main.stackSplit = 600;
--- a/patches/tModLoader/Terraria/ID/BankID.cs
+++ b/patches/tModLoader/Terraria/ID/BankID.cs
@@ -1,0 +1,12 @@
+﻿namespace Terraria.ID;
+
+/// <summary>
+/// Values used to identify special banks with the <see cref="Player.chest"></see> field
+/// </summary>
+public static class BankID
+{
+	public const int PiggyBank = -2;
+	public const int Safe = -3;
+	public const int DefendersForge = -4;
+	public const int VoidVault = 5;
+}
--- a/patches/tModLoader/Terraria/ID/BankID.cs
+++ b/patches/tModLoader/Terraria/ID/BankID.cs
@@ -1,12 +1,17 @@
 ﻿namespace Terraria.ID;
 
 /// <summary>
-/// Values used to identify special banks with the <see cref="Player.chest"></see> field
+/// Values used to identify special banks ("personal storage") with the <see cref="Player.chest"></see> field.
 /// </summary>
 public static class BankID
 {
+	public const int None = -1;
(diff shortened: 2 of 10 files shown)


pbone64:
Hi! While what I implemented certainly works, I'm not really convinced that it's a worthwhile feature to add to tML because it's so niche. It only really makes sense if tML also adds support for _adding_ per-player storages, because you'd be tweaking all of this logic anyways. I think that I only submitted this patch forever ago because I was encouraged to PR tML instead of patching it mod side for compatability reasons, but it really doesn't seem worth it in retrospect haha.


FWIW: since the initial PR, I rewrote my mod-side system into a semi-grotesque ModType wrapper/adaption of portable storages: https://github.com/pbone64/MorePortableStorages

pbone64:
Looking back at my old code, it might actually be possible to rejig it into something that also lets you register more per player storages and PR that. 

Judging from this PR re-opening I'm guessing that there's some interest in a portable storage API - and if one is merged, I'd rather that we do it right instead of dusting off a crusty old patch.

One possibility for a re-worked API: instead of being an all-encompassing portable storages API that handles storing the inventory and tracking logic for projectiles, it could be condensed into a simpler extension to the chest API: i.e. add a way to open arbitrary chest objects that are considered to be at arbitrary positions. That way, a chest object could then be stored on a ModPlayer and it could opened via the new method when the projectile is interacted with via the new projectile interaction API. I'd have to review the vanilla chest system before finalizing any details.

Tyfyter:
There's at least one mod with a use case for adding non-storage "bank IDs", too

JavidPack:
I guess it is true that your mod covers what modders would want to do with this PR, an option for those 2 player storages. 

Maybe this would need to support tml managed player chests to be useful, but it does seem quite involved looking at all the places the player chests are referenced. 

I think I'll put this PR on hold unless there is a clear demand for custom player storages, and I'm not seeing any requests on github right now. If you want to explore and see what it will take, feel free to do that, but I think I'll work reviewing other PRs for now.

PR #4709, opened 2025-07-10
Conversion Fallback/Inheritance

Tyfyter:
### What is the new feature?
Tiles and walls can specify a 'conversion fallback tile', a simpler and more powerful version of `Tile/WallID.Sets.Conversion` in vanilla. When a modded conversion attempts to convert a tile but there is no registered conversion, the tile will be converted as if it was the fallback.

This is handled by calling `Tile/WallLoader.RegisterConversionFallback` with the tile/wall type, the tile/wall type it should inherit from, and any conversion types to exclude. Tiles and walls should exclude the conversion type which creates the tile, to avoid unnecessary redundant reconversion (eg, `Ebonstone` would exclude `Corruption` conversion from the `Stone` fallback).

This PR also introduces a helper method, `Tile/WallLoader.RegisterSimpleConversion` for tiles which simply change from one type to another with a given conversion id, saving the need to write methods or delegates which just call `WorldGen.Convert`.
For example, `RegisterSimpleConversion(Ebonstone, Corruption, Stone)` will also register `Stone` as a fallback for `Ebonstone`, and register a purification/chlorophyte conversion from `Ebonstone` to `Stone`.

This means that overriding `ModTile/Wall.Convert`  is now only required for advanced use cases where `RegisterSimpleConversion` doesn't suffice.

### Sample Usage

```cs
public class MyCustomStoneTile : ModTile
{
		TileLoader.RegisterSimpleConversion(TileID.Stone, ModContent.GetInstance<MyCustomSolutionConversion>().Type, Type);
	}
}
```
```cs
public class MyCustomSolutionConversion : ModBiomeConversion
{
	public override void PostSetupContent() {
		int SandType = ModContent.TileType<MyCustomSandTile>();
		TileLoader.RegisterConversion(TileID.Sand, Type, (i, j, _, _) => …);
		TileLoader.RegisterConversionFallback(SandType, TileID.Sand, Type);
	}
}
```

With this, mods don't need to manually add compatibility with other mod's conversions if their tile is a variant of a vanilla tile.

It also means that mods which add a new stone/sandstone/grass variant etc, don't need to manually add conversion code for crimson/corruption/hallow solutions if they specify a fallback tile which already has such a vanilla conversion.

### ExampleMod updates
Replaced explicit corruption and crimson conversions in desert fossil tiles with conversion fallback.
Updated `ExampleSolutionConversion` and `ExampleHellSolutionConversion` to use fallbacks.

### Porting Notes
Use `Tile/WallLoader.RegisterSimpleConversion` for conversions which don't need any additional logic beyond converting the tile/wall.
Use `Tile/WallLoader.RegisterConversionFallback` for conversions which do need special logic but should still be treated as a pure counterpart for other conversions.
Remove explicit conversion code when the fallback conversion would produce the same result.


Tyfyter:
Marked as draft because discussion on whether or not to completely replace vanilla all conversions is needed.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -718,6 +721,65 @@ public static void RegisterConversion(int tileType, int conversionType, ConvertT
...
+		RegisterConversionFallback(TileID.HallowedGrass, TileID.Grass, BiomeConversionID.Hallow);
+
+		RegisterConversionFallback(TileID.GolfGrassHallowed, TileID.GolfGrass, BiomeConversionID.Hallow);
+		RegisterConversionFallback(TileID.GolfGrass, TileID.Grass, BiomeConversionID.Purity, BiomeConversionID.PurificationPowder, BiomeConversionID.Dirt);
+
+		// removed to preserve vanilla oversight requiring 2 solutions to convert between evil and mushroom
+		// please let me fix this
+		//RegisterConversionFallback(TileID.CorruptJungleGrass, TileID.JungleGrass, BiomeConversionID.Corruption);
+		//RegisterConversionFallback(TileID.CrimsonJungleGrass, TileID.JungleGrass, BiomeConversionID.Crimson);
+		//RegisterConversionFallback(TileID.MushroomGrass, TileID.JungleGrass, BiomeConversionID.GlowingMushroom);
```
Chicken-Bones: Could you just add the mushroom and evil solutions as exceptions to these?

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -718,6 +721,65 @@ public static void RegisterConversion(int tileType, int conversionType, ConvertT
...
+	public static void RegisterConversionFallback(int tileType, int fallbackType, params int[] exceptForConversionTypes)
+	{
+		if (tileConversionFallbacks == null)
+			throw new Exception(Language.GetTextValue("tModLoader.LoadErrorCallDuringLoad", "TileLoader.RegisterConversionFallback"));
+
+		bool[] exceptFor = new bool[BiomeConversionLoader.BiomeConversionCount];
+		for (int i = 0; i < exceptForConversionTypes.Length; i++) {
+			exceptFor[exceptForConversionTypes[i]] = true;
+		}
+		tileConversionFallbacks[tileType] = (fallbackType, exceptFor);
```
Chicken-Bones: conversion fallbacks may as well be an `int[]` which we fill with `fallbackType` and then set `-1` in each of the `exceptFor` slots

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -46,6 +46,7 @@ public static class TileLoader
 	internal static readonly Dictionary<(int, int), int> tileTypeAndTileStyleToItemType = new();
 	public delegate bool ConvertTile(int i, int j, int type, int conversionType);
 	internal static List<ConvertTile>[][] tileConversionDelegates = null;
+	internal static (int fallbackTile, bool[] exceptConversionIds)[] tileConversionFallbacks = null;
```
Chicken-Bones: We should expose a method which lets people lookup the fallback for a conversion and tile id, as well as one that lets them set it directly for certain conversion types (not that I can think of much use case for that, but may as well)

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
Tyfyter: Oh right, I forgot that this approach lets me not fix the vanilla behavior while still registering them.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
Tyfyter: This would improve the versatility of the system a bit, but how would we handle `RegisterConversionFallback` being called twice with the same tile type with this approach?
It seems unlikely enough to me that I think just throwing an exception should be fine, especially with methods to query fallback types and set them individually, but I'd prefer to get a second opinion in case I'm missing something.

Chicken-Bones: I'm happy for repeat calls to just override. People aren't likely to be registering conversion fallbacks for tiles other than their own, and by exposing direct access to the set, people can remove or customize fallbacks in more detail if that's what they really want for some reason

Tyfyter: Good point

PR #4714, opened 2025-07-14
Better Liquid Slope Rendering

steviegt6:
*Development thread:* https://canary.discord.com/channels/103110554649894912/1394418449752789344/1394418449752789344

### What is the new feature?
Liquid touching tiles, slopes and half bricks has been integrated into the standard liquid renderer, rather than being hacked into tile drawing. This means:
- Liquid edges now have proper surfaces, framing and animations
- Liquid edges now have ripple and distortion shaders applied
- Liquid edges now participate in water style color changes perfectly
- `LiquidSlope` textures are now entirely unused (outside of legacy lighting)
- Tiles with partial transparency (thin ice in vanilla) no longer have bits of liquid drawing behind them. 

dotnet_bkMV9K2Wkj

The visual issues bother many modders and artists, `LiquidSlope` textures are limiting and unwieldly, and mod based solutions have compatibility issues. 

### How does it work?

`LiquidRenderer.LiquidCache` has been updated to treat tiles which are adjacent to liquid (edge tiles) as if they contain liquid in most respects. In the first loop of `LiquidRenderer.InternalPrepareDraw`, `LiquidEdgeRenderer.CollectEdgeData` assigns a liquid type and level, as well as a frame based on liquid in the surrounding tiles. With a few minor adjustments, the standard vanilla liquid processing loops can run on the fake liquid in edge tiles as well.

The partial transparency support is achieved by drawing the `tileTarget` render texture into the `waterTarget` as a scissor mask to clear any pixels where the `tileTarget` has an alpha > 0. For tiles with fully transparent inside pixels (like glass in vanilla) which use `BlocksWaterDrawingBehindSelf`, we draw an extra mask texture with the appropriate framing to remove liquid from inside them.

The separate `_drawCacheForShimmer` has been removed in favour of adding `X, Y` fields to the standard `LiquidDrawCache`. `DrawShimmer` still exists separate to `DrawNormalLiquids`, but the underlying draw data is unified.

`TileDrawing.DrawLiquidBehindTiles` and `DrawTile_LiquidBehindTile` are no longer used when the new renderer is enabled.

The old render paths remain for compatibility with lighting modes which don't use render targets, and a new `config.json` setting (`LiquidSlopeFix`) has been added just in case there's any reason to revert to the old renderer.

Development went through several other ideas:
- Rendering liquid into a separate render target and using shaders with the tileTarget as a parameter to perform the masking
  - Discovered we could to the same directly in the liquid target with blend modes and no custom shader
- Keeping track of edges in a separate data structure and calling a version of `DrawTile_LiquidBehindTile` into the liquid target
  - Performance was worse, and there were still visual discrepancies compared to the standard `LiquidCache` based computations. Integrating with `LiquidCache` was a lot more challenging for the developers, but the resulting patches ended up quite small and most of the logic is separate and maintainable.
- Custom liquid masks for irregularly shaped tiles
  - It's simpler, faster and more flexible to change transparent pixels to almost tranparent (1/255 alpha), than having separate mask textures. Human eyes can't tell the difference

A new tile `ExampleTransparentShapedTile` has been added to showcase the potential of the new system. By adding pixels with an alpha of `1/255` we can simulate full transparency while excluding liquid
image

### Porting Notes

This change should be fairly stable. Mods which were tampering with liquid rendering in an attempt to fix some of these issues may now be broken (and hopefully unnecessary).


patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -184,14 +184,15 @@
...
-@@ -396,8 +_,10 @@
+@@ -396,8 +_,11 @@
  		byte b = (byte)(100f + 150f * Main.martianLight);
  		_martianGlow = new Color(b, b, b, 0);
  		TileDrawInfo value = _currentTileDrawInfo.Value;
 -		for (int i = firstTileY; i < lastTileY + 4; i++) {
 -			for (int j = firstTileX - 2; j < lastTileX + 2; j++) {
 +
 +		//TML: Swapped loop iteration order to improve memory access linearity. Minor perf improvement, may cause slight visual discrepancies with tiles that overdraw
++		bool liquidSlopeFix = LiquidEdgeRenderer.Active;
```
Chicken-Bones: Nitpick: this line should go above comment about loop order (line below `TileDrawInfo value = _currentTileDrawInfo.Value`)

patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -204,7 +205,7 @@
 -					if (solidLayer)
 +					// TML(liquid-slopes): Behind-tile liquid rendering is now
 +					// handled as part of actual liquid rendering.
-+					if (solidLayer && !LiquidEdgeRenderer.Active)
++					if (solidLayer && !liquidSlopeFix)
```
Chicken-Bones: Comment good, but it's unclear whether the `solidLayer` condition was added by tML or not, so it's best to write in a way which shows what the vanilla code was

Either:
```cs
/*
if (solidLayer)
*/
if (solidLayer && !liquidSlopeFix)
```

or

```cs
if (liquidSlopeFix) {} else
if (solidLayer)
```

patches/tModLoader/Terraria/GameContent/Drawing/TileDrawing.cs.patch
```diff
@@ -438,12 +439,20 @@
...
+ 		Tile tile3 = Main.tile[tileX, tileY - 1];
 @@ -2649,8 +_,12 @@
  			Main.tile[tileX, tileY + 1] = tile4;
  		}
  
 -		if (!tileCache.active() || tileCache.inActive() || _tileSolidTop[tileCache.type] || (tileCache.halfBrick() && (tile2.liquid > 160 || tile.liquid > 160) && Main.instance.waterfallManager.CheckForWaterfall(tileX, tileY)) || (TileID.Sets.BlocksWaterDrawingBehindSelf[tileCache.type] && tileCache.slope() == 0))
-+		if (!tileCache.active() || tileCache.inActive() || _tileSolidTop[tileCache.type] || (tileCache.halfBrick() && (tile2.liquid > 160 || tile.liquid > 160) && Main.instance.waterfallManager.CheckForWaterfall(tileX, tileY)) || (TileID.Sets.BlocksWaterDrawingBehindSelf[tileCache.type] && tileCache.slope() == 0 && (!tileCache.halfBrick() || !LiquidEdgeRenderer.Active)))
++		if (!tileCache.active() || tileCache.inActive() || _tileSolidTop[tileCache.type] || (tileCache.halfBrick() && (tile2.liquid > 160 || tile.liquid > 160) && Main.instance.waterfallManager.CheckForWaterfall(tileX, tileY)) || (TileID.Sets.BlocksWaterDrawingBehindSelf[tileCache.type] && tileCache.slope() == 0 && (!tileCache.halfBrick() || !liquidSlopeFix)))
```
Chicken-Bones: Same comment here, unfortunately I don't think there's any fun hack so just keep the original condition commented so we can diff the lines visually. Maybe add a comment as to why we're no longer bailing on `halfBrick` tiles

patches/tModLoader/Terraria/GameContent/Liquid/LiquidMask.hlsl
```diff
@@ -0,0 +1,39 @@
...
+texture uMaskTexture;
+sampler2D maskTexture = sampler_state
+{
+    texture = <uMaskTexture>;
+    magfilter = POINT;
+    minfilter = POINT;
+    mipfilter = POINT;
+    AddressU = wrap;
+    AddressV = wrap;
+};
```
Chicken-Bones: do these actually work? I thought we had to set sampler states on the `GraphicsDevice`

patches/tModLoader/Terraria/GameContent/Liquid/LiquidMask.hlsl
```diff
@@ -1,18 +1,9 @@
...
     float alpha = (tex2D(uImage0, uv) * drawColor).a;
-    if (usePartialAlpha)
-    {
-        return (alpha > 0. && alpha < 1.) ? 1. : 0;
-    }
-    else
-    {
-        return (alpha > 0.) ? 1. : alpha;
-    }
+    return (alpha > 0.) ? 1. : alpha;
```
Chicken-Bones: this would read clearer as `alpha > 0 ? 1 : 0`, since `alpha` can never be less than `0`

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -0,0 +1,377 @@
...
+	public static Texture2D MaskTile => (maskTileAsset ??= ModLoader.ModLoader.ManifestAssets.Request<Texture2D>("Terraria.GameContent.Liquid.DefaultTileLiquidMask", AssetRequestMode.ImmediateLoad)).Value;
+
+	private static Asset<Texture2D>? maskTileAsset;
+
+	/// <summary>
+	/// Contains liquid mask textures for specific block types that have funny shapes but still need to hide water.
+	/// <br />
+	/// Only shows up when the tile is part of the <see cref="TileID.Sets.BlocksWaterDrawingBehindSelf"/> set.
+	/// </summary>
+	public static Dictionary<int, Asset<Texture2D>> CustomTileMasks = new Dictionary<int, Asset<Texture2D>>();
```
Chicken-Bones: This should be an `Asset<Texture2D>[]`, and reinitialized in `TileLoader`, also please add an example tile which sets a funky liquid mask :)

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -0,0 +1,377 @@
...
+		AlphaSourceBlend = Blend.Zero,
+		ColorDestinationBlend = Blend.InverseSourceAlpha,
+		AlphaDestinationBlend = Blend.InverseSourceAlpha
+	};
+
+	/// <summary>
+	/// Tiles which mask rendered liquid (tiles on the edge of bodies of
+	/// liquid).
+	/// </summary>
+	public static HashSet<Point> Edges { get; } = [];
```
Chicken-Bones: `List` now

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -223,13 +223,27 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
...
+		}
+		else if (slope is SlopeType.SlopeUpRight or SlopeType.SlopeDownRight) {
+			noLiquidInDiagonals = tileUpLeftCache.LiquidAmount <= 0;
+		}
+		else {
+			noLiquidInDiagonals = tileUpLeftCache.LiquidAmount <= 0 && tileUpRightCache.LiquidAmount <= 0;
+		}
+
+		// If air or a top slope is above itself, like half bricks next to covered full liquids
+		bool airAbove = (tileUpCache.Slope is SlopeType.SlopeUpLeft or SlopeType.SlopeUpRight || !(tileUpCache.HasTile && Main.tileSolid[tileUpCache.TileType]));
```
Chicken-Bones: should all your `tileSolid` checks be replaced with `WorldGen.SolidTile`?

steviegt6:
🎉🎉🎉🎉

aleislost:
This is such a freaking cool additions. I have to ask, how can a texture pack with slightly lighter and more transparent liquid be done?

PR #4715, opened 2025-07-14
`icon_small` generates by default when creating mods and shows up in the configs list besides the mod's name

flye-name:
It encourages more modders to create an `icon_small.png` for their mod, its only current use is the bestiary filter but certain technical mods may benefit from representing other mods as their small icon, a popular example that does this is DragonLens. 

The icon showing up in the config list helps with readability as it can get hard to find a specific mod when you have many active.

Note that the icon is only used and displayed if it has been changed from the default template. This means that the presence of a icon_small.png in a mod is still an intentional decision by the modder and not a surprise. 

The [wiki ](https://github.com/tModLoader/tModLoader/wiki/Basic-tModLoader-Modding-Guide)has also been updated:
image

Yes but this was the simplest way to do it. 

image

# Discussion
Also, is there any reason for mods without configs to show up in the list at all?


absoluteAquarian:
> Also, is there any reason for mods without configs to show up in the list at all?

This used to not be the case.  However, many users thought that since a mod didn't end up in the list, it wasn't enabled; they treated the menu as a "mod list" rather than the actual _Manage Mods_ list.

Hence, now all mods appear in the configs list even if they don't have a config.

JavidPack:
I feel like the template icon_small.png deviates from the template icon.png. Can you rework it in the style of the existing icon.png? Probably just the letter "M" (white letter, purple outline) in front of the dark purple/pink checkerboard would work.

Tyfyter:
> > Also, is there any reason for mods without configs to show up in the list at all?
> 
> This used to not be the case. However, many users thought that since a mod didn't end up in the list, it wasn't enabled; they treated the menu as a "mod list" rather than the actual _Manage Mods_ list.
> 
> Hence, now all mods appear in the configs list even if they don't have a config.

A setting for it would be a really nice addition, so that people who _do_ understand what the menu is don't have to install a mod to get rid of what they'd understand is clutter if they read the button they just clicked (assuming the non-English translations for the mod configuration menu aren't just really bad)

JavidPack:
> > > Also, is there any reason for mods without configs to show up in the list at all?
> > 
> > 
> > This used to not be the case. However, many users thought that since a mod didn't end up in the list, it wasn't enabled; they treated the menu as a "mod list" rather than the actual _Manage Mods_ list.
> > Hence, now all mods appear in the configs list even if they don't have a config.
> 
> A setting for it would be a really nice addition, so that people who _do_ understand what the menu is don't have to install a mod to get rid of what they'd understand is clutter if they read the button they just clicked (assuming the non-English translations for the mod configuration menu aren't just really bad)

A show all mods toggle in `UIModConfigList` could probably be useful. I think given the earlier confusion we'd default it to show all still. The current coloring and hover tooltips are clear enough I think, but you can make an issue and we can see if it gets upvotes or support from other users.

PR #4716, opened 2025-07-15
New drawing-related hooks for ModMenus

flye-name:
### What is the new feature?
Adds `PreDrawSky()`, `PostDrawSky()`, `PreDrawBackground()`, `PostDrawBackground()`, and `ModifyDrawSky()` to `ModMenu`

This allows modders to more freely customize how their mod's menu looks like without resorting to hacky solutions.

I don't think so

```cs
public override void PostDrawSky(SpriteBatch spriteBatch)
{
    // Exact same code you'd put in a CustomSky.Draw()
}
```
```cs
public override void PostDrawBackground(SpriteBatch spriteBatch)
{
    // Draw a slightly transparent color across the screen as a tint without affecting the logo and menu buttons.
}
```

### ExampleMod updates
Added the following snippet to `ExampleModMenu`:
```cs
// ModifyDrawSky lets you modify some parameters in the menu such as whether or not the sun and moon should draw.
public override void ModifyDrawSky(ref SkyDrawParams drawParams) {
	drawParams.DrawStars = false; // Prevents the stars in the background from drawing.

	drawParams.CloudAlpha = 0.2f; // Set the cloud opacity to 20%
}
```
Showcasing the others would be good too but I think a new `ExampleAdvancedMenu` would be better for that.


Destructor-Ben:
It would be nice if this could also fix an issue with mod menus where a background image is drawn in PreDrawLogo, which stops the credits from playing since they are part of the sky (which gets drawn over). Calamity does this: [CalamityMainMenu](https://github.com/CalamityTeam/CalamityModPublic/blob/1.4.4/MainMenu/CalamityMainMenu.cs) on line 55.
image

flye-name:
> It would be nice if this could also fix an issue with mod menus where a background image is drawn in PreDrawLogo, which stops the credits from playing since they are part of the sky (which gets drawn over). Calamity does this: [CalamityMainMenu](https://github.com/CalamityTeam/CalamityModPublic/blob/1.4.4/MainMenu/CalamityMainMenu.cs) on line 55. 

Ideally these new hooks would replace that, I'll look into this though

flye-name:
Oh, the credits are a `CustomSky`, I completely missed that in your comment

Destructor-Ben:
I think there would need to be a hook that runs after the backgrounds/sky is drawn, but before the credits. If that can't be done, then a hook like PreDraw credits with options to disable the drawing of anything that might come after (perhaps backgrounds).

flye-name:
`PostDrawBackground` should cover that

patches/tModLoader/Terraria/ModLoader/ModMenu.cs
```diff
@@ -94,4 +96,36 @@ public virtual void Update(bool isOnTitleScreen)
 	public virtual void PostDrawLogo(SpriteBatch spriteBatch, Vector2 logoDrawCenter, float logoRotation, float logoScale, Color drawColor)
 	{
 	}
+
+	/// <summary>
+	/// Called just before the sky is drawn, and allows you to modify some of the parameters of the sky draw code.
+	/// <br>Return false to prevent the usual sky from drawing.</br>
+	/// </summary>
+	public virtual bool PreDrawSky(SpriteBatch spriteBatch, ref bool drawStars, ref bool drawSunAndMoon)
+	{
```
Mirsario, line 112: Create a `SkyDrawParams` struct for `drawStars` and `drawSunAndMoon`, similar to `BossBarDrawParams`, so that future upgrades to this hook do not change its signature.

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6701,6 +6710,57 @@
...
+@@ -51113,7 +_,11 @@
+ 		sceneArea.totalWidth = screenWidth;
+ 		sceneArea.SceneLocalScreenPositionOffset = Vector2.Zero;
+ 		SceneArea sceneArea2 = sceneArea;
++		bool drawStars = true;
++		bool drawSunAndMoon = true;
++		bool drawSky = gameMenu ? MenuLoader.CurrentMenu.PreDrawSky(spriteBatch, ref drawStars, ref drawSunAndMoon) : true;
+ 		if (!mapFullscreen) {
++			if (drawSky) {
+ 			unityMouseOver = false;
```
Mirsario, line 6728: This will get accidentally auto-formatted. Use goto statements to jump over what you have defined as a block.

Mirsario (changes requested):

Tyfyter:
Could something be added to this to allow shaders? The vanilla game already has support for it, but that exclusively applies to the ``"Sepia"`` filter, so IL editing is required in order to actually use one in a ``ModMenu``.

Chicken-Bones:
You'll have to specify specifically which points you want to be able to apply a shader, and we can see about it

Tyfyter:
A shader as in a screen shader, you can already use other kinds of shaders exactly how you would anywhere else.

flye-name:
the autoformatting wasn't there locally ???

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6701,6 +6710,57 @@
...
++		if (gameMenu)
++			MenuLoader.CurrentMenu.PostDrawSky(spriteBatch);
+ 		spriteBatch.End();
+ 		Vector2 vector = screenPosition;
+ 		int num8 = screenWidth;
+@@ -51162,7 +_,10 @@
+ 		transformationMatrix.Translation -= BackgroundViewMatrix.ZoomMatrix.Translation * new Vector3(1f, BackgroundViewMatrix.Effects.HasFlag(SpriteEffects.FlipVertically) ? (-1f) : 1f, 1f);
+ 		spriteBatch.Begin(SpriteSortMode.Deferred, BlendState.AlphaBlend, SamplerState.LinearClamp, DepthStencilState.None, Rasterizer, null, transformationMatrix);
+ 		CurrentFrameFlags.Hacks.CurrentBackgroundMatrixForCreditsRoll = transformationMatrix;
++		if (gameMenu ? MenuLoader.CurrentMenu.PreDrawBackground(spriteBatch) : true)
```
Mirsario:
```suggestion
+		if (!gameMenu || MenuLoader.CurrentMenu.PreDrawBackground(spriteBatch))
```

Tyfyter:
It would be great if something was added to `Main.DoDraw` immediately after
```
if (Filters.Scene["Sepia"].IsInUse())
{
	flag/*this flag controls whether or not the game even tries to apply filters*/ = false;
}
```
to allow modded menu styles to enable filters and set that flag.

Mirsario (approved):

PR #4723, opened 2025-07-19
Add GetAllMoney()

kelibkali:
### What is the new feature?
Get the total amount of money the player has.
It can be useful to utilize the player's total money in certain effects, as it allows for dynamic calculations based on their wealth.
No
```
    public override void UpdateAccessory(Player player, bool hideVisual)
    {
        //Increases attack speed by 1% for every 1 Platinum the player holds.
        player.GetAttackSpeed(DamageClass.Generic) += (float)player.GetAllMoney() / 1000000 / 100f;
    }
```

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -5190,6 +5190,20 @@
  	}
  
 +	/// <summary>
++	/// Get the total amount of money the player has.
++	/// </summary>
++	public long GetAllMoney()
++	{
++		bool overFlowing;
++		long num = Utils.CoinsCount(out overFlowing, inventory, 58, 57, 56, 55, 54);
++		long num2 = Utils.CoinsCount(out overFlowing, bank.item);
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Is there some reason this should be added to tML rather than modders writing the same code in their own mods?

kelibkali:
> Is there some reason this should be added to tML rather than modders writing the same code in their own mods?

This feature should be added to tModLoader because it provides a standardized interface for retrieving the total amount of money a player has across all inventory and bank slots. The method is already used internally in `CanAfford()`, making it a natural fit for inclusion. Without this, modders would have to look into `Player.cs` and copy the logic from the original `CanAfford() `implementation to get the player's total money, which leads to code duplication and maintenance issues. Having a unified `GetAllMoney()` method greatly simplifies development for mod authors.

Examples of usage include: balancing buffs based on wealth, creating shop interaction systems, or implementing currency-based progression mechanics.

So I think it will be useful.

Jofairden:
Could be useful, but how does it take modded slots with money into account? Modders would need to register these somehow.

PR #4727, opened 2025-07-30
tModCodeAssist port

BasicallyIAmFox:
### What is the new feature?
Reworked analyzers for tML.
- Renamed tModLoader.CodeAssist to tModCodeAssist to fit with the tML branding
- Moved tModCodeAssist to the tML repository and added it as an for mods analyzer in `tModLoader.targets`
- Ported the old analyzer rules

Having tModCodeAssist in the tModLoader repository is much better than having it separately and publishing it as a nuget package because we don't have to consider supporting 1.3, 1.4.3, and 1.4.4 at the same time in one codebase. New analyzer rules will also help new modders port vanilla code and can help them simplify their own code.

- Leaving the name as tModLoader.CodeAssist (doable)
- Leaving the analyzer in a separate repository and making tML a submodule of it (wouldn't make a difference)
- Leaving the analyzer as a nuget package (horrendous)

image

### ExampleMod updates
N/A

### TODO
* [x] Magic Number rules
* [X] Port `rand` simplify rule


Tyfyter:
Could attributes to mark fields, parameters, etc. in mods as associated with a specific ID class be added?

BasicallyIAmFox:
> Could attributes to mark fields, parameters, etc. in mods as associated with a specific ID class be added?

i plan to do that in a next pr; for now we had settled down on smaller scope for easier and faster review

PR #4728, opened 2025-07-30
Add support for multiple tilemaps

Moneylover3246:
## Overview

This PR adds support for multiple tilemaps. Originally, this feature was cut from release due to performance issues (to be specific, tilemap access was two-dimensional which created a lot of overhead). This PR optimizes tilemap access through only one extra `Offset` field. Instead of tiles being accessed this way:

`return new Tile((uint)(y + (x * Height)));`

They are now accessed as follows (minimal overhead!) :
`return new Tile((uint)(y + (x * Height) + Offset));`

Internally, `TileData`s are now set to an initial size (8400 * 2400) upon creation, which allow for minimal copying when future Tilemap instances are added. `TileData<T>` also now uses native memory under the hood, so going forward there's no need to pin `TileData<T>.data` using a `GCHandle`. Most importantly, disposed tilemaps are now cached to an internal pool of 'freed' tilemaps. Any time a new tilemap is created of the same size, it'll take the same offset of the disposed tilemap to eliminate the otherwise necessary data allocation.

The addition of multiple tilemaps enable modders to create features such as larger/infinite worlds, or to cache tile data for use in schematics and other mass world operations.

This PR also adds and modifies utility methods to help facilitate development with multiple tilemaps. These include:

1.) `Tilemap.Clear()`, which clears stored data of the given tilemap
2.) `Tilemap.Copy(Tilemap other)`, which copies all data from the current tilemap instance to `other`.
3.) `Tilemap.Dispose()`, which clears the current tilemap's data and adds its offset and size info to the freed tilemap pool.


Chicken-Bones:
The code quality and implementation looks good. Before I scrutinize and nitpick, I need to be sure any performance impacts are acceptable. Please run the `test\TilemapPerformance\TilemapPerformance.csproj` project on a `Release` version of tModLoader before and after your changes, and report the results.

I'm okay with `Tilemap` being `IDisposable`, but I'm concerned that the removal of `readonly` will affect JIT optimizations. I understand you've made it so that accidentally using a disposed `Tilemap` will throw an exception, rather than crashing the process or other undefined behavior, but I'm not willing to pay a performance tradeoff for this safety. It's also not 100% safe because `Dispose` could be called on a copy of the `struct`.

Moneylover3246:
Running the tests showed practically no performance hit with the revamped tilemaps. The difference is likely to have come from the added `Offset` field:

image
image

Regarding the issues with `IDisposable`, it may be favorable to change `Tilemap` to a class. This way we can also include a finalizer in the circumstance where a developer forgets to call `Dispose()` on a Tilemap instance.

Chicken-Bones:
The performance margin is acceptable. Unfortunately we're not willing to make any ABI breaking changes until 1.4.5. However I have an idea...

Can we keep the vanilla `Tilemap` the way it is, non-disposable, with no `Offset` field, and make a class version for extra tilemaps (say, `ExtraTilemap`)? We could define an implicit conversion between the vanilla tilemap and this class, where `Dispose` is a no-op.

Moneylover3246:
That could work, but why would the conversion be necessary when we don't expect the vanilla `Tilemap` to mutate?

Chicken-Bones:
Would you like to come to Discord so we can discuss in more detail? Feel free to ping me or send a DM https://discord.com/invite/tmodloader

PR #4731, opened 2025-08-02
Add opt-out of mod building in tModLoader.targets

Destructor-Ben:
### What is the new feature?
Projects can import `tModLoader.targets` to reference tML without being built as a mod by setting the property `BuildMod` instead of redefining the target that would build a mod.

So tools libraries/tools that aren't necessarily mods can easily reference tMLs code. [TerraUtil](https://github.com/Destructor-Ben/TerraUtil) is one example that I made (and never finished unfortunately, but the build system does work).

Making another tModLoader.targets file, but this would require more effort. and would require splitting tMLMod.targets into multiple files which is probably worse for maintainability.

See porting notes.

### Porting Notes
If you previously redefined the `BuildMod` target like this: `<Target Name="BuildMod"></Target>`
You can opt-in to disabling mod building by putting this in your `.csproj`:
```csproj
<!-- This must go before tModLoader.targets is imported -->
<PropertyGroup>
    <BuildMod>false</BuildMod>
</PropertyGroup>

<Import Project="..\tModLoader.targets" />
```
This will also disable setting PlatformTarget, TargetFramework, and LangVersion to their default values, as well as setting DotNetName and adding files to AdditionalFiles.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/tMLMod.targets
+++ b/patches/tModLoader/Terraria/release_extras/tMLMod.targets
@@ -1,27 +1,36 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-	<!-- Build Defaults -->
+	<!-- Set the default value for BuildMod -->
 	<PropertyGroup>
+		<BuildMod Condition="'$(BuildMod)' == ''">true</BuildMod>
+	</PropertyGroup>
+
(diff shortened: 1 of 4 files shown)


JavidPack:
Looks like it should work to me.

Chicken-Bones (approved):
Please test that Rider on Windows and Unix still works

Destructor-Ben:
Works on Windows with Rider, will ask someone to test on linux.

PR #4733, opened 2025-08-02
Minor Config List Fixes

Destructor-Ben:
### What is the bug?

- The config side indicator has coloured text, which is a leftover from when it used to be a coloured dot. It also doesn't use a tooltip background like the rest of the config list UI.
- The config side indicator has very minor alignment issues
- The small mod icon has very minor alignment issues

### Before/After

<details>
<summary>Before</summary>

image
image

image

image

#### Analysis

image

- Incorrect distances to each edge
- Incorrect size
- Padding shows up in the icon
- Same issues present in the green side indicator icons, but padding is on the other side

image

- Incorrect left padding (may be intentional to stop the small icon hitting the left side
- Codewise the icon is implemented weirdly, incorrectly aligning the text
- The icon isn't always shown since the mod name can be too wide, but the icon should always show if it exists to avoid confusing users

</details>

<details>
<summary>After</summary>

image
image

image
image

image
image

#### Analysis

image

- Fixed size
- Fixed padding being shown
- Fixed the inconsistent margins for the icon

image

- Same margins as before
- Ensured the left and right margins are the same
- Icon now appears for all mods that have one by shrinking the mod name instead of hiding the icon

</details>

### How did you fix the bug?

- Adding a tooltip background and removing the text colour from the config side indicator
- Fixing the alignment of the config side indicator and small mod icon
- Made the small mod icon use the same code as the config side indicator to ensure text is aligned properly
- The tooltip background required adding `OnDraw` to `UIElement`

No

### Other Notes
- I found a strange bug where the offset of the mod icon/side indicator from the top of the parent panel would increase while scrolling. This wasn't super noticeable and was probably because of floating point error but I have no confirmation for this.
image


JavidPack:
Was there a reason the coloring was removed?

Destructor-Ben:
It was originally because the client/server icons were a colored coded dot (which was because they used to have a colored background), and the tooltip was also coloured, but now the colouring is unneeded and is inconsistent with other tooltips.

JavidPack:
> It was originally because the client/server icons were a colored coded dot (which was because they used to have a colored background), and the tooltip was also coloured, but now the colouring is unneeded and is inconsistent with other tooltips.

K, makes sense.

JavidPack:
Is this still WIP?

Destructor-Ben:
Yea, I forgot about it. iirc I was also aligning the small mod icon properly

JavidPack:
K, OnDraw is fine. The mod icon changes are not looking good, I think too much space on the left. They look a lot different in spacing than the config side indicator. Finish the PR description up and either revert or fix the mod icon and it should be good.

PR #4742, opened 2025-08-06
tModCodeAssist: BadIDType diagnostic, bindings improvements

### LINKED ISSUE #4738: Detect mistakenly using ItemID in Recipe.AddTile

Yes

Mod capability as a Modder

### Description

New modders frequently make the mistake of putting an `ItemID` in `Recipe.AddTile` or a `TileID` in `Recipe.AddIngredient` or vice versa. This is usually because they just make a mistake or can't find the TileID of a tile with multiple place styles and aren't familiar with the fact that multiple tile styles exist on the same tile, but then they see what they were looking for in ItemID.

```
public override void AddRecipes() {
	CreateRecipe(999)
		.AddIngredient(TileID.WorkBenches, 3)
		.AddTile(ItemID.OrichalcumAnvil)
		.Register();
}
```

The fix should mark these lines as errors, since there is no conceivable reason these should be written this way.

_No response_

_No response_

BasicallyIAmFox:
##
adds analyzer for detecting "bad id types", e.g. `AddIngredient(TileID.Dirt)`, closes #4738
BadIDType: The expression ItemID.OrichalcumAnvil is not type of TileID

not 100% sure whether or not cases such as this should be reported too:
```cs
const int a = 420;

item.type = a;
```
(`UnknownIDType` diagnostic perhaps? would require ID type attributes at very least)

##
additionally moves bindings stuff to its own file and re-adds bindings for `Tile.TileColor`, `Tile.WallColor` and `Tile.LiquidType` from https://github.com/tModLoader/tModLoader/commit/dbb8a6b4682f68373e544a9df41d8fc77b3952f1#diff-22928eba7d35b676db7176f2ec2c6ab674b5ef194eec9f4ac81a2f7fcde4b6cfR144-R146

### DIFF
--- a/patches/tModLoader/Terraria/ID/LiquidID.cs.patch
+++ b/patches/tModLoader/Terraria/ID/LiquidID.cs.patch
@@ -1,0 +1,14 @@
+--- src/TerrariaNetCore/Terraria/ID/LiquidID.cs
++++ src/tModLoader/Terraria/ID/LiquidID.cs
+@@ -1,3 +_,5 @@
++using ReLogic.Reflection;
++
+ namespace Terraria.ID;
+ 
+ public static class LiquidID
+@@ -7,4 +_,5 @@
+ 	public const short Honey = 2;
+ 	public const short Shimmer = 3;
+ 	public static readonly short Count = 4;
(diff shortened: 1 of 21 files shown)


JavidPack:
Another look at the new feature/fixes

image

PR #4746, opened 2025-08-07
tModCodeAssist: Common Collision Name analyzer

BasicallyIAmFox:
Errors when one of existing types is detected in the namespace or type, e.g.:
```
namespace EtherealMod.Item
                      ^~~^ Error CommonCollisionName: Name EtherealMod.Item collides with an existing type
{
  public class NPC : ModNPC {
               ^~^ Error CommonCollisionName: Name EtherealMod.Item.NPC collides with an existing type
  }
}
```

the list is hardcoded for peak performance reasons


JavidPack:
image

PR #4748, opened 2025-08-07
feat: simplify `Main.player[Main.myPlayer]` -> `Main.LocalPlayer`

### LINKED ISSUE #4745: Main.player[Main.myPlayer] => Main.LocalPlayer code fix

Yes

Mod capability as a Modder

### Description

The tModCodeAssist analyzer should suggest replacing `Main.player[Main.myPlayer]` with `Main.LocalPlayer` to encourage more readable code.

_No response_

_No response_

steviegt6:
*Fixes #4745.*

Simplifies `Main.player[Main.myPlayer]` access expressions into `Main.LocalPlayer`.

### DIFF
--- a/ExampleMod/Content/Items/Accessories/AbsorbTeamDamageAccessory.cs
+++ b/ExampleMod/Content/Items/Accessories/AbsorbTeamDamageAccessory.cs
@@ -42,7 +42,7 @@ namespace ExampleMod.Content.Items.Accessories
 
 			// Remember that UpdateAccessory runs for all players on all clients. Only check every 10 ticks
 			if (player.whoAmI != Main.myPlayer && player.miscCounter % 10 == 0) {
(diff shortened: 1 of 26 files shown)


JavidPack:
Just a note to myself: There is a bug in how VS handles resx files. It adds empty string entries instead of leaving them out, so if you use any other language, the analyzer messages will show as blank instead of English. Apparently this issue [was addressed](https://developercommunity.visualstudio.com/t/Backward-compatibility-behavior:-Resourc/10915672?entry=suggestion&ftype=problem&preview2=true&utm_source=vshelp&stateGroup=active&sort=newest+.) on August 6th 2025, so maybe it'll be fixed in a future VS update.

To fix this ourselves, we need to make sure entries like this are not included in commits:
```
  <data name="SimplifyLocalPlayerTitle" xml:space="preserve">
    <value />
  </data>
```

PR #1410, opened 2021-03-31
ExampleSentry and ExampleSentry Item

Arcri:
ExampleSentry has been brought up multiple times in #help-with-code, and this can be used as a base.

Includes
- Example Sentry
- Example Sentry Item

PR #4762, opened 2025-08-23
Example sentry

JavidPack:
(Note, the initial commit is a port of #1410 by @arcri, I was not able to rebase that branch due to how old it was.)

### What is the new feature?
Adds an example sentry (item, sentry projectile, and sentry shot projectile) to ExampleMod.

Questions about "how to make a sentry" are frequent on Discord. The common response is "it's just a minion with `.sentry` set to true instead of `.minion`", but it turns out there is a lot more to it.

This example should serve as a great starting place for modders making a sentry and is feature complete as well. There are many small aspects to sentries that would not be easily noticed from trying to make a sentry without a full example.

There are 2 main types of sentries, floating and grounded. This example showcases both allowing modders to use whichever design they want.

Please use `ExampleSentry`, `ExampleSentryItem`, and `ExampleSentryShot` as a guide when implementing a sentry, there are many small details that are easy to miss.

### ExampleMod updates
See `ExampleSentry`, `ExampleSentryItem`, and `ExampleSentryShot`.

Here is the new example in action (note how it demonstrates both grounded and floating behaviors):

https://github.com/user-attachments/assets/d7db651f-d6d9-49ab-9eb5-04ac3f30bd2b

### Porting Notes

- Mods with existing sentries should verify the following:
  - Projectiles spawned by sentries should set `ProjectileID.Sets.SentryShot`. This means that re-using an existing projectile directly won't work correctly.
  - Sentry projectiles should set `Projectile.DamageType = DamageClass.Summon;` so that damage values properly update with player damage bonus values changing.
  - Implement correctly supporting the `ProjectileID.Sets.MinionTargettingFeature` by checking `Projectile.OwnerMinionAttackTargetNPC` in targeting logic.
  - Sentry projectiles need `Projectile.netImportant = true;`.
  - Sentry projectiles should use `Projectile.damage` when spawning their shots, not a constant value.
  - `Player.UpdateMaxTurrets` should be called in the sentry item after spawning the sentry projectile (in `ModItem.Shoot`).


ExampleMod/Content/Projectiles/Minions/ExampleSentry.cs
```diff
@@ -0,0 +1,173 @@
...
+				float distanceToTargetNPC = Vector2.Distance(Projectile.Center, ownerMinionAttackTargetNPC.Center);
+				if (distanceToTargetNPC < closestTargetDistance && Collision.CanHit(Projectile.position, Projectile.width, Projectile.height, ownerMinionAttackTargetNPC.position, ownerMinionAttackTargetNPC.width, ownerMinionAttackTargetNPC.height)) {
+					closestTargetDistance = distanceToTargetNPC;
+					hasValidTarget = true;
+					TargetNPCIndex = ownerMinionAttackTargetNPC.whoAmI;
+				}
+			}
+
+			// Otherwise check for the closest enemy.
+			if (!hasValidTarget) {
```
Chicken-Bones: Can you just use `TargetNPCIndex < 0` or `targetNPC == null` instead of a separate bool?

ExampleMod/Content/Projectiles/Minions/ExampleSentry.cs
```diff
@@ -0,0 +1,173 @@
...
+				Projectile.velocity.Y += 0.2f;
+				if (Projectile.velocity.Y > 16f) {
+					Projectile.velocity.Y = 16f;
+				}
+			}
+
+			// Find an enemy to target.
+			bool hasValidTarget = false;
+			float closestTargetDistance = TargetingRange;
+			TargetNPCIndex = -1;
```
Chicken-Bones: Couldn't `TargetNPCIndex` be a local variable? If so, would it be better to just make it `NPC targetNPC`?

Chicken-Bones (changes requested):

ExampleMod/Content/Projectiles/Minions/ExampleSentryItem.cs
```diff
@@ -0,0 +1,79 @@
...
+			}
+			else {
+				position.Y -= 15; // Adjust in-air spawn to spawn with bottom at cursor.
+			}
+
+			// Spawn the sentry projectile at the calculated location.
+			Projectile sentryProjectile = Projectile.NewProjectileDirect(source, position, canPlaceInAir ? Vector2.Zero : new Vector2(0f, 15f), type, damage, knockback, Main.myPlayer, ai2: canPlaceInAir ? 0 : 1);
+
+			// originalDamage facilitates the Projectile.ContinuouslyUpdateDamageStats feature inherent to sentries and minions.
+			sentryProjectile.originalDamage = Item.damage;
```
direwolf420: This is technically not necessary, because TML sets this already in `Projectile.ApplyStatsFromSource` as long as the `source` is `IEntitySource_WithStatsFromItem`.

ExampleMod/Content/Projectiles/Minions/ExampleSentryShot.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Content.Projectiles.Minions
+{
+	// The projectile shot by ExampleSentry.
+	// Note that this class inherits from Sparkling ball and changes a few properties, but otherwise behaves very similarly
+	public class ExampleSentryShot : SparklingBall
+	{
```
direwolf420, line 14: I don't like this. Examples should be isolated and "portable", with no hard dependencies on other ExampleMod content.

ExampleMod/Content/Projectiles/Minions/ExampleSentryItem.cs
```diff
@@ -0,0 +1,79 @@
...
+					j++;
+				}
+				j--; // Move back up to the empty space right above the found solid tile
+
+				// Spawn immediately over the tile.
+				// Depending on the height of the projectile you may need to adjust this, but it works as is for this projectile's height.
+				position = new Vector2(Main.MouseWorld.X, j * 16);
+			}
+			else {
+				position.Y -= 15; // Adjust in-air spawn to spawn with bottom at cursor.
```
direwolf420: This is hardcode. I personally use the following helper extension to accomplish this dynamically:
```cs
/// <summary>
/// Variant of <see cref="Player.FindSentryRestingSpot"/> but counting modded projectiles and easier use with Vector2 parameter.
/// <br/>Only suitable for ground-based sentries. Airborne ones just spawn using Main.MouseWorld
/// </summary>
/// <param name="player">The player.</param>
/// <param name="checkProj">The projectile to check and to offset position with.</param>
/// <param name="position">The position to spawn the sentry in, matching so that it spawn on solid tiles.</param>
public static void FindSentryRestingSpot(this Player player, int checkProj, out Vector2 position)
{
	player.FindSentryRestingSpot(checkProj, out int worldX, out int worldY, out int pushYUp);
	position = new Vector2(worldX, worldY);

	if (checkProj <= -1 || checkProj >= ProjectileLoader.ProjectileCount)
	{
		position.Y -= pushYUp;
		return;
	}

	Projectile p = ContentSamples.ProjectilesByType[checkProj];
	pushYUp = (int)Math.Ceiling(p.height / 2f); //Always round up in case height is not even

	position.Y -= pushYUp;
}
```

And calling it before NewProjectile like this: `player.FindSentryRestingSpot(type, out position);`

Maybe this method can be added to TML directly, or the vanilla method can be adjusted to take into account modded projectiles?

JavidPack: Changing the existing implementation might cause issues to existing mods. Adding the method as another option, but I think the potential confusion over which overload to use might not be worth it, plus having the height calculation in examplemod does have some teaching potential. I've updated the code to use FindSentryRestingSpot

ExampleMod/Content/Projectiles/Minions/ExampleSentryShot.cs
JavidPack: Good point, I've made it independent.

ExampleMod/Content/Projectiles/Minions/ExampleSentry.cs
JavidPack: yes

ExampleMod/Content/Projectiles/Minions/ExampleSentry.cs
JavidPack: I've changed this. I saw that HoundiusShootius was using `ai[1]` and `netUpdate` for this, but from what I can tell the code is running on all clients anyway, so it shouldn't be necessary to use an ai slot. I do wonder if I'm missing some potential netcode issue with how the eye draws, but maybe it's dead code.

ExampleMod/Content/Projectiles/Minions/ExampleSentryShot.cs
```diff
@@ -1,27 +1,41 @@
...
 using Terraria.ModLoader;
 
 namespace ExampleMod.Content.Projectiles.Minions
 {
 	// The projectile shot by ExampleSentry.
-	// Note that this class inherits from Sparkling ball and changes a few properties, but otherwise behaves very similarly
+	// The most important things needed for a projectile spawned by a sentry are:
+	//		ProjectileID.Sets.SentryShot and Projectile.DamageType = DamageClass.Summon
 	public class ExampleSentryShot : SparklingBall
 	{
```
direwolf420:
```suggestion
	public class ExampleSentryShot : ModProjectile
	{
```

Issue #4771, opened 2025-09-02
Ensure backwards compatibility for MonoMod hooks when a method receives a new overload


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently, there is a HookGen task incorporated into TML which builds the TerrariaHooks.dll file to provide modders with MonoMod hooks on almost all methods of the assembly. If the `ExampleName` method has no overloads, its generated name simply becomes the method name `ExampleName`. If there are overloads however, the member's naming scheme is `ExampleName_argtype1_argtype2_...` for each overload, ensuring uniqueness.

If TML adds new overloads to a method that didn't have them before, that member name changes automatically to accomodate, which breaks mods built with the previous versions, if they were to reference these hook names.

The latest example has been a change to TML adding a new `WorldGen.Convert` overload, where there was only 1 previously. As an example, `On_WorldGen.Convert` turns into `On_WorldGen.Convert_int_int_int_int`, and running the mod with the previous naming will cause a "System.TypeLoadException: Could not load type" exception, as tied to these hooks are special delegate types which update aswell.

The proposal is to insert a new step into the HookGen process to "backport" these old names, ideally with some annotation to remove them come a breaking TML release which requires mods to be rebuilt entirely (such as when a major Terraria update releases).

With the proposal and the above example, there will be 3 names:
```cs
Convert
Convert_int_int_int_int
Convert_int_int_int_int_bool_bool
```

### What does this proposal attempt to solve or improve?

This helps preventing mod breakage for mods that don't receive constant updates or no updates at all.

### Additional Info
As an extra, tModPorter can be updated to accomodate for these "backported" hooks, and provide a suggestion to port it to the new name.

PR #4774, opened 2025-09-04
Added PressurePlate set, HitSwitch and SwitchTiles hooks

Tyfyter:
### What is the new feature?
Added `TileID.Sets.PressurePlate`.
Added `(Mod|Global)Tile.HitSwitch` and `(Mod|Global)Tile.SwitchTiles`.

Currently every mod that adds a pressure plate has to use IL editing to do so, multiple mods which add pressure plates are extremely likely to conflict.

Additionally, there are currently no hooks in `Collision.SwitchTiles` allowing the creation of tiles that react to entity collisions.

Finally, a hook in `Wiring.HitSwitch` is also missing. While only specific content calls this method, it is extremely convenient for implementing tile interaction behaviors that should run on all clients and the server. By adding a hook here, tiles can sync effects without writing custom `ModPacket`s.

Custom pressure plate tile:
```cs
public class MyPressurePlateTile : ModTile {
		…
		TileID.Sets.PressurePlate[Type] = 0; // Reacts to all entities. Different values do different things, see the documentation.
		…
	}
}
```

Using `HitSwitch` instead of `HitWire` or `RightClick` for effects that should run on all clients and the server.
```cs
public class MyInteractionTile : ModTile {
	public override bool RightClick(int i, int j) {
		Wiring.HitSwitch(i, j);
		NetMessage.SendData(MessageID.HitSwitch, number: i, number2: j);
	}

	public override void HitWire(int i, int j) {
		Wiring.HitSwitch(i, j);
		NetMessage.SendData(MessageID.HitSwitch, number: i, number2: j);
	}

	public override void HitSwitch(int i, int j) {
		Wiring.TripWire(i, j, 1, 1); 
		Dust.NewDust(new Vector2(i * 16, j * 16), 16, 16, DustID.GreenFairy);
		SoundEngine.PlaySound(SoundID.Unlock, new Vector2(i * 16, j * 16));
	}
}
```

### ExampleMod Updates
`ExamplePressurePlate` has been added. It demonstrates a weighted pressure plate but also has comments showing how to convert to a normal pressure plate tile.

`ExampleSlopeTile` has been updated to be a one-way sensor, sending a wire signal when a player passes through it in the indicated direction. This example shows off the new `ModTile.SwitchTiles` hook and demonstrates custom entity collision logic with a tile.

Here is a video of the new behavior of `ExampleSlopeTile`:

https://github.com/user-attachments/assets/5c1a9d64-381a-49c1-a668-037bc6cd94eb

### Discussion
- `ExampleSlopeTile` can only place on solid tiles, making it a bit difficult to place if intended for puzzle gates. It might be worth moving the example to a more appropriately named tile and use `TileID.Sets.CanPlaceNextToNonSolidTile` like `ExampleLivingFireTile` does.
- The new hook names match the vanilla method names where they are located, which is helpful for modders seeking to adapt vanilla code. `SwitchTiles`, however, isn't really indicative of what it does, and is plural which is not what we usually do.
- `TileID.Sets.PressurePlate` value options might seem a bit strange, but I wanted to easily support all vanilla options without requiring modders to write collision checking logic for a simple pressure plate. It might be better as a `int?[]` rather than a `int[]`.
- ~~Should we add a `Wiring.HitSwitchAndSync` helper method? `HitSwitch` is almost always followed by `NetMessage.SendData(MessageID.HitSwitch...)`.~~ - Added

### Porting Notes
- Replace IL edits with setting the appropriate pressure plate type in `TileID.Sets.PressurePlate`
- If you are playing sounds or spawning dust in `ModTile.HitWire` or `ModTile.RightClick`, they likely don't work correctly in multiplayer. You might want to use `ModTile.HitTile` instead by calling `Wiring.HitSwitch(i, j);` and `NetMessage.SendData(MessageID.HitSwitch, number: i, number2: j);` or just `Wiring.HitSwitchAndSync(i, j);`
- `Collision.SwitchTiles` has a new `Entity` parameter and the old method is `Obsolete`, adjust MonoMod hooks or method calls as appropriate.


direwolf420:
the sample usage won't compile, PressurePlate is an int[]

Tyfyter:
> the sample usage won't compile, PressurePlate is an int[]

Right, forgot I set it to let you pick what triggers the pressure plate and not just that it's a pressure plate.

JavidPack:
Is making a custom pressure plate useful? I guess you could make it a specific color to match a biome. Do we need an ExamplePressurePlate? Is this really the only thing needed to fully support a custom pressure plate? I had thought there might be a lot more to it from when I looked at the pressure plate code in the past. The docs will need to be updated, I still don't understand what the various values means just from reading the docs.

Tyfyter:
I made this PR because I noticed that two mods adding their own pressure plates for unrelated reasons had caused an incompatibility;
Adding a hook would be a good idea to make detouring the method that handles actually doing something when the pressure plate is pressed unnecessary (and could provide additional functionality in the form of letting modders disable vanilla switches via GlobalTile) but this should be all the code that's absolutely necessary for basic compatibility between mods that add pressure plates.

Tyfyter:
Added a hook to make this useful without detouring or IL editing.

direwolf420:
hooks having a default behavior is not TML design (such as ModTile.HitSwitch), you'd need to include the default behavior via a patch instead.

direwolf420:
To clarify, it should probably be a bool return, with true indicating default behavior (which you now removed for some reason), and false for any custom behavior inside the hook

Tyfyter:
> To clarify, it should probably be a bool return, with true indicating default behavior (which you now removed for some reason), and false for any custom behavior inside the hook

I removed the default behavior because I felt it would be better not to have opaque default behavior controlled by magic booleans when that default behavior is just a single line the mod developer is likely already familiar with;

This way it's shorter for anything that doesn't use what was the default behavior and more readable for anything that does.

JavidPack:
So, after studying the code, I can finally make sense of the docs, but as they are they are not clear. They need to describe how each style maps to which objTypes, or at least link to https://terraria.wiki.gg/wiki/Pressure_Plates and explain how those styles listed are the default behaviors. And this really needs an example. Also, am I correct that this won't support weighted pressure plates?

Tyfyter:
This indeed does not currently support weighted pressure plates.

patches/tModLoader/Terraria/Wiring.cs.patch
```diff
@@ -74,6 +74,24 @@
...
+ 
++	/// <summary>
++	/// Contains shared code for tile interactions. Usually called on the server and all clients. Typically called in conjunction with <c>NetMessage.SendData(MessageID.HitSwitch...)</c> to have the method run everywhere else.
++	/// </summary>
+ 	public static void HitSwitch(int i, int j)
+ 	{
+ 		if (!WorldGen.InWorld(i, j) || Main.tile[i, j] == null)
+ 			return;
+ 
++		if (!TileLoader.HitSwitch(i, j, Main.tile[i, j].TileType))
```
Chicken-Bones, line 98: The docs mention something about the return value of `HitSwitch` 

```
Return true to indicate that the tile had some sort of interaction occur. This return value is only used to force specific friendly NPC to keep walking, preventing them from resting on pressure plates, for example.
```

But I don't see that reflected in the code here

ExampleMod/Content/Tiles/ExampleSlopeTile.cs
```diff
@@ -22,5 +26,48 @@ public override bool Slope(int i, int j) {
...
+			if (!tileCoordinates.Intersects(oldPlayerHitbox)) {
+				return false;
+			}
+
+			// Check if the player is now past the bounds of this tile
+			bool success = (Main.tile[i, j].TileFrameX / 18) switch {
+				0 => playerHitbox.Left >= tileCoordinates.Right,
+				1 => playerHitbox.Top >= tileCoordinates.Bottom,
+				2 => playerHitbox.Right <= tileCoordinates.Left,
+				3 => playerHitbox.Bottom <= tileCoordinates.Top,
```
Chicken-Bones, line 57: Should use `>` and `<`

Chicken-Bones (changes requested):
Generally looks good. The `PressurePlates` set has very specific usage, but I don't have better ideas.

patches/tModLoader/Terraria/Wiring.cs.patch
JavidPack: That's the `ModTile.SwitchTiles` hook docs. It's in `NPC.UpdateNPC_Inner` where `Collision.SwitchTiles` is called.

JavidPack:
There has been a suggestion [here](https://github.com/tModLoader/tModLoader/commit/eb4df93a320e92647c56959b087504e8ca37be9e#r166474509) to add the actual Entity instance to `Collision.SwitchTiles` and subsequently the `ModTile.SwitchTiles` hook. I think it's worth considering before finalizing.

Chicken-Bones:
> There has been a suggestion [here](https://github.com/tModLoader/tModLoader/commit/eb4df93a320e92647c56959b087504e8ca37be9e#r166474509) to add the actual Entity instance to `Collision.SwitchTiles` and subsequently the `ModTile.SwitchTiles` hook. I think it's worth considering before finalizing.

I see no reason why not

direwolf420:
The safest option would be to introduce a static field for this (as there is precedent like `Collision.honey`), as introducing a new parameter would break method usage in existing mods, and adding a new overload would [break MonoMod usage](https://github.com/tModLoader/tModLoader/issues/4771) as there are no overloads for `Collision.SwitchTiles` at the moment.

ExampleMod/Content/Tiles/ExampleSlopeTile.cs
JavidPack: Did some testing, they need to check equals since when left and right are equal, they are no longer `Intersects`, so changing this would means that walking really slow will not trigger the tile since the player wasn't intersecting last game update.

JavidPack:
PR has been updated with recent suggestions, and uses the new hookgen fixes to properly handle the new method and obsolete method.

ExampleMod/Content/Tiles/ExampleSlopeTile.cs
Chicken-Bones: Yes, you're correct, touching borders are not intersecting, good call.

ExampleMod/Content/Tiles/ExampleSlopeTile.cs
```diff
@@ -62,6 +62,7 @@ public override bool SwitchTiles(int i, int j, Entity entity, Vector2 position,
 				// Since SwitchTiles is called locally, rather than calling Wiring.TripWire(i, j, 1, 1); here, we delegate that work to HitSwitch which is called on the server and all clients if we send MessageID.HitSwitch as well.
 				Wiring.HitSwitch(i, j);
 				NetMessage.SendData(MessageID.HitSwitch, number: i, number2: j);
+				// Another option is Wiring.HitSwitchAndSync(i, j), for simplicity.
```
Chicken-Bones: If we add a helper, we should use it :)

PR #4778, opened 2025-09-06
Add OnTileConverted hook

Tyfyter:
### What is the new feature?
Added ``ModTile.OnTileConverted`` and ``GlobalTile.OnTileConverted``. which run after a tile is converted by ``WorldGen.Convert``.

Modders may wish to track tiles the player converts, make a biome that reacts to being converted, or otherwise do something which requires running some code when a tile is converted.

```cs
public class ExampleOnTileConverted : GlobalTile {
	public override void OnTileConverted(int i, int j, int fromType, int toType, int conversionType) {
		if (TileID.Sets.Corrupt[toType]) {
			Main.NewText("Hello, cruel world!");
		} else if (TileID.Sets.Corrupt[fromType]) {
			Main.NewText("Goodbye, cruel world!");
		}
	}
}
```


Chicken-Bones:
Assuming it works, lgtm, add one for walls too

Tyfyter:
Tested by adding
```
public override void OnTileConverted(int i, int j, int fromType, int toType, int conversionType) {
	switch ((TileID.Sets.Corrupt[toType] || TileID.Sets.Crimson[toType]).ToInt() - (TileID.Sets.Corrupt[fromType] || TileID.Sets.Crimson[fromType]).ToInt()) {
		case 1:
			Main.NewText($"Hello, cruel world! {TileID.Search.GetName(fromType)} --> {TileID.Search.GetName(toType)} (converted by {conversionType})");
			break;
		case 0:
			Main.NewText($"I guess you're here, cruel world! {TileID.Search.GetName(fromType)} --> {TileID.Search.GetName(toType)} (converted by {conversionType})");
			break;
		case -1:
			Main.NewText($"Goodbye, cruel world! {TileID.Search.GetName(fromType)} --> {TileID.Search.GetName(toType)} (converted by {conversionType})");
			break;
	}
}
```
to `CorruptFossilTile`, fixed issues which arose in testing.

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -875,6 +875,7 @@ public static bool Convert(int i, int j, int conversionType)
 
 		if (tile.TileType == type && TryGetConversionFallback(type, conversionType, out var fallback)) {
 			tile.TileType = (ushort)fallback;
+			using var recursionCounter = new WorldGen.ConversionRecursion();
```
Chicken-Bones: Why? It seems to me that anything calling `Tile/WallLoader.Convert` should have already established a `ConversionRecursion` so that they can call `OnTileConverted` at the end of the vanilla code

patches/tModLoader/Terraria/Projectile.cs.patch
```diff
@@ -1555,21 +1555,32 @@
...
  		}
-@@ -17290,6 +_,8 @@
+@@ -17289,7 +_,11 @@
+ 						if (!(base.position.X + (float)width > vector15.X) || !(base.position.X < vector15.X + 16f) || !(base.position.Y + (float)height > vector15.Y) || !(base.position.Y < vector15.Y + 16f) || !Main.tile[num96, num97].active())
  							continue;
  
++						int tileType = Main.tile[num96, num97].type;
++						int wall = Main.tile[num96, num97].wall;
  						if (type == 10) {
 +							if (!TileLoader.Convert(num96, num97, BiomeConversionID.PurificationPowder))
```
Chicken-Bones, line 1565: You should create and use a conversion recursion context, incase some mod calls `WorldGen.Convert` as part of their hook

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -3013,6 +3054,7 @@
 +
 +					if (preHookType != Main.tile[num8, num9].type) {
 +						flag = true;
++						TileLoader.OnTileConverted(num8, num9, preHookType, Main.tile[num8, num9].type, BiomeConversionID.Chlorophyte);
```
Chicken-Bones, line 3053: This doesn't support all the vanilla conversions. You could merge this into the `else if` chain below with something like:
```cs
if (preHookType != Main.tile[num8, num9].type) {
    flag = true;
}
else if (!convertTile)
{}
else
if (Main.tile[num8, num9].type == 661 || Main.tile[num8, ...
```

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -1996,19 +2006,23 @@
...
+-						continue;
++						goto SkipTilePurityConversion;
++						//continue;
  					case 2:
 -						if (type > TileID.Count || wall > WallID.Count)
+-							continue;
 +						if (type >= TileLoader.TileCount || wall >= WallLoader.WallCount)
- 							continue;
++							goto SkipTilePurityConversion;
++							//continue;
```
Chicken-Bones: These vanilla bounds checks are silly and could just be commented out entirely.
```cs
						/*
						if (type > TileID.Count || wall > WallID.Count)
							continue;
						*/
```

patches/tModLoader/Terraria/WorldGen.cs.patch
```diff
@@ -3234,10 +3276,34 @@
...
+ 				NetMessage.SendTileSquare(-1, num2, num3);
+@@ -55301,7 +_,16 @@
  					if (!(Math.Abs((double)m - vector2D.X) + Math.Abs((double)n - vector2D.Y) < (double)num2 * 0.5 * (1.0 + (double)genRand.Next(-10, 11) * 0.015)))
  						continue;
  
++					ushort oldTileType = Main.tile[m, n].type;
++					ushort oldWallType = Main.tile[m, n].wall;
++
 +					bool convertWall = WallLoader.Convert(m, n, good ? BiomeConversionID.Hallow : crimson ? BiomeConversionID.Crimson : BiomeConversionID.Corruption);
 +					bool convertTile = TileLoader.Convert(m, n, good ? BiomeConversionID.Hallow : crimson ? BiomeConversionID.Crimson : BiomeConversionID.Corruption);
```
Chicken-Bones, line 3312: See above about using a `RecursionContext`, but is there a reason why these can't use `WorldGen.Convert`? Do they differ in some way?

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
Tyfyter: Nothing but ``WorldGen.Convert`` needs to check for recursion before calling the hook because they can't have been called by ``Tile/WallLoader.Convert``.

Chicken-Bones: But if something else in a modded `TileLoader.Convert` hook calls `WorldGen.Convert` the `OnTileConverted` hook will be called twice, because `WorldGen.Convert` will not know that there is already a conversion in progress

Tyfyter: I'm not sure I understand the situation you're saying it could happen in fully because the only interpretation I can think of in which that could happen seems highly impractical and *should* have OnTileConverted be called, but this one line could be moved to the beginning of the method to stop that

patches/tModLoader/Terraria/WorldGen.cs.patch
Tyfyter: Yes, [the hardmode V converts a few more tile types than normal conversion](https://terraria.wiki.gg/wiki/Hardmode_conversion#Converted_tiles), notably silt and — in worlds with over 200,000 of them as a measure to allow it to work in "not the bees" worlds — [Hive](https://terraria.wiki.gg/wiki/Hive) blocks, we would need to reimplement 9 conversions in order to preserve these while switching to ``WorldGen.Convert``.

patches/tModLoader/Terraria/Projectile.cs.patch
Tyfyter: Fixed in https://github.com/tModLoader/tModLoader/pull/4778/commits/4cbdeb9eca2388a0792d1809d1df6293679bcf4e

PR #4782, opened 2025-09-10
Touch-ups and bugfixes regarding biome conversion registries and biome spread

Mellohi138:
### What is the bug?
Modded infection blocks can spread over other infected blocks due to `WorldGen.SpreadInfectionToNearbyTile` using `WorldGen.Convert`
Vanilla blocks are not affected by this as they are either hardcoded, or are using `TileLoader.Convert`, which doesn't suffer from the same issue.

### How did you fix the bug?
I used @Chicken-Bones idea, and created `TileID.Sets.Infectables`, which contains every pure convertible block in vanilla Terraria including the mossified stones, and made `WorldGen.SpreadInfectionToNearbyTile` check to see if the block it's about to convert is in this id set. 

There was an idea to make `WorldGen.SpreadInfectionToNearbyTile` check to see if the block it's about to convert is a fallback block, but it was discarded as it could have gotten complicated, and creating a TileID set is much easier and simpler.

The infected desert fossil blocks now add desert fossil block to the `TileID.Sets.Infectables`.

### Additionally
This is mainly a bugfix PR, but a bugfix isn't the only thing it does. 
There is now a new overload of `TileLoader.RegisterConversion`, allowing modders to register conversions with just a simple tileID instead of creating a new delegate bool altogether

```diff
-TileLoader.RegisterConversion(TileID.ClayBlock, Type, ConvertToAsh);

-public bool ConvertToAsh(int i, int j, int type, int conversionType) {
-	WorldGen.ConvertTile(i, j, TileID.Ash, true);
-	return false;
-}

+TileLoader.RegisterConversion(TileID.ClayBlock, Type, TileID.Ash);

### CODE REVIEW

File: patches/tModLoader/Terraria/ID/TileID.TML.cs [hunk 1]
```diff
@@ -19,6 +19,11 @@ partial class Sets
 		/// <summary> Allows non-solid tiles to be sloped (solid tiles can always be sloped, regardless of this set). </summary>
 		public static bool[] CanBeSloped = Factory.CreateBoolSet();
 
+		/// <summary> Whether or not this tile can be infected by the natural spreading of biomes such as corruption, crimson, or hallow.
+		/// <br>Doesn't affect the spreading in <see cref="WorldGen.hardUpdateWorld"/>, only used in <see cref="WorldGen.SpreadInfectionToNearbyTile"/></br>
```
Chicken-Bones, line 23: `Doesn't affect the spreading in <see cref="WorldGen.hardUpdateWorld"/>`

Why not? Why would a user of this set care?

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -730,13 +730,17 @@ public static void RegisterConversion(int tileType, int conversionType, ConvertT
...
+	/// <param name="fallback">If true, automatically registers tileType as fallback tile of toType.</param>
+	public static void RegisterSimpleConversion(int tileType, int conversionType, int toType, bool purification = true, bool fallback = true)
 	{
 		RegisterConversion(tileType, conversionType, (int i, int j, int type, int conversionType) => {
 			WorldGen.ConvertTile(i, j, toType);
 			return false;
 		});
-		RegisterConversionFallback(toType, tileType, conversionType);
+
+		if(fallback) {
```
Chicken-Bones: Space between `if` and `(`

patches/tModLoader/Terraria/ModLoader/TileLoader.cs
```diff
@@ -730,13 +730,17 @@ public static void RegisterConversion(int tileType, int conversionType, ConvertT
 	/// <param name="conversionType">The conversion type for which the tile should use this conversion.</param>
 	/// <param name="toType">The tile type that this conversion should convert the tile to.</param>
 	/// <param name="purification">If true, automatically registers purification conversions from toType to tileType as well.</param>
-	public static void RegisterSimpleConversion(int tileType, int conversionType, int toType, bool purification = true)
+	/// <param name="fallback">If true, automatically registers tileType as fallback tile of toType.</param>
+	public static void RegisterSimpleConversion(int tileType, int conversionType, int toType, bool purification = true, bool fallback = true)
```
Chicken-Bones: I don't feel good about the ABI breakage of adding a new parameter, and I figured out a better solution in Discord.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ID/TileID.TML.cs
Mellohi138: I just assumed it would create confusion, thats all. 
Disabling vanilla infection spread is already possible in the current stable, so I see no reason to add a new way to disable it.

Chicken-Bones: Maybe the code in `hardUpdateWorld` should just be re-routed to our `SpreadInfectionToNearbyTile` method, unless there are important differences? This would also simplify #4778

Mellohi138: I don't see it happening without a major change to vanilla code. Aside from trying to do modded conversions first then moving onto vanilla conversions, they are too distinct.
Since `WorldGen.hardUpdateWorld` is used for hardmode blocks to spread their infections to purity, it checks to see if the block is a purity block. `WorldGen.SpreadInfectionToNearbyTile` on the other hand uses `WorldGen.Convert`, which was actually made for the clentaminator, so it never had a need to check for purity blocks. That's the entire reason this new `TileID.Set` had to be made in the first place.

PR #4785, opened 2025-09-13
New dropdown style (and more) Enum ModConfig elements

JavidPack:
### What is the new feature?

This PR adds a new Enum config element options in addition to the current slider option. Options include an expanded panel, a dropdown, and a cycle button.  The new expanded panel element will become the new default, applying to all mods built on stable once this reaches stable. The slider option will still be available if the modder desires it.

**Video showcasing all the enum config element options:**

https://github.com/user-attachments/assets/ec3fd11a-f009-4c05-82e2-c39861a75be4

In addition, this PR also adds tooltip support for enum entries.

Modders have asked for a "dropdown" style selector for config elements for a while. Currently enums are portrayed as a slider, but this is a bit strange to some users and doesn't match conventional expectations. 

After some feedback from modders on Discord, the design has arrived at the current implementation.

Any `enum` in a `ModConfig` will use this new expanded panel option automatically. Modders can customize the behavior with attributes.

```cs
// No attribute, will use the expanded panel option
public EquipType SomeEnum;

[Slider]
public EquipType SomeEnum;

[Dropdown]
public EquipType SomeEnum;

[Cycle]
public EquipType SomeEnum;
```

### ExampleMod updates
`ModConfigShowcaseMisc` has been updated with a showcase of each enum option, and some other fixes to make it easier to learn from.

### Porting Notes

- The new expanded panel element will be the default for any mod built after this is in stable. If you want the new element, publish a new updated to your mod.
- If you want the old slider approach, add `[Slider]` to your enum field/property to keep using it. If you want the dropdown or cycle button, use `[Dropdown]` or `[Cycle]`.
- Your localization files will automatically update with tooltip entries for each enum entry.


Tyfyter:
B with a customizable column count seems like it'd be a good approach, a search bar could theoretically be useful for extremely large enums, but enums large enough for it to be useful are probably not a good idea in the first place.

PR #4786, opened 2025-09-14
Adjustable Potion Delay APIs

### LINKED ISSUE #4775: Adjustable Potion Delay

Yes

Mod capability as a Modder

### Description

Implementing custom potion delay, similar to strange brew is a frequent question on Discord. 

Adjusting potion delay globally like Philosopher's Stone has also been requested.

We need a hook that supports both of these options, probably with multipliers as we usually do for order of operations agnostic calculations. Or maybe just a Player field. 

_No response_

Modders can currently detour `Player.ApplyPotionDelay`, but it isn't the best approach for compatibility.

steviegt6:
### What is the new feature?

Resolves #4775.

Adds a new `StatModifier` field, `Player.PotionDelayModifier`, for adjusting the modifier of healing potion delay alongside hooks for modifying the behavior behind applying delays.

`(GlobalItem|ModItem).ModifyPotionDelay` allows for dynamically modifying potion sickness delay calculations.

`(ModPlayer|GlobalItem|ModItem).ApplyPotionDelay` allows for skipping potion sickness delay application.

`CommonItemTooltip.ReducesHealingPotionCooldownByPercent` key added: "Reduces the cooldown of healing potions by {0}%"

See #4775.

Nothing as versatile, but discussion is open.

A custom potion could use `ModifyPotionDelay` to have a reduced potion sickness time: `baseDelay -= 30 * 60;`.

A custom potion could return false in `ApplyPotionDelay` to skip the potion sickness altogether.

A custom accessory can adjust `Player.PotionDelayModifier` to implement a "Reduces the cooldown of healing potions by 25%" effect: `player.PotionDelayModifier *= 0.75f;`

### ExampleMod updates

`ExamplePotionDelayAccessory` is a new accessory that acts on `Player.PotionDelayModifier`, applying adjustments in a manner similar to Philosopher's Stone.
`ExampleHealingPotion` has been updated to have a reduced potion cooldown.

### Porting Notes

- Mods that currently adjust potion sickness/cooldown times should consider using the new methods and fields in this PR.
- Potion delay fields (`Player.potionDelayTime`, `Player.restorationDelayTime`, `Player.mushroomDelayTime`) no longer get set to their reduced values when Philosopher's Stone is equipped, it's instead calculated in `ApplyPotionDelay`. They are still used so any current mods adjusting them will still work, but they should consider using the new hooks for better compatibility. 
- API and ABI breakage for any users referencing the private method `AdjustRemainingPotionSickness`.


JavidPack:
Great job, I think this is good to merge, just going to have CB take a look.

ExampleMod/Content/Items/Accessories/ExamplePotionDelayAccessory.cs
```diff
@@ -0,0 +1,36 @@
...
+			// You can learn more about StatModifiers by referring to ExampleStatBonusAccessory.
+
+			// Note that since we want to apply an additive decrease, we use `-=` to subtract the total
+			// instead of `+=` to increase. You can also use `+=` with a negative value.
+			// Decreases the final potion delay after multipliers. Adjusting PotionDelayModifier.Base would adjust the delay before multipliers
+			// Multiply by 60 to convert ticks to seconds.
+			player.PotionDelayModifier.Flat -= FlatDelayDecrease * 60;
+
+			// Increases the potion delay multiplicitively. Philosopher's Stone does player.PotionDelay *= 0.75f to decrease the delay multiplicitively.
+			player.PotionDelayModifier *= 1 + MultiplicativeDelayBonus / 100f;
```
Chicken-Bones: Example probably shouldn't feature an _increase_ to potion delay.

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
```diff
@@ -719,6 +719,30 @@ public static void OnConsumeMana(Player player, Item item, int manaConsumed)
 		}
 	}
 
+	private delegate void DelegateModifyPotionDelay(Item item, ref int baseDelay, ref StatModifier potionDelay);
+	private static HookList HookModifyPotionDelay = AddHook<DelegateModifyPotionDelay>(p => p.ModifyPotionDelay);
+
+	public static void ModifyPotionDelay(Item item, Player player, ref int baseDelay, ref StatModifier potionDelay)
```
Chicken-Bones: This hook is rather overloaded. I'd like it better if it only included the `ref int baseDelay`, and there was no hook on `ModPlayer`. 

If you can find some reason to have an item dependent hook on `ModPlayer`, then the player hook should probably only have the `ref StatModifier potionDelay`.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4808,6 +4850,30 @@
...
++				/*
+ 				float num2 = buffTime[num];
+ 				float philosopherStoneDurationMultiplier = PhilosopherStoneDurationMultiplier;
+ 				num2 = ((!pStone) ? (num2 / philosopherStoneDurationMultiplier) : (num2 * philosopherStoneDurationMultiplier));
+ 				buffTime[num] = (int)num2;
++				*/
++				float oldMultiplier = oldPotionDelayModifier.Multiplicative * oldPotionDelayModifier.Additive;
++				float newMultiplier = PotionDelayModifier.Multiplicative * PotionDelayModifier.Additive;
++				float time = buffTime[num];
++				time = time / oldMultiplier * newMultiplier;
```
Chicken-Bones: We could do this better, taking into account `Flat` and `Base` as well. We could consider making `StatModifier.Undo` which just reverses the fornula. Though make sure to clamp the resulting `time` value.

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -7991,7 +8057,43 @@
...
++			ItemID.RestorationPotion => restorationDelayTime,
++			ItemID.Mushroom => mushroomDelayTime,
++			_ => potionDelayTime,
++		};
++
++		CombinedHooks.ModifyPotionDelay(sItem, this, ref baseDelay, ref PotionDelayModifier);
++
++		int actualDelay = (int)PotionDelayModifier.ApplyTo(baseDelay);
++
++		bool skip = !CombinedHooks.ApplyPotionDelay(sItem, this, actualDelay);
```
Chicken-Bones, line 8086: Might want to check for `actualDelay > 0` before or after `ApplyPotionDelay`

Chicken-Bones (changes requested):

ExampleMod/Content/Items/Accessories/ExamplePotionDelayAccessory.cs
steviegt6: I don't see the problem; it's clear in its language and demonstrates that you can both decrease and increase the delay.

patches/tModLoader/Terraria/ModLoader/PlayerLoader.cs
steviegt6: The `StatModifier` parameter was per the request of jopo; I will remove these.

ExampleMod/Content/Items/Accessories/ExamplePotionDelayAccessory.cs
JavidPack: I went ahead and made it a decrease, and added a CommonItemTooltip for it. I think overall it's better if this example showcases a more typical usage.

patches/tModLoader/Terraria/Player.cs.patch
JavidPack: Implemented

Chicken-Bones (approved):

PR #4788, opened 2025-09-14
Fix tML Launch Config for Rider

Destructor-Ben:
### What is the bug?
Rider cannot launch tML without modifying `launchSettings.json` because `dotnet.exe` needs to be run instead of `dotnet`.

### How did you fix the bug?
By adding `$(DotNetName)` to `launchSettings.json` and `Terraria.csproj` (copied from `tMLMod.targets`).

No.


JavidPack:
Just a note, I did test this on Rider 2025.2 and it is fixed there, but it seems like Rider doesn't autoupdate users to new major versions, so it's fine to fix it directly.

PR #4795, opened 2025-09-25
Preserve backwards compatibility for MonoMod hooks when a method receives a new overload, inherit Obsolete from origin method

### LINKED ISSUE #4771: Ensure backwards compatibility for MonoMod hooks when a method receives a new overload

JavidPack:
Fixes #4771 

As described in the issue, adding a 2nd overload to a method will result in the TerrariaHooks.dll containing "fully qualified" events for each overload. This breaks compatibility with existing mods since the original simplified (no parameter suffix) event no longer exists in the referenced dll.

To solve this, HookGenerator now checks for a `[OriginalOverloadAttribute]`. If it finds it, it outputs an event for that method as if it were the sole overload. 

If we go with this approach, we'll need to remember to add this to any PR that adds a 2nd overload. I guess alternatively we could do this automatically through some checks against vanilla, but it would be complicated.

I tested the solution and verified the issue by adding a new overload to an existing method and having a monomod detour in ExampleMod. ExampleMod stopped loading as expected when hookgen was run without the attribute, but fixed itself once the attribute was added and hookgen run again. Both hooks seem to work together as well.

Note: MonoMod seems to have [`MonoModOriginal`](https://github.com/MonoMod/MonoMod/blob/reorganize/src/MonoMod.Patcher/Modifiers/MonoModOriginal.cs) and [`MonoModOriginalName`](https://github.com/MonoMod/MonoMod/blob/reorganize/src/MonoMod.Patcher/Modifiers/MonoModOriginalName.cs) attributes, but from what I can tell they are not related to this issue. I might be wrong though.

**In addition,** after giving the issue more thought, we realized that this situation is usually done as a result of making a new method and marking the old one as `Obsolete`. The old method is usually preserved as a forwarder and passes in any assumed default parameters when calling the new method. Mods still hooking the old method risk silently breaking as mods and tModLoader use the new method.

As such, we've decided that methods marked as `Obsolete` will now result in monomod hooks that are also `Obsolete`. This will result in build errors for any mods still hooking the old method signature, forcing mods to migrate to the correct hook. 

Existing mods hooking the now obsolete monomod hooks will still load, which will result in silent or inconsistent breakage of the features of that mod. Rather than allow inconsistent execution of the hooks, we've decided that the hooks should not run at all. Any existing mods hooking these methods will result in error messages logged to the logs, allowing users to report the issue to the original author.

VsDebugConsole_2025-10-06_17-21-06

Using monomod hooks for obsolete methods is now a compile time error
devenv_2025-09-29_11-14-17

The obsolete hooks are now "noops", aside from the logging.
image

# Porting Notes
- Several MonoMod hooks (Both Detours and IL Edits) are now marked as `Obsolete` since the corresponding method is also `Obsolete`. These will no longer function and will result in a compile error. Consult the error message to fix these to use the correct hook.
- Existing mods hooking these `Obsolete` methods that require a rebuild will be logged to the logs. If you see these error messages in other mods please inform the mod author.

### DIFF
--- a/setup/Core/Utilities/HookGenerator.cs
+++ b/setup/Core/Utilities/HookGenerator.cs
@@ -315,7 +315,8 @@ namespace MonoMod.RuntimeDetour.HookGen
             il = addHook.Body.GetILProcessor();
             if (obsolete) {
                 if (LogObsoleteHookSubscribed != null) {
-                    il.Emit(OpCodes.Ldstr, "Warning");
+                    il.Emit(OpCodes.Ldtoken, methodRef);
+                    il.Emit(OpCodes.Call, m_GetMethodFromHandle);
                     il.Emit(OpCodes.Call, LogObsoleteHookSubscribed);
                 }
                 il.Emit(OpCodes.Ret);
@@ -661,11 +662,11 @@ namespace MonoMod.RuntimeDetour.HookGen
             namespace TerrariaHooks {
                 public static class Logging
                 {
-                    public static Action<string> OnObsoleteHookSubscribed;
+                    public static Action<MethodBase> OnObsoleteHookSubscribed;
 
-                    public static void LogObsoleteHookSubscribed(string message) {
+                    public static void LogObsoleteHookSubscribed(MethodBase methodBase) {
                         if(OnObsoleteHookSubscribed != null) {
-                            OnObsoleteHookSubscribed.Invoke(message);
+                            OnObsoleteHookSubscribed.Invoke(methodBase);
                         }
                     }
                 }
@@ -675,7 +676,9 @@ namespace MonoMod.RuntimeDetour.HookGen
             var cls_Logging_0 = new TypeDefinition("TerrariaHooks", "Logging", TypeAttributes.AnsiClass | TypeAttributes.BeforeFieldInit | TypeAttributes.Public | TypeAttributes.Abstract | TypeAttributes.Sealed, OutputModule.TypeSystem.Object);
             OutputModule.Types.Add(cls_Logging_0);
 
-            var fld_OnObsoleteHookSubscribed_1 = new FieldDefinition("OnObsoleteHookSubscribed", FieldAttributes.Public | FieldAttributes.Static, OutputModule.ImportReference(modder.FindType("System.Action`1").Resolve().MakeGenericInstanceType(OutputModule.TypeSystem.String)));
+            OutputModule.ImportReference(modder.FindType("System.Reflection.MethodBase").Resolve());
+
+            var fld_OnObsoleteHookSubscribed_1 = new FieldDefinition("OnObsoleteHookSubscribed", FieldAttributes.Public | FieldAttributes.Static, OutputModule.ImportReference(modder.FindType("System.Action`1").Resolve().MakeGenericInstanceType(modder.FindType("System.Reflection.MethodBase").Resolve())));
             //var fld_OnObsoleteHookSubscribed_1 = new FieldDefinition("OnObsoleteHookSubscribed", FieldAttributes.Public | FieldAttributes.Static, OutputModule.ImportReference(typeof(System.Action<>)).MakeGenericInstanceType(OutputModule.TypeSystem.String)); // Generated code that causes assembly reference issue
 
             cls_Logging_0.Fields.Add(fld_OnObsoleteHookSubscribed_1);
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
As part of this PR I'd also like you to:
- Generate hook properties with both the overloaded and non-overloaded name for methods with the `[OriginalOverloadAttribute]`
- Mark the legacy (simple name) one with `[Obsolete("Use {overloadedname} for forward compatibility", error: true)]`
- Propagate `[Obsolete]` attributes from methods onto the hook properties

direwolf420:
If not already, there should ideally be a compile time exception, or a runtime exception at the worst, when applying this attribute on different overloads at once. Which may require resolving `// TODO: Fix possible conflict when other members with the same names exist.` further below in HookGenerator.cs.

Also for the future (1.4.5+), is it possible to get rid of the simplified "no suffix" names entirely and make the suffixes mandatory? This would probably avoid this problem entirely.

direwolf420:
I like the new solution with no-op hooks 👍

setup/Core/Utilities/HookGenerator.cs
```diff
@@ -646,6 +650,181 @@ CustomAttribute GenerateEditorBrowsable(EditorBrowsableState state)
...
+            var assembly = OutputModule.Assembly;
+
+            // Generated on https://cecilifier.me/
+            /*
+            using System;
+
+            namespace TerrariaHooks {
+                public static class Logging
+                {
+                    public static event Action<string> OnObsoleteHookSubscribed;
```
Chicken-Bones: You can probably simplify the amount of code you need to generate dramatically by removing the `event` keyword here. It's largely useless in modern C# anyway.

setup/Core/Utilities/HookGenerator.cs
```diff
@@ -646,6 +650,181 @@ CustomAttribute GenerateEditorBrowsable(EditorBrowsableState state)
...
+            assembly.MainModule.Types.Add(cls_Logging_0);
+
+
+            //Event: OnObsoleteHookSubscribed
+            var fld_OnObsoleteHookSubscribed_1 = new FieldDefinition("OnObsoleteHookSubscribed", FieldAttributes.Static | FieldAttributes.Private, assembly.MainModule.ImportReference(typeof(System.Action<>)).MakeGenericInstanceType(assembly.MainModule.TypeSystem.String));
+            cls_Logging_0.Fields.Add(fld_OnObsoleteHookSubscribed_1);
+            var md_Add_2 = new MethodDefinition("add_OnObsoleteHookSubscribed", MethodAttributes.Public | MethodAttributes.Static | MethodAttributes.SpecialName | MethodAttributes.HideBySig, assembly.MainModule.TypeSystem.Void);
+            md_Add_2.Parameters.Add(new ParameterDefinition("value", ParameterAttributes.None, assembly.MainModule.ImportReference(typeof(System.Action<>)).MakeGenericInstanceType(assembly.MainModule.TypeSystem.String)));
+            var mr_OpenCompExc_3 = assembly.MainModule.ImportReference(typeof(System.Threading.Interlocked).GetMethods().Single(m => m.Name == "CompareExchange" && m.IsGenericMethodDefinition));
```
Chicken-Bones: It looks like other parts of `HookGenerator` use `modder.FindType(...).Resolve()`

Chicken-Bones (changes requested):

setup/Core/Utilities/HookGenerator.cs
JavidPack: Doing this and other changes fixed the issue.

setup/Core/Utilities/HookGenerator.cs
JavidPack: Done

setup/Core/Utilities/HookGenerator.cs
```diff
@@ -271,6 +280,8 @@ public bool GenerateFor(TypeDefinition hookType, TypeDefinition hookILType, Meth
                 name = nameTmp;
             }
 
+            bool obsolete = method.HasCustomAttribute("System.ObsoleteAttribute"); // Do we need to check if error is true?
```
Chicken-Bones, line 283: No, it's generally fine for mods to keep calling old forwarder methods for longer than it is to hook them

Chicken-Bones (approved):

PR #4796, opened 2025-09-26
WIP: Allow screen shaders when using Retro/Trippy lighting modes. (And additionally on the main menu.)

ink-zoey:
### What is the bug(s)?
1. Retro/Trippy lighting modes as of current disable all screen shaders, causing buggy functionality. (This is likely intentional by Relogic however.)
2. Screen shaders -- other than Terraria's 'Sepia' filter -- currently do not work on the main menu; this fix is similar to the [proposal by Tyfyter](https://github.com/tModLoader/tModLoader/pull/4716#issuecomment-3300875862) on Ebonfly's ['New drawing-related hooks for ModMenus' PR](https://github.com/tModLoader/tModLoader/pull/4716), see below for more details as to possible implementations.
3. Screen shaders on the main menu apply over top of the UI.

### How did you fix the bug(s)?
This initial pass takes from my mods [Retro Lighting Fix](https://github.com/ZenTheMod/FixRetroLighting) and [Zen's Sky](https://github.com/ZenTheMod/ZensSky/blob/main/Common/Systems/CaptureInMenuSystem.cs), to allow for use of screen shaders with both Retro/Trippy lighting modes and on the main menu*.
The current implementation does not currently allow modded screen shaders on the menu currently as `Main.ClearVisualPostProcessEffects` is ran every frame. I'm unsure of the best way to handle deactivating modded filters whilst allowing some to be active on the main menu (I'm open to suggestions here).

There are definitely safer ways to patch the methods I've changed here, and likely a vastly better way to handle what I run in Main.TML.InitScreenTargets.

### DIFF
--- a/patches/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs.patch
@@ -1,0 +1,14 @@
+--- src/TerrariaNetCore/Terraria/Graphics/Capture/CaptureCamera.cs
++++ src/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs
+@@ -131,8 +_,9 @@
+ 		Monitor.Enter(_captureLock);
+ 		if (_activeSettings == null)
+ 			return;
+-
+-		bool notRetro = Lighting.NotRetro;
++		
++		// Always allow capturing with screen shaders.
++		bool notRetro = true;
+ 		if (_renderQueue.Count > 0) {
+ 			CaptureChunk captureChunk = _renderQueue.Dequeue();
+ 			_graphics.SetRenderTarget(null);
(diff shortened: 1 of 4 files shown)


Chicken-Bones:
Thanks for contributing this. I don't have time to review it currently, but I'll try and remember it. 

I'm sympathetic to allowing more shaders to run on the main menu, as part of ModMenus.
If we assume that the reason filters don't work on retro/trippy lighting is for low end machines, then I wonder whether it's right for tML to be the one changing this. Those lighting modes don't even have support for water shaders.

A significant question is what the benefits and costs of having this in tModLoader as opposed to an optional mod are.

ink-zoey:
> If we assume that the reason filters don't work on retro/trippy lighting is for low end machines, then I wonder whether it's right for tML to be the one changing this.

I feel that a screen shader toggle should not be bundled with lighting mode of all things, or exsist at all. This entirely depends on how much you'd want tML to change vanilla config; vanilla configuration is rather confusing and scattered between menus, and I feel that tML should take steps to improve these interfaces for the players sake.

JavidPack:
I always figured we'd eventually just get rid of trippy and retro in tModLoader if anything, especially since they perform much, much worse at some specific resolution and zoom levels. In the past mods had to deal with Reach vs HiDef support, but we eventually got rid of Reach as an option.

PR #4801, opened 2025-09-30
Warning messages for downloading mods directly from servers

JavidPack:
This PR adds warning messages to the prompts in `UIServerModsDifferMessage`, specifically those involving downloading a mod from a server host. These messages inform the user of the heightened risk of downloading mods directly from servers.

Warnings are ordered starting from most risky to least. (Banned->Not on workshop->Hash differs from workshop->Low subscriber count->On workshop->Workshop Offline)

This PR integrates #4345 for even more informative messages about hash differences, allowing the game to determine if a mod from a client is "official" or not.

https://github.com/user-attachments/assets/395542d2-89e9-44f4-ba9d-f19d4291033d


Tyfyter:
For mods which are on the workshop but differ from the versions on it, could it possibly check if the host is a contributor to the workshop item and display a different message based on that? Someone who isn't a contributor but has a supposedly indev version is a lot more suspicious than someone who is.

JavidPack:
> For mods which are on the workshop but differ from the versions on it, could it possibly check if the host is a contributor to the workshop item and display a different message based on that? Someone who isn't a contributor but has a supposedly indev version is a lot more suspicious than someone who is.

That's an interesting idea, but I'm not sure how useful or reliable it would be. We're going to merge this for now, but we can consider this as an future feature.

JavidPack:
> For mods which are on the workshop but differ from the versions on it, could it possibly check if the host is a contributor to the workshop item and display a different message based on that? Someone who isn't a contributor but has a supposedly indev version is a lot more suspicious than someone who is.

We did more investigation, it doesn't seem possible to check for contributors. We could potentially check for the owner, but only through steam invites, so anything we add for this would be really limited. We think a "always trust this host" option would be a good idea, if such a feature becomes highly requested.

PR #4807, opened 2025-10-01
Fixed deallocation crashes in libnfd.

Mirsario:
- Fixes all `nativefiledialog` functions other than `OpenDialog` causing crashes on deallocation step.
- Fixes a microscopic memory leak previously introduced in the `OpenDialog` function. 

The original implementation invoked MSVC++ runtime's `free` function, which seemingly fails on CoreCLR? Replacing that with `Marshal.AllocHGlobal` resolved the crashes.


patches/TerrariaNetCore/Terraria/nativefiledialog.cs.patch
```diff
@@ -0,0 +1,18 @@
...
+ 	}
+ 
++	/*
+ 	[DllImport("msvcrt", CallingConvention = CallingConvention.Cdecl)]
+ 	private static extern void free(IntPtr ptr);
++	*/
++	private static void free(IntPtr ptr)
++	{
++		// The proper way to deallocate.
++		Marshal.FreeHGlobal(ptr);
```
Chicken-Bones, line 14: Is this cross-platform? Looking at the `libnfd` source it looks like all the allocs come from `malloc` in the linked native runtime. There's no guarantee that `FreeHGlobal` maps to the same memory manager

Jofairden: > Is this cross-platform? Looking at the `libnfd` source it looks like all the allocs come from `malloc` in the linked native runtime. There's no guarantee that `FreeHGlobal` maps to the same memory manager

I think it should be. Marshal.AllocHGlobal is implemented by calling into LocalAlloc with LMEM_FIXED. LMEM_FIXED communicates to the memory manager that the memory cannot move. So with the right pointer, it should work even with a different memory manager. The MS docs are confusing as per usual, since there aren't many indications of how it works non widnows. On non windows there’s no Kernel32.dll, and thus no LocalAlloc. In these runtimes, the CLR/Mono implementation of Marshal.AllocHGlobal is backed by malloc, not the Win32 API. So the API exists and works cross-platform, even if internally it doesn’t use LocalAlloc. Since Mono implements Marshal.AllocHGlobal using libc allocation, it should be safe. It’ll just grab unmanaged memory from the C runtime heap. MS seems to recommend newer alternatives: NativeMemory.Alloc or AllocCoTaskMem in .NET 6+.

PR #4814, opened 2025-10-06
Custom tile hit crack drawing hook for ModTile

mayli4:
### What is the new feature?
Adding a hook to ModTile to facilitate custom drawing for tile hit cracking.

NVIDIA_Overlay_6eAhTxlOXN

In certain cases its desirable to have custom cracks, or disable tile cracking alltogether. Notably in custom trees implemented via modtile, but could also be useful for multitiles. 

https://discord.com/channels/103110554649894912/1422486618639437854/1422534386687017001
https://discord.com/channels/103110554649894912/1422486618639437854/1422532719413563414

ModTile::DrawCracks could be split into two seperate methods, one for the crack drawn onto the tile itself, and one for the animation played from HitTile::DrawFreshAnimations, though i found this redundant and just added a parameter for whether or not the call to the hook is for an animation or not.

```csharp
class ExampleTile : ModTile {
    public override bool DrawCracks(SpriteBatch spriteBatch, int i, int j, int damage, int crackStyle, bool isAnimation, int animationTimeElapsed) {
        if(isAnimation) { //... }
        
        //...
        
        return true;
    }
}
```

### ExampleMod updates

Increased ExampleBlock MineResist to 1.5f to properly showcase each frame of the custom hit animation.
Added a basic showcase of DrawCracks to ExampleBlock.


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6731,15 +6731,55 @@
...
+-			else if (tile[x, y].type == 72) {
++			else if (currentTile.type == 72) {
+ 				flag4 = true;
+-				if (tile[x, y].frameX <= 34)
++				if (currentTile.frameX <= 34)
+ 					flag3 = true;
+ 			}
+ 
+-			if (flag3 && tile[x, y].slope() == 0 && !tile[x, y].halfBrick()) {
++			if (flag3 && currentTile.slope() == 0 && !currentTile.halfBrick()) {
```
Chicken-Bones, line 6779: Please avoid refactors like this as they make it hard to maintain the patches

patches/tModLoader/Terraria/HitTile.cs.patch
```diff
@@ -17,3 +17,18 @@
...
  			}
  			else if (Main.tileSolid[tile.type] && !Main.tileSolidTop[tile.type]) {
+@@ -346,6 +_,14 @@
+ 
+ 			int animationTimeElapsed = data[j].animationTimeElapsed;
+ 			if (!((float)animationTimeElapsed >= 10f)) {
++				/* Check if the current tile is a modded tile. */
++				if (TileLoader.GetTile(tile.type) is { } modTile) {
++					/* If it is a modtile, custom drawing is handled and we should skip the rest of this loop for this tile. */
++					if (modTile.DrawCracks(spriteBatch, x, y, damage, data[j].crackStyle, true, animationTimeElapsed)) {
```
Chicken-Bones, line 27: you should probably pass in the default animation `color` and `scale` parameters, as if people just want to change the texture they're going to struggle

ExampleMod/Content/Tiles/ExampleBlock.cs
```diff
@@ -27,5 +30,60 @@ public override void NumDust(int i, int j, bool fail, ref int num) {
 		public override void ChangeWaterfallStyle(ref int style) {
 			style = ModContent.GetInstance<ExampleWaterfallStyle>().Slot;
 		}
+
+		public override bool DrawCracks(SpriteBatch spriteBatch, int i, int j, int damage, int crackStyle, bool isAnimation, int animationTimeElapsed) {
```
Chicken-Bones, line 34: Example doesn't make use of the `animationTimeElapsed` parameter

ExampleMod/Content/Tiles/ExampleBlock.cs
```diff
@@ -27,5 +30,60 @@ public override void NumDust(int i, int j, bool fail, ref int num) {
...
+					1.0f,
+					SpriteEffects.None,
+					0f
+				);
+			} else {
+				Vector2 drawOffset = new Vector2(Main.offScreenRange, Main.offScreenRange);
+				if (Main.drawToScreen) {
+					drawOffset = Vector2.Zero;
+				}
+				Vector2 baseDrawPos = new Vector2(i * 16 - (int)Main.screenPosition.X, j * 16 - (int)Main.screenPosition.Y) + drawOffset;
```
Chicken-Bones, line 71: having to hardcode these calculations is a pain. We should just pass in a parameter for the draw pos of the tile center in screen space, and then the example can just use `origin: new Vector2(8f, 8f)` regardless. Ideally we can remove the `isAnimation` argument entirely.

Chicken-Bones (changes requested):

Issue #4802, opened 2025-09-30
Player Nameplates & Chat Bubbles Render Incorrectly in Reverse Gravity with UI Scale (and at the wrong base position)


lost-werewolf:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

When under any gravity-flipping effect like the Gravity Glove or Gravitation Potion, nameplates above players and their chat bubbles render incorrectly if your UI Scale is not 100%.

<img width="410" height="520" alt="Image" src="https://github.com/user-attachments/assets/03c27d2e-b983-474f-b1a1-8f8639127474" />

Additionally, the chat bubbles above player heads do not accommodate for reverse gravity at all. (Below image is at 100% UI Scale)

<img width="358" height="243" alt="Image" src="https://github.com/user-attachments/assets/c7d2a998-ca10-4930-983e-86f0653b8fc7" />

Last minor issues (moreso nitpicks?) - The nameplates & chat bubbles are based on the player's position, instead of the player's top (position + (width / 2)). This results in them not being centered above the player when zooming in.

<img width="355" height="213" alt="Image" src="https://github.com/user-attachments/assets/02ac81f6-3ef2-464e-996d-5a952f804541" />

In reverse gravity, the nameplate is placed too far above (below?) the player's head, which can be seen in the image for chat bubbles not accommodating reverse gravity - though, it is moved more by the chat bubble being present.

Provided screenshots are on a local tML server with only Cheat Sheet enabled for convenience, no other mods in use. 

### Log File

[client.log](https://github.com/user-attachments/files/22627368/client.log)

[Launch.log](https://github.com/user-attachments/files/22627369/Launch.log)

[Natives.log](https://github.com/user-attachments/files/22627370/Natives.log)

[environment-client.log](https://github.com/user-attachments/files/22627371/environment-client.log)

[server.log](https://github.com/user-attachments/files/22627373/server.log)

[environment-server.log](https://github.com/user-attachments/files/22627375/environment-server.log)

### Steps to reproduce

1. Enter a multiplayer server with another player and switch to the same team.
2. Use a gravitation potion and change your UI scale, and observe the misplaced nameplate.
3. Optionally, have the other player who isn't in reverse gravity send a chat message, and observe it appearing in the wrong position even at 100% UI scale.

### Expected Behavior

[...]

PR #4815, opened 2025-10-06
Fix: mirror other players' chat Y when gravDir == -1

Puglord6383:
### What is the bug?
When player has inverted gravity (gravDir == -1), other players' overhead chat bubbles are vertically mirrored in screen space.
Example: a bubble that should appear 100px from the top of the screen instead appears 100px from the bottom.
This addresses part of issue [GH-#4802](https://github.com/tModLoader/tModLoader/issues/4802). The issue also mentions issues when UI scale is not 100%, which have not been addressed by this PR.
### How did you fix the bug?
In the main draw loop in main.cs, a single if statement was added to flip the y coordinate of other players' chat bubbles.
`if (i != myPlayer && player[myPlayer].gravDir == -1f)
    vec.Y = Main.screenHeight - vec.Y - messageSize.Y;`

Other correction methods may work. Simply mirroring* the ypos of player chat bubbles across the center x-axis of the screen seemed the most intuitive.

It may be getting mirrored again somewhere (cause of the issue in the first place?), resulting in mirroring twice, once by the cause of the issue and again by my fix. If this is the case, a better solution could be to instead prevent this initial problematic mirroring.

*note that when I say mirroring I mean mirroring of y position across an x-axis at the center of the screen - the text is never upside down

### Before change:

Screenshot 2025-10-06 163747
Screenshot 2025-10-06 163823

### After change:

Screenshot 2025-10-06 155012
Screenshot 2025-10-06 155100

### DIFF
--- a/patches/tModLoader/Terraria/Main.cs.patch
+++ b/patches/tModLoader/Terraria/Main.cs.patch
@@ -511,7 +511,7 @@
  	public static bool[] tileMergeDirt = new bool[TileID.Count];
 +	/// <summary>
 +	/// If true, Tiles of the indexed Tile Type will be destroyed by weapons and projectiles. These tiles will also be destroyed when placing a tile at it's location, see also <see cref="TileID.Sets.BreakableWhenPlacing"/>, which is limited to breaking this tile only by tile placement. <see cref="ModProjectile.CanCutTiles"/> can be used to customize if a projectile should destroy tiles with this property.
-+	/// <br/><br/> For desirable plant tiles, set <see cref="TileID.Sets.TileCutIgnore.Regrowth"/> as well to ignore them when using the <see cref="ItemID.StaffofRegrowth"/>.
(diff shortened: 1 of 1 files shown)


Puglord6383:
I don't know why the diff file generated has extraneous whitespace changes - this happens even when I pull from 1.4.4, regenerate source, make my 3 line change again, Diff tModLoader, and re-commit.

JavidPack:
> I don't know why the diff file generated has extraneous whitespace changes - this happens even when I pull from 1.4.4, regenerate source, make my 3 line change again, Diff tModLoader, and re-commit.

Don't worry about it, I think this happens if you use Rider, but we know how to fix these and will do it before merging.

Puglord6383:
Okay, thank you! I will use Visual Studio for future contributions to make it easier.

Chicken-Bones (approved):
I more sophisticated fix is coming in vanilla in 1.4.5

Please mark this code as `// Proper fix coming in 1.4.5`

PR #4817, opened 2025-10-07
One click download

JavidPack:
### What is the new feature?

Adds a new way to download mods. Mods will download in the background rather than show the mod download progress menu. 

It is more convenient to continue browsing for other mods than wait for a specific mod to finish downloading.

The UI could take any number of forms, I've opted for a notification style UI since it seems suitable. We've also wanted a notification feature in the past for other tModLoader features, we should be able to re-use it for those situations if needed.

Here is a video showing a user downloading several mods. As each mod is queued, including dependencies, they add a notification. The UIModDownloadItem updates as well. Also note how when the user navigates to the Mods menu, the menu refreshes when mods become downloaded.

https://github.com/user-attachments/assets/8797ed51-3369-4e95-9a2f-9ac1a89872a1

### Discussion
- Should this be the new default? Should we support both option? The current code has this feature hidden behind holding shift for testing.
- I'm not too familiar with the new mod browser systems, it's possible several things might not be in the most ideal place.
- I added in-menu functionality to `InGameNotificationsTracker`, but I could duplicate the code and make a `InMenuNotificationsTracker` class instead for clarity.


patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -86,7 +86,8 @@ internal static LocalMod[] FindAllMods()
 
 		// Workshop folders
 		WorkshopFileFinder.Refresh(new WorkshopIssueReporter());
-		foreach (string repo in WorkshopFileFinder.ModPaths) {
+		List<string> workshopModPaths = [.. WorkshopFileFinder.ModPaths]; // Avoid collection was modified
```
Chicken-Bones, line 89: This snapshotting should be hidden behind the `ModPaths` property, and combined with a `lock`

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadNotification.cs
```diff
@@ -0,0 +1,143 @@
...
+				timeLeft--;
+			return;
+		}
+
+		timeLeft--;
+
+		if (timeLeft < 0) {
+			timeLeft = 0;
+		}
+	}
```
Chicken-Bones, line 68: some of the UI/counter behaviour could/should definitely be generified. The `ModDownloadNotification` subclass should only have to provide information on when to transition between UI states (eg 'still downloading so must persist', 'finished downloading so can last for 3s/be cleared') and drawing

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
```diff
@@ -501,6 +504,64 @@ internal static async Task<bool> DownloadMods(IEnumerable<ModDownloadItem> mods,
...
+		lock (Interface.modBrowser.modSlugsToUpdateInstallInfo) {
+			Interface.modBrowser.modSlugsToUpdateInstallInfo.UnionWith(fullList.Select(x=>x.ModName));
+		}
+
+		if (DownloadTask == null || DownloadTask.IsCompleted) {
+			DownloadTask = Task.Run(DownloadTaskTask);
+		}
+	}
+
+	internal static void DownloadTaskTask()
```
Chicken-Bones, line 535: heh... naming

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
```diff
@@ -501,6 +504,64 @@ internal static async Task<bool> DownloadMods(IEnumerable<ModDownloadItem> mods,
...
+				 
+				Interface.modBrowser.SocialBackend.DownloadItem(mod, mod.notification);
+				//if (!wasInstalled)
+				//	onNewModInstalled?.Invoke(mod);
+
+				ModOrganizer.LocalModsChanged([mod.ModName], isDeletion: false);
+			}
+		}
+		catch (Exception ex) {
+			//LogModBrowserException(ex, previousMenuId);
```
Chicken-Bones, line 559: So... this is the part to think about. What happens to the rest of the `DownloadQueue`? Do we report the error to the user?

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
```diff
@@ -501,6 +504,64 @@ internal static async Task<bool> DownloadMods(IEnumerable<ModDownloadItem> mods,
...
+	internal static void DownloadModsNoUI(IEnumerable<ModDownloadItem> mods)
+	{
+		var set = mods.ToHashSet();
+		Interface.modBrowser.SocialBackend.GetDependenciesRecursive(set); // Causes a bit of lag
+
+		var fullList = ModDownloadItem.NeedsInstallOrUpdate(set).ToList();
+		if (!fullList.Any())
+			return;
+
+		lock (DownloadQueue) {
```
Chicken-Bones, line 516: If you `lock (DownloadQueue)` in one location, you must lock it everywhere.

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
```diff
@@ -501,6 +504,64 @@ internal static async Task<bool> DownloadMods(IEnumerable<ModDownloadItem> mods,
...
+				item.notification = notification;
+			}
+		}
+
+		lock (Interface.modBrowser.modSlugsToUpdateInstallInfo) {
+			Interface.modBrowser.modSlugsToUpdateInstallInfo.UnionWith(fullList.Select(x=>x.ModName));
+		}
+
+		if (DownloadTask == null || DownloadTask.IsCompleted) {
+			DownloadTask = Task.Run(DownloadTaskTask);
```
Chicken-Bones, line 531: You have a race condition here. When `!DownloadTask.IsCompleted` you assume that you don't need to start it again, because it will keep processing `DownloadQueue`, but it could have already exited the `while (DownloadQueue.Count > 0)` loop but not yet reached `IsCompleted` state.

What you want is a producer-consumer channel with only one consumer, that's automatically 'woken' when an item is added to the queue. There are a few patterns on how to do this without race conditions, but by far the easiest is to use https://learn.microsoft.com/en-us/dotnet/core/extensions/channels

An ad-hoc option is to form a task chain, where you start with `DownloadTask = Task.Completed` and when you add an item to the queue, you do `DownloadTask = DownloadTask.ContinueWith(_ => Download(mod))`

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
Chicken-Bones: On looking closer, it seems that `LocalModsChanged` is incredibly not thread-safe already. We should probably make sure that all of the callbacks registered to `OnLocalModsChanged` are using either `lock` (for actually cross thread-updates) or `QueueMainThreadAction` (for UI) properly

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/UIModBrowser.cs
Jofairden: Could provide a red/error notification?

PR #4818, opened 2025-10-08
Changes to liquid rendering fields to be both public and no longer readonly

Lion8cake:
### What is the new feature?

Changes the `LiquidRenderer` class to have `WATERFALL_LENGTH`, `DEFAULT_OPACITY`, `WAVE_MASK_STRENGTH` and `VISCOSITY_MASK` to be both public and to no longer be readonly. This is so these fields can be edited and used by mods.

This also gives each of the fields documentation on how they're used and resets their contents on initialization similar to arrays found in ``Main`` such as ``projFrames``.

Currently (as of v2025.09) only `WATERFALL_LENGTH` and `DEFAULT_OPACITY` are public while all four fields are also set as readonly. This prevents mod's from being able to easily modify certain rendering settings for vanilla liquids.
This is already a planned change for ModLiquid, as it will be required to be changed so modded liquids can add their own rendering settings

Creating dedicated fields/properties for each field in `LiquidRenderer` that both get and set the data of `WATERFALL_LENGTH`, `DEFAULT_OPACITY`, `WAVE_MASK_STRENGTH` and `VISCOSITY_MASK`.

```
public override void Load()
{
    LiquidRenderer.VISCOSITY_MASK[LiquidID.Honey] = 100;
}
```

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Liquid/LiquidRenderer.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Liquid/LiquidRenderer.cs.patch
@@ -38,7 +38,7 @@
  	private struct SpecialLiquidDrawCache
  	{
  		public int X;
-@@ -60,24 +_,51 @@
+@@ -60,43 +_,55 @@
  		public bool IsSurfaceLiquid;
  		public bool HasWall;
  	}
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
`public` I understand, but removing `readonly` doesn't do particularly much. I guess it's needed for mod liquid, but that's another thing. Anyway, this is fine, but needs to follow https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#comment-out-vanilla-code-rather-than-removing-it

Jofairden:
This is gonna be hard to function properly when multiple mods edit these, no ?

Lion8cake:
@Chicken-Bones Added the commented out vanilla code, apologies for not adding that initially
Since mods will need to rework their liquid rendering edits due to the recent addition of the liquid slope changes, it feels appropriate to allow modders to be able to change these settings too.

@Jofairden unsure what you mean by hard to function

PR #4819, opened 2025-10-09
Remove tryBreakTrees parameter from WorldGen.ConvertTile

JavidPack:
Removed `tryBreakTrees ` parameter from `WorldGen.ConvertTile`. It appears to be useless, as even without it trees will destroy themselves during `SquareTileFrame`. I've left in `TryKillingTreesAboveIfTheyWouldBecomeInvalid` since that seems to be the intended vanilla approach, it's possible it handles something more correctly than just letting tile framing handle it.


Chicken-Bones (approved):
Turns out the `if (tryBreakTrees)` is gone in 1.4.5 anyway. Approved for preview why not

PR #2639, opened 2022-06-27
Support ARM Dedicated Servers

Chicken-Bones:
## This fix is incomplete. 
Depends on #2496

An ARM64 or AnyCPU Steamworks.NET.dll is required.

<details>
<summary>I managed to build one with the following patches to https://github.com/rlabrecque/Steamworks.NET</summary>

```patch
 Standalone/BuildPackages.targets          |  3 +++
 Standalone/Steamworks.NET.Standard.csproj | 11 ++++++++++-
 2 files changed, 13 insertions(+), 1 deletion(-)

diff --git a/Standalone/BuildPackages.targets b/Standalone/BuildPackages.targets
index eba41e7..c0c86fc 100644
--- a/Standalone/BuildPackages.targets
+++ b/Standalone/BuildPackages.targets
@@ -39,6 +39,9 @@
 			<ToBuilt Include="Steamworks.NET.Standard.csproj">
 				<Properties>Platform=x86;Configuration=OSX-Linux;PackagingBuild=true</Properties>
 			</ToBuilt>
+			<ToBuilt Include="Steamworks.NET.Standard.csproj">
+				<Properties>Platform=arm64;Configuration=OSX-Linux;PackagingBuild=true</Properties>
+			</ToBuilt>
 			<ToBuilt Include="Steamworks.NET.Standard.csproj">
 				<Properties>
 					Platform=x64;
diff --git a/Standalone/Steamworks.NET.Standard.csproj b/Standalone/Steamworks.NET.Standard.csproj
index 9a7cce6..7f7eb5a 100644
--- a/Standalone/Steamworks.NET.Standard.csproj
+++ b/Standalone/Steamworks.NET.Standard.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>netstandard2.1</TargetFramework>
 		<RootNamespace>Steamworks</RootNamespace>
 		<AssemblyName>Steamworks.NET</AssemblyName>
-		<Platforms>x64;x86</Platforms>
+		<Platforms>x64;x86;arm64</Platforms>
 		<Configurations>Windows;OSX-Linux</Configurations>
 		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 		<RepositoryType>git</RepositoryType>
@@ -59,6 +59,15 @@
 		<ErrorReport>prompt</ErrorReport>
 	</PropertyGroup>
 
+	<PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'OSX-Linux|arm64'">
[...]

PR #4826, opened 2025-10-16
Add Linux-arm64 natives

Chicken-Bones:
### What is the new feature?

Native libraries for arm64 on Linux. I built them all myself... they should match the current versions we're shipping for other platforms and architectures.

Should let people run arm servers (and clients I guess??)

Well, we'll update to SDL3 with 1.4.5, then we can just use fnalibs-dailies for most of them


Mirsario (approved):
These differ from x86 builds in two notable ways:
- You accidentally include a `/home/cb/build/SDL/install/lib` runpath, which you should either try to exclude, or at least move somewhere else before a test run, as it affects the dependency lookup process during your testing.
- ~~`ld-linux(-aarch64).so.1` linker is imported, not sure if the x86 one inlines or avoids it altogether. I do not see an `ld-linux` in NixOS' `/run/current-system/sw/lib`, so this may reduce compatibility with the weirder systems such as that.~~ [There is however an opt-in workaround on their side](https://wiki.nixos.org/wiki/Nix-ld).
  **EDIT:** Disregard, the need for nix-ld is unavoidable. Confirmed working on an aarc64 NixOS system with just `programs.nix-ld = { enable = true; libraries = [ pkgs.icu ]; }` in the configuration files.

But since SDL3 will swap the files, this is not that important.

Chicken-Bones:
Is this just `FNA3D` and `FAudio`? What about `SDL2`, `libtheorafile` and `libnfd`?

I will try a more sensible build process tomorrow

Mirsario:
Disregard the ld-linux comment.
Runpath exists only in FNA3D and FAudio, yes.

Chicken-Bones:
Well, I could figure out a github action to build FNA3D and FAudio, but I have better things to do atm. Should be good enough now

WesCook:
Does this supersede #2639?

Mirsario:
Yes, closed that one now.

PR #4839, opened 2025-11-02
Move ModAchievement.(Auto)SetStaticDefaults to FinishSetup

direwolf420:
### What is the new feature?
The current `ModAchievement` implementation doesn't take into concideration that most often, achievement conditions are based on existing content. These conditions were meant to be set up within the `SetupContent` part of the load process, which happens concurrently to other content.

This PR moves the condition (and tracker) code near the end of the loading process, commonly known as "`FinishSetup`", allowing it to capture existing and modified content properly.

As it stands, if the modder wants to make conditions that are based on data which is set by other content within `SetStaticDefaults`, it is prone to out-of-order execution since it all happens in the same loading step. A basic example of such a content-dependent condition is shown in a section below.

I simply moved the hooks, which are _usually_ called in SetupContent, to a later step. This does break (non-enforced) spec, but other content (such as IPlant or ModPylon) also has delayed invocation of these hooks. I did it that way as it will not be breaking mods (runtime or compile-time) except in rare cases where ModAchievement.SetStaticDefaults had to run in the original timing. I have tested this with an unreleased mod adding 43 new ModAchievements and experienced no such issues.

```cs
public class ObtainAllDrillsAchievement : ModAchievement
{
	{
		//ModItem would do SetStaticDefaults() => ItemID.Sets.IsDrill[Type] = true;
		for (int i = 0; i < ItemLoader.ItemCount; i++) {
			if (ItemID.Sets.IsDrill[i])
				AddCondition(new ItemPickupCondition((short)i));
		}
	}
}
```

### ExampleMod updates
Nothing, it's rather straightforward, but if needed, can implement the sample usage.

### Porting Notes
Potential runtime breakage if code execution relied on SetupContent timing.

### DIFF
--- a/patches/tModLoader/Terraria/Achievements/AchievementManager.cs.patch
+++ b/patches/tModLoader/Terraria/Achievements/AchievementManager.cs.patch
@@ -141,7 +141,7 @@
 +		foreach (ModAchievement modAchievement in ModdedAchievements) {
 +			modAchievement.SetStaticDefaults();
 +			if (modAchievement.Achievement._conditions.Count == 0)
-+				throw BlameAchievementException(new Exception($"The ModAchievement '{modAchievement}' has no conditions, achievements must have at least one condition."));
++				throw BlameAchievementException(new Exception($"The ModAchievement '{modAchievement.Name}' has no conditions, achievements must have at least one condition."));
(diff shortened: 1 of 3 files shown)


JavidPack:
Seems to work.

PR #4846, opened 2025-11-06
fix(bug): Fix the issue of deleting the environment on first run

lqfy-jhc:
### What is the bug?
The issue occurs in `ScriptCaller.sh` which is called by `start-tModLoaderServer.sh`. During the first run, the script incorrectly deletes the prepared `dotnet_dir` environment. This happens because the script checks for the existence of log files (`client.log` and `server.log`) to determine if the previous run failed. Since these log files don't exist on the first run, the condition is met and the environment is unnecessarily deleted and reset.

### How did you fix the bug?
I enhanced the existing log directory check by adding a first-run flag mechanism:

1. When the script runs, it first checks if the `LaunchLogs` directory exists
2. If the directory doesn't exist (indicating first run), it creates the directory and sets `is_first_run=true`
3. If the directory exists, it sets `is_first_run=false`
4. Modified the environment deletion condition to check both the absence of log files AND that it's not the first run: ```sh if [[ ! -f "$LaunchLogs/client.log" && ! -f "$LaunchLogs/server.log" && ! "$is_first_run" ]]; then ```

This ensures that on the first run, even though the log files don't exist, the environment won't be deleted, allowing the script to use the existing environment.

Yes, alternative approaches include:
- Checking for specific error patterns in existing log files to more accurately detect actual failures
- Creating placeholder log files during first run to prevent the deletion condition from being met
- Using a separate flag file to track successful runs

However, the first-run flag approach is the most straightforward and maintains the existing logic while simply adding a necessary exception for initial execution. It doesn't complicate the code and provides a clear solution to the problem.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
+++ b/patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
@@ -29,6 +29,9 @@ LaunchLogs="$root_dir/tModLoader-Logs"
 
 if [ ! -d "$LaunchLogs" ]; then
 	mkdir -p "$LaunchLogs"
+	is_first_run=true
+else
+	is_first_run=false
 fi
 
 LogFile="$LaunchLogs/Launch.log"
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/release_extras/LaunchUtils/ScriptCaller.sh
```diff
@@ -54,31 +57,31 @@ echo "This may take a few moments."
 source ./DotNetVersion.sh
 
 # Attempt to fix first time Crash To Desktop due to dotnet install failure
-if [[ ! -f "$LaunchLogs/client.log" && ! -f "$LaunchLogs/server.log" ]]; then
+if [[ ! -f "$LaunchLogs/client.log" && ! -f "$LaunchLogs/server.log" && ! "$is_first_run" ]]; then
```
Mirsario: Short-circuit FS interactions.
```suggestion
if [[ ! "$is_first_run" && ! -f "$LaunchLogs/client.log" && ! -f "$LaunchLogs/server.log" ]]; then
```

Mirsario (approved):

PR #4852, opened 2025-11-11
Feature: allow mods to specify whether or not to show content count baubles

mayli4:
### What is the new feature?

Adds a bool in the ``Mod`` class that allows developers to decide whether or not their mods display the content count "baubles".

With the increase in popularity of custom mod list item designs (pictured below), i feel it would be a nice QOL feature to be able to decide whether or not these show. Even if a mod does not have any custom panel design, it would be nice to be able to disable it and reduce clutter.

image
image
image

N/A

```csharp
class MyMod : Mod {
    public MyMod() {
        ShowContentCountBaubles = false;
    }
}
```

### ExampleMod updates
N/A

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Mod.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Mod.cs
@@ -55,6 +55,10 @@ public partial class Mod
 	public string SourceFolder { get; internal set; }
 
 	/// <summary>
+	/// Whether or not this mod will show its content count baubles in the mod list.
+	/// </summary>
+	public bool ShowContentCountBaubles { get; init; } = true;
+	/// <summary>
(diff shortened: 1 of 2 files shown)


NEPTTUNE1:
Should be a user-side toggle in tMod to force this information for the low not nonzero number of people who would benefit from the information being available even if the modder doesn't want it there by default

JavidPack:
There has been discussion of just removing the icons for everyone, but a final decision on that hasn't been made yet. This isn't something the mod should decide, however, if it were to be toggleable it would be a toggle on the UIMods state itself so the user would have that choice. That said, there isn't really much room for another toggle, especially something so obscure.

If a modder is already force customizing their UIModItem, I'm sure they can easily prevent those icons from being added as well.

PR #4866, opened 2025-11-19
fix incorrect ModProjectile.OnKill networking documentation

stormytuna:
ModProjectile.OnKill was incorrectly documented as being called only on the side of the client who owns the projectile, but it is called on all sides


JavidPack:
This was already fixed in https://github.com/tModLoader/tModLoader/commit/0aac5c74ac8f78c1abe552309125bb2e9265b78f

PR #4881, opened 2025-11-28
[Fix] Liquid Edge Masking Fixes

### LINKED ISSUE #4838: Underwater Deactuated Blocks No Longer Render With the Water Visual Filter or Distortion
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Player

### Description

Right after I got the November patch which changed liquid slope rendering, deactuated blocks that are underwater now render as if they aren't, save for for water appearing at their edges.

On a related note, players in front of deactuated blocks also render as if in air, despite particle effects and movement mechanics indicating they are underwater. See attached.

Image

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

steviegt6:
Fixes #4838.

Tile masking now only renders necessary points instead of the entire `tileTarget`, fixing the inappropriate masking as seen in #4838. We collect all points with special masks to draw, as well as vertical spans of edge tiles to draw (thereby fixing any issues of improper masking as a result of using the entirety of `tileTarget`).

Also fixes a couple more cases of small gaps between tiles.


patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -366,4 +386,57 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
 			pCache->Type = (byte)liquidType;
 		}
 	}
+
+	private static void AddEdgePoint(ushort tileX, ushort tileY, bool hasSpecialMask)
+	{
+		// Masks always break the span because they need to be rendered separately.
```
Chicken-Bones: This shouldn't be necessary, simply not adding this tile as an edge point should cause it to be broken, because there will be a gap in the span.

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -366,4 +386,57 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
...
+			return;
+		}
+
+		// Begin building the span if there is none.
+		if (!currentSpan.HasValue) {
+			currentSpan = new EdgeSpan {
+				X = tileX,
+				YStart = tileY,
+				Height = 1,
+			};
```
Chicken-Bones: These 4 lines are duplicated in 3 locations, there should be a way to write this method so that this is the common fallthrough path. Probably by making `if (currentSpan is {} span && span.X == tileX && span.YStart + span.Height == tileY) {currentSpan = span with { Height = span.Height + 1}; return;}` the first condition.

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -54,21 +55,33 @@ public static class LiquidEdgeRenderer
 		AlphaDestinationBlend = Blend.InverseSourceAlpha
 	};
 
-	private static List<Point> BlockWaterBehindLocations { get; } = [];
+	private struct EdgeSpan
+	{
+		public ushort X;
+		public ushort YStart;
+		public byte Height;
```
Chicken-Bones: Please make `ushort`, just for peace of mind (though I doubt render area ever goes over 4080px high). The struct size is gonna be the same regardless. 

Code might be ever-so-slightly cleaner if this was `ushort YEnd`.

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -366,4 +386,57 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
...
+		// Make sure the tiles are contiguous; we don't call this for every
+		// tile, only for confirmed edges.
+		if (span.YStart + span.Height == tileY) {
+			span.Height++;
+			currentSpan = span;
+			return;
+		}
+
+		// If we've broken the span, commit it and start building a new one.
+		edgeSpans.Add(span);
```
Chicken-Bones: You also need to 'commit' the current span after all `CollectEdgeData` calls have finished. That or accept the performance cost of just updating the last `EdgeData` in the list each time you add a new tile.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
steviegt6: Good catch

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
steviegt6: Since we draw a rectangle, it's helpful to just have direct access to the length of tiles we need to stretch to.

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
steviegt6: Would you prefer I lift the mask logic up to the calling method and just make it a separate conditional case?

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
Chicken-Bones: mm well, you could have one be a property

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
Chicken-Bones: That seems sensible

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -366,4 +391,31 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
...
+		if (currentSpan is not { } span)
+			return;
+
+		edgeSpans.Add(span);
+		currentSpan = null;
+	}
+
+	private static void AddEdgePoint(ushort tileX, ushort tileY)
+	{
+		if (currentSpan is { } span && span.X == tileX && span.YEnd == tileY - 1) {
```
Chicken-Bones: This should be `span.YEnd == tileY`

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
```diff
@@ -366,4 +391,31 @@ public static unsafe void CollectEdgeData(LiquidRenderer.LiquidCache* pCache, Ti
...
+		}
+
+		if (currentSpan.HasValue)
+			edgeSpans.Add(currentSpan.Value);
+
+		currentSpan = new EdgeSpan {
+			X = tileX,
+			YStart = tileY,
+			Height = 1,
+		};
```
Chicken-Bones, line 419: Looks great. Consider:
```cs
		if (currentSpan is { } span) {
			if (span.X == tileX && span.YEnd == tileY) {
				currentSpan = span with { Height = span.Height + 1 };
				return;
			}

			edgeSpans.Add(span);
		}

		currentSpan = new EdgeSpan {
			X = tileX,
			YStart = tileY,
			Height = 1,
		};
```

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/Liquid/LiquidEdgeRenderer.cs
steviegt6: No, the end tile is the last tile the span encompasses, so you want to make sure it's the tile before this one.

A span with a height of one is the length of one tile, so the start and end are on the same coordinate.

Issue #4892, opened 2025-12-05
Jetbrains.Annotations Integration


steviegt6:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

It'd be great to decorate various parts of the code with attributes from `Jetbrains.Annotations`. It'd also be nice to directly provide this reference for mods so they don't have to reference it through a package, meaning mods can productively share meta information instead of stripping it from the binaries upon compilation.

### What does this proposal attempt to solve or improve?

Relieves many warnings provided by Rider/R# and allows developers using those tools to write code with more assurance.

Chicken-Bones:
Mods can ship the `Jetbrains.Annotations` dll as a dll ref I guess... though I'm not too opposed to shipping it with tML
We do go to the effort of doc commenting a lot of vanilla methods, so adding attributes is probably okay too...
Lets see what you come up with

steviegt6:
@Chicken-Bones For now, I've only annotated `Mod` and `ILoadable` since they're what's most important for mods. I may contribute more as they come up (and others can as well).

The reason I'd specifically like tModLoader to provide the DLL instead of a bunch of mods is to avoid fighting over where the assembly is sourced from. We have no easy model for transitive vs non-transitive dependencies among mods.

Vanilla code would rarely benefit from annotations: nullability contracts would probably be the most useful, but we can always enable nullable annotations for a part of the code.

PR #4893, opened 2025-12-05
Add Jetbrains.Annotations as a Dependency and Annotate Important Types

steviegt6:
### What is the new feature?

Implements what's suggested in #4892 for important tModLoader types (`ILoadable`, `Mod`).

Letting tModLoader directly provide the `Jetbrains.Annotations` dependency allows mods to keep their annotations in their binaries without worrying about how assemblies will be loaded. This also allows tModLoader to mark its commonly-used types that are loaded through reflection as implicitly used, relieving many annoying code quality/maintenance warnings given by R#/Rider.

It's also feasible to maintain an [external XML file](https://github.com/terraria-catalyst/catalyst/blob/1a8b052863aa5c3e55ebf38a894d4cbe9920dd18/src/TeamCatalyst.Catalyst.Annotations/tModLoader/Attributes.xml) providing these annotations.
Read more here: https://www.jetbrains.com/help/resharper/Code_Analysis__External_Annotations.html

### DIFF
--- a/patches/tModLoader/Terraria/Terraria.csproj.patch
+++ b/patches/tModLoader/Terraria/Terraria.csproj.patch
@@ -8,7 +8,7 @@
  
  	<!-- General -->
  	<PropertyGroup>
-@@ -11,59 +_,98 @@
+@@ -11,59 +_,99 @@
  		<Company>Re-Logic</Company>
  		<Copyright>Copyright © 2022 Re-Logic</Copyright>
  		<RootNamespace>Terraria</RootNamespace>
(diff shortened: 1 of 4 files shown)


patches/tModLoader/Terraria/Terraria.csproj.patch
```diff
@@ -76,6 +79,8 @@
 +		<Reference Include="TerrariaHooks" />
  
 +		<PackageReference Include="DotNetZip" Version="1.16.0" />
++
```
Chicken-Bones: Blank line not desired

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Terraria.csproj.patch
steviegt6: Inserted by VS I think? Fixed.

PR #4909, opened 2025-12-19
Add ModifyPickTile and tool usage hooks for custom mining behavior

### LINKED ISSUE #4735: Pickaxe based Tile hit modifying method

No

Mod capability as a Modder

### Description

A new ModTileHook ran in Player.PickTile following the innitial Player.Getpickaxe damage calculation that carries a nullable Entity usually set to the pickaxe preforming the hit, a ref of the current damage, and the player the hit is originating from. This would also require adding an optional Entity parameter to Player.PickBlock. The parameter is an entity rather than an Item to account for some mods that might use PickBlock with projectiles and npcs

It is currently not possible to create custom interactions based on WHAT is hitting a tile. This would allow for modders to create tiles that require specific pickaxe types to break or have pickaxe types that are extra effective against them. This could also be used to create other custom interactions depending on what pickaxe is hitting a tile, for example a molten pickaxe used on an explosive rock could cause it to explode rather than be picked up.

_No response_

NanTingPer:
## What is the new feature?
refer to #4735

### ModPlayer Hooks
Added two new hooks in ModPlayer for modifying tile and wall mining behavior:
- ModifyPickTile(IEntitySource source, int x, int y, ref int pick)  
  Allows developers to modify the player's mining power against specific tiles. The IEntitySource parameter indicates the origin of the mining action:
  - EntitySource_ItemUse when mining with an item
  - EntitySource_Mount when mining with a mount (e.g., Drill)
- ModifyPickWall(IEntitySource source, int x, int y, ref int damage)  
  Allows developers to modify the damage dealt to walls, using the same source detection mechanism.

### ModItem & GlobalItem Methods
Added three methods in ModItem and GlobalItem to enable custom tool behavior similar to Player.SpecialToolUsageSettings:
- IsAValidTool(Player player)  
  Determines whether this item can be used as a tool without having Item.axe, Item.hammer, or Item.pick. When returning true:
  - The original tool usage is skipped
  - MiningUsageCondition is called to check conditions
  - If condition passes, MiningUsage is executed
- MiningUsageCondition(Player player, int targetX, int targetY)  
  Called when IsAValidTool returns true. Determines whether to execute MiningUsage. If this returns false, no tool action is performed.
- MiningUsage(Player player, int targetX, int targetY)  
  Called when both IsAValidTool and MiningUsageCondition return true. This is where custom mining logic should be implemented.

Without these hooks, developers must use workarounds to:
1. Modify mining power for specific tile types
2. Create custom tools with special mining behavior (e.g., a shovel that only mines stone)
3. Disable or modify existing tool functionality (e.g., Gravedigger's Shovel)
These hooks provide a clean, official API that simplifies development and improves mod compatibility.

## Are there alternative designs?
An alternative would be to directly expose Player.SpecialToolUsageSettings to developers.
### ModPlayer.ModifyPickTile / ModifyPickWall
See ExamplePickTilePlayer.cs:
- Modify copper pickaxe to have 999 mining power
- Reduce drill mining power to 1
- Make stone blocks extremely difficult to mine

###  ModItem (Custom Tool)
See ExampleUseMiningTools.cs:
- Create a tool that mines a 3×3 area of stone blocks when player health is below 200
- Only activates when targeting stone blocks

### GlobalItem (Disable Existing Tools)
See UseMiningToolsGlobalItem.cs:
- Disable the Gravedigger's Shovel functionality globally

### ExampleMod updates
- Added ExamplePickTilePlayer.cs to demonstrate ModifyPickTile and ModifyPickWall hooks
- Added ExampleUseMiningTools.cs to demonstrate creating a custom mining tool
- Added UseMiningToolsGlobalItem.cs to demonstrate disabling existing tool functionality

## Need advice
1. Hook Names - Are ModifyPickTile and ModifyPickWall appropriate names?
2. Code Standards - Does the implementation follow tModLoader conventions?
3. Implementation Reasonableness - Is the three-method design (IsAValidTool → MiningUsageCondition → MiningUsage) appropriate?


w-man12:
hey you can add my original request to the issues that will close if this gets added, this solves my issue adequately

w-man12:
actually, do you think you could add these functions for walls as well? it would be slightly less though still useful

PR #4930, opened 2026-01-08
Menu for switching between beta branches in-game

JavidPack:
This PR adds new functionality to the "Switch to a different tModLoader version" button on the main menu.  The button now (for Steam users) opens a new menu where steam beta branches are listed. The user can click on a button and the game will close and Steam will switch to that branch. The user can then launch the game again on the new branch.

Here is a video of the feature in action:

https://github.com/user-attachments/assets/81d72815-e87e-4be9-9a8e-eb06690768c2

### Discussion
- Does this need a confirmation dialog menu?
- GOG users will just open the wiki page directly still, I think that is fine.
- The icons could be customized or tweaked, or we can just get rid of the icons. 
- The code doesn't check that the current executable is actually the steam download, but I think that is fine.
- Order: Right now it's the default order, which is stable, 1.3, 1.4.3, then each preview. We could customize this if we wanted.
- Localization: Beta branch names and descriptions can't be localized on Steam, but we could detect pre-set phrases like "	Pre-release for mod makers to test upcoming features" and provide localization for those in-game.
  - The same is true for the default branch, which right now is just "public"/"public".


JavidPack:
Remaining discussion points resolved.

New confirmation dialog, other fixes:

https://github.com/user-attachments/assets/f3b87e5c-9191-4d1b-9921-b31c408348b9    

Warning if running from non-steam install location:
dotnet_2026-06-10_14-31-45

PR #4931, opened 2026-01-09
Add BossBarDrawParams.BossBarColor and add its example

### LINKED ISSUE #4914: Add BossBarColor property to BossBarDrawParams

No

Mod capability as a Modder

### Description

An additional color property in BossBarDrawParams that defaults to Color.White and will replace the 3 instances of Color.White in BossBarLoader.DrawFancyBar_TML()

This proposal makes it possible for modders to modify the color of the boss bar itself in addition to the boss head on the bar which is already modifiable, this would allow for the creation of Color effects such as fading the bar in and out without the need to create fully custom draw code.

Alternately an alpha float or byte property would be useful though not as useful as a color

hezuo1hao:
### What is the new feature?

fixes #4914

Added the `BarColor` field to `BossBarDrawParams` and modified several `Color.White` values inside `BossBarLoader. DrawFancyBar_TML(),` so that the Color of the Bar can be adjusted using the newly added `BarColor`

This can quickly modify the color of the Bar or achieve effects such as fading without rewriting the drawing

The existing design can actually achieve a similar effect, but if you want to quickly modify the color, there should be no alternative design

For `BossBarDrawParams.BarColor`, refer to `ExampleGlobalBossBar.cs`

### ExampleMod updates

Modified `ExampleGlobalBossBar.cs`

### DIFF
--- a/ExampleMod/Common/GlobalBossBars/ExampleGlobalBossBar.cs
+++ b/ExampleMod/Common/GlobalBossBars/ExampleGlobalBossBar.cs
@@ -6,6 +6,7 @@ using Terraria.DataStructures;
 using Terraria.GameContent;
 using Terraria.ID;
 using Terraria.ModLoader;
+using System;
 
 namespace ExampleMod.Common.GlobalBossBars
 {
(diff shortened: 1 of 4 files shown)


JavidPack:
Thanks for working on this. 

There is an issue with the current implementation. This PR changes the parameters of the constructor and Deconstruct methods. Since they are public, it is possible a mod is using them and would break with these changes. Please keep the original methods intact and add `[Obsolete]` to them, this will keep those mods working. See https://github.com/tModLoader/tModLoader/commit/81ca62cef82947d00ff16c19d4a4310f7f654c90 for an example of the basic process. 

The https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#be-aware-of-breaking-changes section has some more information about breaking changes and binary compatibility if you want to know more.

PR #4937, opened 2026-01-17
Refactor mod icon code

Destructor-Ben:
### What is the bug?
Mod icon code was duplicated and split across multiple files. Also, no easy way for modders to access mod icons.
This was going to get worse in a PR for the config UI, so I decided to refactor it here.

I also made the modder get informed with a warning if the `icon_small.png` file doesn't exist or is the wrong size to encourage more usage of small mod icons without making it mandatory.

### How did you fix the bug?
Moving the logic to `Mod` and `ModLoader` and publicly exposing the `ModIcon` and `SmallModIcon` fields.

- We could populate `Mod.ModIcon` with `Mod.PlaceholderModIcon` if `icon.png` doesn't exist. We assumed practically all 1.4.4 mods have an `icon.png`, but since it was technically not required as far as we could tell, it's possible this change will be disruptive when it makes it to stable. If that is the case, we can replace that logic in a hotfix if it is deemed necessary.

# Porting Notes

- Mods without a valid icon.png will not load. They will need to add an icon.png and rebuild.
- Mods without a valid icon_small.png will be logged. This is to encourage them to add an icon_small.png soon.


patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -47,6 +49,30 @@ internal void Autoload()
...
+		// Load icon.png and icon_small.png
+		ModIcon = null;
+		SmallModIcon = null;
+
+		if (RequestAssetIfExists<Texture2D>("icon", out var iconAsset)) {
+			if (iconAsset.Size() == new Vector2(80)) {
+				ModIcon = iconAsset;
+			}
+			else {
+				Logger.Info("icon.png needs to be 80x80 pixels.");
```
Chicken-Bones: That should be `Error` not `Info`

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -47,6 +49,30 @@ internal void Autoload()
 		if (Code == null)
 			return;
 
+		// Load icon.png and icon_small.png
+		ModIcon = null;
+		SmallModIcon = null;
+
+		if (RequestAssetIfExists<Texture2D>("icon", out var iconAsset)) {
```
Chicken-Bones: Duplication between here and `GetModIcon`? Should `GetModIcon`?

patches/tModLoader/Terraria/ModLoader/UI/UIModItem.cs
```diff
@@ -83,25 +83,9 @@ public override void OnInitialize()
 		base.OnInitialize();
 
 		string text = _mod.DisplayName + " v" + _mod.modFile.Version;
-		var modIcon = Main.Assets.Request<Texture2D>("Images/UI/DefaultResourcePackIcon", AssetRequestMode.ImmediateLoad);
+		var modIcon = _mod.GetModIcon() ?? Mod.PlaceholderModIcon;
```
Chicken-Bones: Should this be allowed? I guess it was in the past, but now...? Maybe it should be impossible to load a `Mod` without a valid icon.

Chicken-Bones (changes requested):
Generally in approval, few minor questions

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -47,6 +49,24 @@ internal void Autoload()
 		if (Code == null)
 			return;
 
+		// Load icon.png and icon_small.png
+		ModIcon = null;
+		SmallModIcon = null;
```
Chicken-Bones: Is this required? Don't we only ever `Autoload` a mod instance once?

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
```diff
@@ -47,6 +49,24 @@ internal void Autoload()
...
+			       ?? throw new MissingResourceException($"Failed to load icon.png. Reason: {error}.");
+
+		if (RequestAssetIfExists<Texture2D>("icon_small", out var smallIconAsset)) {
+			if (smallIconAsset.Size() == new Vector2(30)) {
+				SmallModIcon = smallIconAsset;
+			}
+			else {
+				Logger.Error("icon_small.png needs to be 30x30 pixels.");
+			}
+		}
```
Chicken-Bones: Would be nice to merge these two with overloads of `GetModIcon` that take the icon path and expected size. I guess the only difference is that you don't want to log if the `icon_small.png` doesn't exist at all

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -419,6 +420,38 @@ internal static void LoadConfiguration()
 		Main.Configuration.Get("LiquidSlopeFix", ref LiquidEdgeRenderer.Enabled);
 	}
 
+	internal static Asset<Texture2D> GetModIcon(TmodFile modFile, out string error)
```
Chicken-Bones: This is a fine location yeah.

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
Destructor-Ben: will check to be sure but can probably be removed

patches/tModLoader/Terraria/ModLoader/Mod.Internals.cs
Destructor-Ben: we could log that icon_small doesnt exist to encourage modders to add small icons, especially because they will be seen more often in the config ui rework

JavidPack:
Any progress on this? It seems like there is just a couple remaining concerns, right? We should be able to get it merged to 1.4.4. I think there was some discussion of enforcing some additional logos for 1.4.5, we can do that in a separate pr to 1.4.5 after we merge this from 1.4.4 to 1.4.5.

Chicken-Bones:
@JavidPack feel free to finish this one up. Only a couple of minor code cleanups/refactors suggested outstanding

Destructor-Ben:
Sorry for late reply, i'm finished with exams now so i can still do this if wanted

JavidPack:
> Sorry for late reply, i'm finished with exams now so i can still do this if wanted

Sure, if you can find that time. If not don't worry about it.

PR #4939, opened 2026-01-20
Fixed main engine crash when approaching unloaded mod plants

Tyfyter:
I can't fathom how this didn't get caught when ``PlantLoader`` was introduced, ``PlantLoader.GetTexture`` can return null, and therefore ``asset`` should be treated as potentially null until it's guaranteed not to be (or at least until it can't be null without modders actively trying to make it null), as treating it as not null can lead to exceptions and it's exclusively used somewhere that any exception is guaranteed to crash the game.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/TilePaintSystemV2.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/TilePaintSystemV2.cs.patch
@@ -92,21 +92,22 @@
  			PrepareTextureIfNecessary(asset.Value);
  		}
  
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Can you explain what this makes unloaded plants look like now?

Tyfyter:
Most importantly they look like something other than a main engine crash, but they should theoretically look like the default version of the plant.

JavidPack:
It falls back to the vanilla plant, which makes sense to do:

image

PR #4946, opened 2026-01-22
Preemptively load SDL2 before FNA3D, resolving NixOS & Linux dl-lookup edge-cases.

Mirsario:
### What is the bug?
On Linux (& Mac???) systems, if the `LD_LIBRARY_PATH` was not configured for the game's process, be it manually or by starting the game with our launch scripts, or even by having a system-wide configuration, then resolution of SDL may often fail, as it is a library that happens to be looked up by the OS as a dependency of FNA3D, rather than something requested and processed directly by the managed assembly.  This is a significant usability problem for mod developers, due the fact that their tooling often uses `dotnet ./tModLoader.dll` to start the game when debugging their mods. An even bigger problem is mod building - even if the developer explicitly runs `dotnet build` after manually setting up and exporting the environment variable, MSBuild will not inherit it when running the `<Exec/>` block we use for packaging `.tmod` files, leading to the following crash on systems that cannot find SDL on their own:
image

This issue of course does not occur on all Linux systems, as some will even have SDL installed at a FHS path, or simply have varying LD implementations. It is however nearly always present on NixOS, a developer-oriented system which by default does not even have an FHS-like environment.

### How did you fix the bug?
I have made the `ResolveNativeLibrary` callback load `SDL` before loading `FNA3D`/`FAudio`/`nfd`, thus giving the system a full path to load, instead of having it look for it itself as a dependency. Perhaps there are better ways, but out of the 2 solutions I know with this one seems most reliable, and reliable enough.

New logging will look like this:
image
image

We could modify tMLMod.targets to explicitly pass the environment variable to the Exec block, but while that would help `dotnet build`, it wouldn't assist with other uses like running the game via Rider.
```diff
 <Target Name="BuildMod" AfterTargets="Build" Condition="'$(BuildMod)' == 'true'">
 	<Exec
+		EnvironmentVariables="LD_LIBRARY_PATH=/path/to/tModLoader/Libraries/Native/Linux/"
 		Command="dotnet $(tMLServerPath) -build $(ProjectDir) -eac $(TargetPath) -define &quot;$(DefineConstants)&quot; -unsafe $(AllowUnsafeBlocks) $(ExtraBuildModFlags)"
 		WorkingDirectory="$(tMLSteamPath)"
 	/>
 </Target>
```

### DIFF
--- a/patches/TerrariaNetCore/Terraria/MonoLaunch.cs.patch
+++ b/patches/TerrariaNetCore/Terraria/MonoLaunch.cs.patch
@@ -16,12 +16,12 @@
  	private static void Main(string[] args)
  	{
 +#if NETCORE
-+		AssemblyLoadContext.Default.ResolvingUnmanagedDll += ResolveNativeLibrary;
++		AssemblyLoadContext.Default.ResolvingUnmanagedDll += static (a, n) => ResolveNativeLibrary(a, n);
 +#else
  		AppDomain.CurrentDomain.AssemblyResolve += delegate (object sender, ResolveEventArgs sargs) {
  			string resourceName = new AssemblyName(sargs.Name).Name + ".dll";
  			string text = Array.Find(typeof(Program).Assembly.GetManifestResourceNames(), (string element) => element.EndsWith(resourceName));
-@@ -25,9 +_,73 @@
+@@ -25,9 +_,79 @@
  				return Assembly.Load(array);
  			}
  		};
@@ -44,7 +44,7 @@
 +
 +	// Note: AssemblyLoadContext.Default.ResolvingUnmanagedDll is only called if the dll is not found on PATH
 +	// In a production environment, you need to prepend to PATH prioritise your shipped libraries, making this logging mostly useless.
-+	private static IntPtr ResolveNativeLibrary(Assembly assembly, string name)
++	private static IntPtr ResolveNativeLibrary(Assembly assembly, string name, string indent = null)
 +	{
 +		const string tab = "  ";
 +		static void LogLine(string message) => Console.WriteLine(message);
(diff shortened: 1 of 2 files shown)


stormytuna:
Tested on NixOS with both a newly generated mod and Overhaul. Building via `dotnet msbuild` in a terminal works as expected.

Destructor-Ben:
Tested on NixOS with ExampleMod, Overhaul and a newly made mod. Tested building from CLI with `dotnet build` and from rider, both with and without `steam-run`. All succeed.

For launching tML, rider can only launch tML when building ExampleMod if it is rider is run via `steam-run` (complains about being unable to find a display driver otherwise). Debugger, breakpoints, and hot reload all work in the tModLoader solution if rider is launched with `steam-run`.

Debugger, breakpoints, and hot reload all work with the newly created mod if the launch config has `$(DotnetName)` changed to `dotnet/dotnet` (which I believe is tModLoaders dotnet, and for some reason, rider's run configs don't run in nix devshells - need to verify) and rider is launched via `steam-run`. If rider is not launched via `steam-run`. the same issue that ExampleMod has occurs when trying to debug, and when trying to run, NixOS says it can't run a dynamically linked executable.

Chicken-Bones:
This is a bodge and doesn't scale as native dependency trees change, but I'll accept it

Chicken-Bones:
Alternative solution:
- Forward the `LD_LIBRARY_PATH` in `Exec` task for mod building deliberately, but use the current value of the environment variable
- Require NixOS users to set the env var manually in Rider run configs (it's the NixOS philosophy afterall, not _our_ problem)

Mirsario:
Closing in favor of #5165, where I did much better research and took entirely different paths.

PR #4950, opened 2026-01-27
Bug Fix: [Setup-CLI, MacOS] Fix decompiling error on Apple silicon MacOS

sxysxy:
### What is the bug?

In the current 1.4.4 version, decompiling Terraria on an Apple Silicon MacBook fails.

The specific reasons are:
- 1.	When locating the Terraria executable on macOS, the attempted path is ~/Library/Application\ Support/Steam/steamapps/common/Terraria/Terraria.exe, but the correct path should be ~/Library/Application\ Support/Steam/steamapps/common/Terraria/Terraria.app/Contents/MacOS/Terraria.bin.osx.
- 2.	When decrypting the executable, the key for the Windows version binary is still being used, rather than the OSX version key from keys.json.
- 3.	On OSX, the required libraries for the decompilation process, such as mscorlib and FNA, should be found in ~/Library/Application Support/Steam/steamapps/common/Terraria/Terraria.app/Contents/Resources.

### How did you fix the bug?

I modified the source code of the setup project to adapt the file paths for cases where the operating system platform is OSX.

I think this is the only solution.

### Verification on MacOS

- Make './setup-cli.sh setup' works:

a2ee26ca69a213b6964cba5982bde830

- Build tModLoader on Apple Silicon MacOS:

Use command(Requires installing x64 version dotnet SDK):

```
/usr/local/share/dotnet/x64/dotnet build src/tModLoader/Terraria/Terraria.csproj
```

image

image

### Addtional 

The modifications to the code should not affect setup on non-OSX platforms, but I don’t have the conditions to test it on a non-OSX system (mainly Windows). It may need to be tested on Windows before being merged into the 1.4.4 branch.

### DIFF
--- a/setup/Core/DecompileTask.cs
+++ b/setup/Core/DecompileTask.cs
@@ -26,11 +26,13 @@ namespace Terraria.ModLoader.Setup.Core
 
 			private readonly PEFile baseModule;
 			private readonly UniversalAssemblyResolver _resolver;
+			private readonly IReadOnlyList<string> extraSearchDirs;
 			private readonly Dictionary<string, PEFile> cache = new();
 
 			public TerrariaAssemblyResolver(PEFile baseModule, string targetFramework, IEnumerable<string> extraSearchDirs)
 			{
 				this.baseModule = baseModule;
+				this.extraSearchDirs = extraSearchDirs.ToList();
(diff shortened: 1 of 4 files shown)


sxysxy:
Closed, because there's a better solution

Issue #4257, opened 2024-06-07
Enable low resolution rendering in maximized windowed and full screen mode on Apple sillicon MacOS.


sxysxy:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

tModLoader code as a Contributor/Maintainer

### Description

I am using 16 Inch Apple M1 Max macbook pro. When I play tmodloader, maximize or fullscreen the window, the resolution goes to 4K, causing the game runs slow. I notice that the original Terraria can render in a low resolution when the window is maximized or in fullscreen mode, 
<img width="604" alt="image" src="https://github.com/tModLoader/tModLoader/assets/16396193/b4590e18-1401-41d1-96b9-ca11b85a809d">
But in tmodloader, the resolution is 3456x1904
<img width="492" alt="image" src="https://github.com/tModLoader/tModLoader/assets/16396193/99b2e4af-388f-452d-9d32-365cc6a0335b">
If I want to play in a lower resolution, I can only reduce the size of the window, it greatly affects game operations. Or open the 'skip frame' options, leading a tearing game screen.

### What does this proposal attempt to solve or improve?

Maybe, Render the framebuffer in a low resolution, and blt to a high resolution screen when presenting to the screen.

JavidPack:
Thanks, we'll need to look into this, I'm not sure yet why Terraria and tModLoader would differ in this behavior.

sxysxy:
It seems to be a high-DPI adaptation issue. I found that in the original version of Terraria, passing the command-line argument /enablehighdpi:1 results in the same default resolution of 3456x1904 as in tModLoader. 
Based on this observation, I speculate that on macOS, tModLoader has this enablehighdpi option enabled by default, and testing shows that it cannot be disabled via /enablehighdpi:0.

sxysxy:
I implemented support for disabling the default HiDPI adaptation mechanism via the command-line argument /enablehighdpi:0 (effective on macOS only), and confirmed through testing that it indeed resolves the issue. See [PR #4951](https://github.com/tModLoader/tModLoader/pull/4951). The resolution could be set to the logical resolution of MacOS.

<img width="919" height="467" alt="Image" src="https://github.com/user-attachments/assets/813c9254-2a84-4e0d-9db4-4da9df6e9b64" />

PR #4951, opened 2026-01-27
New-Feature: Allow /enablehighdpi:0 command-line to turn off highdpi (A solution for issue 4257)

sxysxy:
### Background

[Issue 4257](https://github.com/tModLoader/tModLoader/issues/4257) reported that on macOS, 4K high resolution is forcibly enabled, resulting in poor game graphics performance, meanwhile this issue does not exist in the original version of Terraria. 

While reading the source code, I noticed that tModLoader enables HighDPI mode whenever the detected screen DPI is greater than a fixed threshold (96). This behavior makes sense on many platforms and is generally a good idea, as it improves visual clarity on high-DPI displays.

However, on macOS this logic causes an unintended side effect, Because the MacOS already provides a global scaling mechanism (HiDPI / Retina scaling).
Applications are expected to render at a logical resolution, while macOS transparently maps that to the physical pixel resolution of the display. **Adjusting DPI scaling by the program is always not expected on MacOS**.

By forcing HighDPI mode solely based on the raw DPI value, tModLoader ends up:
- Ignoring the user’s macOS global scaling settings
- Rendering directly at the display’s native physical resolution (e.g. 5K or 6K). Note: most macOS users are using high-resolution displays.
- Significantly increasing the internal render resolution and **severely degrade rendering performance**, 

### What is the new feature?

To preserve compatibility with the existing tModLoader behavior, this change does not disable HighDPI support by default on macOS.

Instead, the logic is adjusted to respect the existing command-line flag -enablehighdpi, which is also used by the underlying FNA framework:
- When running on macOS, tModLoader now checks the value of -enablehighdpi
- If /enablehighdpi:0 is specified, HighDPI mode will no longer be forcibly enabled, even if the detected DPI exceeds the threshold
- If the flag is not provided, or is set to 1, the previous behavior is preserved

Importantly:
- This change only applies at runtime when the platform is macOS
- All other platforms continue to use the original DPI-based logic without modification

This issue is triggered by tModLoader’s own HighDPI enabling logic (DPI > 96 ⇒ force HighDPI), and it has a real performance impact on macOS because it can cause the game to render at the display’s native physical resolution rather than respecting macOS’ effective scaling.

Adding this to tModLoader makes sense because:
- It fixes a platform-specific regression/footgun: macOS DPI semantics differ from Windows, so a fixed DPI threshold does not reliably represent the intended render scale.
- It keeps the default behavior unchanged for existing users, while providing a supported way to opt out when needed.
- It aligns with FNA’s existing configuration surface by honoring -enablehighdpi, rather than introducing a new flag or UI setting.
- It is minimal and low-risk: the change is gated to macOS at runtime and does not affect other platforms.

In short, this makes HighDPI behavior on macOS more predictable and prevents accidental “native-resolution rendering” that can severely hurt FPS on high-resolution displays.

Yes. From my view, tModLoader should disable self-adjustment of HighDPI by default on macOS, this is the expected behavior for MacOS applications. After applying my patch, using /enablehighdpi:0 on macOS to forcibly disable HiDPI adaptation does not cause issues like blurry graphics, while improving game graphics performance.

Just add command-line arugment /enablehighdpi:0 to the tModLoader

### ExampleMod updates

There's no update for the ExampleMod

### DIFF
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -392,6 +392,12 @@ public static partial class Program
 			return;
 
 		// Move DPI detection before early return.
+		if (Platform.IsWindows) {
+			[System.Runtime.InteropServices.DllImport("user32.dll")]
+			static extern bool SetProcessDPIAware();
+
+			SetProcessDPIAware();
+		}
 		SDL2.SDL.SDL_VideoInit(null);
 		SDL2.SDL.SDL_GetDisplayDPI(0, out var ddpi, out float hdpi, out float vdpi);
 		Logging.tML.Info($"Display DPI: Diagonal DPI is {ddpi}. Vertical DPI is {vdpi}. Horizontal DPI is {hdpi}");
@@ -404,13 +410,6 @@ public static partial class Program
 			}
 		}
 		
-		if (Platform.IsWindows) {
-			[System.Runtime.InteropServices.DllImport("user32.dll")]
-			static extern bool SetProcessDPIAware();
-
-			SetProcessDPIAware();
-		}
-
 		if (ddpi >= HighDpiThreshold || hdpi >= HighDpiThreshold || vdpi >= HighDpiThreshold) {
 			Environment.SetEnvironmentVariable("FNA_GRAPHICS_ENABLE_HIGHDPI", "1");
 			Logging.tML.Info($"High DPI Display detected: setting FNA to highdpi mode");
--- a/patches/tModLoader/Terraria/Program.TML.cs
+++ b/patches/tModLoader/Terraria/Program.TML.cs
@@ -402,39 +402,14 @@ public static partial class Program
 		SDL2.SDL.SDL_GetDisplayDPI(0, out var ddpi, out float hdpi, out float vdpi);
 		Logging.tML.Info($"Display DPI: Diagonal DPI is {ddpi}. Vertical DPI is {vdpi}. Horizontal DPI is {hdpi}");
 
-		// Only affect OSX path
+		// Only affect OSX path, see dicussion on https://github.com/tModLoader/tModLoader/pull/4951 for the reason.
 		if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) {
-			if (!GetEnableHighDpiOverride()) {
-				Logging.tML.Info($"Will not set FNA highdpi mode due to /enablehighdpi:0");
-				return;
-			}
+			return; // Let FNA decide by /enablehighdpi command-line argument, default: 0
 		}
 		
 		if (ddpi >= HighDpiThreshold || hdpi >= HighDpiThreshold || vdpi >= HighDpiThreshold) {
 			Environment.SetEnvironmentVariable("FNA_GRAPHICS_ENABLE_HIGHDPI", "1");
 			Logging.tML.Info($"High DPI Display detected: setting FNA to highdpi mode");
 		}
-	}
-
-	//Check FNA enablehighdpi command-line argument (/enablehighdpi:0)
-	private static bool GetEnableHighDpiOverride()
-	{
-		string[] args = Environment.GetCommandLineArgs();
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Your explanation is excellent. I am inclined to accept your recommendation that the tML code be disabled entirely on MacOS.

Am I correct in understanding that if we were to remove the setting of `FNA_GRAPHICS_ENABLE_HIGHDPI` on MacOS:
- The `-enablehighdpi` flag would still work if users wanted it for some reason
- Users have some way to adjust the DPI scaling of the tModLoader application (or all applications) natively on a MacOS system and this works correctly
  - Can you point me to where this setting is so I can test myself? I have a Mac for testing, but am not a user.

Finally, how would this affect MacOS users with external monitors?

sxysxy:
@Chicken-Bones 

Regarding to your concerns:
- The -enablehighdpi flag would still work if users wanted it for some reason
Yes, FNA also checkes this command-line argument([FNAPlatform.cs#50](https://github.com/tModLoader/FNA/blob/2d16be6c66ffa9a157ff1bb50ffe215558b40394/src/FNAPlatform/FNAPlatform.cs#L50)). 
- Where to set DPI scaling
No and it's unnecessary.
Display scaling (HiDPI / Retina) is handled globally and per-display by the MacOS, while apps are expected to render in logical coordinates (points), See [BetterDisplay wiki](https://github.com/waydabber/BetterDisplay/wiki/MacOS-scaling%2C-HiDPI%2C-LoDPI-explanation).  The macOS determines the mapping between logical pixels (points) and physical pixels automatically at the system level, based on the display’s PPI and the user-selected logical resolution(in System Setting -> Monitor).  

 I noticed the following log output by the TML on my 16.2 inch macbook pro with a 3456x2234 built-in monitor:
```
[15:36:48.213] [Main Thread/INFO] [tML]: Display DPI: Diagonal DPI is 255. Vertical DPI is 255. Horizontal DPI is 255
```
$\frac{\sqrt{3456^2 + 2234^2}}{16.2} = 254.02$

So it's clear that tModLoader is making DPI decisions based on physical PPI, This cancels out the system’s automatic scaling. The macOS expects applications to rely on logical scaling managed by the system.

- How would this affect MacOS users with external monitors?

It should behave the same as with the built-in display. I used a display with a physical resolution of 3840x2160, and set it to 1920x1080 in the system settings. Under /enablehighdpi:0, tModLoader can also adapt properly.

image

6ae9154bfa54bd941f8ca481e4c3c0ce

Chicken-Bones:
Well, lets just disable the dpi scaling by default on OSX. Leave it to the `-enablehighdpi` flag

sxysxy:
I just pushed a small follow-up commit to [sxysxy/tModLodaer:issue4257](https://github.com/sxysxy/tModLoader/tree/issue4257) for macOS behavior: HiDPI is no longer forcibly enabled when physical DPI > 96.  
Instead, FNA will now decide whether to enable HiDPI based on the `/enablehighdpi` command-line argument, which is consistent with how vanilla Terraria behaves.

Some additional observations from my testing on Apple Silicon Macs:

1. The performance impact of HiDPI itself seems smaller than the choice of backend:
   - Using `/gldevice:Vulkan` instead of OpenGL
   - Running native ARM64 (`-arm`) instead of x64 via Rosetta
   both provide significantly larger performance gains.

   On my M1 Max MacBook Pro, even with HiDPI enabled, the game remains fluent and smooth on heavy visual effects.

2. That said, forcing HiDPI off does slightly reduce rendering sharpness, especially for UI/text and certain sprite edges, although macOS scaling still keeps it reasonably clear.

Overall, from a first-principles perspective, respecting macOS system display scaling seems to be the right default behavior.

At the same time, Users who's Apple Silicon GPUs are powerful enough and prefer sharper visuals can still turn on HiDPI explicitly via `/enablehighdpi:1` parameter. This keeps platform consistency while preserving user choice.

Chicken-Bones:
This should now be on preview. Please test!

PR #4969, opened 2026-02-03
New-Feature: IPv4 & IPv6 Support

7aGiven:
### What is the new feature?
MultiPlayer, Support IPv4 & IPv6, listen on [::]:7777 and 0.0.0.0:7777 . And support IPv4 & IPv6 connect with .NET Framework .
hope support this in vanilla.
idk
start server on IPv4 & IPv6 address.
### Porting Notes
Not need.
### Test
Already tested on Windows, Ubuntu. I not have MacOS.

What is the bug which Terraria changelog v1.4.4.7 say?


JavidPack:
Was this closed on accident? With the new update we haven't been able to review it yet, but we should get to it eventually if it's still relevant.

PR #4975, opened 2026-02-05
Fix -modpack command line argument for modern modpack format 

### LINKED ISSUE #4063: modpack command line argument does not work
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

There is a command line argument/server config argument called `modpack` that should specify a modpack to load. This does not work anymore since modpacks have been changed to be a folder.

The logic in `CommandLineModPackOverride` assumes that modpacks are a .json file directly in the ModPacks folder, this is no longer the case. 

A workaround is to provide the full path to enabled.json: `-modpack ExamplePack/Mods/enabled.json` (command line) or `modpack=ExamplePack/Mods/enabled.json` (config.txt)

Additionally, `ModOrganizer.ModPackActive` is not saved automatically when a modpack is enabled. Also, the logic for `ModOrganizer.ModPackActive` in `_FindMods` checks `if (Directory.Exists(ModPackActive))`, which makes no sense, preventing those mods from loading. As it is now, the feature of a consistent modpack between game sessions does not work. Enabling a modpack does work, but it just enables the mods, not the specific local copies of the mods.

Once #3822 is merged, we can tackle this issue. I believe at least some of the issues presented here still persist in that rework.

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

4RH1T3CT0R7:
Fixes #4063                                                                                                                                                   
                                                                                                                                                                
  - Add InitializeCommandLineModPack() to set ModPackActive before FindAllMods                                                                                  
  - Update CommandLineModPackOverride() to support both modern (folder) and legacy (.json) formats                                                              
  - Modern modpacks now properly load frozen mods from the modpack folder


JavidPack:
Thanks for this. We are currently busy with 1.4.5 patches but will get to this.

4RH1T3CT0R7:
@JavidPack bumping this now that the 1.4.5 crunch is over, still merges cleanly if you get a minute.

PR #4977, opened 2026-02-06
Remove the two letters that broke NoSync libraries for Both mods

Tyfyter:


Tyfyter (commented):
We could alternatively add these two letters to the parameter name

PR #4986, opened 2026-02-12
Clean Up `Terraria`/`TerrariaNetCore` Patches

steviegt6:
Removes some dead code including `GStreamerAudioSupport`, removes some useless patches, wraps all new `TerrariaNetCore` code in appropriate conditionals, and unifies Windows to use SDL3's clipboard.


patches/Terraria/Terraria/Graphics/TileBatch.cs.patch
```diff
@@ -21,9 +14,9 @@
  
  			FillVertexBuffer(texture, sprites, offset, num, 0);
 +
-+			//TODO:
-+			// Decompilation bugfix.
-+			// Remove this change whenever ICSharpCode.Decompiler 8.0.0 or 8.0.0.*-preview3 releases.
++			// Decompiler bug: change Unsafe.SizeOf<T> calls to unsafe
++			// sizeof(T) expressions.  Also requires making the containing
++			// method unsafe.
```
Chicken-Bones, line 19: > Remove this change whenever ICSharpCode.Decompiler 8.0.0 or 8.0.0.*-preview3 releases

patches/TerrariaNetCore/Terraria/System/Windows/Forms/Control.cs.patch
```diff
@@ -0,0 +1,11 @@
+--- src/Terraria/Terraria/System/Windows/Forms/Control.cs
++++ src/TerrariaNetCore/Terraria/System/Windows/Forms/Control.cs
+@@ -4,7 +_,7 @@
+ // Keep the namespace declaration outside of the conditional so the game will
+ // still compile nicely with references to the namespace, even if no types in
+ // it are present.
+-#if !WINDOWS
++#if !WINDOWS && !NETCORE
```
Chicken-Bones, line 8: Are all the references to `Control` removed on net core?

patches/Terraria/ReLogic/OS/Windows/PathService.cs.patch
```diff
@@ -1,11 +0,0 @@
---- src/decompiled/ReLogic/OS/Windows/PathService.cs
-+++ src/Terraria/ReLogic/OS/Windows/PathService.cs
-@@ -7,7 +_,7 @@
- 
- internal class PathService : ReLogic.OS.Base.PathService
- {
--	public override string GetStoragePath() => Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "My Games");
-+	public override string GetStoragePath() => Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), "My Games");
```
Chicken-Bones, line 8: These two folders are the same?

Chicken-Bones (changes requested):
Excellent work!

patches/TerrariaNetCore/Terraria/System/Windows/Forms/Control.cs.patch
steviegt6: There are none. The entire class could have been left alone, but `Form` also isn't present in .NET Core, so I opted to just filter the type out (the old .NET Core patches already did this when the class was defined in `Main`).

patches/Terraria/ReLogic/OS/Windows/PathService.cs.patch
steviegt6: Yes, the enums are even the same values. This patch only exists on Linux/macOS, presumably just because the Mono stdlib declares the MyDocuments member before the Personal member so the decompiler changes what's emitted

PR #4991, opened 2026-02-16
Removed DrawYOffset in ExampleChandelier to avoid double offset

### LINKED ISSUE #4988: ExampleChandelier is drawn higher than expected when placed
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

The preview and placed tile both are affected by DrawYOffset, which for tiles using a vanilla custom renderer like TileID.Sets.MultiTileSway might mean the offset is applied on top of existing vanilla offset logic.

This means examplechandelier when placed is too high, compared to the placement preview and other vanilla chandeliers using the exact same sprite.

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

bedelln:
Closes #4988.
### What is the bug?
ExampleChandelier was being drawn higher than expected, and higher than the vanilla chandeliers it matched due to being affected by both DrawYAsset and TileID.Sets.MultiTileSway, doubling its offset.

### How did you fix the bug?
I removed the DrawYAsset line affecting ExampleChandelier.

Finding a way to avoid double-offsetting it might be helpful, or finding a more root cause.

### DIFF
--- a/ExampleMod/Content/Tiles/Furniture/ExampleChandelier.cs
+++ b/ExampleMod/Content/Tiles/Furniture/ExampleChandelier.cs
@@ -59,7 +59,6 @@ namespace ExampleMod.Content.Tiles.Furniture
 			TileObjectData.newTile.RandomStyleRange = 6;
 			TileObjectData.newTile.StyleHorizontal = true;
 			TileObjectData.newTile.StyleLineSkip = 2;
-			TileObjectData.newTile.DrawYOffset = +2;
 			TileObjectData.addTile(Type);
 
 			AddMapEntry(new Color(235, 166, 135), Language.GetText("MapObject.Chandelier"));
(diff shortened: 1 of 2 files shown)


JavidPack:
Thanks, I'll look into this later

Solxanich:
Thanks for bringing this forward. 

I've picked this up and poked around the code base, A suggestion is to use `SetDrawPositions` to correct the offsetY to remain with vanilla parity. See ModBannerTile for example.
image


-----------------------------
Before; affects ExampleWideBannerTile, ExampleChandlier. 
Visible with Echo Blocks mixed in.

image

With FlutterSlime using ModBannerTile, it is correctly aligned. We undo the effects of DrawYOffset in SetDrawPositions within the same ModBannerTile class.
image
image
image

ModBannerTile is consistent with base Terraria's handling in setting DrawYOffset to -2; 
unable to verify where vanilla does the correction or reasoning.

PR #4992, opened 2026-02-17
Adds check for custom vanilla renderer to avoid double Y offset

Eon02:
Solves #4988.
What is the bug? 
ExampleChandelier was being drawn higher than expected, and higher than the vanilla chandeliers it matched due to being affected by both DrawYAsset and TileID.Sets.MultiTileSway, doubling its offset.

How did you fix the bug?
I added a conditional to add the desired offset depending on whether a custom renderer is enabled.

Adding a more robust fix to account for other sorts of custom renderers might be beneficial in the future.


JavidPack:
This PR changes 260 files? I think you forked the 1.4.5 branch and then made a pull request to the 1.4.4 branch, but I'm not sure. It might be easiest to just make a branch off of 1.4.4 and redo your patches there and make a new PR.

Solxanich:
Merged solution in #4991 for vanilla consistency. 
Thank you for your proposal and greatly appreciate providing a suggested approach

Issue #4195, opened 2024-05-03
Loading a player with Deprecated items in inventory with globalitem data corrupts player


JavidPack:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

The player has StardustHammer in inventory somehow, and enters a world with a mod with a GlobalItem class. That class saves data on the StardustHammer when leaving the world. Attempting to load the player on the player select screen results in the following:

```
[01:40:38.125] [Main Thread/WARN] [tML]: Silently Caught Exception: 
System.Collections.Generic.KeyNotFoundException: Terraria.ModLoader.Default.UnloadedGlobalItem
   at Terraria.Item.GetGlobalItem[T]() in tModLoader\Terraria\Item.TML.cs:line 30
   at Terraria.ModLoader.IO.ItemIO.LoadGlobals(Item item, IList`1 list) in tModLoader\Terraria\ModLoader\IO\ItemIO.cs:line 166
   at Terraria.ModLoader.IO.ItemIO.Load(Item item, TagCompound tag) in tModLoader\Terraria\ModLoader\IO\ItemIO.cs:line 113
   at Terraria.ModLoader.IO.ItemIO.Load(TagCompound tag) in tModLoader\Terraria\ModLoader\IO\ItemIO.cs:line 119
   at Terraria.ModLoader.IO.PlayerIO.LoadInventory(Item[] inv, IList`1 list) in tModLoader\Terraria\ModLoader\IO\PlayerIO.cs:line 121
   at Terraria.ModLoader.IO.PlayerIO.Load(Player player, TagCompound tag) in tModLoader\Terraria\ModLoader\IO\PlayerIO.cs:line 74
   at Terraria.Player.Deserialize(PlayerFileData data, Player newPlayer, BinaryReader fileIO, Byte[] tplrData, Int32 release, Boolean& gotToReadName) in tModLoader\Terraria\Player.cs:line 45095
   at Terraria.Player.LoadPlayerFromStream(PlayerFileData playerFileData, Byte[] plrData, Byte[] tplrData) in tModLoader\Terraria\Player.cs:line 45037
   at Terraria.Player.LoadPlayer(String playerPath, Boolean cloudSave) in tModLoader\Terraria\Player.cs:line 45002
   at Terraria.Player.GetFileData(String file, Boolean cloudSave) in tModLoader\Terraria\Player.cs:line 45702
[...]

PR #4996, opened 2026-02-19
fix: Prevent unobtainable items from breaking characters in TML.

Mirsario:
### What is the bug?
Acquiring an unobtainable item such as one of [these hammers](https://terraria.wiki.gg/wiki/Luminite_Hammers) causes player characters to fail to load in TML, at `GetGlobalItem<UnloadedGlobalItem>()` being called with an empty state, if not earlier.

### How did you fix the bug?
Short-circuit if item IsAir after netDefaults.

Explicitly check for Sets.Deprecated, duplicating vanilla code.


JavidPack:
I assume this also fixes https://github.com/tModLoader/tModLoader/issues/4195?

PR #4998, opened 2026-02-21
Fix: [1.4.5] GetItemSettings docs, Chests and Lang patches (1218 broken patches)

ken-tn:
Updates docs for GetItemSettings to 1.4.5 diff offsets.

Fixes 2 broken patches.


ken-tn:
Chests.cs and Lang.cs also updated, fixed failure matches and added porting notes.

PR #5001, opened 2026-02-23
tModLoader patches for 1.4.5: Batch 9 (1189 broken patches remain)

vasyza:
### What is the bug?

49 FAILURE hunks across 36 patch files fail to apply against the decompiled Terraria 1.4.5 source due to vanilla API changes, line offsets, and removed/restructured code.

### How did you fix the bug?

- **RecipeGroup.cs.patch**: Full rewrite for new 1.4.5 API (`recipeGroupIDs` dictionary removed, `RegisterGroup(string, RecipeGroup)` replaced with instance `Register()` method)
- **RecipeGroupHelper.cs**: Removed `recipeGroupIDs.Clear()` reference
- **ArmorShaderData.cs / HairShaderData.cs**: Rewritten hunks for new `EffectParameter<T>` generic API replacing `Shader.Parameters["name"]`
- **ID patches** (BuffID, MountID, NPCHeadID, MessageID, ItemHoldStyleID, ItemUseStyleID, BiomeConversionID): Renumbered to account for new 1.4.5 constants
- **Obsolete hunk removal**: UIWorldCreation, UIText, DyeInitializer — TML fixes now in vanilla 1.4.5
- **Context line offsets**: ~20 smaller patches with shifted context lines (Lang, Netplay, Program, TcpSocket, NetClientSocialModule, etc.)

No — patches must match the decompiled 1.4.5 source. Hunks were updated to align with the new vanilla code while preserving all TML modifications.

### Test results

`setup-cli.sh patch tml -f -m OFFSET` confirms 1189 remaining failures (down from 1238).

### DIFF
--- a/patches/tModLoader/ReLogic/Content/AssetRepository.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/AssetRepository.cs.patch
@@ -1,11 +1,11 @@
 --- src/TerrariaNetCore/ReLogic/Content/AssetRepository.cs
 +++ src/tModLoader/ReLogic/Content/AssetRepository.cs
-@@ -1,64 +_,152 @@
+@@ -1,78 +_,154 @@
  using System;
 +using System.Collections.Concurrent;
  using System.Collections.Generic;
 +using System.Diagnostics;
-+using System.IO;
+ using System.IO;
  using System.Linq;
 +using System.Runtime.CompilerServices;
 +using System.Threading;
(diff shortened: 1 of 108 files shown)


JavidPack:
We are extremely glad that you are willing to contribute patch fixes, but as we talked about on Discord with other contributors we intend for the tML team to handle the work on the patches due to various reasons. The main reason is that reviewing changes to large amounts of patches while making sure nothing is missed is more time consuming than just fixing the patches ourselves. Another reason is we are keeping an eye out for things in the source code to be fixed or changed on Terraria's end in future patch releases, which is easier for us to do as we come across them while fixing the patches.

Once the majority of patches are done, we'll be able to handle community contributions fixing things that still need fixing and it'll be a lot easier to handle at that stage.

(I want to leave this PR open so other contributors who didn't see the Discord message can see our current reasoning as well.)

PR #5008, opened 2026-02-27
ICustomLoadable

Tyfyter:
### What is the new feature?
Added ICustomLoadable, which allows mods to handle autoloading types they define in their own code and without unnecessary template instances

```
public abstract class Foo : ICustomLoadable
{
	static void ICustomLoadable.CustomLoad(Mod mod, Type type) {
		RegisterAllBazzes();
	}
	static void RegisterAllBazzes(Mod mod, Type foo)…
}
public class Bar(Quux quux) : Foo {
	
}
```


Tyfyter:
Superceded by #5009

PR #5009, opened 2026-02-27
Use new C# static abstract interface methods to provide more customizable and lower boilerplate autoloading

Chicken-Bones:
### What is the new feature?

New interfaces allow implementation of custom autoloading that isn't just `mod.AddContent(Activator.CreateInstance(type))`

```cs
public interface IAutoload<TImpl> where TImpl : IAutoloader
{
}

public interface IAutoloader 
{ 
	public static abstract void Autoload(Mod mod, Type type); 
}

public interface ICustomAutoload : IAutoload<ICustomAutoload.AutoloadImpl>
{
	public static abstract void Autoload(Mod mod);
	
	...
}
```

`ILoadable` has been changed to:
```cs
public interface ILoadable : IAutoload<ILoadable.AutoloadImpl>
{
	...
	
	public class AutoloadImpl : IAutoloader 
	{ 
		public static void Autoload(Mod mod, Type type) 
		{ 
			if (type.GetConstructor(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic, Type.EmptyTypes) is { } ctor) 
				mod.AddContent((ILoadable)ctor.Invoke(null)); 
		} 
	}
}
```

Autoloading via `ILoadable` requires that types have a default constructor which can be called to create a 'template instance' of the content which is passed to `Mod.AddContent`. Sometimes, having to provide default constructors is just needless boilerplate. 

Sometimes, especially for systems and types provided by library mods and consumed by other dependent mods, there may be no notion of a 'template instance', but the library may still want to do `Type` level reflection based registration, on behalf of all consuming mod types. For example, a mod providing a serialization library, where packets are structs and types are auto-registered. Eg:

```cs
public interface IPacket<TSelf> : IAutoload<PacketAutoloader> {}
{
    void Write(BinaryWriter w);
    static abstract TSelf Read(BinaryReader r);
    static abstract void Handle(TSelf packet, int fromClient);
}

public readonly record struct CustomSoundPacket(ushort TileX, ushort TileY) : IPacket<CustomSoundPacket>
{
    public void Write(BinaryWriter w) {
        w.Write(TileX);
        w.Write(TileY);
    }

    public static CustomSoundPacketRead(...) => new(r.ReadUShort(), r.ReadUShort());

    public static void Handle(CustomSoundPacket packet, int fromClient)
    {
        if (Main.netMode == NetmodeID.MultiplayerServer)
            packet.Broadcast(ignoreClient: fromClient);
        else
            SoundEngine.PlaySound(MySound, new Vector2(TileX * 16, TileY * 16);
    }
}
```

1. The existing workaround of making an inner `MyThingLoader : ILoadable` class and overriding `Load`
- Continues to work, but the reduced boilerplate of `ICustomAutoload` seems worthwhile.
2. Removing the `<IAutoloader>` parameter from `IAutoload` and putting the static abstract method in `IAutoload`
  - Since interfaces cannot 'override' static abstract methods, there is nowhere to put the default implementation of `ILoadable` (except separately hardcoded in `Mod.cs`)
  - Implementation would always need to exist in an abstract base `class`, which would be incompatible with `struct` use cases.
  - `ILoadable` would still be privileged and able to do things that mods cannot.
3. Similar to 2, but we make the method `static virtual` instead of `static abstract` and use reflection to look through the hierarchy of classes to find the 'override'
- Not compiler supported. Relies on special names/signatures
- Subclass overriding may make sense for `ILoadable` where the default implementation is insufficient, but not for other modded autoload scenarios.
4. Attribute based loading. Library mods use `ModSystem` or their own `ILoadable` class and scan other mods for attributes or interfaces using `GetLoadableTypes`
- Lower performance
- More difficult for library mods to implement correctly
- Loss of load order guarantees from tML
- Exceptions on loading cannot be easily attributed to the correct mod

If you're a library mod author, see the `ILoadable` code snippet above.
If you're a typical modder, see the `ExampleMod` updates below.

### ExampleMod updates
`NonAutoloadVictoryPose`, `DefaultCloudsLoader`, `ManualMusicRegistrationExample`, and `ExampleTrap` now use `ICustomAutoload`

Example:
```cs
	public class ExampleTrap : ModItem, ICustomAutoload
	{ 
		public static void Autoload(Mod mod) { 
			mod.AddContent(new ExampleTrap(0)); 
			mod.AddContent(new ExampleTrap(1)); 
		}
		
		...
```

### Porting Notes
100% backwards compatible

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using JetBrains.Annotations;
 
 namespace Terraria.ModLoader;
 
@@ -7,6 +8,10 @@ namespace Terraria.ModLoader;
 /// The corresponding <typeparamref name="TImpl"/>.<see cref="IAutoloader.Autoload(Mod, Type)"/> method will be called on all non-abstract subtypes which implement this interface
 /// </summary>
 /// <typeparam name="TImpl">Concrete type containing a static <see cref="IAutoloader.Autoload(Mod, Type)"/> implementation method</typeparam>
+[UsedImplicitly(
+	ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature,
+	ImplicitUseTargetFlags.WithMembers | ImplicitUseTargetFlags.WithInheritors
+)]
 public interface IAutoload<TImpl> where TImpl : IAutoloader
 {
 }
--- a/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
@@ -2,10 +2,18 @@
 
 namespace Terraria.ModLoader;
 
+/// <summary>
+/// Generalized marker interface for autoloading.
+/// The corresponding <typeparamref name="TImpl"/>.<see cref="IAutoloader.Autoload(Mod, Type)"/> method will be called on all non-abstract subtypes which implement this interface
+/// </summary>
+/// <typeparam name="TImpl">Concrete type containing a static <see cref="IAutoloader.Autoload(Mod, Type)"/> implementation method</typeparam>
 public interface IAutoload<TImpl> where TImpl : IAutoloader
 {
 }
 
+/// <summary>
+/// <see cref="IAutoload{TImpl}"/>. Used by <see cref="ILoadable"/> in tML
+/// </summary>
 public interface IAutoloader
 {
 	public static abstract void Autoload(Mod mod, Type type);
--- a/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
+++ b/patches/tModLoader/Terraria/ModLoader/IAutoload.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace Terraria.ModLoader;
+
+public interface IAutoload<TImpl> where TImpl : IAutoloader
+{
+}
+
+public interface IAutoloader
+{
+	public static abstract void Autoload(Mod mod, Type type);
+}
--- a/patches/tModLoader/Terraria/ModLoader/ICustomAutoload.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ICustomAutoload.cs
@@ -1,0 +1,26 @@
+using System;
+using System.Reflection;
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Provides a static abstract <see cref="Autoload(Mod)"/> method for mods to implement their own autoloading for types where <see cref="ILoadable"/> does not suffice.<br/>
+/// Sample use cases:<br/>
+/// <list type="bullet">
+/// <item>When the type has multiple instances of the same class</item>
+/// <item>To manually call some content registration methods during autoloading (eg <seealso cref="MusicLoader.AddMusic"/>)</item>
+/// </list>
+/// To implement completely custom loading behavior in a library, consider making an interface that extends from <see cref="IAutoload{TImpl}"/>
+/// </summary>
+public interface ICustomAutoload : IAutoload<ICustomAutoload.AutoloadImpl>
+{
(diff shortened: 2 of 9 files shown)


Ozzatron (approved):

Tyfyter (approved):

patches/tModLoader/Terraria/ModLoader/IAutoload.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Generalized marker interface for autoloading.
+/// The corresponding <typeparamref name="TImpl"/>.<see cref="IAutoloader.Autoload(Mod, Type)"/> method will be called on all non-abstract subtypes which implement this interface
+/// </summary>
+/// <typeparam name="TImpl">Concrete type containing a static <see cref="IAutoloader.Autoload(Mod, Type)"/> implementation method</typeparam>
+public interface IAutoload<TImpl> where TImpl : IAutoloader
```
steviegt6, line 15: Would be great to mark this interface as MeansImplicitUse for Jetbrains Annotations.

patches/tModLoader/Terraria/ModLoader/ILoadable.cs
```diff
@@ -9,7 +11,7 @@ namespace Terraria.ModLoader;
 	ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature,
```
steviegt6: Can probably remove the annotations from ILoadable and move them to IAutoload now.

patches/tModLoader/Terraria/ModLoader/ICustomAutoload.cs
```diff
@@ -0,0 +1,26 @@
...
+/// <summary>
+/// Provides a static abstract <see cref="Autoload(Mod)"/> method for mods to implement their own autoloading for types where <see cref="ILoadable"/> does not suffice.<br/>
+/// Sample use cases:<br/>
+/// <list type="bullet">
+/// <item>When the type has multiple instances of the same class</item>
+/// <item>To manually call some content registration methods during autoloading (eg <seealso cref="MusicLoader.AddMusic"/>)</item>
+/// </list>
+/// To implement completely custom loading behavior in a library, consider making an interface that extends from <see cref="IAutoload{TImpl}"/>
+/// </summary>
+public interface ICustomAutoload : IAutoload<ICustomAutoload.AutoloadImpl>
```
steviegt6, line 15: It'd be useful to explicitly re-explain that Autoload will run for each sub-type, which would not necessarily be desirable for, say, `ExampleTrap`.

steviegt6 (approved):
Rather simple changes so I'll mark it as approved.

patches/tModLoader/Terraria/ModLoader/ICustomAutoload.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+	public static abstract void Autoload(Mod mod);
+
+	public class AutoloadImpl : IAutoloader
+	{
+		private static void Invoker<T>(Mod mod) where T : ICustomAutoload => T.Autoload(mod);
+		private static readonly MethodInfo _invoker = typeof(AutoloadImpl).GetMethod(nameof(Invoker), BindingFlags.Static | BindingFlags.NonPublic);
+
+		public static void Autoload(Mod mod, Type type) => _invoker.MakeGenericMethod(type).Invoke(null, [mod]);
```
Mirsario, line 24: How plausible and rewarding do you think it would be to switch to `Autoload(Mod, ReadOnlySpan<Type>)` calls in order to reduce the amount of reflected and virtual invocations? As the amount of the former has grown extensively in this PR. Would have setup costs.

Chicken-Bones: I don't know that it would actually save much time. Feel free to followup PR. Please profile

steviegt6 (approved):

Mirsario (approved):

PR #5026, opened 2026-03-09
Fix: During SyncClientMods, related dependencies are not disabled

### LINKED ISSUE #4987: [SyncClientMods] If a client Mod depends on Both Mod, then Both Mod will be enabled
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

If a Client Mod depends on a Mod that needs to be Synced, and when the player enables this Client Mod, both this Mod and its dependent Mods will be loaded upon entering the server.

When playing with friends, I entered the server normally, but encountered type confusion and content mismatches. However, when I later attempted to reproduce the issue, error messages consistently appeared.

Therefore, sometimes I can enter the server normally, and other times errors occur.

Most localization mods are Client-side. An error occurs during mod synchronization from the previous world to the next world if the next world does not have the dependent mods enabled that this localization mod requires.

### Possible Modification
---
During mod synchronization, all client-side mods that depend on any disabled mods should be disabled.
https://github.com/tModLoader/tModLoader/blob/007e287b7bff09179df297c3906fbb154215658d/patches/tModLoader/Terraria/ModLoader/ModNet.cs#L248

reference code
https://github.com/tModLoader/tModLoader/blob/2c7a8e776aebc709949f345c2e15ed117d6627ad/patches/tModLoader/Terraria/ModLoader/ModLoader.cs#L194-L208

change to
```cs
var toDisableCache = clientMods.Where(m => m.Side == ModSide.Both).Select(m => m.Name).Except(SyncModHeaders.Select(h => h.name)).ToHashSet();
HashSet<string> toDisable = [.. toDisableCache];
foreach (var mod in toDisableCache) {
	DisableModAndDependents(mod);
}
void DisableModAndDependents(string mod)
{
	toDisable.Add(mod);

	var dependents = modFiles
		.Where(m => ModLoader.IsEnabled(m.Name) &&
				m.properties.RefNames(includeWeak: false).Any(refName => refName.Equals(mod))) //if any of its references are Both but not available on server
		.Select(m => m.Name);

	foreach (var dependent in dependents) {
		DisableModAndDependents(dependent);
	}
}
```

### Log File

_No response_

### Steps to reproduce

1. Create a Client mod that depends on a mod that needs to be Synced.
```text
modReferences = InfernumMode
side = Client
```
2. Join a server where the dependent mod is not enabled.
3. Prompt to disable and reload.
4. Successfully enter / Error occurs.

### Expected Behavior

Successfully enter the server and handle mod relationships.

### Actual Behavior

The mod that needs to be Synced gets loaded because it is referenced by the Client mod.

### Reproduction frequency

_No response_

### Additional Information

_No response_

NanTingPer:
### What is the bug?
fixes #4987 

### How did you fix the bug?
By reusing ModLoader.Load.DisableModAndDependents, the related dependencies are collected and also disabled when disabling a mod.

Directly insert the code into the location that needs to be fixed.


patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -222,6 +212,26 @@ void DisableModAndDependents(string mod)
 		}
 	}
 
+	internal static void CollectModAndDependents(LocalMod[] modFiles, string name, ISet<string> result) // Note: Recursive
+	{
+		result.Add(name);
```
Chicken-Bones: `if (!result.Add(name)) return;`

patches/tModLoader/Terraria/ModLoader/ModLoader.cs
```diff
@@ -222,6 +212,26 @@ void DisableModAndDependents(string mod)
 		}
 	}
 
+	internal static void CollectModAndDependents(LocalMod[] modFiles, string name, ISet<string> result) // Note: Recursive
```
Chicken-Bones: Lets call this, `CollectEnabledDependents`

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/ModNet.cs
```diff
@@ -245,7 +245,11 @@ internal static bool SyncClientMods(BinaryReader reader, out bool needsReload)
 		if (clientSideMods.Any())
 			Logging.tML.Debug($"Client Side mods: " + string.Join(", ", clientSideMods.Select(x => $"{x.Name} ({x.DisplayNameClean})")));
 
-		var toDisable = clientMods.Where(m => m.Side == ModSide.Both).Select(m => m.Name).Except(SyncModHeaders.Select(h => h.name));
+		var toDisableCache = clientMods.Where(m => m.Side == ModSide.Both).Select(m => m.Name).Except(SyncModHeaders.Select(h => h.name)).ToHashSet();
+		HashSet<string> toDisable = [.. toDisableCache];
```
Chicken-Bones: You should start with empty `toDisable` set here (to work with other suggested feedback of early `return`)

Then you can call `toDisableCache` something like `missingSyncMods` and it does not need to be `.ToHashSet`

NanTingPer: Thanks for the review! I've updated the PR based on your suggestion. Let me know if any further adjustments are needed.

PR #5034, opened 2026-03-14
[1.4.5] Fix `AssetRepository` Compilation Errors and Fix Hot Reloading

steviegt6:
Rather simple.

Required some extra work to reduce redundancies in `UpdateAssetValue` (`CreateUntracked<T>` produces an `Asset<T>` instance we don't need).

### DIFF
--- a/patches/tModLoader/ReLogic/Content/AssetRepository.cs.patch
+++ b/patches/tModLoader/ReLogic/Content/AssetRepository.cs.patch
@@ -115,7 +115,7 @@
  
  	public AssetValueUpdated AssetValueUpdatedHandler { get; set; }
  
-@@ -36,43 +_,49 @@
+@@ -36,43 +_,52 @@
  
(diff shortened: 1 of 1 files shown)


steviegt6:
@Chicken-Bones please make sure this is a sane use of relevant APIs in `CreateUntracked<T>`. I am presently unable to test this code (for obvious reasons) but a similar approach worked in my standalone mod that already ported over tML's asset loading.

I set `Wait` here for posterity but unsure whether it's necessary (vs just invoking the method directly):
```cs
		typedAsset.Wait = () => SafelyWaitForLoad(typedAsset, loadTask, tracked: false);
		typedAsset.Wait();
```

`Continuation` is also mutated, but state should otherwise be left alone (aside from `ownValue`).

Chicken-Bones:
Your implementation probably was sane, but unnecessarily indirect for this use case

PR #5044, opened 2026-03-20
WingGlidingSpeeds hook

Tyfyter:
### What is the new feature?
Added Mod/GlobalItem.WingGlidingSpeeds hook.

### ExampleMod updates
`ExampleWings` now have reduced gravity but increased maximum fall speed while gliding normally, and more complicated gliding when holding up.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -4880,25 +4880,29 @@
  							if (wings == 4) {
  								rocketDelay2--;
  								if (rocketDelay2 <= 0) {
-@@ -21062,7 +_,14 @@
+@@ -21062,7 +_,17 @@
  								}
  							}
  						}
 +						float gravityMultiplier = 1f / 3f;
 +						float maxSpeedMultiplier = 1f / 3f;
+ 
++						// As per method docs, true means apply vanilla logic. False means use the custom logic only
 +						if (!ItemLoader.WingGlidingSpeeds(this, ref gravityMultiplier, ref maxSpeedMultiplier))
 +							goto skipGliding;
- 
++
++						// Begin vanilla logic
 +						float gravity = this.gravity * gravityMultiplier * 3f;
 +						float maxFallSpeed = this.maxFallSpeed * maxSpeedMultiplier * 3f;
 +						
  						velocity.Y += gravity / 3f * gravDir;
  						if (gravDir == 1f) {
  							if (velocity.Y > maxFallSpeed / 3f && !TryingToHoverDown)
-@@ -21071,6 +_,7 @@
+@@ -21071,6 +_,8 @@
  						else if (velocity.Y < (0f - maxFallSpeed) / 3f && !TryingToHoverUp) {
  							velocity.Y = (0f - maxFallSpeed) / 3f;
  						}
++						// End vanilla logic
 +						skipGliding:;
  					}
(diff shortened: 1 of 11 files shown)


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -4839,6 +4839,36 @@
...
++						}
++						/*
+ 						velocity.Y += gravity / 3f * gravDir;
+ 						if (gravDir == 1f) {
+ 							if (velocity.Y > maxFallSpeed / 3f && !TryingToHoverDown)
+@@ -21071,6 +_,7 @@
+ 						else if (velocity.Y < (0f - maxFallSpeed) / 3f && !TryingToHoverUp) {
+ 							velocity.Y = (0f - maxFallSpeed) / 3f;
+ 						}
++						*/
```
Chicken-Bones: We normally use `goto` to avoid copy paste code like this

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Player.cs.patch
Tyfyter: How could I use `goto` to avoid any of this?

Chicken-Bones:
```cs
						float gravityMultiplier = 1f / 3f;
						float maxSpeedMultiplier = 1f / 3f;
						if (ItemLoader.WingGlidingSpeeds(this, ref gravityMultiplier, ref maxSpeedMultiplier))
							goto skipGliding;

						float gravity = this.gravity * gravityMultiplier * 3f;
						float maxFallSpeed = this.maxFallSpeed * maxSpeedMultiplier * 3f;

						velocity.Y += gravity / 3f * gravDir;
						if (gravDir == 1f) {
							if (velocity.Y > maxFallSpeed / 3f && !TryingToHoverDown)
								velocity.Y = maxFallSpeed / 3f;
						}
						else if (velocity.Y < (0f - maxFallSpeed) / 3f && !TryingToHoverUp) {
							velocity.Y = (0f - maxFallSpeed) / 3f;
						}
						skipGliding:;
```

Tyfyter: huh, that does make things simpler

Chicken-Bones (approved):

JavidPack:
I'm a bit confused by this PR, the hook location doesn't seem to match up with any vanilla wing logic, but appears to attempt to replicate some effects of existing wings. Is there not a better place for this hook, or a way to use the existing hooks for this effect? 

The example is a bit contradictory as well, why is it increasing max fall speed but reducing gravity? I guess my question is, "What vanilla effect are we actually trying to replicate?".

Tyfyter:
It is not an attempt to replicate any existing wings' behavior, it is an attempt to give the same control over gliding behavior that existing hooks give over ascending behavior;

Any modded wings which are intended to have even slight differences in gliding behavior would currently have to make substantial and likely incompatible IL edits;

The exact behaviors shown are just what I thought of for examples.

Solxanich:
I have updated the documentation to make it clear what is being applied based on my analysis of the existing code. 
Please confirm that I haven't misrepresented anything. 

The PR makes sense to me with these doc clarifications in mind

JavidPack:
I think the example should move to a different wing before merging, it isn't something commonly wanted and ExampleWings should be a basic example.

PR #5061, opened 2026-03-28
[1.4.5] Fix Modded Town NPCs Not Showing Up in the Housing UI

Rijam:
### What is the bug?

1.4.5 introduced a `NPCHeadID.Sets.HeadListOrder` so that vanilla Town NPCs always appeared in the same order in the housing UI. The issue was since modded Town NPCs weren't in that list, their heads wouldn't be drawn in the housing UI (their heads still worked on the map and housing banners, but not in the menu that lets you relocate them).

### How did you fix the bug?

Appended all Town NPC heads to the end of `NPCHeadID.Sets.HeadListOrder` in `NPCHeadLoader.ResizeAndFillArrays`.

This does works for Town NPCs that have multiple head variants like the Town Pets.

* I don't know if the check if the set already contains the head is necessary. I added it to prevent possible duplicate entries, but that might not be an issue. Let me know if I should remove it.
* Making `NPCHeadID.Sets.HeadListOrder` a `List<int>` instead of int array would make appending it easier, but this would require a couple of extra patches in `NPCHeadID.Sets` and `Main`. The current approach doesn't require any patches.
* I tried appending the head in `Mod.AddNPCHeadTexture` instead of `NPCHeadLoader.ResizeAndFillArrays`, but for some reason the set would reset after the mods were done loading.
* The order that they appear is based on mod load order and then alphabetically based on the class name. This is how it's always been, but maybe modders might want their own NPCs listed in a specific way. This fix doesn't facilitate that.

### DIFF
--- a/ExampleMod/Localization/en-US.hjson
+++ b/ExampleMod/Localization/en-US.hjson
@@ -134,7 +134,6 @@ Mods: {
 
 			ExampleTownPet: {
 				DisplayName: Example Town Pet
-				TownNPCMood.NoHome: *Example Town Pet noises*?
 				Census.SpawnCondition: Use an [i:ExampleMod/ExampleTownPetLicense] Example Town Pet License. Purchased from the Zoologist at 50% Bestiary completetion.
 
 				Names: {
--- a/ExampleMod/Localization/en-US.hjson
+++ b/ExampleMod/Localization/en-US.hjson
@@ -134,6 +134,7 @@ Mods: {
 
 			ExampleTownPet: {
 				DisplayName: Example Town Pet
+				TownNPCMood.NoHome: *Example Town Pet noises*?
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
This works fine for now, but I'd like your opinion on:
- Should we kill NPC head loader and NPC heads being separate things? Shouldn't an NPC just have one head that can be chosen according to conditions?
- Can we use some sort of positioning thing like we do for `ExtraJump/ModAchievement/ModMapLayer`

Rijam:
Isn't the TownNPCProfile already kind of "one head that can be chosen according to conditions?" I'm not familiar with how `ExtraJump/ModAchievement/ModMapLayer` work for positioning. I don't have any problems with how NPC heads are implemented currently, but if you wanted to overhaul it, then go ahead.

PR #5063, opened 2026-03-29
[1.4.5] Fix Modded Town NPCs' NoHome Dialogue

Rijam:
### What is the bug?
If a Town NPC is homeless, the Housing button when speaking to them has the Town NPC's NoHome quote as well as a message about valid housing. If the Town NPC is a modded one, it was trying to use the wrong localization key resulting in something like "TownNPCMood_ModName/NPCName.NoHome".

### How did you fix the bug?
If the Town NPC is a ModNPC, it uses a different localization key. This is similar to how `ShopHelper.AddHappinessReportText` changes the localization key if it is a ModNPC.

Unlikely

### TODO For the Future
Add some way to change the housing requirement text. Maybe tie it into `ModNPC.CheckConditions` which is where any housing requirement changes would be made.

### DIFF
--- a/ExampleMod/Localization/en-US.hjson
+++ b/ExampleMod/Localization/en-US.hjson
@@ -134,6 +134,7 @@ Mods: {
 
 			ExampleTownPet: {
 				DisplayName: Example Town Pet
+				TownNPCMood.NoHome: *Example Town Pet noises*?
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Please add sample dialogue to all examplemod town npcs

Rijam:
Example Person already had a NoHome quote, but I forgot to add one for Example Town Pet. Example Traveling Merchant and Example Bone Merchant don't need one because they don't move in.

Issue #5010, opened 2026-03-01
Unable to walk up slopes


maxdp66:
### Version

Latest 1.4.4-stable

### OS

Mac

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

When launching tModLoader (client or server) as ARM on macOS  with "./start-tModLoader.sh arch -arm64" I am unable to walk up slopes going left to right, going to the other side it works fine. This also applies to mobs and NPCs, when the server is started as ARM, the NPCs and mobs get stuck on the slopes.
Starting both client or server as x86_64 works as intended.

### Log File

[client.log](https://github.com/user-attachments/files/25642338/client.log)

[environment-client.log](https://github.com/user-attachments/files/25642341/environment-client.log)

[environment-server.log](https://github.com/user-attachments/files/25642346/environment-server.log)

[Launch.log](https://github.com/user-attachments/files/25642352/Launch.log)

[Natives.log](https://github.com/user-attachments/files/25642360/Natives.log)

[server.log](https://github.com/user-attachments/files/25642369/server.log)

### Steps to reproduce

1. Launch tModLoader client or server with ./executable arch -arm or ./executable -arm
2. Start a game
3. Place a block on the ground and hit it with the hammer until it's sloped from left to right
4. try to walk over
5. get stuck

### Expected Behavior

Walk over slopes as intended.

### Actual Behavior

Getting stuck, sometimes phasing partially through the blocks.

### Reproduction frequency

Every time

JavidPack:
These logs show a good number of mods enabled. Can you disable them all and then restart the game to verify that the issue happens with 0 mods enabled? Then post the new logs.

maxdp66:
[Unmodded.zip](https://github.com/user-attachments/files/25686307/Unmodded.zip)

[Calamity.zip](https://github.com/user-attachments/files/25686308/Calamity.zip)

It seems to be Calamity that causes the issue, here are logs from both TML without any mods and one with only Calamity.

JavidPack:
Well, that is certainly interesting. Calamity does seem to make edits to various player movement methods, but the mod somehow only affecting the game when launched with ARM is unexpected. I imagine the issue might be some issue with monomod support for ARM.

I'm not sure how complicated this issue will be to track down. If anyone has noticed any similar ARM specific regressions, please comment below so we can figure this out.

denbombard:
[...]

PR #5067, opened 2026-03-31
fix: MacOS Apple Silicon optimized dotnet install as default

AleksanderEvensen:
### What is the bug?
By default on Mac OS Apple Silicon devices, tModLoader downloads by default x64 build of .NET. Since Terraria is a very CPU-heavy game, running the game through Rosetta translation layer makes the performance very bad.

### How did you fix the bug?
Change the install script to detect if we're running on Apple Silicon. If we are, then set the arm flag default to true. This can be overridden by setting the newly introduced x64 flag. 

I don't think so.


Chicken-Bones:
We'll revisit this with 1.4.5 in case it has any breaking issues. For now you can supply `-arm` as a launch arg

Chicken-Bones:
relatedly https://github.com/tModLoader/tModLoader/issues/5010

PR #5071, opened 2026-04-01
1.4.4 Alert User to Changes in Visibility of Mods they are subbed to and include reupload scenario

Solxanich:
This adds hardening to alert users that a mod that still exists on their system has been hidden or removed from Steam Workshop for an unknown reason. 

This adds a capability to detect if a mod has been unlisted or banned and reuploaded to allow transfer over to the new item via querying workshop.

This increases startup times by about 200ms per 50 mods installed; the security feature is valuable.

Tested using Remnants

image

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -247,14 +247,14 @@ internal static class ModOrganizer
 		var messages = new StringBuilder();
 
 		if (abnormalWorkshopDownloads.Any()) {
-			messages.AppendLine("These mods were found installed by Steam but aren't visible on Workshop today. Delete?");
+			messages.AppendLine(Language.GetTextValue("tModLoader.RemovedWorkshopMods"));
 			foreach (var mod in abnormalWorkshopDownloads) {
 				messages.AppendLine($"  {mod.DisplayNameClean}");
 			}
 		}
 		
 		if (reuploadMDItems.Any()) {
-			messages.AppendLine("These mods were reuploaded to Steam under a new Publish ID. Transfer your subscription?");
+			messages.AppendLine(Language.GetTextValue("tModLoader.ReuploadedWorkshopMods"));
 			foreach (var mod in reuploadMDItems) {
 				messages.AppendLine($"  {mod.DisplayNameClean}");
 			}
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -204,6 +204,19 @@ internal static class ModOrganizer
 		return modPath.Contains(Path.Combine("workshop"), StringComparison.InvariantCultureIgnoreCase);
 	}
 
+	internal static HashSet<string> DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads)
+	{
+		// During initialize it forces update of CachedInstalledModDownloadItems
+		WorkshopBrowserModule.Instance.Initialize();
+		var foundMDItems = WorkshopBrowserModule.Instance.CachedInstalledModDownloadItems;
+
+		var workshopDownloads = FindWorkshopMods();
+
+		// if workshopDownloads has a mod not in foundMDItems, then what
+
+		// if foundMDItems has a mod that is newer then those in workshopDownloads, then what
+	}
+
 	internal static HashSet<string> IdentifyMissingWorkshopDependencies()
 	{
 		var mods = FindWorkshopMods();
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -1,5 +1,3 @@
-using Microsoft.CodeAnalysis;
-using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -7,12 +5,16 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
+using Microsoft.CodeAnalysis;
+using Newtonsoft.Json;
 using Terraria.Localization;
 using Terraria.ModLoader.Exceptions;
 using Terraria.ModLoader.UI;
 using Terraria.ModLoader.UI.DownloadManager;
+using Terraria.ModLoader.UI.ModBrowser;
 using Terraria.Social.Base;
 using Terraria.Social.Steam;
+using static Terraria.Social.Steam.WorkshopHelper.UGCBased;
 
 namespace Terraria.ModLoader.Core;
 
@@ -204,17 +206,61 @@ internal static class ModOrganizer
 		return modPath.Contains(Path.Combine("workshop"), StringComparison.InvariantCultureIgnoreCase);
 	}
 
-	internal static HashSet<string> DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads)
+	internal static string DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads)
 	{
 		// During initialize it forces update of CachedInstalledModDownloadItems
 		WorkshopBrowserModule.Instance.Initialize();
+
 		var foundMDItems = WorkshopBrowserModule.Instance.CachedInstalledModDownloadItems;
 
-		var workshopDownloads = FindWorkshopMods();
+		// if found installed mod download item that is newer then those in workshopDownloads and under a different publish ID
+		var reuploadMDItems = foundMDItems.Where(a => a.Installed is null);
 
-		// if workshopDownloads has a mod not in foundMDItems, then what
+		// if a local mod is installed and it doesn't have a corresponding workshop publish item AND isn't the reupload case
+		var abnormalWorkshopDownloads = FindWorkshopMods().Except(foundMDItems.Select(a => a.Installed));
 
-		// if foundMDItems has a mod that is newer then those in workshopDownloads, then what
+		if (!reuploadMDItems.Any() && !abnormalWorkshopDownloads.Any()) {
+			resolveAbnormalDownloads = null;
+			return string.Empty;
+		}
+
+		var toDeleteOldMods = abnormalWorkshopDownloads.Where(a => reuploadMDItems.Select(b => b.ModName).Contains(a.Name));
+		abnormalWorkshopDownloads = abnormalWorkshopDownloads.Where(a => !reuploadMDItems.Select(b => b.ModName).Contains(a.Name));
+
+		resolveAbnormalDownloads = async () => {
+			// Group 1: If the mod is Reuploaded, delete the old and sub to the new.
+			foreach (var mod in toDeleteOldMods)
+				DeleteMod(mod);
+
+			if (reuploadMDItems.Any()) {
+				await UIModBrowser.DownloadMods(
+					reuploadMDItems,
+					Interface.loadModsID);
+			}
+
+			// Group 2: Delete mods that originated from workshop but workshop doesn't have a replacement
+			foreach (var mod in abnormalWorkshopDownloads)
+				DeleteMod(mod);
+		};
+
+		// Messages for Users
+		var messages = new StringBuilder();
+
+		if (abnormalWorkshopDownloads.Any()) {
+			messages.AppendLine("These mods were found installed by Steam but aren't visible on Workshop today. Delete?");
+			foreach (var mod in abnormalWorkshopDownloads) {
+				messages.AppendLine($"  {mod.DisplayNameClean}");
+			}
+		}
+		
+		if (reuploadMDItems.Any()) {
+			messages.AppendLine("These mods were reuploaded to Steam under a new Publish ID. Transfer your subscription?");
+			foreach (var mod in reuploadMDItems) {
+				messages.AppendLine($"  {mod.DisplayNameClean}");
+			}
+		}
+
+		return messages.Length > 0 ? messages.ToString() : null;
 	}
 
 	internal static HashSet<string> IdentifyMissingWorkshopDependencies()
--- a/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
@@ -213,7 +213,7 @@ internal static class ModOrganizer
 		var foundMDItems = WorkshopBrowserModule.Instance.CachedInstalledModDownloadItems;
 
 		// if found installed mod download item that is newer then those in workshopDownloads and under a different publish ID
-		var reuploadMDItems = foundMDItems.Where(a => a.Installed is null);
+		var reuploadMDItems = foundMDItems.Where(a => a.IsReupload());
 
 		// if a local mod is installed and it doesn't have a corresponding workshop publish item AND isn't the reupload case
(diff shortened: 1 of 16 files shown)


solutions/TranslationsNeeded.txt
```diff
@@ -1,7 +1,8 @@
...
-es-ES 414
-de-DE 415
+zh-Hans 2
+ru-RU 3
+pt-BR 44
+pl-PL 41
+it-IT 992
+fr-FR 457
+es-ES 416
+de-DE 417
```
Chicken-Bones: Revert this file, it will create cherry picking conflicts

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -326,6 +339,19 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
...
+					try {
+						foreach (var item in checkReupload) {
+							var match2 = TryGetModDownloadItem(item, out var reupload);
+							if (match2 == WorkshopSearchReturnState.Success) {
+								items.Add(reupload);
+							}
+						}
+					}
+					finally {
+						ReleaseWorkshopQuery();
```
Chicken-Bones: why are we releasing a workshop query without a matching call to `Generate...Query` from `SteamedWraps`?

Chicken-Bones: There's already a release in `TrySearchByInternalName`

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -204,6 +205,63 @@ internal static bool IsModFromSteam(string modPath)
...
 
+	internal static string DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads)
+	{
+		// During initialize it forces update of CachedInstalledModDownloadItems
+		WorkshopBrowserModule.Instance.Initialize();
+
+		var foundMDItems = WorkshopBrowserModule.Instance.CachedInstalledModDownloadItems;
+
+		// if found installed mod download item that is newer then those in workshopDownloads and under a different publish ID
+		var reuploadMDItems = foundMDItems.Where(a => a.Installed is null);
```
Chicken-Bones: This is confusing the heck out of me.

- `CachedInstalledModDownloadItems`  uses `DirectQueryInstalledMDItems`
  - `DirectQueryInstalledMDItems` uses `GetInstalledMods`
    - `GetInstalledMods` uses `ModOrganizer.FindWorkshopMods`
   - `DirectQueryInstalledMDItems` uses `DirectQueryItems` which uses `QueryItemsSynchronously` to create `ModDownloadItem` instances for installed workshop mods
      - `ModDownloadItem.UpdateInstallState` calls `Installed = Interface.modBrowser.SocialBackend.IsItemInstalled`
        - `IsItemInstalled` uses `GetInstalledMods`, so... we're back to `FindWorkshopMods`

How is it ever possible for a workshop item which comes from a field called Cached**Installed**ModDownloadItems to not be Installed?

Solxanich: This took me down quite the journey to figure why this instinctively worked. 
In essence, we only updated 'Installed' information when `ModDownloadItem.UpdateInstallState()` is called. This includes setting it initially. 

So when a ModDownloadItem is created, the value of Installed is null. 
Added documentation around that for 1.4.4 branch.... 

So then when I added the code in QueryItemsSynchronously to call TryGetModDownloadItem, I didn't add the additional line to call item.UpdateInstallState(). 

This is why the code as included in this PR works at the stage in tModLoader it is used. 

----------------------------------------------------------------------

Now the installed cached is used by Mod Browser as well.
Which then begs the question of how does the code still know it is an update for an existing installed mod...
image

And that's because the UI for mod browser calls `UpdateInstallState` before rendering and since it calculates based on ModName / slug, it creates a weird hybrid of it being the new mod on workshop being associated with the old mod copy under the previous publishID.

-----------------------------------------------------------------------
So it works, but now that this has been flagged I do not like how it coincidentally works based on the safeguards in code today as opposed to a bit more intentional design so I've changed it

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -341,17 +341,15 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
...
-							var match2 = TryGetModDownloadItem(item, out var reupload);
-							if (match2 == WorkshopSearchReturnState.Success) {
-								items.Add(reupload);
-							}
+					foreach (var item in checkReupload) {
+						var match2 = TryGetModDownloadItem(item, out var reupload);
+						if (match2 == WorkshopSearchReturnState.Success) {
+							// We intentionally don't call UpdateInstallStatus here; this is to ensure we know that it has been reuploaded
+							// Specifically, although from a ModName/Slug standpoint it is installed, it is not installed from a PublishID standpoint.
+							reupload.UpdateInstallState();
```
Chicken-Bones: Comment seems to contradict the code?

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadItem.cs
```diff
@@ -73,6 +80,14 @@ internal void UpdateInstallState()
 		AppNeedRestartToReinstall = Installed == null && Interface.modBrowser.SocialBackend.DoesAppNeedRestartToReinstallItem(PublishId);
 	}
 
+	internal bool IsReupload()
+	{
+		if (!WorkshopHelper.GetPublishIdLocal(Installed.modFile, out var localPublishId))
+			return false;
+
+		return localPublishId.ToString() != PublishId.m_ModPubId;
+	}
```
Chicken-Bones, line 88: good

solutions/TranslationsNeeded.txt
```diff
@@ -1,8 +1,7 @@
...
-fr-FR 457
-es-ES 416
-de-DE 417
+ru-RU 1
+pt-BR 42
+pl-PL 39
+it-IT 990
+fr-FR 455
+es-ES 414
+de-DE 415
```
Chicken-Bones, line 7: still need to revert changes to this file

patches/tModLoader/Terraria/ModLoader/UI/ModBrowser/ModDownloadItem.cs
```diff
@@ -73,6 +80,14 @@ internal void UpdateInstallState()
 		AppNeedRestartToReinstall = Installed == null && Interface.modBrowser.SocialBackend.DoesAppNeedRestartToReinstallItem(PublishId);
 	}
 
+	internal bool IsReupload()
+	{
+		if (!WorkshopHelper.GetPublishIdLocal(Installed.modFile, out var localPublishId))
```
Chicken-Bones, line 84: Won't this throw if `Installed is null`?

Chicken-Bones (changes requested):
Definitely an improvement.

I don't like returning `ModDownloadItem`s which are only partially accurate from some methods and not others. Documentation is an improvement, but it is flimsy. Ideally, the `Installed` value of a `ModDownloadItem` is always correct/accurate/kept up to date.

If it's not possible (for performance reasons perhaps) to initialize them with the correct value then you could compute it on demand.

```cs
private bool _installStateInit;
private LocalMod _installed;
public LocalMod Installed { get {
    if (!__installStateInit) CheckInstallState();
    return _installed;
} }
```

If you don't want to automatically compute on demand, and want to expose the 'not computed' state to the user, then the values should be gated behind a property that requires the user to acknowledge the uninitialized state. eg

```cs
struct LocalModInstallState { LocalMod Installed; bool IsInstalled => Installed != null; }

public LocalModInstallState ? InstallState;

....

if (!modDownloadItem.InstallState.HasValue) {
    // handle/assert/call UpdateInstallState()
}

...or

var installed = modDownloadItem.GetOrUpdateInstallState();
```

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
Chicken-Bones: I added a review with more on this issue, but in response to this comment specifically:

> So then when I added the code in `QueryItemsSynchronously` to call `TryGetModDownloadItem`, I **_didn't_** add the additional line to call `item.UpdateInstallState()`.

But, `QueryItemsSynchronously` _**_does_**_ call `item.UpdateInstallState`, so why is the install state not up to date?
https://github.com/tModLoader/tModLoader/blob/1.4.4/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs#L322

Solxanich: Ok, stepping back ten feet; 
This PR is about the edge case of when the files on someone's computer becomes asymmetric with what they have access to on Steam. 
Specifically, when someone still has workshop files despite a query for the publish id returning `WorkshopSearchReturnState.NotFound`.

This happens when any of the following occurs:
1) A mod has been banned (be it DMCA or malware)
2) A mod has been "marked as incompatible"
3) A mod has been changed to Visibility = Friends Only or Private. 

So the existing code would find no `ModDownloadItem` corresponding to the `LocalMod` that Steam says the user is subbed to.

As part of security hardening, we do want to prompt when it no longer exists in case it might have been malware. 

While working on the feature, I also realized that since we have an approval system in place, I can also utilize the same changes to detect if a mod has been reuploaded in support of transferring items between authors or reupload after a DMCA.

That's the context that this PR is targeting, in alignment with the rather brief description

solutions/TranslationsNeeded.txt
Solxanich: Resolving as the PR doesn't show that there is any changes for me. 
Guessing it showed as changes in the specific commit reviewed

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -204,6 +205,63 @@ internal static bool IsModFromSteam(string modPath)
...
+		// During initialize it forces update of CachedInstalledModDownloadItems
+		WorkshopBrowserModule.Instance.Initialize();
+
+		var foundMDItems = WorkshopBrowserModule.Instance.CachedInstalledModDownloadItems;
+
+		// if found installed mod download item that is newer then those in workshopDownloads and under a different publish ID
+		var reuploadMDItems = foundMDItems.Where(a => a.IsReupload());
+
+		// if a local mod is installed and it doesn't have a corresponding workshop publish item AND isn't the reupload case
+		var abnormalWorkshopDownloads = FindWorkshopMods().Except(foundMDItems.Select(a => a.Installed));
```
Chicken-Bones: This is now `installedWorkshopModsNotFoundOnWorkshop`?

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -204,6 +205,63 @@ internal static bool IsModFromSteam(string modPath)
 		return modPath.Contains(Path.Combine("workshop"), StringComparison.InvariantCultureIgnoreCase);
 	}
 
+	internal static string DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads)
```
Chicken-Bones, line 208: Lets replace the word abnormal with `InstalledWorkshopModsNotFoundOnWorkshop`
(or `InstalledWorkshopModsMissingFromWorkshop`)

patches/tModLoader/Terraria/ModLoader/UI/Interface.cs
```diff
@@ -193,16 +193,22 @@ internal static void ModLoaderMenus(Main main, int selectedMenu, string[] button
...
+				message += ModOrganizer.DetectAbnormalSteamWorkshopDownloads(out Action resolveAbnormalDownloads);
+
 				if (missingDeps.Any()) {
 					message += $"{Language.GetTextValue("tModLoader.DependenciesNeededForOtherMods")}\n  {string.Join("\n  ", missingDeps)}";
 				}
 				message = message.Trim('\n');
 
 				bool anyMissingDependency = missingDeps.Any();
-				bool anyRemovedMod = removedMods.Any();
+				bool anyRemovedMod = removedMods.Any() || resolveAbnormalDownloads is not null;
```
Chicken-Bones, line 211: Minor concerns about overloading this UI, but they could be addressed in a followup PR:
- A mod being removed from workshop and possibly being malware is a different class of warning (and deserves red text) than the rest of this normally ignorable message
- Button label may not be correct, `resolveAbnormalDownloads` doesn't just mean 'delete a reupload and sub to new version'
- Comment in `DetectModChangesForInfoMessage` 
```cs
			//TODO: This code was added hastily in response to a popular mod being transferred ownership by reuploading it.
			// Revisit this code at a later date, as its not optimized for UX
```

patches/tModLoader/Terraria/Social/Base/SocialBrowserModule.cs
```diff
@@ -64,13 +64,17 @@ public interface SocialBrowserModule
...
 		foreach (var mod in mods) {
-			if (GetModIdFromLocalFiles(mod.modFile, out var id))
+			if (GetModIdFromLocalFiles(mod.modFile, out var id)) {
 				listIds.Add(id);
+				modSlugs.Add(mod.Name);
+			}
 		}
 
 		qParams.searchModIds = listIds.ToArray();
+		qParams.searchModSlugs = modSlugs.ToArray();
```
Chicken-Bones, line 80: A comment explaining the downstream effects of this (allows searching by mod name as a fallback) would be great

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -304,28 +304,48 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
...
 					}
 					finally {
 						ReleaseWorkshopQuery();
 					}
+
+					// Check if any of the mods flagged for reupload review are actually reuploaded or if the user has a now hidden mod
+					foreach (var item in checkReupload) {
+						var match2 = TryGetModDownloadItem(item, out var reupload);
+						if (match2 == WorkshopSearchReturnState.Success) {
+							items.Add(reupload);
```
Chicken-Bones, line 357: I assume this method has no requirements about the order of items in the output list?

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -304,28 +304,48 @@ internal List<ModDownloadItem> QueryItemsSynchronously(out List<string> missingM
...
 							var match = TryGenerateModDownloadItem((uint)j, out var item);
-							if (match != WorkshopSearchReturnState.Success) {
 
+							// Check for reupload when searchModSlugs is provided
+							if ((item.Banned || match == WorkshopSearchReturnState.NotFound)
+								&& queryParameters.searchModSlugs?.Length == queryParameters.searchModIds.Length) {
+
+								// Currently, only known case is if a mod the user is subbed to is set to hidden & not deleted by the user
+								// Includes DMCA'd mods and banned mods
+								Logging.tML.Warn($"Unable to find Mod with ID {idArray[j]} on the Steam Workshop");
```
Chicken-Bones: Warning should say 'searching via internal name {queryParameters.searchModSlugs[itemsIndex]}` to distinguish the code path taken.

Chicken-Bones: Do we want to search for re-uploads on banned mods?

Chicken-Bones (approved):

PR #5073, opened 2026-04-02
Upgrade to .NET 10 and modernize dependencies

Chicken-Bones:
### What is the new feature?

Updates tModLoader and all related projects to .NET 10 and C#14.
Updates Roslyn `Microsoft.CodeAnalysis` (for mod building, `tModPorter`, `tModCodeAssist`) to support C#14 and .NET 10
Updates `Newtonsoft.Json` to `13.0.4` to reduce conflicts with modern libraries
Replaces old `Ionic.Zip/DotNetZip` with `System.IO.Compression` (removes security warnings)

Performance, language features, and keeping up with the modern .NET ecosystem

We could avoid updating some of the deps, but I think this is best in the long term and the breakage is minimal.

### ExampleMod updates
None!

### Porting Notes
If you were using `ZipFile/Entry` from `DotNetZip`, switch to `ZipArchive/ZipArchiveEntry` from `System.IO.Compression`


patches/tModLoader/ReLogic/Utilities/ZipExtensions.cs
```diff
@@ -0,0 +1,20 @@
+using System.IO;
+
+namespace System.IO.Compression;
+
+/// <summary>
+/// Extension methods for <see cref="ZipArchive"/> and <see cref="ZipArchiveEntry"/> that provide
+/// API compatibility with Ionic.Zip (DotNetZip), reducing diffs against vanilla
+/// </summary>
+public static class DotNetZipCompatExtensions
```
steviegt6: Would it be worthwhile to make this `internal`? I believe ReLogic already exposes itself to Terraria. Less new API surface to potentially maintain for mods.

patches/tModLoader/ReLogic/ReLogic.csproj.patch
```diff
@@ -9,9 +9,9 @@
  	<ItemGroup>
 -		<Reference Include="../Terraria/Libraries/Common/Ionic.Zip.CF.dll" />
  		<ProjectReference Include="../../../FNA/FNA.Core.csproj" />
- 		<Reference Include="../Terraria/Libraries/Common/Newtonsoft.Json.dll" />
+-		<Reference Include="../Terraria/Libraries/Common/Newtonsoft.Json.dll" />
  		<Reference Include="../Terraria/Libraries/Common/SteelSeriesEngineWrapper.dll" />
-+		<PackageReference Include="DotNetZip" Version="1.16.0" />
++		<PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
```
steviegt6, line 14: Should this be replaced at the TerrariaNetCore level since it's also updated, or is it only a concern for tModLoader since it's distributed?

setup/Core/Setup.Core.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>net8.0</TargetFramework>
+		<TargetFramework>net10.0</TargetFramework>
```
steviegt6, line 3: Updating to .NET 10 causes the formatting step in decompilation to fail: 

[error.log](https://github.com/user-attachments/files/26434031/error.log)

https://github.com/dotnet/roslyn/blob/6a8ea0e2e9c1bc4f1f1fd386383bb67fa9c8643e/src/Compilers/CSharp/Portable/Syntax/CSharpSyntaxRewriter.cs#L39

This was also the case in Debug builds targeting .NET 8.

steviegt6:
Builds and runs on my machine, capable of loading mods initially built for .NET 8 (though it doesn't matter).

patches/tModLoader/ReLogic/ReLogic.csproj.patch
Chicken-Bones: I decided to keep `TerrariaNetCore` more of a minimal port

patches/tModLoader/ReLogic/Utilities/ZipExtensions.cs
Chicken-Bones: Yes

steviegt6 (approved):

JavidPack (approved):
Looks good. I tested all I could think of and it seems to all function as before.

PR #5074, opened 2026-04-04
Fix Linux multi-client logging

### LINKED ISSUE #5055: [1.4.4] [Linux] Second client instance does not create a second client log
### Version

Latest 1.4.4-stable

### OS

Linux

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

When starting multiple clients on linux, a second client log is not created, contrary to windows, where it is

### Log File

The first launched client's log file has no indication of a second instance starting, nothing unusual

### Steps to reproduce

Launch first client with steam or the start script, then a second client using the start script. The way the game is first launched does not affect the generation of the first client log, that one always appears.

### Expected Behavior

A second client log should appear in the Logs directory

Image

### Actual Behavior

no second client log appears.

### Reproduction frequency

100% With my system

### Additional Information

Any help trying to debug this issue on my end, if it isn't reproducible for anyone else, would be appreciated

RoNiN-shadow:
### What is the bug?
Fixes #5055 

This PR fixes an issue on Linux where multiple tModLoader clients write to the same `client.log` file instead of creating `client2.log`. This happens because Linux file locking is advisory by default, allowing multiple instances to append to the same file.

### How did you fix the bug?
I explicitly added `FileAccess.Write` and `FileShare.None` to the `FileStream` constructor in `Logging.cs` to enforce an exclusive lock, matching the default Windows behavior.

No.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Logging.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Logging.cs
@@ -212,7 +212,7 @@ public static partial class Logging
 	private static bool CanOpen(string fileName)
 	{
 		try {
-			using (new FileStream(fileName, FileMode.Append)) { };
+			using (new FileStream(fileName, FileMode.Append, FileAccess.Write, FileShare.None)) { };
 
 			return true;
 		}


JavidPack:
Thanks, @direwolf420 can you verify that this fixes your issue once the build completes? https://github.com/tModLoader/tModLoader/actions/runs/23988385887

Chicken-Bones:
Looks reasonable. Doesn't change the behaviour of the actual logger file handle.

direwolf420:
I can confirm it fixes the issue, tested with both launching two clients and two servers. Though I cannot verify if it causes any regression on other OSes

PR #5075, opened 2026-04-05
Backport PatchReviewer/Setup updates for 1.4.4

Mirsario:
Since PatchReviewer and DiffPatch had a reorganization that moved their projects and code into inner `PatchReviewer` in first case and into an inner `src/` folder in the latter case, switching from 1.4.5 to 1.4.4 branches is now likely to cause the following error, caused by newer `obj` folders becoming part of build processes:
image
This PR simply cherrypicks all changes to the submodule and imports in the setup project in order to avoid that. Excluding hash/gameversion bits.

### Alternatives
An alternative would be to go and make a diffpatch branch that excludes `src\` and `PatchReviewer\` from `Compile`/`None`/`Content`/`EmbeddedResource` groups, but I see no reason to choose that at the moment.

### DIFF
--- a/setup/CLI/Commands/PatchSubCommands.cs
+++ b/setup/CLI/Commands/PatchSubCommands.cs
@@ -1,5 +1,5 @@
 using System.ComponentModel;
-using DiffPatch;
+using CodeChicken.DiffPatch;
 using Spectre.Console.Cli;
 using Terraria.ModLoader.Setup.Core;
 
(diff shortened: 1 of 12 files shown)


Chicken-Bones:
I recommend a second checkout for 1.4.4 development, but this is probably fine

Mirsario:
I'm seeing no behavioral differences at the moment, both in diffing and in patching. Can rebranch if any "breaking" changes are introduced.

PR #5077, opened 2026-04-05
Fix for PR 5071... I think. Needs Non-Dev testing

Solxanich:
Fixes an issue with PR 5071 where the Access Denied state didn't return NotFound and instead returned Retrieval Failed.

As this code is unable to be tested with a developer steam account, it must be tested by a community member before merge

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -316,10 +316,10 @@ public partial class WorkshopHelper
 							// Because we are directly querying the state of the mod via its publish ID, developers are able to see that it is banned.
 							// For regular users, they will see WorkshopState NotFound. The banned state includes DMCA and Malware, so its a mix bag
 							if (match == WorkshopSearchReturnState.Success && item.Banned) {
-								// Currently, only known case is if a mod the user is subbed to is set to hidden & not deleted by the user
-								// Includes 'Hide As Incompatible', Changes in Visibility (Friends-Only, Private), and as of April 2026 banned states (DMCA, Malware)
+								// Includes as of April 2026 banned states (DMCA, Malware)
 								Logging.tML.Warn($"Mod ID {idArray[j]} is banned on Steam Workshop. Consult a developer for more details");
 								missingMods.Add(idArray[j]);
+
 								//checkReupload.Add(queryParameters.searchModSlugs[itemsIndex]); // Enable this line if doing dev testing for NotFound state in DMCA
 								continue;
 							}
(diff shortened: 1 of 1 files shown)


patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
```diff
@@ -542,6 +543,9 @@ internal WorkshopSearchReturnState TryGenerateModDownloadItem(uint index, out Mo
 					item = GenerateModDownloadItemFromQuery(index);
 					return WorkshopSearchReturnState.Success;
 				}
+				catch (UnauthorizedAccessException) {
+					return WorkshopSearchReturnState.NotFound;
```
Chicken-Bones, line 547: Since we have the ability to distinguish between `NotFound` and `AccessDenied`... are there any scenarios where we should?

Chicken-Bones (approved):

patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
Solxanich: Had to think on it. 

fundamentally, the only extra piece of information you learn from an State.AccessDenied is that the publish ID corresponds to an actual item still. Inherently, that implies you knew of the publish ID beforehand belonging to tmod, which is nigh impossible to guess without prior context since publish ids are global to steam. 
From a UI standpoint, you would never be able to find it in Steam or in-game browser.

The scenario where we might care is during multiplayer sync. 
An Access Denied state might mean something different to a player than not found within that particular context. Say for example, someone uploaded a friends-only mod and the person joining isn't a steam friend. 
But since that is done based on ModName / Slug, there is no publish IDs so it doesn't value there either.

So I can't foresee any useful information because it is too niche and contextual of a piece of information.

Solxanich:
Testing summary for documentation:

Method: 
1) Tester Subscribed to https://steamcommunity.com/sharedfiles/filedetails/?id=2593761992
2) Solxan then ban his own mod
3) Tester uploaded a new mod with the same internal name
4) Tester unsubbed from the new mod
5) Tester launching this build and verify the startup info message https://github.com/tModLoader/tModLoader/actions/runs/24004564784/artifacts/6278192089

The Remnants dev, LazyWombat, performed testing. 
This fix correctly addresses the calculation issue flagged with PR 5071. 
Tester noted that calls to 'UGC.UnsubscribeItem()' failed so they were subbed to both at end state. Local deletion of files was successful during transfer of sub.

image
image

From testing without a reupload copy - still works
image

Chicken-Bones:
> image

^ Text should probably say. Your current version will no longer be updated unless you resubscribe. Buttons should probably say. `Ignore` and `Resubscribe`.

> From testing without a reupload copy - still works image

^ Buttons here should definitely not say `Redownload Mods`. I see in #5089 you have added a `Delete or Transfer Mods` localization... I still don't think these two actions should be combined into one button.

PR #5078, opened 2026-04-05
Add Debian Steam Installation path to Linux auto-detection

### LINKED ISSUE #5076: [Setup] [Linux] add an additional path lookup for the steam directory

No

Mod capability as a Modder

### Description

The Setup tool automatically detects the steam directory to look for the Terraria game files. On linux however, it can fail if the current provided test strings fail to return a valid path. In my case and many others, one such valid path that is not covered by current test strings is `.steam/debian-installation`, appended to the user directory. Full path for clarity: `/home/myuser/.steam/debian-installation/steamapps/common/Terraria`. It would need to be included in the check below

https://github.com/tModLoader/tModLoader/blob/8b395b31628420e3d4b6b55a2642d4625cab6509/setup/Core/Utilities/SteamUtils.cs#L89-L93

Less friction with a modder trying to set up a dev environment. A current workaround is to specify the directory manually via `./setup-cli.sh update-workspace-info "my path here"`.

`.steam/debian-installation` is the default install directory for steam installed via apt on debian-based distros.

None

RoNiN-shadow:
### What is the new feature?
Resolves #5076 

Currently, when Steam is installed via `apt` on Debian-based distributions, the Terraria game files are located in `.steam/debian-installation`. Because this path was missing from the auto-detection logic, modders were forced to manually specify the directory using `./setup-cli.sh update-workspace-info <path>`. 

This PR adds a fallback check for the `.steam/debian-installation` directory if the default `.local/share/Steam` and `.steam/steam` paths do not exist.

It significantly reduces friction for modders setting up a development environment on Debian/Ubuntu by removing the need for a manual workaround.

No. Appending the specific path to the existing fallback chain is the most direct and reliable solution.

Simply running `./setup-cli.sh setup` will now automatically find the game files on Debian/Ubuntu distributions without requiring any manual path configuration.

### DIFF
--- a/setup/Core/Utilities/SteamUtils.cs
+++ b/setup/Core/Utilities/SteamUtils.cs
@@ -90,6 +90,8 @@ namespace Terraria.ModLoader.Setup.Core.Utilities
 				path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local/share/Steam");
 				if (!Directory.Exists(path))
 					path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".steam/steam");
+                if (!Directory.Exists(path))
+					path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".steam/debian-installation");
 			}
 
 			return path != null && Directory.Exists(path);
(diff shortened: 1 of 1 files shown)


direwolf420:
Typo in PR title "Daebian", also for readability there should be a newline above the code you added, other than that, if you tested that it works, LGTM

RoNiN-shadow:
To be honest, I haven't been able to run a full test locally for two reasons:

1. I am on Arch, so I don't natively have the Debian `apt` Steam directory structure.

2. My local `setup-cli.sh` is currently blocked by a DRM decryption error (`Cannot decrypt Terraria Windows executable`) which I am still trying to troubleshoot on my end.

However, since this is just adding a safe fallback `if` check, it shouldn't affect the existing detection logic for other users.

Could someone with a working Debian/Ubuntu verify this?

PR #5079, opened 2026-04-05
[1.4.5] Fix ModBiome

lost-werewolf:
### What is the bug?
Modded biomes aren't functional whatsoever, and exceptions may be thrown when modded music is played. Additionally, `ScreenShaderData` lacks null checks from 1.4.4.9, so modded shaders may outright crash the game.
### How did you fix the bug?
A missing patch from 1.4.4 which called `BiomeLoader.UpdateBiomes` has been re-added in an equivalent location. Arrays that need to be resized for modded biome & audio support have been resized. ~`ScreenShaderData` has also had its null checks re-added.~
N/A - While I personally can't think of any, there may always be a better approach somewhere.

Should note: the rejected player patches needs updating, since `LoaderManager.Get<BiomeLoader>().UpdateBiomes(this);` has been (hopefully) migrated correctly. If this is something I should do, let me know and I'll get on it (though, an explanation as to how would be appreciated if so!).


Chicken-Bones:
A proposed alternative is to change `ShaderData.cs:36` from `return null;` to `return new (t => {});`

Can you test this out for me and tell me if you think that has any downsides? If it works fine, then I'll commit it directly and we can focus on `ModBiome` in this PR.

lost-werewolf:
> A proposed alternative is to change `ShaderData.cs:36` from `return null;` to `return new (t => {});`
> 
> Can you test this out for me and tell me if you think that has any downsides? If it works fine, then I'll commit it directly and we can focus on `ModBiome` in this PR.

This seems to work fine on its own. I am not very savvy with shaders, so I'm unsure if there's any downsides from doing this. Would be good to note this change to modders, just in case it breaks anything (though this seems unlikely it would).
Let me know if I should remove the patches I made to ScreenShaderData from this PR.

JavidPack:
I think a separate PR for that sounds like a good idea, just to keep things organized and so we can review it specifically.

Chicken-Bones:
I have applied the `ShaderData` change upstream, so you can revert those changes and merge the latest `upstream/1.4.5`

lost-werewolf:
I removed the patches made to ``ScreenShaderData``. Visual Studio didn't have the option to merge from ``upstream``, so I synced my fork's 1.4.5 branch and then merged that, hopefully that was the correct way to do it. Still wrapping my head around branch management as I've never really needed to merge branches in VS until now. Can revert that if it was the wrong way to go about it.

JavidPack:
> I removed the patches made to `ScreenShaderData`. Visual Studio didn't have the option to merge from `upstream`, so I synced my fork's 1.4.5 branch and then merged that, hopefully that was the correct way to do it. Still wrapping my head around branch management as I've never really needed to merge branches in VS until now. Can revert that if it was the wrong way to go about it.

Looks like you did it right.

> Should note: the rejected player patches needs updating, since `LoaderManager.Get<BiomeLoader>().UpdateBiomes(this);` has been (hopefully) migrated correctly. If this is something I should do, let me know and I'll get on it (though, an explanation as to how would be appreciated if so!).

So, what you would do is open the .cs.rej file, scroll down to the patch, then delete that patch by deleting the lines from the @@ line down to but not including the next @@ line.

image

If you are around to do that, and want to do it for the experience, it shouldn't take more than a minute. I can do it too, just let me know.

Once either you or I do that this PR should be good to merge.

lost-werewolf:
> So, what you would do is open the .cs.rej file, scroll down to the patch, then delete that patch by deleting the lines from the @@ line down to but not including the next @@ line.

>If you are around to do that, and want to do it for the experience, it shouldn't take more than a minute. I can do it too, just let me know.

Done as shown in the screenshot, thanks!

PR #5080, opened 2026-04-05
[1.4.5] Fix mannequins

Jadams505:
### What is the bug?
Mannequins in 1.4.5 have an extra mount and misc slot that was not present in 1.4.4. When using a mannequin the misc slot would not save properly when leaving the world and rejoining.
### How did you fix the bug?
Add the necessary patches and networking to make the misc slot work properly.
No

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.TML.cs
@@ -10,6 +10,7 @@ public partial class TEDisplayDoll
 	{
 		tag["items"] = PlayerIO.SaveInventory(_equip);
 		tag["dyes"] = PlayerIO.SaveInventory(_dyes);
+		tag["misc"] = PlayerIO.SaveInventory(_misc);
 	}
 
 	public override void LoadData(TagCompound tag)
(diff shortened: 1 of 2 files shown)


patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.TML.cs
```diff
@@ -10,41 +10,52 @@ public override void SaveData(TagCompound tag)
...
 
 	public override void LoadData(TagCompound tag)
 	{
 		// TML 1.4.4 saved as "items", 1.4.5 changed field from _items to _equip.
 		PlayerIO.LoadInventory(_equip, tag.GetList<TagCompound>("items"));
 		PlayerIO.LoadInventory(_dyes, tag.GetList<TagCompound>("dyes"));
+		PlayerIO.LoadInventory(_misc, tag.GetList<TagCompound>("misc"));
 	}
 
 	public override void NetSend(BinaryWriter writer)
```
Chicken-Bones, line 24: How would you feel about refactoring this to use the tML `BitReader/Writer` classes, that way we can have all `9*3 = 27` bits in a single array and just read them out in the same order we write them

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/GameContent/Tile_Entities/TEDisplayDoll.TML.cs
Jadams505: Sure I can do that.

PR #5084, opened 2026-04-07
tModPorter AddParameter Feature, port for new parameter of Main.GetPlayerArmPosition

JavidPack:
Adds `AddParameter` tModPorter code to a new parameter for a changed method, in this case the new `Player` parameter for `Main.GetPlayerArmPosition`.

```diff
-Vector2 playerArmPosition = Main.GetPlayerArmPosition(Projectile);
+Vector2 playerArmPosition = Main.GetPlayerArmPosition(Projectile, default /* New parameter: Now expects a Player object, try 'player'. */);
```

After implementing, I realized this probably isn't necessary or useful, since rather than an error telling the modder to fix the code, the ported fix will actually compile and lead to silent errors. Maybe these sort of changes don't need porting, the compilation error message speaks for itself. 

I've made a branch with the code anyway, since maybe it will prove useful for some other porting we will want to do later. Or, we could keep the comment and remove the `default` parameter, but that might complicate things with how the code will behave when running tModPorter multiple times on the same code.

### DIFF
--- a/ExampleMod/Content/Projectiles/ExampleAdvancedFlailProjectile.cs
+++ b/ExampleMod/Content/Projectiles/ExampleAdvancedFlailProjectile.cs
@@ -433,7 +433,7 @@ namespace ExampleMod.Content.Projectiles
 
 		// PreDraw is used to draw a chain and trail before the projectile is drawn normally.
 		public override bool PreDraw(Player player, ref Color lightColor) {
-			Vector2 playerArmPosition = Main.GetPlayerArmPosition(Projectile);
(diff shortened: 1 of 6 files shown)


Tyfyter:
I would expect that these sorts of changes often have some consistent change which would achieve the previous behavior, such as copying the first parameter and using its owner, maybe that should be done instead?

Chicken-Bones:
Thanks for drawing my attention to a stupid function. It doesn't even use the `proj` parameter, so I'm moving it to `Player` as an instance method.

JavidPack:
Closing: 1.4.5.8 now has a `Player.GetArmPosition` method. We'll need to handle that once we update patches, but this PR doesn't have any more utility.

Issue #5086, opened 2026-04-10
NPC to NPC Damage simplification / Parity


w-man12:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

There are a lot of complexities in developing npc's capable of hitting each other, as well as certain features that are missing, this feature proposal comes in 3 parts, first, add a new NPC.StrikeOtherNPC hook to act as a simplification of the following code (note this is an excerpt from a mod I work on so some inputs are hard coded here):

                        NPC.HitModifiers modifiers = Target.GetIncomingStrikeModifiers(DamageClass.Default, NPC.direction, default);
                        NPCLoader.ModifyHitNPC(NPC, Target, ref modifiers);
                        NPC.HitInfo hitinfo = modifiers.ToHitInfo(Damage, false, knockback, true);
                        Target.StrikeNPC(hitinfo, false, true);
                        NetMessage.SendStrikeNPC(NPC, hitinfo);
                        NPCLoader.OnHitNPC(NPC, Target, hitinfo);

This should include properties for the Target, Damage, Knockback, Hitdirection, and rather or not the hit is a crit, has randomization, and the luck stat (really just every variable used by these functions)

Secondly and what I think is most neccesary is that the Dryad Ward and Bee's self damaging effects should be moved to NPCLoader.OnHitNPC(NPC npc, NPC target, NPC.hitinfo hit), with an OnHitByNPC hook being added to ModNPC

Thirdly and most importantly though least relatedly add a Warning / Note / whichever to state clearly that NPC.CalculateHitInfo doesn't include entity related hit modifying hooks and shouldn't be used in normal contexts.

### Which (other) solutions should be considered?

If adding an independent function for this is too unnecessary then more detailed instructions as to how to make NPCs hit each other (possibly included as part of issue 3) would suffice for the first listed issue

w-man12:
also I don't think town npc Melee applies modifiers to the npcs they hit, so probably include that as well

Chicken-Bones:
I am willing to accept and review a PR

PR #5087, opened 2026-04-11
Add NPC.StrikeOtherNPC helper method

RoNiN-shadow:
### What is the new feature?
Implements part 1 and 3 of #5086.

1. Adds new `NPC.StrikeOtherNPC` helper method to simplify writing code
2. Adds a XML remark to `NPC.CalculateHitInfo` warning that it doesn't include entity-related hit modifying hooks.

Reduces boilerplate code when making NPCs hit each other.

N/A - Implemented exactly as requested in the issue.e

```csharp
// Simplify 6 lines of boilerplate into a single call
NPC.StrikeOtherNPC(NPC target, int damage, float knockback, int hitDirection, bool crit, bool randomizeDamage);
```

### DIFF
--- a/patches/tModLoader/Terraria/NPC.cs.patch
+++ b/patches/tModLoader/Terraria/NPC.cs.patch
@@ -2402,7 +2402,7 @@
  	public int checkArmorPenetration(int armorPenetration, float armorPenetrationPercent)
  	{
  		if (ichor)
-@@ -64591,26 +_,165 @@
+@@ -64591,26 +_,169 @@
  
  		return armorPenetration / 2;
  	}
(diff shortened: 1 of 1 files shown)


Chicken-Bones (approved):
Generally good. @JavidPack to review docs

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -2623,8 +2627,32 @@
 -		return 0.0;
 +		return 0;
  	}
++
++    ///<summary>
++    /// A helper method for an NPC striking another NPC
++    /// Handels modifiers, hit info, network, syncing, and hooks automatically.
```
direwolf420: > network, syncing

Either remove one of these, or clarify what you mean by each, because I think it's the same thing

RoNiN-shadow: thanks, I removed duplication

direwolf420: Missed another one, oops
Handels -> Handles

RoNiN-shadow: Alright, fixed it :)

PR #5088, opened 2026-04-11
[1.4.5] Restore functionality for multiple required crafting tiles

wr7:
### What is the new feature?
Restores functionality for multiple required crafting tiles

This brings parity to how crafting stations work in 1.4.4

- To keep the changed behavior from 1.4.5 and then to document it to mod developers.
- ~Filtering recipes by station could also work differently. Currently, it shows all recipes that require at-least one of the stations filtered. However, it could instead only show recipes where all of the required stations are in the filter.~

### Porting Notes

None AFAIK

### Notes
- This change builds, ~but I have not yet been able to test this feature. I plan on testing it when I have more time.~
- I've tested the following things, and they seem to work properly:
   - Recipes that require 0, 1, or 2 different tiles
   - Looking up recipes with the Guide
   - Filtering by crafting station
   - Walking away from crafting stations with the new crafting menu open (which will cause the recipes to gray out and display their missing requirements).


JavidPack:
We can probably manage `requiredTile` by making it a private property, letting the vanilla code still access it so we don't have to change all the vanilla recipes.

As for this PR, we haven't discussed or come to a consensus whether or not to restore multiple crafting stations yet, but if we do this is a good start.

wr7:
> We can probably manage `requiredTile` by making it a private property, letting the vanilla code still access it so we don't have to change all the vanilla recipes.

Just changed it to be a property.

Chicken-Bones:
The patches for this are relatively small, but I do wonder whether it's better if mods follow the new vanilla paradigm. Having multiple crafting stations required in range just makes those recipes less discoverable and requires placing more stations all together.

direwolf420:
I vote for not having multiple crafting stations

Tyfyter:
Multiple crafting station support is trivial to reimplement _poorly_, so some mods will almost certainly keep using multiple, but the recipes will only show up when they click on one crafting station (although with that _still_ showing recipes crafted by hand, that doesn't matter much)

wr7:
I don't really care whether we decide to keep this functionality or not, but just to play devil's advocate, here's a reason why we should preserve the old behavior:

In 1.4.4, these were the only recipes that required multiple tiles:
- Table + Chair:
   - All watches
- Workbench + Chair:
   - Goggles
   - Sunglasses

In 1.4.5, they were all changed to require just a workbench. Notably, in 1.4.4, the table and chair are not crafting stations by themselves. They only become one when used together.

I can see how a mod may have a crafting station composed of multiple, separate tiles which only work when placed next to eachother (such as Table+Chair in vanilla Terraria). I'm not sure how many mods do this, but I think it still might be nice to preserve this behavior due to the low complexity of doing so.

wr7:
Alright, I just dusted off my Windows laptop to test this out. After fixing a bug, everything seemed to work (viewing recipes from the guide, crafting station filters, and recipes that require 0, 1, or 2 different tiles).

It seems like crafting station filtering is done by right-clicking the crafting station that you want to filter by, and it only lets you filter by one station at a time. This means that my approach of "the filter shows all items that either require no station or at least the filtered station" works and is the only sane design that doesn't change vanilla behavior.

One thing that is outside of the scope of this PR is that vanilla crafting stations show an icon when you hover your mouse over them to signify that you can right-click them, but if you add a recipe with your own custom station, you can still right-click it to filter recipes, but its icon will not appear when you hover your mouse over it.

Chicken-Bones:
> I don't really care whether we decide to keep this functionality or not, but just to play devil's advocate, here's a reason why we should preserve the old behavior:
> 
> In 1.4.4, these were the only recipes that required multiple tiles:
> 
> * Table + Chair:
>   
>   * All watches
> * Workbench + Chair:
>   
>   * Goggles
>   * Sunglasses
> 
> In 1.4.5, they were all changed to require just a workbench. Notably, in 1.4.4, the table and chair are not crafting stations by themselves. They only become one when used together.

In 1.4.5 the standard is that every crafting station is interactable to bring up the window with its filter. It would be rather annoying if tables and chairs were now interactable, just to never show any extra recipes unless you had both table and chair, and the required materials for a watch. It also adds more work for the guide to show all the required crafting stations and isn't worth it just for the tiny by of realism for lens/watchmaking.

wr7:
> > I don't really care whether we decide to keep this functionality or not, but just to play devil's advocate, here's a reason why we should preserve the old behavior:
> > In 1.4.4, these were the only recipes that required multiple tiles:
> > 
> > * Table + Chair:
> >   
> >   * All watches
> > * Workbench + Chair:
> >   
> >   * Goggles
> >   * Sunglasses
> > 
> > In 1.4.5, they were all changed to require just a workbench. Notably, in 1.4.4, the table and chair are not crafting stations by themselves. They only become one when used together.
> 
> In 1.4.5 the standard is that every crafting station is interactable to bring up the window with its filter. It would be rather annoying if tables and chairs were now interactable, just to never show any extra recipes unless you had both table and chair, and the required materials for a watch. It also adds more work for the guide to show all the required crafting stations and isn't worth it just for the tiny by of realism for lens/watchmaking.

The table+chair thing is just an example. I think that it's good, actually, that they changed the recipes. I just wanted to show that the multiple tiles thing is useful for things other than recipes that require multiple crafting stations and can instead be for crafting stations that are composed of multiple, independent tiles. Adding recipes that require a table, chair, or other tile will already make them interactable (regardless of this PR).

Also, this PR already includes code that allows the guide to show multiple crafting tiles for a recipe; AFAIK this PR is fully comprehensive.

I could, however, see how this PR may not be worth it due to it being a potential vector for bugs. I have tested it pretty well, but you never know.

hezuo1hao:
My Mod contains some designs that require multiple synthesis stations, either to meet the background settings or to appear more reasonable 
So i vote for having multiple crafting stations

PR #5089, opened 2026-04-12
Cleanup Startup Info Message System - Split the 'New/Updated/missing' from other systems.

Solxanich:
As a followup to PR #5071, it was identified that there is too many systems running through a single UI interface. 

In this PR, we changed to have two UI info message.
First, more commonly encountered:
1) New and updated Mods
2) Missing Mods (mods removed since last launch that may have been in error or unexpected)

Second: (rare UI)
1) Missing Dependency Check
2) Asynchronous Steam mod states; both the reupload scenario and no-access scenario

Due to the unique code path that is used in UI load mods to trigger these info messages in cascading order without getting stuck in a loop, a GoTo statement is used to bypass issues associated with the scenario where there is no info message and it needs to cascade in to the next UI directly. 
When we display UI InfoMessage the return from that message always restarts the loop so no additional action is needed in that scenario.


Solxanich:
Relevant comment to address under this PR:
https://github.com/tModLoader/tModLoader/pull/5077#issuecomment-4231422369

Solxanich:
Current State of the UI Chain:
1.1) Updated / New Mod:
image

1.2) Missing Mods from last session:
image

2.1) Mod is no longer on Steam but you have the files:
image
 
2.2) Reuploaded mod
image

3) Missing Dependencies
image

Chicken-Bones:
> 2.1) Mod is no longer on Steam but you have the files:
> image

Why would this say `Fix My Subscriptions`? In my mind, neither of those buttons mean `Delete my local copy of the mod`.
They should say something like `Keep Installed`, `Delete Mods`

> 2.2) Reuploaded mod
> image

Does this need to say `(Reupload)` next to each entry now?
Maybe `may disappear` -> `may be automatically removed by Steam`

Otherwise good.

Solxanich:
Revised 2.1. Don't know why the vertical alignment is different on the buttons, but that's a different problem unrelated to this PR.
image

Revised 2.2.  `Remnants (Reupload)` just happens to be the Display Name of the reuploaded copy.
image

PR #5090, opened 2026-04-12
[1.4.5] Implemented tModPorter Journeymode-check changes

noahDP1911:
This is my first merge-request to this project, therefore i am uncertain if followed the code and usage guidelines correctly. Or even if my code is adequate. If someone could give some feedback on these changes, it would be greatly appreciated. Afterwards, i would mark it as ready for review.

### What is the bug?
For the 1.4.5 API changes, `tModPorter` should rewrite `Main.GameModeInfo.IsJourneyMode` to `Main.IsJourneyMode`.
The migration entry for this existed, but the config-based `GameModeInfo` refactor did not actually apply to the full access chain, so the old 1.4.4-style was left unchanged.

### How did you fix the bug?
I replaced the unused `GameModeInfo` member refactor with a localized rewrite in `MemberUseRewriter.VisitMemberAccessExpression` that matches the full `...GameModeInfo.IsJourneyMode` access and rewrites it to the 1.4.5 property:
  - `Main.GameModeInfo.IsJourneyMode` to `Main.IsJourneyMode`
  - `Terraria.Main.GameModeInfo.IsJourneyMode` to `Terraria.Main.IsJourneyMode`
  - `global::Terraria.Main.GameModeInfo.IsJourneyMode` to `global::Terraria.Main.IsJourneyMode`

I also removed the unused `RefactorStaticMember("Terraria.Main", "GameModeInfo", ...)` registration from `Config.Terraria` and added a regression test covering unqualified, qualified, and `global::`-qualified forms.

The alternative could be trying to keep this in the config-driven `RefactorStaticMember` path, but this migration needs visibility of the full member-access chain, not just the `GameModeInfo` identifier. Handling it in `VisitMemberAccessExpression` is the smaller and more direct fix probably.

### DIFF
--- a/tModPorter/tModPorter.Tests/TestData/IsJourneyModeTest.Expected.cs
+++ b/tModPorter/tModPorter.Tests/TestData/IsJourneyModeTest.Expected.cs
@@ -1,0 +1,11 @@
+using Terraria;
+
+public class IsJourneyModeTest
+{
+	void Method()
+	{
+		bool isJourney = Main.IsJourneyMode;
+		bool isJourneyQualified = Terraria.Main.IsJourneyMode;
+		bool isJourneyGlobalQualified = global::Terraria.Main.IsJourneyMode;
+	}
+}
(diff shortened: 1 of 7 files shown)


JavidPack:
Thanks for this. 

In reviewing, I managed to find an approach that fixes other similar cases of qualified member access issues and avoids hardcoding specific checks for this specific refactor.

PR #5096, opened 2026-04-15
[1.4.5] Town NPC Chat Buttons

Rijam:
### What is the new feature?
Updates the Town NPC chat buttons to the new system that was added in 1.4.5.

#### What 1.4.5 changed

* There is a new `NPCInteraction` system that registers which chat buttons are shown on which Town NPCs and when.
  * The display name of the button, the condition for the button showing, and the action when clicking on the button are easily customizable. A new 1.4.5 feature also allows for a small exclamation point to be shown next to the button.
* The new system supports up to 12 buttons shown at one time.
  * 13 or more buttons and the system breaks down due to there not being defined gamepad UI links.

#### New features/changes

* Increases and caps the number of buttons that can be shown at one time to 32. Any button past the 32nd button will not be shown.
* `ModNPC.RegisterChatButtons(NPCInteractionList interactions)`
  * The modder has control of where/what order they want the buttons in.
    * `interactions.Append(NPCInteraction interaction)`, `interactions.Prepend(NPCInteraction interaction)`, `interactions.InsertAfter(NPCInteraction interactionToRegister, NPCInteraction interactionAfter)`, or `interactions.InsertBefore(NPCInteraction interactionToRegister, NPCInteraction interactionBefore)`
      * Append will add the button to the end of the list (after the Happiness and Housing buttons, too).
      * Prepend will add the button to the beginning of the list.
      * InsertAfter and InsertBefore will add the button after or before another specified button.
      * Each method returns the supplied NPCInteraction as an `NPCInteractionList.Entry` if caching the result is desired for one of the below methods.
    * `interactions.Disable(NPCInteraction interaction)` can remove a button. It needs to be the exact instance.
    * `interactions.FindInteractionByType(NPCInteraction interaction)` or `FindInteractionByType(Type interaction)` can be used to search for a specific interaction. The search matches the class type (any Close button will match any other Close button).
      * Try versions of these methods exist that returns a bool if found: `TryFindInteractionByType(NPCInteraction searchInteraction, out NPCInteractionList.Entry foundInteraction)` and `TryFindInteractionByType(Type searchInteraction, out NPCInteractionList.Entry foundInteraction)`
    * `insteractions.FindInteractionByInstance(NPCInteraction interaction)` can be used to search for a specific interaction. The search matches the exact instance.
      * Try version of this method exists that returns a bool if found: `TryFindInteractionByInstance(NPCInteraction searchInteraction, out NPCInteraction foundInteraction, out int index)`
    * `interactions.GetInteractions()` returns a `List<NPCInteraction>` with all of the buttons.
    * `interactions.GetInteractionEntries()` returns a `List<NPCInteractionList.Entry>`.
  * Registration implicitly assigns the button to the NPC.
  * The Close, Happiness, and Housing buttons are automatically registered before RegisterChatButtons runs.
    * If the NPC is a Town Pet, the Pet button will be added, too.
    * The Close, Happiness, Housing, and Pet buttons are already predefined and can be used with `NPCInteractionDatabase.CloseButton`, etc.
  * Use `NPCInteractions.Shop(string shopName = "Shop", string customTextKey = null)` to assign shops.
    * The shopName must match the shopName for the `NPCShop`.
    * Example: ModNPC calling `interactions.Append(NPCInteractions.Shop())` will register a button for the shop with full name "ModName/ModNPCName/Shop".
    * Vanilla shops can easily be added with the full name string "Terraria/Merchant/Shop". ("Decor" for the Painter's second shop)
    * Shops from other modded Town NPCs can be added with the full name string "ModName/ModNPCName/ShopName". For example: "ExampleMod/ExamplePerson/Shop".
    * Alternatively, use `NPCShopDatabase.GetShopName(NPCID, "Shop")` to get the full shop name for an NPC.
  * Use `interactions.X(new NPCInteraction...)` to register other buttons.
    * Example: `interactions.Append(new NPCInteractions.Actions.CloseChat());` 
* Create a class inheriting `NPCInteraction` to create custom buttons.
  * Added docs for `NPCInteraction`
  * The `GetText()`, `Condition()`, and `Interact()` must be defined.
    * Since the buttons are assigned to NPCs, there is no need to set the Condition to be only for specific NPC (no need for `TalkNPCType == ...`)
  * `ShowExcalmation` can be set to true to display a for a small exclamation point to be shown next to the button.
  * `TryAddCoins(ref Color chatColor, out int coinValue)` can be used to display a coin count next to the button.
  * *new* `TextColor(ref Color chatColor, ref Color chatColorShadow, bool hoveringOverButton)` can be used to modify the color of the button.
 * Restored `Mod|GlobalNPC.OnChatButtonClicked` and `GlobalNPC.PreChatButtonClicked`.
   * Note: the hooks do not run when the buttons of a sign are clicked.
 * Added `GlobalNPC.RegisterChatButtons(NPC npc, NPCInteractionList interactions)`
 * The hint text explaining what a valid house is in the Housing button can be customized. If the localization key `Mods.ModName.NPCs.NPCName.HousingText.HousingRequirements` exists for the NPC, it will automatically use that over the default text.
   * `{NPCName}` placeholder can be used for the NPC's full name.

#### Technical changes
* `NPCInteractions.Actions.OpenShop` changed to allow for vanilla integer shop IDs, single strings, and full shop name strings.
* `NPCInteractions.All` made internal (from public) so modders don't accidentally use it since it does nothing now.
* *Minor vanilla bug*: Added a safety check for `Terraria.GameContent.NPCInteraction.TalkNPC` otherwise it would create a silent error message when the player opened a sign (since there is no NPC).
* The drawing of the chat panel needed to be updated to support 32 buttons.
  * The vanilla drawing makes the panel really tall when there are a lot of buttons. Some numbers were changed to make the visual appearance better. Vanilla NPCs are unaffected because most vanilla NPC don't have more than 4 buttons showing at a time. 
  * Added more gamepad UI links to support up to 32 buttons. Theoretically 100 buttons could be supported.
* Added `TextColor(ref Color chatColor, ref Color chatColorShadow, bool hoveringOverButton)` to `NPCInteraction`
* Updated `NPC.CanShowHomelessText` to check for `ModNPC.TownNPCStayingHomeless` and return false.

The method for registering the buttons was inspired by how NPCLoot and the ItemDropDatabase work.
* Created `ModLoader.NPCInteractionList`
  * Used in the `Mod|GlobalNPC.RegisterChatButtons` hooks.
  * Conaints a `List<NPCInteractionList.Entry>` with all of the interactions for the specified NPC.
  * Contains the methods explained above.
* Created `ModLoader.NPCInteractionDatabase`
  * Contains a `Dictionary<int, NPCInteractionList>` that holds all the interaction data.
  * `NPCInteractionDatabase.Populate()` used to initialize and register all the buttons.
    * Registers the buttons for signs and all of the vanilla NPCs.
    * Registers the buttons for ModNPCs.
* Added `Main.NPCInteractionDB` to initialize the NPCInteractionDatabase.

The old system for assigning a first and second button for Town NPCs was removed, and thus modded Town NPCs couldn't assign any buttons.

Certainly:
* ~~Controlling the order of the chat buttons.~~ DONE!
  * Vanilla registers the buttons in a very specific order. Shops, collecting taxes, and nurse healing will come before the Close button, then every other type of button, with Happiness and Housing buttons being last.
  * The issue is that if modded Town NPCs register their buttons after vanilla does, the Close, Happiness, and Housing buttons will come before any buttons the modded Town NPC has.
    * ~~This was solved by editing those buttons to only assign to specific vanilla Town NPC instead of all Town NPCs at once. Then, when loading the modded Town NPCs, those buttons are added at the end.~~
    * ~~But wait, that means the Close button will come after any other buttons the modded Town NPC has instead of only after shops like vanilla NPCs have. So, I added ref bools to skip assigning those three buttons so the modder can do it themselves.~~
    * ~~Using the new `GlobalNPC.SetChatButtons` means the buttons you add will always be at the end of the list.~~
  * ~~I think having some sort of way to edit the order of the buttons would be beneficial. Maybe a `List<NPCInteraction>` parameter that you can edit or something.~~ Done!
* ~~The current implementation of `GlobalNPC.SetChatButtons` is maybe not that pretty (on top of the ref bool skipping). See `InitializeVanillaGlobal()` in `NPCInteractions.TML.cs`~~
* Assigning an invalid shop string name will open an empty shop. Maybe not the worst thing. At least it doesn't crash or something. I think this is how it worked in 1.4.4, too.
  * Added a chat and logs warning for an invalid shop name.
* ~~Since chat buttons are assigned per NPC now, I was thinking of genericizing all of the vanilla NPCInteractions.Actions to allow modders to use the vanilla buttons on their Town NPCs.~~
  * Example: If you wanted to assign `TinkererReforge` to your NPC, it won't show up because its condition is to only show for the Goblin Tinkerer. I was thinking of changing the condition to show for any NPC.
  * Right now, you could create a new NPCInteraction class that inherits the button you want and override the Condition to get it to work.
* `NPCInteraction.ShowExcalmation` is spelled wrong. I've already pointed this out for 1.4.5.7, but I was wondering if we should rename it right now.
* Signs are registered to NPC ID 0. I chose 0 because `NPCInteraction.TalkNPCType` returns 0 if the player is not talking to an NPC. Maybe it should it be -1?
* ~~When initializing the buttons for the Town Pets/Slimes, I skipped registering a happiness button. Similarly, the Old Man, Traveling Merchant, and Skeleton Merchant didn't get happiness and housing buttons. Technically, the vanilla system would've registered those buttons to them even if they never show up for them. I was wondering if I should register those buttons to them to match vanilla and for consistency.~~ Done
* The names of all of the methods could be changed, as usual.

```cs
// In the ModNPC:
public override void RegisterChatButtons(NPCInteractionList interactions) {
	// Here is how to register chat buttons to your NPC.
	// There are many method that you can use to change the order of the buttons.
	// interactions.Append(NPCInteraction interaction)	This will add the button to the end of the list (after the Happiness and Housing buttons, too).
	// interactions.Prepend(NPCInteraction interaction)	This will add the button to the beginning of the list.
	// interactions.InsertAfter(NPCInteraction interactionToRegister, NPCInteraction interactionAfter)		This will add the button after another specified button.
	// interactions.InsertBefore(NPCInteraction interactionToRegister, NPCInteraction interactionBefore)	This will add the button before another specified button.
	
	// In this example we are registering our Shop button to before the Close button.
	// The Close button instance is provided for us for convenience.
	interactions.InsertBefore(NPCInteractions.Shop(ShopName), NPCInteractionDatabase.CloseButton); // NPCInteractions.Shop() is a helper that creates a Shop button.
	
	// Next, add the rest of our buttons before the Happiness button (which is before the Housing button).
	interactions.InsertBefore(new AwesomeifyButton(), NPCInteractionDatabase.HappinessButton); // These are custom buttons.
	interactions.InsertBefore(new UpgradeButton(), NPCInteractionDatabase.HappinessButton);
	interactions.InsertBefore(new OpenShopOnlyAvailableDuringDay(ShopName, "Day Only Shop"), NPCInteractionDatabase.HappinessButton);

	// Showcase of other things you can do
	NPCInteractionList.Entry awesomeifyButton = interactions.InsertBefore(new AwesomeifyButton(), NPCInteractionDatabase.HappinessButton); // Return the interaction instance
	interactions.InsertAfter(new OpenShopOnlyAvailableDuringDay(ShopName, "Day Only Shop"), awesomeifyButton); // Insert after the instance we saved above.
	interactions.Prepend(NPCInteractions.Shop(ShopName)); // Insert at the beginning
	interactions.Append(NPCInteractions.Shop(ShopName)); // Insert at the end (after the happiness and housing buttons, too)
	
	// Don't want a close, happiness, or housing button? Just disable it!
	interactions.Disable(NPCInteractionDatabase.HousingButton);

	// Adding existing shops from other NPCs
	interactions.Append(NPCInteractions.Shop("Terraria/Painter/Shop", "Painter Shop"));
	interactions.Append(NPCInteractions.Shop("Terraria/Painter/Decor", "Painter Decor"));
	interactions.Append(NPCInteractions.Shop("ExampleMod/ExamplePerson/Shop", "Example Person"));
}

// A custom interaction that inherits OpenShop with the condition changed to only show during the day.
public class OpenShopOnlyAvailableDuringDay(string shopName, string customTextKey = null) : NPCInteractions.Actions.OpenShop(shopName, customTextKey)
{
	public override bool Condition() => base.Condition() && Main.dayTime;

	public override bool ShowExcalmation => true; // Show an exclamation point next to the button.

	public override void TextColor(ref Color chatColor, ref Color chatColorShadow, bool hoveringOverButton) // Edit the color of the button.
{
	chatColor = Color.Black * (Main.mouseTextColor / 255f); // * (Main.mouseTextColor / 255f) makes it pulse like the normal buttons.
	chatColorShadow = Color.LightGray;
	if (hoveringOverButton)
	{
		chatColor = Color.DarkGray * (Main.mouseTextColor / 255f);
		chatColorShadow = Color.White;
	}
}
```

Adding new buttons in GlobalNPC.
```cs
// In GlobalNPC
public override void RegisterChatButtons(NPC npc, NPCInteractionList interactions) {
	// Here we can add additional chat buttons to Town NPCs.
	if (npc.type == NPCID.Guide) {
		// Add a shop button that open the Zoologist's shop.
		// Vanilla shops can specified with "Terraria/NPCName/Shop" ("Decor" for the Painter's second shop)
		// Modded shops can be specified with "ModName/NPCName/ShopName"
		interactions.InsertBefore(NPCInteractions.Shop("Terraria/BestiaryGirl/Shop", "Shop"), NPCInteractionDatabase.CloseButton);

		// Here we are going to disable the Guide's tips button.
		// This way matches the type of the interaction and returns the first that matches or null if not found.
		// If the interaction wasn't found, nothing happens.
		NPCInteraction guideTipNPCInteraction = interactions.Interactions.OfType<NPCInteractions.Actions.GuideTip>().FirstOrDefault();
		interactions.Disable(guideTipNPCInteraction); // If the instance is null (aka not found), Disable won't do anything.

		// Alternate way: this way does the same thing, but searches the Entries instead and returns the NPCInteractionList.Entry if found.
		// NPCInteractionList.Entry guideTipEntry = interactions.Entries.Where(e => e.NPCInteraction.GetType() == typeof(NPCInteractions.Actions.GuideTip)).FirstOrDefault();
		// interactions.Disable(guideTipEntry); // If the instance is null (aka not found), Disable won't do anything.
	}
}

public override void OnChatButtonClicked(NPC npc, NPCInteraction interaction) {
	// With OnChatButtonClicked, we can do additional things when any chat button is clicked. The interaction is the type of button that was clicked.
	// We can use pattern matching to match the interaction type.
	if (npc.type == NPCID.Guide && interaction is NPCInteractions.Actions.GuideReverseCrafting) {
		// OnChatButtonClicked only runs for the local player who clicked the button. Any multiplayer functionality will need to be synced with a packet.
Main.NewText($"<{npc.FullName}> Simply place a material in the slot and I'll tell you everything you can craft with it!");
	}
}

public override bool PreChatButtonClicked(NPC npc, NPCInteraction interaction) {
	// Here we can stop buttons from running their interaction.
	// We can use pattern matching to match the interaction type.
	if (npc.type == NPCID.Guide && interaction is NPCInteractions.Actions.CloseChat) {
		Main.npcChatText = "You can't close my chat window!";
		return false;
	}

	return base.PreChatButtonClicked(npc, interaction);
}
```

### ExampleMod updates
Updated Example Person, Example Traveling Merchant, Example Bone Merchant, Example Town Pet.
* Example Person has the shop, close button, and three additional custom buttons.
  * AwesomeifyButton
    * A simple button.
    * Side note: this button didn't do anything in 1.4 EM. In 1.3 it opened a custom UI that looked like the Goblin Tinkerer's reforge UI.
  * UpgradeButton
    * A button that upgrades the Hive Pack into the Wasp Nest item.
  * OpenShopOnlyAvailableDuringDay
    * Showcases inheriting OpenShop and changing its condition.
* Example Town Pet no longer requires any chat buttons to be set. The Pet button is automatically set now because NPCID.Sets.IsTownPet is set to true.
* Created GlobalNPCInteractions.cs to showcase adding and disabling a button, OnChatButtonClicked, and PreChatButtonClicked.

### Porting Notes:
* `ModNPC.SetChatButtons(ref string button, ref string button2)` replaced with -> `RegisterChatButtons(NPCInteractionList interactions)`
* `ModNPC.OnChatButtonClicked(bool firstButton, ref string shop)` -> `ModNPC.OnChatButtonClicked(NPCInteraction interaction)`
  * Previously OnChatButtonClicked was used to assign a shop to a button, but that is handled in RegisterChatButtons now. Most modders could probably remove this hook.
* `GlobalNPC.OnChatButtonClicked(NPC npc, bool firstButton)` -> `GlobalNPC.OnChatButtonClicked(NPC npc, NPCInteraction interaction)`
* `GlobalNPC.PreChatButtonClicked(NPC npc, bool firstButton)` -> `GlobalNPC.PreChatButtonClicked(NPC npc, NPCInteraction interaction)`
* If your Town Pet has NPCID.Sets.IsTownPet set to true, you no longer need to define a Pet button. It is automatic now.


Rijam:
I completely redid the way buttons are registers inspired by NPCLoot and ItemDropDatabase. This solves the issue with being unable to edit the order of the buttons. It now implicitly registering the buttons to the NPC which means you don't have to worry about accidentally adding your custom button to all NPCs if you forgot to write the condition correctly.

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
+using System;
+using System.Collections.Generic;
+using Terraria.GameContent;
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Allows for <see cref="NPCInteraction"/> chat buttons to be assigned to an NPC.
+/// </summary>
+public readonly struct NPCInteractionList(int npcNetId)
```
Chicken-Bones: `NPCInteractionList` should just be a `class` rather than a `readonly struct`. Having to add `readonly` to all members just makes it a bit noisy, and it could briefly confuse modders into thinking it's a value/immutable type.

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+		/// <br/>Use <c>interactions.Disable(interaction)</c> to disable.
+		/// <br/>Defaults to <see langword="true"/>.
+		/// </summary>
+		public bool Enabled { get; internal set; } = enabled;
+
+		// Added for debugging convenience. NPCInteraction gets printed as the full namespace + name.
+		public override string ToString() => $"(NPCInteraction: {NPCInteraction}, Enabled: {Enabled})";
+	}
+
+	public readonly int NPCNetID => npcNetId;
```
Chicken-Bones: Lets just call it `Type`, negative NPC types are barely supported in tML and they're a legacy hack solely for `SetDefaults` in vanilla to alter some parameters (like size and color) of some NPC types. There are no interactable NPCs with negative types.

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+
+	private readonly List<Entry> _interactionEntries = new List<Entry>();
```
Chicken-Bones: Rename to `_entries`

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+
+	/// <summary>
+	/// Returns the full <c>List&lt;NPCInteraction&gt;</c> of the NPC.
+	/// </summary>
+	/// <returns>An empty list if not found.</returns>
+	public readonly List<NPCInteraction> GetInteractions()
```
Chicken-Bones: Replace with
```cs
IEnumerable<NPCInteraction> Interactions => _entries.Select(e => e.NPCInteraction);
```

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+	{
+		List<NPCInteraction> list = new();
+		foreach (Entry entry in _interactionEntries) {
+			list.Add(entry.NPCInteraction);
+		}
+		return list;
+	}
+
+	public readonly List<Entry> GetInteractionEntries()
```
Chicken-Bones: Change to property
```cs
public IReadOnlyList<Entry> Entries => _entries;
```

`IReadOnlyList` here prevents modders from trying to insert entries into the list manually rather than using the `Insert` methods on `NPCInteractionList`

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+	public readonly Entry InsertBefore(NPCInteraction interactionToRegister, NPCInteraction interactionBefore)
+	{
+		Entry entry = new(interactionToRegister);
+		int index = -1;
+		foreach (Entry searchEntry in _interactionEntries) {
+			if (searchEntry.NPCInteraction.Equals(interactionBefore)) { // Instances are the same.
+				index = _interactionEntries.IndexOf(searchEntry);
+				break;
+			}
+		}
```
Chicken-Bones: Use https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.findindex?view=net-10.0

```cs
var index = _entries.FindIndex(e => e.NPCInteraction == interactionBefore);
```

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,297 @@
...
+	/// <summary>
+	/// Enables an interaction for the current NPC.
+	/// <para>The <paramref name="interaction"/> needs to be the exact instance of the button.
+	/// <br/>The Close, Happiness, Housing, and Pet buttons are already predefined and can be used with <see cref="NPCInteractionDatabase.CloseButton"/>, etc.
+	/// <br/>If you don't already have the instance, use <see cref="FindInteractionByType(NPCInteraction)"/> or <see cref="FindInteractionByType(Type)"/> to get it.
+	/// </para>
+	/// </summary>
+	/// <param name="interaction">The NPCInteractionList.Entry to enable.</param>
+	/// <remarks>Interactions are enabled by default, so this is only needed if the interaction was disabled.</remarks>
+	public readonly void Enable(Entry interaction)
```
Chicken-Bones: I generally prefer not to provide this method, `NPCShop.Entry` doesn't have it for example.

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
```diff
@@ -0,0 +1,349 @@
...
+	/// Returns the NPCInteractionList of the NPC.
+	/// </summary>
+	/// <param name="npcNetId">The NPC to get the buttons for.</param>
+	/// <returns>Registers the NPC with an empty list if not found.</returns>
+	public NPCInteractionList GetInteractionListForNPCID(int npcNetId)
+	{
+		if (_interactionDatabase.TryGetValue(npcNetId, out NPCInteractionList value))
+			return value;
+
+		return RegisterNewNPC(npcNetId);
```
Chicken-Bones: Should we be registering here? I thought modders npc interactions were meant to be automatically registered in `Populate`, and modders were meant to use hooks on the `NPCInteractionList`, not go and get their own handle to the interaction list and try and modify it at runtime.

In addition, I think it makes sense to just have the one method, that returns `IEnumerable<NPCInteraction>` and shows only the enabled interactions. Since that's all that matters during gameplay.

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
```diff
@@ -0,0 +1,349 @@
...
+
+		return RegisterNewNPC(npcNetId).GetInteractionEntries();
+	}
+
+	/// <summary>
+	/// Adds the NPC to the database and registers the provided interactions.
+	/// </summary>
+	/// <param name="npcNetId">The NPC to register.</param>
+	/// <param name="interactions">List of interactions to pre-register</param>
+	private NPCInteractionList RegisterNewNPC(int npcNetId, params NPCInteraction[] interactions)
```
Chicken-Bones: Do we use the return value?

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
```diff
@@ -0,0 +1,349 @@
...
+	{
+		if (!_interactionDatabase.ContainsKey(npcNetId))
+			_interactionDatabase[npcNetId] = new NPCInteractionList(npcNetId);
+
+		foreach (NPCInteraction interaction in interactions) {
+			_interactionDatabase[npcNetId].Append(interaction);
+		}
+		return _interactionDatabase[npcNetId];
```
Chicken-Bones: Should we be allowing duplicate calls to this method? It's private after-all. It can just throw if the npc has already been registered

patches/tModLoader/Terraria/GameContent/UI/NPCChatPanel.cs.patch
```diff
@@ -0,0 +1,75 @@
...
+ 			UILinkPointNavigator.SetPosition(2500 + num, rectangle.Center.ToVector2());
+@@ -125,7 +_,11 @@
+ 				num2 = num;
+ 				if (Main.mouseLeft && Main.mouseLeftRelease) {
+ 					Main.mouseLeftRelease = false;
++					if (!ModLoader.NPCLoader.PreChatButtonClicked(interaction))
++						return;
++
+ 					interaction.Interact();
++					ModLoader.NPCLoader.OnChatButtonClicked(interaction);
```
Chicken-Bones, line 51: @JavidPack should we deprecate these hooks in favour of the interaction system?

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
```diff
@@ -8,7 +8,82 @@
...
-@@ -191,12 +_,25 @@
+@@ -33,20 +_,41 @@
+ 			}
+ 		}
+ 
++		/// <summary> The helper method <see cref="NPCInteractions.Shop(string, string)"/> can be used instead of creating a new instance of this class.</summary>
+ 		public class OpenShop : NPCInteraction
+ 		{
+-			private int _shopIndex;
+-			private int _npcType;
```
Chicken-Bones: Is `_shopIndex` even used anymore? Recommend
```cs
			/*
			private int _shopIndex;
			private int _npcType;
			*/
```

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
```diff
@@ -48,3 +145,56 @@
...
+ 		Shop(17, 1);
+ 		Shop(19, 2);
+ 		Shop(20, 3);
+@@ -403,7 +_,26 @@
+ 
+ 	private static void Shop(int npcType, int shopIndex, string customTextKey = null)
+ 	{
+-		Register(new Actions.OpenShop(npcType, shopIndex, customTextKey));
++		// Register(new Actions.OpenShop(npcType, shopIndex, customTextKey)); TML: Removed/Obsolete
++	}
```
Chicken-Bones, line 186: I wonder if we should just make `NPCInteractionDatabase` an `NPCInteractions` partial class with the tML implementation. Then just block comment out all the vanilla registration code we don't use

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
Rijam: The reason was structs can't be null. But, if I change NPCInteractionList to a class, then returning null is no problem.

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
Rijam: Only in one spot below in Populate, but I can easily replace those with just `_interactionDatabase[pair.Key]`

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
Rijam: I see what you mean. You could call this multiple times to continue registering interactions.

patches/tModLoader/Terraria/GameContent/UI/NPCChatPanel.cs.patch
JavidPack: Well, I think it's certainly possible a modder might want to act on an existing interaction, rather than be forced to replace it outright. 

For example, "if you collect taxes, then spawn some additional mod-specific currency as well", or "After pet animal, if some world flag, spawn a boss"

ExampleMod/Common/GlobalNPCs/GobalNPCInteractions.cs
```diff
@@ -17,16 +18,14 @@ public override void RegisterChatButtons(NPC npc, NPCInteractionList interaction
...
-			// NPCInteractionList.Entry guideTip = interactions.FindInteractionByType(typeof(NPCInteractions.Actions.GuideTip));
-			// interactions.Disable(guideTip);
+			// This way matches the type of the interaction and returns the first that matches or null if not found.
+			// If the interaction wasn't found, nothing happens.
+			NPCInteraction guideTipNPCInteraction = interactions.Interactions.OfType<NPCInteractions.Actions.GuideTip>().FirstOrDefault();
+			interactions.Disable(guideTipNPCInteraction); // If the instance is null (aka not found), Disable won't do anything.
+
+			// Alternate way: this way does the same thing, but searches the Entries instead and returns the NPCInteractionList.Entry if found.
+			// NPCInteractionList.Entry guideTipEntry = interactions.Entries.Where(e => e.NPCInteraction.GetType() == typeof(NPCInteractions.Actions.GuideTip)).FirstOrDefault();
+			// interactions.Disable(guideTipEntry); // If the instance is null (aka not found), Disable won't do anything.
```
Chicken-Bones: Lets remove this comment

ExampleMod/Content/NPCs/TownPets/ExampleTownPet.cs
```diff
@@ -163,9 +163,11 @@ public override string GetChat() {
 			return chat;
 		}
 
-		public override void SetChatButtons(ref string button, ref string button2) {
-			button = Language.GetTextValue("UI.PetTheAnimal"); // Automatically translated to say "Pet"
+		/* Setting NPCID.Sets.IsTownPet will automatically add the pet button, but here is one way on how to do it if it weren't automatic.
+		public override void RegisterChatButtons(NPCInteractionList interactions) {
+			interactions.InsertAfter(NPCInteractionDatabase.PetButton, NPCInteractionDatabase.CloseButton);
 		}
+		*/
```
Chicken-Bones: No need for this, we have other `RegisterChatButtons` examples

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
```diff
@@ -8,7 +8,82 @@
...
  namespace Terraria.GameContent;
-@@ -191,12 +_,25 @@
+@@ -33,20 +_,42 @@
+ 			}
+ 		}
+ 
++		/// <summary> The helper method <see cref="NPCInteractions.Shop(string, string)"/> can be used instead of creating a new instance of this class.</summary>
+ 		public class OpenShop : NPCInteraction
+ 		{
+ 			private int _shopIndex;
```
Chicken-Bones: Isn't this now unused? Comment it out as well? Use block comment for members instead of `//` (and also in the constructor below)

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
```diff
@@ -8,7 +8,82 @@
...
+-			public override bool Condition() => base.TalkNPCType == _npcType;
++			public override bool Condition() => true; // TML: Changed condition
+ 
+ 			public override string GetText()
+ 			{
+@@ -58,7 +_,18 @@
+ 
+ 			public override void Interact()
+ 			{
++				// If vanilla shop, open by ID.
```
Chicken-Bones: probably leave the commented out vanilla code/function so it's easier for future readers

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
```diff
@@ -48,3 +145,56 @@
...
++		// TML: Buttons are no longer initialized here. They are set in Terraria.ModLoader.NPCInteractionDatabase.Populate()
+ 		Shop(17, 1);
+ 		Shop(19, 2);
+ 		Shop(20, 3);
+@@ -403,7 +_,26 @@
+ 
+ 	private static void Shop(int npcType, int shopIndex, string customTextKey = null)
+ 	{
+-		Register(new Actions.OpenShop(npcType, shopIndex, customTextKey));
++		// Register(new Actions.OpenShop(npcType, shopIndex, customTextKey)); TML: Removed/Obsolete
```
Chicken-Bones: Better to block comment out the `Shop` calls in `Initialize` along with this method.

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
```diff
@@ -0,0 +1,352 @@
...
+
+	// int Key: NPC Type
+	private readonly Dictionary<int, NPCInteractionList> _interactionDatabase = new Dictionary<int, NPCInteractionList>();
+
+	/// <summary>
+	/// Returns the NPCInteractionList of the NPC.
+	/// </summary>
+	/// <param name="npcNetId">The NPC to get the buttons for.</param>
+	/// <returns><see langword="null"/> if not found.</returns>
+	public NPCInteractionList GetInteractionListForNPCID(int npcNetId)
```
Chicken-Bones: Just `GetInteractionList` will do. Recommend replacing `npcNetId` everywhere with just `type` or `npcType`

Net ids are a legacy feature only relevant for the spawning of certain enemy variants.

patches/tModLoader/Terraria/ModLoader/NPCInteractionDatabase.cs
```diff
@@ -0,0 +1,352 @@
...
+
+		return null;
+	}
+
+	/// <summary>
+	/// Returns the full <c>List&lt;NPCInteractionList.Entry&gt;</c> of the NPC.
+	/// </summary>
+	/// <param name="npcNetId">The NPC to get the buttons for.</param>
+	/// <returns><see langword="null"/> if not found.</returns>
+	public IReadOnlyList<NPCInteractionList.Entry> GetInteractionEntriesNPCID(int npcNetId)
```
Chicken-Bones: Suggest, `GetInteractionEntries`

patches/tModLoader/Terraria/ModLoader/NPCInteractionList.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Terraria.GameContent;
+
+namespace Terraria.ModLoader;
+
+/// <summary>
+/// Allows for <see cref="NPCInteraction"/> chat buttons to be assigned to an NPC.
+/// </summary>
+public class NPCInteractionList(int npcNetId)
```
Chicken-Bones: As above, rename `npcNetId` to `type`

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -4694,14 +4694,20 @@
...
+ 
+-	public void OpenShop(int shopIndex)
++	public void OpenShop(string shopName) // TML: Changed to use strings instead of integer indexes.
+ 	{
  		playerInventory = true;
  		stackSplit = 9999;
  		npcChatText = "";
 -		SetNPCShopIndex(shopIndex);
-+		SetNPCShopIndex(1);
 -		shop[npcShop].SetupShop(npcShop);
```
Chicken-Bones: As above, comment out vanilla code instead of removing.

Chicken-Bones (approved):
API surface looks good. `ExampleMod` examples look good. Some minor patch cleanup and names to adjust. @JavidPack to review documentation (I didn't have time to read it in detail)

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
Rijam: The shop index is still being used in `NPCInteractionDatabase.RegisterVanilla()`, but those could easily be changed to the full shop name if you want me to. 
`new NPCInteractions.Actions.OpenShop(1), // Equivalent to NPCInteractions.Shop("Terraria/Merchant/Shop")`
https://github.com/tModLoader/tModLoader/pull/5096/changes#diff-96623fe71f5309151228534b3faabdf3185128994dfb5e26f3f603237e315e2bR111

patches/tModLoader/Terraria/Main.cs.patch
Rijam: The new diff looks odd, but the patch is much cleaner. I left the original method as is, just commented out. https://github.com/Rijam/tModLoader/blob/1.4.5-TownNPCChatButtons/patches/tModLoader/Terraria/Main.cs.patch#L4701

patches/tModLoader/Terraria/GameContent/NPCInteractions.cs.patch
Rijam: I could also use `NPCInteractions.Shop(NPCShopDatabase.GetShopNameFromVanillaIndex(1))` instead of `NPCInteractions.Shop("Terraria/Merchant/Shop")` if you want to continue to use the vanilla shop IDs.

Chicken-Bones (approved):

PR #5097, opened 2026-04-15
Fixed the encoding of "Host & Play" 's Console

LogSpiral:
### What is the bug?
When I uses Multiplayer 's Host & Play
I found that the Chinese Character in the new console is complete mess.
Reason was the Encoding mismatched.

When call AllocConsole() in Windows with Simplified Chinese Language, It uses the CodePage of **GBK** Encoding
But the default Encoding of StreamWriter seems to be **UTF8**

Here are the screenshots

### How did you fix the bug?
I add two NativeMethod in Relogic.OS.Windows.NativeMethods.TML.cs
SetConsoleCP
SetConsoleOutputCP
Which is used to set the Encoding of Input and the OutPut Code Page of the new console

Then, I Call the two function after calling AllocConsole() in Relogic.OS.Windows.WindowService
the number 65001 means the UTF-8 Encoding
these worked when I uses Host & Play again, the Encoding Matched

The other way to fix this is to use another constructor of StreamWriter
Setting the Encoding of the StreamWriter manually can fix this as well
Like this:
```csharp
// before
TextWriter writer = new StreamWriter(Console.OpenStandardOutput()) { AutoFlush = true };

// after
Encoding encoding = Encoding.GetEncoding((int)NativeMethods.GetConsoleOutputCP());
TextWriter writer = new StreamWriter(Console.OpenStandardOutput(), encoding) { AutoFlush = true };
```
This requires add **GetConsoleOutputCP()** method in **NativeMethods**

### DIFF
--- a/patches/tModLoader/ReLogic/OS/Windows/NativeMethods.TML.cs
+++ b/patches/tModLoader/ReLogic/OS/Windows/NativeMethods.TML.cs
@@ -22,6 +22,12 @@ static partial class NativeMethods
 	[DllImport("kernel32.dll")]
 	public static extern IntPtr GetStdHandle(uint nStdHandle);
 
+	[DllImport("kernel32.dll")]
+	public static extern uint SetConsoleCP(uint wCodePageID);
+
+	[DllImport("kernel32.dll")]
+	public static extern uint SetConsoleOutputCP(uint wCodePageID);
+
 	public static void HideConsole()
 	{
 		IntPtr hWnd = GetConsoleWindow();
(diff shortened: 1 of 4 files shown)


LogSpiral:
Here are the screenshots:
serverscreenshot2
serverscrenshot1
The first shot showing the mismatch encoding
and the second shot showing what it should be when encoding matches

JavidPack:
I've reimplemented the fix using a different approach using some existing code. It seemed to work on my end, but I can't be sure since there are a lot of variables at play here and my computer isn't set to Chinese directly. 

If you could test this, that would be great. Either pull the changes and patch/build or download this build here https://github.com/tModLoader/tModLoader/actions/runs/24752089791/artifacts/6567554343.

I would also like other Chinese users to verify the issue and the fix, just so we are sure this isn't an issue that was only affecting your computer configuration. Since others haven't mentioned this before to me, I just want to check with any other users with computers set to Chinese.

LogSpiral:
Thanks! I've test the build and it worked well on my computer. For results from other users, maybe need to wait for a day to collect enough samples. Up to now, all the users testing the build reports that encoding is right.

PR #5101, opened 2026-04-18
Example Mod: Update ExampleFoodItem.cs

A5TR0spud:
Instead of setting the frame time of the ExampleFoodItem to `int.MaxValue`, set it to -1 and set the `NotActuallyAnimating` field of the animation object to true.

No


JavidPack:
Interestingly, vanilla code does the old approach, but I can see this being more readable.

PR #5104, opened 2026-04-21
[1.4.5] Fix slime body ore drop stack size

noahDP1911:
### What is the bug?
Some vanilla slime body drops that should use the `3..13` stack range were falling back to the default stack size of `1`. The existing tModLoader patch moved normal ore stack sizes to `ItemID.Sets.OreDropsFromSlime`, but it commented out the entire vanilla `3..13` case group. That group also contained other vanilla items, such as hardmode ores, Hellstone, and Desert Fossil, which are not currently registered in `OreDropsFromSlime`.

### How did you fix the bug?
The patch now removes only the normal ore item IDs that are handled by `ItemID.Sets.OreDropsFromSlime`, while preserving the remaining vanilla `3..13` slime body drop cases.
This keeps the existing `OreDropsFromSlime` behavior intact and restores the correct vanilla stack range for the affected special slime body drops. The default `1..1` behavior remains unchanged for items without a special stack range.
Tested in-game with a custom mod that spawned the affected special slimes and verified the drop stack sizes.

The affected vanilla items could be added to `ItemID.Sets.OreDropsFromSlime`, but that set is also used to choose which ore items can appear inside slimes. Preserving the existing vanilla switch cases is more targeted and avoids changing the slime body item selection pool.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/SlimeBodyItemDropRule.cs.patch
@@ -18,16 +18,29 @@
  		ItemDropAttemptResult result = default(ItemDropAttemptResult);
  		result.State = ItemDropAttemptResultState.Success;
  		return result;
-@@ -50,4 +_,0 @@
--			case 11:
--			case 12:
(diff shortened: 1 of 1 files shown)


JavidPack:
Thanks.

PR #5109, opened 2026-04-25
Debuff Length Migration

noahDP1911:
I hope this time the patches have the correct format :D
Otherwise i need some further input on how to do it correctly

### What is the bug?
The 1.4.5 porting notes call out that `BuffID.Sets.LongerExpertDebuff` was renamed to `BuffID.Sets.BuffTimeIsExtendedWithGameDifficulty`, and that the docs for this set and `BuffTimeIsExtendedByDeadCellsPotionStationBuff` needed cleanup.
Some migration surfaces still referenced or preserved the old name, and the docs did not fully describe the current 1.4.5 behavior.

### How did you fix the bug?
Removed the stale `BuffID.Sets.LongerExpertDebuff` alias from the TML `BuffID` partial and updated ExampleMod to use `BuffID.Sets.BuffTimeIsExtendedWithGameDifficulty`.
Updated the `BuffTimeIsExtendedWithGameDifficulty` docs to describe the current behavior and reference `GameDifficultyData.DebuffTimeMultiplier`. Also added documentation for `BuffTimeIsExtendedByDeadCellsPotionStationBuff`. `tModPorter` already had the static field rename, but the separate removed-member note for `ModBuff.longerExpertDebuff` still pointed modders to the old name. This updates that note and its expected test output to point to `BuffID.Sets.BuffTimeIsExtendedWithGameDifficulty`.
Also updated the tracked `Player.cs.rej` context to use the new set name.

The old `LongerExpertDebuff` alias could be kept as a compatibility shim, but this is 1.4.5 migration cleanup and the migration guide already points modders to `BuffTimeIsExtendedWithGameDifficulty`.
Removing the stale alias and updating the examples, docs, and migration hints keeps the API surface consistent with the porting notes.


JavidPack:
This is great, you did pretty much everything correctly.

You'll notice from my commits I made a few additional changes. The `.cs.rej` files represent code patches that weren't merged into 1.4.5 from 1.4.4. Rather than updating the text there, the whole patch was removed since the patch is no longer necessary. I also updated PortingNotes_1.4.5.md so we can track what still needs to be done.

PR #5117, opened 2026-04-29
[1.4.5] Complete NPCSpawnInfo migration to NPC.Spawner

noahDP1911:
I realized this was a bit more in-depth than my previous PRs, if anything needs some more tinkering or is just incorrect, just call it out and i'll look at it again over the next few days.

### What is the bug?
`NPCSpawnInfo` was no longer the intended spawn context type for 1.4.5, but the migration was incomplete.
The old `NPCSpawnInfo` placeholder still existed, spawn-related porting notes still called out missing docs/hooks, and modders did not have equivalent hook points for adjusting `NPC.Spawner` state at the same stages of the natural spawn pipeline. tModPorter also needed to fully migrate old `NPCSpawnInfo` usages to `NPC.Spawner`, including removed fields.

### How did you fix the bug?
Removed the stale `NPCSpawnInfo` file and completed the migration to `NPC.Spawner`.
Added two `GlobalNPC` hooks for the intended spawn pipeline stages:
`EditSpawnFlags` enables mods to change player-level spawn context before spawn rate, spawn range, and spawn tile selection happen. For example, a modded safe-zone effect can set `spawner.noWorms = true` before vanilla spawn logic uses that flag. `SpawnChance` and `EditSpawnPool` run too late to affect those earlier decisions.
`EditSpawnInfo` enables mods to change tile-level spawn context after the spawn tile is known but before `SpawnChance` and `EditSpawnPool` run. For example, a modded tile can set `spawner.nearGranite = true` so later spawn chance and spawn pool logic treats the chosen tile as granite-like.
Updated `NPCLoader` and `NPC.cs.patch` to call these hooks at those points.
Updated docs and migration notes to describe the replacement fields and the removed `PlayerFloorX` / `PlayerFloorY` fields.
Updated tModPorter to:
  - Rename `Terraria.ModLoader.NPCSpawnInfo` to `Terraria.NPC.Spawner`.
  - Migrate old lower-case `NPCSpawnInfo` field names to the current `NPC.Spawner` field names.
  - Mark `PlanteraDefeated`, `PlayerFloorX`, and `PlayerFloorY` with migration notes.
  - Support nested type output such as `NPC.Spawner` when rewriting renamed types.

A compatibility `NPCSpawnInfo` wrapper could be kept, but that would duplicate spawn state and risk drifting from vanilla `NPC.Spawner`.
Another option would be to only update `SpawnChance` and `EditSpawnPool`, but that would leave no supported way for mods to adjust spawn state before spawn rate/range/tile selection or after tile-derived spawn info is computed.
Using `NPC.Spawner` directly keeps the migration local to the new vanilla spawn model and makes tModPorter guide modders toward the current API.


patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -105,6 +108,9 @@
  			if (!PostCheckChosenSpawnTile(spawnTileX, spawnTileY, spawnTileType, spawnWallType))
  				return false;
  
+@@ -172,0 +_,2 @@
```
JavidPack: How come these patches are weird, are you manually writing these patches? The patcher doesn't generate short patches like this.

Your other PR was like this too. I wouldn't want you to waste effort manually constructing these patches, that can be tedious. To generate the patches you just need to run `Diff tModLoader` in the setup tool.

noahDP1911: Yeah those were manual since it was the last thing that i did and they were small as you said. But since you point it out i probably should have just used the patcher again, my bad

JavidPack: If you (or I) don't use it before this is merged, the fixed patch will be generated the next time someone works on the 1.4.5 branch, showing changed files, and that might be confusing to them since they know they didn't touch this code.

Anyway, I'll try to get to reviewing soon. The only other thought I had so far is the new hooks probably need more explanation. For example, what is something they facilitate that wasn't possible before. Both in the PR description and the docs.

noahDP1911: Tried to clean it up a bit as you recommended, tried to be a bit more verbose with the explanations, but i dont know if my examples are the best

Chicken-Bones:
@Kogsey would be great if you could take a look at this/the 1.4.5 changes and how they affect your work in #3699

Kogsey:
From a quick look it probably won't effect a massive amount. Vanilla refactoring the spawn code might even help. I'll take a proper look after this weekend

PR #5122, opened 2026-05-01
[1.4.5] Resolve WorldGenerator._seed PortingNotes TODO

a-grasso:
### What is the bug?
After the 1.4.5 patch port, `WorldGenerator._seed` is `private readonly int _seed;`. The previous tModLoader patch elevating this field to `internal` was lost during the port. Tracked in `PortingNotes_1.4.5.md` (line 17): "WorldGenerator._seed needs to be internal again. The patch was lost".

### How did you fix the bug?
~~Re-applied the visibility patch: changed `private readonly int _seed;` to `internal readonly int _seed;` in `src/tModLoader/Terraria/WorldBuilding/WorldGenerator.cs` and regenerated via `setup-cli diff tml`. Result is a single 7-line context hunk in `patches/tModLoader/Terraria/WorldBuilding/WorldGenerator.cs.patch`. Also removed the corresponding line from `PortingNotes_1.4.5.md`.~~

I took another look, seems like the patch isn't actually doing anything useful since the WorldGenerator instance is private anyway. TODO deleted.

~~### Are there alternatives to your fix?
None reasonable. The field needs to be readable by tModLoader code outside the `WorldGenerator` class; `internal` is the minimal scope and matches the previous (lost) patch~~

### DIFF
--- a/PortingNotes_1.4.5.md
+++ b/PortingNotes_1.4.5.md
@@ -14,7 +14,6 @@ Once all patches are fixed, these items need to be fixed or double checked:
 - We need to add ModItem.SummonPrefix(), add ModPrefix.Summon
 - ShimmerTransforms.IsItemTransformLocked seems to have been split, need to verify RecipeLoader.DecraftAvailable and other logic still applies.
 - Update ModPylon docs to removed danger check from check listing, and remove the ValidTeleportCheck_AnyDanger hook
-- WorldGenerator._seed needs to be internal again. The patch was lost
 - Consider updating FlexibleTileWand.Reload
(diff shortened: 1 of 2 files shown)


JavidPack:
Thanks for this. I took another look and it seems like this patch isn't useful anymore. Maybe it was useful long ago, but it doesn't seem to provide anything useful for modders since the WorldGenerator instance in WorldGen is private as well.

If any modder needs the world seed, `Main.ActiveWorldFileData.Seed` is how you would access that.

a-grasso:
@JavidPack  Thanks for looking at it :D🙂 i tried to pick an easy looking task as entry.
Could you recommend a next contribution that might not be useless going forward? BR

Chicken-Bones:
For recommendations, it depends on your experience as a modder, ideally pick things where you understand how they're meant to work or be used. Please come to the discord (https://discord.com/channels/103110554649894912/1466306663374913687) and you can mention any tasks you're thinking about undertaking and plenty of people will be happy to discuss or help estimate the difficulty

Issue #4772, opened 2025-09-03
When an item having too many tooltipline can't show the tooltipline


zlzhaidou:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

When an item having too many tooltipline will show "index was outside the bounds of the array 

<img width="726" height="30" alt="Image" src="https://github.com/user-attachments/assets/3269a06e-24ab-4a5c-8ec3-71a3348a099b" />

"

### Additional Information

Make it won't report and even player can not view all tooltipline

PR #5128, opened 2026-05-05
[1.4.5] Refix Too Many Tooltip Lines Can't Show the Tooltip Line

Rijam:
### What is the bug?

If an item were to draw more than 30 tooltip lines (vanilla default maximum), an index out of bounds of the array error would be thrown.

This bug was fixed in [Commit 6e7c00b](https://github.com/tModLoader/tModLoader/commit/6e7c00b3da4197c48caab6cfaf6b8134ce508827) based on issue #4772, however, 1.4.5 changed how tooltips were drawn and this patch was not applied.

The error would occur in `Main.MouseText_DrawItemTooltip` on line 18407 (accurate at the time of writing this).
```cs
for (int k = 0; k < mouseTextTooltipLine_Text.Length; k++) {
	mouseTextTooltipLine_Color[k] = new Microsoft.Xna.Framework.Color((byte)((float)(int)mouseTextTooltipLine_Color[k].R * num2), (byte)((float)(int)mouseTextTooltipLine_Color[k].G * num2), (byte)((float)(int)mouseTextTooltipLine_Color[k].B * num2), mouseTextColor);
}
```
Earlier in the method, `mouseTextTooltipLine_Text` was resized in `ItemLoader.ModifyTooltips`, but `mouseTextTooltipLine_Color` (new to 1.4.5) was not. Thus, if there were more than 30 tooltip lines, it was trying to assign indexes outside the bounds of `mouseTextTooltipLine_Color`.

### How did you fix the bug?

In `ItemLoader.ModifyTooltips`, resize the `lineColors` array (which is a ref of `mouseTextTooltipLine_Color`), then assign the default value of white to the elements above 30.

* I would've used `Main._mouseTextTooltip_MaxLines` for the start of the for loop instead of 30, but it is private.
* The `mouseTextTooltipLine_Text` aka `text` array gets resized in a different way. The contents get copied into a `List<TooltipLine>`. The `text` array is reassigned with the new size. Then it is refilled with the original data that was stored in the `List<TooltipLine>`. A similar thing could be done with the `lineColors` array where the contents are copied, it gets reassigned and filled with the default white color, then gets the original data that was stored.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/Creative/ItemFilters.cs.patch
+++ b/patches/tModLoader/Terraria/GameContent/Creative/ItemFilters.cs.patch
@@ -8,7 +8,7 @@
  using Terraria.UI;
  
  namespace Terraria.GameContent.Creative;
-@@ -22,19 +_,33 @@
+@@ -22,19 +_,34 @@
  		private int _unusedResearchLine;
  		private string _search;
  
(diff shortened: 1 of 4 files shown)


JavidPack:
Your original fix only fixes exceeding the limit with dynamically added tooltips (ModifyTooltips). I fixed the issue here for the creative mode search filter. It was also broken in 1.4.4. I also fixed the issue for normal tooltips (Item.Tooltip), which is why the 1.4.4 approach was where it was originally.

PR #5129, opened 2026-05-07
Add TileLoader.SetGlow hook and add its example

hezuo1hao:
### What is the new feature?
Added SetGlow function, which can be used to create glow effects for tiles

In the vanilla, the glow effect can be created, but the related code uses special judgment. After adding this function, similar effects can be written to the vanilla without using on or il

I don't have any better ideas

Added corresponding examples in the ExampleBar

### ExampleMod updates
Added corresponding examples in the ExampleBar

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleBar.cs
+++ b/ExampleMod/Content/Tiles/ExampleBar.cs
@@ -1,5 +1,7 @@
 using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
 using Terraria;
+using Terraria.GameContent;
 using Terraria.ID;
 using Terraria.Localization;
 using Terraria.ModLoader;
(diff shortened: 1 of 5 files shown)


JavidPack:
I think we'd want to auto-assign `glowTexture` if a `ModBlockType.GlowTexture` exists, similar to how `ModTile.HightlightTexture` or `ModProjectile.GlowTexture` work. The hook can be used to customize the draw color and source rectangle, if needed, but we'd want to support it automatically setting the default values even if the hook is not overridden, as long as the texture exists.

This might get complicated since many mods and even ExampleMod already use a `_Glow` texture, so we'll need to confirm that things work if we do make these changes. This idea might need to be a 1.4.5 thing.

hezuo1hao:
> I think we'd want to auto-assign `glowTexture` if a `ModBlockType.GlowTexture` exists, similar to how `ModTile.HightlightTexture` or `ModProjectile.GlowTexture` work. The hook can be used to customize the draw color and source rectangle, if needed, but we'd want to support it automatically setting the default values even if the hook is not overridden, as long as the texture exists.
> 
> This might get complicated since many mods and even ExampleMod already use a `_Glow` texture, so we'll need to confirm that things work if we do make these changes. This idea might need to be a 1.4.5 thing.

I agree to use a solution similar to modproj, which should indeed be carefully considered, but it is important. If necessary, I can write another version in 1.4.5 according to the solution you proposed 
At present, there are no hooks that give glow to tile affected by wind effects. If this change is successfully added, it will be a big step for TML to adapt to wind effects

Issue #4703, opened 2025-07-08
Better Summon Tag Damage Handling


stormytuna:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently, whip tag damage is hard coded and impossible to generically get/set like a regular item stat. I'm not sure what the best solution would be, I could see sets that map projectile type to tag debuff type, then a set mapping tag debuff type to extra tag damage, however it's possible for a mod to add percentage tag damage rather than the flat damage vanilla ones add.

This would need more discussion and I'm open to suggestions regarding how to solve this.

### What does this proposal attempt to solve or improve?

It's currently impossible to even read tag damage of a whip generically due to how tag damage is handled. One example of how this change could be useful is a mod adding whip specific prefixes that modify tag damage.

Chicken-Bones:
There will be a standard way to access and modify the `TagDamage` of all vanilla whip _items_ (and other things interacting with the minion tag system) in 1.4.5

Solxanich:
Closed in #5130

PR #5130, opened 2026-05-08
[1.4.5] Whip Updates

Rijam:
### What was changed?

* Adds docs for `WhipTagEffect` and `UniqueTagEffect`
* Updated Example Mod's whips with the new tag effect system.
* Updated Example Mod's whip projectiles with improved code and comments.
* Fixed 1000 damage being passed into TagEffectState.OnHit which meant the Starcrash's asteroids were doing 1330 damage.
  * tModLoader changed the variable that it used to 1000 and divided by 1000 for setting the modifiers.SourceDamage
  * Somehow the Firecracker didn't seem to be wrong even though it was setting the damage to 1000 * 1.75? I couldn't figure it out.
  * The Vulgar Display of Flower and Possession were unaffected because they set their own damage.
* The new vanilla system solves #4703

### Discussion

For clarity, I'm calling the old tag system "Tag buff" or "Tag debuff" and the new tag system "Tag effect".

* Need to discuss what to do with the previous whip debuff immunity.
  * The Dark Harvest and Electric Eel are the only whips that still use some kind of debuff. If an NPC were to be immune to tags, they should also be immune to these debuffs, too.
  * ~~I say we just keep the existing system. Maybe it is slightly odd that you'd need to use the buffs system to specify tag immunity, but I think keeping is fine for these reasons:~~
    * ~~Buff immunity and damage immunity are the only type of immunity systems that people are familiar with.~~
    * ~~As mentioned, the Dark Harvest and Electric Eel still use a debuff as part of their tag effects. So, immunity between those debuffs and the tag effect should be the same.~~
    * ~~No additional patches would be needed for tag effect immunity because it's already handled by the tag debuff immunity.~~
    * ~~It would require no additional porting for modders. If they already have their NPCs immune to tag debuffs, they are already also immune to the tag effects.~~
  * We decided that `NPCID.Sets.ImmuneToAllBuffs` would be removed and replaced with `NPCID.Sets.ImmuneToWhipTags`. Modders should continue to use `NPCID.Sets.ImmuneToRegularBuffs` and if immunity to tags is desired, additionally set `NPCID.Sets.ImmuneToWhipTags`.

# TODOs and notes in PortingNotes_1.4.5.md
* Commented out line in Projectile.Damage_PVE_Inner: `// modifiers.FlatBonusDamage += num6; // TODO: Handled now by TagEffectState.ModifyHit maybe?` 
  * Yes, handled by `WhipTagEffect.ModifyTaggedHit` now.
  * num6 was the tag damage for each debuff + the SummonTagDamageMultiplier (`num6 = (int)((float)num6 * num13);`), which is in `WhipTagEffect.ModifyTaggedHit` now.
* Projectile.cs.rej patch `+ float num13 = ProjectileID.Sets.SummonTagDamageMultiplier[type];`
  * No longer needed. It is now handled `WhipTagEffect.ModifyTaggedHit`, too.
  * The penalty for projectile type 864 (Blade Staff) was removed in 1.4.5. The other three types are the three lightning aura projectiles which are covered in ProjectileID.Sets.SummonTagDamageMultiplier.
* Projectile.cs.rej patch `+ modifiers.ScalingBonusDamage += 1.75f * num13; //num6 += num17;`
  * No longer needed. It is for the Firecracker's proc, and it is covered in `WhipTagEffect_Firecracker.ModifyProcHit` now.
* One more note about reverting the `IsAnNPCWhipDebuff` rename to `IsATagBuff`.
  * No need to revert it. Keep it renamed imo.

Example Mod's whips have been updated with new examples and additional comments.

* ExampleWhip and ExampleWhipProjectile have been simplified.
  * It no longer has the charging ability that it did before (ExampleWhipAdvanced still has it).
  * The draw code has been changed to be more generic.
    * It is almost an exact copy of the Leather Whip's drawing.
	* It assumes each segment in the sprite are equal size, like most sprite sheets are.
* ExampleWhipAdvanced and ExampleWhipProjectileAdvanced have been updated.
  * Now has its own sprites instead of reusing ExampleWhip's sprites. 
  * The item creates a `WhipTagEffect_ExampleWhipAdvanced` to show a few examples of the new tag system.
  * Added `Projectile.drawLayer = ProjectileDrawLayerID.HeldProj` to the projectile's SetDefaults.
  * Replaced `float swingTime = owner.itemAnimationMax * Projectile.MaxUpdates` with `Projectile.GetWhipSettings(Projectile, out float timeToFlyOut, out _, out _)`
	* Projectile.GetWhipSettings has new functionality for when the whip is displayed on a mannequin.
  * Added `owner.MatchItemTimeToItemAnimation()` after setting the heldProj to match vanilla.
  * The draw code has been changed to work better for different segment amounts.
    * Previously, the draw code was specific for ExampleWhipProjectileAdvanced. Now it will work for any number of segments.
	* Even if your whips seem to draw fine, double check the code because it is likely that the third segment of your whip wasn't being drawn.
* ExampleWhipDebuff and ExampleWhipAdvancedDebuff we removed and replace with the tag system.
* Added ExampleWhipBuff that increases SummonMeleeSpeed attack speed.

### Porting Notes

See migration guide. Short version:
* Add `ItemID.Sets.UniqueTagEffects[Type] = new WhipTagEffect() { ... };` in SetStaticDefaults to you whip items to specify `TagDamage` and more.
  * Create a new class that inherits `WhipTagEffect` if you want extra functionality.
* Most buffs that were made to implement tag damage can be removed.
* Whip AI has changed slightly. `Projectile.ai[1]` is now used to set the swing direction. You'll need to spawn in the projectile manually in `Shoot()` to set it. See the migration guide for the code you need.
* Example Mod's whips have been updated with new examples and additional comments. See the migration guide or the above section for details.
* `NPCID.Sets.ImmuneToAllBuffs` was removed. Continue to use `NPCID.Sets.ImmuneToRegularBuffs` and if immunity to tags is desired, additionally set `NPCID.Sets.ImmuneToWhipTags`.


Tyfyter:
> Buff immunity and damage immunity are the only type of immunity systems that people are familiar with.
This is because they were the only type of immunity systems, adding tag immunity should have changed that, as it is immunity to an entirely disconnected concept.

Here is a full list of all NPCs which are immune to tag buffs:
Dungeon Guardian
Blazing Wheel
Martian Saucer (the part which can't be hit regardless of difficulty)
Lunatic Cultist Clone
Pirate Ship (the ship itself)
DD2LanePortal
Pink Fairy
Green Fairy
Blue Fairy
Clumsy Balloon Slime
Mystic Frog

Of these, one is the Dungeon Guardian, 4 could not be tagged by any vanilla whip as they only have 5 health, 1 could not be tagged by any whip as it has 1 health, 1 could not be tagged by any whip as it is destroyed instantly despite having more than 1 health, and the rest should be impossible to tag as they can never be hit to begin with.
Considering this, I would argue that `NPCID.Sets.ImmuneToAllBuffs` should be completely removed and replaced with a set which only affects tag immunity, as immunity to tag effects is essentially unique to the Dungeon Guardian in vanilla and should never be given without active consideration.

patches/tModLoader/Terraria/ID/NPCID.TML.cs
```diff
@@ -67,13 +67,13 @@ public NPCBestiaryDrawModifiers() : this(0) { }
...
-		/// If the NPC should be attacked, it is recommended to set <see cref="ImmuneToRegularBuffs"/> to <see langword="true"/> instead. This will prevent all debuffs except "tag" buffs (<see cref="BuffID.Sets.IsATagBuff"/>), which are intended to affect enemies typically seen as immune to all debuffs. Tag debuffs are special debuffs that facilitate combat mechanics, they are not something that adversely affects NPC.<br/><br/>
+		/// If the NPC should be attacked, it is recommended to set <see cref="ImmuneToRegularBuffs"/> to <see langword="true"/> instead. This will prevent all debuffs except "tag" effects and debuffs (<see cref="GameContent.Items.WhipTagEffect"/> and <see cref="BuffID.Sets.IsATagBuff"/>), which are intended to affect enemies typically seen as immune to all debuffs. Tag debuffs are special debuffs that facilitate combat mechanics, they are not something that adversely affects NPC.<br/><br/>
 		/// Modders can specify specific buffs to be vulnerable to by assigning <see cref="SpecificDebuffImmunity"/> to false.
 		/// </summary>
 		public static bool[] ImmuneToAllBuffs; // derived from DebuffImmunitySets
```
Tyfyter: Should be removed as it makes it simple for new modders to accidentally grant blanket immunity to a system unrelated to buffs and in the rare cases when a modder should use it, this only saves one line over a set specifically for immunity to whips.

patches/tModLoader/Terraria/GameContent/Items/UniqueTagEffect.cs.patch
```diff
@@ -1,33 +1,126 @@
...
  	public virtual void OnTaggedHit(Player owner, Projectile optionalProjectile, NPC npcHit, int calcDamage)
 +	*/
++	/// <summary>
++	/// Allows you to create special effects when tagged enemy takes damage from a minion, sentry, or one of their projectiles.
++	/// </summary>
++	/// <param name="owner">The player who inflicted the tag effect.</param>
++	/// <param name="optionalProjectile">The projectile that hit the NPC.</param>
++	/// <param name="npcHit">The NPC who was hit.</param>
++	/// <param name="calcDamage">The damage after calculations from <see cref="ModifyTaggedHit"/> and <see cref="ModifyProcHit"/>.</param>
 +	public virtual void OnTaggedHit(Player owner, Projectile optionalProjectile, NPC npcHit, float calcDamage)
```
Chicken-Bones: Please adjust `OnTagged/ProcHit` to pass `NPC.HitInfo hit` instead of `int calcDamage`

patches/tModLoader/Terraria/GameContent/Items/UniqueTagEffect.cs.patch
```diff
@@ -1,33 +1,126 @@
...
 +	*/
++	/// <summary>
++	/// Allows the damage dealt to be modified when a tagged enemy takes damage from a minion, sentry, or one of their projectiles.
++	/// </summary>
++	/// <param name="owner">The player who inflicted the tag effect.</param>
++	/// <param name="optionalProjectile">The projectile that hit the NPC.</param>
++	/// <param name="npcHit">The NPC who was hit.</param>
++	/// <param name="damageDealt">The original damage that was dealt to the NPC.</param>
++	/// <param name="crit">Whether the hit is a critical hit.</param>
 +	public virtual void ModifyTaggedHit(Player owner, Projectile optionalProjectile, NPC npcHit, ref NPC.HitModifiers damageDealt, ref bool crit)
```
Chicken-Bones: Rename `damageDealt` to `modifiers`, and remove `ref bool crit`. Replace vanilla usages with `modifiers.SetCrit()`

Chicken-Bones (changes requested):
Excellent!

Chicken-Bones (approved):

PR #5131, opened 2026-05-08
[1.4.5] Remove pylon danger teleport hooks

noahDP1911:
Deleted a bit, i hope it was not too much.

### What is the bug?
`ModPylon.ValidTeleportCheck_AnyDanger` and `GlobalPylon.ValidTeleportCheck_PreAnyDanger` still existed even though pylons no longer check for danger when teleporting in Terraria 1.4.5.
This left tModLoader exposing hooks for a vanilla pylon validation step that no longer exists.

### How did you fix the bug?
Removed the obsolete pylon danger teleport hooks from `ModPylon`, `GlobalPylon`, and `PylonLoader`.
Updated `ExamplePylonTileAdvanced`, migration/porting notes, and tModPorter so old overrides are marked as removed with an explanation.

The hooks could be kept as tModLoader-only behavior, but that would preserve an outdated validation step and make modded pylon behavior differ from vanilla. So, no reason really :D


Chicken-Bones (approved):

JavidPack:
Great job, thanks.

PR #5136, opened 2026-05-11
 ExampleMod: Demonstrate active ability handling for minions

dark-but-spark:

  This updates `ExampleSimpleMinion` to demonstrate how a minion item can support an active ability using existing tModLoader hooks.

  The example now shows:

  - using `ModItem.CanUseItem` to block normal left-click item use while the minion is already active
  - preserving right-click minion targeting behavior
  - handling held left-click input from the minion projectile AI
  - firing a summon-damage projectile from the active minion
  - adding the minion item to ExampleMod starting items for easier testing

  This is intended as an ExampleMod-only pattern. It does not add a new core API.

  The behavior is intentionally simple: once the example minion is active, using the staff again triggers the minion's active shot instead of summoning another copy.

  No additional documentation is required for this PR.

  This example is meant to document the pattern directly in ExampleMod. It also reflects the current recommendation from discussion: modders can implement this kind of active minion behavior with existing hooks instead of requiring a dedicated tModLoader API.

### DIFF
--- a/ExampleMod/Common/Players/ExampleInventoryPlayer.cs
+++ b/ExampleMod/Common/Players/ExampleInventoryPlayer.cs
@@ -1,6 +1,7 @@
 ﻿using ExampleMod.Content.Items;
 using ExampleMod.Content.Items.Placeable;
 using ExampleMod.Content.Items.Placeable.Furniture;
+using ExampleMod.Content.Projectiles.Minions;
 using System.Collections.Generic;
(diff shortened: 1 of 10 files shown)


JavidPack:
I've started a thread on Discord, there have been some interesting points brought up so far. Please take a look if you can. https://discord.com/channels/103110554649894912/1503505995677171802

PR #5137, opened 2026-05-11
Tooltip line color changes

JavidPack:
1.4.5 made some changes to how tooltip line colors are determined, and these changes can be leveraged to simplify the `TooltipLine` and `DrawableTooltipLine` code and corresponding patches to `Main.cs`.

In 1.4.4, tooltip line colors were determined after collecting the tooltip lines based on various hard-coded conditions, such as if the line was the journey mode research line, a bad prefix modifier, or a good prefix modifier. This approach led to the `TooltipLine` needing to contain `IsModifier` and `IsModifierBad` fields as well as a separate `OverrideColor` for bypassing vanilla color choice logic, which happened after lines were collected.

In 1.4.5 now, the colors are calculated as the lines are collected, allowing us to greatly simplify the logic.

Changes: 
* `TooltipLine`'s `Color? OverrideColor` is now `Color Color`.
  * Since it is the line color directly, the old "override" moniker no longer makes sense.
  * `DrawableTooltipLine`'s `public new Color? OverrideColor` was also removed. 
  * `DrawableTooltipLine`'s `public Color Color` is now `public new Color Color { get; internal set; }` to once again prevent changes.
* `TooltipLine.IsModifier` and `TooltipLine.IsModifierBad` were removed in 355f07e. 
  * `Terraria.ID.Colors.PrefixGood` and `Terraria.ID.Colors.PrefixBad` have been added, modders will now set `TooltipLine.Color` directly. 
  * `DrawableTooltipLine`'s `public new readonly bool IsModifier;` and `public new readonly bool IsModifierBad;` were also removed.

Discussion
* Since `TooltipLine.IsModifier` and `TooltipLine.IsModifierBad` have been removed, it is no longer possible to accurately determine if a tooltip line is for a modifier or not, aside from checking the color.
* We'll need to check with modders to see if these changes prevents implementing the same logic as 1.4.4 era mods. Some mods might have been checking `DrawableTooltipLine.Color` and `OverrideColor`, but I can't imagine why.

# Other Changes
* The "SocialDesc" tooltip line has been removed. Docs updated.
* `Item.hasVanityEffects` is now used, it was previously unused by vanilla. Docs updated.

# TODO
- [x] Need to update MigrationGuide with the porting notes for the `TooltipLine`/`DrawableTooltipLine` class changes.

### DIFF
--- a/ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
+++ b/ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
@@ -105,9 +105,9 @@ namespace ExampleMod.Common.GlobalItems
 
 		public override void ModifyTooltips(Item item, List<TooltipLine> tooltips) {
 			if (experience > 0) {
-				tooltips.Add(new TooltipLine(Mod, "level", LevelText.Format(level)) { OverrideColor = Color.LightGreen });
+				tooltips.Add(new TooltipLine(Mod, "level", LevelText.Format(level)) { Color = Color.LightGreen });
 				int nextLevelExperience = (level + 1) * experiencePerLevel - experience;
-				tooltips.Add(new TooltipLine(Mod, "experience", ExperienceText.Format(experience, nextLevelExperience)) { OverrideColor = Color.White });
+				tooltips.Add(new TooltipLine(Mod, "experience", ExperienceText.Format(experience, nextLevelExperience)) { Color = Color.White });
 			}
 		}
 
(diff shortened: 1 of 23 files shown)


JavidPack:
There was some discussion in https://discord.com/channels/103110554649894912/1503510653145583656 mostly discussing if it was OK to removed the IsModifier fields, and in general it seems that mods should be able to adapt. Just because it happened to be treated specially in past versions, we don't feel it is necessary to treat those lines specially just because the old code design forced the weird fields into those versions of our abstraction.

PR #5140, opened 2026-05-12
Add support for modded crowd control effects

### LINKED ISSUE #5025: CrowdControled Property

No

Mod capability as a Modder

### Description

Player.CCed is currently unable to be enabled independently of the debuffs that impliments it in vanilla, adding a 4th bool to the or chain with a custom property modders can enable without triggering the secondary visual effects of these other debuffs would be very convenient

_No response_

_No response_

hezuo1hao:
### What is the new feature?
Add Player.SetCCed. Fixes #5025

At present, it is difficult to modify Player.CCed in mods, so corresponding methods have been added to enable mods to set it to true, This is also proposed by #5025

This should be the simplest and quickest solution

Please refer to ExampleFrozenDebuff. cs for details

### ExampleMod updates
Add ExampleFrozenDebuff.cs and modify MinionBossEye


JavidPack:
Thanks for this, should be useful.

PR #5141, opened 2026-05-12
Declarative Armor set bonuses support

JavidPack:
1.4.5 added the `ArmorSetBonuses` and `ArmorSetBonus` classes. These replace the old 1.4.4 `Player.UpdateArmorSets` approach. Now, all vanilla armor sets are registered as data, rather than hard-coded logic. As such, armor sets can now be queried directly by mods interested in armor sets and armor set effects.

This PR updates the modder API to utilize this new approach and documents all relevant new members. We could just leave it as it is and let mods work as before, but having all mods and vanilla on the new system should be beneficial in the long term.

# Changes

- `Player.setBonus` was removed. It was left in the 1.4.5 source code but was not used at all. `Player.UpdateArmorSetsOld` was also commented out to prevent misleading search results.
- `ModItem.UpdateArmorSet` and `GlobalItem.UpdateArmorSet` now have an `ArmorSetBonus armorSetBonus` parameter.
- `ModItem.IsArmorSet` and `GlobalItem.IsArmorSet` removed.
- Many new methods to facilitate registering modded armor sets, including `ModItem.AddArmorSet`/`CreateArmorSet`.

# Migration Information

Armor sets are now declarative, meaning that they are stored as data rather than hard-coded logic. This change helps facilitate showing the "Set Bonus (NumEquipped/NumTotal)" tooltip line and other features. The new armor set system, found in the `ArmorSetBonuses` and `ArmorSetBonus` classes, allows armor sets to be created and registered. Each armor set is an `ArmorSetBonus` object that contains the head, body, and legs armor items, code for the armor set effects, a description, and optionally an indication of a primary equipment slot. The system supports optional equipment slots and multiple options for each equipment slot.

Mods will need to update their code to use the new system, `ModItem.IsArmorSet` has been removed in favor of ensuring that all armor sets are properly declared as armor sets to take advantage of the features of the new system.

#### Examples
Here are some examples of declaring armor sets: 
```cs
// These examples show code found in SetStaticDefaults of a helmet item (ExampleHelmet):
int body = ModContent.ItemType<ExampleBreastplate>();
int legs = ModContent.ItemType<ExampleLeggings>();

// Basic armor set. Consists of a full set of items. No substitutions. 
AddArmorSet(Type, body, legs);

// Basic armor set alternate approach. Generics can be used if all items are ModItems.
AddArmorSet<ExampleHelmet, ExampleBreastplate, ExampleLeggings>(); 

// Basic armor set with no legs. This is how the Wizard Hat set is implemented.
AddArmorSet(Type, body, ItemID.None);

// Armor set with multiple options and the same behavior. In this case, both ExampleHelmet and ExampleHood will satisfy this armor set. This is how the Necro armor set is implemented
// For armor sets with multiple options, the CreateArmorSet method is followed by any number of Set methods to define the options, then Add must be called at the end.
int alternateHelmet = ModContent.ItemType<ExampleHood>();
CreateArmorSet().Set(Type, body, legs).Set(alternateHelmet, body, legs).Add();

// Same as above but using generics.
CreateArmorSet().Set<ExampleHelmet, ExampleBreastplate, ExampleLeggings>().Set<ExampleHood, ExampleBreastplate, ExampleLeggings>().Add();

// Armor set with multiple options and the same behavior. In this case either complete set will satisfy and apply the armor set. Note that the items aren't interchangeable and can't be mixed and matched. This is how the Wood armor sets are implemented.
int alternateBody = ModContent.ItemType<ExampleBreastplate_Alt>();
int alternateLegs = ModContent.ItemType<ExampleLeggings_Alt>();
CreateArmorSet().Set(Type, body, legs).Set(alternateHelmet, alternateBody, alternateLegs).Add();

// Armor set with multiple options and the same behavior. In this case every equipment is interchangeable. For example ExampleLeggings_Alt can be used with ExampleHelmet and ExampleBreastplate. This is how the Shadow/Ancient Shadow armor sets are implemented.
CreateArmorSet().Set([Type, alternateHelmet], [body, alternateBody], [legs, alternateLegs]).Add();

// Armor sets sharing armor items except and having a primary item that changes the tooltip. 
// The PartType parameter lets the game know that the head and legs items are part of multiple armor sets with different set bonus tooltips, so "Changes with (head/torso/leg) piece" is shown instead of the the set bonus tooltip. This is how the Beetle armor set is implemented.
// Customizing localization and the identifier is explained in the next section.
AddArmorSet(Type, body, legs, this.GetLocalization("SetBonus_Melee"), ArmorSetBonus.PartType.Body, "ExampleMod/ExampleArmor_Melee");
AddArmorSet(Type, alternateBody, legs, this.GetLocalization("SetBonus_Ranged"), ArmorSetBonus.PartType.Body, "ExampleMod/ExampleArmor_Ranged");
```

The examples above all used `ModItem` helper methods that automate registering a localization key, assigning the armor set effect to `ModItem.UpdateArmorSet`, and assigning a default identifier string. These can all be customized by providing values for the optional parameters of the methods. The set bonus text can be customized by passing in a `LocalizedText`. The code that executes can be customized as well by passing in a delegate. The identifier can also be passed in, which is useful if you need to differentiate between multiple set bonuses that use some of the same items. The `ArmorSetBonuses.Add` and `ArmorSetBonuses.Create` methods could be used instead of the `ModItem.AddArmorSet` and `ModItem.CreateArmorSet` helper methods as well, such as if used in `GlobalItem`.

```cs
// Customize the set bonus tooltip. We can pass in any localization key rather than the automatically derived "Mods.{ModName}.Items.{ItemName}.SetBonus" key. We can also pass in a LocalizedText object.
// In this example, we are using the automatic key but populating it with a value. This helps keep the set bonus effect in line with the tooltip as a mod is updated.
LocalizedText SetBonusText = this.GetLocalization("SetBonus").WithFormatArgs(AdditiveGenericDamageBonus);
AddArmorSet(Type, body, legs, SetBonusText);

// The SetBonus_Melee and SetBonus_Ranged examples above show using a custom key and a custom identifier.

// And of course, we can use a specific localization key, either by typing it out or using Mod.GetLocalizationKey, which can be useful for organization.
AddArmorSet(Type, body, legs, Mod.GetLocalizationKey("CommonSetBonuses.Example"));

// Finally, we can customize the code that runs for each armor set. By default, the UpdateArmorSet method is called. We could check the data on the ArmorSetBonus passed into the method to implement custom code for alternate armor sets, but another option is to pass in a ArmorSetBonus.ArmorSetEffect delegate directly. Here is a full example of that:

public static readonly int AdditiveMeleeDamageBonus = 15;
public static readonly int AdditiveRangedDamageBonus = 25;

	int body = ModContent.ItemType<ExampleBreastplate>();
	int alternateBody = ModContent.ItemType<ExampleBreastplate_Alt>();
	int legs = ModContent.ItemType<ExampleLeggings>();

	var SetBonusTextMelee = this.GetLocalization("SetBonus_Melee").WithFormatArgs(AdditiveMeleeDamageBonus);
	ArmorSetBonuses.Add(Type, body, legs, SetBonusTextMelee, ArmorSetBonus.PartType.Body, "ExampleMod/ExampleArmor_Melee", UpdateArmorSetMelee);
	var SetBonusTextRanged = this.GetLocalization("SetBonus_Ranged").WithFormatArgs(AdditiveRangedDamageBonus);
	ArmorSetBonuses.Add(Type, alternateBody, legs, SetBonusTextRanged, ArmorSetBonus.PartType.Body, "ExampleMod/ExampleArmor_Ranged", UpdateArmorSetRanged);
}

private void UpdateArmorSetMelee(Player player) {
	player.GetDamage(DamageClass.Melee) += AdditiveMeleeDamageBonus / 100f;
}
private void UpdateArmorSetRanged(Player player) {
	player.GetDamage(DamageClass.Ranged) += AdditiveRangedDamageBonus / 100f;
}
```

# Porting Notes
* 🤖: `ModItem.UpdateArmorSet` and `GlobalItem.UpdateArmorSet` now have an `ArmorSetBonus armorSetBonus` parameter.
* ⚙️: `ModItem.IsArmorSet` and `GlobalItem.IsArmorSet` removed. Use `AddArmorSet` as described above to replace this.
* Using vanilla armor from existing armor sets in modded sets is now potentially confusing to the user due to how the set bonus tooltip is calculated and displayed. The game will show the armor set tooltip for an item in the inventory (not just the equipment slots), so if a modded armor set also uses that item, it might potentially show the modded armor set bonus tooltip rather than the original vanilla armor set bonus tooltip the player might be looking for. 

# Discussion
- ~~The current vanilla code for registering an armor set effect takes a delegate: `public delegate void ArmorSetEffect(Player player);`. We could adopt this technique easily, but using this approach directly would prevent using instance-specific logic in armor set effects. This is probably unacceptable and would likely make some mods unable to port their effects properly.~~ I've decided to let mods continue using `UpdateArmorSet`, which now passes in a `ArmorSetBonus` that can have its `Identifier` checked if necessary.
- ~~I'd like to add a `ArmorSetBonus` parameter to `ModItem.UpdateArmorSet` for differentiating between multiple sets registered from the same class, but it might be hard.~~ I left `ArmorSetBonus.Effect` unchanged and just call `ModItem.UpdateArmorSet` always. The complication was avoided.
- Since `Player.setBonus` was removed, modifying the set bonus text directly for vanilla tweaks or other dynamic considerations is not possible yet. We'll need to revisit this.


ExampleMod/Content/Items/Armor/ExampleHood.cs
```diff
@@ -34,7 +35,7 @@ public override void SetDefaults() {
 		}
 
 		// UpdateArmorSet allows you to give set bonuses to the armor.
-		public override void UpdateArmorSet(Player player) {
+		public override void UpdateArmorSet(Player player, ArmorSetBonus armorSetBonus) {
 			player.manaCost -= ManaCostReductionPercent / 100f; // Reduces mana cost by 10%
 			player.GetModPlayer<ExampleArmorSetBonusPlayer>().ExampleSetHood = true;
```
Chicken-Bones, line 40: Seems to me that we should remove `UpdateArmorSet` from `ModItem` at least

JavidPack: ModItem.UpdateArmorSet allows access to instanced data (this.ExtraDefenseBonus). If a modder instead passes in a delegate, they might unintentionally be accessing instanced data of the template item instead of the equipped item instance. I figure that would be a pitfall, and it would force mods to change where their armor set effect code is for pretty much no reason.

It also allows sets with variants and common logic to keep shared code all in one place and differentiate base on the passed in armorSetBonus.Identifier.

Chicken-Bones: Hmm... I'm only partially convinced. This does 'keep it working' so to speak, but our examples don't show accessing instance data. In practice, I think most armor set bonuses would require a `ModPlayer` and should go on/near the `ModPlayer` rather than on one of the pieces. Having the code in the head is fine, but we could at least pass a static method for the effect. It's pretty easy for modders who have custom instance data to go and retrieve the actual head item. 

To summarise, I think keeping the hook is fine, especially because it's guarded on `armorSetBonus.Head/Body/Legs == head/body/legs.type`, but I'm not sure we should 'teach' it in ExampleMod. Let me know your thoughts

patches/tModLoader/Terraria/DataStructures/ArmorSetBonus.cs.patch
```diff
@@ -0,0 +1,93 @@
...
++	/// </summary>
+ 	public ArmorSetEffect Effect;
+ 	public LocalizedText Description;
+ 	public int Head;
+@@ -205,5 +_,5 @@
+ 		});
+ 	}
+ 
+-	public static Builder Create(ArmorSetEffect effect, string textKey, PartType primaryPart = PartType.None) => new Builder(effect, textKey, primaryPart);
++	public static Builder Create(ArmorSetEffect effect, string textKey, PartType primaryPart = PartType.None, string Identifier = null) => new Builder(effect, textKey, primaryPart, Identifier);
```
Chicken-Bones, line 92: Probably, the no `Identifier` overload should be `private`, and should pass `textKey = Identifier` Then you don't need `AssignKeysToVanillaArmorSets`

Should we force mods to pass `LocalizedText` for the key?

Chicken-Bones (commented):
We can probably make an example of 'adding to' a vanilla set bonus, and updating the text by replacing `ArmorSetBonus.Description`, we can showcase the loc being `{0}\nExtra effect` and using `set.Description = NewDescription.WithFormatArgs(set.Description)` as a way to 'concatenate with the existing description, even if multiple mods add to the same vanilla set bonus'

PR #5142, opened 2026-05-13
chore: Add necessary VSCode layout & optimization configs.

Mirsario:
Out of the box, VSCode is unusable as an editor for this repo, as it:
- Uses gitignore to hide files in the explorer view, hiding the `src/` folder.
- Has all `patches/` participate in quick file pickers.
- Stresses the language server provider too much for nesting resolution, resulting in extremely poor performance.

This workspace config solves that, with its only bias being exclusion of non-tModLoader steps from quick file picking. `Terraria` and `TerrariaNetCore` files can still be opened via the solution explorer window, i.e. the mouse.


Chicken-Bones:
> This workspace config solves that, with its only bias being exclusion of non-tModLoader steps from quick file picking. `Terraria` and `TerrariaNetCore` files can still be opened via the solution explorer window, i.e. the mouse.

Would it be possible to cascade further `.vscode/settings` in `src/Terraria` and `src/TerrariaNetCore`, since I'm thinking to move their respective `.sln` files there?

Mirsario:
Using multi-root workspaces maybe. Would need to look into it.
https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces
I'm not sure how search/quickpickers would work in that context.

Chicken-Bones:
> Using multi-root workspaces maybe. Would need to look into it. https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces I'm not sure how search/quickpickers would work in that context.

Please consider/test with #5143

PR #5143, opened 2026-05-14
Move solution files to suitable root directories

Chicken-Bones:
### What is the new feature?

`tModLoader.sln` moved to respository root
`Terraria.sln` moved to `src/Terraria`
`TerrariaNetCore.sln` moved to `src/TerrariaNetCore`

This will break developers shortcut paths and reset their local `.vs` folder state stores, but after that everything should be fine again.

A followup PR will look at renaming the `solutions` directory to something like `buildextras`

Rider and VS Code both have some features that strongly prefer the opened `.sln` file to be at the root of all the 'working files' for the solution. 

Note that `TerrariaNetCore.sln` uses `FNA` so it may be better served at the repo root per that rule, but it's a tradeoff between further cluttering the base directory. 

Move `TerrariaNetCore.sln` to the root directory.

### DIFF
--- a/patches/Terraria/Terraria.sln
+++ b/patches/Terraria/Terraria.sln
@@ -3,9 +3,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31903.59
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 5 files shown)


Mirsario (approved):
Placing the other solution files in the root as well does not sound bad either, would deduplicate them between patches and src.

PR #5146, opened 2026-05-16
[1.4.5] Restore ItemLoader.Update and PostUpdate hooks for WorldItem

JerryAZR:
### What is the bug?

 In 1.4.5, vanilla moved all in-world item physics from Item.MoveInWorld() and Item.Update() into the new WorldItem class. The tModLoader patches that injected ItemLoader.Update and ItemLoader.PostUpdate into those methods failed because the target methods no longer exist in Item.cs. This left behind two stale hunks in patches/tModLoader/Terraria/Item.cs.rej and broke modded items' ability to:
 - Modify gravity and max fall speed while lying in the world
 - Run custom logic every tick via PostUpdate (e.g., spawning dust or emitting light)

 ### How did you fix the bug?

 - Added ItemLoader.Update(this, ref gravity, ref maxFallSpeed); at the very start of WorldItem.MoveInWorld(), before vanilla's gravity logic runs. This restores the pre-1.4.5 behavior where mods can override gravity and fall speed.
 - Added ItemLoader.PostUpdate(this); inside WorldItem.UpdateItem(), immediately after timeSinceItemSpawned++. This restores the post-update hook so mods can run per-tick logic on world items.
 - Removed the two now-obsolete rejected patch hunks from Item.cs.rej since the hooks now live in WorldItem.cs.patch.

 - Placing the hooks back in Item.cs: This is not viable. Item no longer has position, velocity, or MoveInWorld() in 1.4.5, so the hooks must live in WorldItem where the actual world state exists.
 - Calling Update later inside MoveInWorld: We deliberately placed it at the top, before vanilla checks ItemID.Sets.ItemNoGravity. This allows a mod to set gravity = 0 and have it take effect on the same frame, matching pre-1.4.5 behavior.
 - Calling PostUpdate elsewhere in UpdateItem: The current placement (after timeSinceItemSpawned++) mirrors the original location in Item.Update() and is the natural point where all vanilla update work for the tick is complete.

### DIFF
--- a/PortingNotes_1.4.5.md
+++ b/PortingNotes_1.4.5.md
@@ -79,7 +79,7 @@ Once all patches are fixed, these items need to be fixed or double checked:
   - "This is unused, replaced with this.ArmorPenetration." patch might be incorrect as well. Nearby switch table also changed a lot, might need to apply them elsewhere.
 - Vanilla CanHavePrefixes logic changed, might be able to use it rather than tml changes.
   - #StackablePrefixWeapons needs to be searched for and removed
-- Item Shimmer/Update/CheckLavaDeath/MoveInWorld/GetPickedUpByMonsters_Special/FindOwner/getRect/GetShimmered/CombineWithNearbyItems/related methods have moved to World Item. Need to move docs/patches over.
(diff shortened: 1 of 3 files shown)


JavidPack:
Looks good, thanks.

PR #5147, opened 2026-05-17
[1.4.5] Restore CanStack, CanStackInWorld, and OnStack hooks for WorldItem stacking

JerryAZR:
### What is the bug?

In 1.4.5, vanilla moved all in-world item stacking logic from `Item.cs` into the new `WorldItem.TryCombiningIntoNearbyItems()` method. The tModLoader patch that injected `ItemLoader.CanStack`, `ItemLoader.CanStackInWorld`, and `ItemLoader.StackItems` (which fires `OnStack`) failed because its target code no longer exists in `Item.cs`. This left a stale hunk in Item.cs.rej and broke three modding hooks:

| Hook | What broke | Observable effect |
|---|---|---|
| CanStack | Never called during world merging | Mods can't block stacking based on prefix/type |
| CanStackInWorld | Never called during world merging | Mods can't block world-only stacking |
| OnStack | Never called during world merging | No custom effects (dust, sound, etc.) when items merge on the ground |

 Inventory and chest stacking still worked because those paths already call TryStackItems/StackItems separately.

### How did you fix the bug?

 Added the three hook calls to `WorldItem.TryCombiningIntoNearbyItems()` in `src/tModLoader/Terraria/WorldItem.cs`:

 1. `ItemLoader.CanStack(inner, worldItem.inner)` — general compatibility check (type, prefix, etc.)
 2. `ItemLoader.CanStackInWorld(this, worldItem)` — world-specific stacking permission
 3. `ItemLoader.StackItems(inner, worldItem.inner, out int numTransferred)` — performs the transfer and fires `OnStack`

 Also added explanatory TML comments clarifying:
 - Why `CanStack` comes before `CanStackInWorld`
 - Why `CanStackInWorld` is world-only and doesn't affect inventory/chests
 - Why `OnStack` firing in all contexts matches 1.4.4 behavior

 Removed the now-obsolete `@@ -48669` hunk from `Item.cs.rej`.

 - Call hooks in a different order: `CanStack` → `CanStackInWorld` is the same order as the original 1.4.4 patch. Reversing it would let `CanStackInWorld` run before type/prefix validation, which doesn't make sense.
 - Use manual arithmetic instead of `StackItems`: Would skip `OnStack`, which is exactly the bug we're fixing.
 - Merge `CanStackInWorld` into `CanStack`: These are intentionally separate hooks with different purposes. `CanStack` is for general compatibility; `CanStackInWorld` is specifically for world physics decisions (e.g., "don't merge while this item is exploding").

### DIFF
--- a/patches/tModLoader/Terraria/Item.cs.rej
+++ b/patches/tModLoader/Terraria/Item.cs.rej
@@ -165,34 +165,6 @@
  				flag3 = true;
  				NPC.TransformCopperSlime(j);
  				break;
-@@ -48669,14 +_,26 @@
- 				num2 *= 2;
- 
- 			if (num < (float)num2) {
-+				if (!ItemLoader.CanStack(this, item))
-+					continue;
-+
-+				if (!ItemLoader.CanStackInWorld(this, item))
-+					continue;
-+
- 				position = (position + item.position) / 2f;
- 				velocity = (velocity + item.velocity) / 2f;
-+
-+				ItemLoader.StackItems(this, item, out _);
-+
-+				/* #OnStackHook
- 				int num3 = item.stack;
- 				if (num3 > maxStack - stack)
- 					num3 = maxStack - stack;
- 
- 				item.stack -= num3;
- 				stack += num3;
-+				*/
-+
- 				if (item.stack <= 0) {
- 					item.SetDefaults();
- 					item.active = false;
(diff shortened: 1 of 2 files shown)


JavidPack:
This is good for now. We will be moving ItemLoader.CanStack into Item.CanStack sometime later, but fixing these for now is great.

Issue #5135, opened 2026-05-10
Expand BossLoot hook to include heart drop amount


direwolf420:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Currently `ModNPC.BossLoot` hook only gives control over the dropped potion type (i.e. lesser healing potion for pre-hm bosses, and regular ones for hm bosses), which is a valid use case to match vanilla behavior. 
An additional behavior modders may want to tweak is the amount of Hearts dropped (the ones that heals for 20 HP each), the amount is currently fixed for every boss in the game.

The proposal aims to give modders control over this amount, and potentially even set it to 0 so hearts don't drop at all.

### What does this proposal attempt to solve or improve?

Give modders more control without having to resort to workarounds like `BossLoot() => NPCLoader.blockLoot.Add(ItemID.Heart);` (which is buggy, see below) or `PreKill() => NPC.boss = false;` (which has alot more side effects).

### Which (other) solutions should be considered?

Multiple issues I discovered while looking into this:
* Modders may also want to have no potion drop at all. That is currently possible by setting the potionType to 0, but that will still spawn an "empty item" including getting hooks called on it during creation, so maybe the NewItem base method should have a check to return early if the type is detected to be `<= 0`, or the code spawning the potion should have that.
* The `NPCLoader.blockLoot.Add()` method does not work for drops that happen after `NPCLoader.OnKill`, which includes the use cases mentioned above, so the workaround above has to be used in an otherwise unusual callsite (which is intended to be `PreKill`, and not `BossLoot`). For that to function correctly however, the `NPCLoader.blockLoot.Clear()` should be placed at the end of `NPC.NPCLoot` instead of `NPCLoader.OnKill`, otherwise hearts will no longer drop from non-enemy sources like Heart Statues, until an enemy dies or a tree is shaken.

JavidPack:
Seems like updating the hook to have the potion stack and heart stack should satisfy the desired customizability.

CPeluti:
Opened a PR for this issue (#5154).

direwolf420:
@JavidPack do you think "the NPCLoader.blockLoot.Clear() should be placed at the end of NPC.NPCLoot instead of NPCLoader.OnKill" should be a separate issue/PR?

CPeluti:
[...]

PR #5154, opened 2026-05-20
Expand BossLoot hook to include heart and potion drop amount 

CPeluti:
### What is the new feature?
Allows mods to modify the amount of potions and hearts dropped from a boss loot (#5135).
Currently mods need to workaround the current implementation to be able to specify the amount of hearts and potions dropped by a boss when killed, that usually causes buggy behavior or unexpected side effects.
Instead of dropping a amount of single hearts, it could be just a one or a few hearts with a more dense stack.
```C#
public override void BossLoot(ref int potionType, ref int potionStack, ref int heartStack) {
	potionStack = 50;
	heartStack = 99;
}
```
### ExampleMod updates
Incremented the BossLoot comment to cover the new feature at the MinionBossBody Class

### Porting Notes
It should work as it is. The feature shouldn't affect existing mods.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
@@ -447,13 +447,17 @@ public abstract class ModNPC : ModType<NPC, ModNPC>, ILocalizedModType
 	{
 	}
 
+	[OriginalOverload]
 	[Obsolete("Use BossLoot(ref int potionType) instead. The name can be changed by modifying DeathMessage")]
 	public virtual void BossLoot(ref string name, ref int potionType)
 	{
 	}
-
+	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.")]
+	public virtual void BossLoot(ref int potionType)
+	{
+	}
 	/// <summary>
-	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops.
+	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops, the amount of potions and the amount of hearts.
 	/// <para/> Called in single player or on the server only.
 	/// </summary>
 	public virtual void BossLoot(ref int potionType, ref int potionStack, ref int heartStack)
--- a/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
@@ -456,7 +456,7 @@ public abstract class ModNPC : ModType<NPC, ModNPC>, ILocalizedModType
 	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops.
 	/// <para/> Called in single player or on the server only.
 	/// </summary>
-	public virtual void BossLoot(ref int potionType)
+	public virtual void BossLoot(ref int potionType, ref int potionStack, ref int heartStack)
 	{
 	}
 
--- a/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
@@ -447,19 +447,18 @@ public abstract class ModNPC : ModType<NPC, ModNPC>, ILocalizedModType
 	{
 	}
 
-	[OriginalOverload]
-	[Obsolete("Use BossLoot(ref int potionType) instead. The name can be changed by modifying DeathMessage")]
+	[Obsolete("Use BossLoot(ref int potionType) instead. The name can be changed by modifying DeathMessage", error: true)]
 	public virtual void BossLoot(ref string name, ref int potionType)
 	{
 	}
 
-	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.")]
+	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.", error: true)]
 	public virtual void BossLoot(ref int potionType)
 	{
 	}
 
 	/// <summary>
-	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops, the amount of potions and the amount of hearts.
+	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops, the amount of potions, and the amount of hearts.
 	/// <para/> Called in single player or on the server only.
 	/// </summary>
 	public virtual void BossLoot(ref int potionType, ref int potionStack, ref int heartStack)
--- a/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
+++ b/patches/tModLoader/Terraria/ModLoader/ModNPC.cs
@@ -452,10 +452,12 @@ public abstract class ModNPC : ModType<NPC, ModNPC>, ILocalizedModType
 	public virtual void BossLoot(ref string name, ref int potionType)
 	{
 	}
+
 	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.")]
 	public virtual void BossLoot(ref int potionType)
 	{
 	}
+
 	/// <summary>
 	/// Allows you to customize what happens when this boss dies, as well as letting you modify the type of potion it drops, the amount of potions and the amount of hearts.
 	/// <para/> Called in single player or on the server only.
--- a/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
+++ b/patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
@@ -567,10 +567,14 @@ public static class NPCLoader
 			g.ModifyGlobalLoot(globalLoot);
 		}
 	}
-
-	public static void BossLoot(NPC npc, ref string name, ref int potionType, ref int potionStack, ref int heartStack)
+	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.")]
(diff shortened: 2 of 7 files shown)


JavidPack:
> ### Porting Notes
> It should work as it is. The feature shouldn't affect existing mods.

This is not true. Your changes here break both binary and source code compatibility. All mods using this method will need to update in order to load once these changes goes live.

We'll want to preserve compatibility by preserving the old method and marking it obsolete. You'll also need to add code to tModPorter to handle updating to the new method signature.

Please read this section:
https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#be-aware-of-breaking-changes

In that section it links to the [Contributing to tModPorter](https://github.com/tModLoader/tModLoader/wiki/Contributing-to-tModPorter) guide, which you will also want to read.

The [GlobalInfoDisplay.ModifyDisplayParameters refactor](https://github.com/tModLoader/tModLoader/commit/3ca9bf16a3722c35ea86377b7a28f4456b072743) example linked in that section is a good example of what needs to be added to this PR to preserve compatibility.

If any of that doesn't seem like something you want to learn, let me know and I can finish off the PR for you.

CPeluti:
> > ### Porting Notes
> > It should work as it is. The feature shouldn't affect existing mods.
> 
> This is not true. Your changes here break both binary and source code compatibility. All mods using this method will need to update in order to load once these changes goes live.
> 
> We'll want to preserve compatibility by preserving the old method and marking it obsolete. You'll also need to add code to tModPorter to handle updating to the new method signature.
> 
> Please read this section: https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide#be-aware-of-breaking-changes
> 
> In that section it links to the [Contributing to tModPorter](https://github.com/tModLoader/tModLoader/wiki/Contributing-to-tModPorter) guide, which you will also want to read.
> 
> The [GlobalInfoDisplay.ModifyDisplayParameters refactor](https://github.com/tModLoader/tModLoader/commit/3ca9bf16a3722c35ea86377b7a28f4456b072743) example linked in that section is a good example of what needs to be added to this PR to preserve compatibility.
> 
> If any of that doesn't seem like something you want to learn, let me know and I can finish off the PR for you.

I'm going to take a look later today. That said, where's the right place to ask questions about this?

JavidPack:
> I'm going to take a look later today. That said, where's the right place to ask questions about this?

Chatting here it fine, but if you want quicker answers or a back and forth our [Discord chat](https://discord.gg/tmodloader) is where development discussions primarily happen. You can ping me (jopojelly) there and we can go from there.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -447,16 +447,20 @@ public virtual void ModifyNPCLoot(NPCLoot npcLoot)
 	{
 	}
 
+	[OriginalOverload]
 	[Obsolete("Use BossLoot(ref int potionType) instead. The name can be changed by modifying DeathMessage")]
 	public virtual void BossLoot(ref string name, ref int potionType)
 	{
 	}
-
```
CPeluti, line 454: I've marked the old method with obsolete as required.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -447,16 +447,20 @@ public virtual void ModifyNPCLoot(NPCLoot npcLoot)
 	{
 	}
 
+	[OriginalOverload]
```
CPeluti: I've also marked the original method as OriginalOverload as recommended in the [Style Guide](https://github.com/tModLoader/tModLoader/wiki/tModLoader-Style-Guide)

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -567,12 +567,16 @@ public static void ModifyGlobalLoot(GlobalLoot globalLoot)
 			g.ModifyGlobalLoot(globalLoot);
 		}
 	}
-
+	[Obsolete("BossLoot has now parameters to control the amount of potions and hearts.")]
```
CPeluti, line 571: I also marked this method as obsolete since it needed more params.

CPeluti (commented):
@JavidPack As you've recommended me. I've marked the old methods with obsolete and added the necessary  modifications to the tModPorter. That said, I had some problems related to the later and would recommend you to test to see if everything is alright there. 

There were too much tests failing and some were in a infinite loop state, so I had to make some alterations to be running on my environment.

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
```diff
@@ -447,16 +447,20 @@ public virtual void ModifyNPCLoot(NPCLoot npcLoot)
...
 	[Obsolete("Use BossLoot(ref int potionType) instead. The name can be changed by modifying DeathMessage")]
```
direwolf420: Incorrect text, what is DeathMessage here?

patches/tModLoader/Terraria/ModLoader/NPCLoader.cs
```diff
@@ -567,12 +567,16 @@ public static void ModifyGlobalLoot(GlobalLoot globalLoot)
...
 	public static void BossLoot(NPC npc, ref string name, ref int potionType)
 	{
 		npc.ModNPC?.BossLoot(ref name, ref potionType);
 		npc.ModNPC?.BossLoot(ref potionType);
 	}
+	public static void BossLoot(NPC npc, ref string name, ref int potionType, ref int potionStack, ref int heartStack)
```
direwolf420: To stay consistent with the rest of the file, a line break between end of method and start of next method is needed here.

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -2083,15 +2083,22 @@
...
-+		NPCLoader.BossLoot(this, ref typeName, ref num);
+-		Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);
++
 +
- 		Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);
  		int num2 = Main.rand.Next(5) + 5;
++
++		NPCLoader.BossLoot(this, ref typeName, ref num, ref stack, ref num2);
++
++		Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);
```
direwolf420: As per the original issue, I would suggest surrounding this with a check if `num > 0 && stack > 0`, otherwise an empty item is spawned. Not necessary for the hearts as the loop condition prevents that already

direwolf420 (commented):
Thanks for working on the issue! I left a few comments

patches/tModLoader/Terraria/ModLoader/ModNPC.cs
CPeluti: > Incorrect text, what is DeathMessage here?

I have no clue since that was the obsolete message that existed before that.

CPeluti: Ok, I've read a bit the old code and the previous implementation had BossLoot(ref string name, ref int potionType). I suppose the string was an message to be exhibited when the boss died (thus DeathMessage)

direwolf420: Alright makes sense

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -2093,8 +2093,10 @@
  		int num2 = Main.rand.Next(5) + 5;
 +
 +		NPCLoader.BossLoot(this, ref typeName, ref num, ref stack, ref num2);
++		if (stack > 0 && num > 0) {
++			Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);
++		}
 +
-+		Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);
```
direwolf420: A more elegant solution here would be to write it like
```cs
if (stack == 0 || num == 0)
    goto skipPotionDrop;

Item.NewItem(GetItemSource_Loot(), (int)position.X, (int)position.Y, width, height, num, stack);

skipPotionDrop:
```
This should reduce the patch as you will no longer have an indentation change and makes it more maintainable

CPeluti: First time I've encountered a suggestion to use goto. Gonna use that right now.

JavidPack: Looks like you manually edited the patch file. Please use setup to generate patches next time, otherwise the patcher errors out due to malformed patches.

JavidPack:
Thanks, good job.

patches/tModLoader/Terraria/NPC.cs.patch
CPeluti: > Looks like you manually edited the patch file. Please use setup to generate patches next time, otherwise the patcher errors out due to malformed patches.

I did generate with the setup-cli. But I had to pick the commits manually. Am I supposed to do that? Or should I just commit everything that comes from the 1.4.5 branch? Every time I pull the updates and run the diff for tmodloader, appears a new change which isn't related to my changes.

Issue #5149, opened 2026-05-17
Centralize stacking compatibility checks: merge `ItemLoader.CanStack` into `Item.CanStack`


JerryAZR:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

tModLoader code as a Contributor/Maintainer

### Description

`tModLoader` currently maintains two independent stacking compatibility checks:

1. `Item.CanStack(Item, Item)` — vanilla-only; checks `type == type && prefix == prefix`
2. `ItemLoader.CanStack(Item, Item)` — mod hooks only; checks prefix + `GlobalItem.CanStack` + `ModItem.CanStack`

This split is error-prone. Every call site that actually transfers items must remember to call **both**, in the right order. When new vanilla code is introduced, it typically only calls `Item.CanStack`, silently bypassing all mod stacking rules. We then have to hunt down each new call site and patch it manually.

This refactor would fix the `CanStack` inconsistency class of bugs by making `Item.CanStack` invoke mod hooks automatically. The separate `OnStack` hook still requires using `ItemLoader.TryStackItems` / `StackItems` instead of raw arithmetic, but at least the stacking *permission* check would be consistent everywhere.

### What does this proposal attempt to solve or improve?

The current architecture requires ongoing maintenance: whenever vanilla Terraria adds a new item-stacking code path, tML maintainers must notice it and manually chain `ItemLoader.CanStack` after `Item.CanStack`. This is fragile and has already caused regressions in 1.4.5.

If mod hooks were integrated into `Item.CanStack` itself, new vanilla stacking code would automatically respect `GlobalItem.CanStack` / `ModItem.CanStack` without any patches.

### Evidence: 1.4.5 regressions caused by this split

During the 1.4.5 port, Re-Logic introduced a new `QuickStacking` class for the quick-stack-to-nearby-chests feature. It calls `Item.CanStack(item2, item)` to decide whether inventory items can stack into chest slots:

```cs
// QuickStacking.cs — vanilla 1.4.5
if (Item.CanStack(item2, item) && item.stack < item.maxStack) {
    FillStack(item2, dest, i, source.position);
}
```

`FillStack` then performs the transfer with raw `stack +=` / `stack -=`, and `Consolidate` merges inventory stacks before transfer. Neither calls `ItemLoader.CanStack` or `ItemLoader.StackItems`.

With the current architecture, tML had to patch both methods manually:
[...]

PR #5157, opened 2026-05-21
[1.4.5] Merge ItemLoader.CanStack into Item.CanStack, fix stack-related bugs

JerryAZR:
### What is the bug?

In 1.4.5, multiple code paths that transfer or stack items bypass `ItemLoader.CanStack`, breaking modded stacking rules. This was a pre-existing class of bugs in 1.4.4 that became worse after the `Item` → `WorldItem` migration because tML's per-call-site patches were lost or became stale.

Concrete examples a player can observe:
- **Emergency Stacking** (new 1.4.5 feature): Items on the ground can merge despite `CanStackInWorld` returning `false`.
- **ItemSlot swap**: Holding a non-stackable item and clicking a same-type slot does nothing (swap is skipped because `type == type`, but merge is also skipped because `CanStack` was never checked).
- **Crafting UI, wiring, right-click pickup, ammo refill, and more**: All call `Item.CanStack` but the vanilla method does not consult `ItemLoader.CanStack`, so mod hooks are silently ignored.

Additionally, `EmergencyStacking.FindBestTransfers` crashes with an `IndexOutOfRangeException` when modded items exist, because its internal `GroupLookup` cache is sized to `ItemID.Count` but modded item types are `>= ItemID.Count`.

There is also a pre-existing tML patch bug in `Player.ItemSpace` where the type check was accidentally dropped in favor of `ItemLoader.CanStack`, which only checks prefix + mod hooks.

### How did you fix the bug?

Rather than playing whack-a-mole with every individual call site, this PR implements the architecture approved in #5149 : **merge `ItemLoader.CanStack` into `Item.CanStack`** so that the vanilla stackability check automatically respects modded rules.

**Core change:**
- `Item.CanStack(Item, Item)` now calls `ItemLoader.CanStack` after its own type/prefix checks. Every vanilla call site that uses `Item.CanStack` (~20 locations) now implicitly respects `CanStack` hooks with no additional patches.

**Bug fixes discovered along the way:**
- **EmergencyStacking crash**: `GroupLookup` is now sized to `ItemLoader.ItemCount` with a safe `TryGetValue` fallback for modded items. Added `ResetGroupLookup()` cache invalidation on mod reload.
- **EmergencyStacking bypass**: Added `CanStackInWorld` guard in `FindBestTransfers` so world-only stacking rules are respected.
- **ItemSlot swap bypass**: The `SkipSwap` guard now checks `Item.CanStack(...)` so same-type-but-non-stackable items correctly swap instead of doing nothing.
- **ItemSpace ammo check reverted**: Restored `Item.CanStack` (which now includes mod hooks) instead of `ItemLoader.CanStack` alone, fixing the missing type check.

**Cleanups removing now-redundant checks:**
- `WorldItem.TryCombiningIntoNearbyItems`: Removed separate `ItemLoader.CanStack` call (now fully covered by `Item.CanStack`).
- `Player.CanItemSlotAcceptPickup`: Removed separate `ItemLoader.CanStack` call.
- `ChestUI.RefillAmmo`: Removed separate `ItemLoader.CanStack` call.
- `Player.FillAmmo`, `GetItem_FillIntoOccupiedSlot`, `GetItem_FillIntoOccupiedSlot_VoidBag`: Replaced `TryStackItems` with `StackItems` since `Item.CanStack` was already verified.
- `ChestUI.DepositAll_IntoWorldChest`, `DepositAll_IntoLocalChest`, quick stack to chest: Same `TryStackItems` → `StackItems` cleanup.
- `ItemSlot.LeftClick` stacking, `ItemSorting` sort stacking: Same cleanup.
- Removed obsolete `.rej` hunks in `ItemSlot.cs.rej` and `ChestUI.cs.rej` that are now handled by this refactor.

**Testing:**
- Created a test mod (`CanStackDemo`) with an item whose `CanStack` hook blocks stacking between items with different internal state. Verified every affected code path before and after the fix.
- Ran `regen-source --no-prompts` → `dotnet build` → confirmed 0 errors.

The alternative is the status quo: patching each call site individually. We already did this for some paths in earlier PRs, but new bypasses keep appearing (e.g. `EmergencyStacking` in 1.4.5). Centralizing the check in `Item.CanStack` is the only way to prevent future regressions when vanilla adds new stacking code paths.

---

**Porting notes for modders:** None. This is fully backward-compatible. Mods that implement `CanStack`/`CanStackInWorld` will now see those hooks respected in code paths that previously ignored them.


JavidPack:
Thanks. Good job.

PR #5163, opened 2026-05-24
[1.4.5] Fix item flasks

Jadams505:
### What is the bug?
Placing an Item Flask in the world causes the world to not save properly.

### How did you fix the bug?
Updated how the Item Flask saves its data.
Updated the netcode to support proper multiplayer syncing.

I don't think so. Basically made the same changes as Item Frames.


JavidPack:
Perfect, thanks.

PR #5164, opened 2026-05-26
[1.4.5] Updated GlobalNPC.BuffTownNPC

Rijam:
### What is the new feature?

Updated GlobalNPC.BuffTownNPC to allow for changing the attack speed multiplier and the max life of the Town NPC.
`GlobalNPC.BuffTownNPC(ref float damageMult, ref int defense)` -> `GlobalNPC.BuffTownNPC(ref float damageMult, ref float attackSpeedMult, ref int defense, ref int maxLife)`

1.4.5 made it so defeating bosses give +1.5% attack speed, combat books give +20%, and Tipsy gives +10%. The combat books also give +250 max life each.

* There's a special case for the Dryad if in `infectedSeed` world ("Purify this" world) where they are immortal and given 99999 defDefense (as well as changing which direction they walk in? Idk I didn't look too hard at that).
* There's a special case for the Cyborg where their attack speed gets divided by 3.

Perhaps we should pass an `NPC npc` parameter so modders could do something like that if they wanted to.

### Porting Notes:
`GlobalNPC.BuffTownNPC(ref float damageMult, ref int defense)` -> `GlobalNPC.BuffTownNPC(ref float damageMult, ref float attackSpeedMult, ref int defense, ref int maxLife)`. Multiply `attackSpeedMult` by a number or add to `maxLife` if desired.


JavidPack:
I've added the NPC parameter. I don't see why not and it should allow dryad-style effects.

PR #5169, opened 2026-05-29
DebugBuildMod_PublishError

Solxanich:
Publishing workflow now tosses an error if the loaded mod assembly is a debug build.


patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -298,6 +299,13 @@ internal static List<Mod> InstantiateMods(List<LocalMod> modsToLoad, Cancellatio
 
 	public static IEnumerable<Assembly> GetModAssemblies(string name) => GetLoadContext(name).assemblies.Values;
 
+	public static bool IsLoadedModAssemblyDebugBuild(string name)
+	{
+		var assembly = GetModAssemblies(name).Where(asm => asm.FullName.Contains(name)).FirstOrDefault();
```
Mirsario: This is an error-prone way of getting the mod's main assembly, as what if a library is named `ModName_Utils`? 
```suggestion
	public static IEnumerable<Assembly> GetModAssemblies(string name) => GetLoadContext(name).assemblies.Values;

	public static Assembly GetMainModAssembly(string name) => GetLoadContext(name).assembly;

	public static bool IsLoadedModAssemblyDebugBuild(string name)
	{
		var assembly = GetMainModAssembly(name);
		if (assembly == null) return false;
```

patches/tModLoader/Terraria/ModLoader/Core/AssemblyManager.cs
```diff
@@ -298,6 +299,13 @@ internal static List<Mod> InstantiateMods(List<LocalMod> modsToLoad, Cancellatio
...
+
+		return assembly.GetCustomAttributes(false).OfType<DebuggableAttribute>().Any(da => da.IsJITTrackingEnabled);
```
Mirsario: CB's suggestion.
```suggestion
		return assembly.GetCustomAttributes(false).OfType<DebuggableAttribute>().Any(da => da.IsJITOptimizerDisabled);
```

Mirsario (approved):

PR #5175, opened 2026-05-30
[1.4.5] World Item Stack Creation in Shimmer Transform/De-craft Should Respect `maxStack` of Items

JerryAZR:
### What is the bug?
When an item stack is shimmer-transformed into another item type with a lower `maxStack`, the resulting item retains the original stack count, producing an **invalid stack** (`stack > maxStack`).

For example, if a modded item with `maxStack = 9999` shimmer-transforms into an item with `maxStack = 10`, dropping a stack of 500 produces a single pile of 500 — which exceeds the new item's `maxStack`. The pile is still pickable and still contains all 500 items, but it is in an invalid state that may confuse inventory logic, stacking behavior, or mods that assume `stack <= maxStack`.

A related issue exists in the decrafting branch of the same method: vanilla hardcodes a `9999` cap when spawning decraft ingredients. If a recipe ingredient has `maxStack < 9999` (common for modded items), decrafting can similarly produce invalid stacks.

### How did you fix the bug?
**Shimmer transform** (`transformToItem > 0` branch):
1. Cap the current stack to the new item's `maxStack` after `SetDefaults`
2. Spawn additional `WorldItem` instances for the overflow, each capped to `maxStack`
3. Set shimmer state on all spawned items and sync via `NetMessage.SendData(145)`

**Decrafting** (`decraftingRecipeIndex >= 0` branch):
1. Spawn the ingredient `WorldItem` first
2. Read its `maxStack` (which reflects the ingredient's actual limit)
3. Cap the spawned stack to `maxStack` instead of hardcoded `9999`

Both fixes mirror the existing pattern already used by the decrafting branch for spawning items.

None that preserve valid item state. The vanilla behavior produces an invalid stack.

### Porting notes
- Backward-compatible. No modder action required.
- The shimmer transform fix was originally written for 1.4.4's `Item.GetShimmered()` (see `.rej` hunk `@@ -47808` in `Item.cs.rej`). Vanilla 1.4.5 moved the method to `WorldItem.GetShimmered()`, so the fix was adapted to the new location.

### DIFF
--- a/patches/tModLoader/Terraria/Item.cs.rej
+++ b/patches/tModLoader/Terraria/Item.cs.rej
@@ -58,31 +58,6 @@
  			IEnumerable<Item> enumerable = recipe.requiredItem;
  			if (recipe.customShimmerResults != null)
  				enumerable = recipe.customShimmerResults;
-@@ -47869,17 +_,22 @@
- 
- 				num15++;
- 				int num16 = num13 * item2.stack;
-+
-+				/* TML: Recipe.alchemy is removed due to being replaced by consume item hooks
- 				if (recipe.alchemy) {
- 					for (int num17 = num16; num17 > 0; num17--) {
- 						if (Main.rand.Next(3) == 0)
- 							num16--;
- 					}
- 				}
-+				*/
-+
-+				RecipeLoader.ConsumeIngredient(recipe, item2.type, ref num16, true);
- 
- 				while (num16 > 0) {
- 					int num18 = num16;
--					if (num18 > 9999)
--						num18 = 9999;
-+					if (num18 > item2.maxStack)
-+						num18 = item2.maxStack;
- 
- 					num16 -= num18;
- 					int num19 = NewItem(GetItemSource_Misc(8), (int)position.X, (int)position.Y, width, height, item2.type);
 @@ -48004,7 +_,7 @@
  		if (Main.netMode == 1) {
  			int num = (int)(position.X + (float)(width / 2)) / 16;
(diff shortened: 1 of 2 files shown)


Tyfyter:
As of 1.4.5 there should no longer be a lower max stack size for the item's current stack to be above

JerryAZR:
If that's the case, I'll close this PR. However, I'd like to understand what the team is planning to do with maxStack modding support going forward:

1. Is maxStack being deprecated as a moddable field? Should modders stop setting custom maxStack values on their items?
2. Are stacks exceeding maxStack intentionally acceptable now? If a modded item has maxStack = 10 and receives a stack of 500, is that valid state the modder is expected to handle?
3. Should existing code that enforces maxStack be removed or left as-is? There are several places in the codebase that cap transfers to maxStack (e.g., ItemSorting, Player.FillAmmo, various Item.NewItem callers). If the field is no longer meaningful, those checks may be misleading to future contributors.

Just want to make sure I'm not working against the design direction. Thanks!

JavidPack:
I think there is some confusion here. This is just fixing a patch that was moved, right? Tyfyter, what do you mean, why wouldn't that happen anymore?

Tyfyter:
As of 1.4.5 there are no items with different max stack sizes aside from coins and coal to let you stack one item to higher than the max stack of a different item;
While it's not technically impossible for a mod to end up making an item shimmer into one with a lower max stack, that there are only 4 possible result items from vanilla — 3 of which should have custom handling anyway and 1 of which is specifically meant to be an inconvenience — it seemed like a completely pointless increase in maintenance to add support for it, as items with different max stack sizes would be going against vanilla convention, and I apparently missed the mention that it was re-adding an old patch rather than being an entirely new feature.

PR #5177, opened 2026-06-02
Grab Bag Updates

KingXerxes512:
### What is the bug?
Loot drops from grab bags needed updating to 1.4.5 standard drops.  
This includes: Plantera, Duke Fishron, Moonlord, Eye of Cthulu.  
Golden Lock box drops also needed updating to add the Moon Lord Torso and the Ram Rune.  
Huge Dragon Egg needed to be added to the item drop database.  

### How did you fix the bug?
Plantera whip added to boss bag.  
Duke Fishron whip added to boss bag.  
Moon lord whip added to boss bag.  
Eye of cthulu max ore increased from 87 to 90.  
Eye of cthulu unholy arrow max increased from 49 to 50.  
Golden lock box drops now include MoonLordBody and DeadCellsRamRune items.  
Huge Dragon Egg method for registering in the database added. This method just registers the drops for the egg which are the Palworld Chillet and Palworld Chillet Ignis.

I think not.

### Notes
I am not totally sure on the Huge Dragon Egg drop chances being what they are supposed to be. It is supposed to drop one or the other items. Please double check that code in your reviews.


JavidPack:
Fixed some issues, unholy arrows should now drop regardless of world evil, moonlordbody should only drop on for the worthy seeds, and now only 1 of the 2 chillet will drop.

Thanks.

PR #5178, opened 2026-06-02
Fix various boss bag drops and biome crate drops

JavidPack:
Fixes remaining boss bag drops that have changed in 1.4.5

Also fixes biome crate drops that have changed in 1.4.5

Also fixes oyster drop rates.

Finally, this PR also fixes several drops that have been using luck when they shouldn't (All "container" drops in vanilla don't use luck at all). A new `ItemDropRule.ByConditionNotScalingWithLuck` method and `ItemDropWithConditionNotScalingWithLuckRule` class have been added as well to support not consulting luck for rules with a condition.

(done using 1.4.4->1.4.5 source code diff and consulting https://terraria.wiki.gg/wiki/1.4.5.0#Drop_and_Obtainment_Changes)

# TODO
We have a lot of helper methods and classes for drops using luck vs not using luck. These have arisen due to adding features during 1.4.4 development, but in 1.4.5 maybe we can make luck just be a property of each rule. A future PR could remove all extraneous classes/helper methods and just have applying luck be a field in the rule itself, as mentioned in comments below.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/ItemDropDatabase.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/ItemDropDatabase.TML.cs
@@ -265,7 +265,7 @@ partial class ItemDropDatabase
 		};
 		IItemDropRule[] hardmodeThemed = new IItemDropRule[]
 		{
-			ItemDropRule.ByCondition(new Conditions.IsHardmode(), ItemID.Sundial, 200),
+			ItemDropRule.ByConditionNotScalingWithLuck(new Conditions.IsHardmode(), ItemID.Sundial, 200),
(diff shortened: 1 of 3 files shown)


JavidPack:
CB: Is adding a new `ItemDropRule.ByConditionNotScalingWithLuck` method what we want to do? We've added a bunch of "NotScalingWithLuck" methods already, just wondering if you think we should take another approach instead.

Chicken-Bones:
I think comparing the drop and entity source system to many of the other systems (like NPC shops) has shown the downsides of inheritance. 'Scale with luck' should be a composable behaviour (a flag) that can be applied to any rule, just like conditions. Neither should require separate subclasses.

But... that's the job of a rework PR to assess benefits and damages.

PR #5187, opened 2026-06-06
fix: Resolve full solution builds failing on first runs.

Mirsario:
This slightly hacky patch fixes the following type of compilation error that can be encountered by simply wiping the destination folder and running `dotnet build tModLoader.sln`:
image
image

### Changes
- The repo's `/tModLoader.targets` file is updated to evaluate `tMLMod.targets` directly from `src/.../release_targets`, removing the dependency cycle between evaluations and builds. Paths still point to the destination, but that should be fine as they are not fully acted upon before builds. In this process, `tMLMod.targets`'s paths are modified to only be set if they are missing.
- The solution is modified to explicitly build EM after TML. Adding references at a later stage via the means of targets did not work out.

### Alternative approaches
- Since `/tModLoader.targets` already adds project references for TML, ReLogic and FNA, that trend might be able to be continued with it changed to copy all of TML's original references to the child projects that use it. But that would be even more hacky.

### DIFF
--- a/patches/tModLoader/Terraria/release_extras/tMLMod.targets
+++ b/patches/tModLoader/Terraria/release_extras/tMLMod.targets
@@ -22,9 +22,9 @@
 
 	<PropertyGroup>
 		<!-- Legacy property name for tMLSteamPath -->
-		<TerrariaSteamPath>$(MSBuildThisFileDirectory)</TerrariaSteamPath>
-		<tMLSteamPath>$(MSBuildThisFileDirectory)</tMLSteamPath>
-		<tMLLibraryPath>$(tMLSteamPath)Libraries</tMLLibraryPath>
+		<TerrariaSteamPath Condition="'$(TerrariaSteamPath)' == ''">$(MSBuildThisFileDirectory)</TerrariaSteamPath>
+		<tMLSteamPath Condition="'$(tMLSteamPath)' == ''">$(MSBuildThisFileDirectory)</tMLSteamPath>
+		<tMLLibraryPath Condition="'$(tMLLibraryPath)' == ''">$(tMLSteamPath)Libraries</tMLLibraryPath>
 		<tMLName>tModLoader</tMLName>
 		<tMLPath>$(tMLName).dll</tMLPath>
 		<tMLServerPath>$(tMLPath) -server</tMLServerPath>
--- a/patches/tModLoader/Terraria/release_extras/tMLMod.targets
+++ b/patches/tModLoader/Terraria/release_extras/tMLMod.targets
@@ -21,10 +21,12 @@
 	</PropertyGroup>
 
 	<PropertyGroup>
-		<!-- Legacy property name for tMLSteamPath -->
-		<TerrariaSteamPath Condition="'$(TerrariaSteamPath)' == ''">$(MSBuildThisFileDirectory)</TerrariaSteamPath>
-		<tMLSteamPath Condition="'$(tMLSteamPath)' == ''">$(MSBuildThisFileDirectory)</tMLSteamPath>
-		<tMLLibraryPath Condition="'$(tMLLibraryPath)' == ''">$(tMLSteamPath)Libraries</tMLLibraryPath>
+		<!-- GetDirectoryName is used to get a path without a trailing slash. -->
+		<TerrariaSteamPath Condition="'$(TerrariaSteamPath)' == ''">$([System.IO.Path]::GetDirectoryName('$(MSBuildThisFileDirectory)'))</TerrariaSteamPath>
+		<tModLoaderSteamPath Condition="'$(tModLoaderSteamPath)' == ''">$([System.IO.Path]::GetDirectoryName('$(MSBuildThisFileDirectory)'))</tModLoaderSteamPath>
+		<tMLLibraryPath>$(tModLoaderSteamPath)/Libraries</tMLLibraryPath>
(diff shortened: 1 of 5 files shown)


tModLoader.targets
```diff
@@ -1,8 +1,14 @@
+<!-- This is a special targets file that is only to be used by TML's own project files, such as ExampleMod. -->
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <!-- special targets file for ExampleMod. Fits in place of the one in /Mod Sources -->  
+  <!-- Set game paths from WorkspaceInfo.targets and avoid having them overwritten by tMLMod.targets. -->
   <Import Project="src/WorkspaceInfo.targets" />
-  <Import Project="$(tModLoaderSteamPath)/tMLMod.targets" />
-  
+  <PropertyGroup>
+		<tMLSteamPath>$(tModLoaderSteamPath)/</tMLSteamPath>
+		<tMLLibraryPath>$(tMLSteamPath)Libraries</tMLLibraryPath>
```
Chicken-Bones: If you remove ` Condition="'$(tMLLibraryPath)' == ''` from `tMLMod.targets` then you don't need this line.

tModLoader.targets
```diff
@@ -1,8 +1,14 @@
+<!-- This is a special targets file that is only to be used by TML's own project files, such as ExampleMod. -->
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <!-- special targets file for ExampleMod. Fits in place of the one in /Mod Sources -->
```
Chicken-Bones, line 2: Keep this comment, but extend it

tModLoader.targets
```diff
@@ -1,8 +1,14 @@
...
-  <!-- special targets file for ExampleMod. Fits in place of the one in /Mod Sources -->  
+  <!-- Set game paths from WorkspaceInfo.targets and avoid having them overwritten by tMLMod.targets. -->
   <Import Project="src/WorkspaceInfo.targets" />
-  <Import Project="$(tModLoaderSteamPath)/tMLMod.targets" />
-  
+  <PropertyGroup>
+		<tMLSteamPath>$(tModLoaderSteamPath)/</tMLSteamPath>
+		<tMLLibraryPath>$(tMLSteamPath)Libraries</tMLLibraryPath>
+  </PropertyGroup>  
+  <!-- Import targets from src, do not use the destination folder in order to not create an eval-build dependency cycle. -->
```
Chicken-Bones: Comment not really required, this is just straight up sensible

patches/tModLoader/Terraria/release_extras/tMLMod.targets
```diff
@@ -22,9 +22,9 @@
 
 	<PropertyGroup>
 		<!-- Legacy property name for tMLSteamPath -->
-		<TerrariaSteamPath>$(MSBuildThisFileDirectory)</TerrariaSteamPath>
-		<tMLSteamPath>$(MSBuildThisFileDirectory)</tMLSteamPath>
-		<tMLLibraryPath>$(tMLSteamPath)Libraries</tMLLibraryPath>
+		<TerrariaSteamPath Condition="'$(TerrariaSteamPath)' == ''">$(MSBuildThisFileDirectory)</TerrariaSteamPath>
+		<tMLSteamPath Condition="'$(tMLSteamPath)' == ''">$(MSBuildThisFileDirectory)</tMLSteamPath>
```
Chicken-Bones: If only this had the same name as the property in `WorkspaceInfo.targets`. Can we change one or the other?

tModLoader.targets
Chicken-Bones, line 13: Did we not ever need `$(MSBuildThisFileDirectory)`? Or do paths in `ProjectReference  Include=` somehow resolve differently to ones in `Import Project=`

tModLoader.sln
```diff
@@ -10,6 +10,9 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FNA.Core", "FNA\FNA.Core.csproj", "{1EED6B32-5FC7-41A8-91C3-08147C46DD4F}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ExampleMod", "ExampleMod\ExampleMod.csproj", "{F646F0CE-2C51-4610-B097-F16FB58CECC5}"
+	ProjectSection(ProjectDependencies) = postProject
+		{4EFE2B31-D61A-4C07-BAD2-991D3633C237} = {4EFE2B31-D61A-4C07-BAD2-991D3633C237}
+	EndProjectSection
```
Chicken-Bones, line 15: I don't understand why the `ProjectReference` wouldn't set build order for this

Chicken-Bones (changes requested):
Most of the changes seem good. I'd just like to do some more diagnostics to make sure we get it right this time

tModLoader.sln
Mirsario: ProjectReferences set build order, but do not delay expansion of "**/*.dll"-esque wildcards in Reference inclusions. Solution-level dependencies do.
> Adding references at a later stage via the means of targets did not work out.

tModLoader.targets
Mirsario: Seems like it's different, MSBuild is a mess: `warning [MSB9008](https://go.microsoft.com/fwlink/?LinkId=MSBuild.CommonSdk.NonExistentProjectReference): The referenced project FNA/FNA.Core.csproj does not exist`

PR #5193, opened 2026-06-09
[1.4.5] tModPorter: remove plr parameter from Player.GetItem

bwendell:
### What is the bug?

`Player.GetItem` changed signature in the 1.4.5 port: the leading `int plr` parameter was removed. It went from

```cs
Item GetItem(int plr, Item newItem, GetItemSettings settings)   // 1.4.4
```

to

```cs
Item GetItem(Item newItem, GetItemSettings settings)            // 1.4.5
Item GetItem(WorldItem newItem, GetItemSettings settings)       // 1.4.5
```

Mod code calling the old 3-argument form doesn't compile against 1.4.5, and tModPorter didn't rewrite it automatically. This is the open `# tModPorter TODOs` item in `PortingNotes_1.4.5.md`: "Player.GetItem no longer has plr parameter".

### How did you fix the bug?

Added a tModPorter rule mirroring the existing `Item.SetDefaults(noMatCheck)` removal:

```cs
RefactorInstanceMethodCall("Terraria.Player", "GetItem", RemoveParameter(0, "plr", "int"));
```

This drops the first argument from `Player.GetItem(...)` calls when it's an `int` (or a named `plr:` argument), so `player.GetItem(plr, item, settings)` becomes `player.GetItem(item, settings)`. The rule is scoped to `Terraria.Player.GetItem` and only removes an `int`/`plr:` first argument, so already-updated calls (whose first argument is an `Item`/`WorldItem`) are left untouched, and it's idempotent.

Also:
- Added test coverage to `RemoveParameterTest.cs` / `.Expected.cs` (a positional literal `int`, a non-literal `int` `Main.myPlayer`, and a named `plr:` argument).
- Documented the change for modders in `MigrationGuide_1.4.5.md` (🤖).
- Checked the item off the `# tModPorter TODOs` list in `PortingNotes_1.4.5.md`.

Verified locally with `dotnet test tModPorter/tModPorter.Tests` — the `RemoveParameter` transform test and the `ExpectedModCompiles` test both pass (the expected output compiles against the 1.4.5 API).

- Match the parameter purely by name (`plr:`) rather than also by positional `int` index — but that would miss the common positional call style, so matching both (as the existing `SetDefaults` rule does) is better.
- Leave it to modders to fix manually — but tModPorter exists specifically to automate these mechanical signature changes.

#### Note
- I used Claude to help with this PR.

### DIFF
--- a/MigrationGuide_1.4.5.md
+++ b/MigrationGuide_1.4.5.md
@@ -424,6 +424,7 @@ All classes are in the `Terraria` or `Terraria.ID` namespaces unless otherwise i
 * 🤖: `Main.ShouldShowInvisibleWalls` -> `Main.ShouldShowInvisibleBlocksAndWalls`
 * 🤖: `NPC.ShouldBestiaryGirlBeLycantrope` now static.
 * ⚙️: `NPC.SpawnWithHigherTime` removed. No longer used.
+* 🤖: `Player.GetItem` changed. The `plr` parameter has been removed.
 * 🤖: `Player.IsProjectileInteractibleAndInInteractionRange` -> `Player.IsProjectileInteractableAndInInteractionRange`
 * ⚙️: `Player.CheckForGoodTeleportationSpot` removed. Use `Utils.CheckForGoodTeleportationSpot` instead.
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
Excellent, thanks

Issue #4042, opened 2024-01-23
Multilure and Sonar buff cause item-pickup-texts to be increased by 1


BarzenX:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

When using multiple fishing lines all works fine until I apply the Sonar buff. After that, the item-pickup-texts from the fished loot get increased by 1 for the "extra spawned" boppers.

### Log File

not needed

### Steps to reproduce

1. Get the MultiLure mod from Adbluescarab (https://steamcommunity.com/sharedfiles/filedetails/?id=2942015919)
2. Configure one of the mods lures to have more than 1 fishing line (I took 10 as example)
3. Create a fresh world and enter with a fresh player
4. Go to a fishing spot (f.ex. the ocean) and equip/buff to have a sufficiently high fishing power to have most of the boppers splashing
5. Reel in the loot with and without the sonar buff
6. Compare the item-pickup-texts item-amount with the actual added amount in the inventory

### Expected Behavior

The pickup texts should always match the added items to the inventory

### Actual Behavior

When the sonar buff is active some pickup-texts state an increase by 1. I assume the original bopper is the one with the correct amount.
Screenshot: https://imgur.com/yntQBIX

### Reproduction frequency

Always

### Additional Information

nothing

BarzenX:
I'm really honored to see that my reported issue really pointed out to something in the tModLoader and that I could take part in improving it 😍
You, Sir, are awesome! 🥰

JavidPack:
Thanks for the report. Turns out there was another bug with the sonar texts not despawning immediately. They should despawn right as the item is given to the player.

Fixed behavior:

https://github.com/tModLoader/tModLoader/assets/4522492/50845be0-553a-41d5-986b-6ac269c295ce

Technically Terraria doesn't support multiple lures, but since it's so common in mods, I went ahead and made a fix for this in tModLoader itself. This fix does technically have the potential to conflict with something some mod will be doing, so we'll make sure it is tested during February for March release.

There is a simpler fix that would only fix the popuptext stack size (and not the despawning), if this fix causes issues with mods during the preview period we can revert to this change in `PopupText.NewText`:

```diff
	if (!popupText.active || popupText.notActuallyAnItem || (!(popupText.name == newItem.AffixName()) && (!flag || !popupText.coinText)) || popupText.NoStack || noStack)
		continue;

+	if (popupText.sonar)
[...]

PR #5194, opened 2026-06-09
[1.4.5] ItemID.Sets.IsQuestFish and Fishing Fixes

Rijam:
### What is the new feature?

Removed `ModItem.IsQuestFish` in favor of using the vanilla `ItemID.Sets.IsQuestFish`.

It was redundant to have two things defining if something was a quest fish. It was also a noob trap because `ItemID.Sets.IsQuestFish` did nothing before, so somebody could've set `ItemID.Sets.IsQuestFish` and not `ModItem.IsQuestFish` and then they'd be confused on why their quest fish wasn't working.

### What is the bug?

Fishing did not let you catch anything because there was an index out of bounds error when setting the sonar text.

### How did you fix the bug?

#4042 (Commit [247c77b](https://github.com/tModLoader/tModLoader/commit/247c77b853aa299c39328d3484ab9e072d4cc1d1)) added a fix for the sonar text and supported multi-bobber fishing poles. It used `Projectile.localAI[2]` to store the index of the popup text. Well, in 1.4.5, `Projectile.localAI[2]` is now used to store the bait item ID. I switched the sonar popup text to use `Projectile.ai[2]` which was not used by bobbers.

* I had to move the part that adds ModItems to `Main.anglerQuestItemNetIDs` from `ResizeArrays` to `FinishSetup` because `ResizeArrays` is too early for `ModItem.SetStaticDefaults` to run. That meant `ItemID.Sets.IsQuestFish` for the ModItem was always false. Maybe it could go somewhere else?
* I was going to include the AFishDropRulePopulator stuff in this PR, but it is way more complicated and needs a lot of changes for modded item support, so that'll be another PR.

ExampleQuestFish (and the new ExampleBasicFish)

### ExampleMod updates

* Added ExampleBasicFish
  * A basic fish that can be caught in the Example Surface Biome.
  * Shows off the new `ItemID.Sets.IsBasicFish`.
* ExampleQuestFish updated to use `ItemID.Sets.IsQuestFish` and `ModItem.IsQuestFish` was removed.

### Porting Notes
Remove `ModItem.IsQuestFish` from your quest fish and use `ItemID.Sets.IsQuestFish` in SetStaticDefaults instead.


JavidPack:
Looking good. 

As you hinted at, fishing API is still not up to date yet, the visuals and drop logic still need updating. I'm curious if there is any need for CatchFish to exist, hopefully we want all mods to be using the drop rules, but maybe some mod needs to act after the drop rules have been consulted.

Issue #1661, opened 2021-08-01
Game crashes at end of settling water step when entering a world with terraria set to high monitor resolution (i.e. 3840x1600p)


Ph-enix:
### Description

When I try to enter into a world that has already been generated at a higher resolution (the resolution of my monitor is 3840x1600p), the game crashes to desktop at the end of the 'settling water' step.

### Log file

[client.log](https://github.com/tModLoader/tModLoader/files/6913392/client.log)

### Steps to reproduce

1. Set Terraria resolution to a large resolution (I am at 3840x1600p using the in-game video settings, I saw another report for 5120x1440p). Windowed or Fullscreen.
2. Try to enter into a world with any character.

### Expected behavior 

The world should load and the game can be played at the resolution of the monitor.

### Actual behavior
Game closes and will shutdown at the 'Settling water 99%/100%' step.

### Reproduction frequency
Every time I try to enter a world at the resolution of my monitor.

### Additional Information
The crash happens consistently at the resolution of my monitor when full screen or windowed. If I switch to windowed using the game settings and go to a lower resolution (i.e. the default one) the world can be loaded into and the game played as expected.

Edit: Changing the tmodloader setting "Remove Forced Minimum Zoom" to 'Yes' instead of 'No' fixes this crash and the world loads as expected at any resolution fullscreen or windowed.

coleboggus:
Does this still occur if you disable minimum zoom under tmodloader settings? I was having a similar issue and this resolved it for me.

Ph-enix:
> 
> 
> Does this still occur if you disable minimum zoom under tmodloader settings? I was having a similar issue and this resolved it for me.

That's a good call - when I toggle that setting "Remove Forced Minimum Zoom" under the tmodloader settings to 'Yes' instead of the default 'No' then the world does load as expected. The GUI will initially be tiny, but using the in game slider you can increase the scale manually again and it'll look as intended. 

Thank you.

freezebladeJD:
I'm having the same issue when running 4k, except when I load up a world, it finishes loading completely but upon entering the world it crashes. I've tried full screen, windowed, and borderless but nothing affects it.

Solxanich:
[client (1).log](https://github.com/tModLoader/tModLoader/files/6986953/client.1.log)
Just tossing up another log here from someone else with different GPU manufacturer. 

[...]

PR #5202, opened 2026-06-15
feat: shared lighting buffer acquisition api

steviegt6:
### What is the new feature?

Introduces a new API for acquiring the current lighting buffer as a texture.

You may request the lighting buffer with `Lighting.GetBufferTexture()`, which returns a `LightMapBuffer` object. This object has the buffer `Texture` as well as the tile bounds defined as `TileArea`. Read the documentation for specifics on what the properties mean. A buffer texture pixel corresponds to a single tile in the world.

Also tweaks LegacyLighting to copy its lighting state to a LightMap.

Many mods that work with lighting need this same general data, but collecting it is expensive; this reduces redundancy across mods. It also implements it in a manner that is considerably more efficient than any mod could implement it, by making fundamental changes to LightMap that are not possible with MonoMod.

Nitrate uses a lighting buffer to apply lighting to entities. So does my 1.4.5 rendering mod. Starlight River uses it for lighting up backgrounds. This is generally useful for applying light to large meshes.

Nothing substantial. The only other approach is manually copying data with GetColor calls. The advantage of this is it doesn't require modifying any of the lighting engines or LightMap, but it's also considerably slower.

This short mod draws the buffer both as a small image in the corner of the screen and as a full image overlayed on the corresponding tiles.

```cs
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;
using Microsoft.Xna.Framework.Input;
using Terraria;
using Terraria.ModLoader;
using Terraria.UI;

namespace LightBufferTest;

public class LightBufferTest : Mod;

internal sealed class System : ModSystem
{
    public enum Mode
    {
        Disabled,
        SmallWindow,
        Fullscreen,
        Count
    }

    private Mode mode;

    public void CycleMode()
    {
        mode = (Mode)(((int)mode + 1) % (int)Mode.Count);
    }

    public override void ModifyInterfaceLayers(List<GameInterfaceLayer> layers)
    {
        base.ModifyInterfaceLayers(layers);

        layers.Add(new LegacyGameInterfaceLayer("SmallWindow", () =>
        {
            if (mode != Mode.SmallWindow)
            {
                return true;
            }

            var buffer = Lighting.GetBufferTexture();
            Main.spriteBatch.Draw(buffer.Texture, new Vector2(256f), new Rectangle(0, 0, buffer.TileArea.Width, buffer.TileArea.Height), Color.White);
            return true;
        }, InterfaceScaleType.UI));

        layers.Add(new LegacyGameInterfaceLayer("Fullscreen", () =>
        {
            if (mode != Mode.Fullscreen)
            {
                return true;
            }

            Main.spriteBatch.End();
            Main.spriteBatch.Begin(SpriteSortMode.Deferred, null, SamplerState.PointClamp, null, null, null, Main.GameViewMatrix.ZoomMatrix);

            var buffer = Lighting.GetBufferTexture();

            var area = new Rectangle(0, 0, buffer.TileArea.Width, buffer.TileArea.Height);
            var rect = buffer.TileArea;
            rect.Width *= 16;
            rect.Height *= 16;
            rect.X *= 16;
            rect.Y *= 16;
            rect.X -= (int)Main.screenPosition.X;
            rect.Y -= (int)Main.screenPosition.Y;
            Main.spriteBatch.Draw(buffer.Texture, rect, area, Color.White * 0.5f);

            Main.spriteBatch.End();
            Main.spriteBatch.Begin(SpriteSortMode.Deferred, null, null, null, null, null, Main.GameViewMatrix.ZoomMatrix);
            return true;
        }, InterfaceScaleType.Game));
    }

    public override void PostUpdateInput()
    {
        base.PostUpdateInput();

        if (Main.keyState.IsKeyDown(Keys.F6) && !Main.oldKeyState.IsKeyDown(Keys.F6))
        {
            CycleMode();
            Main.NewText("Cycled to mode: " + mode.ToString());
        }
    }
}
```

### ExampleMod updates

Not really useful for anything in EM, but a sample mod is attached above.

### Porting Notes

Nothing is *necessary* to port, but if you already had a system like this, you should use this.

A screenspace version of the buffer is not provided.

If you implement `ILightingEngine` yourself, you will need to implement `GetBufferTexture`.

### DIFF
--- a/patches/tModLoader/Terraria/Graphics/Light/LegacyLighting.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Light/LegacyLighting.cs.patch
@@ -24,6 +24,16 @@
  		if (x < _expandedRectLeft || x >= _expandedRectRight || y < _expandedRectTop || y >= _expandedRectBottom)
  			return Vector3.Zero;
  
+@@ -276,6 +_,9 @@
+ 			PreRenderPhase();
+ 
+ 		_lastCameraPosition = Main.Camera.UnscaledPosition;
++
++		// TML: Mark light map as dirty to prompt buffer regenration.
++		_lightMap.MarkDirty();
+ 	}
+ 
+ 	private void TryUpdatingMapWithLight()
 @@ -581,6 +_,7 @@
  		num3 = Utils.Clamp(_expandedRectTop, 5, Main.maxTilesY - 1);
  		num4 = Utils.Clamp(_expandedRectBottom, 5, Main.maxTilesY - 1);
--- a/patches/tModLoader/Terraria/Graphics/Light/LegacyLighting.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Light/LegacyLighting.cs.patch
@@ -14,7 +14,7 @@
  {
  	public struct RectArea
  	{
-@@ -133,6 +_,21 @@
+@@ -133,6 +_,11 @@
  	private static FastRandom _swipeRandom = FastRandom.CreateWithRandomSeed();
  	private LightMap _lightMap = new LightMap();
  
@@ -22,16 +22,6 @@
 +	// lighting and mask data.  This new light map contains a copy of the data
 +	// in _states so it can be used with the LightMapBuffer API.
 +	private LightMap exposedLightMap = new();
-+	private int cachedExpandedLeft;
-+	private int cachedExpandedTop;
-+
-+	private void UpdateLightMap(int x, int y, Vector3 color)
-+	{
-+		// x += (_expandedRectLeft - cachedExpandedLeft);
-+		// y += (_expandedRectTop - cachedExpandedTop);
-+
-+		exposedLightMap[x, y] = color;
-+	}
 +
  	public int Mode { get; set; }
  
@@ -57,14 +47,13 @@
  		if (_states != null && _states.Length >= num && _states[0].Length >= num2)
  			return;
  
-@@ -276,6 +_,10 @@
+@@ -276,6 +_,9 @@
  			PreRenderPhase();
  
  		_lastCameraPosition = Main.Camera.UnscaledPosition;
 +
 +		// TML: Mark light map as dirty.
 +		exposedLightMap.MarkDirty();
-+		Main.NewText($"{_expandedRectLeft - cachedExpandedLeft} {_expandedRectTop - cachedExpandedTop}");
  	}
  
  	private void TryUpdatingMapWithLight()
@@ -74,7 +63,7 @@
  				}
 +
 +				// TML: LightMapBuffer state maintenance.
-+				UpdateLightMap(i - _expandedRectLeft, j - _expandedRectTop, lightingState.ToVector3());
++				exposedLightMap[i - _expandedRectLeft, j - _expandedRectTop] = lightingState.ToVector3();
  			}
  		}
  	}
@@ -84,7 +73,7 @@
  				obj.B = lightingState.B;
 +
 +				// TML: LightMapBuffer state maintenace.
-+				UpdateLightMap(i, j, obj.ToVector3());
++				exposedLightMap[i, j] = obj.ToVector3();
  			}
  		}
  	}
@@ -94,7 +83,7 @@
  					obj.B = lightingState.B2;
 +
 +					// TML: LightMapBuffer state maintenance.
-+					UpdateLightMap(i, j, obj.ToVector3());
++					exposedLightMap[i, j] = obj.ToVector3();
  				}
  			}
  
@@ -104,7 +93,7 @@
  				obj2.B = lightingState2.R2;
 +
 +				// TML: LightMapBuffer state maintenance.
-+				UpdateLightMap(k, l, obj2.ToVector3());
++				exposedLightMap[k, l] = obj2.ToVector3();
  			}
  		}
  	}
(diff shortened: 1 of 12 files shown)


patches/tModLoader/Terraria/Graphics/Light/LightMap.cs.patch
```diff
@@ -1,6 +1,100 @@
...
+@@ -4,9 +_,15 @@
+ 
+ namespace Terraria.Graphics.Light;
+ 
+-public class LightMap
++// TML: Switched backing color store from Vector3[] to Vector4[], changed store
++// from column-major to row-major; see comments below for explanations.
++public partial class LightMap
+ {
++	// TML: Colors are now stored as a Vector4 array instead of a Vector4 array
```
oli-x64:
```suggestion
+	// TML: Colors are now stored as a Vector4 array instead of a Vector3 array
```

oli-x64:
I support this feature as it would be valuable to optimisation mods such as Nitrate, as well as those that do custom background rendering such as Starlight River

LGTM

Ozzatron (approved):
My only concern would be enforcing the main thread safety on the method which retrieves the lighting buffer.

steviegt6:
There would be two ways to handle Main Thread safety. You could either use RunOnMainThread or throw a more explicit exception, but everything in Lighting is already not-thread safe, so I'm considering it relatively unimportant in practice. Especially since it's well established that graphics resource acquisition on secondary threads is illegal in tMod.

Tyfyter:
Since virtually any use of the buffer has to be on the main thread, a lack of thread safety seems a given;
I'm not experienced enough with lower-level graphics code to provide input on the implementation, but the API looks good to me.

patches/tModLoader/Terraria/Graphics/Light/LightMapBuffer.cs
```diff
@@ -0,0 +1,33 @@
...
+	///		indicate the top-left position of the buffer in tile coordinates,
+	///		and the Width and Height indicate the full area being procesesed.
+	///		<para />
+	///		The light map buffer is subject to two possible kinds of padding:
+	///		1) padding in the allocated color/mask buffers, and 2) padding with
+	///		valid data, but is exempt from further processing (i.e. blurring in
+	///		<see cref="LightingEngine"/>).
+	///		<br />
+	///		Width and Height will exclude the former but, include the latter.
+	/// </summary>
```
Chicken-Bones, line 31: How are the padding notes here relevant to a caller?

patches/tModLoader/Terraria/Graphics/Light/LightMap.TML.cs
```diff
@@ -0,0 +1,86 @@
...
+	//
+	// The reason I don't currently do this is it relies on the memory layout
+	// of Vector3 and Vector4 to be Explicit or Sequential, which it currently
+	// isn't.  (In practice, this would probably work fine on 64-bit machines).
+	//
+	// As it currently stands, I don't notice a massive performance impact, so
+	// the JIT likely handles these cases quite well already.
+
+	[MethodImpl(MethodImplOptions.AggressiveInlining)]
+	private static Vector3 ToVector3(in Vector4 value)
```
Chicken-Bones, line 76: you could set a breakpoint when running in release and check the disassembly

patches/tModLoader/Terraria/Graphics/Light/LightMap.cs.patch
Chicken-Bones, line 109: This was added in response to https://github.com/tModLoader/tModLoader/issues/1661 but I think the solution is super dubious -.-

Any chance you can validate the original 'error'?

patches/tModLoader/Terraria/Graphics/Light/LightMap.cs.patch
```diff
@@ -1,6 +1,104 @@
...
+ 		FastParallel.For(0, Width, delegate (int start, int end, object context) {
+ 			for (int j = start; j < end; j++) {
+ 				BlurLine(IndexOf(j, 0), IndexOf(j, Height - 1 - NonVisiblePadding), 1);
+@@ -94,6 +_,26 @@
+ 				BlurLine(IndexOf(Width - 1, i), IndexOf(NonVisiblePadding, i), -Height);
+ 			}
+ 		});
++		*/
++
++		// TML: Because we switch from column-major to row-major (see IndexOf
```
Chicken-Bones, line 72: Accepted, hopefully the performance cost is minimal. Our tile rendering is column major :(

patches/tModLoader/Terraria/Graphics/Light/LegacyLighting.cs.patch
```diff
@@ -1,13 +1,31 @@
...
+ 	public struct RectArea
+ 	{
+@@ -133,6 +_,11 @@
+ 	private static FastRandom _swipeRandom = FastRandom.CreateWithRandomSeed();
+ 	private LightMap _lightMap = new LightMap();
+ 
++	// TML: _lightMap is used by vanilla to get TileScanner data for basic
++	// lighting and mask data.  This new light map contains a copy of the data
++	// in _states so it can be used with the LightMapBuffer API.
++	private LightMap exposedLightMap = new();
```
Chicken-Bones, line 24: THis comment doesn't _quite_ explain why you can't use `_lightMap`. Also lets call it `_bufferTextureLightMap` if we do need it

Chicken-Bones (changes requested):

PR #5203, opened 2026-06-16
fix: correct library arch paths for Linux

Zyian:
### What is the bug?
With the change to update libraries in https://github.com/tModLoader/tModLoader/commit/e7e0719f5c329353455c8ca84cd13d5c8b48330f

Linux directory
```
Libraries/Native/Linux
❯ readelf -h libSDL3.so.0
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           AArch64
  Version:                           0x1
  Entry point address:               0x443c0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          3988696 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         7
  Size of section headers:           64 (bytes)
  Number of section headers:         30
  Section header string table index: 29
```

Linux-arm64 directory
```
Libraries/Native/Linux-arm64
❯ readelf -h libSDL3.so.0
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x39f00
  Start of program headers:          64 (bytes into file)
  Start of section headers:          4080784 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         9
  Size of section headers:           64 (bytes)
  Number of section headers:         31
  Section header string table index: 30
```

The paths for X64 Linux libraries and ARM64 were swapped making it unable to launch

### How did you fix the bug?
Swapped the directories, nothing else was changed


Chicken-Bones:
Thanks and my apologies. I've swapped just the files from that recent commit

PR #5206, opened 2026-06-18
[1.4.5] Add perspective of player to tile sight hooks

noahDP1911:
### What is the bug?
`IsTileSpelunkable` and `IsTileBiomeSightable` do not receive the player whose perspective is being rendered. Mods must rely on `Main.LocalPlayer`, producing incorrect results when rendering another player's perspective, such as while spectating.

### How did you fix the bug?
Added a `Player` parameter to the corresponding `ModTile` and `GlobalTile` hooks and passed `_perspectivePlayer` through `TileDrawing`, `Main`, and `TileLoader`.
Existing public `Main` overloads remain available and default to `Main.LocalPlayer`. Hook documentation, migration guidance, and tModPorter rules were also updated.

Hooks could read `Main.SceneMetrics.PerspectivePlayer` directly, but an explicit parameter avoids hidden global state and matches the existing `IsTileDangerous` design.

Updated `ExampleHerb.IsTileSpelunkable` and `ExampleOre.IsTileBiomeSightable` to use the new signatures.

The migration guide documents the signature changes and advises using the supplied player instead of `Main.LocalPlayer`. tModPorter updates existing overrides and emits a migration suggestion.

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleHerb.cs
+++ b/ExampleMod/Content/Tiles/ExampleHerb.cs
@@ -153,7 +153,7 @@ namespace ExampleMod.Content.Tiles
 			}
 		}
 
-		public override bool IsTileSpelunkable(int i, int j) {
+		public override bool IsTileSpelunkable(int i, int j, Player player) {
 			PlantStage stage = GetStage(i, j);
 
 			// Only glow if the herb is grown
--- a/ExampleMod/Content/Tiles/ExampleOre.cs
+++ b/ExampleMod/Content/Tiles/ExampleOre.cs
@@ -35,7 +35,7 @@ namespace ExampleMod.Content.Tiles
 		}
 
 		// Example of how to enable the Biome Sight buff to highlight this tile. Biome Sight is technically intended to show "infected" tiles, so this example is purely for demonstration purposes.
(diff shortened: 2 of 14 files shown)


Chicken-Bones:
This is fine, but really needs a motivating example use case. The calls to these hooks are already gated behind whether the perspective player has a spelunker or biome sight potion active, so there's no need for mods to implement player specific checks in their tiles

noahDP1911:
> This is fine, but really needs a motivating example use case. The calls to these hooks are already gated behind whether the perspective player has a spelunker or biome sight potion active, so there's no need for mods to implement player specific checks in their tiles

Hm, you’re right. I looked into it and realized that tile rendering uses the spectated player, so passing the perspective player would be internally consistent with IsTileDangerous.

However, the outer Spelunker/Biome Sight checks already use that player and I couldn’t find an existing mod where the tile classification itself varies by player. A filtered ore scanner or progression-based sight mechanic is possible, but unimplemented :D. Mods needing it can also just read Main.SceneMetrics.PerspectivePlayer.

Given that this changes four public hook signatures, I don’t think API consistency alone provides enough benefit. So unless someone comes in here and demands this feature, closing this is better than imposing migration work for a speculative feature.

PR #5207, opened 2026-06-19
[1.4.5] Ported tile-entity updates

noahDP1911:
### What is the bug?
Terraria 1.4.5 only updates tile entities registered in `TileEntity.UpdateEntities`.
tModLoader directly added and removed modded tile entities from `ByID` and `ByPosition`, bypassing the new registration logic. Consequently, modded tile entities were not registered for updates.

### How did you fix the bug?
Modded tile entities now require updates by default, preserving their previous behavior.
Placement, loading, and removal now use `TileEntity.Add` and `TileEntity.Remove`, keeping `UpdateEntities`, `ByID`, and `ByPosition` synchronized.
Documentation and tests were added for both updating tile entities and tile entities that opt out by setting `RequiresUpdates` to `false`.

Mods could be required to set `RequiresUpdates` themselves, but this would silently break existing tile entities.
tModLoader could maintain a separate update loop, but using Terraria's existing registration methods avoids duplicating update behavior.


Chicken-Bones:
Please remove the tests, they don't add much value relative to their maintenance cost. We will discuss the default for `RequiresUpdates`

Chicken-Bones:
You can initialize the default value of `RequiresUpdates` to `LoaderUtils.HasOverride(this, e => e.Update)`

Chicken-Bones (approved):

PR #5208, opened 2026-06-20
[1.4.5] Display Doll (Mannequin) Settings

Rijam:
### What is the new feature?

* New `bool (ModProjectile|GlobalProjectile).DisplayDollSettings(Player doll, TEDisplayDoll.DisplayDollPose pose, ref int aiStyle, ref int aiType)` hook to allow for changing how a held projectile is drawn when a mannequin is holding it.
  * `doll` is the player instance for the mannequin.
  * `pose` is the current pose of the mannequin.
  * `ref aiStyle` is to use one of the vanilla aiStyles for drawing on the mannequin even if your projectile doesn't necessarily use that vanilla aiStyle. `ref aiType` further specifies a specific projectile type to emulate.
  * Return `false` to stop the rest of the vanilla code from running.
* Made `TEDisplayDoll.RegisterUsePose` public.

* Projectile.AI_DisplayDoll is the method that handles changing how held projectiles look based on their aiStyle. Projectiles that use custom aiStyles were not drawn or were drawn incorrectly.

Vanilla supports the following aiStyles:
* 19 Spear
* 20 Drill
* 161 ShortSword
  * 75 HeldProjectile (This has a bunch of misc held projectiles and uses the same settings as Shortsword)
* 142 ForwardStab (Ghastly Glaive and Sky Dragon's Fury)
  * 140 SleepyOctopod (Same settings as ForwardStab)
* 99 Yoyo
* 15 Flail
* 100 MedusaRay
* 165 Whip

* I have hook be a bool so the modder can return false to stop the vanilla code from running. I did it this way in case your projectile uses a supported aiStyle, but you want to do your own settings instead of the vanilla drawing.
* The name of the hook. We could change it to "MannequinDisplaySettings".
* ~~The `botherDrawing` parameter could be renamed to something like "shouldDrawProjectile".~~ Removed param

* Updated Example Mod content.

### ExampleMod updates
* `ExampleAdvancedFlailProjectile`, `ExampleCustomSwingProjectile`, `ExampleDrillProjectile`, `ExampleFailProjectile`, `ExampleHeldProjectileWeaponProjectile`, `ExampleJoustingLanceProjectile`, `ExampleShortswordProjectile`, `ExampleSpearProjectile`, `ExampleWhipProjectile`, `ExampleWhipProjectileAdvanced`, and `ExampleYoyoProjectile` have been updated to support being held by mannequins.
* `ExampleCustomUseStyleWeapon` (`ExampleCustomUseStyleGlobalItem`) has been updated to support being held by mannequins using `TEDisplayDoll.RegisterUsePose` and `player.isDisplayDollOrInanimate`.

### Porting Notes
* To support mannequins on your held projectiles:
  * If you are using one of the supported vanilla aiStyles, or one that mimics one of the vanilla aiStyles, override DisplayDollSettings, set the aiStyle parameter to the vanilla aiStyle, and return true. You can further set aiType as well if the vanilla logic takes that into account.
  * If one of the vanilla aiStyles doesn't work for your projectile, override DisplayDollSettings and implement custom logic. Return false to stop the vanilla code from running.
  * Projectile PreDraw/Draw/PostDraw code needs to be updated to use the new `player` parameter instead of using `Main.player[Projectile.owner]`.
  * `Projectile.drawLayer = ProjectileDrawLayerID.HeldProj` will likely need to be added to `SetDefaults` for the projectile to draw on the correct layer.
  * If necessary, use `Projectile.isAPreviewDisplayDoll` to check if the projectile is held by a mannequin.
* If you had a custom use style, add `TEDisplayDoll.RegisterUsePose()` in Load.
  * Any code that uses the player's mouse should be changed to something more generic for mannequins.
  * Use `player.isDisplayDollOrInanimate` to check if the player is a mannequin.
  * ModPlayers still work on mannequins, so no need to check for null or anything there.


JavidPack:
I added a `aiStyle` parameter to allow reusing vanilla logic and removed a lot of code that wasn't necessary anymore. I'm still not sure about the parameter and method names.

This is looking good. Maybe we should merge after asking around if the docs and parameter/method names make sense.

Chicken-Bones:
This will need at least minor updates with 1.4.5.7, so I'll leave it to you both to decide whether to merge it now so modders can get started using the feature and revise it with the update, or wait till then

PR #5212, opened 2026-06-21
[1.4.5] Example Kite

max93713:
Adds ExampleKite and ExampleKiteProjectile, demonstrating how to create a basic modded kite, the kite string visuals do match the 1.4.5 changes as well.

I would assume not, as the code is commented throughout to explain the kite and all the drawing logic.


JavidPack:
I've fixed a few things, like supporting the kite anchor, but there are still a few issues that might take more thought or community discussion.

dotnet_2026-06-22_16-03-08

The issue is the custom drawing path used by anchored projectiles. `Main.DrawKite` is used to draw the kite while anchored, the usual projectile drawing hooks are all skipped. This is necessary because `DrawKite` has a parameter for the anchor position. The current code uses `PreDraw` to skip the vanilla drawing code while held by a player. This leads to 2 separate paths for drawing the same projectile. As seen above, this actually has the effect of properly drawing the kite tail, which was missing from the held kite this PR originally implemented. The anchored projectile is not using any of the custom draw code from `ExampleKiteProjectile`, so as an example it would deviate in visuals once a modder customizes it. 

We need some approach to allowing modders to fully customize the drawing without duplicating code. It's possible we could make a hook similar to how `AdjustMultiTileVineParameters` was implemented to customize the drawing customization within `DrawKite`, but we'd want to be able to fully customize the drawing code as well. A lot of the current `ExampleKiteProjectile` code is duplicated from `DrawKite`, which would also be good to avoid. Using the existing `PreDraw` hook would be useful, but there might not be a good way to pass in the anchor position. A `DrawKite` hook that allowed customizing the parameters and optionally skipping the logic altogether might work, but separating it from the normal projectile drawing hooks might be confusing to modders who wouldn't understand why it needs it's own hook.

All this is to say, we probably need to discuss this more in depth later, once more modders are on 1.4.5.

JavidPack:
Also, ALL anchored kites don't seem to be affected by wind and just fall to the ground. I haven't figured out why yet.

- [ ] We also need to add in the normal tooltips that the vanilla kites have. ` "{$CommonItemTooltip.Kite}",`

PR #5215, opened 2026-06-22
fix: Allow SteamAppId env var to be pre-set for Family Share launches

### LINKED ISSUE #5174: Launch bug
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

I had to launch manually because of steam family share but whenever I try to launch it using the family .bat file it gives me and error message(the error message is attached below)

Image

### Log File

_No response_

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

AdvenRanises:
Adds a guard clause in `Steam.SetAppId()` to skip setting the `SteamAppId` environment variable if it already matches the expected app ID. This fixes a crash when launching tModLoader via Steam Family Share .bat files, which pre-set `SteamAppId` before launch.

Fixes #5174.

### DIFF
--- a/patches/tModLoader/Terraria/ModLoader/Engine/Steam.cs
+++ b/patches/tModLoader/Terraria/ModLoader/Engine/Steam.cs
@@ -71,8 +71,12 @@ internal class Steam
 
 		//Solxan: While testing family share, noticed this inherently fails if game is not initiated from the 'start' button in Steam.
 		// Not normally a problem, as family share should be pathing to game server anyways
-		if (Environment.GetEnvironmentVariable("SteamAppId") != appId.ToString()) {
-			throw new Exception("Cannot overwrite steam env. SteamAppId=" + Environment.GetEnvironmentVariable("SteamAppId"));
+		var existingSteamAppId = Environment.GetEnvironmentVariable("SteamAppId");
+		if (existingSteamAppId == appId.ToString()) {
+			return;
+		}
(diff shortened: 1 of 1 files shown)


JavidPack:
@Solxanich Your opinion here would also be appreciated. It seems like maybe the original issue shouldn't be happening at all, since it wasn't an issue before, so maybe this fix is bypassing some check we have come to rely on.

Solxanich:
image

Fundamentally, the code that throws should be unreachable if SteamedWraps.FamilyShared is being set correctly. 
And if its not being set correctly, everything else will fail downstream.

SteamedWraps.FamilyShared is set when TerrariaSteamClient reports SteamApps.BIsSubscribedFromFamilySharing(). 
The TerrariaSteamClient log doesn't show that anywhere so there is some whole lot of other problems happening with that particular user.

Apologies, but reject current commit as written as it does not solve the reported problem of family share not working; needs further information from the originating issue as to what is going on. 

It looks like they are running familyShared.bat when they shouldn't need to and can just run it normally (assuming that the Steamworks.NET function isn't broken)

AdvenRanises:
Closing per maintainer feedback — the guard clause masks a deeper issue with FamilyShared detection. Will follow up on the issue for more logs.

PR #5217, opened 2026-06-22
fix: Prevent startup crash when Steam Workshop is offline

### LINKED ISSUE #5126: Auto-updating mods causes crash in offline mode
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

If a mod releases an update and I open tModloader in Steam's offline mode, the game will crash upon attempting to update the mod(s).

### Log File

[client.log](https://github.com/user-attachments/files/27375034/client.log)

### Steps to reproduce

(Mod must have an update not yet downloaded)
1. Run tModloader offline without having opened it while online prior

### Expected Behavior

tModloader should skip mod update(s) until it gains access to the Steam Workshop.

### Actual Behavior

Game crashes immediately after Re-Logic intro card.

Image

### Reproduction frequency

Always, as long as at least one of my subscribed mods has an update

### Additional Information

Game will run properly after returning online and re-opening tModloader

### LINKED ISSUE #5196: cant start tmodloader offlin
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

GOG

### This bug affects...

Gameplay as a Player

### Description

i have bad internet conection and tml keep crashing after logo so trying to go in tml offline but steel crash with same error 

im sure tml can used offline 
sombody help ore patch this

### Log File

[client.log](https://github.com/user-attachments/files/28911808/client.log)

[environment-client.log](https://github.com/user-attachments/files/28911811/environment-client.log)
[environment-server.log](https://github.com/user-attachments/files/28911812/environment-server.log)
[Launch.log](https://github.com/user-attachments/files/28911813/Launch.log)
[Natives.log](https://github.com/user-attachments/files/28911814/Natives.log)
[server.log](https://github.com/user-attachments/files/28911815/server.log)

### Steps to reproduce

open tml when offline

see re-logic intro 

main engin  crash

### Expected Behavior

should loading mods or at last load terraria

### Actual Behavior

crash after re-logic screen

### Reproduction frequency

always except some time that im have great internet for a sec

### Additional Information

im know  its becous off steam but plaes dont tell me delate steam its a nightmer to reinstall it with my net 

also im bad at english sory

AdvenRanises:
### What is the bug?
When Steam is offline (or GOG without Steam), the workshop query during startup throws `System.Exception: Error: Unable to access Steam Workshop. ERROR CODE: k_EResultFail` from `WaitForQueryResultAsync`. The existing `catch (TimeoutException)` in `WaitForQueryResult` (added in abca3f4) doesn't catch this, so the exception propagates up through:

`DetectAbnormalSteamWorkshopDownloads` → `Initialize` → `OnLocalModsChanged` → `DirectQueryInstalledMDItems` → `QueryItemsSynchronously` → `WaitForQueryResult` → **uncaught crash**

### How did you fix the bug?
Broadened `catch (TimeoutException e)` to `catch (Exception e)` in `WorkshopHelper.TML.cs:WaitForQueryResult`. This applies the existing graceful degradation (set `SteamAvailable = false`, `SteamClient = false`, show message, continue) to **all** workshop query failures, not just timeouts.

Could add explicit offline checks before calling workshop APIs, but that would require touching multiple call sites. Catching the exception at the common `WaitForQueryResult` entry point is the minimal, safest fix that mirrors the existing timeout handling pattern.

### Issues fixed
- Fixes #5196
- Fixes #5126

### DIFF
--- a/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
+++ b/patches/tModLoader/Terraria/Social/Steam/WorkshopHelper.TML.cs
@@ -535,11 +535,11 @@ public partial class WorkshopHelper
 				try {
 					WaitForQueryResultAsync(query).GetAwaiter().GetResult();
 				}
-				catch (TimeoutException e) {
-					// Solxan: Most likely to occur during game startup if it occurs due to banned/malware startup check code
+				catch (Exception e) {
+					// Catch all workshop query failures (offline mode, timeout, k_EResultFail, etc.)
 					SteamedWraps.SteamAvailable = false;
 					SteamedWraps.SteamClient = false;
-					Utils.ShowFancyErrorMessage("Steam Workshop Timed Out. Steam Workshop related functionality has been disabled.\n" +
+					Utils.ShowFancyErrorMessage("Unable to access Steam Workshop. Steam Workshop related functionality has been disabled.\n" +
 						"Restart Steam and tModLoader to attempt to restore.", Interface.loadModsID);
 				}
 			}


JavidPack:
@Solxanich Does this look good? The original `catch TimeoutException` seemed quite intentional when I looked into this issue, so I wasn't sure if expanding this to catch more would break other things.

Solxanich:
I'm presuming that we would clean the history so that we only merge the offline workshop commit in this PR?

So it is intentional to an extent, yes. When I'm patching a bug in real-time with the intent to fast-forward, I try to avoid introducing other issues.

In the event of TimeoutException, we have been unable to reach Steam for 10 seconds. This is reasonably sufficient to consider it dead and turn off all workshop related code. 
Interestingly, when this issue of 'Offline' was initially reported, it was throwing Timeout Exception as Steam is fully dead. 

Whereas EResult_Fail can occur much more frequently. Is usually not a condition that I would associate with turning off all Steam Workshop functionality just because it happened once. 

In practice, for this particular method, the risk on the table of that particular occurrence is probably lower than the benefits of catching all exceptions; meaning if it is broken on Game launch, it is probably not recovering. 

The other areas that use this are multiplayer mod sync and publishing; they both could trigger that fail and lock out features. But again, this isn't during the high-rate speed of the full mod browser so should be safe?  

Needs to be properly tested on preview so no fastforward.

PR #5222, opened 2026-06-25
[1.4.5] ArmorIDs.Wing.Sets.AlwaysAnimated Docs

Rijam:
### What was changed?

Added docs for `ArmorIDs.Wing.Sets.AlwaysAnimated`. It is solely used by the Velociraptor mount to determine if it should use its flying animation while grappling. If the wing is in this set, the Velociraptor will *not* use its flying animation while grappling.

The set contains:
* 22 Hoverboard
* 28 Lazure's Barrier Platorm
* 45 Celestial Starboard
* 34 Jim's Wings
* 48 Chippy's Cloak
* 39 Leinfor's Prehensile Cloak
* 40 Ghostar's Infinity Eight
* 44 Empress Wings

This demonstration shows the Velociraptor grappling with Angel Wings (which is not in the set) and the Hoverboard (which is in the set). With the Angel Wings, the Velociraptor uses its flying animation while grappling and while attached to a block. With the Hoverboard, the Velociraptor uses its walking animation while grappling and uses the falling animation while attached to a block.

VelociraptorAlwaysAnimated

Relevant code is in Mount.UpdateFrame_Velociraptor()


JavidPack:
Thanks.

PR #5225, opened 2026-06-26
Resolve GameModeData porting note

kirrishima:
### What is changed

This addresses the first `PortingNotes_1.4.5.md` TODO:

> GameModeData.cs no longer exists, patches need to be redistributed

I checked for remaining `GameModeData` references and found the stale reference was in the XML docs for `NPCID.Sets.NeedsExpertScaling`.

The patch now updates the cref from:

```csharp
NPC.ScaleStats(int?, GameModeData, float?)
```

to the current 1.4.5 signature:

```csharp
NPC.ScaleStats(int?, float?)
```

The completed TODO was removed from `PortingNotes_1.4.5.md`.

### Verification

- `rg GameModeData .` finds no remaining references.
- `dotnet build` succeeds with 0 errors.
- existing warnings remain unrelated.

### Feedback wanted

This is my first small 1.4.5 porting contribution, so I would appreciate feedback on whether this is the expected way to resolve this TODO and whether removing the note from `PortingNotes_1.4.5.md` is appropriate.

### DIFF
--- a/PortingNotes_1.4.5.md
+++ b/PortingNotes_1.4.5.md
@@ -2,7 +2,6 @@
 
 Once all patches are fixed, these items need to be fixed or double checked:
 
-- GameModeData.cs no longer exists, patches need to be redistributed
 - NPCSpawnParams.gameModeData no longer exists. This was potentially used in IBestiaryInfoElement.
 - NPCSpawnParams.strengthMultiplierOverride renamed to difficultyOverride. Investigate if behavior changed.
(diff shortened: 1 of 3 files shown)


JavidPack:
Thanks, looks good. I was away for a week, so I didn't get to this until today. 

When I made that note, I wasn't quite sure what needed to be done, just that it needed investigating, so your work looking into this is perfect. I took this chance to update some of the game mode related docs as well as part of this PR so they are more clear.

PR #5229, opened 2026-06-27
fix: Fertilizer now grows modded saplings (resolves #3586)

### LINKED ISSUE #3586: Fertilizer does not grow modded trees
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

TML does not support the code path the [Fertilizer](https://terraria.wiki.gg/wiki/Fertilizer) takes when growing trees. The method `WorldGen.AttemptToGrowTreeFromSapling` it uses, which calls `WorldGen.TryGrowingTreeByType`, are not hooked by tmls `PlantLoader`, which means modded (palm) trees don't grow when fertilizer is used.
ExampleMods `ExampleSapling` should be updated to use the code similar to `AttemptToGrowTreeFromSapling` for its `RandomUpdate`

### Log File

_No response_

### Steps to reproduce

1. Plant Acorn on Example Block
2. Use Fertilizer
3. Nothing happens

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Reproduction frequency

_No response_

### Additional Information

_No response_

Mersal108:
### What is the bug?

Fertilizer (ItemID.Fertilizer) has no effect on modded sapling tiles, even when the mod correctly sets `TileID.Sets.CommonSapling[Type] = true`.

Fixes #3586

There were two separate guards blocking modded tiles:

1. **`Projectile.cs`** (the fertilizer projectile AI): checked `tile.type < TileID.Count` before calling `AttemptToGrowTreeFromSapling`, which unconditionally excluded every modded tile ID regardless of `TileID.Sets.CommonSapling`.

2. **`WorldGen.AttemptToGrowTreeFromSapling`**: a switch over vanilla tile IDs (20 / 595 / 615 / 590) with `default: return false` — so even if a modded tile reached this function it was silently rejected.

### How did you fix the bug?

- **`Projectile.cs.patch`** — removed the `tile.type < TileID.Count` guard. `TileID.Sets.CommonSapling` is already sized to cover modded IDs so the check was unnecessary.
- **`WorldGen.cs.patch`** — replaced `default: return false` in `AttemptToGrowTreeFromSapling` with `return TileLoader.GrowModSapling(x, y, tile.type)`, delegating to the modded tile's own growth logic.
- **`TileLoader.cs`** — added `GrowModSapling(int x, int y, int type)` which invokes a new `ModTile.GrowSapling` hook.
- **`ModTile.cs`** — added virtual `GrowSapling(int x, int y)`. The default implementation calls `WorldGen.GrowTree` and shows the grow particle effect if successful, matching vanilla fertilizer behavior out of the box. Modders can override to use a different growth method (e.g. `WorldGen.GrowPalmTree`) or add custom logic.
- **`ExampleSapling.cs`** — overrides `GrowSapling` with direct `GrowTree`/`GrowPalmTree` logic; `RandomUpdate` now delegates to it so the growth code isn't duplicated and fertilizer correctly bypasses the 1-in-20 random delay.

Calling `TileLoader.RandomUpdate` directly instead of adding a new `GrowSapling` hook, but that would include the modder's random-chance delay in `RandomUpdate`, meaning fertilizer would only sometimes work — inconsistent with vanilla behavior.

### DIFF
--- a/ExampleMod/Content/Tiles/Plants/ExampleSapling.cs
+++ b/ExampleMod/Content/Tiles/Plants/ExampleSapling.cs
@@ -63,25 +63,28 @@ namespace ExampleMod.Content.Tiles.Plants
 				return;
 			}
 
-			Tile tile = Framing.GetTileSafely(i, j); // Safely get the tile at the given coordinates
-			bool growSuccess; // A bool to see if the tree growing was successful.
+			GrowSapling(i, j);
+		}
+
+		// Overridden to handle both ExampleTree (GrowTree) and ExamplePalmTree (GrowPalmTree) styles based on TileFrameX.
+		public override bool GrowSapling(int x, int y) {
+			Tile tile = Framing.GetTileSafely(x, y);
+			bool growSuccess;
 
 			// Style 0 is for the ExampleTree sapling, and style 1 is for ExamplePalmTree, so here we check frameX to call the correct method.
 			// Any pixels before 54 on the tilesheet are for ExampleTree while any pixels above it are for ExamplePalmTree
 			if (tile.TileFrameX < 54) {
-				growSuccess = WorldGen.GrowTree(i, j);
+				growSuccess = WorldGen.GrowTree(x, y);
 			}
 			else {
-				growSuccess = WorldGen.GrowPalmTree(i, j);
+				growSuccess = WorldGen.GrowPalmTree(x, y);
 			}
 
-			// A flag to check if a player is near the sapling
-			bool isPlayerNear = WorldGen.PlayerLOS(i, j);
-
-			// If growing the tree was a success and the player is near, show growing effects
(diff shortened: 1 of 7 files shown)


Mersal108:
@Chicken-Bones hey, sorry for the bother!
if you get a chance can you please take a look!
thank you!!

JavidPack:
I should be able to get to this tomorrow. I was out last week.

JavidPack:
Thanks for working on this. The biggest issue is you seem to have manually typed out these patches. You need to use `setup` to create the patches, otherwise they very likely will be malformed and won't work, as these were before I regenerated them.

I've added more parameters so we can better support infused fertilizer and differentiate between natural and unnatural tree growth.

PR #5230, opened 2026-06-27
Document new fields in NPCID.Sets

efcawesome:
### What is the new feature?
Adds documentation for the fields added by 1.4.5 in NPCID.Sets

Documentation helps modders understand what vanilla code is doing.

N/A

N/A


JavidPack:
Thanks, this is great.

PR #5231, opened 2026-06-28
Fix opening legacy presents

Silencersn:
### What is the bug?
Original：
``` csharp
private static bool TryOpenContainer_GrantItems(Item item, Player player)
{
	if (ItemID.Sets.BossBag[item.type]) {
		player.OpenBossBag(item.type);
	}
	else if (ItemID.Sets.IsFishingCrate[item.type]) {
		player.OpenFishingCrate(item.type);
	}
	else if (item.type == 3093) {
		player.OpenHerbBag(3093);
	}
	else if (item.type == 4345) {
		player.OpenCanofWorms(item.type);
	}
	else if (item.type == 4410) {
		player.OpenOyster(item.type);
	}
	else if (item.type == 1774) {
		player.OpenGoodieBag(1774);
	}
	else if (item.type == 6142) {
		player.OpenChilletEgg(6142);
	}
	else if (item.type == 3085) {
		if (!player.ConsumeItem(327, reverseOrder: false, includeVoidBag: true))
			return false;

		player.OpenLockBox(3085);
	}
	else if (item.type == 4879) {
		if (!player.HasItemInInventoryOrOpenVoidBag(329))
			return false;

		player.OpenShadowLockbox(4879);
	}
	else if (item.type == 1869) {
		player.OpenPresent(1869);
	}
	/*
	else {
		if (item.type != 599 && item.type != 600 && item.type != 601)
			return false;
	*/
	else if (item.type == 599 && item.type == 600 && item.type == 601) { // this is always false
		player.OpenLegacyPresent(item.type);
	}
	else { // Added by TML.
		player.DropFromItem(item.type);
	}

	ItemLoader.RightClickCallHooks(item, player);
	return true;
}
```

### How did you fix the bug?
Change `&&` to `||`.

Leave it as is, because these three items are unobtainable now.

### DIFF
--- a/patches/tModLoader/Terraria/GameContent/ItemDropRules/ItemDropDatabase.TML.cs
+++ b/patches/tModLoader/Terraria/GameContent/ItemDropRules/ItemDropDatabase.TML.cs
@@ -1189,6 +1189,18 @@ partial class ItemDropDatabase
 		RegisterToItem(ItemID.Present, new SequentialRulesNotScalingWithLuckRule(chanceDenominator: 1, rules));
 	}
 
+	private void RegisterLegacyPresent()
+	{
+		IItemDropRule[] rules = new IItemDropRule[]
+		{
+			ItemDropRule.ByConditionNotScalingWithLuck(new Conditions.IsHardmode(), ItemID.SnowGlobe, chanceDenominator: 14),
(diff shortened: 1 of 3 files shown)


JavidPack:
Thanks for finding this. Note that I added code for the actual drops, which were never implemented it seems. Make sure to test your fixes in-game to make sure what you are fixing is fixed completely.

PR #5233, opened 2026-06-28
Fix ExampleLamp flame drawing and update related docs

### LINKED ISSUE #5105: GetTileFlameData docs and ExampleLamp misleading
`GetTileFlameData` docs link to `ExampleLamp` as an example of drawing flame textures, but this is actually misleading. It's just a coincidence that the approach in `ExampleLamp` works.

The issue is `ExampleLamp.PostDraw` shows drawing the whole flame texture all at once. A modder following this example for a wider tile (with flames in those other tiles) would run into issues due to tiles drawing in order from top to bottom, left to right. The flames would be drawn over by subsequent tile draws. It also doesn't account for padding, but ExampleLamp_Flame.png does have padding. Someone following this example for flame in tiles other than the top left would have the flames incorrectly offset.

A more correct approach would be using `PreDraw` to call `AddSpecialLegacyPoint` and then use `SpecialDraw` to actually draw the flame texture. The flame texture itself would likely need to be without padding to correctly draw with the flickering offset logic as well.

Also, docs for `PostDraw` and `PreDraw` should probably mention how tiles are drawn one after the other, and that `PostDraw` for one tile happens before the tile next to it is drawn, not after all tiles on screen are drawn.

ibrahimk21:

ExampleLamp now uses `DrawEffects` + `AddSpecialLegacyPoint` + `SpecialDraw` to draw its flame texture instead of `PostDraw`. The previous approach had two problems:
- Flames drawn in `PostDraw` are overdrawn by sprites of tiles drawn after them, which causes visual issues on multi-tile objects.
- The source rectangle used frame coordinates to index into the flame texture, which produces incorrect results for any tile other than the top-left.

`ExampleLamp_Flame.png` has also been cropped from 36×54 to 16×16 to remove the tile-sheet-sized padding, matching how the texture is now drawn in `SpecialDraw`.

The `GetTileFlameData`, `PreDraw`, and `PostDraw` XML docs have been updated to clarify tile draw order and point to the correct approach for drawing flames on non-MultiTileVine tiles.

Closes #5105


ibrahimk21:
Hi guys, this is my first contribution to the tModLoader project (and my first contribution to an open source project in general). Terraria has been a big part of my life for the last 7 years now and tModLoader has only made my love for it much bigger, so contributing to it is a real pleasure. I am new to this so any feedback is welcomed and I am open to learn from anyone. Thanks for all the people contributing to this project and making it possible to mod Terraria as easily as it is.

JavidPack:
This is great. Thanks a lot.

PR #5235, opened 2026-06-28
[1.4.5] <right> and <left> Click Porting Notes

Rijam:
### What is the bug?

* `<right>` and `<left>` are no longer auto translated to "Right Click" and "Left Click".

### How did you fix the bug?

* Replaced with existing `{InputTrigger_UseOrAttack}` (left click) and `{InputTrigger_ToggleOrOpen}`, `{InputTrigger_InteractWithTile}`, or `{InputTrigger_InteractWithTileUI}` (right click)

* We could register `<right>` and `<left>` as global substitutes if wanted to bring those back. However, `{InputTrigger_XYZ}` has different buttons depending on the context for the controller.
  * Example: `{InputTrigger_ToggleOrOpen}` is right click on keyboard but is the Grapple button on controller.

### Porting Notes

* Replace all `<left>` with `{InputTrigger_UseOrAttack}`.
* Replace `<right>` with one of the three options:
  * `{InputTrigger_ToggleOrOpen}` Used by items that have right click functionality in the inventory. Example: Grab bags and items that transform into other items.
	* `{$CommonItemTooltip.RightClickToOpen}` can still be used for "Right Click to open".
  * `{InputTrigger_InteractWithTile}` Used by things that have right click functionality with tiles or the world. Example: Placing items into a tile or right click alternate fire weapons.
  * `{InputTrigger_InteractWithTileUI}` Used by the crafting window to tell you can right click it to switch between the classic and modern styles.

#### Other keys

Lang.InitGlobalSubstitutions() has many other substitutions that may be useful.
* `{InputTriggerUI_BuildFromInventory}` -> QuickMount button
* `{?InputTriggerUI_TrashEnabled}` -> true if enabled
  * Example: "{?InputTriggerUI_TrashEnabled}If your Inventory is full, you can press {InputTriggerUI_Trash} to send items to the Trash.",
* `{InputTriggerUI_Trash}` -> Control or Shift depending on the user's settings. Or SmartSelect button for controller.
* `{InputTriggerUI_FavoriteItem}` -> FavoriteKey button (left alt) or SmartCursor button for controller.
* `{InputTrigger_QuickEquip}` -> Grapple button
* `{InputTrigger_Grapple}` -> Grapple button
* `{InputTrigger_LockOn}` -> LockOn button
* `{InputTrigger_RadialQuickbar}` -> HotbarMinus button
* `{InputTrigger_RadialHotbar}` -> HotbarPlus button
* `{InputTrigger_SmartCursor}` -> SmartCursor button
* `{InputTrigger_SmartSelect}` -> SmartSelect button
* `{ToggleArmorSetBonusKey}` -> Up or Down depending on the user's settings.

### DIFF
--- a/ExampleMod/Content/Items/Consumables/ExampleFishingCrate.cs
+++ b/ExampleMod/Content/Items/Consumables/ExampleFishingCrate.cs
@@ -14,6 +14,7 @@ namespace ExampleMod.Content.Items.Consumables
 			// Disclaimer for both of these sets (as per their docs): They are only checked for vanilla item IDs, but for cross-mod purposes it would be helpful to set them for modded crates too
 			ItemID.Sets.IsFishingCrate[Type] = true;
 			//ItemID.Sets.IsFishingCrateHardmode[Type] = true; // This is a crate that mimics a pre-hardmode biome crate, so this is commented out
+			ItemID.Sets.OpenableBag[Type] = true;
 
 			Item.ResearchUnlockCount = 10;
 		}
--- a/ExampleMod/Content/Items/HoldStyleShowcase.cs
+++ b/ExampleMod/Content/Items/HoldStyleShowcase.cs
@@ -19,6 +19,8 @@ namespace ExampleMod.Content.Items
 		public override void SetStaticDefaults() {
 			SwitchingText = this.GetLocalization("Switching");
 			ThisIsText = this.GetLocalization("ThisIs");
+
+			ItemID.Sets.HasRightFire[Type] = true;
 		}
 
 		public override void SetDefaults() {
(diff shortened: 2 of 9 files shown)


Rijam:
Btw vanilla bug: ru-RU `GuideHelpText.Help_1065` in `Town.json` still uses `<right>` and `<left>` instead of the new substitutions.

JavidPack:
dotnet_2026-07-06_16-10-07

Signs don't seem to support chat tags when hovered at this moment. It would be good to fix eventually, and didn't work in 1.4.4 either, but it's a bit odd anyway to have input mode-specific text in something other players might see, so I changed it.

PR #5243, opened 2026-07-08
Flail API

### LINKED ISSUE #5240: ProjectileLoader.Colliding is never called for flails and some specific projectiles
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Mod capability as a Modder

### Description

The flail AI style skips running `ProjectileLoader.Colliding` entirely, other projectiles affected by this issue include True Nights Edge and Starlight.

Tyfyter:
### What is the new feature?
Flail API.

An entire vanilla weapon category with its own ai style currently can't be implemented with slight changes without completely reimplementing the ai style.

Sets could be used, but this approach more accurately predicts what modders might want to vary.

See Example Flail.

### ExampleMod updates
Example Flail now has doubled spinning range.

Also fixes #5240, because that mostly affects flails.

### Porting Notes
- `(Mod|Global)Projectile.Colliding` has been fixed to now be called for some projectile types and aiStyle values that were previously skipped. This has a slight chance of affecting some mods.

### DIFF
--- a/ExampleMod/Content/Projectiles/ExampleFlailProjectile.cs
+++ b/ExampleMod/Content/Projectiles/ExampleFlailProjectile.cs
@@ -48,6 +48,7 @@ namespace ExampleMod.Content.Projectiles
 			return base.PreDrawExtras();
 		}
 
+		// These 2 hooks are specific to projectiles using ProjAIStyleID.Flail and allow customizing the behavior of that aiStyle to some degree.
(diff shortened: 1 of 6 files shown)


Tyfyter:
The advanced flail should probably be updated to be a true flail too, or removed since basic customization is no longer "advanced"

Tyfyter:
Maybe we should also add consts or an enum for the ai states.

PR #5246, opened 2026-07-09
[1.4.5] Fix modded bestiary tag text layout

bilimoing:
## What is the bug?

Modded bestiary info tags could appear blank in the bestiary entry details panel, even though their tooltip text appeared correctly when hovered.

This affected mod-provided bestiary tags such as the mod source tag and mod biome tag. Vanilla bestiary tags displayed normally.

The issue was caused by the modded tag text element still using the older offset-only layout. In 1.4.5, the matching vanilla tag UI gives the text element an explicit width and uses left padding, allowing the text to calculate a valid inner width and render properly.

## How did you fix the bug?

Updated the modded bestiary info tag text layout to match the vanilla bestiary tag layout.

The text now uses left padding and a full-width layout instead of only applying a left offset. This gives the text element a usable width for layout and dynamic scaling, so the tag name is drawn inside the panel instead of appearing blank.

## Are there alternatives to your fix?

One alternative would be to special-case modded tag text drawing manually, but matching the vanilla bestiary tag layout is smaller and keeps modded tags consistent with the UI behavior Terraria now expects.

## Testing

Verified in-game with ExampleMod bestiary entries. Mod source and mod biome tags now display in the bestiary entry details panel without needing to hover them.
6C224CBFBE7A412E5DFBC80EFC703C10
53FC3F2910A472F9936D2B00A70FCA02


bilimoing:
I suspected it might be a problem with string acquisition in the previous test, but upon further investigation, I found that it wasn't the case. After changing the UI, it was found that it could be displayed normally.

PR #5248, opened 2026-07-10
Updated AddBuff within Player.cs and MessageBuffer.cs

AjuliDev:
### Why should this be a part of TModLoader?
As per [https://github.com/AjuliDev/tModLoader/blob/1.4.5/PortingNotes_1.4.5.md](url):
`Player.AddBuff parameters changed. Will need to adjust docs and maybe inform modders of any behavior changes.`

- Updated docs for AddBuff to reflect new changes, with mention of the now-gone quiet parameter for current modders.
- Renamed **fromNetPvP** to **skipBuffAndDifficultyScaling** to better reflect it's actual use within the method.

### DIFF
--- a/patches/tModLoader/Terraria/ID/NPCAIStyleID.cs
+++ b/patches/tModLoader/Terraria/ID/NPCAIStyleID.cs
@@ -1,4 +1,5 @@
 using ReLogic.Reflection;
+using static Terraria.ID.NPCID;
 
 namespace Terraria.ID;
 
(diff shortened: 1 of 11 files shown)


Chicken-Bones:
This is good, but please revert the parameter rename. Reasoning:
- It makes it a bit harder to maintain the patches, we might rename it in vanilla. It's for internal net use only
- Maybe, we'll add a hook in the future to allow mods to adjust buff time on apply, it might not be difficulty scaling

AjuliDev:
Will do soon.

PR #5260, opened 2026-07-21
[1.4.5] Support for modded Special Seeds and Secret Seeds

Pointy-Orb:
### What is the new feature?
Adds ModSpecialSeed and ModSecretSeed classes that modders can use to add their own special/secret seeds. The classes have an `Enabled` flag that is saved and loaded per-world, as well as `PostWorldGen` and `ModifyWorldGen` hooks to allow the seeds to more easily change world generation.
A `Known` property within ModSecretSeed allows modders to automatically unlock modded secret seeds at load time if they wish, so they don't need to tell players the seed code to let them access their secret seed. 

The new Special Seeds menu added in 1.4.5 is a great place for modders to add per-world worldgen options and toggles for alternate gamemodes. Being delegated to its own menu, it poses virtually no risk of cluttering up the main world generation page, and a scrollbar allows the special seeds list to be easily expanded to accommodate multiple mods each adding their own seeds. 

The secret seeds list within the Special Seeds menu also has uses, being good for modders that want to add world generation options/modes that are subtler, smaller in scale, or intended to be combined with other secret seeds. Giving these options their own menu reduces the risk of the Special Seeds menu being cluttered with lots of small seeds that each only do one thing.

It is possible to only add support for just custom special seeds (or vise versa), but doing so doesn't leave room for the nuance of if a mod's new feature deserves to be a special seed or a secret seed. 
Someone who wants to add a special seed but can only add secret seeds have to worry about their new feature being forgotten in a sub-submenu. Someone who wants to add, say, 10 secret seeds but can only add special seeds could have their several simple modular additions crowding out larger and self-contained ones.

A new submenu can be added for just modded changes. This could have the potential to allow for more flexible options for modders, but is more work to set up and creates a divide between vanilla and modded content.

<details>
<summary>ExampleSpecialSeed class from ExampleMod</summary>

```
public class ExampleSpecialSeed : ModSpecialSeed
{
	private Asset<Texture2D> iconCorruption;
	private Asset<Texture2D> iconHallowCorruption;
	private Asset<Texture2D> iconCrimson;
	private Asset<Texture2D> iconHallowCrimson;

		IncludeInZenith = true;
		iconCorruption = ModContent.Request<Texture2D>($"ExampleMod/Common/SpecialSeeds/{Name}_IconCorruption");
		iconHallowCorruption = ModContent.Request<Texture2D>($"ExampleMod/Common/SpecialSeeds/{Name}_IconHallowCorruption");
		iconCrimson = ModContent.Request<Texture2D>($"ExampleMod/Common/SpecialSeeds/{Name}_IconCrimson");
		iconHallowCrimson = ModContent.Request<Texture2D>($"ExampleMod/Common/SpecialSeeds/{Name}_IconHallowCrimson");
	}

	public override IEnumerable<string> SpecialSeedNames()
	{
		yield return "example";
	}

	public override IEnumerable<int> SpecialSeedNumbers()
	{
		yield return 1337;
	}

	public override IEnumerable<int> SpecialSeedNumbers() {
		yield return 1337;
	}

	public override ModMenu WorldGenMenu => ModContent.GetInstance<ExampleModMenu>();

	public override Asset<Texture2D> GetSeedTexture(bool isCorruption, bool isHardMode) {
		if (isCorruption) {
			return isHardMode ? iconHallowCorruption : iconCorruption;
		}
		else {
			return isHardMode ? iconHallowCrimson : iconCrimson;
		}
	}

	public override void ModifyWorldGenTasks(List<GenPass> tasks) {
		//Add a GenPass immediately after the "Grass" pass. ExampleOreSystem explains this approach in more detail.
		int index = tasks.FindIndex(i => i.Name.Equals("Grass"));

		if (index != -1) {
			//ExampleSpecialSeedPass replaces grass and dirt with ExampleBlocks
			tasks.Insert(index+1,new ExampleSpecialSeedPass("Example Special Seed Changes", 200f));
		}
	}
}
```
</details>

<details>
<summary>Result in menu</summary>

menu_example
</details>

<details>

<summary>Effect in world</summary>

world_example

This system uses the seed's flag to change the background:
```
public class ExampleSpecialSeedSystem : ModSystem
{
	public override void ModifySunLightColor(ref Color tileColor, ref Color backgroundColor) {
		if (!SeedLoader.SeedEnabled<ExampleSpecialSeed>())
			return;
		backgroundColor = Color.Lerp(backgroundColor, Color.Black, 0.8f);
	}
}
```

</details>

### ExampleMod updates

Added the following:
**ExampleSpecialSeed** - demonstrates basic usage of ModSpecialSeed
**ExampleAdvancedSpecialSeed** - demonstrates more of ModSpecialSeed's hooks as well as adding enough seeds to make the special seed menu scroll
**ExampleSpecialSeedSystem** - demonstrates the usage of the `SeedLoader.SeedEnabled<>()` method to create an active effect for ExampleSpecialSeed worlds
**ExampleSecretSeed** - demonstrates basic usage of ModSecretSeed 
**ExampleSecretSeedPlayer** - Like ExampleSpecialSeedSystem, demonstrates `SeedLoader.SeedEnabled<>()` to reduce player gravity on worlds with ExampleSecretSeed enabled.

### Porting Notes

Any challenge mods, mods that add alternate gamemodes, and mods that exclusively change worldgen should consider moving their mod's features to a special seed or secret seed toggle if they deem it appropriate. 
Mods which used IL edits or detours to add appropriate toggles or customization to the world creation page should consider instead using a special seed that uses the `OnSeedButtonPress` hook to open a submenu for customization.


ink-zoey:
Excited to see this, although I'm curious as to why theres a distinct seperation between 'special' and 'secret' seeds as opposed to some general unlock condition shared between them. Additionally GetSeedTexture should just return a Texture2D, and the patch to `UIScrollbar` adding the `OnScroll` action is very mysterious (same goes for other miscellaneous UI patches here, I think they should be better justified and documented.) 👀

Pointy-Orb:
I'm glad you're excited, sorry if this response is too long.

 Special and secret seeds are distinct mainly because of how different they are in vanilla; special seeds are all singletons which inherit from `AWorldGenerationOption`, while each secret seed is an instance of the `SecretSeed` class. I think while it is technically possible to combine them, it would lead to a lot of places in the code where very different things would have to be done depending on the one field. For example, `ModSpecialSeed` controls an instance of `ModSpecialSeedUIOption`, which is an internal class created to allow for instances of AWorldGenerationOption to be made and used in the creation of the `GroupOptionButton<AWorldGenerationOption>`s in the special seeds menu. On the other hand, `ModSecretSeed` controls an instance of `SecretSeed`, which itself managed in _its_ own list via `SecretSeedsTracker`.

I do agree that `GetSeedTexture` should return the Texture2D directly. It will require a bit of reworking, I had it return an asset initially because the asset returned would be added to a list of the world's different icons, which it would cycle through. I'll see if I can figure out a way to have the list item added be something that connects to the `GetSeedTexture` method like a property, and maybe also provide a way to change the source rectangle so that modders will only have to load a single texture with multiple sections instead of multiple independent textures. Allowing the source rectangle to be changed could also maybe allow for animated textures like the GetFixedBoi icon, which could be fun.

I added `OnScroll` so that the rainbow ring trail around the Secret Seed icon could correct itself when the scrollbar scrolls. Without a hook like that, the trail gets smeared along the scroll line, which looks very stupid. 
I added `CustomColor` to `UIHeader` because I wanted to give modders the ability to change the text color of the world progress message for worlds generating with modded seeds (in the same manner that Celebration changes the text color to hot pink), but for some reason the color change on the world progress message for Celebration is hardcoded. Since UIHeader was used for more than just the world progress message, I felt I couldn't just tack `ModifyWorldProgressText` right before the draw call, and instead opted to have the color saved as a property at the same time that the other special seeds (For the Worthy, Drunk, Not the Bees, etc.) changed the world header text.
`SecretSeedsTracker.ProcessedConfig` was added to expose the private `_processedConfig` field so that `ModSecretSeed.Known`'s setter could use it to determine whether it is safe to add the ModSecretSeed's SecretSeedInstance now or if it should be queued for when the vanilla secret seeds are loaded (SecretSeedsTracker clears its secret seeds list when loading the world creation menu if _processedConfig is set to false). `ResetConfig` was added to set _processedConfig to false during unloading so that the new batch of modded secret seeds can be added instead. I feel that there are definitely ways to optimize the secret seed listing process if that is something you would want from me.

I'll put summaries of these explanations in comments next to the relevant parts, as well as try to rework `GetSeedTexture` as soon as I have time. Thank you for the feedback!

Pointy-Orb:
I did a little more digging regarding the `GetSeedTexture` thing, it seems that the way the cycling seed icons work internally is that a list of `Asset<Texture2D>`s is assembled when the world list icon is made, and that list is used to create a `UICyclingImage`, which is hardcoded to cycle through the `Asset<Texture2D>`s via `SetImage` according to a set interval and a private frame counter. I also looked closer at `Asset` and found that the "asset with a property for a value" I wanted doesn't exist.

In summary, making `GetSeedTexture` return a Texture and not an Asset is possible, but harder than I thought it would be.

I have the idea of maybe adding internal properties to `UICyclingImage` to expose the private `_currentTextureIndex` and `_framesCounted` fields in order to make room for the modded icon textures. A private method in `UIWorldListItem` hooking onto the world icon's `OnUpdate` event can serve as the caller for the `SeedLoader` method(s). What do you think?

ink-zoey:
There are ways of creating an `Asset<T>` from a value of `T` (that don't involve saving the asset to a temporary location and loading it,) but I'd honestly just opt for patching `UICyclingImage` to use regular textures internally; may require a bit of finesse to get working nicely depending on how vanilla passes textures in, but you can probably just have a seperate ctor to minimize the patches needed (and if you want to allow modifying the source rectangle I'd just replace `GetSeedTexture` with something like "ModifySeedDrawInfo" with `ref` args.) Anyway thanks for clearing things up, didn't know about the vanilla secret seed distinction (and for the UI patches, I'm pressing on them because I think every miscellaneous change that comes with a larger PR should be documented, and well at that.)

JavidPack:
I haven't reviewed this yet, but I don't see why you think you need to change `GetSeedTexture` from `Asset<Texture2D>` to `Texture2D`.

ink-zoey:
Modders should be allowed to pass in an arbitrary texture (ie. a RenderTarget2D) without having to create a dummy Asset<Texture2D> for it.

Pointy-Orb:
Do you think it would be cleanly possible to create overloads for both options?

Chicken-Bones:
I will look into providing a way to create `Asset<T>` from `T` separate to this PR

PR #5267, opened 2026-07-25
Updated AddBuff within Player.cs [Remade PR]

AjuliDev:
### Why should this be a part of TModLoader?

As per https://github.com/AjuliDev/tModLoader/blob/1.4.5/PortingNotes_1.4.5.md:

> Player.AddBuff parameters changed. Will need to adjust docs and maybe inform modders of any behavior changes.

- Updated docs for AddBuff to reflect new changes, with mention of the now-gone quiet parameter for current modders.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -1653,7 +1653,7 @@
  			Main.item[num].OverrideWith(theItemWeDrop);
  			theItemWeDrop = new Item();
  			if (slot == 58)
-@@ -3422,6 +_,22 @@
+@@ -3422,6 +_,21 @@
  		return -1;
  	}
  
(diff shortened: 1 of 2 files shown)


Chicken-Bones:
Please also remove the line from the porting notes :)

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -1653,18 +1653,26 @@
...
++	/// <br/>
++	/// <b>This accounts for the following: </b><br/>
++	/// If the player is immune to the buff; <br/>
++	/// Adjusting the buff <paramref name="time"/> for different game modes; <br/>
++	/// If the player already has the buff, where re-apply logic will happen. <br/>
++	/// <br/>
++	/// Vanilla buff types are found in <see cref="BuffID"/> and modded buff types are typically retrieved using
++	/// <see cref="ModContent.BuffType{T}"/>. <br/>
++	/// The network-syncing functions of the now-gone <b>quiet</b> parameter are now handled automatically. <br/>
++	/// Set <paramref name="fromNetPvP"/> to <b>true</b>, if you want to skip multiplying <paramref name="time"/> by 1.2 if <see cref="BuffID.DeadCellsPotionStation"/> is present, as well as skip multiplying by <see cref="Main.Difficulty"/>.
```
Chicken-Bones: Probably remove these 2 lines. They're not necessary to new readers.

You could instead say something like 
'player buff changes are synced by the owner using a `PlayerBuffs` packet, instead of on add/remove. Adding a pvpBuff to a remote player will send an `AddPlayerBuffPvP` packet instead of adding the buff locally.'

AjuliDev: Replaced them with your suggestion, and sorry for the fuss, I have a bad habit of doing these when I'm dead tired xP

Should we note somewhere the exact details / changes that were discussed in the 1.4.5 Initial Patching channel or just leave it as is, with the specifics being on discord?

PR #5269, opened 2026-07-26
[1.4.5] Example Voice Accessory & ModPlayer Hurt/Death Sound Hooks

Rijam:
### What is the new feature?

Allows modders to create a voice accessory to change the sound the player makes when taking damage and/or dying. https://terraria.wiki.gg/wiki/Voice_accessories

Voice accessories:
* Change which sound is played when getting hurt.
* Some change the sound that is played when dying.
* Will play their sound when then item is used. Uses a ParticleOrchestra to broadcast the sound.
* Can randomly generate as chest loot (except the Whoopie Cushion).
* Can randomly spawn inside slimes in Skyblock worlds (except the Whoopie Cushion).
* Slimes that spawn with voice accessories are given triple HP.
* Slimes holding a voice accessory item makes the sound of the voice accessory when taking damage.

Player wearing the item:

https://github.com/user-attachments/assets/49933107-2140-41f8-8503-1db828fef2c1

Slime holding the item:

https://github.com/user-attachments/assets/1fc71dce-77c9-44f2-a7b3-678e303d459e

* Create an item with `Item.voiceSlot = Type`.
  * Or use `Item.DefaultToVoiceOverrideAccessory(Type)`. 
* Two new hooks `ModItem.VoiceItemHurtSoundOverride(Entity entity, Vector2 soundPosition)` and `ModItem.VoiceItemDeathSoundOverride(Entity entity)`.
  * Play your sound in these hooks. Return true to prevent vanilla from playing other sounds.
  * Entity is used instead of Player because slimes run this hook, too.
* Add your item to `ItemID.Sets.IsAVoiceChangeItem[Type] = true` to let naturally spawning slimes in Skyblock that spawn with this item gain triple HP.
* Add your item to the new `ItemID.Sets.VoiceChangeItemForChestLootAndSlimes.Add(Type)` to let the item be randomly generated in chest loot or spawn in slimes in Skyblock.
* Two new hooks `ModPlayer.PlayerHurtSound(Vector2 soundPosition)` and `ModPlayer.PlayerDeathSound()` for more generic modification of the player's sounds.
  * Return true to prevent vanilla from playing other sounds.

Support the ability for modders to create a voice accessory to change the sound the player makes when taking damage and/or dying. 

* The ModItem hooks only work if `Item.voiceSlot = Type` is set and them item is an accessory in the social accessory slots. This might be confusing for modders who try to use the hooks without setting `Item.voiceSlot = Type` and then find they do nothing.
  * Example: this doesn't work with mounts and wouldn't work with armor set bonuses.
* Maybe the ModPlayer hooks could just be part of `ModPlayer.ModifyHurt`/`ModPlayer.OnHurt` and `ModPlayer.PreKill`

<details>
<summary>Other approaches</summary>

#### The "Implement it yourself" approach
* Showcase detouring `Player.PlayHurtSound()` and `Player.PlayDeathSound()` and not calling orig.
  * Pretty simple detours and gives EM more detour examples.
  * Doesn't take into account slimes holding the item.

#### The simple hook approach (What this PR is)
* Create an item with `Item.voiceSlot = Type`. This is an easy way to ensure each item has a unique slot ID.
  * Or use `Item.DefaultToVoiceOverrideAccessory(Type)`. 
* Two new hooks `ModItem.PlayerHurtSoundOverride(Entity entity)` and `ModItem.PlayerDeathSoundOverride(Entity entity)`.
  * Play your sound in these hooks. Return true to prevent vanilla from playing other sounds.
  * Entity is used instead of Player because slimes run this hook, too.
* The `Player.voiceOverride` (Which is the item's `Type`) gets looked up in `ContentSamples.ItemsByType` and runs the hooks.
  * Works with slimes holding the item, too.
* Doesn't work with anything that isn't an accessory in the social accessory slots.
  * Example: this doesn't work with mounts and wouldn't work with armor set bonues.
  * Maybe ModPlayer versions of the hooks could be added.
* The hooks only work if `Item.voiceSlot = Type` is set. This might be confusing for modders who try to use the hooks without setting `Item.voiceSlot = Type` and then find they do nothing.

#### EquipmentLoader approach
* Use attribute `[AutoloadEquip(EquipType.Voice)]` to automatically assign `Item.voiceSlot`
* EquipmentLoader is made to load textures, which these voice accessories don't do.
* We can't just assign a SoundStyle to the equipment because the vanilla Chicken Charm has a rare chance of playing an alternate sound which we want to be able to replicate that behavior.
</details>

### ExampleMod updates

* `ExampleVoiceAccessory` added and `ExampleCostume` (Charm of Example) updated to showcase the ModItem hooks.
* `ExampleHurtDeathSoundPlayer` with two simple examples to showcase the ModPlayer hooks. 

### Porting Notes

None, unless you were detouring/IL editing to change the hurt or death sounds the player makes. You can now use the new ModPlayer hooks.


patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5946,6 +5946,43 @@
  	}
  
  	public void PlayHurtSound()
+ 	{
+ 		Vector2 vector = position;
++		/*
+ 		if (whoAmI == Main.myPlayer)
+ 			vector = new Vector2(-1f, -1f);
```
Chicken-Bones: In vanilla, this makes hurt sounds play 'globally' (like UI sounds) for the local player. Without this, if your camera is not centered on your player (eg scrying orb, fpv drone) and you get hurt, you may miss the sound.

I don't remember what the tML way is to mark a sound as global play off the top of my head

patches/tModLoader/Terraria/Player.cs.patch
```diff
@@ -5946,6 +5946,43 @@
...
++		*/
+ 
+ 		if (stoned) {
+ 			SoundEngine.PlaySound(0, vector);
+@@ -31210,6 +_,13 @@
+ 				return;
+ 		}
+ 
++		// Run ModItem.PlayerHurtSoundOverride to override the hurt sound if voiceOverride == ModItem.Type. PlayerHurtSoundOverride returning true will return.
++		if (voiceOverride > PlayerVoiceOverrideID.Other) {
```
Chicken-Bones: Don't really need this outer if

patches/tModLoader/Terraria/NPC.cs.patch
```diff
@@ -2585,6 +2585,25 @@
...
+ 					default:
++					/*
+ 						if (HitSound != null)
+ 							SoundEngine.PlaySound(HitSound, position);
++					*/
+ 						break;
++				}
++
++				// type == slime, ai[1] == the item the slime is carrying.
++				if (ai[1] > 0 && (!ContentSamples.ItemsByType[(int)ai[1]]?.ModItem?.PlayerHurtSoundOverride(this) ?? true)) {
```
Chicken-Bones: Seems like this would be better placed in the `default` block above.

```cs
if (ai[1] > 0 && ContentSamples.ItemsByType[(int)ai[1]]?.ModItem?.PlayerHurtSoundOverride(this) is true) {}
else
if (HitSound != null)
    SoundEngine.PlaySound(HitSound, position);
```

patches/tModLoader/Terraria/Player.cs.patch
Tyfyter: > I don't remember what the tML way is to mark a sound as global play off the top of my head

Playing the sound with ``null`` as the position

JavidPack: `vector = new Vector2(-1f, -1f);` should be fine to play the sound globally. If you follow the methods it'll translate that to null later on.

Rijam: Leaving it as `Vector2(-1f, -1f)` makes the sound play in the top left corner of the world. This might be a bigger tModLoader issue. For example, drinking a recall potion far away from spawn cuts off the sound effect after teleporting.

JavidPack: Ah, yeah, seems like recalling pans while moving fast on 1.4.5 and 1.4.4. Looks like the issue is some of the code here is passing in a SoundStyle which is calling into a path that doesn't check for -1, -1. I guess the solution might be to do something similar to the XYToOptionalPosition method somewhere in the call chain to validate the Vector2

Rijam: I added an XYToOptionalPosition overload and added it to two of the methods. This fixed the voice accessories, but it didn't fix the magic mirror/recall potion issue.

Rijam:
Updated for 1.4.5.8

PR #5272, opened 2026-07-27
Support new languages added in 1.4.5

JavidPack:
### What is the new feature?
Adds support for Japanese, Traditional Chinese, and Korean languages for tModLoader and mods.

### Changes
- Generated `tModLoader.json` files for the 3 new languages 
- Updated `tModLoader.json` files with localization entries for each workshop language tag
- Renamed the `tModLoader.TagsLanguage_Chinese` localization key to `tModLoader.TagsLanguage_SimplifiedChinese` for clarity. The internal name of the tag on the Steam workshop will not change since we are unsure how that would affect things.
- Fixed some vanilla languages selection localization keys that seemed wrong and inconsistent.
- Added pluralization support to the new languages, they all do not pluralize.
- Updated ExampleMod localization files to mention the new language options
- Updated `serverconfig.txt` to mention the new language options
- Updated localization update scripts to handle the new languages

### TODO
- [ ] We need to verify and update the font creation process
  - [ ] We'll need a new font template file containing the correct new character sets
  - Moved to be tracked in https://github.com/tModLoader/tModLoader/issues/5361
- [x] Update https://github.com/tModLoader/tModLoader/wiki/Localization#supported-languages
- [ ] We need to actually add the tags to the steam workshop and then update and run `UpdateWorkshopTagFiles.py`
- [x] We want #5247 changes on 1.4.5, so hold off on merging this PR until we do the regularly scheduled sync between 1.4.4 and 1.4.5 branches.

### Porting Notes
- Supporting these new languages in mods is the same process as the existing languages.
- If you have custom fonts, you may need to build your font files again and include the new characters in your font creation.
  - TODO: We will have new tooling for this made available sometime.


H-A-M-G-E-R:
Finally, Japanese mods!

Issue #4895, opened 2025-12-07
DrawData with the default parameters for NPCs and Projectiles


flye-name:
### Do you intend to personally contribute/program this feature?

Yes

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

A way to get typical `DrawData` for NPCs and Projectiles, enabling use cases such as:
```cs
Main.EntitySpriteDraw(
    NPC.GetDrawData() with { texture = Glow, color = Color.White }
);
```

```cs
for (int i = 0; i < Projectile.oldPos.Length; i++)
    Main.EntitySpriteDraw(
        Projectile.GetDrawData() with { position = Projectile.oldPos[i] + Projectile.Size / 2f, scale = new Vector2(1f - i / (float)Projectile.oldPos.Length) }
    );
```

```cs
Main.EntitySpriteDraw(
    NPC.GetDrawData() with { texture = Wings, sourceRect = WingsFrame, origin = Wings.Size() / 2 }
);
```

etc

### What does this proposal attempt to solve or improve?

This would make typical drawing much more convenient to write as well as making it less intimidating for newcomers if enough examples are used in ExampleMod

### Which (other) solutions should be considered?

Not really alternatives but I'm curious if it would be better to include optional parameters in the Get method over using the `with` keyword?

```cs
GetDrawData(position: whateverPosition, color: whateverColor)
```

```cs
GetDrawData() with { position = whateverPosition, color = whateverColor } 
```

`with` seems more appropriate in this case and would likely make it more readable
<sup>It would also easily introduce the keyword to those that don't learn C# properly before they start modding</sup>

but I'm not too sure

JavidPack:
This is a good idea, I'm just not sure how complicated it would be to implement. It'll take some investigating. If the method just returned a default DrawData without taking into account any custom logic from DrawNPCDirect, that might be easy to do, I'm just not sure if the intention is for the method to get all the tweaks or represent the defaults before any tweaks.

flye-name:
Personally I'd be fine with it only being default values for the use cases I'd have in mind but more opinions are needed methinks

JavidPack:
K, 5273 was merged with the projectile support so far.

PR #5273, opened 2026-07-28
Feature/projectile draw helper

### LINKED ISSUE #4764: Helper method for default projectile drawing parameters or add those parameters to Pre/PostDraw

No

Mod capability as a Modder

### Description

The Projectile drawing function of tmodloader is really difficult to use. I need to rewrite the drawing function every time, hoping to improve it. This is translated by a translation software

_No response_

_No response_

darronese:
### What is the new feature?
A helper function that returns the draw parameters vanilla would use for a projectile, so a custom ``PreDraw``/``PostDraw`` can reuse them instead of recreating the math. Closes #4764.

Modders who want custom drawing currently have to reimplement vanilla drawing logic by hand. This helper function removes that duplicated work and makes it easier to preserve vanilla behavior while customizing rendering.

- Adding parameters to ``PreDraw/PostDraw``
- ~~Returning ``DrawData`` instead of the new struct type~~ Done
- The vanilla code does a few "weird" things, but they were all replicated in this helper method because that is the whole point of the PR. We could consider making other methods that allow more customization, or that handle flipping, scale, and texture framing differently. 
- Does `GetDefaultDrawParameters` convey the correct purpose of the hook? Should it be in `ModProjectile` instead?

```csharp
public override bool PreDraw(Player player, ref Color lightColor) {
    var drawData = Projectile.GetDefaultDrawParameters(player, lightColor);

    // Do whatever customizations needed here
    Main.EntitySpriteDraw(drawData with { color = drawData.color * 0.5f });

    return false;
}
```

### ExampleMod updates
Updated `ExampleBullet.PreDraw` (afterimage), ‎`ExampleInteractableProjectile.PostDraw` (custom highlight rendering), and `ExampleGravityDebuffBullet.PreDraw` (afterimage) to use `Projectile.GetDefaultDrawParameters` and then tweak the resulting `DrawData` instead of instead of recomputing the draw parameters manually. 

Other projectiles used a lot of custom logic and were not suitable for using the new helper method.


ink-zoey:
[DrawData with the default parameters for NPCs and Projectiles](https://github.com/tModLoader/tModLoader/issues/4895) Related? I'd also opt for something like `GetDefaultDrawData` or just `GetDrawData`. also don't provide a custom struct for this, `DrawData` is fine, likely better since it has `SpriteBatch.Draw` extensions.
 It's also probably worth providing this for other entities.

patches/tModLoader/Terraria/Projectile.DefaultDrawParameters.cs
```diff
@@ -0,0 +1,82 @@
...
+		}
+		else {
+			sourceRectangle = new Rectangle(0, 0, texture.Width, texture.Height);
+
+			if (player != null && ownerHitCheck && player.gravDir == -1f) {
+				if (player.direction == 1)
+					effects = SpriteEffects.FlipHorizontally;
+				else if (player.direction == -1)
+					effects = SpriteEffects.None;
+			}
```
ink-zoey: Why are we changing the sprite direction based on the player's? `effects` should be based solely on `projectile.spriteDirection`. Modders can modify `Projectile.direction` in `AI` to match the owner's if needed.
Additionally this seems to be placed in the `else` case and will only run if `Main.projFrames[type] <= 0`, mistake?

patches/tModLoader/Terraria/Projectile.DefaultDrawParameters.cs
```diff
@@ -0,0 +1,82 @@
...
+
+		SpriteEffects effects = spriteDirection == -1 ? SpriteEffects.FlipHorizontally : SpriteEffects.None;
+
+		// Vanilla splits these into two separate paths in DrawProj_DrawNormalProjs:
+		// Animated projectiles draw a single frame while everything else draws the whole texture
+		Rectangle sourceRectangle;
+
+		if (Main.projFrames[type] > 1) {
+			int frameHeight = texture.Height / Main.projFrames[type];
+			sourceRectangle = new Rectangle(0, frameHeight * frame, texture.Width, frameHeight - 1);
```
ink-zoey: Why are we subtracting 1 from the frame height here? Framing/padding issues should be handled in the sprite.

JavidPack: The purpose of the hook is to replicate the exact logic in `DrawProj_DrawNormalProjs` so it can be used in tandem with vanilla logic easily. This is what vanilla uses.

patches/tModLoader/Terraria/Projectile.DefaultDrawParameters.cs
JavidPack: Same with this.

JavidPack:
I've made some adjustments, and I've asked for feedback in Discord.

darronese:
Hey, I've made some changes from the Discord feedback:

- Renamed ``GetDefaultDrawParameters`` -> ``GetDefaultDrawData`` since that seemed to be the consensus
- Added another doc listing explaining which projectiles the helper doesn't apply to (from direwolf420's suggestion), and used golf balls as an example since they were handled by a hardcoded type-specific branch in ``DrawProj_DrawNormalProjs``
- Documented a little more to the player parameter (from the Discord discussion)

PR #5279, opened 2026-08-01
feat: Display obsolete preview versions as such in the version switcher.

Mirsario:
The version switcher now grabs the highest major-minor from the running version + the available branches' names, highlighting previews as obsolete if they run behind that value.

image

Will need a cherrypick to 1.4.5, I should have done things the other way around.


JavidPack:
> Will need a cherrypick to 1.4.5, I should have done things the other way around.

Solxan regularly cherrypicks everything in a batch, it's fine.

Issue #4983, opened 2026-02-09
Developer tools added in 1.4.5


JavidPack:
### Description

The 1.4.5 update exposes some developer tools intended to be used by modders. These tools were compiled out of releases in previous updates, but the Terraria team has left them in the 1.4.5 assembly to allow modders to use them.

These tools include the following:

### World Generation Debugger
While generating a world, a preview of the world as it is generating is displayed and a list of generation passes is also displayed. 

(Add pictures)

### Chat Commands
Many additional chat commands allow modders access to many useful features.

(Add pictures)

# Discussion
tModLoader currently has a "developer mode", `ModCompile.DeveloperMode`, enabled by any of the following: `Debugger.IsAttached || Program.LaunchParameters.ContainsKey("-build") || FindModSources().Length > 0`. If the user is actively debugging the program with an attached IDE, they have any mod source folders in the mod sources folder, or if the `-build` command line argument is being used to build a mod. This existing developer mode enables the following: "This mod did not fully unload" warning and adding a "Force Reload" button to the mods menu.

There is also another flag, `ModCompile.activelyModding`, which is set to true if `Debugger.IsAttached` during mod loading or if any mod is built this game session. This flag, unlike the other, is intended to enable while actively doing modding work, allowing modders to have the normal game experience when not working on their mod. This flag controls the following currently: Additional achievement logging, displaying all exceptions in the in-game chat, additional logging for merging named custom sets, and additional logging for bad TileObjectData values.

## Questions
- How should the Terraria-provided modding tools be exposed to modders? We could just add a toggle, or we could piggy-back off of one of these existing modding flags.
-  Should we allow regular users to enable these new developer tools? They might be distracting, or be too much of a temptation for some users, especially since they can be catastrophic in multiplayer scenarios.
- What API need to be developed? What vanilla API exists that need to be documented?

Anyways, please share thoughts here or in Discord (`#tml-development` channel).

PR #5282, opened 2026-08-01
Restore error reporting when a world generation pass fails

dabbielocks:
### What is the bug?

If a world generation pass throws an exception, `WorldGenerator.RunPass` catches it, logs it, and then just carries on like nothing happened. The logging (`ReportException`) only writes to a debug trace file, and only shows anything in-game if a debug flag is on, which isn't the case for normal players. So `GenerateWorld()` still reports success, and the world gets saved anyway — meaning a broken/incomplete world could get saved with no indication anything went wrong.

There's a TODO about this in `PortingNotes_1.4.5.md` mentioning that a patch handling this ("shutdown the thread without saving") was removed because the new code "seems to handle it." From what I could trace through, it looks like that part didn't quite carry over — it logs the error but doesn't stop the save or show the player anything.

### How did you fix the bug?

- In `RunPass`, re-throw the exception after logging it instead of swallowing it.
- For the normal "Create New World" flow, that turned out to be all that was needed: `WorldGen.worldGenCallback` already has a try/catch that shows a proper error message and only saves if generation succeeds — it just wasn't getting the exception before.
- There's another path though: world autogen through `WorldFile.LoadWorld()`, which had no try/catch at all around this call. Just rethrowing there would've turned a silent failure into a crash instead, so I added a try/catch that sets the existing `loadFailed` flag, matching how this same method already handles a similar case a few lines up (cloud inaccessible).
- Also deleted the now-resolved TODO from `PortingNotes_1.4.5.md` and added a short note to `MigrationGuide_1.4.5.md`, since modders with custom world gen passes will now see failures instead of silent corruption.

Tested by temporarily forcing an exception in `RunPass`: the error dialog showed up correctly, the world didn't get saved, and normal generation still worked fine once I removed the forced exception.

One thing that gave me more confidence here: the 1.4.4 branch (current stable) already has almost this exact fix in its patch file for this same method — try/catch, show an error, then throw to stop the save. Even the comment "We need to shutdown the thread without it saving" is the same line quoted in the 1.4.5 TODO. So this is restoring behavior that's already shipped and working in 1.4.4, not something new. The one difference is 1.4.4 shows the error message directly inside `RunPass`, while 1.4.5's newer `Controller`/`ReportException` setup already handles showing the error one level up in `worldGenCallback`, so I didn't need to add that part myself.

Thought about having `GenerateWorld()` return a failure result instead of throwing, but rethrowing seemed simpler since it reuses the error handling already in `worldGenCallback`, and it matches what 1.4.4 already does.

First PR here, so let me know if I'm missing anything or should be approaching this differently!


Chicken-Bones:
Thanks for your contribution. This has some overlap with https://github.com/tModLoader/tModLoader/issues/4983, we need to decide how we want to surface the new world gen previewer/debug tool to modders (and potentially power users)

PR #5284, opened 2026-08-02
[1.4.5] Example Armor Set Bonus

bilimoing:
## What is the new feature?

Updates the 1.4.5 armor set bonus documentation and ExampleMod examples to use the new `ArmorSetBonus` registry flow for both vanilla and modded armor set bonuses.

This PR demonstrates how to:

- Register modded armor sets with `ArmorSetBonus.Create(...).Set<Head, Body, Legs>().Add()`.
- Apply the set effect directly through the registered `ArmorSetBonus.Effect` delegate.
- Modify vanilla armor set descriptions and replace vanilla armor set effects in `ModPlayer.ModifyArmorSetBonuses`.
- Provide set bonus descriptions through localization.
- Support partial set tooltip progress and the `SetBonusSinglePiece` tooltip.
- Register multiple valid armor set variants when needed.
- Keep `ModItem.IsArmorSet` only for vanity visual matching, not for the actual set bonus effect.

The 1.4.5 armor tooltip system can display partial set progress and identify which armor piece determines the set bonus. The old pattern of setting `Player.setBonus` and applying the effect through `ModItem.UpdateArmorSet` does not integrate with that registry-based behavior.

Documenting and showcasing the new flow gives modders a clear migration path and prevents new examples from relying on the legacy set bonus implementation.

This PR also keeps the API surface smaller by using `ArmorSetBonus.Effect` as the single runtime effect path. Modded sets and vanilla set modifications are both handled from `ModPlayer.ModifyArmorSetBonuses`, and the registered effect receives the active `Player`, so per-player logic does not need a separate `PreUpdateArmorSetBonus` hook.

### Use the `ArmorSetBonus` registry

- Modded sets are registered once during loading.
- Vanilla set definitions can be edited in the same hook.
- Tooltip text and partial set progress come from the registered set definition.
- Runtime behavior is handled by the registered `ArmorSetBonus.Effect` delegate.

### Add a generic set registration overload

The new generic overload lets modded sets use:

```cs
ArmorSetBonus.Create(ApplyExampleSetBonus, Mod.GetLocalization("ArmorSetBonus.ExampleSet").Key, ArmorSetBonus.PartType.Head)
	.Set<ExampleHelmet, ExampleBreastplate, ExampleLeggings>()
	.Add();
```

This avoids repeatedly writing `ModContent.ItemType<T>()` for the common modded head/body/legs case while preserving the existing integer ID overloads for vanilla items, mixed sets, partial sets, and variant arrays.

### Keep using `ModItem.IsArmorSet` and `ModItem.UpdateArmorSet`

- This preserves the familiar 1.4.4-style implementation.
- It does not make full use of the 1.4.5 armor set tooltip system.
- It requires manually managing displayed set bonus text through `Player.setBonus`.
- It cannot provide the same registry-based partial set information.

### Add a separate runtime armor set hook

- This would allow code to inspect the active armor set before its registered effect runs.
- It overlaps with `ArmorSetBonus.Effect`.
- It makes simple armor sets require one method for registration and another method for runtime behavior.
- Replacing or assigning `ArmorSetBonus.Effect` in `ModPlayer.ModifyArmorSetBonuses` provides the same practical capability with less API surface.

### Use `GlobalItem` armor set hooks

- This can be useful when a mod needs broad behavior across armor sets from multiple items or mods.
- It is more general than necessary for normal mod-owned armor sets.
- The `ArmorSetBonus` registry is a better fit for declarative armor combinations and tooltip integration.

## Implementation details

- Added a generic `ArmorSetBonus.Builder.Set<THead, TBody, TLegs>()` overload.
- Updated ExampleMod armor set registrations to use the generic `Set<THead, TBody, TLegs>()` form.
- Updated `MigrationGuide_1.4.5.md` with the new registry-based migration guidance.
- Documented `ModPlayer.ModifyArmorSetBonuses` as the single place to register modded sets and modify vanilla set definitions.
- Demonstrated replacing a vanilla armor set effect by assigning `ArmorSetBonus.Effect`.
- Migrated `ExampleHelmet` away from legacy `IsArmorSet` and `UpdateArmorSet` set bonus logic.
- Migrated `ExampleHood` so its stats are registered through `ArmorSetBonus` while retaining `IsArmorSet` only for vanity shadow matching.
- Removed the redundant `PreUpdateArmorSetBonus` runtime flow from the source implementation.

## ExampleMod updates

- Added `ExampleArmorSetBonusPlayer` as the central ExampleMod armor set registration example.
- Updated `ExampleHelmet` to use the new armor set bonus registration flow.
- Updated `ExampleHood` to separate vanity visual matching from actual set bonus behavior.
- Added localized armor set bonus descriptions under the `ArmorSetBonus` localization section.
- Updated `MigrationGuide_1.4.5.md` with examples for registering modded sets and modifying vanilla set definitions.

## Bugs

No known bugs have been identified.

Armor set effects still depend on the standard armor equipment slots. Vanity shadow behavior remains handled separately through `ArmorSetShadows`.
- Updated `MigrationGuide_1.4.5.md` with the migration guidance.
image
image
image


JavidPack:
Hi, thanks for putting in this effort. The existing PR, #5141, does several more things that we want, like LocalizedText support, an Identifier field, and supporting instanced ArmorSetEffect.

We will likely be going with that PR, as that is what we have been discussing on Discord for the last while. I think I will end up porting over support for ModItem generics, that is a good idea.

Issue #5114, opened 2026-04-27
Critical bug  Occurs only with the Copliot keyboard


miku1881:
### Version

Latest 1.4.4-stable

### OS

Windows

### Platform

Steam

### This bug affects...

Gameplay as a Player

### Description

When I switched the Copliot keyboard from half-width alphanumeric input to Japanese input, tmodloader crashed. Please fix this.

### Log File

[client.log](https://github.com/user-attachments/files/27108715/client.log)

[Launch.log](https://github.com/user-attachments/files/27108723/Launch.log)

[Natives.log](https://github.com/user-attachments/files/27108727/Natives.log)

[environment-client.log](https://github.com/user-attachments/files/27108732/environment-client.log)

### Steps to reproduce

I used the Copliot keyboard for text input

I switched the input from half-width alphanumeric characters to Japanese

It crashed while I was typing.

### Expected Behavior

I'd like to be able to type using the Copliot keyboard as well

### Actual Behavior

As soon as I tried to type, it froze and crashed.

### Reproduction frequency

This happens every time I use Japanese input with the Copliot keyboard.

### Additional Information

I think you’ll see what I mean once you try the Copliot keyboard.

JavidPack:
Well, I wasn't able to replicate the issue by trying a bunch of similar things, but I don't have a keyboard with a copilot button. We might need to see if anyone else reports something similar and see if we can narrow down the issue further.

Can you send the `C:\Steam\steamapps\common\tModLoader\tModLoader-Logs\client_v2026.2.3.2_04-27-26_10-03-30-8916_4.dmp.zip` file? It was mentioned in the logs and might help me further diagnose the issue.

miku1881:
[client_v2026.2.3.2_04-28-26_09-44-57-7929_4.dmp.zip](https://github.com/user-attachments/files/27146414/client_v2026.2.3.2_04-28-26_09-44-57-7929_4.dmp.zip)

I see. “Copliot Keyboard” refers to the IME.　 https://www.microsoft.com/ja-jp/area/atlife/copilotkeyboard

JavidPack:
I see, I was not aware of that and assumed you meant just a keyboard key.

Seems that program needs Windows 11 to run, so I won't be able to replicate your issue yet on my machine. I'll try and see if we can find someone else to test this issue.

miku1881:
If there are any Japanese Windows users on the dev team,

I think we should be able to reproduce the issue.

[...]

Issue #3635, opened 2023-07-17
System IME panel appear and cannot close 


LingYi0110:
### Version

Latest 1.4.4-preview

### OS

Windows

### Platform

Steam

### This bug affects

Gameplay as a Player

### Description

When I try to type, the in-game IME panel and the system IME panel appear at the same time, while the system IME panel cannot be closed.
![problem](https://github.com/tModLoader/tModLoader/assets/79823959/5f8305d2-12a1-422e-a004-68cc5ba4c583)
Also I noticed #2571 and #3375 , the in-game input method was not available in the previous version 1.4.3, so this issue may be related to the newly merged code.

In fact there's no need for a separate in-game panel, maybe a switch button to choose whether to use the in-game panel or the system panel would be better.

### Steps to reproduce

1.Open tModLoader 1.4.4 and type some characters by input method.
2.Close the textbox and the system IME panel still exist.

### Expected Behavior

When typing, the system panel shouldn't appear, because game-in panel replaces it, and it shouldn't not be able to close.

### Actual Behavior

Actually when you type the first character, it appears and stop updates the content what you type. It's like the program stopped responding.

### Reproduction frequency

frequently

Chicken-Bones:
Please provide a `client.log` file. I also need to know which IME you are attempting to use (a screenshot of your language & region settings would be good)

LingYi0110:
This is my log file :[client.log](https://github.com/tModLoader/tModLoader/files/12074541/client.log)
And I'm using Microsoft Pinyin IME.
This is a screenshot of my language and region settings
![image](https://github.com/tModLoader/tModLoader/assets/79823959/b6f8d5af-4d4a-40e9-a405-a6eaecc17a75)
![image](https://github.com/tModLoader/tModLoader/assets/79823959/ba026926-fead-4b72-8656-6738827143b6)

HydroH:
Wasn't able to reproduce, could you provide your Windows version? (Specific to build numbers)

LingYi0110:
OK, I am using Windows10 now. And the build numbers is 19045.3324
![{26471207-5E78-42b3-8AD1-25D0092A5A5E}](https://github.com/tModLoader/tModLoader/assets/79823959/0b382dd4-7c11-48cd-80ae-16d1a8eac10c)

LingYi0110:
By the way, according to my test, it seems like there's something wrong with windows message hook .

[...]

PR #5285, opened 2026-08-03
Fix incompatibility with certain IMEs

HydroH:
### What is the bug?
#3635 (partially), #5114 ,delayed candidate list update for some IMEs

### How did you fix the bug?
The cause of #3635 is that Windows 10 versions of Microsoft Pinyin that use the new TSF interface instead of old Imm32 framework, do not behave well on the Windows TSF-Imm32 compatibility layer. By consuming `WM_IME_COMPOSITION` messages, incompatible IMEs will properly draw their own candidate lists and in-game candidate lists are hidden, which is a valid fallback.
There are two causes for #5114 :
- The previous WinImm32Ime implementation made an incorrect assumption that strings are tightly packed in candidate lists. Certain IMEs do not follow this convention and the previous implementation will throw an OOB error in this case. The new implementation does not make this assumption and instead searches for UTF-16 null terminators from each offset start.
- The previous implementation updates candidate lists as soon as a `IMN_CHANGECANDIDATE` message is received. Some IMEs trigger multiple `IMN_CHANGECANDIDATE` messages in one frame, and some of them contain corrupted buffer. Delaying the update to the next frame fixes this problem. This also fixes candidate window getting delayed updates in some IMEs.

There is an ongoing experimentation on reimplementing ReLogic.Native.dll in 64-bit, and would support a full vanilla-like IME experience.

### DIFF
--- a/patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
+++ b/patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
@@ -18,6 +18,7 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
 	private string[] _candList = Array.Empty<string>();
 	private uint _candSelection;
 	private uint _candPageSize;
+	private bool _isCandDirty;
 
 	public uint SelectedPage => _candPageSize == 0 ? 0 : _candSelection / _candPageSize;
 
@@ -27,7 +28,16 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
 
 	public override uint SelectedCandidate => _candPageSize == 0 ? 0 : _candSelection % _candPageSize;
 
-	public override uint CandidateCount => Math.Min((uint)_candList.Length - SelectedPage * _candPageSize, _candPageSize);
+	public override uint CandidateCount {
+		get {
+			// Lazily update candidate list at most once per frame to avoid race conditions.
+			// CandidateCount is always read first so we use it.
+			if (_isCandDirty) {
+				UpdateCandidateList();
+			}
+			return Math.Min((uint)_candList.Length - SelectedPage * _candPageSize, _candPageSize);
+		}
+	}
 
 	public WinImm32Ime(WindowsMessageHook wndProcHook, IntPtr hWnd)
 	{
@@ -83,6 +93,7 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
 		try {
 			int size = NativeMethods.ImmGetCandidateList(hImc, 0, ref MemoryMarshal.GetReference(Span<byte>.Empty), 0);
 			if (size == 0) {
+				// This usually means candidate list is not ready, wait for next frame
 				_candList = Array.Empty<string>();
 				_candPageSize = 0;
 				_candSelection = 0;
@@ -109,16 +120,31 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
 					}
 					end += 2;
 				}
+
 				candStrList[i] = Encoding.Unicode.GetString(buf[start..end]);
 			}
 
+			_isCandDirty = false;
 			_candList = candStrList;
 			_candPageSize = candList.dwPageSize;
 			_candSelection = candList.dwSelection;
 		}
+		catch (Exception e) when (e is ArgumentOutOfRangeException or IndexOutOfRangeException) {
+			// Some IME occasionally send malformed candidate buffers due to race conditions.
+			// Ignore them until next correct buffer is available.
+			Console.WriteLine($"Failed to parse candidate list: {e}");
+		}
 		finally {
 			NativeMethods.ImmReleaseContext(_hWnd, hImc);
 		}
+	}
+
+	private void ClearCandidateList()
+	{
+		_isCandDirty = false;
+		_candList = Array.Empty<string>();
+		_candPageSize = 0;
+		_candSelection = 0;
 	}
 
 	public override string GetCandidate(uint index)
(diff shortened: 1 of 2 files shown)


JavidPack:
I didn't ask on the issue, but does this issue apply to terraria as well? It would be good to know if we need to make this fix there as well.

HydroH:
> I didn't ask on the issue, but does this issue apply to terraria as well? It would be good to know if we need to make this fix there as well.

Terraria doesn't have this issue. Terraria uses a dll to handle ime, and it was removed when tml upgraded to 64-bit. WinImm32Ime is only present on tml.

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -58,17 +68,18 @@ private void FinalizeString(bool bSend = false)
...
 
-	private string GetCompositionString()
+	private string GetCompositionString(uint dwIndex)
 	{
 		IntPtr hImc = NativeMethods.ImmGetContext(_hWnd);
 		try {
-			int size = NativeMethods.ImmGetCompositionString(hImc, Imm.GCS_COMPSTR, ref MemoryMarshal.GetReference(Span<byte>.Empty), 0);
+			int size = NativeMethods.ImmGetCompositionString(hImc, dwIndex,
+				ref MemoryMarshal.GetReference(Span<byte>.Empty), 0);
 			if (size == 0) {
```
Chicken-Bones, line 77: Needs to be `size <= 0`

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -96,27 +109,53 @@ private void UpdateCandidateList()
 			var offsets = MemoryMarshal.CreateReadOnlySpan(ref candList.dwOffset, (int)candList.dwCount);
```
Chicken-Bones, line 109: probably best to guard `dwCount` here too. Make sure it's in range of `buf`, since `CreateReadOnlySpan` will not ensure it.

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -96,27 +109,53 @@ private void UpdateCandidateList()
...
+			_isCandDirty = false;
 			_candList = candStrList;
+			_candPageStart = candList.dwPageStart;
 			_candPageSize = candList.dwPageSize;
 			_candSelection = candList.dwSelection;
 		}
+		catch (Exception e) when (e is ArgumentOutOfRangeException or IndexOutOfRangeException) {
+			// Some IME occasionally send malformed candidate buffers due to race conditions.
+			// Ignore them until next correct buffer is available.
+			Console.WriteLine($"Failed to parse candidate list: {e}");
```
Chicken-Bones, line 139: mm, hopefully the race conditions aren't our fault, but if they're not, why doesn't vanilla get them.

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -83,7 +94,9 @@ private void UpdateCandidateList()
 		try {
 			int size = NativeMethods.ImmGetCandidateList(hImc, 0, ref MemoryMarshal.GetReference(Span<byte>.Empty), 0);
 			if (size == 0) {
+				// This usually means candidate list is not ready, wait for next frame
```
Chicken-Bones, line 97: Call `ClearCandidateList`?

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -17,17 +17,27 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
...
-	public override uint CandidateCount => Math.Min((uint)_candList.Length - SelectedPage * _candPageSize, _candPageSize);
+	public override uint CandidateCount {
+		get {
+			// Lazily update candidate list at most once per frame to avoid race conditions.
+			// CandidateCount is always read first so we use it.
+			if (_isCandDirty) {
+				UpdateCandidateList();
+			}
+
+			return Math.Min((uint)_candList.Length - _candPageStart, _candPageSize);
```
Chicken-Bones, line 38: Do we need to bounds check/clamp `_candPageStart <= _candList.Length`

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
```diff
@@ -17,17 +17,27 @@ internal class WinImm32Ime : PlatformIme, IMessageFilter
...
+	public override uint SelectedCandidate => _candSelection - _candPageStart;
 
-	public override uint CandidateCount => Math.Min((uint)_candList.Length - SelectedPage * _candPageSize, _candPageSize);
+	public override uint CandidateCount {
+		get {
+			// Lazily update candidate list at most once per frame to avoid race conditions.
+			// CandidateCount is always read first so we use it.
+			if (_isCandDirty) {
+				UpdateCandidateList();
+			}
```
Chicken-Bones, line 36: Better to extract this to a common method and call it from all properties that require update candidate list

Chicken-Bones (changes requested):

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
HydroH: Vanilla handles both TSF-based and Imm32-based IMEs natively. We only have Imm32 path here, so we need to rely on Windows' TSF-Imm32 compatibility layer, which does not work perfectly to say at least. That's also why many Microsoft IMEs have a "compatibility mode" that switches to Imm32 backends.

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
HydroH: We still need isCandDrity to be true. ClearCandidateList() resets dirty state.

Chicken-Bones: either `bool resetDirty = true` param, or just set `_isCandDirty = true` afterwards :)

patches/TerrariaNetCore/ReLogic/Localization/IME/WinImm32Ime.cs
Chicken-Bones: Okay, vanilla's `ReLogic.Native` is mostly just a fork of <https://github.com/microsoft/DXUT/blob/master/Optional/ImeUi.h>/`.cpp`

The fact that tML implementation uses an 'older' ime interface is a bit sad for future FNA projects based on tML, but hey, thanks for all your work!

PR #5286, opened 2026-08-04
[1.4.5] Enable screen shaders and render targets in legacy lighting modes

ink-zoey:
Based on my prior [WIP: Allow screen shaders when using Retro/Trippy lighting modes. (And additionally on the main menu.)](https://github.com/tModLoader/tModLoader/pull/4796)

### What is the new feature?
Legacy lighting modes will always use required targets (`WorldSceneLayerTarget`'s,`skyTarget`, `screenTarget`, and `screenTargetSwap`) when rendering, regardless of zoom.
`FilterManager::Begin`/`EndCapture` will now run regardless of lighting mode, should be noted that vanilla requirements (e.g. `Filters.Scene.CanCapture` and `Main.ForegroundSunlightEffects`) must still be `true`.
The `WaterDistortion` filter remains disabled in legacy modes.

Some mods may rely on screen shaders for important visual information and thus break in legacy lighting modes.

- Obsolete members could be removed, but would require extra patches and cause unnecessary porting friction.
- The new behaviour could be provided as a setting either for the user, or enabled by mods.
  - The old behaviour is done in vanilla to support older hardware, TML's system requirements already exceed this so a setting is likely not beneficial.

N/A

### ExampleMod updates
N/A

### Notes
- Vanilla's fallback case in `Main.InitTargets` has been replaced with a hard crash, this error needs localizing.

### Porting Notes
- Common modded logic pertaining to tile rendering similar to `var zero = Main.drawToScreen ? Vector2.Zero : new Vector2(Main.offScreenRange);` is legacy, and should be replaced with `var zero = new Vector2(Main.offScreenRange)` as `Main.drawToScreen` is always false. (Perhaps possible as a tModPorter task?)

### DIFF
--- a/patches/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs.patch
+++ b/patches/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs.patch
@@ -1,0 +1,30 @@
+--- src/TerrariaNetCore/Terraria/Graphics/Capture/CaptureCamera.cs
++++ src/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs
+@@ -180,19 +_,27 @@
+ 
+ 	public void BeginDrawCapture()
+ 	{
++		/*
+ 		if (Lighting.NotRetro) {
+ 			Main.instance.GraphicsDevice.SetRenderTarget(Main.skyTarget);
+ 			Main.instance.GraphicsDevice.Clear(Microsoft.Xna.Framework.Color.Transparent);
+ 			Filters.Scene.BeginCapture(_filterFrameBuffer1);
+ 		}
++		*/
++		Main.instance.GraphicsDevice.SetRenderTarget(Main.skyTarget);
++		Main.instance.GraphicsDevice.Clear(Microsoft.Xna.Framework.Color.Transparent);
++		Filters.Scene.BeginCapture(_filterFrameBuffer1);
+ 
+ 		Main.waterTarget = _waterTarget;
+ 	}
+ 
+ 	public void EndDrawCapture(Vector2 screenSize, Vector2 sceneSize, Vector2 sceneOffset)
+ 	{
++		/*
+ 		if (Lighting.NotRetro)
+ 			Filters.Scene.EndCapture(_frameBuffer, _filterFrameBuffer1, _filterFrameBuffer2, screenSize, sceneSize, sceneOffset);
++		*/
++		Filters.Scene.EndCapture(_frameBuffer, _filterFrameBuffer1, _filterFrameBuffer2, screenSize, sceneSize, sceneOffset);
+ 	}
+ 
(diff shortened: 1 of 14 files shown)


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -1006,6 +1014,22 @@
...
+ 	}
+ 
++
++	[Obsolete("Always true in TML")]
+ 	public static bool RenderTargetsRequired {
+ 		get {
++			/*
+ 			if (!(GameZoomTarget > 1f))
+ 				return GameViewMatrix.TransformationMatrix.M11 > 1f;
+-
```
Chicken-Bones: extra removed newline (just put it back after the `*/`)

patches/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs.patch
```diff
@@ -0,0 +1,30 @@
+--- src/TerrariaNetCore/Terraria/Graphics/Capture/CaptureCamera.cs
++++ src/tModLoader/Terraria/Graphics/Capture/CaptureCamera.cs
+@@ -180,19 +_,27 @@
+ 
+ 	public void BeginDrawCapture()
+ 	{
++		/*
+ 		if (Lighting.NotRetro) {
```
Chicken-Bones: Typically we'd make a patch which comments out the `if` using a block comment, and then inserts a new line directly after it with an opening `{`

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6681,6 +6705,20 @@
...
+ 			MapRenderer.DrawToMap(default(Microsoft.Xna.Framework.Rectangle));
+ 		}
+ 
++		/*
+ 		if (Lighting.UpdateEveryFrame)
+ 			drawToScreen = true;
+ 		else
+ 			drawToScreen = false;
++		*/
++		drawToScreen = false;
```
Chicken-Bones: Patch not required? Since `Lighting.UpdateEveryFrame` is now always `false` (you can mark that one as obsolete too)

patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6715,6 +6753,15 @@
  		if (gameMenu || player[myPlayer].gravDir == 1f)
  			Rasterizer = RasterizerState.CullCounterClockwise;
  		else
+@@ -53663,7 +_,7 @@
+ 		}
+ 
+ 		bool flag2 = Terraria.Graphics.Effects.Filters.Scene.CanCapture() || ForegroundSunlightEffects;
+-		bool flag3 = !drawToScreen && !mapFullscreen && !onlyDrawFancyUI && Lighting.NotRetro && flag2;
++		bool flag3 = /* !drawToScreen && */ !mapFullscreen && !onlyDrawFancyUI /* && Lighting.NotRetro */ && flag2;
```
Chicken-Bones: Only need to comment out the `Lighting.NotRetro` part here.

Chicken-Bones (changes requested):

Chicken-Bones (approved):

Mirsario (approved):

PR #5287, opened 2026-08-04
Update Steam Workshop Chinese Localization

ayuyoiM:


JavidPack:
Thanks, should be live soon.

PR #5288, opened 2026-08-04
Brazilian store page translation + Update and revise Brazilian localization

Pixelnando:


JavidPack:
Thanks, should be live soon.

PR #5295, opened 2026-08-05
[1.4.5] Force screen capture API and provide Main.finalScreenTarget

ink-zoey:
### What is the new feature?
Adds `ModSystem.RequiresScreenTarget` to force the screen to capture, and `Main.finalScreenTarget` that gives a reference to the last screen target used for filters.

Currently modders that want to reference the screen target are unable to do so without edits as `Main.screenTarget`/`Main.screenTargetSwap` are unreliable, and the screen is not always captured to, needing edits to `FilterManager.CanCapture` and `FilterManager.EndCapture` as well as a blank edit to `Main.DoDraw` to prevent inlining.
Additionally modders may want to implement their own screen filter systems, this minimizes the number of edits needed for that.

TODO

### Sample usage for the new feature:
Could be used for screen previews in UI
```cs
// In a class inheriting ModSystem:
public override bool RequiresScreenTarget()
    => exampleUserInterface?.CurrentState is not null;

// In some UIElement part of a UIState
protected override void DrawSelf(SpriteBatch spriteBatch)
{
    var dims = GetInnerDimensions().ToRectangle();
    spriteBatch.Draw(Main.finalScreenTarget, dims, Color.White);
}
```
Additionally has applications for post processing in UI e.g. liquid glass like effects.

### ExampleMod updates
N/A

### Porting Notes
- Relevant edits to `FilterManager.CanCapture` should be replaced with the `ModSystem.RequiresScreenTarget` hook.
- Code that relied on `Main.screenTarget` in UI contexts should be replaced with `Main.finalScreenTarget`.


patches/tModLoader/Terraria/Main.cs.patch
```diff
@@ -6721,6 +6721,15 @@
  			_isDrawingOrUpdating = false;
  		}
  	}
+@@ -53408,6 +_,8 @@
+ 			return;
+ 		}
+ 
++		finalScreenTarget = null;
```
Chicken-Bones: why set here rather than right before `EndCapture`?

patches/tModLoader/Terraria/ModLoader/ModSystem.cs
```diff
@@ -438,6 +438,14 @@ public virtual void ModifySunLightColor(ref Color tileColor, ref Color backgroun
 	/// <param name="scale">Brightness scale</param>
 	public virtual void ModifyLightingBrightness(ref float scale) { }
 
+	/// <summary>
+	/// Use this method and return true if you need the game to capture the screen (excluding UI,) ideal for use with <see cref="Main.finalScreenTarget"/>, and edits to FilterManager.EndCapture. <para/>
+	/// The game may capture the screen regardless of if this returns false explicitly, (for instance: if any vanilla filters are active.)
+	/// </summary>
+	/// <returns><see langword="true"/> if the game should capture the screen.</returns>
+	public virtual bool RequiresScreenTarget()
```
Chicken-Bones, line 446: may as well be a property instead of a method

patches/tModLoader/Terraria/Graphics/Effects/FilterManager.cs.patch
```diff
@@ -1,6 +1,35 @@
...
++		Main.spriteBatch.Begin(SpriteSortMode.Immediate, BlendState.AlphaBlend);
++
++		Main.spriteBatch.Draw(t2, Vector2.Zero, value, Color.White);
++
++		Main.spriteBatch.End();
+ 		for (int i = 0; i < 8; i++) {
+ 			graphicsDevice.Textures[i] = null;
+ 		}
+ 
++		Main.finalScreenTarget = t2;
```
Chicken-Bones: New code needs to be enabled if and only if the screen target is required by mods. Otherwise there's an extra blit for nothing.

patches/tModLoader/Terraria/Graphics/Effects/FilterManager.cs.patch
```diff
@@ -1,6 +1,35 @@
 --- src/TerrariaNetCore/Terraria/Graphics/Effects/FilterManager.cs
 +++ src/tModLoader/Terraria/Graphics/Effects/FilterManager.cs
-@@ -183,4 +_,13 @@
+@@ -158,7 +_,7 @@
+ 			linkedListNode = next;
+ 		}
+ 
+-		graphicsDevice.SetRenderTarget(finalTexture);
++		graphicsDevice.SetRenderTarget(t2);
```
Chicken-Bones: either comment original code line with a block comment, or include comment about what changed here

Chicken-Bones (changes requested):

patches/tModLoader/Terraria/Main.cs.patch
ink-zoey: Should the screen target be provided at the start of the draw cycle? Would it be better to not reset it to null?

patches/tModLoader/Terraria/ModLoader/ModSystem.cs
ink-zoey: resolved in collaborators chat

patches/tModLoader/Terraria/Main.cs.patch
Chicken-Bones: I don't see why not. What if someone wants to reference last frame's values? Isn't that exactly what the horizon render/sunlight effect shaders do?

Chicken-Bones (approved):

PR #5153, opened 2026-05-19
[1.4.5] Mod rating button

flye-name:
### What is the new feature?
a like / dislike button in the local mods menu, this directly rates the mod on the steam workshop right from the game.
image
(this PR also changes the delete button's sprite to be consistent with the other buttons in that corner)

the rating button's sprite was made by pilchardheck

### Why should this be part of tModLoader?
a lot of players aren't aware of the fact that you can rate mods, this feature would result in more data when checking the reception of a mod. I'll also be making a separate PR incorporating the star rating into the mod browser to add onto this factor, both of these features have been directly requested and approved of in the collaborators channel.

### Are there alternative designs?
the button could be placed in the description alongside the homepage and such but that feels redundant since the workshop would be one click away from that.

PR #5299, opened 2026-08-08
[1.4.5] Rating display in the Mod Browser (+ a minor fix for #5153)

flye-name:
### What is the new feature?
A way to see the ratings a mod has in the mod browser.

This PR also addresses an issue with https://github.com/tModLoader/tModLoader/pull/5153/ where the rating button would show up for all mods instead of the enabled ones only as originally intended.

..

By default, a mod will display the percentage of likes it has, hovering over its panel will instead display the total rating count.

If the mod doesn't have enough ratings it'll just display "No ratings" to avoid exaggerated percentages on smaller mods (the threshold was arbitrarily placed at 25 but it would be better for the maintainers to decide this.)

image

Currently, there's no way for players to gauge the general reception of a mod without going into the workshop and reading the comments or looking at the (very vague) star count. 

This would also be a consistent change with https://github.com/tModLoader/tModLoader/pull/5153/.

Alternative designs have been discussed in the Discord, most notably was displaying the ratings outright instead of the ratio between them.
image


JavidPack:
I've made a few changes. Got rid of the decimal place in the ratio, changed the ratio to match the weighted ratio used on the workshop, changes "No ratings" to "Not enough ratings" to match the workshop.

image

PR #5310, opened 2026-08-14
[1.4.5] Fixed item tags being drawn in the tooltip shadow pass, removed 8 obsolete rejected patches

justdoshiba:
### What is the bug?
During the 1.4.5 port, the patch that prevents item tags from being drawn in the shadow pass of tooltips was rejected.

Item tags such as `[i:...]` inside tooltips are drawn a second time in the shadow pass, producing a ghost item icon behind the tooltip text. Shadow draw colors can be black with reduced alpha (faded tooltips), and vanilla's `color != Color.Black` check misses these, so the shadow version of the item tag was drawn when it should not have been.

### How did you fix the bug?
Changed the check in `ItemTagHandler.UniqueDraw` to compare the RGB components (`color.R != 0 || color.G != 0 || color.B != 0`) instead of `color != Color.Black`, so black colors with any alpha value are correctly treated as shadow draws. The other hunk of the rejected patch (the `scale *= 0.75f` double-scale fix) was already absorbed into vanilla 1.4.5.

Additionally, 8 other rejected patch files were deleted after verifying they are obsolete in Terraria 1.4.5:
- `UIVirtualKeyboard.cs.rej`: vanilla now has the chest name editing system and a `maxLength` parameter
- `ChatMessageContainer.cs.rej`: vanilla added `Main.ChatLineWidthLimit`
- `CreativeUI.cs.rej`: vanilla absorbed the `SacrificeItem` refactor
- `UIText.cs.rej`: vanilla absorbed `ChatManager.GetStringSize` and the wrapped `MinWidth` handling
- `LiquidRenderer.cs.rej`: vanilla split shimmer into its own draw cache and method
- `DyeInitializer.cs.rej`: `FixRecipes` no longer exists in 1.4.5
- `WorldFile.cs.rej`: the cached field changes no longer apply
- `RecipeGroup.cs.rej`: the recipe group lookup system is disabled in 1.4.5

The shadow color construction in `DrawColorCodedStringWithShadow` could be changed instead, but that would alter vanilla behavior globally rather than at the item tag draw site. The obsolete `.rej` files could have been left in place, but they do not apply to the 1.4.5 code and would need to be resolved for the porting effort to be completed.


Tyfyter:
Shadows could be drawn in any color, and black text could be drawn without being a shadow, while this is vanilla behavior, it is arguably not correct behavior, and I would like to suggest that this assumption that all black text and only black text is shadows be replaced with an actual check for shadow drawing.

justdoshiba:
@Tyfyter Would a static flag in DrawColorCodedStringShadow work as the fix? That covers all shadow colors without touching the UniqueDraw signature.

Tyfyter:
I believe it would, while there's technically the question of how a snippet drawing text with shadows during shadow drawing should be handled, it's probably safe to ignore that possibility because it would require someone actually doing that

justdoshiba:
> I believe it would, while there's technically the question of how a snippet drawing text with shadows during shadow drawing should be handled, it's probably safe to ignore that possibility because it would require someone actually doing that

Implemented. Added ChatManager.IsDrawingShadow, set while the shadow pass draws, and ItemTagHandler now checks it instead of the color. The flag is reset in a finally, so it stays correct even if a snippet draws shadowed text mid-pass.

JavidPack:
Thanks. In the future, including unrelated rejected patch fixes isn't ideal. Those should go in a separate PR if possible, to keep the change history readable and to make reviewing and merging PRs more straightforward.

PR #5312, opened 2026-08-16
OnTileKilled hooks for ModTile and GlobalTile

queueAngel:
### What is the new feature?
Hook that runs when a tile is destroyed.

There is currently no way to guarantee an event is carried out when a terrain tile or vanilla multi-tile is destroyed without an IL edit. `KillTile` hooks in `ModTile` and `GlobalTile` may be used along with checking `if (!fail)`, however, there is the chance that a subsequent `GlobalTile` might override `fail`, making the condition not true anymore.

The placement for the new call to `TileLoader.OnTileKilled` could be moved further back, as the functionality isn't affected and may provide additional context to modders. It is currently placed immediately before tile data is cleared to allow modders to access this data inside the hook.
The hooks could also be split into `PreTileKilled` and `PostTileKilled`, where the latter runs after tile data is cleared, however I don't see an immediate use case.

Modders can use `OnTileKilled` to do things when a tile is destroyed by regular means, such as a player mining it.

Consider the following code. It makes a tile break neighboring tiles of the same type when broken, and has a 10% chance to release a firework.

```cs
public override void OnTileKilled(int i, int j) {
	Tile thisTile = Main.tile[i, j];
	thisTile.HasTile = false; // Prevents stack overflow
	for (int x = i - 1; x <= i + 1; x++) {
		for (int y = j - 1; y <= j + 1; y++) {
			if (y == i && x == j)
				continue;
			Tile t = Main.tile[x, y];
			if (t.HasTile && t.TileType == Type)
				WorldGen.KillTile(x, y);
		}
	}
	if (Main.netMode != NetmodeID.MultiplayerClient && Main.rand.NextBool(10)) {
		Projectile.NewProjectile(new EntitySource_TileBreak(i, j), i * 16f, j * 16f, 0f, -4f, ProjectileID.RocketFireworksBoxRed + Main.rand.Next(4), 0, 0, Main.myPlayer);
	}
}
```

Video showcase of the above:

https://github.com/user-attachments/assets/e42d8f42-865d-43d0-ac37-5ac498453018

### ExampleMod updates
Usage example including the firework effect from the above code is included in ExampleBlock.cs, though the effect could be changed or the example could be moved to another tile.

### DIFF
--- a/ExampleMod/Content/Tiles/ExampleBlock.cs
+++ b/ExampleMod/Content/Tiles/ExampleBlock.cs
@@ -30,9 +30,9 @@ namespace ExampleMod.Content.Tiles
 		}
 
 		public override void OnTileKilled(int i, int j) {
-			// 10% chance to spawn a mouse when broken.
-			if (Main.netMode != NetmodeID.MultiplayerClient && Main.rand.NextBool(10)) {
-				NPC.NewNPC(new EntitySource_TileBreak(i, j), i * 16, j * 16, NPCID.Mouse);
+			// 1% chance to release a firework when broken.
+			if (Main.netMode != NetmodeID.MultiplayerClient && Main.rand.NextBool(100)) {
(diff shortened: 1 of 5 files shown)


queueAngel:
by the way, I'm not sure if that's what the patch file should look like or if I messed up at some point, I only added two lines (one for the call to the hook and one blank line for spacing)

PR #5313, opened 2026-08-16
[1.4.5] Japanese, Traditional Chinese, Korean localization update

bilimoing:


bilimoing:
There are some newly updated localized contents. This is because at the beginning I didn't notice this branch, but I still localized it. This way, it will be more convenient during the future merge process, without the need for secondary localization.

JavidPack:
I'm a bit surprised this PR has 3 different languages, are you fluent in all these languages?

bilimoing:
> I'm a bit surprised this PR has 3 different languages, are you fluent in all these languages?

Yes, I have been learning continuously for a long time.

JavidPack:
Thanks.

PR #5317, opened 2026-08-18
[ExampleMod] ExampleInstancedItem expansion + comments

direwolf420:
When trying to help modders understand how instanced items work (items that track instanced data), I noticed there is no definitive example that goes in detail and explains which hooks are necessary. There are actually many applications of this concept in the mod, but with some inconsistencies, missing hooks (some intended, some not), or undocumented.

This PR expands the ExampleInstancedItem example to include most of the information modders would need, all in one file.

Common things modders wanted to know is how to handle control flow and lists/arrays for netcode, the example lent itself to showcase and explain it.

It also edits many of the partial examples of instanced items to refer to the ExampleInstancedItem one.

It also edits the "Hot Potato" example to include NetSend/NetReceive as I believe they are necessary for multiplayer to be more faithful to that specific example, otherwise dropped items will instantly reset. Load/SaveData are omitted as the timer is so low it doesn't really matter, it is not designed to be persistent.

I tested all the code in this PR in multiplayer and with saving/loading.

The other examples could be expanded to feature the full set of hooks to bring consistency, but I didn't deem it as necessary.

### 1.4.5
I believe this can be forwarded to 1.4.5 directly, at a glance I didn't see any differences in the code or comments used.

I created this PR on 1.4.4 because the current mod development still happens on that branch.

### DIFF
--- a/ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
+++ b/ExampleMod/Common/GlobalItems/WeaponWithGrowingDamage.cs
@@ -13,6 +13,9 @@ using Terraria.ModLoader.IO;
 // Related to GlobalProjectile: ProjectileWithGrowingDamage
 namespace ExampleMod.Common.GlobalItems
 {
+	// This class features an implementation of instance fields stored on any item.
+	// As the hooks and approach is mostly the same except an additional InstancePerEntity override,
+	// it is sufficient to refer to ExampleInstancedItem for the understanding of the concept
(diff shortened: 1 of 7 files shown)


direwolf420:
Actually, isn't ExampleDataItem ("Hot Potato") kinda redundant now? The only thing it showcases is a value type field + assigning it on craft. ExampleInstanceItem assigns its data on craft aswell but in a different way but uses a reference type field. Maybe the former can be merged into the latter, to showcase both value+ref type handling?

PR #5319, opened 2026-08-18
Recognize that mods required by both-side mods are essentially both-side & add sortIgnore

Tyfyter:
### What is the bug?
Sort ordering between mods guaranteed to be present on all sides (both-side mods and any mods directly or indirectly required by them) mods may be affected by mods which are not.

### How did you fix the bug?
Rewrote mod sorting.

Almost certainly.


Tyfyter:
The complete rewrite in [3f7ce7c](https://github.com/tModLoader/tModLoader/pull/5319/commits/3f7ce7ce5e2ad55cbaf2db13523d6a6436037b62) might be able to be replaced with just using ``knownSyncedMods.Contains`` as ``syncedSort``'s condition, but recognizing that mods required by synced mods are always present if all synced mods are present and ``sortIgnore`` are definitely important for cross-mod between mods with different sides.

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
```diff
@@ -687,8 +687,21 @@ private static TopoSort<LocalMod> BuildSort(ICollection<LocalMod> mods)
...
+				if (requirement.properties.side is not ModSide.Both and not ModSide.NoSync)
+					throw new ModSortingException([processing], "TODO: write error or warning, mods should never require mods with more restrictive sides");
+				if (!knownSyncedMods.Add(requirement))
+					toProcess.Enqueue(requirement);
+			}
+		}
+
 		var preSorted = mods.OrderBy(mod => mod.Name, StringComparer.InvariantCulture).ToList();
-		var syncedSort = BuildSort(preSorted.Where(mod => mod.properties.side == ModSide.Both).ToList());
+		var syncedSort = BuildSort(preSorted.Where(knownSyncedMods.Contains).ToList());
```
Chicken-Bones, line 704: Lgtm, but I'd prefer a recursive function.
```cs
var synced = new HashSet<LocalMod>();
void AddToSyncSet(LocalMod mod) {
    if (!synced.Add(mod)) return;
    foreach (var r in mod.properties.modReferences) {
        var dep = nameMap[r.mod];
        ...
        AddToSyncSet(dep);
}
foreach (var mod in preSorted.Where(mod => mod.properties.side == ModSide.Both))
    AddToSyncSet(mod);
```

patches/tModLoader/Terraria/ModLoader/Core/BuildProperties.cs
```diff
@@ -395,6 +400,10 @@ internal static void InfoToBuildTxt(Stream src, Stream dst)
 			sb.AppendLine($"sortAfter = {string.Join(", ", properties.sortAfter)}");
 		if (properties.sortBefore.Length > 0)
 			sb.AppendLine($"sortBefore = {string.Join(", ", properties.sortBefore)}");
+
+		string[] sortIgnore = properties.modReferences.Union(properties.weakReferences).Select(dep => dep.mod).Except(properties.sortAfter).Except(properties.sortBefore).ToArray();
```
Chicken-Bones, line 409: Isn't this a breaking change?

Tyfyter: How so? Unless I'm missing something no mod could have ``sortIgnore.Length > 0`` without having used sortIgnore (which no mod could have done before) or being edited after building (pretty sure that's extremely unsupported) so for existing mods this should just make the method take a little longer, right?

patches/tModLoader/Terraria/ModLoader/Core/ModOrganizer.cs
Tyfyter: Also an option, I think I tend to forget recursion is a viable solution for some problems;
No objections to yelling at modders for trying to make a both-side or nosync mod only load on one side by requiring a client or server side mod?
I should probably add a test for that, but it'll need its error message to be finalized first, and I'm bad at human languages.

PR #5320, opened 2026-08-18
Added Mod.Index

Tyfyter:
### What is the new feature?
``Mod.Index``, a property containing each mod's index.

Allows more optimized access to data one mod associates with all mods.

Could also add a ``SetFactory`` so mods can share sets.


Tyfyter:
Only a performance benefit over dictionaries without a ``SetFactory``, but I will personally stare dumbfounded at anyone who manages to make this need maintenance.

PR #5335, opened 2026-08-23
[1.4.5] Fixed ModTile.AdjTiles not being applied to crafting station lookups

Reisenbug:
# PR Description

## What is the bug?

`ModTile.AdjTiles` has no effect. A modded crafting station declaring `AdjTiles = [TileID.WorkBenches]` is not treated as a workbench: standing next to it, recipes requiring a workbench stay unavailable.

1.4.5 moved tile equivalences into `Recipe.TileCountsAs`. Both `Player.SetAdjTile` and `NewCraftingUI.CraftStationRecipeFilter` resolve equivalences by recursing through that table. Nothing registers `ModTile.AdjTiles` into it, and the `TileLoader.AdjTiles(this, ...)` call in `Player.AdjTiles` was left commented out during the port:

```csharp
// TODO: Move to SetAdjTile: TileLoader.AdjTiles(this, Main.tile[j, k].type);
```

So `TileLoader.AdjTiles` had no call site at all.

## How did you fix the bug?

- Added `Recipe.SetupModdedTileInheritance()`, called right after `SetupTileInheritance()`, registering each `ModTile.AdjTiles` entry via `AddTileCountsAs`. Modded stations now use the same mechanism as vanilla equivalences such as `AddTileCountsAs(134, 16)` (Mythril Anvil counts as Anvil), so both `Player.SetAdjTile` and the crafting station filter UI pick them up.
- Restored the `TileLoader.AdjTiles(this, tileType)` call in `Player.SetAdjTile` and removed the TODO comment, as the porting note suggested.
- `TileLoader.AdjTiles` no longer iterates `ModTile.AdjTiles`, since `Recipe.TileCountsAs` now covers it. It still runs the `GlobalTile.AdjTiles` hook, which is queried per lookup and can vary at runtime.
- Documented the split on both members.

Removed the corresponding `PortingNotes_1.4.5.md` entry.

## Are there alternatives to your fix?

Restoring only the `TileLoader.AdjTiles` call, as the porting note's sub-item literally says, without touching `TileCountsAs`. That fixes `player.adjTile` so recipes become craftable, but `CraftStationRecipeFilter` reads `TileCountsAs` directly, so right clicking the station would still not list the equivalent station's recipes.

## Notes

Tested with ExampleWorkbench (`AdjTiles = [TileID.WorkBenches]`) placed away from any vanilla workbench. Before, workbench recipes were unavailable next to it; after, they are.

### DIFF
--- a/PortingNotes_1.4.5.md
+++ b/PortingNotes_1.4.5.md
@@ -43,8 +43,7 @@ Once all patches are fixed, these items need to be fixed or double checked:
 - TileLoader.HasWalkDust and WalkDust never pass in the Tile or Frame values. New vanilla logic sometimes checks frameX/Y, so hook could be updated with more parameters to facilitate.
 - Collision.SwitchTiles `objType` parameter in Player.Update changed from 1 to 5. Why, isn't it still Player? Adjust docs accordingly after investigating.
   - Seems to control MessageID.PlayerControls being sent but I don't know why. 1 is unused.
-- Adjacent tiles are now contained in Recipe.TileCountsAs rather than hard-coded. Need to adjust TileLoader.AdjTiles hooks/docs/exampels to prioritize using Recipe.TileCountsAs
-  - Restore `TileLoader.AdjTiles(this, Main.tile[j, k].type);` patch to new Player.SetAdjTile method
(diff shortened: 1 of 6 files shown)


Reisenbug:
Oops. I think #5334 should be merged first so this pr will be safe

PR #5358, opened 2026-08-24
[1.4.5] Fixed modded campfires not toggling on right click

Reisenbug:
### What is the bug?

Modded campfires do not toggle when right clicked. Toggling them with wire works, and vanilla campfires right click fine.

### How did you fix the bug?

`TileInteractionsUse` runs two separate statements. First the vanilla campfire branch in the else if chain:

```cs
else if (TileID.Sets.Campfires[tile.type]) {
    ...
    tile2.frameY += num23;
}
```

then, at the end of the method, outside that chain:

```cs
if (TileLoader.RightClick(myX, myY))
    flag2 = true;
```

`TileID.Sets.Campfires` is an open set, so a modded campfire enters the vanilla branch and gets its frameY flipped, then `ModTile.RightClick` flips it back. Wire toggling only goes through `HitWire`, so it flips once and works.

Excluded modded tiles from the vanilla branch with `&& tile.type < TileID.Count`, the same way the chest and torch branches in this chain already do.

No.

### DIFF
--- a/patches/tModLoader/Terraria/Player.cs.patch
+++ b/patches/tModLoader/Terraria/Player.cs.patch
@@ -5058,7 +5058,7 @@
  				for (int i = 0; i < pylons.Count; i++) {
  					TeleportPylonInfo info = pylons[i];
  					if (TeleportPylonsMapLayer.IsRevealed(info))
-@@ -25892,7 +_,7 @@
+@@ -25892,12 +_,15 @@
  				Wiring.HitSwitch(myX, myY);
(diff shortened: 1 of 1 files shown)


Chicken-Bones:
Did you test this with example campfire?

Reisenbug:
> Did you test this with example campfire?

Yes. I tested the example campfire and found this bug

JavidPack:
I've been looking into this and I think we might want to embrace the automatic `TileID.Sets.Campfires` logic. I'm still testing locally, but I think it makes sense. No need to update the PR.

Issue #5376, opened 2026-08-27
Separate the library mods from the regular mods


Sangar2728:
### Do you intend to personally contribute/program this feature?

No

### I would like to see this change made to improve my experience with

Mod capability as a Modder

### Description

Large mods increasingly require more mod libraries to function, but having them mixed in the menu with content mods creates confusion for players and overloads the mod list.

### What does this proposal attempt to solve or improve?

This could help modders use a wider ecosystem of libraries (their own/third-party) without the player feeling like they have 9 mods just for downloading one.

### Which (other) solutions should be considered?

I suggest a secondary tab for libraries, nothing sophisticated, just to visually separate them in the UI.

PR #5382, opened 2026-08-29
[1.4.4] Separate the library mods from the regular mods

bilimoing:
Closes [#5376](https://github.com/tModLoader/tModLoader/issues/5376)

### What is the new feature?

Adds a library mod filter to the Mods menu, allowing players to switch between showing all mods, only library mods, or only non-library mods.

Library mods are detected through either the local build property or the Steam Workshop Library tag. Modders can mark a local mod as a library mod by adding this to their build properties:

```properties
libMod = true
```

For Workshop mods, the Mods menu also reads Workshop tag data and treats mods with the Library tag as library mods.

Large mods increasingly depend on library/API mods. When those libraries are mixed together with regular content mods, the Mods menu becomes harder for players to understand and manage.

This feature addresses [#5376](https://github.com/tModLoader/tModLoader/issues/5376) by giving players a simple way to separate required library mods from the mods they usually interact with directly.

A separate Libraries tab was considered, but a filter toggle fits better with the existing Mods menu design. It reuses the same interaction pattern already used for enabled state, mod side, sorting, and search filters.

Another option would be to only use the local build property, but that would miss already-published Workshop mods that are categorized with the Steam Workshop Library tag. Supporting both sources lets existing Workshop library mods work without requiring every mod author to republish only for this change.

Players can open the Mods menu and use the new library filter toggle to choose one of these views:

- Show all mods.
- Show only library mods.
- Show only non-library mods.

A mod author can mark a mod as a library mod in the build file like this:

```properties
displayName = Example Library Mod
author = Example Author
version = 1.0
libMod = true
```

### ExampleMod updates

No ExampleMod updates are included. This change only affects the Mods menu filtering and library-mod classification logic.
image
image


patches/tModLoader/Terraria/Social/Steam/SteamedWraps.cs
```diff
@@ -18,6 +18,21 @@ namespace Terraria.Social.Steam;
...
+	private static readonly HashSet<string> LibraryTagNames = new(StringComparer.OrdinalIgnoreCase) {
+		LibraryTagInternalName,
+		LibraryTagSteamId,
+		"模组类库",
+		"Библиотека",
+		"Biblioteca",
+		"Biblioteka (lib)",
+		"Librerías",
+		"Programmbibliothek"
+	};
```
Solxanich: please double check we need this; 
At a glance, https://partner.steamgames.com/doc/api/ISteamUGC#GetQueryUGCTagDisplayName suggests that your implementation should only need to check `LibraryTagInternalName`

patches/tModLoader/Terraria/ModLoader/UI/UIMods.cs
```diff
@@ -41,6 +43,7 @@ internal class UIMods : UIState, IHaveBackButtonCommand
 	public ModsMenuSortMode sortMode = ModsMenuSortMode.RecentlyUpdated;
 	public EnabledFilter enabledFilterMode = EnabledFilter.All;
 	public ModSideFilter modSideFilterMode = ModSideFilter.All;
+	public ModLibraryFilter modLibraryFilterMode = ModLibraryFilter.All;
```
Solxanich: Set this default to ModLibraryFilter.NonLibraryOnly.

Solxanich (changes requested):
Thanks for the work getting this done quickly relative to the issue filed. 

Beyond the changes flagged, two general comments for when merging for tml team awareness:
1) Given the default will show the libraries as 'Filtered', I'm hoping that will still clue in people that they exist. As is, there isn't any code that auto-disables unused libraries so players still have to be able to find these to disable them.
2) Doing a test run with this PR, I do see several mods (like Spirit Reforged) for example that are using the Library tag. Will have to flag this in tml's porting notes to update tags used.

Solxanich (approved):

PR #5396, opened 2026-09-05
[1.4.5]Add ModPlayer.CanDoWireStuffHere And ModPlayer.CanShowWireStuffHere hook and add its example

hezuo1hao:
### What is the new feature?
It allows modders to modify whether players can see or place and destroy wires
Currently, achieving this functionality would require an IL edit, which is inconvenient and less maintainable. These hooks provide a cleaner, more accessible way to add custom logic or override vanilla wire interaction rules.
None
`ExampleCanDoAndShowWireStuffHerePlayer.cs` to demonstrate usage.
### ExampleMod updates
The ExampleMod has been updated with `ExampleCanDoAndShowWireStuffHerePlayer.cs` to showcase the new functionality.

### DIFF
--- a/ExampleMod/Common/Players/ExampleCanDoAndShowWireStuffHerePlayer.cs
+++ b/ExampleMod/Common/Players/ExampleCanDoAndShowWireStuffHerePlayer.cs
@@ -1,0 +1,36 @@
+﻿using ExampleMod.Content.Walls;
+using Terraria;
+using Terraria.ID;
+using Terraria.ModLoader;
+
+namespace ExampleMod.Common.Players
+{
+	// Showcases preventing the player from seeing or modifying wires under specific wall or condition.
+	public class ExampleCanDoAndShowWireStuffHerePlayer : ModPlayer
+	{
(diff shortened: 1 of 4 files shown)


ExampleMod/Common/Players/ExampleCanDoAndShowWireStuffHerePlayer.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+	// Showcases preventing the player from seeing or modifying wires under specific wall or condition.
+	public class ExampleCanDoAndShowWireStuffHerePlayer : ModPlayer
+	{
+		public override bool? CanDoWireStuffHere(int x, int y) {
+			Tile tile = Framing.GetTileSafely(x, y);
+
+			// Allows the player to modify wires within the Jungle Temple regardless of conditions.
+			if (tile.WallType == WallID.LihzahrdBrickUnsafe)
+				return true;
```
Chicken-Bones, line 16: probably should be commented

Chicken-Bones (changes requested):

PR #5408, opened 2026-09-08
[1.4.5] Migrate tModPorter Item.SetDefaults()

FreshinS:
### What is the bug?

In 1.4.5, `Item.SetDefaults(int Type = 0)` no longer has a default value for the `Type` parameter.

### How did you fix the bug?

I added a `ConvertItemSetDefaults` handler which:

- converts `Item.SetDefaults()` to `Item.TurnToAir()`
- preserves the existing migration which removes the obsolete `noMatCheck` parameter

Added regression tests covering both behaviors.

The parameterless migration could be implemented as a separate handler, but `InvokeRewriter` uses the first matching handler for a method, so keeping the `Item.SetDefaults` migrations in a single handler avoids overlapping registrations.

### DIFF
--- a/PortingNotes_1.4.5.md
+++ b/PortingNotes_1.4.5.md
@@ -188,8 +188,6 @@ Once all patches are fixed, these items need to be fixed or double checked:
 
 - ItemVariants.EverythingWorld renamed to MechdusaWorld
 - Main.GameModeInfo.IsJourneyMode -> Main.IsJourneyMode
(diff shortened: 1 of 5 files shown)


Chicken-Bones:
Yeah, okay, the proper migration is to change it to `TurnToAir`. Since it early exits if the item is already air. If you update this PR we can accept it

FreshinS:
I updated PR with migration to `TurnToAir()`

PR #5419, opened 2026-09-17
[1.4.5] Fix zh-Hans Localization

bilimoing:
The recent update caused all the TML Chinese localization to fail. I have fixed it.


JavidPack:
Fixed them all at once in https://github.com/tModLoader/tModLoader/commit/c45ae08d9d680e58c8a057b06c128741bf8f1d72 instead.

PR #5427, opened 2026-09-19
[1.4.5] Fixed input method display

bilimoing:
### What is the bug?
tModLoader does not correctly display the system IME candidate window when using CJK and other non-Latin input methods. This causes inconvenience for modders in multi-language regions when typing in-game text.
### How did you fix the bug?
Fixed window message handling to properly forward IME-related messages, allowing the system input method UI to display normally on the game window.
The only alternative is manually copying text from external editors, which is unfriendly and inefficient for non-English modders. This fix provides native global IME support for all mods.

image


bilimoing:
Of course, this might not be a bug in some way

JavidPack:
Note: This PR just fixes UIFocusInputTextField and UIInputTextField, which previously didn't display the IME candidate window. There are other remaining issues in pending PRs that fix other related issues.

Chicken-Bones:
I want to explore https://github.com/tModLoader/tModLoader/pull/5293 as a possible universal solution to these issues first

PR #5428, opened 2026-09-19
[1.4.5] Add Coat EquipType

bilimoing:
### What is the new feature?
Adds `EquipType.Coat`, `Item.coatSlot`, and public support for `Player.coat`.
Coat textures use the body composite layout and are drawn over equipped body armor. Optional lower extensions are automatically loaded using `_Coat_Front.png` and `_Coat_Back.png`.
Vanilla already uses the coat drawing system for Chicken Bones' Robe, but it was hardcoded and unavailable to mods. This exposes the existing system so mods can create accessories with the same layered coat behavior.
A custom `PlayerDrawLayer` could reproduce this behavior, but every mod would need to duplicate vanilla drawing and framing logic.
```csharp
[AutoloadEquip(EquipType.Coat)]
public class ExampleCoat : ModItem
{
		Item.accessory = true;
		Item.vanity = true;
	}
}
```
### ExampleMod updates
Added `ExampleCoat` with body, front extension, and back extension textures.
### Porting Notes
No action is required for existing mods.


bilimoing:
The coat texture originates from ChickenBones' developer vanity set.

PR #5429, opened 2026-09-19
[1.4.5] Fix bad invariant in MapHelper.TileToLookup

ppebb:
### What is the bug?
The invariant check in MapHelper.TileToLookup checks TileID.Count instead of tileLookup.Length which causes it to continuously fail.

### How did you fix the bug?
The tileType is checked against tileLookup.Length instead of TileID.Count

No.


oli-x64:
LGTM!

hakusaro (approved):
lgtm

direwolf420:
LTGM :rocket:

Mellozx:
LGTM* 🚀

PR #5430, opened 2026-09-20
[1.4.4] Fix + test message frame length deadlock

djfariel:
### What is the bug?

CheckBytes accepts frame lengths below the minimum three bytes. A zero-length frame prevents the framing loop from advancing, stalling packet processing.

### How did you fix the bug?

Reject lengths 0–2 before payload dispatch, terminate the affected connection, and clear its buffered input.

Add regression tests to the existing test project, exercising the checkout's framing implementation. All 47 tests passed, including 14 packet-framing cases.

This PR intentionally targets the 1.4.4 branch.

Breaking out of the loop alone would leave malformed input buffered. Rejecting the connection prevents it from being processed repeatedly. So, not really, no.

### Version scope

This PR targets tModLoader's 1.4.4 branch.

Vanilla Terraria 1.4.5.0 introduced minimum frame-length validation, which remains present in 1.4.5.8. Our vanilla 1.4.5.8 dedicated-server tests confirmed that malformed lengths 0–2 no longer stall the server.

The 1.4.4 branch lacks that guard. This change rejects those lengths and terminates the offending connection. In the vanilla 1.4.5.8 tests, the malformed input was discarded, but the connection remained open during the observation window.

Retaining connections after malformed input may allow offending clients to continue occupying connection slots or repeatedly trigger error handling and logging. Connection-slot exhaustion and log amplification were not tested, and existing timeout or connection-limit behavior may mitigate them. This PR disconnects the offending peer immediately; equivalent hardening for 1.4.5 would be a separate consideration.


djfariel:
FWIW I haven't been able to keep a server online for more than an hour since I posted this PR.

Chicken-Bones:
The fix looks fine. Delete the docs and test.

djfariel:
Done.

Chicken-Bones:
This fix is live on preview as `2026.08.2.2`, if you can confirm it works I can fast-forward it to stable

djfariel:
My server is up and running on `2026.08.2.2`. I had an incompatibility with Biome Colored Lava but that's a non-issue. My tests pass against the new server and I'll be leaving it up. I'll report back if it starts failing again, otherwise 24 hours or so.

